PR #245, opened 2016-11-29
Closes #243

### LINKED ISSUE #243: Automatic Compatibility with Divi Blog Module
Per this doc - http://docs.wp-rocket.me/article/657-divi-theme
Can we just auto-exclude that JS file from minification, it will prevent some unnecessary tickets.

Tabrisrp:
Exclude Salvattore script from JS minification

### DIFF
--- a/inc/3rd-party/themes/divi.php
+++ b/inc/3rd-party/themes/divi.php
@@ -17,7 +17,7 @@ if ( 'Divi' === $current_theme->get( 'Name' ) ) :
      */
     add_filter( 'rocket_exclude_js', 'rocket_exclude_js_divi' );
     function rocket_exclude_js_divi( $excluded_js ) {
-        $excluded_js[] = ET_BUILDER_URI . '/scripts/salvattore.min.js';
+        $excluded_js[] = str_replace( home_url(), '', ET_BUILDER_URI ) . '/scripts/salvattore.min.js';
 
         return $excluded_js;
     }
--- a/inc/3rd-party/themes/divi.php
+++ b/inc/3rd-party/themes/divi.php
@@ -17,7 +17,7 @@ if ( 'Divi' === $current_theme->get( 'Name' ) ) :
      */
     add_filter( 'rocket_exclude_js', 'rocket_exclude_js_divi' );
     function rocket_exclude_js_divi( $excluded_js ) {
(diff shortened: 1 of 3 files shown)


GeekPress:
Could we use this filter instead? 
rocket_excluded_handle_js

We will exclude the JS file from its slug and not from the URL uri. If the filename is updated for any reason, we will have to update WP Rocket too :/

Tabrisrp:
I tried that first, but the handle for the script is not available at that time because it's not loaded during the wp_enqueue_scripts action

GeekPress:
Is there a constant used by Divi to have the URI to the JS file? We "hack" the URL with "wp-content". If the WP_CONTENT_URL is used, the file won't be excluded.

inc/3rd-party/themes/divi.php
```diff
@@ -0,0 +1,24 @@
+<?php
+defined( 'ABSPATH' ) or die( 'Cheatin\' uh?' );
+
+$current_theme = wp_get_theme();
+
+if ( 'Divi' === $current_theme->get( 'Name' ) ) :
```
GeekPress: Could we check the ET_BUILDER_URI Divi constant instead?

Tabrisrp: defined( 'ET_BUILDER_URI' ) returns false, I don't know why. ET_BUILDER_URI alone works

inc/3rd-party/themes/divi.php
```diff
@@ -0,0 +1,24 @@
...
+     *
+     * @since 2.9
+     * @author Remy Perona
+     *
+     * @param Array $excluded_js An array of JS paths to be excluded
+     * @return Array the updated array of paths
+     */
+    add_filter( 'rocket_exclude_js', 'rocket_exclude_js_divi' );
+    function rocket_exclude_js_divi( $excluded_js ) {
+        $excluded_js[] = str_replace( home_url(), '', ET_BUILDER_URI ) . '/scripts/salvattore.min.js';
```
GeekPress: Could we check if the constant exists before to use it? If it's removed for any reasons, we will do the trick instead of having a fatal error ^^

inc/3rd-party/themes/divi.php
GeekPress: Ok, so we don't have the choice ^^

PR #250, opened 2016-12-05
Fixes #249

### LINKED ISSUE #249: Everything under CPT archive is cleared when updating a post of this CPT
Only the archive cache should be cleared and not everything in the folder

Tabrisrp:
Purge only the cache for the  CPT archive instead of the whole directory

### DIFF
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -139,7 +139,8 @@ function rocket_clean_post( $post_id ) {
 	// Add Post Type archive
 	if ( 'post' !== $post->post_type ) {
 	    if ( $post_type_archive = get_post_type_archive_link( get_post_type( $post_id ) ) ) {
-	    	array_push( $purge_urls, $post_type_archive );
+	    	array_push( $purge_urls, $post_type_archive . 'index.html' );
+	    	array_push( $purge_urls, $post_type_archive . 'index.html_gzip' );
 	    }
     }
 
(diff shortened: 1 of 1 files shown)


GeekPress:
What was the issue before? We create an other issue by trying to fix one: the archive pagination won't be cleared.

Tabrisrp:
If you purge the cache for a custom post - for example /project/test/ - it will clear everything under /project/, so clearing all posts of the CPT instead of the one to purge + CPT archive.

We can purge the pagination by adding one more URL to the purge (like we do for dates URLs with $GLOBALS['wp_rewrite']->pagination_base)

GeekPress:
Ok, I understand the reason, it was a big issue :/

Yes please add the purge of the pagination.

inc/common/purge.php
```diff
@@ -139,7 +139,8 @@ function rocket_clean_post( $post_id ) {
 	// Add Post Type archive
 	if ( 'post' !== $post->post_type ) {
 	    if ( $post_type_archive = get_post_type_archive_link( get_post_type( $post_id ) ) ) {
-	    	array_push( $purge_urls, $post_type_archive );
+	    	array_push( $purge_urls, $post_type_archive . 'index.html' );
```
GeekPress: What do you think about using trailingslashit() function to be sure that post_type_archive will always contains "/" at the end?

inc/common/purge.php
```diff
@@ -139,7 +139,9 @@ function rocket_clean_post( $post_id ) {
 	// Add Post Type archive
 	if ( 'post' !== $post->post_type ) {
 	    if ( $post_type_archive = get_post_type_archive_link( get_post_type( $post_id ) ) ) {
-	    	array_push( $purge_urls, $post_type_archive );
+	    	array_push( $purge_urls, trailingslashit( $post_type_archive ) . 'index.html' );
```
GeekPress: To avoid to use 3 times the same function, we can do before the first array_psuh:
$post_type_archive = trailingslashit( $post_type_archive );

PR #265, opened 2016-12-21
Closes #264

### LINKED ISSUE #264: Improve compatibility with WP Retina
Prevent the issue we had wit them removing a function

Tabrisrp:
Move WP Retina code to 3rd party and improve resilience

### DIFF
--- a/inc/3rd-party/wp-retina-2x.php
+++ b/inc/3rd-party/wp-retina-2x.php
@@ -1,6 +1,8 @@
 <?php
 
 if ( class_exists( 'WR2X_Admin' ) ) :
+	add_filter( 'wr2x_img_retina_url', 'rocket_cdn_on_images_from_wp_retina_x2' );
+	add_filter( 'wr2x_img_url', 'rocket_cdn_on_images_from_wp_retina_x2' );
 	/**
 	 * Conflict with WP Retina x2: Apply CDN on srcset attribute.
 	 *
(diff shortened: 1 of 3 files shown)


inc/3rd-party/wp-retina-2x.php
```diff
@@ -0,0 +1,27 @@
...
+	 * @since 2.5.5
+	 *
+	 * @param string $url URL of the image.
+	 * @param string Updated URL with CDN
+	 */
+	add_filter( 'wr2x_img_retina_url', 'rocket_cdn_on_images_from_wp_retina_x2' );
+	add_filter( 'wr2x_img_url', 'rocket_cdn_on_images_from_wp_retina_x2' );
+	function rocket_cdn_on_images_from_wp_retina_x2( $url ) {
+		global $wr2x_admin;
+		if ( $wr2x_admin->is_pro() ) {
```
GeekPress: Check before if the method exists with method_exists() function to avoid php error if the method doesn't exist in the future

inc/front/plugin-compatibility.php
```diff
@@ -158,25 +158,6 @@ function __rocket_cdn_on_sliders_with_lazyload() {
 }
 
 /**
- * Conflict with WP Retina x2: Apply CDN on srcset attribute.
- *
- * @since 2.5.5
- */
-add_filter( 'wr2x_img_retina_url', '__rocket_cdn_on_images_from_wp_retina_x2' );
-add_filter( 'wr2x_img_url', '__rocket_cdn_on_images_from_wp_retina_x2' );
-function __rocket_cdn_on_images_from_wp_retina_x2( $url ) {
```
GeekPress: Don't forget to deprecated this function __rocket_cdn_on_images_from_wp_retina_x2() as it's renamed rocket_cdn_on_images_from_wp_retina_x2() without __ at the beginning

Tabrisrp: Most probably yes, we need  to remove them all completely

PR #290, opened 2017-01-06
Fixes #286

### LINKED ISSUE #286: Fatal error with Move Login 2.4
In Move Login 2.4, released a few days ago, I changed a method: `SFML_Options::get_slugs()` is not static anymore, and should be called `SFML_Options::get_instance()->get_slugs()` instead.
Calling it as a static method triggers a fatal error with php 7. It is used in `get_rocket_logins_exclude_pages()`.

In the mean time, the version 2.4 uses the constant `DONOTCACHEPAGE` now, so there is no need to add ML slugs in the exclusion list anymore.

Sorry, this one is my fault, I haven't realized we used it in Rocket :/

Tabrisrp:
Update compatibility with SF Move Login.

No longer exclude the URLs for version 2.4 and above since it's using DONOTCACHEPAGE constant.

### DIFF
--- a/inc/3rd-party/plugins/sf-move-login.php
+++ b/inc/3rd-party/plugins/sf-move-login.php
@@ -16,22 +16,28 @@ endif;
  * @return array Updated array of URLs
  */
 function rocket_add_sfml_exclude_pages( $urls ) {
-	if ( defined( 'SFML_PLUGIN_DIR' ) ) { 
-		if ( ! class_exists( 'SFML_Options' ) ) {
+	if ( ! function_exists( 'sfml_get_slugs' ) ) {
+		if ( file_exists( SFML_PLUGIN_DIR . 'inc/utilities.php' ) ) {
 			include( SFML_PLUGIN_DIR . 'inc/utilities.php' );
-			include( SFML_PLUGIN_DIR . 'inc/class-sfml-options.php' );
+		} else {
+			return $urls;
 		}
-		
-		$sfml_slugs = SFML_Options::get_slugs();
-
-		foreach( $sfml_slugs as $slug ) {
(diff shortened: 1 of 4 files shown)


inc/3rd-party/plugins/sf-move-login.php
```diff
@@ -0,0 +1,71 @@
...
+
+		foreach( $sfml_slugs as $slug ) {
+			$sfml_urls[] = rocket_clean_exclude_file( trailingslashit( home_url( $slug ) ) );
+		}
+
+		$urls = array_merge( $urls, $sfml_urls );
+	}
+
+	return $urls;
+}
```
Screenfeed: My proposal:

	function rocket_add_sfml_exclude_pages( $urls ) {
		if ( ! function_exists( 'sfml_get_slugs' ) ) {
			if ( file_exists( SFML_PLUGIN_DIR . 'inc/utilities.php' ) ) {
				include( SFML_PLUGIN_DIR . 'inc/utilities.php' );
			} else {
				return $urls;
			}
		}

		if ( ! class_exists( 'SFML_Options' ) && ! defined( 'SFML_NOOP_VERSION' ) ) {
			if ( file_exists( SFML_PLUGIN_DIR . 'inc/class-sfml-options.php' ) ) {
				include( SFML_PLUGIN_DIR . 'inc/class-sfml-options.php' );
			} else {
				return $urls;
			}
		}

		$sfml_slugs = sfml_get_slugs();
		$sfml_slugs = array_map( 'home_url', $sfml_slugs );
		$sfml_slugs = array_map( 'trailingslashit', $sfml_slugs );
		$sfml_slugs = array_map( 'rocket_clean_exclude_file', $sfml_slugs );

		return array_merge( $urls, $sfml_slugs );
	}

Screenfeed:
OK for me ;)

Issue #296, opened 2017-01-11
All GET parameters get cached on Nginx


junaidbhura:
Even if the field under **Advanced Options -> Cache pages that use the following query strings (GET parameters)** is left completely blank, WP Rocket still caches **all** GET variables on an Nginx server.

The problem is in the file /wp-rocket/inc/front/process.php which does not add the correct exception for Nginx, causing the condition to exclude GET parameters to fail everytime.

junaidbhura:
An update: This was caused due to an additional query parameter ?q= being passed with all requests, which was set on our Nginx configuration:

`try_files $uri $uri/ /index.php?q=$uri&$args;`

Once this is changed to:

`try_files $uri $uri/ /index.php?$args;`

The problem goes away!

The configuration with the "?q=$uri" was an older configuration, so perhaps this issue and pull request should still be implemented to account for websites running the older configuration?

Tabrisrp:
The caching of the `q` parameter on NGINX is a deliberate choice we made. Without it, there was no caching happening if the NGINX configuration uses `try_files $uri $uri/ /index.php?q=$uri&$args;`, because URLs with query strings are excluded by default.

As an alternative solution, we can define the `q` parameter as a parameter to ignore when parsing the URL, and that should prevent URLs with other query strings from being cached.

arunbasillal:
Related discussion - https://wp-media.slack.com/archives/G8L4EE8N9/p1577117490016100
Current compatibility is included here - https://github.com/wp-media/wp-rocket/blob/master/inc/3rd-party/hosting/nginx.php

GeekPress:
I'm closing this one as the `q` parameter isn't used anymore on NGINX configurations and we auto-exclude it anyway.

DahmaniAdame:
Related - https://secure.helpscout.net/conversation/1711029079/311308/ (nothing we can do, just for the record)

PR #297, opened 2017-01-11
Fix Nginx GET parameters caching

junaidbhura:
Even if the field under **Advanced Options -> Cache pages that use the following query strings (GET parameters)** is left completely blank, WP Rocket still caches **all** GET variables on an Nginx server.

This fixes: #296

### DIFF
--- a/inc/front/process.php
+++ b/inc/front/process.php
@@ -72,14 +72,18 @@ if ( ! $continue ) {
 	return;
 }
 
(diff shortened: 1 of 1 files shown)


Tabrisrp:
Hi!

I will be closing this PR, as it's modifying a file that is now deprecated. `process.php` is not used anymore, the system was split into multiple classes to handle cache-related code.

PR #300, opened 2017-01-12
Closes #288

### LINKED ISSUE #288: Add filters to remove query string and static file creation
It's needed to be able to exclude files from those as it causes issues with some particular themes/plugins using wp_enqueue_style() or wp_enqueue_script() in unexpected ways.

Tabrisrp:
Add filters to cache busting and static dynamic resources to excludefiles from the process

### DIFF
--- a/inc/front/enqueue.php
+++ b/inc/front/enqueue.php
@@ -165,8 +165,9 @@ function rocket_cache_dynamic_resource( $src ) {
 	 *
 	 * @param array $excluded_files An array of filepath to exclude.
 	 */
-	$excluded_files = apply_filters(  'rocket_exclude_static_dynamic_resources', array( '/wp-admin/admin-ajax.php' ) );
-	$excluded_files = array_flip( $excluded_files );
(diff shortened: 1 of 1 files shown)


inc/front/enqueue.php
```diff
@@ -33,6 +33,13 @@ function rocket_browser_cache_busting( $src, $current_filter = '' ) {
     if ( false === strpos( $src, '.css' ) && false === strpos( $src, '.js' ) ) {
         return $src;
     }
+
+	$excluded_files = apply_filters(  'rocket_exclude_cache_busting', array() );
```
GeekPress: Add filter comment here

inc/front/enqueue.php
```diff
@@ -142,6 +149,13 @@ function rocket_cache_dynamic_resource( $src ) {
         return $src;
     }
 
+	$excluded_files = apply_filters(  'rocket_exclude_static_dynamic_resources', array( '/wp-admin/admin-ajax.php' ) );
```
GeekPress: Add filter comment here too :)

PR #323, opened 2017-01-25
Disable embeds feature

Tabrisrp:


inc/common/embeds.php
```diff
@@ -0,0 +1,81 @@
...
+if ( version_compare( $wp_version, '4.4' ) >= 0 && get_rocket_option( 'embeds', 0 ) ) :
+
+	/**
+	 * Disable embeds on init.
+	 *
+	 * - Removes the needed query vars.
+	 * - Disables oEmbed discovery.
+	 * - Completely removes the related JavaScript.
+	 *
+	 * @since 1.0.0
```
GeekPress: @since 3.0 instead of 1.0.0?

GeekPress:
@Tabrisrp Could you add the @author on the PHP comments?

PR #330, opened 2017-01-30
Closes #329

### LINKED ISSUE #329: Improve compatibility with Autoptimize
Instead of displaying a notice saying it's not compatible, we'll reflect Autoptimize configuration in WP Rocket. If an option is selected in Autoptimize, we unselect it in WP Rocket and gray-out the field

Tabrisrp:
Autoptimize compatibility improvement


GeekPress:
This PR is weird, there are other files in the commit. is it normal?

Tabrisrp:
it should be fine, it's because I selected the wrong base branch initially

PR #334, opened 2017-02-02
Fixes #304 + minify without concatenation

### LINKED ISSUE #304: JS & CSS minification should reflect uncached position of enqueued scripts & styles and not $wp_scripts->in_footer array
Hey guys,

In `inc/front/minify.php` on line `576` `$wp_scripts->in_footer` is used to loop trough all js files that were enqueued in footer, or better, enqueued with the `$in_footer` parameter set to `true` when `wp_enqueue_scripts()` is used in a theme or plugin.

That doesn't reflect the uncached situation in many cases. For example, you can use the following theme code to make sure built-in WordPress core scripts are outputted in footer trough the `wp_footer` action:

```
// when executed in the init or after_theme_setup hook
remove_action( 'wp_head', 'wp_print_scripts' );
remove_action( 'wp_head', 'wp_print_head_scripts', 9 );
remove_action( 'wp_head', 'wp_enqueue_scripts', 1 );

add_action( 'wp_footer', 'wp_print_scripts', 5 );
add_action( 'wp_footer', 'wp_enqueue_scripts', 5 );
add_action( 'wp_footer', 'wp_print_head_scripts', 5 );
```

When using the above code in your theme, you will see that all core scripts normally enqueued in head, are now placed in footer. This is a common performance practice.

**Therefor: the minification process should minify js files on the location that they are outputted when uncached. So that the cached situation reflects the uncached one.**

In my point of view, it could be a good starting point to look at the `done` array in the `$wp_scripts` variable right after both the `wp_head` and `wp_footer` action. The `done` array represents the script handles that were in queue before and got outputted in either of those hooks. It seems to me that that could be the perfect timing on which to determine which js files should be minified at that position.

**EDIT: Exactly the same could be considered for the enqueued styles also, using the `$wp_styles` global variable and using the `done` array of that var too.**

Good luck guys, keep it up!

// T

Tabrisrp:


inc/front/minify.php
```diff
@@ -314,6 +387,70 @@ function rocket_minify_css( $buffer )
...
+ * Minify JS files without concatenation.
+ *
+ * @since 3.0
+ * @author Remy Perona
+ *
+ * @param string $buffer HTML code to parse.
+ * @return string Updated HTML code
+ */
+function rocket_minify_js_only( $buffer ) {
+	// Get host of CNAMES.
```
GeekPress: Could we create a function for the lines from 399 to 413. We use exactly the same code for the CSS part. With an argument (css or js in value) to use the right exclusion function.

inc/functions/minify.php
```diff
@@ -4,179 +4,268 @@
...
-			rocket_mkdir_p( $cache_path );
-		}
-		
-		// Apply CDN on CSS properties
-		if( strrpos( $pretty_path, '.css' ) ) {
-			$content = rocket_cdn_css_properties( $content );	
-		}
+	$minify_result = wp_safe_remote_get( $url );
+
+	if ( 200 !== wp_remote_retrieve_response_code( $minify_result ) ) {
```
GeekPress: Check before if the response isn't a WP Error.

Tabrisrp: is_wp_error() is included in wp_remote_retrive_response_code https://developer.wordpress.org/reference/functions/wp_remote_retrieve_response_code/

PR #354, opened 2017-02-21
Load CSS asynchronously + critical css

Tabrisrp:


inc/admin/options.php
```diff
@@ -741,6 +746,26 @@ function rocket_settings_callback( $inputs ) {
 	}
 
 	/*
+	 * Option : critical CSS content for async CSS
```
GeekPress: If we don't use it anymore, we should remove it, nope?

Tabrisrp: We still need the field to input the critical css, I only removed the automatic generation of it

GeekPress: I'm talking about the commented PHP line in fact ^^

Tabrisrp: oops it should not be, will fix

PR #368, opened 2017-03-06
Fixes #156

### LINKED ISSUE #156: Imagify install button doesn't work since WP 4.6
Reference ticket on the script repo: https://github.com/thomasgriffin/tgm-plugin-update-modal/issues/2

Tabrisrp:


GeekPress:
Does it work on WordPress < 4.7? If not, we must have a fallback to be also able to install Imagify on previous WordPression versions, not only since 4.7

Tabrisrp:
Tested on WP 4.5.7 (so before shiny updates integration). It works without any fallback necessary, it's just not as fancy (no ajax)

PR #374, opened 2017-03-07
WP coding standards

Tabrisrp:


Tabrisrp:
There is a remaining number of warnings/errors we need to discuss I think to see if we fix them or not and how

uninstall.php
```diff
@@ -1,51 +1,48 @@
...
-		return;	
+		rocket_direct_filesystem()->delete( $dir );
+		return;
 	}
 
-    if ( $globs = glob( $dir . '/*', GLOB_NOSORT ) ) {
+	if ( $globs = glob( $dir . '/*', GLOB_NOSORT ) ) {
 	    foreach ( $globs as $file ) {
-			is_dir( $file ) ? __rocket_rrmdir($file) : @unlink( $file );
+			is_dir( $file ) ? rocket_rrmdir( $file ) : rocket_direct_filesystem()->delete( $file );
```
GeekPress: It must be rocket_uninstall_rrmdir instead of rocket_rrmdir?

uninstall.php
```diff
@@ -1,51 +1,48 @@
...
- * @param string $dir Directory path to remove.
+ * @param string $dir Path to directory.
  */
-function __rocket_rrmdir( $dir ) {
+function rocket_uninstall_rrmdir( $dir ) {
 
 	if ( ! is_dir( $dir ) ) {
-		@unlink( $dir );
-		return;	
+		rocket_direct_filesystem()->delete( $dir );
```
GeekPress: Are you sure that function is available here? The the plugin is already deactivated, it musn't be available.

Tabrisrp: You're right. So we have two solutions:

- Revert to unlink
- Duplicate the function for the filesystem to keep with the coding standards

Which do you prefer?

inc/functions/options.php
```diff
@@ -558,28 +558,32 @@ function get_rocket_exclude_async_css() {
...
 /**
  * Determine if the key is valid
  *
  * @since 2.2 The function do the live check and update the option
+ *
+ * @param string $type Transient type.
+ * @param string $data Data.
+ * @return array WP Rocket options array
  */
 function rocket_check_key( $type = 'transient_1', $data = null ) {
```
GeekPress: This function has been updated in the 2.9.7. The content of this one is the old content.

uninstall.php
```diff
@@ -1,51 +1,48 @@
...
-
-// If uninstall not called from WordPress exit
 defined( 'WP_UNINSTALL_PLUGIN' ) or die( 'Cheatin&#8217; uh?' );
 
-// Delete all transients
+// Delete all transients.
 delete_site_transient( 'update_wprocket' );
 delete_site_transient( 'update_wprocket_response' );
 delete_transient( 'wp_rocket_settings' );
 delete_transient( 'rocket_check_licence_30' );
```
GeekPress: This has been deleted in 2.9.7. It seems that this branch isn't updated with the latest master content.

uninstall.php
GeekPress: We can revert to unlink

PR #415, opened 2017-05-17
UX enhancements

glueckpress:
Finished all the tabs. Left minor string issues (like untranslated “e.g.” in placeholders), but I think this you go as another internal beta.

### DIFF
--- a/inc/admin/ui/modules/varnish.php
+++ b/inc/admin/ui/modules/varnish.php
@@ -8,8 +8,9 @@ defined( 'ABSPATH' ) or die( 'Cheatin&#8217; uh?' );
 	*
 	* @param bool true will display the "Varnish" tab
  */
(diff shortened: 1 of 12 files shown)


inc/admin/ui/modules/varnish.php
```diff
@@ -8,34 +8,56 @@
 	*
 	* @param bool true will display the "Varnish" tab
  */
-if ( apply_filters( 'rocket_display_varnish_options_tab', true ) ) :
+if ( ! apply_filters( 'rocket_display_varnish_options_tab', true ) )
+	return false;
```
Tabrisrp: always use brackets even for one line, coding standards ☝️

glueckpress: Oh, cool. 👍

PR #418, opened 2017-05-22
White-label support + enhanced license validation panel

glueckpress:
- removed all branding (incl. docs links) from options when WL is enabled
- enhanced license validation panel

### DIFF
--- a/inc/admin/options.php
+++ b/inc/admin/options.php
@@ -572,10 +572,8 @@ function rocket_display_options() {
 				<?php if ( ! rocket_is_white_label() ) { ?>
 					<a href="#tab_support" class="nav-tab"><?php _e( 'Support', 'rocket' ); ?></a>
 				<?php } ?>
-			<?php }else { ?>
+			<?php } else { ?>
 				<a href="#tab_apikey" class="nav-tab"><?php _e( 'License', 'rocket' ); ?></a>
-				<a href="#tab_faq" class="nav-tab"><?php _e( 'FAQ', 'rocket' ); ?></a>
-				<a href="#tab_support" class="nav-tab file-error"><?php _e( 'Support', 'rocket' ); ?></a>
 			<?php }  ?>
 			<?php
 			do_action( 'rocket_tab', rocket_valid_key() );
--- a/inc/admin/options.php
+++ b/inc/admin/options.php
@@ -885,8 +885,8 @@ function rocket_settings_callback( $inputs ) {
 		$inputs['cloudflare_domain'] = '';
 	}
 
-	$inputs['cloudflare_devmode'] = ( is_numeric( $inputs['cloudflare_devmode'] ) ) ? (int) $inputs['cloudflare_devmode'] : 0;
-	$inputs['cloudflare_auto_settings'] = ( is_numeric( $inputs['cloudflare_auto_settings'] ) ) ? (int) $inputs['cloudflare_auto_settings'] : 0;
+	$inputs['cloudflare_devmode'] = ( isset( $inputs['cloudflare_devmode'] ) && is_numeric( $inputs['cloudflare_devmode'] ) ) ? (int) $inputs['cloudflare_devmode'] : 0;
+	$inputs['cloudflare_auto_settings'] = ( isset( $inputs['cloudflare_auto_settings'] ) && is_numeric( $inputs['cloudflare_auto_settings'] ) ) ? (int) $inputs['cloudflare_auto_settings'] : 0;
 
 	/*
 	 * Option : WL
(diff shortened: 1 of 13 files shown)


glueckpress:
@GeekPress The support does not seem to send anything when the license is not active, that’s why I removed the support tab there. I tested multiple times, only saw a perpetuating sending animation, but my message did not get transmitted.

glueckpress:
@GeekPress Hold on, it may have been my bad. I had actually removed the license data from my install. That would hardly be the case for a customer, admittedly. 

Can we confirm the support form works even when the connection to our servers does not work for license validation?

inc/admin/options.php
```diff
@@ -885,8 +883,8 @@ function rocket_settings_callback( $inputs ) {
 		$inputs['cloudflare_domain'] = '';
 	}
 
-	$inputs['cloudflare_devmode'] = ( is_numeric( $inputs['cloudflare_devmode'] ) ) ? (int) $inputs['cloudflare_devmode'] : 0;
-	$inputs['cloudflare_auto_settings'] = ( is_numeric( $inputs['cloudflare_auto_settings'] ) ) ? (int) $inputs['cloudflare_auto_settings'] : 0;
+	$inputs['cloudflare_devmode'] = ( isset( $inputs['cloudflare_devmode'] ) && is_numeric( $inputs['cloudflare_devmode'] ) ) ? (int) $inputs['cloudflare_devmode'] : 0;
+	$inputs['cloudflare_auto_settings'] = ( isset( $inputs['cloudflare_auto_settings'] ) && is_numeric( $inputs['cloudflare_auto_settings'] ) ) ? (int) $inputs['cloudflare_auto_settings'] : 0;
```
Tabrisrp: this should alway be set, did you encounter a case that require this additional check?

inc/admin/ui/modules/cdn.php
```diff
@@ -137,7 +141,7 @@
 			'type'         => 'textarea',
 			'label_for'    => 'cdn_reject_files',
 			'label_screen' => __( 'Exclude files:', 'rocket' ),
-			'placeholder'  => "e.g. /wp-content/themes/some-theme/style.css\ne.g. /wp-content/plugins/some-plugin/(.*).css"
+			'placeholder'  => "/wp-content/themes/some-theme/style.css\n/wp-content/plugins/some-plugin/(.*).css"
```
Tabrisrp: \n doesn't work in the placeholder

inc/admin/ui/modules/optimization.php
```diff
@@ -202,7 +216,7 @@
 			'type'         => 'textarea',
 			'label_for'    => 'exclude_css',
 			'label_screen' => __( 'CSS files to be excluded from minification and concatenation:', 'rocket' ),
-			'placeholder'  => "e.g. /wp-content/themes/some-theme/style.css\ne.g. /wp-content/plugins/some-plugin/(.*).css",
+			'placeholder'  => "/wp-content/themes/some-theme/style.css\n/wp-content/plugins/some-plugin/(.*).css",
```
Tabrisrp: Same here \n doesn't work

inc/admin/ui/modules/optimization.php
```diff
@@ -233,7 +247,7 @@
 			'type'         => 'textarea',
 			'label_for'    => 'exclude_js',
 			'label_screen' => __( 'JS files to be excludeded from minification and concatenation:', 'rocket' ),
-			'placeholder'  => "e.g. /wp-content/themes/some-theme/(.*).js\ne.g. /wp-content/plugins/some-plugin/script.js",
+			'placeholder'  => "/wp-content/themes/some-theme/(.*).js\n/wp-content/plugins/some-plugin/script.js",
```
Tabrisrp: + here

inc/admin/options.php
glueckpress: Yes, when my CF credentials when I saved CF settings with a wrong API key and wrong email, I got a warning without those `isset()`s.

inc/admin/ui/modules/cdn.php
glueckpress: [MDN reference](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) wasn’t very clear about it, but [spec is](https://www.w3.org/TR/html5/forms.html#the-placeholder-attribute), so changing it.

glueckpress: Done in https://github.com/wp-media/wp-rocket/pull/418/commits/6696b048b0799dbd6882b820ee493ae18814fa38

inc/admin/ui/modules/optimization.php
glueckpress: Done in https://github.com/wp-media/wp-rocket/pull/418/commits/6696b048b0799dbd6882b820ee493ae18814fa38

inc/admin/ui/modules/optimization.php
glueckpress: Done in https://github.com/wp-media/wp-rocket/pull/418/commits/6696b048b0799dbd6882b820ee493ae18814fa38

PR #420, opened 2017-05-24
Jetpack: add compatibilty functions for the EU Cookie Widget.

jeherve:
We'd like to add Jetpack's [EU Cookie Law WIdget](https://jetpack.com/support/extra-sidebar-widgets/eu-cookie-law-widget/) to this list:
http://docs.wp-rocket.me/article/687-which-cookie-law-plugins-are-compatible-with-wp-rocket

I'm not super familiar with the way WP Rocket handles Cookies, but looked at how other third-party plugins were handled in your plugin, and created this quick PR. I would appreciate a review!

This conflict was reported here:
https://wordpress.org/support/topic/coockie-banner-issue-maybe-cause-by-cache-plugin/


inc/3rd-party/jetpack.php
```diff
@@ -1,56 +1,124 @@
...
+			add_filter( 'rocket_htaccess_mod_rewrite'    , '__return_false' );
+			add_filter( 'rocket_cache_mandatory_cookies' , 'rocket_add_jetpack_cookie_law_mandatory_cookie' );
+
+			// Update the WP Rocket rules on the .htaccess file.
+			flush_rocket_htaccess();
+
+			// Regenerate the config file.
+			rocket_generate_config_file();
+		}
+		add_action( 'activate_jetpack/jetpack.php', 'rocket_activate_jetpack_cookie_law' );
```
jeherve: This isn't great as I'd like this code to run once, either when the widget is added, or when someone activates WP Rocket while our widget is already active, or when both are already active and they update to the next version of WP Rocket. I'll try to come up with something better.

jeherve:
This should now be ready for review. Thank you!

jeherve:
> Remove any underscore or double underscores prefixing functions, we're moving away from that

Done in 74a49fef6e239fc5f0192d996e15548f17c92373

> Add the author to any new function comment block

Done in 6aa821cdf59a3b25d1ce1aa3425f9fb5048fd1d3

> For information we won't merge it directly into the master

👍  Let me know if I should rebase this branch against a specific release branch and submit a new PR!

PR #436, opened 2017-06-09
i18n: minor string update, added missing text domains, fixed RTL support

glueckpress:
This branch fixes missing text domains on a few strings and will enable localisers to complete their translations.

It also fixes some misplaced icons in RTL setups.


Tabrisrp:
Can you also update any occurence of "in case" by "if" as a customer suggested (see contact mailbox)

GeekPress:
As we don't display anymore the FAQ page, we must deleted the file which contains all FAQ's questions.

PR #460, opened 2017-06-26
Closes #452

### LINKED ISSUE #452: Auto-exclude purchase confirmation page from EDD
the title says it all :)

Tabrisrp:


inc/functions/plugins.php
```diff
@@ -177,6 +177,11 @@ function get_rocket_ecommerce_exclude_pages() {
 		$urls = array_merge( $urls, $checkout_urls );
 	}
 
+	if ( function_exists( 'EDD' ) && isset( $edd_settings['success_page'] ) ) {
```
Screenfeed: I would use `! empty()` instead of `isset()`. Same for L175.

PR #466, opened 2017-07-04
Fixes #462

### LINKED ISSUE #462: Cache doesn't work on Windows server
config path doesn't match in `stripos()` because of `realpath()` changing backslash to anti-backslash on Windows server.

We need to do a `realpath()` on the second parameter too to be sure separators match.

Tabrisrp:


Screenfeed:
I see 2 "problems" here:
1. Instead of doing multiple `realpath( $rocket_config_path )`, it would be simpler to add a single `$rocket_config_path = realpath( $rocket_config_path );` somewhere.
2. `realpath()` removes the trailing slash, so I think the test will pass with the following values:
```php
$rocket_config_path = '/public_html/wp-content/wp-rocket-config/';
$host = '../wp-rocket-config-hacked';
```
So imho, the solution is:
```php
$rocket_config_path = realpath( $rocket_config_path ) . DIRECTORY_SEPARATOR;
```
Unless I'm drunk, `realpath()` will use `\` on Windows, that's why I use `DIRECTORY_SEPARATOR`. Do you agree for all of this? (We should look at the whole code to make sure it won't create more problems in the following lines.)

Tabrisrp:
updated the code, what do  you think of this version?

Screenfeed:
That's exactly what I suggested so of course I think it's good x)

Issue #78, opened 2016-07-06
Serve mobile cache for tablet for compatibility with wp_is_mobile()


jbma:
We could add a filter in process.php to allow customers who want to serve the mobile cache to tablets too

arunbasillal:
Now that [wp_is_mobile()](https://codex.wordpress.org/Function_Reference/wp_is_mobile) is out of beta and its officially recommended for use in themes and plugins, WP Rocket [not treating tablets as mobiles](https://docs.wp-rocket.me/article/708-mobile-caching#what-does-mobile-mean-for-wp-rocket) is going to be an issue. 

`wp_is_mobile()` treats tablets as mobiles.

21stcn:
For the record, I was baffled by this functionality after setting up my caches and moving on to compatibility testing only to find tablets were being served non-mobile cached files. To reiterate my (rather frustrated) concerns:

> It seems absurd that the plugin would not use the same checks as the wordpress wp_is_mobile function. Developers build their contingencies using this check. Mobiles generally have touch, they don't autoplay video, etc. In fact the wp_is_mobile function literally identifies an ipad by checking the HTTP_USER_AGENT for "Mobile". This means that apple have literally defined their tablets as mobile devices in the device user agent strings.

`And now my calculated division between mobile and desktop, as defined by the hardware manufacturers and native wordpress is function, is undermined by your plugin because you don't define tablets as mobile, unlike every other standard on the web. So my site now loads hover effects and autoplay videos which dont play on an ipad, and the whole process is a lot slower than if it has just loaded the correct version, even uncached. `

In any case, I changed the following, line  187, in process.php

`if ( $detect->isMobile() && ! $detect->isTablet() )`

to 

`if ( $detect->isMobile() || $detect->isTablet() )`

So far so good. I'm aware of the pitfalls of editing plugin core files,  but honestly the other options are not caching for a huge quantity of devices, or redeveloping my entire site to treat tablets as non-mobile, pointer, hover-enabled, large screen devices which isn't going to happen.

arunbasillal:
There were some internal discussions on this matter, for team mates, the details can be found here: https://wp-media.slack.com/archives/CBJ2KQ0MU/p1536151743000100

To summarize, what is preventing us from simply switching is the fact that some mobile plugins does treat tablets as desktops. So we need to find a way to accommodate them and not break their websites after an update. 

There is always a way.

arunbasillal:
[...]

PR #472, opened 2017-07-11
Check wp-content/advanced-cache-mobile-detect.php for custom mobile d…

foliovision:
#78 

Modification to allow users to define their own mobile detection in wp-content/advanced-cache-mobile-detect.php.

Very powerful as the function can return any string, so there can be a separate caching group for tablet devices.


Tabrisrp:
Hi!

I will be closing this PR, as it's modifying a file that is now deprecated. process.php is not used anymore, the system was split into multiple classes to handle cache-related code.

The related issue #78 was also closed with version 3.2, bringing a solution to choose which cache file tablets will receive.

Any feedback on the solution, and additional improvements you might need is welcome.

PR #473, opened 2017-07-19
Fixes #157

### LINKED ISSUE #157: Add a filter to keep empty lines in the htaccess file
A customer requested that we don't remove empty lines when writing in the htaccess file. We could add a filter in htaccess.php to allow that

Tabrisrp:
add filter to determine if empty lines should be kept in htaccess


Tabrisrp:
Yes config is not correct for stickler I need to correct it

PR #474, opened 2017-07-19
Fixes #369

### LINKED ISSUE #369: Apply CDN URL on internal images links
The CDN URL could be applied on links to internal images (like full size images link on a thumbnail for example)

Tabrisrp:
Add images link to CDN rewrite


inc/front/cdn.php
```diff
@@ -269,7 +269,7 @@ function rocket_cdn_custom_files( $html ) {
 		 *
 		 * @param array $filetypes Array of file types.
 		 */
-		$filetypes = apply_filters( 'rocket_cdn_custom_filetypes', array( 'mp3', 'ogg', 'mp4', 'm4v', 'avi', 'mov', 'flv', 'swf', 'webm', 'pdf', 'doc', 'docx', 'txt', 'zip', 'tar', 'bz2', 'tgz', 'rar' ) );
+		$filetypes = apply_filters( 'rocket_cdn_custom_filetypes', array( 'mp3', 'ogg', 'mp4', 'm4v', 'avi', 'mov', 'flv', 'swf', 'webm', 'pdf', 'doc', 'docx', 'txt', 'zip', 'tar', 'bz2', 'tgz', 'rar', 'jpg', 'jpeg', 'png', 'gif', 'webp', 'bmp', 'tiff' ) );
```
Screenfeed: What about jpe?

Tabrisrp: Added you see more than should be?

Screenfeed: Erf, we also forgot `ico`.
OK, a good place to start is here: https://wpseek.com/function/wp_get_mime_types/
Maybe we can use it? For example:

```php
$filetypes = wp_get_mime_types();
$filetypes = array_keys( $filetypes );
$filetypes = implode( '|', $filetypes );
$filetypes = explode( '|', $filetypes );
```

The only problem is the `'mime_types'` filter, that can be used to add or remove types :/
At least you can look at the mime types it uses, and complete our list.

Tabrisrp: I'm trying to cover common filetypes with the default values, if customers need more the filter is there for that. I don't think ico is really needed since we're matching `<a href` only here.

Screenfeed: Ha, ok then, go for it.

PR #476, opened 2017-07-20
Closes #404

### LINKED ISSUE #404: Cache doesn't clear when ACF Options page is updated
ACF Pro allows us to create an Options page, but when this page is updated the Rocket cache isn't cleared. It can be done with a simple function (my workaround is below) but would be good to see this in the plugin.

```
add_action('acf/save_post', 'clear_cache_on_options_save');
function clear_cache_on_options_save($post_id) {
    if ($post_id == 'options') {
        rocket_clean_domain();
    }
}

### PR DESCRIPTION
**Pull request by Tabrisrp**
Clear cache when ACF options page is updated

### DIFF
--- a/inc/3rd-party/plugins/advanced-custom-fields.php
+++ b/inc/3rd-party/plugins/advanced-custom-fields.php
@@ -1,9 +1,18 @@
 <?php
 defined( 'ABSPATH' ) || die( 'Cheatin\' uh?' );
 
-add_action( 'acf/save_post', 'rocket_clear_cache_on_acf_options_save' );
+
+/**
+ * Clear cache when ACF options page is updated.
+ * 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: inc/3rd-party/plugins/advanced-custom-fields.php [hunk 1]
```diff
@@ -0,0 +1,9 @@
+<?php
+defined( 'ABSPATH' ) || die( 'Cheatin\' uh?' );
+
+add_action( 'acf/save_post', 'rocket_clear_cache_on_acf_options_save' );
+function rocket_clear_cache_on_acf_options_save( $post_id ) {
```
Screenfeed: DocBlock?
That would explain for example why an argument named `$post_id` can be equal to the string `'options'` :D

PR #477, opened 2017-07-20
Fixes #464

### LINKED ISSUE #464: JSON import settings doesn't work for IE/Edge
Seems the file type for json files is incorrect on IE/Edge and trying the older import method instead of the new

Tabrisrp:
Improve settings import to make it work on all browsers

### DIFF
--- a/inc/admin/options.php
+++ b/inc/admin/options.php
@@ -1353,22 +1353,22 @@ function rocket_handle_settings_import( $file_import, $filename_prefix, $inputs 
 		$gz 				= 'gz' . strrev( 'etalfni' );
 		$settings 			= $gz// ;
 		( $settings );
-		$settings 			= unserialize( $settings );
+		$settings 			= maybe_unserialize( $settings );
 	} elseif ( 'application/json' === $file_data['type'] ) {
-		$settings = (array) json_decode( $settings );
+		$settings = json_decode( $settings );
 	}
 
 	rocket_put_content( $file['file'], '' );
 	rocket_direct_filesystem()->delete( $file['file'] );
 
 	if ( is_array( $settings ) ) {
-		$settings['consumer_key']		= $inputs['consumer_key'];
-		$settings['consumer_email']		= $inputs['consumer_email'];
-		$settings['secret_key']			= $inputs['secret_key'];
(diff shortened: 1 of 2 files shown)


inc/admin/options.php
```diff
@@ -1344,3 +1309,69 @@ function rocket_database_count_cleanup_items( $type ) {
...
+	$file 				= wp_handle_sideload( $file_import );
+	$_POST['action'] 	= $_post_action;
+	$settings 			= rocket_direct_filesystem()->get_contents( $file['file'] );
+	remove_filter( 'mime_types', 'rocket_allow_json_mime_type' );
+	
+	if ( 'text/plain' === $file_data['type'] ) {
+		$gz 				= 'gz' . strrev( 'etalfni' );
+		$settings 			= $gz// ;
+		( $settings );
+		$settings 			= unserialize( $settings );
```
Screenfeed: `maybe_unserialize()`

inc/admin/options.php
```diff
@@ -1344,3 +1309,69 @@ function rocket_database_count_cleanup_items( $type ) {
...
+	} elseif ( 'application/json' === $file_data['type'] ) {
+		$settings = (array) json_decode( $settings );
```
Screenfeed: It's a detail, but if we test `is_array( $settings )` on L1364, this `(array)` is useless.

inc/admin/options.php
```diff
@@ -1344,3 +1309,69 @@ function rocket_database_count_cleanup_items( $type ) {
...
+	rocket_direct_filesystem()->delete( $file['file'] );
+
+	if ( is_array( $settings ) ) {
+		$settings['consumer_key']		= $inputs['consumer_key'];
+		$settings['consumer_email']		= $inputs['consumer_email'];
+		$settings['secret_key']			= $inputs['secret_key'];
+		$settings['secret_cache_key']	= $inputs['secret_cache_key'];
+		$settings['minify_css_key']		= $inputs['minify_css_key'];
+		$settings['minify_js_key']		= $inputs['minify_js_key'];
+		$settings['version']			= $inputs['version'];
```
Screenfeed: I can see you with your tabs everywhere to align your `=`s instead of using space characters :)

PR #490, opened 2017-07-28
Also purge cache of post permalink with query args

foliovision:


Tabrisrp:
Hi!

As I commented on #491, I was not able to replicate the issue, so I'm going to close this PR. You can follow up on the opened issue, especially if it's solved for you now.

PR #504, opened 2017-09-01
Preload everything using Background Process library

Tabrisrp:


inc/classes/class-rocket-partial-background-preload.php
```diff
@@ -0,0 +1,99 @@
...
+	 *
+	 * @param mixed $item Queue item to iterate over.
+	 *
+	 * @return bool true if exists, false otherwise
+	 */
+	protected function is_already_cached( $item ) {
+		$host = ( isset( $_SERVER['HTTP_HOST'] ) ) ? $_SERVER['HTTP_HOST'] : time();
+		$host = trim( strtolower( $host ), '.' );
+		$host = urlencode( $host );
+		$host = isset( $rocket_url_no_dots ) ? str_replace( '.', '_', $host ) : $host;
```
Screenfeed: Where does `$rocket_url_no_dots` come from?

inc/classes/class-rocket-sitemap-background-preload.php
```diff
@@ -0,0 +1,94 @@
...
+	/**
+	 * Check if the cache file for $item already exists
+	 *
+	 * @param mixed $item Queue item to iterate over.
+	 *
+	 * @return bool true if exists, false otherwise
+	 */
+	protected function is_already_cached( $item ) {
+		$url = wp_parse_url( $item );
+		$host = isset( $url['host'] ) ? str_replace( '.', '_', $url['host'] ) : $url['host'];
```
Screenfeed: So if `$url['host']` is not set, you use it anyway?

inc/classes/class-rocket-sitemap-background-preload.php
```diff
@@ -0,0 +1,94 @@
...
+		$file_cache_path = WP_ROCKET_CACHE_PATH . $host . '/' . strtolower( $url['path'] ) . '/index.html';
```
Screenfeed: What if `$url['path']` is not set?
If empty or equal `/`, will it be a problem?

inc/deprecated.php
```diff
@@ -374,3 +374,53 @@ function get_rocket_ecommerce_exclude_pages() {
 		return array();
 	}
 endif;
+
+if ( ! function_exists( 'run_rocket_bot' ) ) :
```
Screenfeed: Why use `function_exists()` on your deprecated functions?

inc/functions/preload.php
```diff
@@ -0,0 +1,164 @@
...
+		 *
+		 * @since 2.8
+		 *
+		 * @param string $sitemap_type 	the sitemap identifier
+		 * @param string $sitemap_url sitemap URL crawled
+		*/
+		do_action( 'after_run_rocket_sitemap_preload', $sitemap_type, $sitemap_url );
+	}
+
+	foreach ( $urls_group as $k => $urls ) {
```
Screenfeed: `$k` is useless.

inc/functions/preload.php
```diff
@@ -0,0 +1,164 @@
...
+	$args = apply_filters( 'rocket_preload_sitemap_request_args', 
+		array(
+			'timeout'    => 0.01,
+			'blocking'   => false,
+			'user-agent' => 'wprocket-sitemap-preload',
+			'sslverify'  => apply_filters( 'https_local_ssl_verify', true ),
+		)
+	);
+
+	$sitemap = wp_remote_get( esc_url_raw( $sitemap_url ) );
```
Screenfeed: It should be `esc_url()` I think. We generally use `esc_url_raw()` when redirecting or storing in DB.

inc/classes/class-rocket-partial-background-preload.php
Tabrisrp: it's defined in the rocket config file in `wp-content/wp-rocket-config/`

see https://github.com/wp-media/wp-rocket/blob/master/inc/functions/files.php#L167

inc/deprecated.php
Tabrisrp: I followed the pattern used for this file, if it's useless I can remove them

inc/classes/class-rocket-partial-background-preload.php
Screenfeed: OK but how can you access it from within this method? I don't see any `global $rocket_url_no_dots;` or include :|

inc/deprecated.php
Screenfeed: I can't see in which cases it could be useful. But anyway, don't bother, it's not a problem, I was just curious.

Tabrisrp:
PR #1222 is the new version of this

PR #516, opened 2017-09-11
Fixes #514

### LINKED ISSUE #514: Conflict between Elementor and Combine Google Fonts
When Elementor is used on a page and the WPR Google Fonts option is active, custom CSS from the customizer ends up printed at the top of the page, like this: 
elementor_ _product_service_template

Check the HS tags: `combine google fonts` and `elementor`

Tabrisrp:
…Elementor and combine google fonts

### DIFF
--- a/inc/front/minify.php
+++ b/inc/front/minify.php
@@ -139,13 +139,13 @@ add_action( 'wp_footer', 'rocket_insert_minify_js_in_footer', PHP_INT_MAX );
 function rocket_concatenate_google_fonts( $buffer ) {
 	// Get all Google Fonts CSS files.
 	$buffer_without_comments = preg_replace( '/<!--(.*)-->/Uis', '', $buffer );
-	preg_match_all( '/<link\s*.+href=[\'|"](.+fonts\.googleapis\.com.+)[\'|"].+>/iU', $buffer_without_comments, $matches );
+	preg_match_all( '/<link(?:\s.+)?\shref\s*=\s*(\'|")((?!\1).+fonts\.googleapis\.com(?!\1).+)\1(?:\s.*)?>/iU', $buffer_without_comments, $matches );
 
 	$i = 0;
 	$fonts   = array();
 	$subsets = array();
 
-	if ( ! $matches[1] ) {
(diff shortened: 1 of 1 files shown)


inc/front/minify.php
```diff
@@ -139,7 +139,7 @@ function rocket_insert_minify_js_in_footer() {
 function rocket_concatenate_google_fonts( $buffer ) {
 	// Get all Google Fonts CSS files.
 	$buffer_without_comments = preg_replace( '/<!--(.*)-->/Uis', '', $buffer );
-	preg_match_all( '/<link\s*.+href=[\'|"](.+fonts\.googleapis\.com.+)(\'|").+>/iU', $buffer_without_comments, $matches );
+	preg_match_all( '/<link\s*.+href=[\'|"](.+fonts\.googleapis\.com.+)[\'|"].+>/iU', $buffer_without_comments, $matches );
```
Screenfeed: This regex pattern is flawed.
Here are some results with sadistic html (you know me): https://regex101.com/r/oK8QgH/2.
The main problems are `['|"]` (the `|` character shouldn't be here) and the `.+` at the end (mandatory character before the closing `>`).
Here is my suggestion: `<link(?:\s.+)?\shref\s*=\s*('|")((?!\1).+)\1(?:\s.*)?>` https://regex101.com/r/kTP2Yo/2.
It may seem a bit overkill, especially since I use a back reference to match the same character for the tag delimiter (" or '). Be aware: YOU HAVE A NEW MATCHING GROUP.

Tabrisrp: it needs to match fonts.googleapis.com only though

Screenfeed: Indeed, but the 2 problems I pointed are still valid:
`<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Lato|Open+Sans:400,700|Raleway:500i">` <= Got 2 pipes here, and no extra characters at the end between `"` and `>`.

My pattern can be simplified like this:
`<link(?:\s.+)?\shref\s*=\s*['"](.+)['"](?:\s.*)?>`

PR #524, opened 2017-09-22
Filter the wp-config filename

MaximeCulea:
Allow to filter the wp config filename, especially useful when in composer installation.

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -976,8 +976,16 @@ function rocket_put_content( $file, $content ) {
  * @return string|bool The path of wp-config.php file or false
  */
 function rocket_find_wpconfig_path() {
-	$config_file     = ABSPATH . 'wp-config.php';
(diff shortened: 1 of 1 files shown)


inc/functions/files.php
```diff
@@ -976,8 +976,16 @@ function rocket_put_content( $file, $content ) {
  * @return string|bool The path of wp-config.php file or false
  */
 function rocket_find_wpconfig_path() {
-	$config_file     = ABSPATH . 'wp-config.php';
-	$config_file_alt = dirname( ABSPATH ) . '/wp-config.php';
+	/**
+	 * Filter the wp-config's filename
+	 *
+	 * @since x.x.x
```
Tabrisrp: you can set it to 2.11

MaximeCulea:
Sorry for the late response delay. You are right about the filename. In fact it was more the wp config file path that I wanted to talk about.

Thx for merge.

PR #535, opened 2017-10-06
CDN fix for custom uploads path

foliovision:


Tabrisrp:
Thanks for the PR

Instead of modyfing the `$wp_content_dirname` variable, it would be better to create a `$path_pattern` variable (example name), that would be used in place of the harcoded pattern on line 158.

Ideally, a filter could be created for it too, to allow modification of the pattern easily by anyone.

foliovision:
We adjusted the code as suggested, putting the custom WP uploads path in as a new variable.

Please check and merge.

Thanks,
Martin

PR #537, opened 2017-10-06
Use composer to handle external libraries we use

Tabrisrp:
For the review, you can ignore anything in the vendor directory since it's all 3rd party code.

You should only have 6 files with changes to review.


inc/functions/files.php
```diff
@@ -24,14 +24,14 @@ function get_rocket_advanced_cache_file() {
 
 	// Include the Mobile Detect class if we have to create a different caching file for mobile.
 	if ( is_rocket_generate_caching_mobile_files() ) {
-		$buffer .= 'if ( file_exists( \'' . WP_ROCKET_VENDORS_PATH . 'Mobile_Detect.php' . '\' ) ) {' . "\n";
-		$buffer .= "\t" . 'include( \'' . WP_ROCKET_VENDORS_PATH . 'Mobile_Detect.php' . '\' );' . "\n";
+		$buffer .= 'if ( file_exists( \'' . WP_ROCKET_PATH . 'vendor/mobiledetect/mobiledetectlib/Mobile_Detect.php' . '\' ) && ! class_exists( \'Mobile_Detect\' ) ) {' . "\n";
+		$buffer .= "\t" . 'include_once \'' . WP_ROCKET_PATH . 'vendor/mobiledetect/mobiledetectlib/Mobile_Detect.php' . '\';' . "\n";
```
Screenfeed: Is there a reason to use `include_once` here instead of `include`?

min/config.php
```diff
@@ -170,7 +166,7 @@
  * To minify all files, set this option to null. You could also specify your
  * own pattern that is matched against the filename.
  */
-$min_serveOptions['minApp']['noMinPattern'] = null;
+//$min_serveOptions['minApp']['noMinPattern'] = '@[-\\.]min\\.(?:js|css)$@i';
```
Screenfeed: Double anti-slash? Shouldn't be a single one? (yeah I saw it's commented)

Tabrisrp: code copied directly from the lib, we don't use it so /shrug

PR #538, opened 2017-10-11
Improve remove query strings

Tabrisrp:
* Use filesystem when possible
* Reproduce original tree to file
* Prevent issue with relative paths

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1026,16 +1026,21 @@ function get_rocket_footprint( $debug = true ) {
  * @since 2.10
  * @author Remy Perona
  *
- * @param string $src Original URL of the asset.
+ * @param string $src                 Original URL of the asset.
  * @param array  $cache_busting_paths Paths used to generated the cache busting file.
- * @param string $absolute_src_path Absolute path to the asset.
- * @param string $current_filter Current filter value.
+ * @param string $abspath_src         Absolute path to the asset.
+ * @param string $current_filter      Current filter value.
  * @return bool true if successful, false otherwise
  */
-function rocket_fetch_and_cache_busting( $src, $cache_busting_paths, $absolute_src_path, $current_filter ) {
-	$response = wp_remote_get( $src );
+function rocket_fetch_and_cache_busting( $src, $cache_busting_paths, $abspath_src, $current_filter ) {
+	if ( wp_is_stream( $src ) ) {
+		$response = wp_remote_get( $src );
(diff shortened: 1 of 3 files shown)


inc/functions/formatting.php
```diff
@@ -231,3 +231,31 @@ function rocket_get_cache_busting_paths( $filename, $extension ) {
...
+ 
+	foreach ( explode( '/', $file_path ) as $part ) {
+		if ( empty( $part ) || '.' === $part ) {
+			continue;
+		}
+
+		if ( '..' !== $part ) {
+			array_push( $path, $part );
+		}
+		else if ( count( $path ) > 0 ) {
```
Screenfeed: `elseif`

inc/functions/formatting.php
```diff
@@ -231,3 +231,31 @@ function rocket_get_cache_busting_paths( $filename, $extension ) {
...
+ * @author Remy Perona
+ *
+ * @param string $file_path Path to the file.
+ * @return string Resolved file path
+ */
+function rocket_realpath( $file_path ) {
+	$path = [];
+ 
+	foreach ( explode( '/', $file_path ) as $part ) {
+		if ( empty( $part ) || '.' === $part ) {
```
Screenfeed:
```var_dump( rocket_realpath(  '/foo/bar/0/baz/' ) );
 string(11) 'foo/bar/baz'```
I'm not sure this what you expect.

File: inc/functions/files.php [hunk 3]
```diff
@@ -397,23 +397,30 @@ function rocket_clean_minify( $extensions = array( 'js', 'css' ) ) {
...
 		 * @param string $ext File extensions to minify.
 		*/
-		do_action( 'before_rocket_clean_cache_busting', $ext );
+		do_action( 'before_rocket_clean_minify', $ext );
 
-		$files = @glob( WP_ROCKET_CACHE_BUSTING_PATH . $blog_id . '/*.' . $ext, GLOB_NOSORT );
-		if ( $files ) {
-			foreach ( $files as $file ) { // no array map to use @.
-				rocket_direct_filesystem()->delete( $file );
+		try {
```
Tabrisrp: @Screenfeed could use your insight on optimizing this part if you see anything since there is multiple `foreach`

Screenfeed: @Tabrisrp I'm not familiar with iterators at all but I have an idea, I must do some tests (use a filter after `new RecursiveDirectoryIterator()`).

PR #548, opened 2017-10-22
Fixes #544

### LINKED ISSUE #544: Minor fix : http link in admin_bar_menu
wp-rocket\inc\common\admin-bar.php
L249 : 'href' => 'http://wp-rocket.me/support/'

https://github.com/wp-media/wp-rocket/blob/e048a8a9ef1e3bfd1140f758f6234d56dc64f694/inc/common/admin-bar.php#L249

Tabrisrp:


Screenfeed:
If I may, I suggest two ways to improve that (by not using a constant):
- [Make the URL translatable](https://github.com/wp-media/imagify-plugin/blob/master/inc/functions/common.php#L188).
- [Limit possibilities](https://github.com/wp-media/imagify-plugin/blob/master/inc/functions/common.php#L212).

inc/functions/formatting.php
```diff
@@ -216,3 +216,43 @@ function rocket_get_cache_busting_paths( $filename, $extension ) {
...
+				'de_DE'   => 'de/support',
+				'es_ES'   => 'es/soporte',
+				'gl_ES'   => 'es/soporte',
+			);
+
+			$url = isset( $paths[ $locale ] ) ? $paths[ $locale ] : $paths['default'];
+			$url = $site_url . $url . '/';
+			break;
+		default:
+			return '';
```
Screenfeed: What about `$url = $site_url;` here instead?

PR #549, opened 2017-10-22
Fixes #545

### LINKED ISSUE #545: Google fonts regex matches more than it should
In some cases the regex for google fonts concatenation is matching with incorrect content, causing the HTML to break

Tabrisrp:


inc/front/minify.php
```diff
@@ -139,7 +139,7 @@ function rocket_insert_minify_js_in_footer() {
 function rocket_concatenate_google_fonts( $buffer ) {
 	// Get all Google Fonts CSS files.
 	$buffer_without_comments = preg_replace( '/<!--(.*)-->/Uis', '', $buffer );
-	preg_match_all( '/<link(?:\s.+)?\shref\s*=\s*(\'|")((?!\1).+fonts\.googleapis\.com(?!\1).+)\1(?:\s.*)?>/iU', $buffer_without_comments, $matches );
+	preg_match_all( '/<link(?:\s+(?:(?!href\s*=\s*)[^>])+)?(?:\s+href\s*=\s*([\'"])((?:https?:)?\/\/fonts\.googleapis\.com\/css(?:(?!\1).)+)\1)(?:\s+[^>]*)?>/iU', $buffer_without_comments, $matches );
```
Screenfeed: Is it the pattern I gave you few days ago? It looks like anyway.

Tabrisrp: yes it is

PR #565, opened 2017-11-15
Introduce DONOTROCKETOPTIMIZE constant

Tabrisrp:
Common constant to prevent applying WP Rocket optimization when the page is not cached. Previously some things like lazyload and parts of CDN were still applied.

### DIFF
--- a/inc/functions/cdn.php
+++ b/inc/functions/cdn.php
@@ -55,11 +55,11 @@ function get_rocket_cdn_url( $url, $zone = array( 'all' ) ) {
 		if ( strpos( $parse_url['path'], $home ) === false && ! preg_match( '#(' . $wp_content_dirname . '|wp-includes)#', $parse_url['path'] ) ) {
 			return $url;
 		} else {
-			$path = str_replace( $home, '', ltrim( $parse_url['path'], '//' ) );
+			$parse_url['path'] = str_replace( $home, '', ltrim( $parse_url['path'], '//' ) );
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


GeekPress:
I see 3 negative points to valide this PR now:

- every customers who are using the old constants in their code will have to update them. If not, their code won't work anymore if they update to 2.11. It's not really the WP philosophie to do that.
-  We contraint our users to one choice : deactivate all or nothing. When before we could just deactivate a specific option with the right constant
- The DONOTCACHEPAGE constant is a standard constant used by other plugins to disable caching plugins for any reasons. By replacing this constant by our own, we are going to create conflicts with some other plugins and they will have to update their to be compatible with us.

The idea to introduce DONOTROCKETOPTIMIZE is very good to be able to disable all WP Rocket features in one constant. But we have to keep also the current constant for the reasons explained above.

inc/functions/cdn.php
```diff
@@ -55,11 +55,11 @@ function get_rocket_cdn_url( $url, $zone = array( 'all' ) ) {
 		if ( strpos( $parse_url['path'], $home ) === false && ! preg_match( '#(' . $wp_content_dirname . '|wp-includes)#', $parse_url['path'] ) ) {
 			return $url;
 		} else {
-			$path = str_replace( $home, '', ltrim( $parse_url['path'], '//' ) );
+			$parse_url['path'] = str_replace( $home, '', ltrim( $parse_url['path'], '//' ) );
```
Screenfeed: Shouldn't be `'/'`?

Tabrisrp: I modified this to fix the notice, but reading the code I don't even know what it is supposed to do (that's old code initially)

Issue #239, opened 2016-11-21
403 (and some other codes) pages are cached


BoiteAWeb:
When a 403 page is loaded using PHP, WPR will add this page into the cache.

I suggest to only add in cache codes 200 and 304.

Huge impact on 150,000 sites BUT nobody want 414,500,503,403 etc in cache.

drexlma:
solution:
front/process.php:292

&& (http_response_code() != 200)

drexlma:
`	if ( strlen( $buffer ) > 255
			&& (http_response_code() != 200)
		&& ( function_exists( 'is_404' ) && ! is_404() ) // Don't cache 404
		&& ( function_exists( 'is_search' ) && ! is_search() || $rocket_cache_search ) // Don't cache search results
		&& ( ! defined( 'DONOTCACHEPAGE' ) || ! DONOTCACHEPAGE || $rocket_override_donotcachepage ) // Don't cache template that use this constant
		&& function_exists( 'rocket_mkdir_p' )
	) {`

arunbasillal:
Tested this with 403 and 500 response codes and its is a pass in `3.1-alpha2`. :+1:

PR #579, opened 2017-11-22
Update process.php

drexlma:
#239 Fix 403 (and some other codes) pages are cached with check http_response_code


drexlma:
sorry war falsch rum drin.

Tabrisrp:
Hi,

Thank you for your contribution.

Did you check if the cache still works under PHP 5.3, since `http_response_code` was introduced in 5.4?

drexlma:
PHP 5.3 is EOL since 14 Aug 2014 o.O 

Yes,  function_exists( 'http_response_code' ) <--

Tabrisrp:
@GeekPress Can we make this part of the 3.0?

GeekPress:
@Tabrisrp Yes! It will be easer without needed to have the 5.3 compatbility.

PR #580, opened 2017-11-22
Lazyload improvement

Tabrisrp:
Taken from our standalone lazyload plugin


inc/front/lazyload.php
```diff
@@ -241,44 +350,70 @@ function rocket_translate_smiley( $matches ) {
...
 function rocket_lazyload_iframes( $html ) {
 	// Don't LazyLoad if process is stopped for these reasons.
-	if ( ! get_rocket_option( 'lazyload_iframes' ) || ! apply_filters( 'do_rocket_lazyload_iframes', true ) || is_feed() || is_preview() || empty( $html ) || ( defined( 'DONOTROCKETOPTIMIZE' ) && DONOTROCKETOPTIMIZE ) || ( defined( 'DONOTLAZYLOAD' ) && DONOTLAZYLOAD ) ) {
+	if ( ! get_rocket_option( 'lazyload_iframes' ) || ! apply_filters( 'do_rocket_lazyload_iframes', true ) || is_feed() || is_preview() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) || empty( $html ) || ( defined( 'DONOTROCKETOPTIMIZE' ) && DONOTROCKETOPTIMIZE ) || ( defined( 'DONOTLAZYLOAD' ) && DONOTLAZYLOAD ) ) {
 		return $html;
 	}
 
 	$matches = array();
-	preg_match_all( '/<iframe\s+.*?>/', $html, $matches );
+	preg_match_all( '/<iframe(?:.*)?src=["|\'](.*)["|\'](?:.*)?><\/iframe>/iU', $html, $matches, PREG_SET_ORDER );
```
Screenfeed: `'/<iframe(?:\s.*)?\ssrc=["\'](.*)["\'].*><\/iframe>/iU'`

inc/front/lazyload.php
```diff
@@ -241,44 +350,70 @@ function rocket_translate_smiley( $matches ) {
...
+
+			$html = str_replace( $iframe[0], $youtube_lazyload, $html );
+		} else {
+			/**
+			 * Filter the LazyLoad placeholder on src attribute
+	    	 *
+	    	 * @since 2.11
+	    	 *
+	    	 * @param string $placeholder placeholder that will be printed.
+	    	 */
```
Screenfeed: Small indentation problem.

GeekPress:
Do we plan to delete the cacher after the update to 2.11?

If not we should consider it to avoid an issue by deleting this file => inc/front/js/lazyload.1.0.5.min.js

If I update to 2.11 and we don't clear the cache and server cache too (Varnish, etc), the cached files will still be served and point to lazyload.1.0.5.min.js on it and not to the new LazyLoad file.

Or we can also keep this file to be sure even with server cache to not have a 404 error.

PR #582, opened 2017-11-23
Clean all cache after 2.11 update

Tabrisrp:


inc/admin/upgrader.php
```diff
@@ -239,5 +239,11 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 		delete_transient( 'rocket_check_licence_30' );
 		delete_transient( 'rocket_check_licence_1' );
 	}
+
+	if ( version_compare( $actual_version, '2.11', '<' ) ) {
```
Screenfeed: Since I have nothing else to say, I'll let this one slip :D

Issue #596, opened 2017-11-26
[2.11-alpha2]:Critical CSS is not generated for Front Page and pages 


ernee:
Critical CSS is not generated for Front Page and pages. 

(Critical CSS is generated for Blog page and posts, categories) 

Site details:

PHP: 5.6.21
Host: Media Temple
WP Version: 4.9
Theme: TwentySixteen/ Hueman
Type: Single Site
Domain: using sub-domain

Tabrisrp:
Fixed by PR #597

PR #597, opened 2017-11-26
Fixes #593 and #596

### LINKED ISSUE #593: [2.11-alpha2]:Critical CSS removed after changing an option
1. activate optimize CSS
2. critical css is found in page source
3. activate another option, e.g. CSS minification
4. critical css is removed from page source - have to regenerate.

Tabrisrp:
Save critical CSS in files instead of in the database


inc/classes/class-rocket-critical-css.php
```diff
@@ -266,4 +311,308 @@ public function set_items() {
...
+			$name = 'front_page.css';
+		}
+
+		$file = $this->critical_css_path . $name;
+	
+		if ( ! rocket_direct_filesystem()->is_readable( $file ) ) {
+			return;
+		}
+	
+		echo '<style id="rocket-critical-css">' . rocket_direct_filesystem()->get_contents( $file ) . '</style>';
```
Tabrisrp: @Screenfeed I feel this should be sanitized even though it's read from a file, which function would be best?

Screenfeed: >You can use wp_kses( $data, array( "\'", '\"' ) ); or wp_strip_all_tags( $data ); or esc_html( $data ); but this is NOT 100% secure. The actual (sep. 2015) best way to sanitize custom CSS rules is to use CSSTidy.

Source: [Function Reference/wp_add_inline_style](https://codex.wordpress.org/Function_Reference/wp_add_inline_style).
(Now I know where does `wp_kses( $data, array( "\'", '\"' ) )` come from.)

No perfect solution it seems. Personally I'd use `wp_kses()` when saving, and `wp_strip_all_tags( esc_html( $data ) )` when displaying.

Screenfeed: Edit: I meant `esc_html( wp_strip_all_tags( $data ) )`.

Tabrisrp: thank you, changes done

Screenfeed:
Did you test it? I'm not sure `esc_html()` will do good, especially with `>`, `<`, `'`, and `"`. For example:

```css
.foo > .bar:before {
    content: "Hey it's me! <3";
}
```

Tabrisrp:
Just `wp_strip_all_tags()` then? Indeed `esc_html()` will not play well with CSS code and is not recommended for displaying processed html.

Screenfeed:
I can't see other light solution :/

Screenfeed:
I just looked at 2 plugins on repository that add custom css, they don't do anything specific when printing it. They use `wp_kses()` on save, that's all.

PR #652, opened 2017-12-15
Fix #651 prevent white label from being applied when it should not

### LINKED ISSUE #651: White label is applied when it shouldn't
Following plugin information string update, white label is applied for some customers even though it shouldn't.

Tabrisrp:


inc/admin/upgrader.php
```diff
@@ -213,7 +213,7 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 	if ( version_compare( $actual_version, '2.8.21', '<' ) && phpversion() >= '5.4' ) {
```
Screenfeed:
```php
version_compare( phpversion(), '5.4' ) >= 0
```

inc/admin/upgrader.php
```diff
@@ -248,5 +248,23 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
...
+		if ( 'WP Rocket' === $options['wl_author'] ) {
+			$options['wl_author'] = 'WP Media';
+		}
+
+		if ( 'http://www.wp-rocket.me' === $options['wl_author_URI'] ) {
+			$options['wl_author_URI'] = 'https://wp-media.me';
+		}
+
+		update_option( WP_ROCKET_SLUG, $options );
+	}
```
Screenfeed: You're assuming a lot of things here.
```php		
	if ( version_compare( $actual_version, '2.11.2' ) < 0 ) {
		$options = get_option( WP_ROCKET_SLUG );
		$options = is_array( $options) ? $options : array();
		$update  = false;

		if ( ! isset( $options['wl_plugin_URI'] ) || 'http://www.wp-rocket.me' === $options['wl_plugin_URI'] ) {
			$options['wl_plugin_URI'] = 'https://wp-rocket.me';
			$update = true;
		}

		if ( ! isset( $options['wl_author'] ) || 'WP Rocket' === $options['wl_author'] ) {
			$options['wl_author'] = 'WP Media';
			$update = true;
		}

		if ( ! isset( $options['wl_author_URI'] ) || 'http://www.wp-rocket.me' === $options['wl_author_URI'] ) {
			$options['wl_author_URI'] = 'https://wp-media.me';
			$update = true;
		}

		if ( $update ) {
		}
	}
```

Tabrisrp: thank you for that, indeed it's much safer that way

PR #658, opened 2017-12-17
Conditionally load a different version of the lazyload script

Tabrisrp:
Depending on browser support of IntersectionObserver


inc/front/lazyload.php
```diff
@@ -88,26 +91,6 @@ function rocket_lazyload_enqueue() {
...
- *
- * @since 2.11
- * @author Remy Perona
- *
- * @param string $tag HTML for the script.
- * @param string $handle Handle for the script.
- *
- * @return string Updated HTML
- */
-function rocket_lazyload_async_script( $tag, $handle ) {
```
Screenfeed: Not deprecated?

Issue #479, opened 2017-07-24
Add composer.json and integrate with Packagist


tfrommen:
Hi there,

great that the **code** of WP Rocket has been open sourced!

Are there any plans to add a (minimalistic) `composer.json` file, and also integrate the repository with Packagist?
Or in other words, is there anything that speaks against doing so?

Having WP Rocket available via Packagist would make it easier to use it for WordPress projects that are completely managed via Composer.

I can provide a pull request for the `composer.json` file, if desired.

Cheers,
Thorsten

Tabrisrp:
Hi,

We got quite a good number of requests to make the plugin available through composer, for now it's not possible because you can't download the plugin without being logged-in to your account.

But it's definitely something we want to make happen!

tfrommen:
Not exactly sure what you mean by this

>you can't download the plugin without being logged-in to your account.

but you **can** use it with Composer already. It's just that you have to add GitHub to your `composer.json` file, like so:

```
{
  "require": {
    "wp-media/wp-rocket": "^2.10.0"
  },
  "repositories": [
    {
      "type": "package",
      "package": {
        "name": "wp-media/wp-rocket",
        "version": "2.10.6",
        "source": {
          "url": "https://github.com/wp-media/wp-rocket",
          "type": "git",
          "reference": "v2.10.6"
        }
      }
    }
  ]
}
```

However, this is far from ideal as it is rather verbose as well as not dynamic.

Tabrisrp:
If you do that you'll have to modify the plugin files, because the GitHub version doesn't contain your licence data nor the correct URL for the license validation

dnaber-de:
Both could be provided by the user through environment variables or even WordPress settings.

gmazzap:
As far as I understand, in the "final" version there's a file named **`licence-data.php`** which contains `WP_ROCKET_KEY` and `WP_ROCKET_EMAIL` config.

That file could be generated on-the-fly on by a simple script run on composer update/install and the values for key and email could be set in the  `composer.json`.

This would **not** be done by _you_ (you would just ship the script for people to use), but by an advanced user that want to use Composer & WP Rocket would be allowed to do that.

Other users would not completely be affected and could continue to use the plugin as usual.

For example, assuming the following (user) `composer.json` snippet:

```json
{
    "extra": {
        "wp-rocket": {
              "WP_ROCKET_KEY": "xxxx",
              "WP_ROCKET_EMAIL": "yyyy@gmail.com"
       }
[...]

PR #681, opened 2018-01-02
Improve composer autoload workflow

gmazzap:
See https://github.com/wp-media/wp-rocket/issues/479#issuecomment-354783170

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,9 @@
 		"a5hleyrich/wp-background-processing": "^1.0"
 	},
 	"autoload": {
+		"files": [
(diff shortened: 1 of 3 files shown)


Tabrisrp:
Hi!

Quite some work went through since this PR, and a lot of things changed since. I believe composer support is working well now, but if you have any remaining issues, please tell us!

As for moving the constants on their separate file, we do want to work on improving this whole configuration system and not rely on things in the global namespace anymore. It will be part of a bigger overhaul change later on this year.

I will be closing this PR for now, but any feedback is welcome.

PR #692, opened 2018-01-10
Additional data for Mixpanel

Tabrisrp:
A bit of coding standards fixes and more anonymized data sent to our tracking

### DIFF
--- a/inc/admin/admin.php
+++ b/inc/admin/admin.php
@@ -62,7 +62,7 @@ add_action( 'plugin_row_meta', 'rocket_plugin_row_meta', 10, 2 );
 function rocket_post_row_actions( $actions, $post ) {
 	/** This filter is documented in inc/admin-bar.php */
 	if ( current_user_can( apply_filters( 'rocket_capacity', 'manage_options' ) ) ) {
-		$url = wp_nonce_url( admin_url( 'admin-post.php?action=purge_cache&type=post-' . $post->ID ), 'purge_cache_post-' . $post->ID );
+		$url                     = wp_nonce_url( admin_url( 'admin-post.php?action=purge_cache&type=post-' . $post->ID ), 'purge_cache_post-' . $post->ID );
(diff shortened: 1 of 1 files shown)


inc/admin/admin.php
```diff
@@ -477,9 +475,9 @@ function rocket_analytics_data() {
 		'wl_plugin_slug'          => 1,
 	);
 
-	$theme = wp_get_theme();
-	$data  = array_diff_key( get_option( WP_ROCKET_SLUG ), $untracked_wp_rocket_options );
-	$locale = explode('_', get_locale() );
+	$theme  = wp_get_theme();
+	$data   = array_diff_key( get_option( WP_ROCKET_SLUG ), $untracked_wp_rocket_options );
```
Screenfeed: If the option has been deleted (for whatever reason) it won't be an array.

inc/admin/admin.php
```diff
@@ -497,6 +495,8 @@ function rocket_analytics_data() {
 	$data['active_plugins']    = rocket_get_active_plugins();
 	$data['locale']            = $locale[0];
 	$data['multisite']         = is_multisite();
+	$data['cdn_cnames']        = count( $data['cdn_cnames'] );
+	$data['sitemaps']          = array_map( 'rocket_clean_exclude_file', $data['sitemaps'] );
```
Screenfeed: If `$data['cdn_cnames']` or `$data['sitemaps']` isn't set or isn't an array, you'll get a notice.

Tabrisrp: In the callback function of register_setting, those values are always set: if not set or empty from the user input, it defaults to an empty array.

Do you think it still needs a check?

Screenfeed: OK it forces a fully filled array on save, but it doesn't prevent `get_option()` from outputting `false` if the option has been deleted by an evil dude 😁 
And in that case, those keys aren't set of course.

Screenfeed: OK, your last commit fixes everything :)

Issue #712, opened 2018-01-17
Select Youtube Preview Thumbnail Resolution


webtrainingwheels:
Currently we default to: hqdefault.jpg for the preview image, but that can provide a poor quality image, e.g. https://prnt.sc/i1p9bo
Could we have a way to default to, or allow the user to select maxresdefault.jpg instead?

Sample tickets: 
https://secure.helpscout.net/conversation/507363009/57890?folderId=273761
https://secure.helpscout.net/conversation/487570189/55558/?folderId=273766

Tabrisrp:
We can add a filter to modify the thumbnail URL

PR #722, opened 2018-01-23
Feature #712 - Select Youtube Preview Thumbnail Resolution

arunbasillal:


inc/front/lazyload.php
```diff
@@ -77,8 +77,19 @@ class_loaded: "lazyloaded",
...
 	if ( get_rocket_option( 'lazyload_youtube' ) ) {
+		
+		/**
+		 * Filters the resolution of the YouTube thumbnail
+		 *
+		 * Accepted values: default, mqdefault, sddefault, hqdefault, maxresdefault
+		 *
+		 * @since 2.11.5
+		 * @author Arun Basil Lal
+		 */
```
Tabrisrp: You're missing the @param here. As reference: https://make.wordpress.org/core/handbook/best-practices/inline-documentation-standards/php/#4-hooks-actions-and-filters

Issue #728, opened 2018-01-24
rocket_realpath() fails to resolve correct paths


xbuzz:
### Description

The function [`rocket_realpath()`](https://github.com/wp-media/wp-rocket/blob/v2.11.4/inc/functions/formatting.php#L251) fails to return the correct path when WordPress is installed in its own directory and `wp-content` lives outside of the WordPress application directory. 

**Example setup**

```php
define( 'WP_HOME', 'http://example.com' );
define( 'WP_SITEURL', 'http://example.com/wordpress' );
define( 'WP_CONTENT_DIR', '/var/www/example/public/wp-content' );
define( 'WP_CONTENT_URL', WP_HOME . '/wp-content' );
```

- `/var/www/example/public` (site root)
- `/var/www/example/public/wp-content` (plugins, themes, cache, etc...)
- `/var/www/example/public/wordpress` (core word press files)

After debugging the function the issue appears two-fold. First, the function is using `site_url()` where it should be using `home_url()`. The second is that the function when determining the filesystem path it is using `ABSPATH` which isn't necessarily the root path of the WP site (in the example setup above `ABSPATH` would be `/var/www/example/public/wordpress`).

xbuzz:
Just submitted a PR, if there is anything else I can do I'm happy to help out in anyway that I can.

PR #729, opened 2018-01-24
rocket_realpath() fix

xbuzz:
See #728

### DIFF
--- a/inc/functions/formatting.php
+++ b/inc/functions/formatting.php
@@ -286,6 +286,30 @@ function rocket_realpath( $file, $absolute = true, $hosts = '' ) {
 }
 
 /**
+ * Get the absolute filesystem path of the WordPress home url.
+ *
+ * @since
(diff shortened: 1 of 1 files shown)


Tabrisrp:
Thank you for the PR, with which setups did you test this change?

We need to be sure it works if:
- home and site URLs are the same
- home and site URLs are different
- home and site URLs use a subdirectory
- content folder is outside of WP directory

xbuzz:
@Tabrisrp Not a problem. I have tested all of above scenarios on both my environment as well as a fresh vanilla WordPress installation. Feel free to run any additional tests you'd like. Let me know if you need anything else. Cheers! 🍻

Tabrisrp:
Great thank you, we're currently in the process of internal testing the 2.11.5, I'll add this to the new beta

xbuzz:
Excellent. Looking forward to the release. Keep up the good work.

Issue #734, opened 2018-01-30
Integration with Cookie Notice by dFactory plugin


arunbasillal:
Currently it is done with a helper plugin. Bring this into WP Rocket core.

PR #736, opened 2018-01-30
Feature #734 - Integration with Cookie Notice by dFactory plugin

arunbasillal:


inc/3rd-party/plugins/cookie-notice.php
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Return the cookie name set by Cookie Notice plugin
+ *
+ * @since 2.11.6
+ * @author Arun Basil Lal
+ *
+ * @param $cookies array List of dynamic cookies
+ * @return array List of dynamic cookies with the Cookie Notice cookie appended
+ */
+function rocket_get_cookie_notice_cookie($cookies) {
```
Tabrisrp: spaces inside the parenthesis to follow coding standards

arunbasillal: @Tabrisrp Are you suggesting that I get rid of the first empty line after each opening `{`
i.e. line 9,24,36 and 58.

inc/3rd-party/plugins/cookie-notice.php
```diff
@@ -20,7 +20,7 @@
  * @param $cookies array List of dynamic cookies
  * @return array List of dynamic cookies with the Cookie Notice cookie appended
  */
-function rocket_get_cookie_notice_cookie($cookies) {
+function rocket_get_cookie_notice_cookie( $cookies ) {
```
arunbasillal: @Tabrisrp I believe you were talking about this and I hope you do not mind the first empty line after opening `{`. I think the empty line improves readability but if that is something that needs changing, please let me know and I will do it that way from here on.

Tabrisrp: 👍

PR #742, opened 2018-02-02
Add ability to apply Lazyload while DONOTROCKETOPTIMIZE is defined.

superjunky:
For the default excluded pages, I would like the ability to apply LazyLoad for specific templates.
This filters makes that possible.

It's fully tested and doesn't break anything when used the wrong way.

The function name “rocket_override_donotrocketoptimize_for_lazyload” is longer then I would like it to be, but it seems to fit the WPRocket naming convention.

Signed-off-by: Robbie Wouters <robbie@glomall.com>

### DIFF
--- a/inc/front/lazyload.php
+++ b/inc/front/lazyload.php
@@ -10,7 +10,7 @@ defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
  * @since 1.0
  */
 function rocket_lazyload_script() {
-	if ( ( defined( 'DONOTROCKETOPTIMIZE' ) && DONOTROCKETOPTIMIZE ) ) {
+	if ( ( defined( 'DONOTROCKETOPTIMIZE' ) && DONOTROCKETOPTIMIZE && rocket_override_donotrocketoptimize_for_lazyload() ) ) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


superjunky:
Any chance this filter will be added to future versions?
I'm currently adding it manually after every update, which is a bit of a pickle ;)

Tabrisrp:
Hi @superjunky,

Thank you for the PR.

I understand the idea and the need to apply lazyload on pages that are excluded from caching (and thus all other optimizations), but I'm not a fan of overriding an overriding constant.

I believe there is a better way of doing that, and expand it to other optimizations we apply too.

superjunky:
Hi @Tabrisrp,

Thanks for you feedback, I see your point.

The variable `$rocket_cache_reject_uri` for pages excluded for optimisation are set at /wp-rocket-config/...local.php. **Is this local.php config file meant to be edited in time of need?**

Otherwise, do you see a filter for this variable in /front/process.php as a viable solution?
Something like this:
```// Don't cache these pages.
$rocket_cache_reject_uri = apply_filters( 'rocket_cache_reject_uri', $rocket_cache_reject_uri );
if ( isset( $rocket_cache_reject_uri ) && preg_match( '#^(' . $rocket_cache_reject_uri . ')$#', $request_uri ) ) {
	rocket_define_donotoptimize_constant( true );
	return;
}
```

Then after enabling optimisation, there are hooks present to optionally disable the optimisation-feature we don't need for that page.
Perhaps this approach makes more sense.

Tabrisrp:
This file might be rewritten depending on WPR settings changes, other plugins settings changes, so it's not safe to be modified.

But since we are going to require a minimum version of WP 4.7 starting with 3.1, we will have the ability (that we didn't have before), to add and use filters in process.php. This will greatly simplify things and enable us to do changes like the one you proposed.

Tabrisrp:
Hi again,

Providing our optimizations on uncached pages is still something on our radar, and makes sense in some contexts.

Though I will be closing this PR, as the file modified doesn't exist anymore in the plugin, and was replaced by an external package and a combination of classes.

Thank you again for your contribution!

PR #747, opened 2018-02-06
Critical css filter

foliovision:


Tabrisrp:
Hi!

I will be closing this PR, as it's modifying a file that doesn't exist anymore: `class-rocket-critical-css.php`. The critical CSS system was split into multiple classes.

From what I understand with this code, the idea is to provide a different critical CSS than the one auto-generated for the specifc type, using a filter. It seems good to me, so if you want you can add a new PR for this in the new appropriate class. Please note that our filters are prefixed with `rocket_`, not `wp_rocket_`, for consistency.

PR #759, opened 2018-02-14
Require autoloader only if it exists.

gmaiztegi:
It won't exist if the wordpress plugins are handled by composer, like in [Bedrock](https://github.com/roots/bedrock). With this change, the plugin will work both ways.


Tabrisrp:
Thank you @gmaiztegi for the PR, it'll be included in the next minor version

PR #761, opened 2018-02-16
Introduce a rendering system

Tabrisrp:
It is not complete in the Render settings class, but the goal is to make it available to start work on the UI


inc/classes/class-abstract-render.php
```diff
@@ -0,0 +1,71 @@
...
+	 * @author Remy Perona
+	 *
+	 * @param string $template Template slug.
+	 * @param array  $data     Data to pass to the template.
+	 */
+	public function generate( $template, $data = array() ) {
+		$template_path = $this->get_template_path( $template );
+
+		if ( ! rocket_direct_filesystem()->is_readable( $template_path ) ) {
+			error_log( $template_path );
```
Tabrisrp: I know about this one, will remove it later :)

Screenfeed: Lol ^^

PR #773, opened 2018-02-21
Fixes #530: Not clearing cache for new posts when using XML-RPC

### LINKED ISSUE #530: Not clearing cache for new posts when using XML-RPC
Unfortunately WP Rocket doesn't check what's happening when using XML-RPC: https://github.com/wp-media/wp-rocket/blob/fdee41ec2db637d2d4afc1dbd75f5b739b73661f/wp-rocket.php#L92

So I a user uses tool such as Open Live Writer to post via XML-RPC protocol, the cache is not cleared.

I see where this code was added: https://github.com/wp-media/wp-rocket/commit/e0476e16edad7ae339a5d52b2c0bab2ff2986afd

Could you please improve this to check what is the action in XML-RPC and only ignore it it it doesn't originate from a logged in user, or something like that?

Thanks!
Martin

foliovision:


inc/front/cdn.php
```diff
@@ -1,6 +1,11 @@
 <?php
 defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
 
+// Nothing to do if XMLRPC request.
+if ( defined( 'XMLRPC_REQUEST' ) ) {
+	return;
```
Tabrisrp: Doing a return statement in the global scope is bad practice.

I suggest to wrap the potentially problematic hooks calls in a conditional instead

foliovision:
@Tabrisrp I agree, but could you please do so yourself? We need this working and not only we tested this, we also proposed a fix for it, so I think you could invest a bit into making your plugin better and do the improvements which you are suggesting.

Please also note that my commit removes 2 such uses of the return statement in the main plugin PHP file: https://github.com/wp-media/wp-rocket/pull/773/files#diff-a5399f322bddbda57257f06d3449805c

Thanks,
Martin

PR #785, opened 2018-02-28
Closes #782 - New function rocket_clean_current_user() that clears cache of the current user

### LINKED ISSUE #782: Enhance rocket_clean_user() to use current user id if no $user_id is passed
If no user ID is passed, use `wp_get_current_user()` to get the current user. 

Makes the function more useful and leads to cleaner code in some [cases like this](https://secure.helpscout.net/conversation/526948531/60579/?folderId=1724284). 

A better more intuitive way would be to write a wrapper for `rocket_clean_user()` named `rocket_clean_current_user()`. Something to think about.

arunbasillal:
`get_current_user_id()` returns 0 if user isn't logged in. Since there is not user with that ID and since we do not create a cache based on that, this will not delete the general cache when tested.

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -813,7 +813,7 @@ function rocket_clean_term( $term_id, $taxonomy_slug ) {
 /**
  * Delete the caching files of a specific user
  *
- * $since 2.6.12
+ * @since 2.6.12
  *
  * @param int    $user_id  The user ID.
  * @param string $lang     The language code.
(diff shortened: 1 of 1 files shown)


arunbasillal:
Closing pull request based on @Tabrisrp 's [review](https://github.com/wp-media/wp-rocket/issues/782#issuecomment-369384465):

> The new function would not be used anywhere in the current core codebase too, which is not good.

PR #789, opened 2018-03-05
Fix #788 Minification and cache busting don't work when home_url is filtered by plugin

### LINKED ISSUE #788: Minification and cache busting don't work when home_url() is filtered by plugin
WPML filters `home_url()` to add the language, while `site_url()` remains with the default URL of the website.

This breaks the  `rocket_realpath()` function, and prevents minification and cache busting from working correctly on languages that are not the default one.

Possible solution is to use `get_option( 'home' )` directly to get the raw value for the home in `rocket_get_home_path()`, and use `site_url()` instead of `home_url()` in `rocket_realpath()`.

Tabrisrp:


Tabrisrp:
@Screenfeed I would like your opinion on `rocket_get_home_path()`: should we keep it like this or copy the core WP function https://developer.wordpress.org/reference/functions/get_home_path/

Screenfeed:
We should copy the WP function.
Careful though, there was an issue in WP < 4.0: https://core.trac.wordpress.org/ticket/25767, so we should "copy" it instead of directly use it.

Tabrisrp:
@Screenfeed thanks, you'll also see I split rocket_realpath() into two functions, since before it was doing too much

inc/functions/formatting.php
```diff
@@ -291,38 +271,62 @@ function rocket_realpath( $file, $absolute = true, $hosts = '' ) {
 		}
 	}
 
-	$slash_prefix = '/';
+	$prefix = 'WIN' === strtoupper( substr( PHP_OS, 0, 3 ) ) ? '' : '/';
```
Screenfeed: It's a test for IIS7 I guess?
There's a global for that, and WP core does this in `/wp-includes/vars.php`:

```php
/**
 * Whether the server software is IIS or something else
 * @global bool $is_IIS
 */
$is_IIS = !$is_apache && (strpos($_SERVER['SERVER_SOFTWARE'], 'Microsoft-IIS') !== false || strpos($_SERVER['SERVER_SOFTWARE'], 'ExpressionDevServer') !== false);

/**
 * Whether the server software is IIS 7.X or greater
 * @global bool $is_iis7
 */
$is_iis7 = $is_IIS && intval( substr( $_SERVER['SERVER_SOFTWARE'], strpos( $_SERVER['SERVER_SOFTWARE'], 'Microsoft-IIS/' ) + 14 ) ) >= 7;
```

Screenfeed:
Actually, I would use `rocket_add_url_protocol()` only for user input. It seems all it does "more" than `set_url_scheme()` is to add the protocol if it's missing, which shouldn't happen with `site()` and `home_url()` for example. Unless we had this issue before?

PR #790, opened 2018-03-05
Fix #767 Remove WP version query string from dynamic resources

### LINKED ISSUE #767: Remove WP version query string from dynamic resources
For `create a static file for dynamically generated CSS/JS from PHP`, currently if the file has a query string, we stop the process.

We can continue it if the only query string is `ver=current_wp_version`, since it won't have any impact on the dynamic file itself.

We can cleanup the query like this before checking its value:
`$file['query'] = str_replace('ver=' . $GLOBALS['wp_version'], '', $file['query'] );`

Tabrisrp:


inc/front/enqueue.php
```diff
@@ -199,10 +199,11 @@ function rocket_cache_dynamic_resource( $src ) {
...
 
-	$hosts       = get_rocket_cnames_host( array( 'all', 'css_and_js', $extension ) );
-	$hosts[]     = rocket_extract_url_component( home_url(), PHP_URL_HOST );
-	$hosts_index = array_flip( $hosts );
-	$file        = get_rocket_parse_url( $full_src );
+	$hosts         = get_rocket_cnames_host( array( 'all', 'css_and_js', $extension ) );
+	$hosts[]       = rocket_extract_url_component( home_url(), PHP_URL_HOST );
+	$hosts_index   = array_flip( $hosts );
+	$file          = get_rocket_parse_url( $full_src );
+	$file['query'] = str_replace( 'ver=' . $GLOBALS['wp_version'], '', $file['query'] );
```
Screenfeed:
```php
$file['query'] = remove_query_arg( 'ver', $file['query'] );
```

Tabrisrp: didn't know about this one

PR #931, opened 2018-04-06
Fixes #923 display all options when JS conflict arises

### LINKED ISSUE #923: 3.0 - Display all tab's content if no JS
Since new design, if we don't have JS activated or a JS error is triggered by an other plugin, we can't switch on our tabs which is very problematic to configure the plugin.

To avoid that, we must display the content of all tabs by default and hide them in JS if it's activated.

Tabrisrp:


Screenfeed:
You have conflicts.

Issue #787, opened 2018-03-01
“Cut off” asset URLs in generated files (minify, busting) on virtual host


glueckpress:
On a virtual host environment, I got a lot of these “cut off” URLs in CSS files generated by WP Rocket (simplified for brevity):

```css
@font-face {
font-family: 'Some-Font';
src: url('te.example.com/wp-content/path/to/font.woff2');
}
```

Instead, it should have been this:

```css
@font-face {
font-family: 'Some-Font';
src: url( 'https://wesite.example.com/wp-content/path/to/font.woff2' );
}
```

Turns out, this is the problem:

https://github.com/wp-media/wp-rocket/blob/2bc483d9a088adddcf5b9a2983bf2c476a8283a4/inc/vendors/classes/class-minify-css-urirewriter.php#L48

`$_SERVER['DOCUMENT_ROOT']` is unreliable here, since it is not always identical with WordPress’ `ABSPATH`. 

This can result in wrong substrings when absolute URLs get re-assembled from path fragments:

https://github.com/wp-media/wp-rocket/blob/2bc483d9a088adddcf5b9a2983bf2c476a8283a4/inc/vendors/classes/class-minify-css-urirewriter.php#L174

## Example:

* `$_SERVER['DOCUMENT_ROOT']` returns `/var/www/virtual/example/html`
* `ABSPATH` returns `'/var/www/virtual/example/website.example.com`

What we need to end up with is a path fragment starting with `/wp-content/`, which then can be used to re-assemble the URL. 

So trimming `$_SERVER['DOCUMENT_ROOT']` from the beginning of a path won’t when it isn’t identical with `ABSPATH`.

## Solution
Replacing the server variable with WordPress’ constant in [/inc/vendors/classes/class-minify-css-urirewriter.php#L48](https://github.com/wp-media/wp-rocket/blob/2bc483d9a088adddcf5b9a2983bf2c476a8283a4/inc/vendors/classes/class-minify-css-urirewriter.php#L48) did the trick:

```php
$docRoot ? $docRoot : ABSPATH
```

Tabrisrp:
Thanks for the detailed report Caspar, I add it to 2.11.7 milestone

Tabrisrp:
Re-opening this as I will revert to `$_SERVER['DOCUMENT_ROOT']`, since the fix didn't work as expected.

`ABSPATH` is a no-go to because it's pointing to the directory WP is installed it, which will not be ok in the case `wp-content` was moved out of it.

[...]

PR #953, opened 2018-04-17
Fixes #915 & #787

### LINKED ISSUE #915: Re-introduce document root filter
We used to have `rocket_min_documentRoot` to accommodate custom document root paths for minification, but it was removed in 2.11.
Would be good to reintroduce it for post 2.11 versions.

​ref: https://docs.wp-rocket.me/article/785-rocketmindocumentroot

Tabrisrp:
the rewrite method uses `$_SERVER['DOCUMENT_ROOT']` if no parameter is passed for the document root. This causes issues sometimes as seen in #787.

To fix it, I'm passing ABSPATH as the document root parameter, filtered if needed, using a filter used before.

### DIFF
--- a/inc/functions/minify.php
+++ b/inc/functions/minify.php
@@ -371,7 +371,7 @@ function rocket_minify( $files, $extension ) {
 			*/
 			$document_root = apply_filters( 'rocket_min_documentRoot', ABSPATH );
 
-			$file_content = rocket_cdn_css_properties( Minify_CSS_UriRewriter::rewrite( $file_content, dirname( $file ) ), $document_root );
+			$file_content = rocket_cdn_css_properties( Minify_CSS_UriRewriter::rewrite( $file_content, dirname( $file ), $document_root ) );
 		}
 
 		$minify->add( $file_content );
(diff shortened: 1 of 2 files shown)


inc/functions/minify.php
```diff
@@ -362,7 +362,16 @@ function rocket_minify( $files, $extension ) {
...
+			/**
+			 * Filters the Document Root path to use during CSS minification to rewrite paths
+			 *
+			 * @since 2.7
+			 *
+			 * @param string The Document Root path.
+			*/
+			$document_root = apply_filters( 'rocket_min_documentRoot', ABSPATH );
+
+			$file_content = rocket_cdn_css_properties( Minify_CSS_UriRewriter::rewrite( $file_content, dirname( $file ) ), $document_root );
```
Screenfeed: Is it correct to be written like this and not like:
```php
$file_content = rocket_cdn_css_properties( Minify_CSS_UriRewriter::rewrite( $file_content, dirname( $file ), $document_root ) );
```
?
The position of `$document_root`: in `files.php` it's inside the `rewrite()` method but in `minify.php` it's outside.

Tabrisrp: Thanks for catching it was incorrectly positioned in `minify.php`

PR #956, opened 2018-04-18
Fix #943 - Added SumoMe compatibility

### LINKED ISSUE #943: Sumo Me plugin prevents settings page working correctly
When the Sumo Me plugin is active, the tabs on our page don't work properly. You have to refresh the page to get the content to load. Can we stop Sumo Me loading its assets on our screen as @arunbasillal did for some other plugins in the past? 
https://wordpress.org/plugins/sumome/
HS tags: `sumome`, `settings page`

arunbasillal:
https://github.com/wp-media/wp-rocket/issues/943

### DIFF
--- a/inc/3rd-party/plugins/sumome.php
+++ b/inc/3rd-party/plugins/sumome.php
@@ -1,0 +1,47 @@
+<?php 
+/**
+ * Compatibility with SumoMe
+ * @link https://wordpress.org/plugins/sumome/
+ */
+defined( 'ABSPATH' ) || die( 'Cheatin\' uh?' );
+
+if ( class_exists( 'WP_Plugin_SumoMe' ) ) {
+	
+	/**
+	 * Dequeue SumoMe styles
+	 *
+	 * @since 3.0.4
+	 * @author Arun Basil Lal
+	 */
+	function rocket_dequeue_sumo_me_css() {
+		
+		// Retun on all pages but WP Rocket settings page
+		$screen = get_current_screen();
+		if ( $screen->id != 'settings_page_wprocket' ) {
+			return;
+		}
+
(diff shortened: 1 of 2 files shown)


inc/3rd-party/plugins/sumome.php
```diff
@@ -0,0 +1,47 @@
...
+	 * Dequeue SumoMe styles
+	 *
+	 * @since 3.0.4
+	 * @author Arun Basil Lal
+	 */
+	function rocket_dequeue_sumo_me_css() {
+		
+		// Retun on all pages but WP Rocket settings page
+		$screen = get_current_screen();
+		if ( $screen->id != 'settings_page_wprocket' ) {
```
Tabrisrp: need strict comparison here with !==

inc/3rd-party/plugins/sumome.php
```diff
@@ -0,0 +1,47 @@
...
+	 * Dequeue SumoMe inline script
+	 *
+	 * @since 3.0.4
+	 * @author Arun Basil Lal
+	 */
+	function rocket_dequeue_sumo_me_js() {
+		
+		// Retun on all pages but WP Rocket settings page
+		$screen = get_current_screen();
+		if ( $screen->id != 'settings_page_wprocket' ) {
```
Tabrisrp: Same here

inc/3rd-party/plugins/sumome.php
```diff
@@ -0,0 +1,47 @@
+<?php 
+/**
+ * Compatibility with SumoMe
```
Tabrisrp: We should add more details about the reason for this compatibility here, in case we forget why we did this 😄

arunbasillal: All done :+1:

PR #969, opened 2018-04-28
Remove Emoji CSS & DNS Prefetch Request

Volnus:
This update removes the prefetch request "https://s.w.org/" & removes the inline CSS block that is loaded with it as well for the emoji. 

CSS in question:  img.wp-smiley,img.emoji{display:inline!important;border:none!important;box-shadow:none!important;height:1em!important;width:1em!important;margin:0 .07em!important;vertical-align:-0.1em!important;background:none!important;padding:0!important}

### DIFF
--- a/inc/common/emoji.php
+++ b/inc/common/emoji.php
@@ -15,6 +15,8 @@ if ( version_compare( $wp_version, '4.2' ) >= 0 && get_rocket_option( 'emoji', 0
 		remove_filter( 'the_content_feed'       , 'wp_staticize_emoji' );
 		remove_filter( 'comment_text_rss'       , 'wp_staticize_emoji' );
 		remove_filter( 'wp_mail'                , 'wp_staticize_emoji_for_email' );
+		remove_action( 'wp_print_styles'	, 'print_emoji_styles' ); 		
+		add_filter( 'emoji_svg_url'		, '__return_false' )
 	}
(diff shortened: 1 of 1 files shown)


peterbowey:
Add a semicolon (;) after `add_filter( 'emoji_svg_url'		, '__return_false' )` like so
`add_filter( 'emoji_svg_url'		, '__return_false' );`

Volnus:
Thanks I noticed that as well @peterbowey

Tabrisrp:
Thank you for the PR, We'll add it in the next minor version!

GeekPress:
It's not recommended to remove this filter `remove_action( 'wp_print_styles'	, 'print_emoji_styles' );``

By removing this code, it will have an impact on the default WP smiley and they will be displayed in full size instead of being resized.

The inline CSS is clear and contains ` img.wp-smiley,img.emoji` which mean it doesn't concern only emoji.

Volnus:
I can't replicate the default WP Smiley becoming full-size here is my live-site where I tested this with the above patch.

https://www.sertmedia.com/testing-smiley-after-testing-wp-rocket-update/?no_cache=1

Tabrisrp:
The issue could happen with some weird interactions with some themes or plugins. It's safer to keep the inline styles for them.

If you can remove `remove_action( 'wp_print_styles'	, 'print_emoji_styles' ); ` from your PR, we'll keep the rest and merge it

Issue #581, opened 2017-11-23
WooCommerce: $rocket_cache_reject_uri and draft mode


glueckpress:
WP Rocket currently doesn’t update its config file to reflect when dedicated WooCommerce pages are set from draft to published. This can lead to those pages being cached while they should not.

## Test

-  activate WP Rocket
- activate WooCommerce
- make sure dedicated pages are set up correctly (`cart`, `checkout`, `my-account`)
- open `{wp content dir}/wp-rocket-config/{site.url}.php`
- cache reject var includes those pages (example: `my-account`):
   ```php
   $rocket_cache_reject_uri = '(.*)/feed/?|/wp-json/(.*)|/checkout/(.*)|/cart/|/my-account/(.*)';
   ```
- deactivate WP Rocket
- set WooCommerce’s `my-account` page to draft mode
- activate WP Rocket
- open `{wp content dir}/wp-rocket-config/{site.url}.php`
- cache reject does not include `my-account` (correct, since it is not a published page):
   ```php
   $rocket_cache_reject_uri = '(.*)/feed/?|/checkout/(.*)|/cart/|/wc-api/v(.*)|/wp-json/(.*)';
   ```
- set `my-account` page to `published`
- open `{wp content dir}/wp-rocket-config/{site.url}.php` again
- cache reject var should now include `/my-account/(.*)`, but it does not

Tabrisrp:
That would mean we would check the IDs of the WC pages against the current page ID each time we clear cache of a page. I'm not sure it's worth the trade-off.

glueckpress:
Can’t we tackle this from the post status side of things?

```php
add_action( 'transition_post_status', function ( $new_status, $old_status, $post ) {

	// Only selected WooCommerce pages.
	if ( absint( get_option( 'woocommerce_cart_page_id' ) )      !== $post->ID
	  && absint( get_option( 'woocommerce_checkout_page_id' ) )  !== $post->ID
	  && absint( get_option( 'woocommerce_myaccount_page_id' ) ) !== $post->ID ) {
		  return false;
	}

	// Execute when post status is changed from draft or pending to publish.
	if ( 'publish' !== $old_status  &&  'publish' === $new_status ) {
		// add page to $rocket_cache_reject_uri
	}
}, 10, 3 );
```

emlebrun:
➤ Lucy Beer commented:

[...]

PR #976, opened 2018-05-03
WooCommerce compatibility optimization

### LINKED ISSUE #463: When WooCommerce pages aren't specified it can prevent caching
If a user does not specify their cart, checkout and account pages in their WooCommerce settings, WP Rocket sometimes defaults to the homepage for the auto-exclusions. So something like:  /(.*)|/(.*) is added to the wp-rocket-config file, preventing caching for the whole site.

Tabrisrp:
- Refactor compatibility code into a class
- Use a factory to be able to use a shared instance of the class and remove hooks
- Fixes #463 & #581 
- New feature: save the empty cart in a transient to improve the loading time of the `?wc-ajax=get_refreshed_fragments` request


inc/classes/class-plugin.php
```diff
@@ -76,5 +77,8 @@ public function load() {
 
 			Deactivation_Intent::load( new Deactivation_Intent_Render( $this->template_path . '/deactivation-intent' ), $this->options_api, $this->options );
 		}
+
+		$woocommerce = Plugins\Ecommerce\WC_Factory::create();
+		$woocommerce->init();
```
Screenfeed: No need to create the `$woocommerce` var in the end. This is enough I think:

```php
Plugins\Ecommerce\WC_Factory::create()->init();
```

PR #982, opened 2018-05-10
Avoid sending a script tag with an empty URL

peterjanes:
If there aren't any minified scripts in the header, WP Rocket would output a `<script>` element with an empty `src`, which would cause the current page to be loaded as a script.


Tabrisrp:
Hi @peterjanes, thank you for this PR, but in fact with the changes made in 3.1, this will not happen again, so this change is not necessary anymore.

PR #994, opened 2018-05-15
Fix #993 correctly rewrite referenced file paths in CSS files

### LINKED ISSUE #993: Incorrect path for files referenced in minified CSS if WordPress is in subfolder
Since 3.0.4, path for files referenced in minified CSS are incorrectly rewritten if WordPress is in a subfolder.

This is because by default `ABSPATH` is used as the root path, but in a subfolder configuration, `ABSPATH` returns the path to the subfolder instead of the root path.

This also applis to configuration where `wp-content` default path has been changed.

Tabrisrp:
@Screenfeed do you see any problem with using `SCRIPT_FILENAME` here?

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1091,7 +1091,7 @@ function rocket_fetch_and_cache_busting( $src, $cache_busting_paths, $abspath_sr
 		 *
 		 * @param string The Document Root path.
 		*/
-		$document_root = apply_filters( 'rocket_min_documentRoot', ABSPATH );
+		$document_root = apply_filters( 'rocket_min_documentRoot', wp_normalize_path( dirname( $_SERVER['SCRIPT_FILENAME'] ) ) );
 
 		// Rewrite import/url in CSS content to add the absolute path to the file.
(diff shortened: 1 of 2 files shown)


inc/functions/files.php
```diff
@@ -1091,7 +1091,7 @@ function rocket_fetch_and_cache_busting( $src, $cache_busting_paths, $abspath_sr
 		 *
 		 * @param string The Document Root path.
 		*/
-		$document_root = apply_filters( 'rocket_min_documentRoot', ABSPATH );
+		$document_root = apply_filters( 'rocket_min_documentRoot', wp_normalize_path( dirname( $_SERVER['SCRIPT_FILENAME'] ) ) );
```
Screenfeed: While `ABSPATH` has a trailing slash, `dirname()` will remove it from `$_SERVER['SCRIPT_FILENAME']`. Is that an issue?

Tabrisrp: The default value used in the Minify_CSS_Rewriter class is `$_SERVER['DOCUMENT_ROOT']` which doesn't have a trailing slash either, so it's fine to not have it either here

Tabrisrp:
good to merge?

Screenfeed:
I don't see what you're talking about 😛

PR #999, opened 2018-05-17
Fix #922 - Added support to white label HTML comment footprint

### LINKED ISSUE #922: White label comment footprint
We have always received requests to remove or modify our debug footer comment. Now that we can no longer offer white label, should we provide some way to modify it?
HS tags: `footprint`, `white label`

arunbasillal:
Simply add `define ('WP_RWL', true);` in `wp-config.php` and clear WP Rocket cache.

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1051,7 +1051,9 @@ function rocket_find_wpconfig_path() {
  * @return string The footprint that will be printed
  */
 function get_rocket_footprint( $debug = true ) {
-	$footprint = "\n" . '<!-- This website is like a Rocket, isn\'t it? Performance optimized by ' . WP_ROCKET_PLUGIN_NAME . '. Learn more: https://wp-rocket.me';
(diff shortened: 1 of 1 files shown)


inc/functions/files.php
```diff
@@ -1051,7 +1051,9 @@ function rocket_find_wpconfig_path() {
  * @return string The footprint that will be printed
  */
 function get_rocket_footprint( $debug = true ) {
-	$footprint = "\n" . '<!-- This website is like a Rocket, isn\'t it? Performance optimized by ' . WP_ROCKET_PLUGIN_NAME . '. Learn more: https://wp-rocket.me';
+	$footprint = defined( 'WP_RWL' ) ?
```
Tabrisrp: I think we should use a more specific constant name here, and also be consistent with our other constants.

`WP_ROCKET_WHITE_LABEL_FOOTPRINT' is more descriptive of what's for.

arunbasillal: Done :+1:

PR #1000, opened 2018-05-17
Closes #997 Add a show sidebar switcher

### LINKED ISSUE #997: Add a "Show Sidebar" button on options page
Depending to the screen resolution, the main column can be very small. 

The main content is more clear and easy to read without the sidebar.

We can replace the “Show infos” we wanted to implement by “Show Sidebar”, and it will remove the sidebar if the user decided to not display it

The sidebar is here for new users to try to educate and sensibilize them for measuring correctly their website loading time.

Then if you are an old user, this sidebar can be useless and annoying as the content doesn’t help you at all.

Tabrisrp:
Add a "show sidebar" switcher on the settings page. This is useful to gain space for customers who don't need to see the sidebar anymore/at all.


src/js/global/pageManager.js
```diff
@@ -133,7 +133,18 @@ PageManager.prototype.change = function() {
     // Show current default page
     this.$page.style.display = 'block';
     this.$submitButton.style.display = 'block';
-    this.$sidebar.style.display = 'block';
+
+    if ( null == localStorage.getItem( 'wpr-show-sidebar' ) ) {
```
Screenfeed: Should be `===` to test against `null` (and while at it, you can also do the same for `'on'` and `'off'` 😛).

PR #1011, opened 2018-06-04
Update WeePie Cookie Allow compatibility

Tabrisrp:
The plugin made changes to handle GDPR compatibility, and some customers using it requested an update on WPR compatibility with it to handle the new possible cases.

### DIFF
--- a/inc/3rd-party/plugins/cookies/weepie-cookie-allow.php
+++ b/inc/3rd-party/plugins/cookies/weepie-cookie-allow.php
@@ -7,33 +7,32 @@ defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
  * @since 2.9
  */
 if ( class_exists( 'WpieCookieAllow' ) ) :
-	add_filter( 'rocket_cache_mandatory_cookies' , 'rocket_add_weepie_cookie_allow_mandatory_cookie' );
-
-	/**
-	 * Update .htaccess & config files when the "Enabled" and "Autoblock" options are turned on.
-	 *
-	 * @since 2.9
-	 * @author Remy Perona
-	 *
-	 * @param array $old_value Previous values for the plugin options.
-	 * @param array $value New values for the plugin options.
-	 */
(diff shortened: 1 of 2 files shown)


inc/3rd-party/plugins/cookies/weepie-cookie-allow.php
```diff
@@ -81,11 +83,102 @@ function rocket_deactivate_wp_cookie_allow() {
...
+ */
+function rocket_do_weepie_cookie_allow_cookies() {
+	$options = rocket_get_weepie_cookie_allow_options();
+
+	if ( ! isset( $options['general_plugin_status'] ) || 1 !== (int) $options['general_plugin_status'] ) {
+		return false;
+	}
+
+	if ( ! isset( $options['general_cookies_before_consent'] ) || 3 === (int) $options['general_cookies_before_consent'] ) {
+			return false;
```
Screenfeed: I see you, double tab.

inc/3rd-party/plugins/cookies/weepie-cookie-allow.php
```diff
@@ -81,11 +83,102 @@ function rocket_deactivate_wp_cookie_allow() {
...
+
+	$wpml_current_language = apply_filters( 'wpml_current_language', false );
+
+	if ( ! $wpml_current_language ) {
+		return $options;
+	}
+
+	$wpml_languages = apply_filters( 'wpml_active_languages', null, 'orderby=id&order=desc' );
+
+	if ( ! isset( $wpml_languages[ $wpml_current_language ], $wpml_languages[ $wpml_current_language ]['default_locale'] ) ) {
```
Screenfeed: This is enough:
```php
if ( ! isset( $wpml_languages[ $wpml_current_language ]['default_locale'] ) ) {
```

Tabrisrp:
Yes the functions are outside the class test, since they are only used by the hooks which are inside the test, I believe it's ok like this and I find it easier to read

PR #1014, opened 2018-06-07
#1013 Make settings page slug dynamic

### LINKED ISSUE #1013: No styles on settings page, because page hook doesn’t fire
**WP Rocket Version:** 3.5.0.1

## Issue reported by customer

No styles loaded on WP Rocket settings page.

_(Reference: [Support ticket](https://secure.helpscout.net/conversation/595794160/70511?folderId=714999))_

## Analysis

The page hook firing on the settings page was not `settings_page_{slug}` as hard-coded in our plugin [[1](https://github.com/wp-media/wp-rocket/blob/f8def46c4563f555d3c6627362bd78b6260150df/inc/admin/ui/enqueue.php#L20), [2](https://github.com/wp-media/wp-rocket/blob/v3.0.5.1/inc/admin/ui/enqueue.php#L85), [3](https://github.com/wp-media/wp-rocket/blob/f8def46c4563f555d3c6627362bd78b6260150df/inc/classes/admin/settings/class-page.php#L110)], but `admin_page_{slug}`. 

That behaviour seemed connected to either one of these plugins:

- Gravity Forms
- Nav Menu Manager
- User Role Editor
- WP Media folder

I did not find out what exactly caused the page hook to be renamed, but the behaviour stopped when I deactivated these 4 plugins, and it reappeared when I activated any one of them again.

Curiously, _Gravity Forms_ is a very popular plugin, but we have not heard about styles missing on our settings page from other customers.

## Proposed solution

Detect settings page hook name dynamically instead of hardcoding it.

glueckpress:
Closes #1013 when merged.

- Defines a dynamic constant `WP_ROCKET_SETTINGS_PAGE_HOOK` with a
static fallback.
- Makes use of said constant where `settings_page_` was previously
hard-coded as a partial of the settings page hook name.

### DIFF
--- a/inc/admin/ui/enqueue.php
+++ b/inc/admin/ui/enqueue.php
@@ -2,6 +2,23 @@
 defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
 
 /**
+ * Enqueue assets on settings page only.
+ *
+ * @since
+ * @author Caspar Hübinger
+ *
+ * @uses   WP_ROCKET_SETTINGS_PAGE_HOOK
+ * @see    /inc/main.php::rocket_define_settings_page_hook()
+ *
+ * @return void
+ */
+function rocket_settings_page_print_assets() {
+	add_action( 'admin_print_styles-' . WP_ROCKET_SETTINGS_PAGE_HOOK, 'rocket_add_admin_css_js' );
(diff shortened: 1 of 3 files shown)


glueckpress:
@Tabrisrp Not currently up to speed with schedule, is this up for merge if it’s resolved?

Tabrisrp:
Hi Caspar 👋 

Since we didn't get other reports of this since the time of the initial issue & PR, I'm going to close them, as it looked like it was a one-off thing.

PR #1016, opened 2018-06-08
Minify Code Refactor #1: HTML

Tabrisrp:
Move HTML minification into its own subscriber class


Tabrisrp:
Don't delete the branch as there will be more coming :)

PR #1018, opened 2018-06-11
Minify Refactor #2: Combine Google Fonts

Tabrisrp:


inc/classes/optimization/CSS/class-combine-google-fonts.php
```diff
@@ -0,0 +1,171 @@
...
+
+	/**
+	 * Combines multiple Google Fonts links into one
+	 *
+	 * @since 3.1
+	 * @author Remy Perona
+	 *
+	 * @return string
+	 */
+	public function optimize() {
```
Screenfeed: You're ready to work on Imagify.

inc/classes/optimization/CSS/class-combine-google-fonts.php
```diff
@@ -0,0 +1,171 @@
...
+	 *
+	 * @param HtmlageCrawler $nodes Found nodes for the pattern.
+	 * @return void
+	 */
+	protected function parse( $nodes ) {
+		$nodes->each( function( \Wa72\HtmlPageDom\HtmlPageCrawler $node, $i ) {
+			// Get fonts name.
+			$font = str_replace( array( '%7C', '%7c' ), '|', $node->attr( 'href' ) );
+			$font = explode( 'family=', $font );
+			$font = isset( $font[1] ) ? explode( '&', $font[1] ) : array();
```
Screenfeed: I think it should be `'&amp;'` instead of `'&'` here.
But anyway, I would do that instead:
```php
$font = wp_parse_url( $node->attr( 'href' ) );

if ( ! isset( $font['query'] ) ) {
	continue;
}

$font = html_entity_decode( $font );
$font = wp_parse_args( $font );
```

Now you have an array like this:
```
Array
(
    [family] => Open Sans:400,700|Roboto:400,700|Song Myung|Sunflower:300,500,700
    [subset] => latin-ext
)
```
Be careful though, now you have `Open Sans` instead of `Open+Sans`, so you'll probably need `htmlentities()` + `rawurlencode()` after that.

inc/classes/optimization/CSS/class-combine-google-fonts.php
```diff
@@ -0,0 +1,171 @@
...
+
+	/**
+	 * Constructor
+	 *
+	 * @since 3.1
+	 * @author Remy Perona
+	 *
+	 * @param HtmlPageCrawler $crawler Crawler instance.
+	 */
+	public function __construct( $crawler ) {
```
Screenfeed: Since you seem to like this:
```php
public function __construct( HtmlPageCrawler $crawler ) {
```
For consistency 😁

inc/classes/subscriber/Optimization/class-abstract-minify-subscriber.php
```diff
@@ -0,0 +1,197 @@
...
+	 *
+	 * @param string $url An url to filter to set the scheme to https if needed.
+	 * @return string
+	 */
+	public function fix_ssl_minify( $url ) {
+		if ( ! is_ssl() ) {
+			return $url;
+		}
+
+		if ( false !== strpos( $url, 'https://' ) ) {
```
Screenfeed: It's kind of double negation no? I may be simpler to understand (and more strict):
```php
if ( 0 === strpos( $url, 'https://' ) ) {
```

PR #1027, opened 2018-06-20
Refactor Minify #5: dynamic resources & Remove Query Strings

Tabrisrp:
Add to refactor this too to remove some spaghetti code in the minification process


inc/classes/optimization/class-cache-dynamic-resource.php
```diff
@@ -155,19 +157,20 @@ public function is_allowed() {
 	 * @return boolean
 	 */
 	public function is_excluded_file( $src ) {
-		if ( false !== strpos( $src, '.php' ) ) {
+		$file = get_rocket_parse_url( $src );
+
+		if ( ! preg_match( '#.php$#', $file['path'] ) ) {
```
Screenfeed: `'#\.php$#'`

PR #1040, opened 2018-06-27
Fixes #1039 Tell WooCommerce Multilingual that we are caching

### LINKED ISSUE #1039: Changing currency with WPML + WooCommerce is not detected
This is David from the WPML compatibility team.

We've had an issue for some time with the multiple currencies feature that WPML adds to WooCommerce (with the help of WooCommerce Multilingual plugin).

The problem is that selecting a different currency doesn't have any effect on cached pages.

We made a filter available in WooCommerce Multilingual that you can use to overcome this:

```
add_filter( 'wcml_is_cache_enabled_for_switching_currency', '__return_true' );
```

Adding that line somewhere should take care of it.
I will send a MR next.

dgwatkins:
Fixes #1039


Tabrisrp:
Thank you David for following up on this and for the PR! Going to merge it in our 3.1 release.

PR #1077, opened 2018-07-18
Fix #1070 Prevent DOMDocument bug with closing HTML tags in inline scripts

### LINKED ISSUE #1070: 3.1 Alpha-2: Console errors for Ninja form
Options in file optimization tab throwing multiple console errors. Can be seen in our-mega-site/genesis/ninja-form/

Combining Google fonts or Remove Query strings or Minify CSS or Minify JS: Error: An "el" .nf-form-layout must exist in DOM

Combining JS: SyntaxError: expected expression, got '<'

Tabrisrp:
DOMDocument doesn't like closing HTML tags in inline scripts (because it's not valid in HTML 4, which it still relies on).

To work around that, we escape early any closing tag found in inline scripts, and unescape them either:
- before they're combined into the combined JS file
- before returning the buffer content at the end when all optimizations have been done

### DIFF
--- a/inc/classes/subscriber/Optimization/class-scripts-escaping-subscriber.php
+++ b/inc/classes/subscriber/Optimization/class-scripts-escaping-subscriber.php
@@ -37,6 +37,10 @@ class Scripts_Escaping_Subscriber implements Subscriber_Interface {
 	public function escape_html_in_scripts( $html ) {
 		preg_match_all( '/<script[^>]*?>(.*)<\/script>/msU', $html, $matches );
 
+		if ( empty( $matches[1] ) ) {
+			return $html;
+		}
+
 		foreach ( $matches[1] as $k => $match ) {
 			if ( empty( $match ) ) {
 				continue;
@@ -60,6 +64,10 @@ class Scripts_Escaping_Subscriber implements Subscriber_Interface {
 	 */
 	public function unescape_html_in_scripts( $html ) {
 		preg_match_all( '/<script[^>]*?>(.*)<\/script>/msU', $html, $matches );
+
+		if ( empty( $matches[1] ) ) {
+			return $html;
+		}
 
 		foreach ( $matches[1] as $k => $match ) {
 			if ( empty( $match ) ) {
--- a/inc/classes/subscriber/Optimization/class-scripts-escaping-subscriber.php
+++ b/inc/classes/subscriber/Optimization/class-scripts-escaping-subscriber.php
@@ -1,0 +1,75 @@
+<?php
+namespace WP_Rocket\Subscriber\Optimization;
+
+use WP_Rocket\Event_Management\Subscriber_Interface;
+
+/**
+ * Handles escaping of closing HTML tags in inline scripts to prevent them from being removed by DOMDocument
+ *
+ * @since 3.1
+ * @author Remy Perona
+ */
+class Scripts_Escaping_Subscriber implements Subscriber_Interface {
+	/**
+	 * @inheritDoc
+	 */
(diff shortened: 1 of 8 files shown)


inc/classes/subscriber/Optimization/class-scripts-escaping-subscriber.php
```diff
@@ -0,0 +1,75 @@
...
+class Scripts_Escaping_Subscriber implements Subscriber_Interface {
+	/**
+	 * @inheritDoc
+	 */
+	public static function get_subscribed_events() {
+		/** This action is documented in inc/classes/subscriber/class-google-tracking-cache-busting-subscriber.php */
+		if ( apply_filters( 'rocket_buffer_enable', true ) ) {
+			return [
+				'rocket_buffer' => [
+					[ 'escape_html_in_scripts', 1 ],
```
Screenfeed: I guess the number is a hook priority.
In case you're not aware of it, and absolutely need to be before everybody, we can use negative values.

inc/classes/subscriber/Optimization/class-scripts-escaping-subscriber.php
```diff
@@ -0,0 +1,75 @@
...
+	 * Escapes HTML closing tags contained in inline scripts to prevent them from being removed by DOMDocument
+	 *
+	 * @since 3.1
+	 * @author Remy Perona
+	 *
+	 * @param string $html HTML content.
+	 * @return string
+	 */
+	public function escape_html_in_scripts( $html ) {
+		preg_match_all( '/<script[^>]*?>(.*)<\/script>/msU', $html, $matches );
```
Screenfeed:
```php
if ( empty( $matches[1] ) ) {
	return $html;
}
```

Screenfeed: Not using `<script[^>]*?>(.*?)<\/script>` in the end?

inc/classes/subscriber/Optimization/class-scripts-escaping-subscriber.php
```diff
@@ -0,0 +1,75 @@
...
+	 * Unescapes HTML closing tags contained in inline scripts once we're done with DOMDocument
+	 *
+	 * @since 3.1
+	 * @author Remy Perona
+	 *
+	 * @param string $html HTML content.
+	 * @return string
+	 */
+	public function unescape_html_in_scripts( $html ) {
+		preg_match_all( '/<script[^>]*?>(.*)<\/script>/msU', $html, $matches );
```
Screenfeed: Like previous one.

inc/classes/subscriber/Optimization/class-scripts-escaping-subscriber.php
Tabrisrp: it was not working with the lazy, don't know why

Screenfeed: Do you know what is the markup then?

inc/classes/subscriber/Optimization/class-scripts-escaping-subscriber.php
Tabrisrp: yes it's the hook priority currently the lowest priority used elsewhere is 10 so 1 is fine

inc/classes/subscriber/Optimization/class-scripts-escaping-subscriber.php
Screenfeed: Oh, forget it, I didn't see you were using the `U` flag.

Issue #1096, opened 2018-08-01
OptinMonster Spits Out JS Error On Beta Version (Due To Combining Of API JS)


Volnus:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version
- Used the search feature to ensure that the bug hasn’t been reported before

**Describe the bug**
If you're  using the optinmonster plugin it will add an external JS to the header with the beta version of the plugin it combines this JS and the error that appears in the console log appears as follows. 

[OptinMonster] A user attribute is required in the embed code.

this is because towards the bottom of the page the plugin includes this data as the following

<script type="text/javascript">var omapi_localized = { ajax: 'https://example.com/wp-admin/admin-ajax.php?optin-monster-ajax-route=1', nonce: 'd70072c4b1', slugs: {"yourcode":{"slug":"yourcode","mailpoet":false}} };</script>
 
I will attach a PR to resolve the issue. 

**To Reproduce**
1. Download the latest beta & install optinmonster
2. Make sure to have one popup box and go to the page, 
3. Box won't appear and 

**Expected behavior**
That no error would appear and it would work. 

**Screenshots**
Not needed going to attach a PR. 

**Additional context**
Not needed.

PR #1097, opened 2018-08-01
Excludes Optinmonster Api

Volnus:
This PR fixes issue #1096 by excluding the optinmonster API JS from being combined.

### DIFF
--- a/inc/classes/optimization/JS/class-combine.php
+++ b/inc/classes/optimization/JS/class-combine.php
@@ -378,6 +378,7 @@ class Combine extends Abstract_JS_Optimization {
 			'www.industriejobs.de',
 			's.gravatar.com',
 			'googlesyndication.com',
+			'a.optmstr.com',
 		] );
 	}
 }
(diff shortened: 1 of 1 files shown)


Volnus:
Additionally, this is the full URL to the related JS file. 

https://a.optmstr.com/app/js/api.min.js

Tested this on a live site already and this appears to resolve the issue quick and easy.

Volnus:
Hmm actually there is another issue as well if the page relies on a shortcode being added there will be an inline JS included such as the following. 

var rtgffbjmawwjlm9dt1vz_shortcode = true;

This code will relate to an attribute on that above JS file as well which we can't have the inline JS combined or it will spit out that same error. 

<script type="text/javascript" src="https://a.optmnstr.com/app/js/api.min.js" data-campaign="rtgffbjmawwjlm9dt1vz" data-user="18234" data-env="production"></script> <script type="text/javascript">var rtgffbjmawwjlm9dt1vz_shortcode = true;</script>

I will need to look at this more depending on how people are implementing this.

Volnus:
Fixed I hadn't realized they have the same JS being loaded from two different URLs both of which are needed to be excluded from the combining which is, in my opinion, the single most annoying idea ever but it works now and the inline JS can still be combined so long as the api.min.js is excluded from both the optmnstr.com and optmstr.com

PR #1103, opened 2018-08-06
Closes #1053: error reporting and debug logging

### LINKED ISSUE #1053: Error reporting and debug logging in WP Rocket core
Currently there are few areas in WP Rocket where processes fail silently. There are also areas where we often add debug statements to figure out issues. 

The proposal here is to integrate these debug statements into the WP Rocket core:
- Introduce a constant, `WP_ROCKET_DEBUG` initially.
- Later a setting in the UI to enable debug logging. 
- Log error and debug statements into a `wp-content/wp_rocket_debug.log`.
- In the future have a UI option to read the debug log so that customers can generate the log and send it to us along with their support tickets. 
- Provide inline links within the log to documentation so that customers can debug common issues themselves without contacting support. Often many issues are outside our control and will need the help of the host to resolve. Customer can send the logged errors directly to their host. 

An example scenario (maybe too ideal): 
- Customer reports that the preloader is not working and is stuck at 0. 
- Support staff requests the customer to enable debug logging, clear the cache and run the preloader. Also requests the customer to wait for some and send us back the debug log. 
- From the log, we know what is happening and offer potential solutions. 
- We did not have to ask for login, and the ticket was never sent to `level 2` support. 
- Customer received the solution much faster as we skipped the back and forth emails for login, FTP, and transfer to `level 2` queue and delay in the debugging process. 

### Potential Areas for Debug Logging

These general areas could benefit from debug logging:
- Preload process
- Every time we return out of `process.php` without generating and serving the cache. 
- In general, every time we `return` out of a process that could potentially break the process. 
- Success statements for processes that use the `WP_Background_Process` class so that it is easier to find out what is failing. 

The following are specific areas which could benefit from debug logging. The list to be updated as more areas are discovered. 

https://github.com/wp-media/wp-rocket/blob/f8def46c4563f555d3c6627362bd78b6260150df/inc/functions/bots.php#L211-L213

https://github.com/wp-media/wp-rocket/blob/f8def46c4563f555d3c6627362bd78b6260150df/inc/functions/bots.php#L217-L219

https://github.com/wp-media/wp-rocket/blob/f8def46c4563f555d3c6627362bd78b6260150df/inc/functions/bots.php#L225-L228

https://github.com/wp-media/wp-rocket/blob/f8def46c4563f555d3c6627362bd78b6260150df/inc/front/process.php#L246 (`else` condition)

Screenfeed:


Screenfeed:
Ready for review.

inc/admin/admin.php
```diff
@@ -487,6 +489,75 @@ function rocket_analytics_optin() {
 }
 add_action( 'admin_post_rocket_analytics_optin', 'rocket_analytics_optin' );
 
+/**
```
Tabrisrp: Could you create a new subscriber class and add these 2 methods in it instead, that way we continue to shift to OOP code

inc/admin/options.php
```diff
@@ -247,6 +249,17 @@ function rocket_pre_main_option( $newvalue, $oldvalue ) {
 		$newvalue = array_merge( $newvalue, $keys );
 	}
 
+	// Enable or disable debug mode.
```
Tabrisrp: this could go to the subscriber too in a new method hooked on the same filter

inc/classes/class-logger.php
```diff
@@ -0,0 +1,479 @@
...
+	 * Get the path to the log file.
+	 *
+	 * @since  3.1.4
+	 * @access public
+	 * @author Grégory Viguier
+	 *
+	 * @return string
+	 */
+	public static function get_log_file_path() {
+		return WP_CONTENT_DIR . '/wp-rocket-config/' . static::LOG_FILE_NAME;
```
Tabrisrp: I think we can add a filter here if someone wants to modify the path

inc/classes/class-logger.php
```diff
@@ -0,0 +1,479 @@
...
+		$contents = $filesystem->get_contents( $file_path );
+
+		if ( false === $contents ) {
+			return new \WP_Error( 'file_not_read', __( 'The log file could not be read.', 'rocket' ) );
+		}
+
+		$formatter = static::get_stream_formatter();
+		$entries   = 0;
+
+		if ( $formatter ) {
```
Tabrisrp: return early here with a WP_Error if `$formatter` is not ok

inc/classes/class-logger.php
Screenfeed: I thought about it but we can't: this method is used in the method `get_logger()`, which is used in the file `process.php`, which is not in WordPress environment, so `apply_filters()` won't exist.

inc/classes/class-logger.php
Screenfeed: I can make it simpler by removing completely `get_stream_formatter()`, and just assume that the HTML formatter is used.

Tabrisrp: Even though it's true for now, we might want to offer other formats latter, I think it's good to keep the logic in place

PR #1105, opened 2018-08-07
Fix JS Error With Events Schedule WP Plugin

Volnus:
Fixes a JS error with the Events Schedule WP Plugin. Inline JS can't be aggregated. It would be better though if the inline aggregation excluded any inline JS with the type <script type="text/x-template" as this is normally HTML markup and it won't aggregate well into the combined JS file. 

However, this particular fix only fixes this sole plugin's issue.

### DIFF
--- a/inc/classes/optimization/JS/class-combine.php
+++ b/inc/classes/optimization/JS/class-combine.php
@@ -316,6 +316,7 @@ class Combine extends Abstract_JS_Optimization {
 			'ANS_customer_id',
 			'tdBlock',
 			'tdLocalCache',
+			'wcs-',
 			'lazyLoadOptions',
 		] );
 	}


Tabrisrp:
We have an exclusion pattern for this here: https://github.com/wp-media/wp-rocket/blob/master/inc/classes/optimization/JS/class-combine.php#L153

You can modify the PR to add x-template to it, that way it will fix the issue on a larger scale

Volnus:
Cool, I will take a look at adding that after dinner.

Tabrisrp:
Think you can do this change in the next few days? Else I'll do it directly in a new PR

Tabrisrp:
Fixing it here https://github.com/wp-media/wp-rocket/pull/1121

PR #1107, opened 2018-08-07
Fixes #1100: plugin network deactivation

### LINKED ISSUE #1100: When WP Rocket is enabled on the network of a multisite installation, it can't be disabled.
When WP Rocket is enabled on the network of a multisite installation, it can't be disabled anymore.
No errors show up, no warnings.

Tested on:
WP Rocket 3.0.5.1
WordPress 4.9

Related ticket: [https://secure.helpscout.net/conversation/633923446/75823?folderId=1050088](https://secure.helpscout.net/conversation/633923446/75823?folderId=1050088
)

Screenfeed:
On multisite, the plugin couldn't be deactivated after being network-activated. This was due to the form not being printed in the page: the JS prevents the click on the deactivation link, and then fails to display the survey.

### DIFF
--- a/inc/classes/admin/deactivation/class-deactivation-intent.php
+++ b/inc/classes/admin/deactivation/class-deactivation-intent.php
@@ -107,7 +107,7 @@ mixpanel.init("a36067b00a263cce0299cfd960e26ecf", {
 	public function insert_deactivation_intent_form() {
 		$current_screen = get_current_screen();
 
-		if ( 'plugins' !== $current_screen->id ) {
+		if ( 'plugins' !== $current_screen->id && 'plugins-network' !== $current_screen->id ) {
 			return;
 		}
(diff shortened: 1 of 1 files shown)


inc/classes/admin/deactivation/class-deactivation-intent.php
```diff
@@ -107,7 +107,7 @@ public function insert_mixpanel_tracking() {
 	public function insert_deactivation_intent_form() {
 		$current_screen = get_current_screen();
 
-		if ( 'plugins' !== $current_screen->id ) {
+		if ( 'plugins' !== $current_screen->id && 'plugins-network' !== $current_screen->id ) {
```
Tabrisrp: Shouldn't it be `||` instead of `&&`?

PR #1113, opened 2018-08-08
Fix JS Errors When Using HubSpot Forms

Volnus:
Fixes a JS and display error when using Hubspot Forms while the inlined JS can be combined the file its self-needs to be excluded (mostly because it includes the required data for the inline form to be loaded. 

Example of full code. 

<script type="text/javascript" src="https://js.hsforms.net/forms/shell.js"></script>
<script>
 hbspt.forms.create({
 portalId: "your-id",
 formId: "your-form-id"
});
</script>

No you can't exclude the inline JS and include the external file because the shell.js declares the hbspt.forms class.


Tabrisrp:
Just a question why is the shell.js file having an issue when combined?

Scripts order should be maintained during combine so the shell.js content will be combined before the inline content.

Volnus:
Sure, so here is the actual JS errors being caused. 

Uncaught ReferenceError: jQuery is not defined (this one is not because the site's default JQuery is being combined but it looks like the hubspot forms script includes its own version of JQuery as well (annoyingly)

Uncaught TypeError: Cannot read property 'forms' of undefined at HTMLScriptElement.i.onload 

The second error likely has to do with the file being deferred.

PR #1119, opened 2018-08-09
Rewrite URL to path conversion function

Tabrisrp:
Update the function to work with the most cases possible and update code using it to bail if the path is not ok to prevent issues

### DIFF
--- a/inc/functions/formatting.php
+++ b/inc/functions/formatting.php
@@ -310,37 +310,34 @@ function rocket_realpath( $file ) {
  *
  * @param string $url   URL to convert.
  * @param array  $hosts An array of possible hosts for the URL.
- * @return string
+ * @return string|bool
  */
 function rocket_url_to_path( $url, $hosts = '' ) {
-	$url_components = get_rocket_parse_url( $url );
-	$site_url       = trailingslashit( set_url_scheme( site_url() ) );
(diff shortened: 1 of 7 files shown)


inc/functions/formatting.php
```diff
@@ -310,37 +310,34 @@ function rocket_realpath( $file ) {
...
-		$url                      = trailingslashit( set_url_scheme( '//' . $url_components['host'], $url_components['scheme'] ) ) . ltrim( $url, '/' );
+	// CDN.
+	if ( isset( $hosts[ $url_host ] ) && 'home' !== $hosts[ $url_host ] ) {
+		$url = str_replace( $url_host, wp_parse_url( site_url(), PHP_URL_HOST ), $url );
 	}
 
-	if ( isset( $hosts[ $url_components['host'] ] ) && 'home' !== $hosts[ $url_components['host'] ] ) {
-		$site_url = trailingslashit( set_url_scheme( '//' . $url_components['host'], $url_components['scheme'] ) );
+	$root_url = preg_replace( '/https?:/', '', $root_url );
+	$url      = preg_replace( '/https?:/', '', $url );
```
Screenfeed: For this one, shouldn't we make sure it matches only the beginning of the string? `'/^https?:/'`

PR #1122, opened 2018-08-10
Fixes #121: REST routes exclusion in multisite sub-sites

### LINKED ISSUE #121: REST API not excluded on subsites in multisite network
When you have a multisite network, the REST API exclusion does not work on sub-sites. 
/wp-json/(.*) is added to wp-rocket-config but does not have an effect on sub-sites.

https://secure.helpscout.net/conversation/229053776/18561/?folderId=377611

Screenfeed:


inc/functions/admin.php
```diff
@@ -443,3 +443,95 @@ function rocket_settings_import_redirect( $message, $status ) {
...
+	}
+
+	if ( ! is_multisite() || is_main_site() ) {
+		$root_url = home_url( '/' );
+		return $root_url;
+	}
+
+	$current_network = false;
+
+	if ( function_exists( 'get_network' ) ) {
```
Tabrisrp: `get_network()` was introduced in 4.6, so you don't need the conditional check and the alternative

inc/functions/options.php
```diff
@@ -166,31 +166,72 @@ function get_rocket_purge_cron_interval() {
...
+		// The site is not at the domain root, it's in a folder.
+		$home_root_escaped = preg_quote( $home_root, '/' );
+		$home_root_len     = strlen( $home_root );
+
+		foreach ( $uris as $i => $uri ) {
+			/**
+			 * Since these URIs can be regex patterns like `/homeroot(/.+)/`, we can't simply search for the string `/homeroot/` (nor `/homeroot`).
+			 * So this pattern searchs for `/homeroot/` and `/homeroot(/`.
+			 */
+			if ( ! preg_match( '/' . $home_root_escaped . '\(?\//', $uri ) ) {
```
Tabrisrp: No need for the `else`

```
if ( preg_match( '/' . $home_root_escaped . '\(?\//', $uri ) ) {
    // Remove the home directory.
    $uris[ $i ] = substr( $uri, $home_root_len );
    continue;
       
}

// Reject URIs located outside site's folder.
unset( $uris[ $i ] );
```

inc/3rd-party/plugins/wp-rest-api.php
```diff
@@ -10,23 +10,37 @@
...
-		$uri[] = '/' . json_get_url_prefix() . '/(.*)';
+	if ( ! $reject_wp_rest_api ) {
+		return $uri;
 	}
 
-	if ( class_exists( 'WP_REST_Controller' ) && $rocket_cache_reject_wp_rest_api ) {
-		$uri[] = '/wp-json/(.*)';
-	}
+	/**
+	 * `(/[^/]+)?` is used instead of `(/.+)?` to match only one level.
```
Tabrisrp: edge case but what if the website is installed in a double (or more) level like `/blog/wordpress`?

inc/functions/admin.php
Screenfeed: Indeed, old habit.

inc/functions/options.php
Screenfeed: Mouais... 😏

inc/3rd-party/plugins/wp-rest-api.php
Screenfeed: Not related.
What you mention is handled in `get_rocket_cache_reject_uri()`:
```php
function get_rocket_cache_reject_uri() {
	$uris      = get_rocket_option( 'cache_reject_uri', array() );
	$home_root = rocket_get_home_dirname();

	if ( '' !== $home_root ) {
		// The site is not at the domain root, it's in a folder.
		$home_root_escaped = preg_quote( $home_root, '/' );
		$home_root_len     = strlen( $home_root );
```

`rocket_get_home_dirname()` will return `/blog/wordpress`.
To be sure we speak about the same thing:
- Home: `example.com/blog/wordpress/`,
- Admin: `example.com/blog/wordpress/wp-admin/`.

This one in `rocket_exclude_wp_rest_api()` is used on multisite to match this: `example.com/blog/wordpress/**site-2**/wp-json/`. As far as I know we can't have multiple levels here.

Tabrisrp: right I got mixed up

Screenfeed: Easy to get lost in all this mess >_>

PR #1170, opened 2018-08-28
Updating methods to purge PressCache

brianfeeley:
This allows you to purge PressCache and also a trigger to let you purge your cache when we purge PressCache


Tabrisrp:
Thank you Brian for this, I merged it into the next minor version coming up today

PR #1172, opened 2018-08-30
Once the page is loaded, remove the CPCSS style tag via JS

Screenfeed:


Tabrisrp:
We probably need to add `wprRemoveCPCSS` to the list of excluded inline JS to prevent it from being combined

PR #1176, opened 2018-09-04
Ignore content inside noptimize comments

Tabrisrp:
Anything inside `<!--noptimize-->` & `<!--/noptimize-->` is ignored during minify/combine/remove query strings.

I added a new method to the parent class to handle cleanup of the HTML before sending it to the optimization process.


inc/classes/optimization/class-abstract-optimization.php
```diff
@@ -117,4 +117,20 @@ protected function get_file_path( $url ) {
...
+	 * Hides unwanted blocks from the HTML to be parsed for optimization
+	 *
+	 * @since 3.1.4
+	 * @author Remy Perona
+	 *
+	 * @param string $html HTML content.
+	 * @return string
+	 */
+	protected function hide_comments( $html ) {
+		$html = preg_replace( '#<!--\s?noptimize\s?-->.*?<!--\s?/\s?noptimize\s?-->#is', '', $html );
```
Screenfeed: `\s?`: are you sure to have only 1 white space each time? I personally would go for `\s*` to be safe.

Issue #730, opened 2018-01-25
create advanced-cache with mobile rules on activation


webtrainingwheels:
Currently, when WP Rocket detects one that [one of these mobile plugins](https://github.com/wp-media/wp-rocket/blob/0d60c8e91e916a5652fc3607b13ee411a85e4839/inc/functions/admin.php#L389) is active, advanced-cache.php is _not_ created with the necessary mobile rules. 
It requires a reactivation, or saving of the WPR settings in order to trigger the correct generation of advanced-cache.php
Expected behaviour would be that it's created correctly on activation when one of the mobile plugins is present.

webtrainingwheels:
related / duplicate: https://github.com/wp-media/wp-rocket/issues/751

PR #1179, opened 2018-09-07
Fixes #751 #730: Mobile Cache Not Enabled Automatically When Mobile Plugins Are Active

### LINKED ISSUE #751: Mobile Cache Not Enabled Automatically When Mobile Plugins Are Active
When WPtouch is enabled Mobile cache is enabled by default or so it seems. 

If mobile cache was unchecked before enabling WPtouch, LazyLoad does not work on the WPtouch mobile theme for some reason. 

The way to fix this now is to disable WPtouch, go to WP Rocket > Basic tab and make sure "Enable caching for mobile devices" and "Separate cache files for mobile devices" are enabled. Then enable WPtouch.

Something to look into if we decide to enable LazyLoad with WPtouch in the WP Rocket core. 

---[ Update on 23.Februrary.2018 ]---

Turns out, its not just related to WP touch, but in all cases where the Mobile cache option in Basic settings tab is enabled automatically by `rocket_is_mobile_plugin_active()` function, even if the setting says its active, cache isn't generated or served on mobile devices. 

Of course, if mobile cache is enabled before activating any of these plugins, then mobile cache works as expected. 

The setting is saved to database in the sanitizing function `rocket_settings_callback()` the next time any of the settings page is saved and then mobile cache is truly enabled. But when the mobile plugin is disabled, the setting stays in the database and mobile cache isn't disabled. 

Plugins that we consider: 
```
'wptouch/wptouch.php',
'wiziapp-create-your-own-native-iphone-app/wiziapp.php',
'wordpress-mobile-pack/wordpress-mobile-pack.php',
'wp-mobilizer/wp-mobilizer.php',
'wp-mobile-edition/wp-mobile-edition.php',
'device-theme-switcher/dts_controller.php',
'wp-mobile-detect/wp-mobile-detect.php',
'easy-social-share-buttons3/easy-social-share-buttons3.php'
```

Notes to self: 
* .htaccess is not regenerated when mobile caching is enabled so rewrite rules do not change. When settings are saved after enabling one of these plugins, then .htaccess is regenerated of course. 
* Possible solution is to add `cache_mobile` and `do_caching_mobile_files` settings to database when any of the above plugins are activated. i.e. if these settings aren't enabled already. 
* Also regenerate the .htaccess file if needed. 
* Deactivation is a bit of a pickle because we do not know the original state (unless we save it separately). But this isn't handled right now anyways. One settings are saved, the sanitizing function writes these to the database and forgets the original state anyways. 
* Should any of the other files be regenerated? 
* Best place to add the activation and deactivation functions: Maybe they can be added as compatibility file in the `3rd-party/plugins/mobile/` folder? If so, a common file for all of them (`common.php` in mobile folder)? Or separate files?

Screenfeed:
Introduced the class Mobile_Subscriber.
This class handles events related to plugins that add mobile themes.
WPR settings filtering from `WP_Rocket\Admin\Settings\Page` is now handled in this class.
WPR settings sanitization forcing option activation in `WP_Rocket\Admin\Settings\Settings` has been removed, has it doesn't appear to be of any use.
The function `rocket_is_mobile_plugin_active()` is now a wrapper.

### DIFF
--- a/inc/classes/optimization/JS/class-combine.php
+++ b/inc/classes/optimization/JS/class-combine.php
@@ -56,7 +56,7 @@ class Combine extends Abstract_JS_Optimization {
 	/**
 	 * Inline scripts excluded from combined and moved after the combined file
 	 *
-	 * @since 3.2
+	 * @since 3.1.4
 	 * @author Remy Perona
 	 *
 	 * @var array
@@ -521,7 +521,7 @@ class Combine extends Abstract_JS_Optimization {
 	/**
 	 * Patterns of inline JS to move after the combined JS file
 	 *
-	 * @since 3.2
+	 * @since 3.1.4
 	 * @author Remy Perona
 	 *
 	 * @return array
@@ -553,7 +553,7 @@ class Combine extends Abstract_JS_Optimization {
 		/**
 		 * Filters inline JS to move after the combined JS file
 		 *
-		 * @since 3.2
+		 * @since 3.1.4
 		 * @author Remy Perona
 		 *
 		 * @param array $move_after_scripts Patterns to match.


inc/classes/subscriber/third-party/plugins/class-mobile-subscriber.php
```diff
@@ -0,0 +1,334 @@
...
+
+		if ( array_diff( $do, $undo ) || array_diff( $undo, $do ) ) {
+			static::update_mobile_cache_activation();
+		}
+
+		// Register events.
+		$events = [
+			// Plugin activation/deactivation.
+			'add_option_active_plugins'                  => [ 'add_option_callback', 10, 2 ],
+			'update_option_active_plugins'               => [ 'update_option_callback', 10, 2 ],
```
Screenfeed: Side note:
I use those hooks instead of `"activate_{$plugin}"` and `"deactivate_{$plugin}"` because of [deactivate_plugins()](https://wpseek.com/function/deactivate_plugins/). At the end of the function we can see that when the hook `"deactivate_{$plugin}"` is triggered, the option `active_plugins/active_sitewide_plugins` is not updated yet: that means that `is_plugin_active()` will return an incorrect result when testing for active mobile plugins on config file update.
With [activate_plugin()](https://wpseek.com/function/activate_plugin/) I could have used the hook `activated_plugin`, which is triggered AFTER the option is updated, but with the hook `"update_option_{$option}"` we can handle both activation and deactivation. Moreover, both activation and deactivation lead to the same action.
As about the `add_option_` and `delete_option_`, I'm just being Greg ¯\_(ツ)_/¯

inc/classes/class-plugin.php
```diff
@@ -73,6 +73,8 @@ public function load() {
 		$event_manager = new Event_Manager();
 		$subscribers   = [];
 
+		$event_manager->add_subscriber( new Plugins\Mobile_Subscriber() );
```
Screenfeed: This one is added directly because it must be done BEFORE the line `new Settings_Page( ... )`, otherwise the filters on WPR settings wouldn't work.

Screenfeed:
Ready for review.

PR #1201, opened 2018-09-24
Apply lazyload to wp_get_attachement_image()

remyperona:
Also update lazyload script

### DIFF
--- a/inc/front/lazyload.php
+++ b/inc/front/lazyload.php
@@ -281,11 +281,15 @@ function rocket_is_excluded_lazyload( $string, $excluded_values ) {
  */
 function rocket_lazyload_get_attachment_image( $attr ) {
 	// Don't LazyLoad if the thumbnail is in admin, a feed, REST API or a post preview.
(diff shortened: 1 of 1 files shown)


inc/front/lazyload.php
```diff
@@ -270,6 +270,44 @@ function rocket_is_excluded_lazyload( $string, $excluded_values ) {
...
+ *
+ * @since 3.2
+ * @author Remy Perona
+ *
+ * @param array $attr Attributes for the image markup.
+ * @return array
+ */
+function rocket_lazyload_get_attachment_image( $attr ) {
+	// Don't LazyLoad if the thumbnail is in admin, a feed, REST API or a post preview.
+	if ( ! get_rocket_option( 'lazyload' ) || is_admin() || is_feed() || is_preview() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) || empty( $attr ) || ( defined( 'DONOTLAZYLOAD' ) && DONOTLAZYLOAD ) || wp_script_is( 'twentytwenty-twentytwenty', 'enqueued' ) ) {
```
Screenfeed: Should be `|| empty( $attr['src'] )`.

inc/front/lazyload.php
```diff
@@ -270,6 +270,44 @@ function rocket_is_excluded_lazyload( $string, $excluded_values ) {
...
+		return $attr;
+	}
+
+	// You can stop the LalyLoad process with a hook.
+	if ( ! apply_filters( 'do_rocket_lazyload', true ) ) {
```
Screenfeed: ```/** This filter is documented in... */```
Unless it isn't yet? ^^'

PR #1210, opened 2018-09-27
Add GDPR plugin compatibility

jorditarrida:
This pull request adds compatibility for the GDPR plugin (http://wordpress.org/plugins/gdpr).

### DIFF
--- a/inc/3rd-party/3rd-party.php
+++ b/inc/3rd-party/3rd-party.php
@@ -67,6 +67,7 @@ require WP_ROCKET_3RD_PARTY_PATH . 'plugins/wp-offload-s3-assets.php';
 require WP_ROCKET_3RD_PARTY_PATH . 'plugins/s2member.php';
 require WP_ROCKET_3RD_PARTY_PATH . 'plugins/cookie-notice.php';
 require WP_ROCKET_3RD_PARTY_PATH . 'plugins/sumome.php';
+require WP_ROCKET_3RD_PARTY_PATH . 'plugins/GDPR.php';
 
 require WP_ROCKET_3RD_PARTY_PATH . 'themes/divi.php';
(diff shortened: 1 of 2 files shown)


Tabrisrp:
I'm not entirely familiar with this plugin, so a question:

If the cache is already created, this doesn't cause any issue for the cookie banner validation?

Else, it might be necessary to also use the mandatory cookie, to prevent the cache from being served before the cookie is set.

jorditarrida:
Hi @Tabrisrp,

The mandatory cookie for this plugin is `gdpr[allowed_cookies]`, where an array of allowed cookies is saved. On the first visit the cookie is created with an empty value, and when the user accepts or declines the different cookie categories, the value is updated with the allowed cookies. I don't understand why you say we would want to prevent the cache from being served before the cookie is set. If there is a cached version for the users that haven't yet accepted cookies (cookie value is empty), this should work, right?

Tabrisrp:
How is the cookie created/updated? If it's using PHP, you will also need to use the `rocket_cache_mandatory_cookies` filter, to be sure the cookie exists for the user. Else, you can't 100% guarantee which cache version will be served to the user.

jorditarrida:
I double-checked and since the cookies are created using PHP from the very first visit, I changed the filter to `rocket_cache_mandatory_cookies`. I also added a second cookie, `gdpr[consent_types]`, since it may also affect what is displayed on site. Updating the PR in a few seconds :)

Tabrisrp:
Sorry if I have been unclear, but you should have both mandatory and dynamic cookies registered.

The first one will prevent the cache from being served if the cookies are not set, the second will create  specific cache versions depending on the cookies values.

jorditarrida:
Sorry my mistake. Already fixed in a new commit. I also moved the file to 3rd-party/plugins/cookies following #1219.

PR #1211, opened 2018-09-28
Fixed WeePie Cookie Allow plugin compatibility

webRtistik:
This Pull request fixes the compatibility of our our [WeePie Cookie Allow plugin](https://codecanyon.net/item/weepie-cookie-allow-easy-complete-cookie-consent-plugin/10342528),

--

Lately we've improved our **WeePie Cookie Allow plugin** regarding **caching compatibility** by moving the PHP logic to the JavaScript. In short this means:

- The bar/box is now in depended form the server: the JavaScript decides, based on the cookie values, to show/not show the bar/box
- The blocking/un-blocking of content in the HTML (scripts, iFrames, etc) is 'prepared' server-side, but the JavaScript decides again to block/un-block the content

This means that, also with cached pages, our WeePie Cookie Allow plugin should function well. 

The only thing is that the cache (I think) should be cleared when:
- a user saves one of the settings tabs of our plugin (because they can impact the frontend like colors, texts and cookie settings)
- a user activate/de-activate our plugin (optional because there is also the message from WP Rocket already: **WP Rocket: One or more plugins have been enabled or disabled, clear the cache if they affect the front end of your site.**)

Thanks in advance!

### DIFF
--- a/inc/3rd-party/plugins/cookies/weepie-cookie-allow.php
+++ b/inc/3rd-party/plugins/cookies/weepie-cookie-allow.php
@@ -7,11 +7,10 @@ defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
  * @since 2.9
  */
 if ( class_exists( 'WpieCookieAllow' ) ) :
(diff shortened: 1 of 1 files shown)


webRtistik:
Thanks!

PR #1218, opened 2018-10-02
Fixes #1199: the fatal staging debug logger

### LINKED ISSUE #1199: PHP Fatal Error: Composer Autoloader Cannot declare class
Over the last few days we have been seeing a PHP fatal error when WP Rocket is activated on some hosts. 

`
Fatal error: Cannot declare class ComposerAutoloaderInit92972d118ced8ff642fe76cb6658cf52, because the name is already in use in /wp-content/plugins/wp-rocket/vendor/composer/autoload_real.php on line 5
`

- https://wp-media.slack.com/archives/C43T1AYMQ/p1537456207000100
- We have seen this on multiple hosts, so it doesn't look like a host related problem.
- When I manually rename `ComposerAutoloaderInit92972d118ced8ff642fe76cb6658cf52` to something else, the error is gone. Which is strange because it suggests that someo other theme or plugin that is using composer has the same class name. (Very unlucky if that is the case)

Helpscout tags: `autoloader error` and `php fatal error`

Screenfeed:
Create a separate class autoloader for the debug logger. This will prevent fatal errors when advanced-cache.php from a staging environment is pushed to the live environment.
This file still must be refreshed afterwards, but it will prevent instant failure.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,6 +51,7 @@
 		],
 		"exclude-from-classmap": [
 			"vendor/monolog/monolog/src/",
+			"vendor/psr/log/",
 			"inc/classes/class-logger.php",
 			"inc/classes/class-secure-stream-logger.php"
 		]
--- a/composer.json
+++ b/composer.json
@@ -48,6 +48,11 @@
 		"classmap": [
 			"inc/classes",
 			"inc/vendors/classes"
+		],
+		"exclude-from-classmap": [
+			"vendor/monolog/monolog/src/",
+			"inc/classes/class-logger.php",
+			"inc/classes/class-secure-stream-logger.php"
 		]
 	}
 }
--- a/composer.json
+++ b/composer.json
@@ -48,12 +48,6 @@
 		"classmap": [
 			"inc/classes",
 			"inc/vendors/classes"
-		],
-		"exclude-from-classmap": [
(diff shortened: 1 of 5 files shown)


Tabrisrp:
Is it necessary to create a custom autoload for the logging in main.php and exclude them from the composer autoload?

I believe if the classes are already loaded, they won't be loaded a second time by composer, so the default behaviour should be fine. Did you test?

Screenfeed:
Indeed there is no "need" to exclude them from composer and duplicate the custom loader. I did this for 1 reason: we have a readable version of the custom loader. I also thought it would help to get warned when we forget to add a file to this autoloader, but actually it won't help at all (since this forgotten file will probably in composer).

Now with a bit of hindsight I can see a better way:
- Create a file `process-autoloader.php` in plain php, containing the autoloader, so it is easily readable.
- `file_get_contents()` on it and add its content to `advanced-cache.php`. Or, add a `include 'process-autoloader.php';` to it instead.
- Then no need to add it to `main.php`, nor remove the classes from composer autoloader.

What do you think?

Tabrisrp:
Agreed it seems to be much better to go that direction

Screenfeed:
Ready 👍

PR #1223, opened 2018-10-03
Fixes #78 Allow serving mobile cache version to tablets

### LINKED ISSUE #78: Serve mobile cache for tablet for compatibility with wp_is_mobile()

Tabrisrp:
Introduces a new filter to activate serving the mobile cache version to tablets if required

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -99,6 +99,17 @@ function get_rocket_config_file() {
 		$buffer .= '$rocket_common_cache_logged_users = 1;' . "\n";
 	}
 
+	/**
+	 * Filters the use of the mobile cache version for tablets
+	 * 0 will serve desktop to tablets, 1 will serve mobile to tablets
+	 *
+	 * @since 3.2
+	 * @author Remy Perona
+	 *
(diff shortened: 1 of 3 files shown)


inc/functions/files.php
```diff
@@ -99,6 +99,17 @@ function get_rocket_config_file() {
...
+	/**
+	 * Filters the use of the mobile cache version for tablets
+	 * 0 will serve desktop to tablets, 1 will serve mobile to tablets
+	 *
+	 * @since 3.2
+	 * @author Remy Perona
+	 *
+	 * @param int $mobile_for_tablet valid values are 0 or 1.
+	 */
+	$buffer .= '$rocket_cache_mobile_files_tablet = ' . apply_filters( 'rocket_cache_mobile_files_tablet', 0 ) . ';' . "\n";
```
Screenfeed: Can't we use something like this instead?

```php
$rocket_cache_tablet = 'mobile';
$rocket_cache_tablet = 'desktop';
```

I think it can be understood more easily.

PR #1233, opened 2018-10-08
Closes #1087 Clear Cache On Theme Update

### LINKED ISSUE #1087: Automatically clear WP Rocket cache on theme updates
While its safe to assume that clearing WP Rocket cache during plugin update is not required in many cases (for example, updating Akismet as @GeekPress pointed out), updating a theme is very likely to need a cache flush. 

Right now, when the theme is updated, no notice is displayed in the admin suggesting that a cache clear might be needed. I think most theme updates would need a cache clean. 

A helper plugin to make this happen until this feature is available in core: https://www.dropbox.com/s/curme88kmgb08al/wp-rocket-helper-clear-cache-theme-upgrade.zip?dl=1

Related ticket: https://secure.helpscout.net/conversation/626671965/74866?folderId=1724284

Volnus:
This clears the cache when themes or plugins are updated.

Reference https://developer.wordpress.org/reference/hooks/upgrader_process_complete/
Reference https://developer.wordpress.org/reference/classes/theme_upgrader/

#1087 - Ticket referencing the issue. 

Some notes, it's totally possible to only do this when THEMES are updated by changing the parameters see the above references. However, doing so seems like a half-baked solution especially when updating things that affect the front-end and there are many plugins that do this. It's safer, in my opinion, to clear the cache on both plugin & theme updates. 

Especially since there is no prompt to clear the cache when plugins or themes are updated.

### DIFF
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -19,7 +19,7 @@ add_action( 'edit_link'                                             , 'rocket_cl
 add_action( 'delete_link'                                           , 'rocket_clean_domain' );  // When a link is deleted.
 add_action( 'customize_save'                                        , 'rocket_clean_domain' );  // When customizer is saved.
 add_action( 'update_option_theme_mods_' . get_option( 'stylesheet' ), 'rocket_clean_domain' ); // When location of a menu is updated.
-
+add_action( 'upgrader_process_complete'                             , 'rocket_clean_domain' );  // When a theme or plugin is updated.
 /**
  * Purge cache When a widget is updated
  *


Tabrisrp:
Doing this automatically on theme update makes sense, as it will almost always impact the front-end.

On plugins update, it's trickier, as it doesn't necessarily mean the front-end will be impacted. We try to clean the cache as little as possible to preserve the performance.

Instead for this case, we could do the same as we do on plugin activation/deactivation, and display a notice to invite the user to clear the cache if needed.

On a code note, `rocket_clean_domain` should not be used as the callback for this hook, as the first parameter of the function doesn't match with what is provided by the action. There should be another function serving as wrapper instead.

inc/common/purge.php
```diff
@@ -19,7 +19,7 @@
 add_action( 'delete_link'                                           , 'rocket_clean_domain' );  // When a link is deleted.
 add_action( 'customize_save'                                        , 'rocket_clean_domain' );  // When customizer is saved.
 add_action( 'update_option_theme_mods_' . get_option( 'stylesheet' ), 'rocket_clean_domain' ); // When location of a menu is updated.
-
+add_action( 'upgrader_process_complete'                             , 'rocket_clean_domain' );  // When a theme or plugin is updated.
```
Tabrisrp: `rocket_clean_domain` can't be used as the direct callback for this, as the parameters of the hook and the function don't match.

Also we don't want to auto clean for plugins update, so we need to check in the callback function that the  type of update is `theme` before triggering the clean.

Tabrisrp:
I'd like to add this in 3.2.1, but changes need to be made before that, as I wrote in the review

Tabrisrp:
Closing in favor of #1437

PR #1237, opened 2018-10-12
Custom logger html formatter

Screenfeed:
It's a custom HTML formatter for the debug logger that removes the "Channel" row.
Also grouped classes related to the logger into a new folder/namespace.


Screenfeed:
I think it will be simpler to review commit by commit.

PR #1321, opened 2018-11-09
Homepage preload fix

foliovision:
Our fix for the homepage preload issues #1322


inc/functions/formatting.php
```diff
@@ -109,15 +109,22 @@ function rocket_remove_url_protocol( $url, $no_dots = false ) {
...
  *
  * @param string $url The URL to parse.
  * @return string $url The URL with protocol
  */
 function rocket_add_url_protocol( $url ) {
-	if ( strpos( $url, 'http://' ) === false && strpos( $url, 'https://' ) === false ) {
+  
+	$link_data = wp_parse_url( $url );
+
+	if( empty($link_data['host']) ) {
```
Tabrisrp: This is making the function have 2 different goals, which is problematic, and the name becomes misleading.

This should be either done directly in the homepage preload method, or with a new function.

foliovision:
Done, the rocket_add_url_protocol() remains unaffected now.

inc/functions/formatting.php
foliovision: Done, the rocket_add_url_protocol() remains unaffected now.

PR #1328, opened 2018-11-09
Add support for the impreza theme

TheZoker:
This adds support for the impreza theme:
https://themeforest.net/item/impreza-retina-responsive-wordpress-theme/6434280

### DIFF
--- a/inc/3rd-party/themes/impreza.php
+++ b/inc/3rd-party/themes/impreza.php
@@ -18,7 +18,7 @@ if ( 'impreza' === strtolower( $current_theme->get( 'Name' ) ) || 'impreza' === 
 
 		return $excluded_js;
 	}
-    add_filter( 'rocket_exclude_js', 'rocket_exclude_js_impreza' );
-    add_filter( 'rocket_exclude_defer_js', 'rocket_exclude_js_impreza' );
-    add_filter( 'rocket_exclude_cache_busting', 'rocket_exclude_js_impreza' );
+	add_filter( 'rocket_exclude_js', 'rocket_exclude_js_impreza' );
+	add_filter( 'rocket_exclude_defer_js', 'rocket_exclude_js_impreza' );
+	add_filter( 'rocket_exclude_cache_busting', 'rocket_exclude_js_impreza' );
 }
--- a/inc/3rd-party/themes/impreza.php
+++ b/inc/3rd-party/themes/impreza.php
@@ -1,0 +1,24 @@
+<?php
+defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
+
+$current_theme = wp_get_theme();
+
+if ( 'impreza' === strtolower( $current_theme->get( 'Name' ) ) || 'impreza' === strtolower( $current_theme->get( 'Template' ) ) ) {
+	/**
+	 * Excludes Impreza files from minification/combine, defer and cache busting
+	 *
+	 * @since 3.2
+	 * @author TheZoker
+	 *
+	 * @param array $excluded_js Array of JS filepaths to be excluded.
+	 * @return array
+	 */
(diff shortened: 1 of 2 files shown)


TheZoker:
@Tabrisrp What exactly needs to be changed?

inc/3rd-party/themes/impreza.php
```diff
@@ -0,0 +1,24 @@
...
+	/**
+	 * Excludes Impreza files from minification/combine, defer and cache busting
+	 *
+	 * @since 3.2
+	 * @author TheZoker
+	 *
+	 * @param array $excluded_js Array of JS filepaths to be excluded.
+	 * @return array
+	 */
+	function rocket_exclude_js_impreza( $excluded_js ) {
```
Tabrisrp: too much indentation here

TheZoker: I think it's too less on the `add_filter` lines instead. I fixed it

arunbasillal:
@TheZoker Hello, I am trying to test it out using the demo website provided by theme forest. 

I cannot find `framework/js/us.core.min.js` in the source of the uncached page. Is there anything that I should enable to replicate the issue?

TheZoker:
@arunbasillal It seems, that this has been changed since I've created this PR.
I fix the path within this PR:
https://github.com/wp-media/wp-rocket/pull/1932

arunbasillal:
@TheZoker Thanks for the update. 
However, I still cannot find it in the source code. You can see the demo website here - http://impreza-ceklr69.themetest.net/ (expires in about 6 hours)

I am looking for `js/us.core.min.js` in the source-code and it doesn't exit. 

If you want to see the uncached version, please see - http://impreza-ceklr69.themetest.net/?nocache

I feel like I am missing something here. Please let me know what I missed.

TheZoker:
The file is right here (this is the main demo):
http://impreza.us-themes.com/wp-content/themes/Impreza/js/us.core.min.js

In your installation the file is also available:
http://impreza-ceklr69.themetest.net/wp-content/themes/Impreza/js/us.core.min.js?nocache

arunbasillal:
Thanks for the quick reply. The file does exist for sure.

What I cannot find is where the file is being used. 

- It's absent in the source code for the homepage - https://jmp.sh/0JUdv8s
- Or when I search for it - https://jmp.sh/PDn9oya

Can you advise where it is used?

TheZoker:
But you're right, it seems that the file itself is not included anymore. It also seems, that the file is included in another file, that is located here:
http://impreza.us-themes.com/wp-content/uploads/us-assets/impreza.us-themes.com.js

I will speak with the developers about that file and get back

arunbasillal:
Thanks!

I think it would be better to add this compatibility at your end. I see that there is compatibility for gravity forms, polylang and other plugins. 

This will help your developers change and update files at their pace rather than send in a PR each time. Let's say the file changes again in the future, things would break for Impreza's and WP Rocket's customers. Since Impreza's developers are the ones making the change, it would be easy for them to change the compatibility code at your end. 

Would you propose that to them and see what they think please?

thekendog:
@TheZoker there is an option in the theme settings to optimize CSS and JS size.  When you do this the file is located at /wp-content/uploads/us-assets/www.siteurl.com.js.  Perhaps adding something like this would work?

$uploads = wp_get_upload_dir();
$protocols = array('http://', 'https://');
$excluded_js[] = rocket_clean_exclude_file( $uploads['baseurl'] . '/us-assets/' . str_replace($protocols, '', site_url()) . '.js' );

arunbasillal:
We decided to remove this from 3.4. It's best to add this to the theme itself as mentioned here - https://github.com/wp-media/wp-rocket/pull/1328#issuecomment-527410346

PR #1339, opened 2018-11-12
Correctly check for https cache version during preload

Tabrisrp:
We were not checking for the correct cache file in the preload when the website was using https. This would cause requests to be made even though the file is already in cache.

### DIFF
--- a/inc/classes/preload/class-full-process.php
+++ b/inc/classes/preload/class-full-process.php
@@ -98,7 +98,7 @@ class Full_Process extends \WP_Background_Process {
 			$url['host'] = str_replace( '.', '_', $url['host'] );
 		}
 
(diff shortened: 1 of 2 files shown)


inc/classes/preload/class-full-process.php
```diff
@@ -85,14 +85,20 @@ protected function task( $item ) {
 	 * @return bool
 	 */
 	protected function is_already_cached( $item ) {
+		static $https;
+
+		if ( ! isset( $https ) ) {
+			$https = ( is_ssl() && get_rocket_option( 'cache_ssl' ) ) ? '-https' : '';
```
Screenfeed: `$https` is not used.
Maybe it should be `'index' . $https . '.html'` later then?

PR #1350, opened 2018-11-13
Don't preload excluded pages and improve homepage URLs discovery

Tabrisrp:
- Remove excluded from cache pages from all preload
- Improve homepage URLs discovery by removing URLs to files, fragment URLs, URLs to excluded from cache pages
- Fix partial preload already cached method

### DIFF
--- a/inc/classes/preload/class-homepage.php
+++ b/inc/classes/preload/class-homepage.php
@@ -18,6 +18,33 @@ class Homepage extends Abstract_Preload {
 	 * @return void
 	 */
 	public function preload( $urls ) {
+		foreach ( $urls as $home_url ) {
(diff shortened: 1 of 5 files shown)


inc/classes/preload/class-homepage.php
```diff
@@ -18,6 +18,33 @@ class Homepage extends Abstract_Preload {
 	 * @return void
 	 */
 	public function preload( $urls ) {
+		foreach ( $urls as $home_url ) {
+			$urls      = $this->get_urls( $home_url );
```
Screenfeed: This `$urls` is confusing with the one above. It should be something like `$home_urls` imho.

Issue #1379, opened 2018-11-22
Rank Math SEO Plugin's Preload


MyThemeShop:
**Is your feature request related to a problem? Please describe.**
This feature was requested here: https://wordpress.org/support/topic/function-to-get-sitemap-url/

**Describe the solution you'd like**
An option in the plugin as shown here: https://mythemeshop.d.pr/BRH8Sf

PR #1380, opened 2018-11-22
 Added support for Rank Math SEO Plugin's Sitemap

MyThemeShop:


MyThemeShop:
Related to #1379

Tabrisrp:
Thank you for the PR, it looks all good, so we're going to incorporate it into 3.2.3 👍

PR #1397, opened 2018-11-29
Exclude Report Broken Link JS

ScottTravisHartley:
The following plugin: https://wordpress.org/plugins/report-broken-links/

This plugin includes inline JS that includes that page's URL (for obvious reasons) this simply excludes this inline JS to avoid, the cache from generating a unique JS file per page / post / archive / etc


Tabrisrp:
Thank you for the PR!

PR #1403, opened 2018-12-01
Fixes #1388: Logged-in user cache: customizer link in admin bar is hidden

### LINKED ISSUE #1388: Logged-in user cache: customizer link in admin bar is hidden
**Describe the bug**
When logged-in user cache is active, the customize link in the admin bar is hidden.

This is happening because of the method injecting back the IE conditionals in the buffer, in the `IE_Conditionals_Subscriber` class.

the `preg_replace()`  function is stripping the double backslashes in one inline JS for IE, transforming `\\s` into `\s`.

The conditionals content should be escaped before being used by `preg_replace()` to prevent this.

Screenfeed:


Screenfeed:
[Test](https://regex101.com/r/vYMMgY/1).

PR #1409, opened 2018-12-03
Fixes #1370 Prevent PHP warning and notice on preload

### LINKED ISSUE #1370: PHP Warning: class-homepage.php
Several customers are reporting this PHP Warning: 
`Warning: Invalid argument supplied for foreach() in /wp-content/plugins/wp-rocket/inc/classes/preload/class-homepage.php on line 25`

HS tag: `class-homepage.php`

Tabrisrp:


inc/classes/preload/class-homepage.php
```diff
@@ -19,7 +19,12 @@ class Homepage extends Abstract_Preload {
 	 */
 	public function preload( $home_urls ) {
 		foreach ( $home_urls as $home_url ) {
-			$urls      = $this->get_urls( $home_url );
+			$urls = $this->get_urls( $home_url );
+
+			if ( ! $urls || empty( $urls ) ) {
```
Screenfeed: No need for both tests, they do the same thing.
```if ( ! $urls ) {```

File: inc/classes/preload/class-homepage.php [hunk 2]
```diff
@@ -57,7 +62,7 @@ private function get_urls( $url ) {
 		$response = wp_remote_get( $url, $args );
 
 		if ( 200 !== wp_remote_retrieve_response_code( $response ) ) {
-			return;
+			return false;
```
Screenfeed: `@return` tag should be updated then.

PR #1413, opened 2018-12-03
Closes #1361 Add new notice about WP Rocket Footer JS

### LINKED ISSUE #1361: Add Rocket Footer JS to list of plugins that trigger warning
When https://wordpress.org/plugins/rocket-footer-js/ is enabled, it deliberately prevents our Deferred JS options being activated. 
We should display a warning to deactivate this plugin.

Tabrisrp:


pcfreak30:
@Tabrisrp @Screenfeed 

Hello, a client of mine just pinged me about this. May I ask why this was felt to be needed? My plugin has been promoted as un-official from day one and so to me, this just only serves to scare some non-technical people off?

If there is an actual problem it would have been more productive to open an issue in my GitHub to resolve it.

Thanks!

Tabrisrp:
@pcfreak30 @webtrainingwheels will be able to provide more details as to why support felt this was necessary, but I know one issue was your plugin preventing the defer JS option from being ever checked in WPR, which confused users and made them open support tickets with us.

PR #1419, opened 2018-12-05
Domain cache deletion background process

dotSILENT:
As for now, Clear Cache (`rocket_clean_domain()` for domain cache) uses `rocket_rrmdir()` to recursively delete all the domain directories. On huge websites with thousands of cached posts, this operation can take a very long time and exceed the maximum PHP execution time.
I've added an option in the Settings/Cache page called `Domain Cache Invalidation` which is disabled by default, when enabled the domain cache directories aren't directly deleted but their name is 'invalidated' to a unique one and picked up by the Domain_Cache_Deletion_Subscriber which queues them for deletion in the background (Domain_Cache_Deletion_Process)
In case the deletion exceeds the execution time, WP Background Processing will re-run the process after 5 minutes, so everything will be eventually deleted.
If renaming wasn't possible due to any file system error (shouldn't happen unless playing with the cache files directly from the system), it falls back to the default `rocket_rrmdir()` behaviour.

I have also noticed that sometimes (at least on Windows), `rocket_rrmdir()` deletes all the files but cannot delete the directories unless you retry it a few times. The background process handles this and has a limit of max 20 retries to prevent any looping, it never took me more than 3 repeats though.

This however doesn't preserve the directories from `get_rocket_i18n_to_preserve()` since it invalidates the entire parent directories.

### DIFF
--- a/inc/admin/options.php
+++ b/inc/admin/options.php
@@ -52,6 +52,7 @@ function rocket_after_save_options( $oldvalue, $value ) {
 		'analytics_enabled'           => true,
 		'sucury_waf_cache_sync'       => true,
 		'sucury_waf_api_key'          => true,
+		'invalidate_domain_cache'     => true,
 	];
 
 	// Create 2 arrays to compare.
--- a/inc/admin/upgrader.php
+++ b/inc/admin/upgrader.php
@@ -153,6 +153,7 @@ function rocket_first_install() {
 				'cache_ssl'                   => rocket_is_ssl_website() ? 1 : 0,
 				'emoji'                       => 1,
 				'embeds'                      => 1,
+				'invalidate_domain_cache'     => 1,
 				'cache_reject_uri'            => [],
 				'cache_reject_cookies'        => [],
 				'cache_reject_ua'             => [],
@@ -381,6 +382,13 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 
 	if ( version_compare( $actual_version, '3.2.2', '<' ) ) {
 		flush_rocket_htaccess();
+	}
+
+	if( version_compare( $actual_version, '3.2.3', '< ') ) {
+		$options = get_option( WP_ROCKET_SLUG );
+		$options['invalidate_domain_cache'] = 0;
+
+		update_option( WP_ROCKET_SLUG, $options );
 	}
 }
 add_action( 'wp_rocket_upgrade', 'rocket_new_upgrade', 10, 2 );
--- a/inc/admin/upgrader.php
+++ b/inc/admin/upgrader.php
@@ -153,7 +153,6 @@ function rocket_first_install() {
 				'cache_ssl'                   => rocket_is_ssl_website() ? 1 : 0,
 				'emoji'                       => 1,
 				'embeds'                      => 1,
-				'invalidate_domain_cache'     => 1,
 				'cache_reject_uri'            => [],
 				'cache_reject_cookies'        => [],
 				'cache_reject_ua'             => [],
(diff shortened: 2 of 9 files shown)


Surgeonito:
I humbly insist on giving high priority to this PR. It's crucial must-have for all high traffic websites with large post's count.

Tabrisrp:
Thank you for the PR, it's something I tought about before and that could help a lot indeed.

As it's a major change, it won't be included in a minor version. It could be added in the next major version, we'll have to review and test it.

> This however doesn't preserve the directories from get_rocket_i18n_to_preserve() since it invalidates the entire parent directories.

The current behaviour should be preserved if you want this change included in.

Surgeonito:
Hi guys.
I'm wondering if this is the right approach.
The problem we're addressing here is the case of huge, high-traffic websites with tons of posts.
Right now this code changes the name of all the folders inside the domain folder. Which is again - more that 10000 folders to process by php script. Which is not good, not at all.

Why not just renaming entire domain folder?
I know it will also purge all the languages in case when WordPress is in multilang mode. But this feature will be used by users on purpose by enabling this in the settings screen. So everyone will be aware of the consequences. 
@Tabrisrp what do you say?

dotSILENT:
I made a commit to address the issue above, the problem was in how I passed the paths to directories to the background process, now they are relative to the cache directory, so nested directories can also be invalidated.
The way it's done now, ``get_rocket_i18n_to_preserve`` could be easily added again, however after checking into it I noticed that the actual code in ``rocket_rrmdir`` isn't working at all, at least not on Windows. I'm not sure but here are the issues I think I found:

1. ``get_rocket_i18n_to_preserve`` creates an invalid regex pattern, at least on Windows since it uses different slashes for file paths. The pattern wasn't working at all, I fixed it by adding ``preg_quote`` for escaping:

```php
$langs_to_preserve[] = preg_quote( WP_ROCKET_CACHE_PATH . $parse_url['host'] ) . '(.*)' . preg_quote( '/' . trim( $parse_url['path'], '/' ) );
```
(replaces line 310 in inc/functions/i18n.php)
2. Even with the correct regex above, it's still not working. Why?
The directories from the domain directory, obtained via ``glob`` in ``rocket_clean_domain`` (inc/functions/files.php) are passed into ``rocket_rrmdir``, these directories actually contain the directories from ``get_rocket_18n_to_preserve``, in example:
``C:\web\wordpress/wp-content/cache/wp-rocket/wp.test/en``
The regex pattern returned from `get_rocket_i18n_to_preserve` would look something like this:
``C:\web\wordpress/wp-content/cache/wp-rocket/wp.test(.*)/en``
But, ``rocket_rrmdir`` doesn't check the pattern (``preg_grep``) for the directory passed to it (``$dir``), but it's contents (after using ``glob`` with appended /* to the ``$dir`` path)
Therefore the pattern isn't matching at all, I got it working by modifying ``get_rocket_i18n_to_preserve`` again and adding a ``/(.*)`` at the end, so any path that starts with ``.../en/`` will match:
```php
$langs_to_preserve[] = preg_quote( WP_ROCKET_CACHE_PATH . $parse_url['host'] ) . '(.*)' . preg_quote( '/' . trim( $parse_url['path'], '/' ) . '/' ) . '(.*)';
```

I'm not really sure about that though, so I haven't done any commits yet. In my opinion this function isn't really that important, especially if it wasn't working at all, even if the cache is deleted it will be generated again, I may be wrong though.

Surgeonito:
@Tabrisrp any plans for this PR?

Tabrisrp:
As I said previously, this will have to wait until the next major version at least, which currently doesn't have an ETA. I'll update when we make a decision on that.

PR #1436, opened 2018-12-14
Fixes #757 Correctly rewrite url() path in CSS

### LINKED ISSUE #757: Relative URLs without leading slash (folder relative instead of root relative) replaced incorrectly during minification
Relative references like this: `background: url(images/break-f.png) right repeat-y;`

Instead of `background: url(/images/break-f.png) right repeat-y;` <-- root-relative with leading `/`

Minification turns it into `/xxxxx/wp-content/themes/DynamiX/images/break-f.png`
Folder named `xxxxx` is the folder in the server where the WordPress is installed. 

Related tickets: 
- https://secure.helpscout.net/conversation/488092939/55631/
- https://secure.helpscout.net/conversation/596683842/70638/?folderId=273766
- https://secure.helpscout.net/conversation/619863450/73838/?folderId=273764

Tabrisrp:
Use Converter class of MatthiasMullie\PathConverter instead of the old Minify_CSS_Rewriter class.

We no longer rely on `DOCUMENT_ROOT` to perform the conversion, but on 2 absolute paths instead: the source path and the target path.


inc/classes/optimization/CSS/trait-path-rewriter.php
```diff
@@ -14,20 +16,155 @@
...
+		// find all relative urls in css.
+		$matches = array();
+		foreach ( $relativeRegexes as $relativeRegex ) {
+			if ( preg_match_all( $relativeRegex, $content, $regexMatches, PREG_SET_ORDER ) ) {
+				$matches = array_merge( $matches, $regexMatches );
+			}
+		}
+
+		$search  = array();
+		$replace = array();
```
Screenfeed:
```php
$search  = [];
$replace = [];
```

inc/classes/optimization/CSS/trait-path-rewriter.php
```diff
@@ -14,20 +16,155 @@
...
+					(?P<path>.+?)
+
+				# close path enclosure
+				(?P=quotes)
+
+			/ix',
+		);
+
+		// find all relative urls in css.
+		$matches = array();
```
Screenfeed:
```php
$matches = [];
```

inc/classes/optimization/CSS/trait-path-rewriter.php
```diff
@@ -14,20 +16,155 @@
...
+			 * @see https://github.com/matthiasmullie/minify/issues/193
+			 */
+			$url = trim( $url );
+			if ( preg_match( '/[\s\)\'"#\x{7f}-\x{9f}]/u', $url ) ) {
+				$url = $match['quotes'] . $url . $match['quotes'];
+			}
+
+			// build replacement.
+			$search[] = $match[0];
+			if ( $type === 'url' ) {
```
Screenfeed:
```php
'url' === $type
```

inc/classes/optimization/CSS/trait-path-rewriter.php
```diff
@@ -14,20 +16,155 @@
...
+				$replace[] = 'url(' . $url . ')';
+			} elseif ( $type === 'import' ) {
```
Screenfeed:
```php
'import' === $type
```

PR #1437, opened 2018-12-14
Closes #1087 Clear cache on current theme update

### LINKED ISSUE #1087: Automatically clear WP Rocket cache on theme updates

remyperona:


inc/common/purge.php
```diff
@@ -441,7 +442,32 @@ function do_admin_post_rocket_purge_cloudflare() {
...
+		return;
+	}
+
+	if ( 'theme' !== $hook_extra['type'] ) {
+		return;
+	}
+
+	$current_theme = wp_get_theme();
+
+	if ( ! in_array( $current_theme->get_template(), $hook_extra['themes'], true ) ) {
```
Screenfeed: In case of a child theme, we should check for both parent and child.
To test, but this should do the trick:
```php
$themes = [
	$current_theme->get_template(), // Parent theme.
	$current_theme->get_stylesheet(), // Child theme.
];

if ( ! array_intersect( $hook_extra['themes'], $themes ) ) {
	return;
}
```

Screenfeed (approved):

cemerson:
_Does this logic kick in if, for example, the header.php of a theme is updated? I just tested and it appears as if the cache is not dumping/refreshing._

Update: Scratch that - it may have just worked. I updated the header.php via FTP, then I updated it via the built-in WP theme file editor and I think the cache refreshed then. I was in the middle of using the Appearance > Customize menu to add some custom CSS when I think I saw the cache reset. Anyhow - just FYI.

PR #1442, opened 2018-12-17
Fixes #804 Prevent fatal error if Polylang function is not available

### LINKED ISSUE #804: Add check for pll_get_post_language() to prevent error
Sample ticket: https://secure.helpscout.net/conversation/539674804/62247?folderId=377611
(contains full error)

Customer gets fatal error when using WPLR Sync plugin because, presumably, the pll_get_post_language() function isn't available when WPLR is triggering. 
Add check with function_exists to prevent that.

Tabrisrp:


inc/common/purge.php
```diff
@@ -80,15 +80,16 @@ function rocket_clean_post( $post_id ) {
 
 	// Get the post language.
 	$i18n_plugin = rocket_has_i18n();
+	$lang        = false;
 
 	if ( 'wpml' === $i18n_plugin && ! rocket_is_plugin_active( 'woocommerce-multilingual/wpml-woocommerce.php' ) ) {
 		// WPML.
 		$lang = $GLOBALS['sitepress']->get_language_for_element( $post_id, 'post_' . get_post_type( $post_id ) );
 	} elseif ( 'polylang' === $i18n_plugin ) {
```
Screenfeed: We all love pyramids but:
```php
} elseif ( 'polylang' === $i18n_plugin && function_exists( 'pll_get_post_language' ) ) {
```

PR #1453, opened 2018-12-19
Fixes #431 WPML returning incorrect home_url() value on the admin

### LINKED ISSUE #431: Plugin Compat : WPML’s option “Use directory for default language”
WP Rocket doesn’t work with WPML’s option “Use directory for default language”.

PHP Config files in /wp-content/wp-rocket-config/ are created for each languages but the extension of the tld is duplicate (e.g. : /domain.fr.fr.php & /domain.es.es.php) so pages are not cached.

Tabrisrp:


inc/functions/i18n.php
```diff
@@ -478,3 +478,33 @@ function get_rocket_i18n_translated_post_urls( $post_id, $post_type = 'page', $r
...
+/**
+ * Returns the home URL, without WPML filters if the plugin is active
+ *
+ * @since 3.2.4
+ * @author Remy Perona
+ *
+ * @param string $path Path to add to the home URL.
+ * @return string
+ */
+function rocket_get_home_url( $path = '' ) {
```
Screenfeed: Two issues here:
- `rocket_has_i18n()` can't be trusted here: even if `$sitepress->get_active_languages()` exists, that doesn't mean that `$wpml_url_filters->home_url_filter()` does (old versions). I could be wrong though, maybe it's there for a long time, but it would require to verify in old versions (maybe you did? Personally I wouldn't have ^^.
- Your use of `$home_url` is wrong, it will store the same home URL for any value of `$path`.

I's prefer something like that:
```php
function rocket_get_home_url( $path = '' ) {
	global $wpml_url_filters;
	static $home_url = [];
	static $has_wpml;

	if ( isset( $home_url[ $path ] ) ) {
		return $home_url[ $path ];
	}

	if ( ! isset( $has_wpml ) ) {
		$has_wpml = $wpml_url_filters && is_opject( $wpml_url_filters ) && method_exists( $wpml_url_filters, 'home_url_filter' );
	}

	if ( $has_wpml ) {
		remove_filter( 'home_url', [ $wpml_url_filters, 'home_url_filter' ], -10 );
	}

	$home_url[ $path ] = home_url( $path );

	if ( $has_wpml ) {
		add_filter( 'home_url', [ $wpml_url_filters, 'home_url_filter' ], -10, 4 );
	}

	return $home_url[ $path ];
}
```

But what I'd prefer more, is to see all code related to WPML in `/inc/classes/subscriber/third-party/plugins/multilang/class-wpml-subscriber.php`.

Tabrisrp: Subscriber would be good but things are too mixed up currently to split it correctly in a minor version imo.

PR #1454, opened 2018-12-19
Closes #1366 Improve license validation messaging

### LINKED ISSUE #1366: License Validation - Detect when connection is blocked by firewall or security plugins
A common issue with license validation is the connection to our validation system is blocked by a security plugin. 

Right now, we show the notice and point them to our doc - https://docs.wp-rocket.me/article/100-resolving-problems-with-license-validation

file-qfu04rjlwy

Since the validation error says to contact support, they often contact us and all we can do is tell them that it's a connection issue. Which can be automated. 

- We can definitely confirm if it's a connection issue or not. If it's a connection issue, displaying that info to the user directly in the UI and asking them to whitelist our IP or contact their host would be better. 
- We could go one-step further and check to see if any of the security plugins that often cause issues are installed and point that out to the user as well with a solution.

Tabrisrp:
- Save the error message(s) encountered in a transient
- Display them in a notice if validation can't be done

### DIFF
--- a/inc/functions/admin.php
+++ b/inc/functions/admin.php
@@ -7,10 +7,20 @@ defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
  * @since 1.0
  */
 function rocket_need_api_key() {
+	$message = '';
+	$errors  = get_transient( 'rocket_check_key_errors' );
+
+	if ( false !== $errors ) {
+		foreach ( $errors as $error ) {
+			$message .= '<p>' . esc_html( $error ) . '</p>';
+		}
+	}
+
 	?>
 	<div class="notice notice-warning">
-		<p><strong><?php echo WP_ROCKET_PLUGIN_NAME; ?></strong>: <?php _e( 'There seems to be an issue with outgoing connections from your server. Resolve per documentation, or contact support.', 'rocket' ); ?>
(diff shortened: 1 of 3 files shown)


inc/functions/admin.php
```diff
@@ -7,10 +7,20 @@
...
+	if ( false !== $errors ) {
+		foreach ( $errors as $error ) {
+			$message .= '<p>' . esc_html( $error ) . '</p>';
+		}
+	}
+
 	?>
 	<div class="notice notice-warning">
-		<p><strong><?php echo WP_ROCKET_PLUGIN_NAME; ?></strong>: <?php _e( 'There seems to be an issue with outgoing connections from your server. Resolve per documentation, or contact support.', 'rocket' ); ?>
+		<p><strong><?php echo esc_html( WP_ROCKET_PLUGIN_NAME ); ?></strong>: <?php esc_html_e( 'There seems to be an issue validating your license. You can see the error message(s) below. Resolve per documentation, or contact support.', 'rocket' ); ?>
```
Screenfeed: Suggestion:
```php
<p>
	<?php
	echo esc_html(
		sprintf(
			/* translators: 1 and 2 can be... anything. */
			__( '%1$s: %2$s', 'rocket' ),
			'<strong>' . WP_ROCKET_PLUGIN_NAME . '</strong>',
			_n( 'There seems to be an issue validating your license. You can see the error message below. Resolve per documentation, or contact support.', 'There seems to be an issue validating your license. You can see the error messages below. Resolve per documentation, or contact support.', count( $errors ), 'rocket' )
		)
	);
	?>
</p>
```

It would allow to translate the `:`, notably the space character placed before it in some languages (like, you know, French). And maybe it's better for RTL languages.
Also, `_n()`.

PR #1456, opened 2018-12-20
Closes #1377 Display error messages when preload fails

### LINKED ISSUE #1377: Provide UI Feedback for Preload Fails
Preload fails for a variety of reasons and to find them we normally have to access the site to debug. Could we provide some feedback in the UI as to why it's failing?

As an example, while not perfect, is the messaging we have for the CPCSS generation, which is helpful. It gives the customer something concrete they can provide us with, and we have some ideas about the cause.

Tabrisrp:
Add a transient to store the error messages when an URL can't be reached or the content is not good, and display them in the preload running notice.

### DIFF
--- a/inc/classes/preload/class-homepage.php
+++ b/inc/classes/preload/class-homepage.php
@@ -61,7 +61,21 @@ class Homepage extends Abstract_Preload {
 
 		$response = wp_remote_get( $url, $args );
 
+		$errors = get_transient( 'rocket_preload_errors' );
+
+		if ( is_wp_error( $response ) ) {
+			// Translators: %1$s is an URL, %2$s is the error message.
+			$errors['errors'][] = sprintf( __( 'Could not gather links on %1$s because of the following error: %2$s', 'rocket' ), $url, $response->get_error_message() );
+
+			set_transient( 'rocket_preload_errors', $errors );
+			return false;
+		}
+
 		if ( 200 !== wp_remote_retrieve_response_code( $response ) ) {
+			// Translators: %1$s is an URL, %2$s is the HTTP response code.
+			$errors['errors'][] = sprintf( __( 'Could not gather links on %1$s because it returned the following response code: %2$s', 'rocket' ), $url, wp_remote_retrieve_response_code( $response ) );
+
+			set_transient( 'rocket_preload_errors', $errors );
 			return false;
 		}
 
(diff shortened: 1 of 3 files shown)


inc/classes/preload/class-homepage.php
```diff
@@ -61,7 +61,21 @@ private function get_urls( $url ) {
 
 		$response = wp_remote_get( $url, $args );
 
+		$errors = get_transient( 'rocket_preload_errors' );
```
Screenfeed: To prevent notices:
```php
$errors           = get_transient( 'rocket_preload_errors' );
$errors           = is_array( $errors ) ? $errors : [];
$errors['errors'] = isset( $errors['errors'] ) && is_array( $errors['errors'] ) ? $errors['errors'] : [];
```

inc/classes/preload/class-sitemap.php
```diff
@@ -92,20 +92,32 @@ public function process_sitemap( $sitemap_url, $urls = [] ) {
...
+		$args = apply_filters(
+			'rocket_preload_sitemap_request_args',
+			[
+				'user-agent' => 'WP Rocket/Sitemaps',
+				'sslverify'  => apply_filters( 'https_local_ssl_verify', true ), // WPCS: prefix ok.
+			]
+		);
 
 		$sitemap = wp_remote_get( esc_url( $sitemap_url ), $args );
+		$errors  = get_transient( 'rocket_preload_errors' );
```
Screenfeed: Same here.

inc/classes/subscriber/preload/class-preload-subscriber.php
```diff
@@ -238,10 +238,22 @@ public function notice_preload_running() {
 			return;
 		}
 
+		// translators: %1$d = Number of pages preloaded.
+		$message = '<p>' . sprintf( __( 'Preload: %1$d uncached pages have now been preloaded. (refresh to see progress)', 'rocket' ), $running ) . '</p>';
```
Screenfeed:
```php
// translators: %1$s = Number of pages preloaded.
$message = '<p>' . sprintf( _n( 'Preload: %1$s uncached page has now been preloaded. (refresh to see progress)', 'Preload: %1$s uncached pages have now been preloaded. (refresh to see progress)', $running, 'rocket' ), number_format_i18n( $running ) ) . '</p>';
```

PR #1480, opened 2019-01-15
Prepare WooCommerce product images for Lazyload

nicomollet:
It only applies to images in product loops. Not to the product image in single product page.


Tabrisrp:
Version 3.3 of the plugin will change the way the lazyload is applied, and it will automatically include images that were not previously picked up.

So I'm closing this PR as it won't be necessary anymore.

Still, thank you for your participation!

PR #1482, opened 2019-01-17
Closes #1306 Add fallback for sitemap preload

### LINKED ISSUE #1306: Preload Idea - A fallback when sitemap preload fails
No matter how much we improve the preload process, fetching the sitemap is always going to be a bottle neck. There isn't much we can do if the cURL request takes longer other than asking users to contact their host, upgrade hosting and so on. 

An idea is to come up with a fallback plan. 

- When sitemap fetch fails, get the links from the DB.
- Do something similar to what the SEO plugins do to generate the sitemap in the first place. 
- We know all the custom post types and taxonomies. 
- Show an admin notice that we are switching to fallback. 

I think this would reduce a ton of issues we have related to sitemap based preload. The customer is happy because we are auto-magically switching strategies and his/her pages are preloaded one way or another.

arunbasillal:
An implementation for #1306 for your review. 

Uses `get_posts` to get the latest 1000 posts and then passes the permalinks back. 

It's limited to 1000 by default which is the default entries per sitemap in Yoast. Can be filtered with `rocket_preload_sitemap_fallback_request_args`. 

The fallback will almost never be called more than once:

- If sitemap is 404, then there are no sub-sitemaps to fetch.
- If `$xml_data` is empty or `$xml` is false, then its likely so for all sub-sitemaps.

Risk analysis: 

I think this is low risk since its used only when sitemap preload fails.


arunbasillal:
Screenshot

fallback

inc/classes/preload/class-sitemap.php
```diff
@@ -122,42 +135,80 @@ public function process_sitemap( $sitemap_url, $urls = [] ) {
...
+				
+				return $urls;
+			}
+
 			set_transient( 'rocket_preload_errors', $errors );
 			return [];
 		}
 
 		if ( ! function_exists( 'simplexml_load_string' ) ) {
+
```
arunbasillal: We probably need an error message when `simplexml_load_string` is missing.

inc/classes/preload/class-sitemap.php
```diff
@@ -108,6 +108,19 @@ public function process_sitemap( $sitemap_url, $urls = [] ) {
...
 			$errors['errors'][] = sprintf( __( 'Sitemap preload encountered an error. Could not gather links on %1$s because of the following error: %2$s. %3$sLearn more%4$s.', 'rocket' ), $sitemap_url, $sitemap->get_error_message(), '<a href="https://docs.wp-rocket.me/article/1065-sitemap-preload-is-slow-or-some-pages-are-not-preloaded-at-all#failed-preload" rel="noopener noreferrer" target=_"blank">', '</a>' );
+			
+			// Attempt to use the fallback method.
+			$fallback_urls = $this->get_urls();
+			
+			if ( ! empty( $fallback_urls ) ) {
+				
+				$urls = array_merge( $urls, $fallback_urls );
+				
+				$errors['errors'][] = __( 'WP Rocket will use a fallback method to preload your website.', 'rocket' );
```
Tabrisrp: I think we determined we didn't want to communicate the use of the fallback to the users

inc/classes/preload/class-sitemap.php
```diff
@@ -195,4 +259,56 @@ public function process_sitemap( $sitemap_url, $urls = [] ) {
...
+		 *
+		 * @since 3.2.4
+		 * @author Arun Basil Lal
+		 *
+		 * @param array $args Arguments for get_posts
+		 */
+		$args = apply_filters(
+			'rocket_preload_sitemap_fallback_request_args',
+			[
+				'fields'			=> 'ids',
```
Tabrisrp: alignment issue

inc/classes/preload/class-sitemap.php
```diff
@@ -195,4 +259,56 @@ public function process_sitemap( $sitemap_url, $urls = [] ) {
...
+				'numberposts' 		=> 1000,
+				'posts_per_page'  	=> -1,
+				'post_type'			=> $post_types,
```
Tabrisrp: aligment issue

inc/classes/preload/class-sitemap.php
```diff
@@ -195,4 +259,56 @@ public function process_sitemap( $sitemap_url, $urls = [] ) {
...
+		$urls = [];
+		
+		// Get public post types
+		$post_types = get_post_types( array( 'public' => true ) );
+		$post_types = array_filter( $post_types, 'is_post_type_viewable' );
+		
+		/**
+		 * Filters the arguments for get_posts
+		 *
+		 * @since 3.2.4
```
Tabrisrp: 3.3

inc/classes/preload/class-sitemap.php
```diff
@@ -195,4 +259,56 @@ public function process_sitemap( $sitemap_url, $urls = [] ) {
 		$urls = array_merge( $urls, $tmp_urls );
 		return $urls;
 	}
+	
+	/**
+	 * Get URLs from WordPress
+	 * 
+	 * Used as a fallback when extracting URLs from sitemap fails.
+	 *
+	 * @since 3.2.5
```
Tabrisrp: 3.3

Tabrisrp:
Also, existing error messages should only be displayed if `WP_ROCKET_DEBUG` is `true`

inc/classes/preload/class-sitemap.php
arunbasillal: Removed it.

inc/classes/preload/class-sitemap.php
arunbasillal: Updated to 3.3

inc/classes/preload/class-sitemap.php
arunbasillal: My editor had tabs as 4 spaces. GitHub uses 8 spaces as default. Fixed now.

inc/classes/preload/class-sitemap.php
arunbasillal: Alignment looks correct here. But not here 🤔 - https://github.com/wp-media/wp-rocket/blob/feature/1306-sitemap-preload-fallback/inc/classes/preload/class-sitemap.php

arunbasillal:
@Tabrisrp Since we return when each of these checks fails, and since `get_urls` always gets the same urls, would it be more elegant to call the fallback here - https://github.com/wp-media/wp-rocket/blob/b922ddb737dfd1bde595f579ac7a6204d5ce8287/inc/classes/preload/class-sitemap.php#L53-L55

Tabrisrp:
This would mean the fallback is triggered only if there was no successful read of the sitemap(s). If at least one worked, the fallback would not be triggered.

arunbasillal:
Added `$sitemap_error` to track failures and call `get_urls` once instead.

PR #1489, opened 2019-01-23
Presslabs hosting integration

cristianuibar:
This PR will add support for our hosting environment.

A little explanation:
- Since we disable any advanced-cache.php file inside wp-content folder we need to require it manually;
- We need to clear our entire cache when a domain purge is happening inside wp-rocket;
- We also found out that some inconsistency can show up and hence un-styled pages (from our cache) when the minified files are cleared from wp-rocket but not from our cache as well. So we believe we also need to hook on to  `after_rocket_clean_minify`. Please let me know if this should not be the case and I should hook somewhere else?
- For post actions, we clear only the post, homepage and listings from the cache.

Looking forward to your opinion on how I can improve this. We want to see this merged as soon as possible.


Tabrisrp:
Hi,

Thank you for the PR. Did you test it on your hosting environment?

`advanced-cache.php` contains the call to `process.php`, which is the file determining if the page should be cached AND optimized. If it's not run when it is expected in WP loading process, the optimizations (minify/combine,etc) will not be applied I believe.

The minified files are never cleared from WP Rocket without a clear domain at the same time, so if you're already hooked on the clear domain action (which seems to be the case), it should not be necessary to also hook on the minify action.

Since you have your own caching system, I recommend to add `add_filter( 'do_rocket_generate_caching_files', '__return_false', PHP_INT_MAX );` to prevent WP Rocket from saving static cache files, and also `add_filter( 'rocket_cache_mandatory_cookies', '__return_empty_array', PHP_INT_MAX );` to prevent some incompatibility issues.

cristianuibar:
Thank you @Tabrisrp!

I have just implemented your suggestions, tested them on one of our customers, and it seems to work as expected. I believe it works a lot better than before because of the static files and cookies cache disable.

I am still thinking about cache flush on post publish/update/delete. We already have this mechanic implemented inside our custom code on top of WP using the `cache_post_change` hook with priority 0. Since we already clean our cache on this hook, and WP Rocket does the same, does it still make sense to hook one another using `rocket_clean_post` and `pl_pre_cache_refresh`?

Tabrisrp:
This might not be necessary depending on what you already do on your side. I suggest to have a look at what we already have for Kinsta, since this code is the most integrated and takes into account the different parts of WP Rocket where you can trigger a clean cache https://github.com/wp-media/wp-rocket/blob/master/inc/3rd-party/hosting/kinsta.php

Tabrisrp:
In the next version of the plugin the advanced-cache.php file is not required anymore for things not cache related to work, so you can remove the inclusion in this, that will be even better.

inc/3rd-party/hosting/presslabs.php
```diff
@@ -0,0 +1,142 @@
+<?php
+defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
+
+if ( defined( 'PL_INSTANCE_REF' ) && class_exists( '\Presslabs\Cache\CacheHandler' ) ) {
+	if ( file_exists( WP_CONTENT_DIR . '/advanced-cache.php' ) ) {
```
Tabrisrp: You can remove this starting from our 3.3 version

cristianuibar: Removed the advanced-cache loading as instructed.

pdobrescu:
Hi @Tabrisrp, we've removed that call, please check if everything is ok now. 

Many thanks for your support with this.

inc/3rd-party/hosting/presslabs.php
```diff
@@ -0,0 +1,139 @@
+<?php
+defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
+
+if ( defined( 'PL_INSTANCE_REF' ) && class_exists( '\Presslabs\Cache\CacheHandler' ) ) {
+
+	add_action( 'pl_pre_cache_refresh', 'rocket_clean_post', 0 );
+	add_action( 'after_rocket_clean_domain', 'rocket_pl_clear_cache' );
+	add_action( 'after_rocket_clean_post', 'rocket_pl_clean_post', 2 );
```
Tabrisrp: You don't need this one, as `rocket_clean_post()` is calling `rocket_clean_files()`, so your action on line 13 will take care of cache clearing.

If  you keep it, there will be a double call to clearing the cache, which is unnecessary. You can remove the action and the associated function.

inc/3rd-party/hosting/presslabs.php
```diff
@@ -0,0 +1,139 @@
+<?php
+defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
+
+if ( defined( 'PL_INSTANCE_REF' ) && class_exists( '\Presslabs\Cache\CacheHandler' ) ) {
+
+	add_action( 'pl_pre_cache_refresh', 'rocket_clean_post', 0 );
```
Tabrisrp: You should be using `rocket_clean_files` instead

inc/3rd-party/hosting/presslabs.php
```diff
@@ -0,0 +1,139 @@
...
+	add_action( 'after_rocket_clean_domain', 'rocket_pl_clear_cache' );
+	add_action( 'after_rocket_clean_post', 'rocket_pl_clean_post', 2 );
+	add_filter( 'rocket_display_varnish_options_tab', '__return_false' );
+	add_filter( 'do_rocket_generate_caching_files', '__return_false', PHP_INT_MAX );
+	add_filter( 'rocket_cache_mandatory_cookies', '__return_empty_array', PHP_INT_MAX );
+	add_action( 'after_rocket_clean_home', 'rocket_pl_clean_home', 10, 2 );
+	add_action( 'after_rocket_clean_file', 'rocket_pl_clean_url' );
+	add_action( 'pl_pre_url_button_cache_refresh', 'after_rocket_clean_file' );
```
Tabrisrp: `after_rocket_clean_file` is not a callable. It should be `rocket_clean_files()`

pdobrescu:
@Tabrisrp many thanks for the additional code recommendations, we'll get them implemented soon. 

I have some questions as we're seeing a rather strange situation in our testing:
- how often and when is `rocket_clean_domain` supposed to be called? Should all caches be purged here? We've seen sometimes several global cache purges being done and that's for sure not optimal for performance, considering that all caches are purged in this way
- when is `rocket_clean_post()` being called and how is the list of URL's from `rocket_clean_files()` being constructed? Which URL's should be in this list?

We're just trying to ensure that this integration won't do more harm than good, by purging the caches too often (this is how it looks on our side for the moment). 

Thank you!

pdobrescu:
Hello @Tabrisrp, can you please help me with the answers for the questions from the previous comment? I just want to ensure we're doing this implementation the right way. 

Thank you!

Tabrisrp:
@pdobrescu 

`rocket_clean_domain` is called on a number of actions listed here: https://github.com/wp-media/wp-rocket/blob/master/inc/common/purge.php#L4

It will also be called when doing a manual purge cache from WP Rocket admin bar, when WP Rocket settings are updated, and in various cases for compatibility with other plugins.

For `rocket_clean_post`, it's on the actions here: https://github.com/wp-media/wp-rocket/blob/master/inc/common/purge.php#L238.

It will be manually called when purging the cache for a specific post, and some compatibility with plugins. The list of URLs is constructed inside the function, it includes the post itself, next and previous posts, related terms and archives, and the homepage.

inc/3rd-party/hosting/presslabs.php
pdobrescu: OK, I have removed it completely and modified the one that used to be on line 13.

inc/3rd-party/hosting/presslabs.php
pdobrescu: 👍 change done!

inc/3rd-party/hosting/presslabs.php
pdobrescu: Ok, updated!

pdobrescu:
Hello @Tabrisrp, I've pushed some more changes following your code review, with the following general comments:
- I have removed completely the call to `rocket_clean_domain`, as we consider it's not needed (we have this operation in our dashboard too
- I have added back the requirement of the `advanced-cache.php` file as otherwise the WP Rocket throws an error in wp-admin

Please let me know if there is anything else needed from our side for this PR to be merged.

Thank you!

pdobrescu:
Hi @Tabrisrp, can you please check the latest changes and hopefully we can get this PR merged? We already have common customers, who are successfully using the changes from this PR. Please let us know if there is anything else that we're missing.

Tabrisrp:
Looks good to me now, I'll merge it on our develop branch and it will be included in the next release 👍

pdobrescu:
Great, thanks a lot @Tabrisrp!

PR #1528, opened 2019-02-11
Update varnish.php

poralix:
1. Wrong scheme, the plugin calls http when is should be https. We have nginx in front of Varnish and both http and https are proxied to Varnish.

2. We have multi-host environment we need to specify valid domain with PURGE request. So we remove * from a domain name.


Tabrisrp:
I'm going to close your PR as there is important changes made in #2021 for our Varnish integration, that also includes those.

Issue #1529, opened 2019-02-11
Installing via Composer asks for requirement that doesn't exist


mike-source:
When installing via Composer (as per docs), it fails asking for a requirement version which doesn't appear to exist:

```
$ composer require wp-media/wp-rocket
Using version ^3.2 for wp-media/wp-rocket                
./composer.json has been updated
Loading composer repositories with package information
Updating dependencies (including require-dev)            
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - wp-media/wp-rocket v3.2.4 requires a5hleyrich/wp-background-processing 1.3 -> no matching package found.
    - wp-media/wp-rocket v3.2.3.1 requires a5hleyrich/wp-background-processing 1.3 -> no matching package found.
    - wp-media/wp-rocket v3.2.3 requires a5hleyrich/wp-background-processing 1.3 -> no matching package found.
    - wp-media/wp-rocket v3.2.2 requires a5hleyrich/wp-background-processing 1.3 -> no matching package found.
    - wp-media/wp-rocket v3.2.1.1 requires a5hleyrich/wp-background-processing 1.3 -> no matching package found.
    - wp-media/wp-rocket v3.2.1 requires a5hleyrich/wp-background-processing 1.3 -> no matching package found.
    - wp-media/wp-rocket v3.2.0.1 requires a5hleyrich/wp-background-processing 1.3 -> no matching package found.
    - wp-media/wp-rocket v3.2 requires a5hleyrich/wp-background-processing 1.3 -> no matching package found.
    - Installation request for wp-media/wp-rocket ^3.2 -> satisfiable by wp-media/wp-rocket[v3.2, v3.2.0.1, v3.2.1, v3.2.1.1, v3.2.2, v3.2.3, v3.2.3.1, v3.2.4].

Potential causes:
 - A typo in the package name
 - The package is not available in a stable-enough version according to your minimum-stability setting
   see <https://getcomposer.org/doc/04-schema.md#minimum-stability> for more details.

Read <https://getcomposer.org/doc/articles/troubleshooting.md> for further common problems.

Installation failed, reverting ./composer.json to its original content.
```

**To Reproduce**
Just run `composer require wp-media/wp-rocket` in your project.

**Expected behavior**
It should have a resolvable set of requirements.

**Additional context**
It asks for version 1.3 of https://packagist.org/packages/a5hleyrich/wp-background-processing but the latest version that exists is 1.0.1

mike-source:
[...]

PR #1530, opened 2019-02-11
Fix: https://github.com/wp-media/wp-rocket/issues/1529

mike-source:
Update Composer.json to require version "^1.0", of "a5hleyrich/wp-background-processing" which is the current latest in wp-packagist (instead of 1.3, which does not exist and causes composer install to fail on it's dependencies).

Note: This may need to be "1.0.1 ", I'm not sure what else it will affect.

Also Note: I'm not sure which Topic branch is appropriate here besides master, this should surely be a really simple fix? Unless I'm barking up completely the wrong tree!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
 	],
     "require": {
 		"php": ">=5.4.0",
-		"a5hleyrich/wp-background-processing": "1.3",
+		"a5hleyrich/wp-background-processing": "^1.0",
 		"composer/installers": "~1.0",
 		"jamesryanbell/cloudflare": "^1.11",
 		"matthiasmullie/minify": "1.3.*",


bweiler-marketingflotte:
Same problem here.

Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Installation request for wp-media/wp-rocket 3.2.5 -> satisfiable by wp-media/wp-rocket[v3.2.5].
    - wp-media/wp-rocket v3.2.5 requires a5hleyrich/wp-background-processing 1.3 -> no matching package found.

Tabrisrp:
Hi!

As we use a modified version of the WP Background Processing package, we can't choose the version currently existing on packagist, and that's why we're redirecting to the fork instead in our composer configuration.

In the future we plan to move away from this library, or to create our own version, so this will simplify this matter.

PR #1557, opened 2019-02-28
Add Dependency Injection Container

Tabrisrp:
* Add league/container as the DI Container
* Refactor some classes to use the event manager/subscriber system


inc/classes/optimization/CSS/class-critical-css.php
```diff
@@ -0,0 +1,330 @@
...
+		}
+
+		try {
+			$files = new RecursiveIteratorIterator( $directory, RecursiveIteratorIterator::CHILD_FIRST );
+		} catch ( Exception $e ) {
+			// no logging yet.
+			return;
+		}
+
+		foreach ( $files as $file ) {
```
Screenfeed:
```php
if ( ! $files ) {
	return;
}
```

PR #1572, opened 2019-03-13
Pressable compatibility

Tabrisrp:
* Prevent WP Rocket cache files creation
* Prevent display of notices related to `advanced-cache.php`
* Prevent display of Varnish and NGINX add-ons as they don't make sense on Pressable
* Auto-purge Pressable cache when WP Rocket purge cache action is triggered
* Auto-detect Pressable CDN value
* Modify absolute path for assets in `wp-includes` to fit with Pressable setup

### DIFF
--- a/inc/classes/subscriber/third-party/Hostings/class-pressable-subscriber.php
+++ b/inc/classes/subscriber/third-party/Hostings/class-pressable-subscriber.php
@@ -1,0 +1,117 @@
+<?php
+namespace WP_Rocket\Subscriber\Third_Party\Hostings;
+
+use WP_Rocket\Event_Management\Subscriber_Interface;
+
+/**
+ * Subscriber for compatibility with Pressable hosting
+ *
+ * @since 3.3
+ * @author Remy Perona
+ */
+class Pressable_Subscriber implements Subscriber_Interface {
+	/**
+	 * @inheritDoc
+	 */
+	public static function get_subscribed_events() {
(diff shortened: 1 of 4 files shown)


inc/classes/subscriber/third-party/Hostings/class-pressable-subscriber.php
```diff
@@ -0,0 +1,117 @@
...
+	 * Modify wp-includes absolute path to be able to optimize assets in this directory on Pressable
+	 *
+	 * @since 3.3
+	 * @author Remy Perona
+	 *
+	 * @param string $file Absolute path to the file.
+	 * @return string
+	 */
+	public function fix_wp_includes_path( $file ) {
+		if ( ! preg_match( '#^(.+)(wp-includes(?:.+))$#is', $file ) ) {
```
Screenfeed: This is useless, you can directly return `preg_replace()`.

inc/functions/formatting.php
```diff
@@ -337,6 +337,15 @@ function rocket_url_to_path( $url, $hosts = '' ) {
...
 	$file     = rocket_realpath( $file );
+	/**
+	 * Filters the absolute path to the asset file
+	 *
+	 * @since 3.3
+	 * @author Remy Perona
+	 *
+	 * @param string $file Absolute path to the file.
+	 */
+	$file = apply_filters( 'rocket_url_to_path', $file );
```
Screenfeed: I think it could be useful to add `$url` as 2nd parameter.

PR #1611, opened 2019-04-03
Fixes #1558 Prevent rocket_clean_post from running multiple times for the same post

### LINKED ISSUE #1558: Prevent WP Rocket cache from being cleared multiple times
When a post is being saved using Elementor or Divi Builder (haven't checked other page builders), the 'clean_post_cache' hook is triggered 3 times.

WP Rocket is running the partial cache purging with the array full of posts' related URLs 3 times while saving a single post in this situation. 

When external software is being involved in the cache clearing process, there is a possibility that saving post will take much time.

Example:
Varnish cache clearing on Flywheel takes around 0.5-0.7 seconds for each request, which makes saving a post with the above page builders to take approximately 25 seconds (it depends on the number of related posts in the array)

I'm not sure if 3 'clean_post_cache' calls are necessary, on our end we can make sure that WP Rocket's cache is being cleared only after first one:
https://wp-media.slack.com/archives/C43T1AYMQ/p1551450912003800?thread_ts=1551362494.005200&cid=C43T1AYMQ

Related tickets:
https://secure.helpscout.net/conversation/764336044/94386?folderId=2415573
https://secure.helpscout.net/conversation/786974195/98071?folderId=2415573

Tabrisrp:
Also change function signature to add the second parameter `$post` while staying backward compatible

### DIFF
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -60,6 +60,8 @@ function rocket_clean_post( $post_id, $post = null ) {
 		return;
 	}
 
+	$done[ $post_id ] = 1;
+
 	if ( defined( 'DOING_AUTOSAVE' ) ) {
 		return;
 	}
@@ -67,7 +69,7 @@ function rocket_clean_post( $post_id, $post = null ) {
 	$purge_urls = [];
 
 	// Get all post infos if the $post object was not supplied.
-	if ( empty( $post ) ) {
+	if ( is_null( $post ) ) {
 		$post = get_post( $post_id );
 	}
 
@@ -232,8 +234,6 @@ function rocket_clean_post( $post_id, $post = null ) {
 	 * @param string  $lang       The post language
 	 */
 	do_action( 'after_rocket_clean_post', $post, $purge_urls, $lang );
-
-	$done[ $post_id ] = 1;
 }
 add_action( 'wp_trash_post',           'rocket_clean_post' );
 add_action( 'delete_post',             'rocket_clean_post' );
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -50,17 +50,26 @@ add_filter( 'widget_update_callback', 'rocket_widget_update_callback' );
  * @since 1.1.3 Use clean_post_cache instead of transition_post_status, transition_comment_status and preprocess_comment
  * @since 1.0
  *
- * @param int $post_id The post ID.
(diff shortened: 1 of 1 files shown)


inc/common/purge.php
```diff
@@ -218,16 +227,18 @@ function rocket_clean_post( $post_id ) {
...
-	 * @param obj    $post       The post object
-	 * @param array  $purge_urls URLs cache files to remove
-	 * @param string $lang       The post language
+	 * @param WP_Post $post       The post object
+	 * @param array   $purge_urls URLs cache files to remove
+	 * @param string  $lang       The post language
 	 */
 	do_action( 'after_rocket_clean_post', $post, $purge_urls, $lang );
+
+	$done[ $post_id ] = 1;
```
Screenfeed: This should go right after the `isset()` test imho, otherwise this is will run multiple times if we don't reach this point.

inc/common/purge.php
```diff
@@ -218,16 +227,18 @@ function rocket_clean_post( $post_id ) {
...
 }
-add_action( 'wp_trash_post'             , 'rocket_clean_post' );
-add_action( 'delete_post'               , 'rocket_clean_post' );
-add_action( 'clean_post_cache'          , 'rocket_clean_post' );
-add_action( 'wp_update_comment_count'   , 'rocket_clean_post' );
+add_action( 'wp_trash_post',           'rocket_clean_post' );
+add_action( 'delete_post',             'rocket_clean_post' );
+add_action( 'clean_post_cache',        'rocket_clean_post' );
```
Screenfeed:
```suggestion
add_action( 'clean_post_cache',        'rocket_clean_post', 10, 2 );
```

inc/common/purge.php
```diff
@@ -50,17 +50,26 @@ function rocket_widget_update_callback( $instance ) {
  * @since 1.1.3 Use clean_post_cache instead of transition_post_status, transition_comment_status and preprocess_comment
  * @since 1.0
  *
- * @param int $post_id The post ID.
+ * @param int     $post_id The post ID.
+ * @param WP_Post $post    WP_Post object.
  */
-function rocket_clean_post( $post_id ) {
+function rocket_clean_post( $post_id, $post = '' ) {
```
Screenfeed:
```suggestion
function rocket_clean_post( $post_id, $post = null ) {
```

Issue #1194, opened 2018-09-19
CSS minification: Exclude CSS files from Uncode theme


ernee:
Customers report that the cache dir size increases with Uncode theme. Excluding this CSS file path has resolved several cases: 

/wp-content/themes/uncode/library/css/(.*).css

Perhaps auto-exclude as well?

tickets:

https://secure.helpscout.net/conversation/662328344/79959?folderId=1391600

https://secure.helpscout.net/conversation/640260108/76736/

GeekPress:
These files also needs to be excluded from the cache busted option. 

The best solution will be to auto-replace the dynamic query string by the theme version. Or to contact the theme developer to ask them to put the theme version instead of a ramdom query string

Tabrisrp:
@wp-media/supportrocket Is this a common issue?

GeekPress:
I was in contact with Uncode team. The issue is because of the Production Mode option with Off value.

To fix the issue, we can ask to our users to turn on the Production mode.
https://undsgn.d.pr/eLOsFj

On our side we could override or force the value of the Production Mode if our minification option is turned on.

It's also related to this issue:
https://github.com/wp-media/wp-rocket/issues/1628

GeekPress:
This is fixed since Uncode 2.10.

PR #1689, opened 2019-05-06
Fixes #1628 Exclude Uncode files from Remove Query Strings

### LINKED ISSUE #1628: Auto-exclude Uncode JS files from Remove query strings option
These 3 files from Uncode theme generate random queries which leads to a cache dir size issue when the Remove query strings from static resources option is enable:

- /wp-content/themes/uncode/library/js/init.js
- /wp-content/themes/uncode/library/js/app.js
- /wp-content/themes/uncode/library/js/plugins.js

Tabrisrp:


GeekPress:
I'm in contact with the author of Uncode themes. 

We have the issue because their themes add a random query string on their CSS & JS files. It's because of their "Production Mode" option:
https://undsgn.d.pr/AhAFb6

I asked them to automatically turn on this option when WP Rocket is active.

On our side could we "bypass" the value of their option and then being able to optimize the Uncode files?

It also concern this issue:
https://github.com/wp-media/wp-rocket/issues/1194

We will be able to fix these 2 issues.

PR #1696, opened 2019-05-07
Fixes #1514 Use preg_replace to insert the combined CSS tag

### LINKED ISSUE #1514: Combined CSS added multiple times when <title>is included more than once
Normally, one pair of `<title></title>` tags should exist in the `head` block of a page:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/title

If, for any reason, there are more, WP Rocket adds the combined CSS file multiple times because of how this is implemented here:
https://github.com/wp-media/wp-rocket/blob/c92efa3a52ef285eb6230a04ac4c5e6d4e2f8e26/inc/classes/optimization/CSS/class-combine.php#L115

Maybe it would be best to use a RegEx to find the `<title>` tag.
This can also have a check, about the `<title>`  tag not being there, that will print a warning in the Dashboard, making it easier to troubleshoot CSS combination related issues.

**Related Ticket:**
https://secure.helpscout.net/conversation/768028020/94982?folderId=2135277

### LINKED ISSUE #1547: Merged CSS File Output Multiple Times When Using <title> For SVG's
This issue closely relates to #1514 but I have noticed it on a site where I was making use of https://wordpress.org/plugins/simple-social-icons/

The issue though is exasperated and a bit more annoying because SVG's for accessibility concerns are supposed to use the <title> and (also desc) tags for screenreaders. 

See > https://www.w3.org/TR/SVG11/struct.html#DescriptionAndTitleElements

Now there is a bit of an important bit that might help when deciding to redo the Regex. The <title> attributes for SVG's should include a unique id related to it. Typical title tags don't contain related ID's. 

The only reason I am opening a separate issue as the bug is a bit more annoying when working with SVG's on websites that are also being built for accessibility. 

Solutions: Either check for specific use cases of these title tags and exclude OR output the combined css file after the meta charset tag

Tabrisrp:
Also fixes #1547

### DIFF
--- a/inc/classes/admin/class-options.php
+++ b/inc/classes/admin/class-options.php
@@ -1,8 +1,6 @@
 <?php
 namespace WP_Rocket\Admin;
 
-defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
-
 /**
  * Manages options using the WordPress options API.
  *
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ notifications:
 branches:
   only:
   - master
+  - develop
+  - /branch-.*$/
 cache:
   directories:
   - vendor
--- a/composer.json
+++ b/composer.json
@@ -43,9 +43,13 @@
 	"require-dev": {
 		"brain/monkey": "^2.0",
 		"dealerdirect/phpcodesniffer-composer-installer": "^0.5.0",
-		"phpcompatibility/phpcompatibility-wp": "*",
+		"doctrine/instantiator": "1.0.*",
+		"myclabs/deep-copy": "1.6.*",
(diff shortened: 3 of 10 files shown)


inc/classes/admin/class-options.php
```diff
@@ -1,8 +1,6 @@
 <?php
 namespace WP_Rocket\Admin;
 
-defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
```
Screenfeed: This should not be removed.
A direct access to the file will print a fatal error leading to a path disclosure:
```
Fatal error: Class 'WP_Rocket\Admin\Abstract_Options' not found in /abs/path/to/inc/classes/admin/class-options.php on line 10
```

PR #1733, opened 2019-05-17
Fixes #1196 Exclude common external jQuery from defer JS safe mode

### LINKED ISSUE #1196: Jquery Safe Mode - exclude common externally hosted jquery
Safe mode for Deferred JS doesn't work if the site is using an externally hosted version of jquery instead of the default WordPress file. 

Could we also automatically exclude the most common versions of externally hosted jquery so that we don't have to use a helper plugin? 

e.g. https://ajax.googleapis.com/ajax/libs/jquery/1.12.4/jquery.min.js?ver=1.12.4

Tabrisrp:
Exclude jQuery loaded from `wp.com` CDN and from `ajax.googleapis.com`

### DIFF
--- a/inc/functions/options.php
+++ b/inc/functions/options.php
@@ -468,9 +468,13 @@ function get_rocket_exclude_defer_js() {
 	];
 
 	if ( get_rocket_option( 'defer_all_js', 0 ) && get_rocket_option( 'defer_all_js_safe', 0 ) ) {
-		$jquery = site_url( $wp_scripts->registered['jquery-core']->src );
+		$jquery            = site_url( $wp_scripts->registered['jquery-core']->src );
+		$jetpack_jquery    = 'c0.wp.com/c/(?:.+)/wp-includes/js/jquery/jquery.js';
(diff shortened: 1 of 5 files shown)


inc/functions/options.php
```diff
@@ -468,9 +468,13 @@ function get_rocket_exclude_defer_js() {
 	];
 
 	if ( get_rocket_option( 'defer_all_js', 0 ) && get_rocket_option( 'defer_all_js_safe', 0 ) ) {
-		$jquery = site_url( $wp_scripts->registered['jquery-core']->src );
+		$jquery            = site_url( $wp_scripts->registered['jquery-core']->src );
```
Screenfeed: I remember some people messing with `jquery-core` and `jquery-migrate`, and registering their custom jQuery under `jquery`. Do you think it's still a thing today?

Tabrisrp: not often enough to warrant that I think, and often they replaced with the googleapis one so it should be covered by that

Screenfeed: OK. Let's say that the risk of `$wp_scripts->registered['jquery-core']->src` not being set or being empty can be neglected then.

PR #1734, opened 2019-05-17
Fixes #1685 Auto-active SSL cache for new installations

### LINKED ISSUE #1193: More checks for https?
Sometimes a site isn't cached because it's using HTTPS (for example with a plugin) but https has not been set in the WP settings. So we don't detect it and SSL cache isn't applied. 

It creates some support tickets, and can be overlooked in debugging on our side.

Is it possible to do  more checks and show an admin notice if we suspect https might be used? 
- Site may be forcing HTTPS using a plugin but have not updated their URL settings. We could check for some specific plugins like Really Simple SSL for example.
- Check if the admin URL is using HTTPS?
- Possible to check the front end url?

Admin notice would be something along the lines of: 
"Are you using https on your site? If so, go to WP Rocket settings > Cache and enable SSL Cache"

Even better, if possible, put a check box in the notice to enable it from there.

### LINKED ISSUE #1685: Proposal : Remove SSL Cache option and auto-apply it.
Here the story of the SSL Cache option:

At the beginning, WP Rocket didn't have an option to cache SSL pages. These pages were automatically excluded from the cache.

At this period SSL was just used on checkout pages for e-commerce website. We didn't auto-exclude these pages automatically yet like we can do.

We added an option for SSL Cache to avoid these issues. People who used SSL on few pages could decide to apply the cache or not.

A couple of time later, we decide to auto-exclude cart & checkout pages. 

Then Google recommended to apply SSL on all websites. Some times later, the SSL options make more sense to be auto-activated if we detected SSL on the general settings URL.

Now, people doesn't have anymore mixed pages. I mean some in non-ssl and other in ssl. You have SSL or not on all your pages.

So, now we could have no difference between SSL and non-SSL website. The option is useless and caching should be applied whatever we have SSL or not.

It will help us to avoid this kind of ticket when people as non-ssl in their general setting but using something else for applying SSL on their website:
https://secure.helpscout.net/conversation/843383300/106384/

Tabrisrp:
Also fixes #1193

### DIFF
--- a/inc/admin/options.php
+++ b/inc/admin/options.php
@@ -290,7 +290,7 @@ add_filter( 'pre_update_option_' . WP_ROCKET_SLUG, 'rocket_pre_main_option', 10,
  * @param array $value     An array of submitted options values.
  */
 function rocket_update_ssl_option_after_save_home_url( $old_value, $value ) {
-	if ( $old_value === $value ) {
+	if ( $old_value === $value || get_rocket_option( 'cache_ssl' ) ) {
 		return;
 	}
 
--- a/inc/admin/upgrader.php
+++ b/inc/admin/upgrader.php
@@ -165,7 +165,7 @@ function rocket_first_install() {
 				'cache_mobile'                => 1,
 				'do_caching_mobile_files'     => 0,
 				'cache_logged_user'           => 0,
-				'cache_ssl'                   => rocket_is_ssl_website() ? 1 : 0,
+				'cache_ssl'                   => 1,
 				'emoji'                       => 1,
 				'embeds'                      => 0,
 				'cache_reject_uri'            => [],
--- a/inc/classes/admin/settings/class-page.php
+++ b/inc/classes/admin/settings/class-page.php
@@ -1945,6 +1945,7 @@ class Page {
 				'version',
 				'cloudflare_old_settings',
 				'sitemap_preload_url_crawl',
+				'cache_ssl',
 			]
 		);
 	}
--- a/inc/classes/admin/settings/class-page.php
+++ b/inc/classes/admin/settings/class-page.php
@@ -473,20 +473,6 @@ class Page {
 					],
 					'page'        => 'cache',
 				],
-				'cache_ssl_section'    => [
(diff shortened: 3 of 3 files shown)


arunbasillal:
Looks good to me. 👍 

Wondering if this will have any influence on a website where WP Rocket fails to detect HTTPS - https://github.com/wp-media/wp-rocket/blob/ba3581b37a620fa069e5a92458ef0e2641b5a22c/inc/admin/options.php#L265-L268

Tabrisrp:
It will behave as before for that so I don't think we need to worry about it

arunbasillal:
Thanks. So shall I merge this PR? Not sure if it's you or me.

Screenfeed:
Hi.
I don't understand this one:
- If I read the title, "Auto-active SSL cache for new installations", this PR should stop at `upgrader.php` and not remove the checkbox.
- If I read the issue ticket, the checkbox should indeed be removed, but the option itself and the whole system too (htaccess writes conditions, etc).
- If somehow we keep it, hidden, filterable: the option value should be removed, and a hook should filter a hard-coded value. Moreover, if I disabled it on my site, then upgrade, I can't enable it back anymore.
I feel like we're in the 3rd case, but we don't want to force this change to users who disabled the option, am I right?

Screenfeed:
After some explanations from @arunbasillal I understand more the story.
We still found few quirks though. The option would be in a "zombie state" as I call it: it's dead, but still alive somehow, with ways to change its value.
- Since the checkbox is missing, submitting the settings page will delete the option. If I remember correctly, there is something for "hidden options" somewhere.
- The function `rocket_update_ssl_option_after_save_home_url()` may set the option back to 0. A possible solution:
```php
function rocket_update_ssl_option_after_save_home_url( $old_value, $value ) {
        if ( $old_value === $value || get_rocket_option( 'cache_ssl' ) ) {
            return;
        }
```

The general idea is to prevent the option to be set from 1 to 0, unless the dedicated filter is used.
Looking in the plugin for `rocket_is_ssl_website()` tests may spot other cases.

PR #1751, opened 2019-06-03
Fixes #1333 Prevent multiple CPCSS generations

### LINKED ISSUE #1333: CPCSS generation on infinite Loop - Force exit when this happens
There were couple of tickets where CPCSS generation never stopped. 

screen shot 2018-11-08 at 1 09 57 pm

For some weird reason, the `complete()` method was never called - https://github.com/wp-media/wp-rocket/blob/f8def46c4563f555d3c6627362bd78b6260150df/inc/classes/class-rocket-background-critical-css-generation.php#L181-L201

**Tickets:**
- https://secure.helpscout.net/conversation/680055496/82258?folderId=273766#thread-1887162396
- https://secure.helpscout.net/conversation/693314047/83840?folderId=1724284

**Idea:** 
We could add a check to see if the count ever goes beyond the total count and when it does force the process to exit. It sounds hacky, but when weird things happen, a hack is not such a bad idea imo.

Tabrisrp:
Also prevent regenerating the CPCSS files on option activation when they already exist

### DIFF
--- a/inc/classes/optimization/CSS/class-critical-css.php
+++ b/inc/classes/optimization/CSS/class-critical-css.php
@@ -33,9 +33,9 @@ class Critical_CSS {
 	 *
 	 * @since 2.11
 	 * @var string path to the critical css directory
-	 * @access public
+	 * @access private
 	 */
-	public $critical_css_path;
+	private $critical_css_path;
 
 	/**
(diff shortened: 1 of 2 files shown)


Screenfeed:
Travis is not happy.

PR #1752, opened 2019-06-03
Fixes #1471 Purge Varnish cache if Cloudflare is active

### LINKED ISSUE #1471: Auto-enable Varnish Purge with Proxy
When Varnish is used in conjunction with a proxy like Cloudflare or similar, the Varnish cache does not get purged unless this add-on is used: 
https://docs.wp-rocket.me/article/891-purge-varnish-when-using-a-proxy

It seems we could automate this and negate the need for the helper (except maybe for super-edge cases) at least in cases where we detect both our Varnish add-on and Cloudflare add-ons are active.

Tabrisrp:


Screenfeed:
Travis still not happy 🤔

PR #1769, opened 2019-06-06
Update htaccess WP Rocket missing rules notice

### LINKED ISSUE #1555: Check WP Rocket Rules in .htaccess to avoid notice after including our rules manually
When WP Rocket doesn't have the permission rules to write into the .htaccess file, one of our suggestion is to put the rules manually.

After that, we still display notice which can be confused for your customer. (Ex: https://secure.helpscout.net/conversation/788970317/98393?folderId=273761)

To avoid this confusion,  we can also check if the .htaccess file contains "WP ROCKET" to know if our rules are already inserted.

Then, if we detect our rules, we can disable the notice.

### LINKED ISSUE #1651: Proposal: make htaccess permissions warning less ominous
WP_Rocket_‹_WordPress_Themes_—_WordPress

This admin notice leads people to believe that WP Rocket will not function without htaccess permission. That can lead to some panic, and even refund requests:
https://secure.helpscout.net/conversation/827250883/104247?folderId=377611#thread-2314448219

Since the htaccess file is not _required_ for WP Rocket to function, maybe we can tone it down? 
- Change it to yellow instead of red
- Modify the wording to communicate that while the htaccess rules are helpful, they are not required

Tabrisrp:
Closes #1555 
Closes #1651 

- Check if the WP Rocket rules are already present in the htaccess before displaying the notice
- Update the notice wording and switch the type from error to warning

### DIFF
--- a/inc/admin/ui/notices.php
+++ b/inc/admin/ui/notices.php
@@ -385,6 +385,10 @@ function rocket_warning_htaccess_permissions() {
 			return;
 	}
 
+	if ( rocket_check_htaccess_rules() ) {
+		return;
+	}
+
 	$boxes = get_user_meta( get_current_user_id(), 'rocket_boxes', true );
 
 	if ( in_array( __FUNCTION__, (array) $boxes, true ) ) {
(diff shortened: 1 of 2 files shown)


inc/admin/ui/notices.php
```diff
@@ -385,21 +385,42 @@ function rocket_warning_htaccess_permissions() {
...
+		/* translators: This is a doc title! %1$s = opening link; %2$s = closing link */
+		__( 'Troubleshoot: %1$sHow to make system files writeable%2$s', 'rocket' ),
+		/* translators: Documentation exists in EN, DE, FR, ES, IT; use loaclised URL if applicable */
+		'<a href="' . __( 'https://docs.wp-rocket.me/article/626-how-to-make-system-files-htaccess-wp-config-writeable/?utm_source=wp_plugin&utm_medium=wp_rocket', 'rocket' ) . '" target="_blank">',
+		'</a>'
+	);
+
+	add_filter( 'rocket_htaccess_mod_rewrite', '__return_false' );
+
+	$message .= '<p>' . __( 'Don’t worry, WP Rocket’s page caching and settings will still function correctly.', 'rocket' ) . '<br>' . __( 'For optimal performance, adding the following lines into your .htaccess is recommended (not required):', 'rocket' ) . '<br><textarea readonly="readonly" id="rules" name="rules" class="large-text readonly" rows="6">' . esc_textarea( get_rocket_htaccess_marker() ) . '</textarea></p>';
```
Screenfeed: I don't think that using an id as simple as "rules" on something that can be displayed anywhere is a good idea.

PR #1798, opened 2019-06-18
Check if isset before strpos

mirkodevx:
PHP Notice:  Undefined offset: 1
PHP Notice:  Undefined offset: 2

Check if the array keys are setted for $matches_inline.


Tabrisrp:
👋 
Thank you for your contribution. Avoiding the notices is definitely good.

As an alternative instead of repeating an `isset` check multiple times, this could be used instead, after the `preg_match` and before the conditionals:
```
$matches_inline = array_merge(
    [
        'attrs'   => '',
        'content' => '',
    ],
    $matches_inline
);
```
That way the keys are always set with an empty value if for some reason they were not before.

If you can update your PR with this change, I'll merge it for the next release.

hellofromtonya:
@crystinutzaa How about merging into 3.4.4 instead of `develop`?

crystinutzaa:
Yes, I think all 3 issues you sent can be merged in v3.4.4. 
@arunbasillal what do you think?

arunbasillal:
@crystinutzaa Sure :)

hellofromtonya:
Thank you @mirkodevx for contributing to WP Rocket! @crystinutzaa updated to use @Tabrisrp's default strategy. All looks good and this PR has been approved for merge as part of Rocket 3.4.4.

We appreciate your contribution 👍

PR #1809, opened 2019-06-26
Fixes #201: Mysterious multiple update requests, users can’t update

### LINKED ISSUE #201: Mysterious multiple update requests, users can’t update
Mysterious multiple update requests lead to a temporary ban on our server. As a consequence, some users cannot update WP Rocket, not even via `rocket_force_update`.

Website on StageCoach where the behavior appears:
https://secure.helpscout.net/conversation/260889534/23733/?folderId=714999#thread-696756157

Another ticket where we don’t have access, but issue sounds related:
https://secure.helpscout.net/conversation/251247286/22318/?folderId=714999

@jbma and myself debugged on the StageCoach site. Even after deactivating all plugins except WP Rocket and switching to a default theme, the behavior remained the same: multiple requests, error 503 server response.

JB echoed `BOB` in `rocket_check_update()` – it got printed 9 times during the requests. So 9 requests were made to our server, and we have no idea why.

capture d ecran 2016-10-11 a 15 58 47

Screenfeed:
This will fix the issue where some sites are flooding the updates server.

- The class `Updater_Subscribers` is the service provider, registering the 3 new event subscribers.
- The class `Information_Subscriber` (1) manages plugin information (displayed in the "View version x.x.x details" popup).
- The class `Updater_Subscriber` (2) checks for updates (and has the icon URLs).
- The class `Updater_Api_Common_Subscriber` (3) manages hooks that are common to (1) and (2) (request user agent for example).
- The class `Updater_Api_Subscriber` is abstract and its only purpose is to provide common code to (1) and (2).

(2) is the class that could trigger several requests: it should not anymore, even when the URL arg `rocket_force_update` is used. The request response is stored in a transient for 1, 2, or 12 hours, depending on the response success.

Note: the `rocket_force_update` arg will work only if `set_site_transient( 'update_plugins', $obj )` is triggered in the page.


inc/classes/traits/trait-updater-api.php
```diff
@@ -0,0 +1,97 @@
...
+
+use WP_Rocket\Logger\Logger;
+
+/**
+ * Trait for the plugin updater.
+ *
+ * @since  3.3.6
+ * @author Grégory Viguier
+ */
+trait Updater_Api_Subscriber {
```
Tabrisrp:
```suggestion
trait Updater_Api_Tools {
```

inc/classes/traits/trait-updater-api.php
```diff
@@ -0,0 +1,97 @@
...
+	/**
+	 * An ID to use when a API request fails.
+	 *
+	 * @var    string
+	 * @since  3.3.6
+	 * @access protected
+	 * @author Grégory Viguier
+	 */
+	/*protected $request_error_id;*/
```
Tabrisrp: It can be removed instead of commented

inc/classes/subscriber/Plugin/class-updater-subscriber.php
```diff
@@ -0,0 +1,438 @@
...
+use WP_Rocket\Event_Management\Subscriber_Interface;
+
+/**
+ * Manages the plugin updates.
+ *
+ * @since  3.3.6
+ * @author Grégory Viguier
+ */
+class Updater_Subscriber implements Subscriber_Interface {
+	use \WP_Rocket\Traits\Updater_Api_Subscriber;
```
Tabrisrp:
```suggestion
	use \WP_Rocket\Traits\Updater_Api_Tools;
```

inc/classes/subscriber/Plugin/class-information-subscriber.php
```diff
@@ -0,0 +1,167 @@
...
+use WP_Rocket\Event_Management\Subscriber_Interface;
+
+/**
+ * Manages the plugin information.
+ *
+ * @since  3.3.6
+ * @author Grégory Viguier
+ */
+class Information_Subscriber implements Subscriber_Interface {
+	use \WP_Rocket\Traits\Updater_Api_Subscriber;
```
Tabrisrp:
```suggestion
	use \WP_Rocket\Traits\Updater_Api_Tools;
```

inc/classes/traits/trait-updater-api.php
Screenfeed: Imho it is better to keep it to document it, and prevent forgetting it.

Tabrisrp: Fair enough 👍

inc/classes/traits/trait-updater-api.php
Screenfeed: The file should be renamed into `trait-updater-api-tools` then, shouldn't it?

Tabrisrp: It would be best yes

Screenfeed:
🎉

PR #1841, opened 2019-07-15
Fixes #967 Send a request to the home URL after clearing the cache manually

### LINKED ISSUE #967: Proposal: Simulate a visit to the homepage after clearing the whole cache
When we click on "Clear cache" from the admin bar or the dashboard button, we purge all kind of caches: page, busting, minification/concatenation.

To avoid waiting a first visitor to generate the busted and minified files, we could simulate a visit on the homepage after the "Clear cache" action purged the homepage. It will help us to generate the files in "background".

Tabrisrp:


inc/common/purge.php
```diff
@@ -350,6 +350,8 @@ function do_admin_post_rocket_purge_cache() {
 					update_option( WP_ROCKET_SLUG, $options );
 				}
 
+				wp_safe_remote_get( home_url( $lang ) );
```
Screenfeed: Wouldn't it be a good idea to make it non-blocking?

PR #1866, opened 2019-07-25
CDN Feature rewrite

### LINKED ISSUE #50: CDN on images with new Royal Slider remove all images
When you enable CDN on images with Royal Slider enabled, it removes all the content of the div where there are usually the images.
Plugin is attached to the bug
Source : https://secure.helpscout.net/conversation/118475315/6471/

### LINKED ISSUE #65: CDN rewrite to the original image instead of the thumbnail
Once the CDN is enabled, images are rewritten to the original. From this ticket : https://secure.helpscout.net/conversation/169941850/11297/?folderId=273767
I've enabled CDN linker temporary which correctly rewrite the image.

### LINKED ISSUE #312: Apply CDN on <a> for compatiblity with Magic Zoom Plus.
By default we don't apply on <a> tags. It could be an issue with Magic Zoom Plus (https://www.magictoolbox.com/magiczoomplus/modules/woocommerce/).

This plugin doesn't seem to provide the image src with WordPress function so our CDN can't be applied and it causes an issue with the Zoom feature.

If we can be in touch with their developer or check the plugin's code to add the full compatibility with our CDN feature.

Related ticket support: https://secure.helpscout.net/conversation/307744188/31463?folderId=343182

### LINKED ISSUE #390: CDN URL replacement issue with data-src attribute
There’s an extra space that gets added before the `src` attribute in the `img` tag after the CDN URL replacement has occurred: 

```
<img  src="http://cdn.example.com/wp-content/uploads/2017/03/image.jpg" data-src="http://www.example.com/wp-content/uploads/2017/03/image.jpg" />
```

This isn't a problem per se, unless someone has an `img` tag with an empty `src` attribute but a populated `data-src` attribute:

```
<img src="" data-src="http://www.example.com/wp-content/uploads/2017/03/image.jpg" />
```

at which point, the `data-src` attribute will end up with that extra space inside of _it_ (causing the image to not display):

```
<img src="" data- src="http://cdn.example.com/wp-content/uploads/2017/03/image.jpg"  />
```

A secondary issue is that when both the  `src` and `data-src` attributes exist and are populated, only the `src` attribute gets the URL replacement (as shown in first example above).

Even if you try to get around the problem by entering the URL in the **Rejected files** field, the HTML still gets "touched", leaving the extra space behind even though the URL itself is not actually being changed. (This might be considered a separate bug all on its own.)

### LINKED ISSUE #392: Support Thrive leads
Here is some code I recently used to fix CDN for thrive leads. Adapt as you wish.

```
if ( function_exists( 'get_rocket_cdn_url' ) ) {
add_filter( 'tve_leads_ajax_load_forms', function ( $response ) {
foreach ( $response['res'] as $type => $urls ) {
array_walk( $response['res'][ $type ], function ( &$item ) {
$item = get_rocket_cdn_url( $item );
} );
}
return $response;
} );
}
```

### LINKED ISSUE #398: CDN URL replacement issue with WordPress Playlist Shortcode
Best way to describe is with examples. :)

HTML before CDN replacement:

```
<# if ( data.image ) { #>
<img src="{{ data.thumb.src }}" alt="" />
<# } #>
```

HTML after CDN replacement:

```
<# if ( data.image ) { #>
<img src="https://cdn.example.com/{{ data.thumb.src }}" alt="" />
<# } #>
```

The problem is that the value returned by `{{ data.thumb.src }}` is already a full URL, so it ends up being a value something like this:

```
https://cdn.example.com/https://example.com/image.jpg
```

which, of course, breaks the reference, causing the image to not display.

Edited to add: this HTML output apparently comes from WordPress's `playlist` shortcode; see function `wp_underscore_playlist_templates`: 
https://developer.wordpress.org/reference/functions/wp_underscore_playlist_templates/

### LINKED ISSUE #414: Add support of CDN rewrites for Avada’s parallax feature
Avada’s parallax feature stores image URLs in a `data-bg-image` attribute of a `<div>`. That attribute is not supported by WP Rocket’s CDN rewrites, and since it is on a `<div>`, the `rocket_cdn_images_html` filter cannot be used to sneak in a custom attribute to be considered by `rocket_cdn_images()`.

We could create a custom function to fix this. Patch to follow.

### LINKED ISSUE #561: Mega Menu
Apply the get_rocket_cdn_url function to the `megamenu_custom_icon_url` filter. Code however is generated in admin context.

### LINKED ISSUE #566: CDN Incorrect path rewriting for relative paths
CDN feature doesn’t recognize relative paths but still try to apply CNAME on an empty domain and it results in broken URLs like this one: **https:///wp-content/plugins/...**

### LINKED ISSUE #621: Images not served from CDN with visual composer
Visual composer seems to generate image sizes on the fly and WP rocket is serving the full image instead. 

Related ticket: https://secure.helpscout.net/conversation/478356727/54459/?folderId=1724284

Possibly related to issue #65

### LINKED ISSUE #975: Load svg files via CDN
Currently my SVG files are not loaded via CDN.

I'm aware, that often there is a SVG reference, that should not be loaded via CDN, but an image like
```
<img src="logo.svg" alt="">
```
can also be loaded via CDN

### LINKED ISSUE #1050: CDN image and "data-guid" image attribute both show up in tests
Seen most recently in the Uncode theme, when CDN is enabled, both the CDN replaced image and "data-guid" image attribute which references the original image URL are flagged in tests as loading the image twice.

Example: https://secure.helpscout.net/conversation/614890534/73158?folderId=1391600

### LINKED ISSUE #1585: Add CDN support for the <picture> tag
Currently, images, including WebP ones, added with the `picture` tag are not loaded when the CDN feature is enabled.

**Diff file:**
https://www.diffchecker.com/b2OAC9jS

**Related Ticket:**
https://secure.helpscout.net/conversation/807291484/101107/

Tabrisrp:
- Remove the old code handling the CDN feature in different locations
- New code uses a CDN class + a CDN subscriber class
- CDN is now applied only on `rocket_buffer` instead of various WP filters
- Keep some CDN related functions as an API in a new `api.php` file autoloaded with composer
- Add unit & integration tests

Closes #1585 
Closes #1050 
Closes #975 
Closes #621 
Closes #566 
Closes #561 
Closes #414 
Closes #398 
Closes #392 
Fixes #390 
Closes #312 
Fixes #65 
Fixes #50


inc/classes/CDN/CDN.php
```diff
@@ -0,0 +1,297 @@
...
+	 * @return boolean
+	 */
+	private function is_excluded( $url ) {
+		if ( 'php' === pathinfo( strtok( $url, '?' ), PATHINFO_EXTENSION ) ) {
+			return true;
+		}
+
+		$path = wp_parse_url( $url, PHP_URL_PATH );
+
+		if ( ! isset( $path ) ) {
```
Screenfeed: `wp_parse_url()` will return `false` on parsing failure.

inc/classes/CDN/CDN.php
```diff
@@ -0,0 +1,297 @@
...
+		/**
+			* Filter the excluded files.
+			*
+			* @since 2.5
+			*
+			* @param array $files List of excluded files.
+		*/
+		$files = (array) apply_filters( 'rocket_cdn_reject_files', $files );
+		$files = array_filter( $files );
+		$files = array_flip( array_flip( $files ) );
```
Screenfeed: It should be `preg_quote()`d, maybe by passing the delimiter as method’s parameter.

inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
```diff
@@ -139,7 +139,7 @@ public function insert_lazyload_script() {
 		$polyfill = apply_filters( 'rocket_lazyload_polyfill', false );
 
 		$script_args = [
-			'base_url' => get_rocket_cdn_url( WP_ROCKET_ASSETS_JS_URL . 'lazyload/', [ 'all', 'js', 'css_and_js' ] ),
+			'base_url' => 'base_url'  => WP_ROCKET_ASSETS_JS_URL . 'lazyload/',
```
Screenfeed: Your array is very strange ;)

inc/classes/CDN/CDN.php
Screenfeed: Well, this part is useless now.

inc/classes/CDN/CDN.php
```diff
@@ -0,0 +1,301 @@
...
+
+		return preg_quote( implode( '|', $files ), '|' );
```
Screenfeed: Assuming the result is used as a regex pattern, what I meant was more like:
```php
private function get_excluded_files( $delimiter ) {
	// ...
	$files = array_filter( $files );

	if ( ! $files ) {
		return '';
	}

	$files = array_flip( array_flip( $files ) );

	foreach ( $files as $i => $file ) {
		$files[ $i ] = preg_quote( $file, $delimiter );
	}

	return implode( '|', $files );
}
```

PR #1869, opened 2019-07-29
Exclude combination of jQuery served from c0.wp.com

vmanthos:
Related ticket:
https://secure.helpscout.net/conversation/914613004/117593/

### DIFF
--- a/inc/classes/optimization/JS/class-combine.php
+++ b/inc/classes/optimization/JS/class-combine.php
@@ -610,6 +610,7 @@ class Combine extends Abstract_JS_Optimization {
 			'code.tidio.co',
 			'www.uplaunch.com',
 			'widget.reviewability.com',
+			'c0.wp.com',
 		];
 
 		$excluded_external = array_merge( $defaults, $this->options->get( 'exclude_js', [] ) );


arunbasillal:
Related Slack conversation - https://wp-media.slack.com/archives/C43T1AYMQ/p1577479912012000

Summary: 

- When using JetPack site accelerator (Photon previously), the links are rewritten using this - https://jmp.sh/ueVJpNn
- Excluding `c0.wp.com` would prevent all JS files from being combined
- So we need to find an alternate solution to exclude this specific file.

vmanthos:
**Related ticket:**
https://secure.helpscout.net/conversation/1048999295/139694/

On an additional note, fixing the issue requires as to exclude the following URL:
`c0.wp.com/c/`

We cannot use wildcards, e.g. `c0.wp.com/c/(.*)/wp-includes/js/jquery/jquery.js` because of the way exclusions of URLs are handled here:
https://github.com/wp-media/wp-rocket/blob/72e1f60c4ae6f7f67900a33f559a63dcb7c8d364/inc/classes/optimization/JS/class-combine.php#L174

So, the existing solution has the same issues with this PR, i.e. excluding `c0.wp.com` would prevent all JS files from being combined.

Implementing the way we do it for internal files would allow us to use wildcards and target specifically the jQuery file, regardless of the WordPress version (the parenthesis part of the jQuery URL):
https://github.com/wp-media/wp-rocket/blob/72e1f60c4ae6f7f67900a33f559a63dcb7c8d364/inc/classes/optimization/JS/class-abstract-js-optimization.php#L81-L111

vmanthos:
Created a new PR about this:
https://github.com/wp-media/wp-rocket/pull/2296

PR #1870, opened 2019-07-29
Closes #486: Improve "Cache lifespan" setting to only delete cache entries older than set number of minutes/hours/days

### LINKED ISSUE #486: Improve "Cache lifespan" setting to only delete cache entries older than set number of minutes/hours/days
I checked how the `rocket_purge_time_event` WP cron event works and it just calls `rocket_clean_domain()`. So it purges the entire cache.

It would be great if this could be improved to only purge old cache entries. So if the "Cache lifespan would be set to 4 hours it would only clear the entries which were not regenerated in last 4 hours. So it would have to check that every 5 minutes or so.

It certainly makes more sense to me that way - why purge the entries which are frequently accessed?

Thanks,
Martin

Screenfeed:
Instead of deleting the whole cache folder periodically, delete only the old files.
Use partial preload after that.

Note: the cache folders will now use `#` as query strings delimiter instead of `_`.

Methods and hooks tagged for version 3.5.


inc/classes/subscriber/Optimization/class-automatic-cache-purge-subscriber.php
```diff
@@ -0,0 +1,461 @@
...
+	}
+
+	/**
+	 * Perform the event action.
+	 *
+	 * @since  3.5
+	 * @access public
+	 * @author Grégory Viguier
+	 */
+	public function do_event() {
```
Tabrisrp: can you rename the method to something more descriptive of what it does

PR #1875, opened 2019-07-31
Fixes #1867 - Append date and time to admin notices

### LINKED ISSUE #1867: Append date and time to admin notices
Feature request on behalf of a customer:

Problem:
It is not clear when an admin notice was generated or when an action (Preload, CPCSS etc) was completed

Solution:
Add a date and time stamp to the end of the message stating when the action was completed or message generated.

Link to request: https://secure.helpscout.net/conversation/913483428/117448/

crystinutzaa:
Added a date & time after some messages related to: PreLoad, Clear Cache, CPCSS

### DIFF
--- a/inc/admin/ui/notices.php
+++ b/inc/admin/ui/notices.php
@@ -764,25 +764,29 @@ function rocket_clear_cache_notice() {
 		case 'all':
 			if ( current_user_can( 'rocket_purge_cache' ) ) {
 				// translators: %s = plugin name.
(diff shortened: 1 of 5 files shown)


GeekPress:
We will have to re-translate all the strings. As the date doesn't have any relation with the string, it could be added after like this:

```
$notice = sprintf( __( '%s: Cache cleared.', 'rocket' ), '<strong>' . WP_ROCKET_PLUGIN_NAME . '</strong>' );
$notice .= ' <em>' . date("Y-m-d h:i:sa") .'</em>';
```

Then, it will avoid us to have to translate the strings.

Also, a more user friendly format date could help like:

`date_i18n( "F j, Y @ G:i", time() )
`

crystinutzaa:
I split the dates from orginal messages and I changed the use of date_i18n function.
However I think that for PreLoad Complete Time I still need a transient.
Did I missed any messages which should contain the timestamp?

PR #1884, opened 2019-08-05
Add SEOPress XML sitemaps to WP-Rocket preloading settings

wp-seopress:


remyperona (approved):

wp-seopress:
Hi there,

we have to update this integration to use the new methods.

How can we proceed to handle this?

Thx

piotrbak:
Hello @wp-seopress 

Could you create an issue with the description of proposed change and as many details as possible?

Would be good to know about backward compatibility, etc.

PR #1886, opened 2019-08-13
Closes #1458: webp compatibility

### LINKED ISSUE #1458: Media Webp support
It would be good if wp-rocket supports this type of plugins, to be able to use Webp (Other cache plugins already support it)
Wp-rocket should detect when the browser is compatible and load the image in Webp format
Plugin: https://wordpress.org/plugins/media-webp/

A plugin that does this: https://wordpress.org/plugins/cache-enabler/

Screenfeed:
Returning `true` to the filter `'rocket_disable_webp_cache'` provides a quick way to disable webp cache.
For more advanced management, like adding compatibility with a third party webp plugin, an instance using the interface `WP_Rocket\Subscriber\Third_Party\Plugins\Images\Webp\Webp_Interface` must be added to the array filtered by `'rocket_webp_plugins'`.


inc/classes/subscriber/Media/class-webp-subscriber.php
```diff
@@ -0,0 +1,593 @@
...
+	 * @since  3.4
+	 * @access public
+	 * @author Remy Perona
+	 * @author Grégory Viguier
+	 *
+	 * @param string $html HTML content.
+	 * @return string
+	 */
+	public function convert_to_webp( $html ) {
+		if ( ! $this->options->get( 'cache_webp' ) || $this->get_plugins_serving_webp() ) {
```
Tabrisrp: Shouldn't it be `! $this->get_plugins_serving_webp()`?

inc/classes/subscriber/third-party/plugins/Images/Webp/class-shortpixel-subscriber.php
```diff
@@ -0,0 +1,296 @@
...
+	private $options;
+
+	/**
+	 * CDNSubscriber instance.
+	 *
+	 * @var    CDNSubscriber
+	 * @access private
+	 * @author Grégory Viguier
+	 */
+	private $cdn_subscriber;
```
Tabrisrp:
```suggestion
	private $cdn;
```

inc/classes/subscriber/third-party/plugins/Images/Webp/class-shortpixel-subscriber.php
```diff
@@ -0,0 +1,296 @@
...
+	 * @var    Options_Data
+	 * @access private
+	 * @author Remy Perona
+	 */
+	private $options;
+
+	/**
+	 * CDNSubscriber instance.
+	 *
+	 * @var    CDNSubscriber
```
Tabrisrp:
```suggestion
	 * @var    CDN
```

inc/classes/subscriber/third-party/plugins/Images/Webp/class-shortpixel-subscriber.php
```diff
@@ -0,0 +1,296 @@
...
+	 * Options_Data instance.
+	 *
+	 * @var    Options_Data
+	 * @access private
+	 * @author Remy Perona
+	 */
+	private $options;
+
+	/**
+	 * CDNSubscriber instance.
```
Tabrisrp:
```suggestion
	 * CDN instance.
```

inc/classes/subscriber/Media/class-webp-subscriber.php
```diff
@@ -0,0 +1,593 @@
+<?php
+namespace WP_Rocket\Subscriber\Media;
+
+use WP_Rocket\Admin\Options_Data;
+use WP_Rocket\Subscriber\CDN\CDNSubscriber;
+use WP_Rocket\Event_Management\Subscriber_Interface;
+use WP_Rocket\Subscriber\Third_Party\Plugins\Images\Webp\Webp_Interface;
```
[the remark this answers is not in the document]
Screenfeed: Only in `get_webp_plugins()` but this `use` is indeed useless.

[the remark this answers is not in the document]
Screenfeed: Well, the `use` is only useful for comments, so we may consider removing it.

inc/classes/subscriber/Media/class-webp-subscriber.php
Screenfeed: No.
If a plugin is serving webp, then we don't need to do anything.

PR #1887, opened 2019-08-14
Improvement new TestCase class for unit tests

Screenfeed:
This class contains `setUp()` and `tearDown()`, which are common to all our tests. Tests will then extend the class, and extend the methods if needed.

### DIFF
--- a/tests/Unit/Functions/Options/TestExcludeDeferJS.php
+++ b/tests/Unit/Functions/Options/TestExcludeDeferJS.php
@@ -1,7 +1,7 @@
 <?php
 namespace WP_Rocket\Tests\Unit\Functions\Options;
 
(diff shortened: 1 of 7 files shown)


tests/Unit/Functions/Options/TestExcludeDeferJS.php
```diff
@@ -1,7 +1,7 @@
 <?php
 namespace WP_Rocket\Tests\Unit\Functions\Options;
 
-use PHPUnit\Framework\TestCase;
+use WP_Rocket\Tests\Unit\TestCase;
 use Brain\Monkey;
```
Tabrisrp: You can remove this line too here, and in the `setUp()` function on line 12, remove the `Monkey\setUp();` while keeping the `parent::setUp()`

Issue #1564, opened 2019-03-07
Caching by default utm_term query string


WordPresseur:
WP Rocket actually doesn't provide standard cache file for this query string: `utm_term`

There's no reason it shouldn't.

PR #1892, opened 2019-08-20
Allow standard caching query parameters to be extended. 

clifgriffin:
Howdy!

This PR does two things. 

1. It corrects an oversight in the solution for this issue:
https://github.com/wp-media/wp-rocket/issues/1564

`utm_term` was added to the Buffer testing class, but it doesn't appear to have been added to `inc/front/process.php`, which means the parameter was never fully implemented. (Unless I'm missing something!)

2. It adds a constant `WP_ROCKET_STANDARD_CACHING_QUERY_PARAMS` that allows users to add additional parameters for standard caching. These parameters are merged with the default list, so users are able to extend but not remove the default parameters. 

Happy to make changes if there are any issues with this proposal. 

Thanks!
Clif

### DIFF
--- a/inc/front/process.php
+++ b/inc/front/process.php
@@ -103,6 +103,7 @@ $rocket_remove_query_strings = [
 	'utm_medium'      => 1,
 	'utm_campaign'    => 1,
 	'utm_expid'       => 1,
+	'utm_term'        => 1,
 	'fb_action_ids'   => 1,
 	'fb_action_types' => 1,
 	'fb_source'       => 1,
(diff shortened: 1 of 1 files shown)


Tabrisrp:
Hi,

This is already handled in the PR #1865, using a filter instead.

`process.php` is deprecated, so it won't be updated anymore. It's there as backward-compatibility during updates from older versions.

clifgriffin:
Well I feel dumb!

So sorry for wasting your time.

Tabrisrp:
Don't feel dumb, it was not easy to know if you are not in the knows of the daily what's going on in the team.

Thank you for taking the time to contribute, we appreciate it a lot!

PR #1952, opened 2019-09-16
Exclude inline JavaScript keywords

vmanthos:
Exclude inline JavaScript keywords to avoid the cache directory size issue.


inc/classes/optimization/JS/class-combine.php
```diff
@@ -515,6 +515,10 @@ protected function get_excluded_inline_content() {
 			'styles: \' #custom-menu-',
 			'function svc_center_',
 			'#svc_carousel2_container_',
+			'$("#myCarousel',
```
Tabrisrp: isn't this line JQuery dependant? If yes, it should be in the other exclusion method

vmanthos: My bad @Tabrisrp .

PR #1957, opened 2019-09-18
Fixes #1940: Minify inline JS & CSS - prevent PREG_BACKTRACK_LIMIT_ERROR

### LINKED ISSUE #1940: HTML minification + var wc_product_block_data
WooCommerce recently started inserting this inline script, beginning: 
`var wc_product_block_data`

In some cases it can be really large and seems to cause a problem - a blank screen - with HTML minification. 

Sample tickets:

- https://secure.helpscout.net/conversation/942490237/120952/
- https://secure.helpscout.net/conversation/940796878/120724?folderId=2952229

Related conversation: 
https://wp-media.slack.com/archives/C43T1AYMQ/p1567089731008200

crystinutzaa:
Big inline JS scripts are causing PREG_BACKTRACK_LIMIT_ERROR and the page will be empty.
preg_replace_callback - on errors the return is NULL

Solution: if preg_replace_callback return is NULL, then do not minify any inline JS code to prevent this issue. 

The issue was identified from: WooCommerce Blocks plugin:
https://wordpress.org/plugins/woo-gutenberg-products-block/

There is excessive JS (1.5MB of JS) from it causing the issue with WP Rocket: https://wordpress.org/support/topic/causes-excessive-js/

### DIFF
--- a/inc/vendors/classes/class-minify-html.php
+++ b/inc/vendors/classes/class-minify-html.php
@@ -95,7 +95,7 @@ class Minify_HTML
         }
 
         $this->_replacementHash = 'MINIFYHTML' . md5($_SERVER['REQUEST_TIME']);
-        $this->_placeholders = array();
+        $this->_placeholders    = array();
 
         // replace SCRIPTs (and minify) with placeholders
         // preg_replace_callback - on errors the return is NULL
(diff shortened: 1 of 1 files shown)


Tabrisrp:
Following this comment https://github.com/wp-media/wp-rocket/issues/1940#issuecomment-537278914, can you also update this to handle errors for CSS

PR #1963, opened 2019-09-19
Fixes #1931: Minify HTML do not take into account attribute values that span multiple lines

### LINKED ISSUE #1931: Minify HTML do not take into account attribute values that span multiple lines.
Minify HTML will break at trimming each line when attributes span on multiple lines:

This line is causing the issue:
https://github.com/wp-media/wp-rocket/blob/7f9af51eec7336d4bd4d7a0193be8d9f06750d99/inc/vendors/classes/class-minify-html.php#L131

I fixed it replacing with a " ":
`$this->_html = preg_replace('/^\\s+|\\s+$/m', ' ', $this->_html);`

Checked other solutions, but it seems this is the only one (extra space) or by changing the trim line with this code which adds a line instead an extra space:
`//Remove blank useless space`
`        $search = array(`
`            '@( |\t|\f)+@',    // Shorten multiple whitespace sequences`
`            '@(^[\r\n]*|[\r\n]+)[\s\t]*[\r\n]+@',    //Remove blank lines`
`            '@^(\s)+|( |\t|\0|\r\n)+$@' //Trim Lines`
 `       );`
 `       $replace = array(' ',"\\1",'');`
`        $this->_html = preg_replace($search, $replace, $this->_html);`

https://cl.ly/83cc119afb9e

https://secure.helpscout.net/conversation/939694690/120616?folderId=3083222
https://secure.helpscout.net/conversation/943488407/121105?folderId=3083222

crystinutzaa:
Minify HTML will break at trimming each line when attributes span on multiple lines:

This line is causing the issue:
wp-rocket/inc/vendors/classes/class-minify-html.php

$this->_html = preg_replace('/^\\s+|\\s+$/m', '', $this->_html); 

SOLUTION:

// Fixed attribute values which span on multiple lines. Causes double spaces "  "
$this->_html = preg_replace('/^\\s+|\\s+$/m', ' ', $this->_html);
// Fixed double spaces. Replaced with a single space
$this->_html = preg_replace('/  +/', ' ', $this->_html);

### DIFF
--- a/inc/vendors/classes/class-minify-html.php
+++ b/inc/vendors/classes/class-minify-html.php
@@ -128,7 +128,10 @@ class Minify_HTML
 
         // trim each line.
         // @todo take into account attribute values that span multiple lines.
-        $this->_html = preg_replace('/^\\s+|\\s+$/m', '', $this->_html);
+        // Fixed attribute values which span on multiple lines. Causes double spaces "  "
+        $this->_html = preg_replace('/^\\s+|\\s+$/m', ' ', $this->_html);
+        // Fixed double spaces. Replaced with a single space
+        $this->_html = preg_replace('/  +/', ' ', $this->_html);
 
         // remove ws around block/undisplayed elements
         $this->_html = preg_replace('/\\s+(<\\/?(?:area|article|aside|base(?:font)?|blockquote|body'
@@ -248,7 +251,7 @@ class Minify_HTML
 
     protected function _removeCdata($str)
     {
-	   	$data = array();
+	    $data = array();
 
 	    if ( false !== strpos( $str, '<![CDATA[' ) ) {
 		    $data['content'] = str_replace( array( '/* <![CDATA[ */', '/* ]]> */' ), '', $str );
--- a/inc/vendors/classes/class-minify-html.php
+++ b/inc/vendors/classes/class-minify-html.php
@@ -129,9 +129,9 @@ class Minify_HTML
         // trim each line.
         // @todo take into account attribute values that span multiple lines.
         // Fixed attribute values which span on multiple lines. Causes double spaces "  "
-        $this->_html = preg_replace('/^\\s+|\\s+$/m', ' ', $this->_html);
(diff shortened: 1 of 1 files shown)


inc/vendors/classes/class-minify-html.php
```diff
@@ -128,7 +128,10 @@ public function process()
 
         // trim each line.
         // @todo take into account attribute values that span multiple lines.
-        $this->_html = preg_replace('/^\\s+|\\s+$/m', '', $this->_html);
+        // Fixed attribute values which span on multiple lines. Causes double spaces "  "
+        $this->_html = preg_replace('/^\\s+|\\s+$/m', ' ', $this->_html);
+        // Fixed double spaces. Replaced with a single space
+        $this->_html = preg_replace('/  +/', ' ', $this->_html);
```
Tabrisrp:
```suggestion
        $this->_html = preg_replace('/\s+/', ' ', $this->_html);
```

@Screenfeed You confirm?

Screenfeed: Both should work, but I prefer `'/\s+/'` indeed.
What I don't get is the double `\` in `'/^\\s+|\\s+$/m'`. Unless I'm missing something it should be `'/^\s+|\s+$/m'`.

crystinutzaa: Changed from` '/  +/'` into `'/\s+/'`

As for the double `\` in '`/^\\s+|\\s+$/m'`. in  this was in the original file:
https://github.com/mrclay/minify/blob/8e4b0ceb86a4e89bb8031715d1e0f75e953a925f/lib/Minify/HTML.php#L131

I changed into single `\` and tested and is working.

Screenfeed: I saw was in the original file but I'm surprised it was even working ¯\_(ツ)_/¯
Thanks :)
Is it all fine for you @Tabrisrp?

Tabrisrp: yes all good

PR #1965, opened 2019-09-19
Unit tests for class EwwwSubscriber

Screenfeed:
Unit tests for webp compatibility with EWWW.


Tabrisrp:
There is a conflict to resolve before merging

PR #1966, opened 2019-09-19
Fix preloader not preloading webp cache

Screenfeed:
The header `Accept` was missing and has been added.
Also, the request to the home page was not filtered, it has been corrected.

### DIFF
--- a/inc/classes/subscriber/Media/class-webp-subscriber.php
+++ b/inc/classes/subscriber/Media/class-webp-subscriber.php
@@ -94,15 +94,16 @@ class Webp_Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'rocket_buffer'                           => [ 'convert_to_webp', 23 ],
-			'rocket_cache_webp_setting_field'         => [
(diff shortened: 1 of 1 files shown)


inc/classes/subscriber/Media/class-webp-subscriber.php
```diff
@@ -377,6 +378,7 @@ public function add_accept_header( $args ) {
 			return $args;
 		}
 
+		$args['headers']['Accept']      = 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8';
 		$args['headers']['HTTP_ACCEPT'] = 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8';
```
Tabrisrp: Do we need this if we have the line above?

Screenfeed: Yes we need both: https://github.com/wp-media/wp-rocket/blob/5492dcdf444156077eb887c3bcab3df0b73e4c0c/inc/classes/subscriber/Media/class-webp-subscriber.php#L135-L140

PR #1975, opened 2019-09-24
Capabilities setup

Tabrisrp:
- Add/Remove capabilities on plugin's activation/deactivation


inc/main.php
```diff
@@ -149,6 +149,7 @@ function rocket_deactivation() {
 	wp_clear_scheduled_hook( 'rocket_google_tracking_cache_update' );
 	wp_clear_scheduled_hook( 'rocket_cache_dir_size_check' );
 
+	( new WP_Rocket\Subscriber\Plugin\Capabilities_Subscriber() )->remove_rocket_capabilities();
```
Screenfeed: I wonder if we should add this after `do_action( 'rocket_deactivation' );` instead. A callback hooked to it may need this.

PR #1998, opened 2019-09-30
Display admin notice for CPCSS outside WP Rocket admin page

crystinutzaa:
Added admin notices for CPCSS generation allowed outside WP Rocket settings page

Issue demo - https://jmp.sh/JZiRHTc

Based on:
3.4.1 WP Rocket Minor Version doc

### DIFF
--- a/inc/classes/subscriber/Optimization/class-critical-css-subscriber.php
+++ b/inc/classes/subscriber/Optimization/class-critical-css-subscriber.php
@@ -172,13 +172,13 @@ class Critical_CSS_Subscriber implements Subscriber_Interface {
 
 		$screen = get_current_screen();
 
+		if ( 'settings_page_wprocket' !== $screen->id ) {
+		    return;
+		}
+
 		$transient = get_transient( 'rocket_critical_css_generation_process_running' );
 		if ( ! $transient ) {
 			return;
-		}
-
-		if ( 'settings_page_wprocket' !== $screen->id ) {
-		    return;
 		}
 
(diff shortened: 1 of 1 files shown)


arunbasillal:
It would be ideal if it can be a message like Preload for non-WP Rocket pages - https://jmp.sh/Zfm26vw

`Critical CSS generation is currently running. Go to the WP Rocket settings page to track progress.`

Please remember to check if the user has the capability to see the settings page (`rocket_manage_options`).

QA Test: 
- Give an author user only `rocket_regenerate_critical_css` capability. 
- Visit WordPress dashboard.
- Regenerate Critical Path CSS.
- There should be an admin notice. 
- Repeat the test with an admin user.

crystinutzaa:
Fixed. Displayed the message:
`Critical CSS generation is currently running. Go to the WP Rocket settings page to track progress.`

inc/classes/subscriber/Optimization/class-critical-css-subscriber.php
```diff
@@ -125,15 +172,15 @@ public function critical_css_generation_running_notice() {
...
-		}
-
 		$transient = get_transient( 'rocket_critical_css_generation_process_running' );
 		if ( ! $transient ) {
 			return;
 		}
 
+		if ( 'settings_page_wprocket' !== $screen->id ) {
+		    return;
+		}
```
Tabrisrp: Why change the position of this conditional check?

crystinutzaa: Sorry, I copied it back in the wrong position while testing something. No need to be moved below get_transient();.

PR #2004, opened 2019-10-02
Fix 1917 - Autocorrect wilcard rules on never cache urls

crystinutzaa:
Autocorrect wildcard rules on Never Cache Url(s)

Incorrect wildcard rules as * , .* will be automatically replaced by the correct rule (.*)

Based on:
3.4.1 WP Rocket Minor Version doc

### DIFF
--- a/inc/functions/formatting.php
+++ b/inc/functions/formatting.php
@@ -29,25 +29,27 @@ function rocket_clean_exclude_file( $file ) {
  * @param  string       $path URL which needs to be cleaned
  * @return bool\string  false if $path is empty or cleaned URL
  */
-function rocket_clean_wildcards ( $path ) {
-	if ( ! $path ) {
-		return false;
-	}
+function rocket_clean_wildcards( $path ) {
+    if ( ! $path ) {
+        return false;
+    }
 
-	$path_components = explode( '/', $path );
+    $path_components = explode( '/', $path );
     $arr             = array(
             ".*"   => "(.*)" ,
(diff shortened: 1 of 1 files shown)


inc/functions/formatting.php
```diff
@@ -20,6 +20,35 @@ function rocket_clean_exclude_file( $file ) {
...
+ */
+function rocket_clean_wildcards ( $path ) {
+	if ( ! $path ) {
+		return false;
+	}
+
+	$path_components = explode( '/', $path );
+    $arr             = array(
+            ".*"   => "(.*)" ,
+            "*"    => "(.*)" ,
```
Screenfeed: I suppose we could also add `'(*)' => '(.*)',`.
Also, phpcs won't be happy.

Screenfeed: Nice use of `strtr()` 👍

crystinutzaa: Added also: `'(*)' => '(.*)' rule`

Applied to CSS/JS files exclusions, User Agents exclusion, Always Purged URLs, and CDN exclusions

Screenfeed: Last changes about styling:
- `$arr             = array(`: short array syntax please.
- `".*"   => "(.*)" ,`: no space character before the last `,` for the 4 lines.

crystinutzaa: Coding style changes applied. Also I changed all array() into [] shorthand. It's confusing to see array()

PR #2015, opened 2019-10-07
3.4.0.4

### LINKED ISSUE #2014: 3.4: CDN set to Images only - srcset not rewritten
When the CDN is set only to Images, the URLs in `srcset` are not rewritten, only the `img` tag is.

LazyLoad_-_Mission_Control

Slack: https://wp-media.slack.com/archives/C08N8J6VC/p1570216373020700?thread_ts=1570204755.010300&cid=C08N8J6VC

Related ticket: https://secure.helpscout.net/conversation/970255465/125157?folderId=377611

Tabrisrp:
- Fixes #2014 : Rewrite `srcset` URLs when the zone is set to `images`
- Prevent a PHP fatal error if the `administrator` role doesn't exist when adding custom capabilities
- Remove references to the IT and DE docs

### DIFF
--- a/inc/classes/CDN/CDN.php
+++ b/inc/classes/CDN/CDN.php
@@ -49,14 +49,14 @@ class CDN {
 	/**
 	 * Rewrites URLs in a srcset attribute using the CDN URL
 	 *
-	 * @since 3.4
+	 * @since 3.4.0.4
 	 * @author Remy Perona
 	 *
 	 * @param string $html HTML content.
 	 * @return string
 	 */
 	public function rewrite_srcset( $html ) {
-		$pattern = '#\s+(?:data-lazy-|data-)?srcset\s*=\s*["\']\s*([^"\',\s]+\.[^"\',\s]+(?:\s+\d+(?:w|px|r?em))?(?:\s*,\s*[^"\',\s]+\.[^"\',\s]+\s+\d+(?:w|px|r?em))*)\s*["\']#i';
+		$pattern = '#\s+(?:data-lazy-|data-)?srcset\s*=\s*["\']\s*([^"\',\s]+\.[^"\',\s]+(?:\s+\d+[wx])?(?:\s*,\s*[^"\',\s]+\.[^"\',\s]+\s+\d+[wx])*)\s*["\']#i';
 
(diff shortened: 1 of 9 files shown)


inc/classes/CDN/CDN.php
```diff
@@ -36,16 +36,47 @@ public function __construct( Options_Data $options ) {
...
+				return str_replace( $matches['url'], $this->rewrite_url( $matches['url'] ), $matches[0] );
 			},
 			$html
 		);
 	}
 
+	/**
+	 * Rewrites URLs in a srcset attribute using the CDN URL
+	 *
+	 * @since 3.4
```
Screenfeed:
```
@since 3.4.0.4
```

inc/classes/CDN/CDN.php
```diff
@@ -36,16 +36,47 @@ public function __construct( Options_Data $options ) {
...
+	 * @author Remy Perona
+	 *
+	 * @param string $html HTML content.
+	 * @return string
+	 */
+	public function rewrite_srcset( $html ) {
+		$pattern = '#\s+(?:data-lazy-|data-)?srcset\s*=\s*["\']\s*([^"\',\s]+\.[^"\',\s]+(?:\s+\d+(?:w|px|r?em))?(?:\s*,\s*[^"\',\s]+\.[^"\',\s]+\s+\d+(?:w|px|r?em))*)\s*["\']#i';
```
Screenfeed: `(?:w|px|r?em)`: apparently it should be `[wx]`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-srcset

inc/classes/subscriber/CDN/CDNSubscriber.php
```diff
@@ -69,6 +72,23 @@ public function rewrite( $html ) {
 		return $this->cdn->rewrite( $html );
 	}
 
+	/**
+	 * Rewrites URLs in srcset attributes to the CDN URLs if allowed
+	 *
+	 * @since 3.4
```
Screenfeed:
```
@since 3.4.0.4
```

PR #2021, opened 2019-10-09
Varnish Add-on improvements

### LINKED ISSUE #1128: Please make rocket_varnish_ip an Array
Could you please modify the 'rocket_varnish_http_purge' function so that it can handle an array of Varnish server IPs ?

$varnish_ip would become an array instead of a string (maybe rename to $varnish_ips) then you'd loop over each ip to execute wp_remote_request.

I already made a custom plugin, you can find the useful parts below.

**Additional context**

```php
<?php

add_filter( 'rocket_varnish_ip', function ( $ip ) {
        return array(
'127.0.0.1',
'10.129.21.25',
        );
} );

function mvva_rocket_varnish_http_purge( $url ) {
        $parse_url = get_rocket_parse_url( $url );

        $varnish_x_purgemethod = 'default';
        $regex = '';

        if ( 'vregex' === $parse_url['query'] ) {
$varnish_x_purgemethod = 'regex';
$regex = '.*';
        }

        /**
* Filter the Varnish IP to call
*
* @since 2.6.8
*
* @param string The Varnish IP
*/
        $varnish_ips = apply_filters( 'rocket_varnish_ip', [] );

        if ( defined( 'WP_ROCKET_VARNISH_IP' ) && empty( $varnish_ips ) ) {
$varnish_ips = array( WP_ROCKET_VARNISH_IP );
        }

        /**
* Filter the HTTP protocol (scheme)
*
* @since 2.7.3
*
* @param string The HTTP protocol
*/
        $scheme = apply_filters( 'rocket_varnish_http_purge_scheme', 'http' );

        foreach ( $varnish_ips as $varnish_ip ) {
$parse_url['host'] = ( $varnish_ip ) ? $varnish_ip : $parse_url['host'];
$purgeme = $scheme . '://' . $parse_url['host'] . $parse_url['path'] . $regex;

wp_remote_request(
$purgeme,
array(
'method' => 'PURGE',
'blocking' => false,
'redirection' => 0,
/**
* Filters the headers to send with the Varnish purge request
*
* @since 3.1
* @author Remy Perona
*
* @param array $headers Headers to send.
*/
'headers' => apply_filters( 'rocket_varnish_purge_headers', array(
/**
* Filters the host value passed in the request headers
*
* @since 2.8.15
*
* @param string The host
*/
'host' => apply_filters( 'rocket_varnish_purge_request_host', $parse_url['host'] ),
'X-Purge-Method' => $varnish_x_purgemethod,
) ),
)
);
        }
}

remove_action( 'before_rocket_clean_domain', 'rocket_varnish_clean_domain', 10 );
remove_action( 'before_rocket_clean_file', 'rocket_varnish_clean_file', 10 );
remove_action( 'before_rocket_clean_home', 'rocket_varnish_clean_home', 10 );

add_action( 'before_rocket_clean_domain', function ( $root, $lang, $url ) {
        mvva_rocket_varnish_http_purge( trailingslashit( $url ) . '?vregex' );
}, 10, 3 );

add_action( 'before_rocket_clean_file', function ( $url ) {
        mvva_rocket_varnish_http_purge( trailingslashit( $url ) . '?vregex' );
}, 10, 1 );

add_action( 'before_rocket_clean_home', function ( $root, $lang ) {
        $home_url = trailingslashit( get_rocket_i18n_home_url( $lang ) );
        $home_pagination_url = $home_url . trailingslashit( $GLOBALS['wp_rewrite']->pagination_base ) . '?vregex';

        mvva_rocket_varnish_http_purge( $home_url );
        mvva_rocket_varnish_http_purge( $home_pagination_url );
}, 10, 2 );
```

### LINKED ISSUE #1478: Allow changing Varnish Purge method
Hello,

I'd like to change the Varnish Purge Method to something else.
Most of the time I have BAN method and not PURGE.

Add a filter to the wp_remote_request to change any parameter.

Currently I'm adding a mu-plugin replaying exactly the same request but with the BAN word.

### LINKED ISSUE #1527: HTTPS scheme gets overwritten with HTTP with a PURGE request to Varnish
Hello,

Referenced file: wp-rocket/inc/functions/varnish.php

This way it seems to work for us:

```
        $scheme = apply_filters( 'rocket_varnish_http_purge_scheme', $parse_url['scheme']);

        $parse_url['host'] = ( $varnish_ip ) ? $varnish_ip : str_replace('*','',$parse_url['host']);
        $purgeme           = $scheme . '://' . $parse_url['host'] . $parse_url['path'] . $regex;
```

It used to fail to purge Varnish cache due to 2 reasons:

1. Wrong scheme, the plugin calls http when is should be https. We have nginx in front of Varnish and both http and https are proxied to Varnish.

2. We have multi-host environment we need to specify valid domain with PURGE request. So we remove * from a domain name.

Regards,
Alex.

### LINKED ISSUE #1534: Varnish add-on compatibility
Since version 3 of Varnish PURGE requests are not compatible with the Regex by default. The popular workaround is to use the Varnish BAN function when PURGE request is made.
https://varnish-cache.org/docs/6.0/users-guide/purging.html (_ban lurker_ section)

However, this solution is very dependent on the customers VCL configuration and will fail if the expected configuration is not present.

BAN and PURGE are very similar, the only difference is that banned items are still being stored in the memory (but they can't be used anymore), while Purged are being deleted.

I think that sending both requests BAN and PURGE, one by one would resolve the problem and I'm not able to find any weak sides of this solution:
- If the BAN request works, we will also try to delete cached items using PURGE. 
- If BAN fails, we will also try to delete it
- If PURGE fails in any of the above scenarios, nothing happens at all.
Above invalidate options should work with every version of Varnish.

There also seems to be one blind spot more with the purging of Varnish cache:

I'm not sure why do we hardcode the HTTP scheme here:
https://github.com/wp-media/wp-rocket/blob/430a0691df2c27947f8cb4ba4b4db4c9ab58f319/inc/functions/varnish.php#L41

And why '*' is appended to the end of the variable here:
https://github.com/wp-media/wp-rocket/blob/430a0691df2c27947f8cb4ba4b4db4c9ab58f319/inc/functions/varnish.php#L43

Here's the solution provided by one of our customers which fixed an issue for him and it needs further tests:
https://github.com/wp-media/wp-rocket/issues/1527

Tabrisrp:
* Switch Varnish Add-on from one-click to Rocket Add-on
* Add option for custom Varnish IP
* Allow to clear multiple Varnish IPs from one request
* Switch code to Classes/Subscribers
* Closes #1534 
* Fixes #1527 
* Closes #1478 
* Closes #1128


inc/deprecated/3.5.php
```diff
@@ -1,6 +1,70 @@
...
+ * Purge all the domain
+ *
+ * @since 2.6.8
+ *
+ * @param string $root The path of home cache file.
+ * @param string $lang The current lang to purge.
+ * @param string $url  The home url.
+ */
+function rocket_varnish_clean_domain( $root, $lang, $url ) {
+    rocket_varnish_http_purge( trailingslashit( $url ) . '?vregex' );
```
Screenfeed: For all functions in the file, missing `_deprecated_function()` and DocBlock tags.

tests/Integration/Subscriber/Addons/CloudflareSubscriber/TestGetSubscribedEvents.php
```diff
@@ -0,0 +1,19 @@
...
+namespace WP_Rocket\Tests\Integration\Subscriber\Addons\CloudflareSubscriber;
+
+use PHPUnit\Framework\TestCase;
+use WP_Rocket\Subscriber\Addons\Cloudflare\CloudflareSubscriber;
+
+class TestGetSubscribedEvents extends TestCase {
+	public function testShouldReturnSubscribedEventsArray() {
+        $events = [
+            'rocket_varnish_ip'                 => 'set_varnish_localhost',
+			'rocket_varnish_purge_request_host' => 'set_varnish_purge_request_host',
```
Screenfeed: Alignment issue.

views/settings/fields/text.php
```diff
@@ -33,5 +33,10 @@
 			<label for="<?php echo esc_attr( $data['id'] ); ?>"><?php echo $data['label']; ?></label>
 		<?php endif; ?>
 		<input type="text" id="<?php echo esc_attr( $data['id'] ); ?>" class="" name="wp_rocket_settings[<?php echo esc_attr( $data['id'] ); ?>]" value="<?php echo esc_attr( $data['value'] ); ?>"<?php echo $data['input_attr']; ?>>
+		<?php if ( ! empty( $data['helper'] ) ) : ?>
+		<div class="wpr-field-description wpr-field-description-helper">
+			<?php echo $data['helper']; ?>
+		</div>
+	<?php endif; ?>
```
Screenfeed: Alignment issue.

PR #2022, opened 2019-10-09
Fixes Authorize.net Seal

ScottTravisHartley:
If you're using the authorize.net seal JS on the website and have the defer JS option enabled it breaks. Excluding the seals JS resolves the issues. 

here is the direct url for the JS: https://verify.authorize.net/anetseal/seal.js


hellofromtonya:
@crystinutzaa Merge this one into 3.4.4 to ensure it goes through a full QA and release cycle.

tests/Integration/Functions/Options/TestExcludeDeferJS.php
```diff
@@ -40,7 +41,8 @@ public function testShouldReturnExcludeDeferJSArrayWhenSafeMode() {
             'js.hsforms.net',
             'www.uplaunch.com',
             'google.com/recaptcha',
-            'widget.reviews.co.uk',
+			'widget.reviews.co.uk',
```
hellofromtonya: Fix the array element tab alignment.

tests/Unit/Functions/Options/TestExcludeDeferJS.php
```diff
@@ -26,6 +26,7 @@ public function testShouldReturnExcludeDeferJSArray() {
             'www.uplaunch.com',
             'google.com/recaptcha',
             'widget.reviews.co.uk',
+            'verify.authorize.net/anetseal',
```
hellofromtonya: @crystinutzaa This array of deferred scripts, i.e. `$exclude_defer_js` is repeated in the next test. Let's pull these out of each test and move them into `setUp()` and a property such as `$this->expected_deferred_js`. Then in the next test, add the new deferred scripts to it.

Why? It'll help us as we add more deferred scripts in the future, i.e. one place to make changes instead of 2.

crystinutzaa: I moved it in a function which returns the $exclude_defer array with excluded jquery files based on function param:
`public function get_exclude_defer_js_list( $defer_jquery ) {`

hellofromtonya:
Thank you @ScottTravisHartley for contributing to WP Rocket! @crystinutzaa updated our tests to validate your contribution. All looks good and this PR has been approved for merge as part of Rocket 3.4.4.

We appreciate your contribution 👍

PR #2023, opened 2019-10-09
Fix #1922 Cloudflare flexible SSL redirect loop

### LINKED ISSUE #1922: Enhance Cloudflare plugin warning, re: flexible SSL
@vmanthos previously documented an issue on our internal KB about a "too many redirects" error that can occur when a customer: 

- Is using Flexible SSL in Cloudflare
- Is using the official Cloudflare plugin
- Prompted by us, they deactivate the Cloudflare plugin
- Too many redirects ensue!
- Customer can't access their wp-admin
This is obviously a bad experience for them.

I feel that we should modify our Cloudflare plugin warning to suggest that if they are using Flexible SSL they should deactivate our add-on instead.

crystinutzaa:
Fix for Cloudflare Flexible SSL redirect loop

### DIFF
--- a/inc/common/cloudflare-flexible-ssl.php
+++ b/inc/common/cloudflare-flexible-ssl.php
@@ -1,0 +1,77 @@
+<?php
+defined('ABSPATH') || die('Cheatin&#8217; uh?');
+
+/**
+ * Check if request is from Cloudflare
+ *
+ * @since  3.4.1
+ * @author Soponar Cristina
+ *
+ * @return bool
+ */
+function rocket_is_cloudflare() {
+     if ( ! isset( $_SERVER['HTTP_CF_CONNECTING_IP'] ) ) {
+         return false;
+     } else {
+         // Check if original ip has already been restored, e.g. by nginx - assume it was from cloudflare then
+         if ( $_SERVER['REMOTE_ADDR'] === $_SERVER['HTTP_CF_CONNECTING_IP'] ) {
+             return true;
+         }
+     }
+
+     return rocket_is_cf_ip();
+}
+
+/**
+ * Check if a request comes from a CloudFlare IP.
+ *
+ * @since  3.4.1
+ * @author Soponar Cristina
+ *
+ * @return bool
+ */
+function rocket_is_cf_ip() {
+    // Store original remote address in $original_ip
+    $original_ip = $_SERVER['REMOTE_ADDR'];
+    if ( ! isset( $original_ip ) ) {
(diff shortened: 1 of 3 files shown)


inc/common/cloudflare-flexible-ssl.php
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @since  3.4.1
+ * @author Soponar Cristina
+ *
+ * @return bool
+ */
+function rocket_is_cloudflare() {
+     if ( ! isset( $_SERVER['HTTP_CF_CONNECTING_IP'] ) ) {
+         return false;
+     } else {
```
Tabrisrp: You don't need the `else` since you have a return in the condition already.

inc/common/cloudflare-flexible-ssl.php
```diff
@@ -0,0 +1,77 @@
...
+ * @return bool
+ */
+function rocket_is_cf_ip() {
+    // Store original remote address in $original_ip
+    $original_ip = $_SERVER['REMOTE_ADDR'];
+    if ( ! isset( $original_ip ) ) {
+        return false;
+    }
+
+    $cf_ips_values = get_transient('rocket_cloudflare_ips');
```
Tabrisrp: `get_transient` can also return `false` if it doesn't exist. You need to check that, else you will have errors on an undefined object after.

crystinutzaa: Indeed. At this step Cloudflare IPs are required. So if the transient is empty I filled those with Cloudflare IPs. And I removed an outdated IP based on: https://www.cloudflare.com/ips/ and changed also in cloudflare.php to short array. 

Cloudflare IPs code is currently duplicated in both cloudflare-flexible-ssh.php and also in cloudflare.php. I was thinking to move it in ip_in_range.php. What do you think?

Tabrisrp: The code in `cloudflare.php` is dependent on other functions to do CF API requests. Having the duplicate code is fine for now, we'll do an overhaul of all this at some point

crystinutzaa: Ok. But I thinking only to move this block of code: https://www.screencast.com/t/ez4hSBeOB in ip_in_range.php , in a function rocket_get_cloudflare_default_ips() which will return this code and use this function in both cloudflare.php and cloudflare-flexible-ssl.php . Not to have duplicate code.

PR #2039, opened 2019-10-16
Fix #1942 - WP Smush Lazyload compatibility with Unit tests

### LINKED ISSUE #1942: WP Smush lazyload - Images are hidden when our lazyload is enabled
WP Smush has a lazyload functionality:
https://premium.wpmudev.org/blog/smush-image-lazy-load-free/

When that's enabled in combination to WP Rocket's lazyload, images aren't displayed because of these CSS rules which WP Smuch adds:
```CSS
<style>.no-js img.lazyload {
    display: none
}

figure.wp-block-image img.lazyloading {
    min-width: 150px
}

.lazyload,
.lazyloading {
    opacity: 0
}

.lazyloaded {
    opacity: 1;
    transition: opacity 400ms;
    transition-delay: 0ms
}

</style>
```

**Proposed compatibility solution:**

Detect if WP Smush is installed and the lazyload feature is activated, and display a notification and/or prevent WP Rocket's lazyload from being activated.

**Related ticket:**
https://secure.helpscout.net/conversation/948513882/121859/

crystinutzaa:
WP Smush Lazyload compatibility

Detect if WP Smush is installed and the lazyload feature is activated, and display a notification and/or prevent WP Rocket's lazyload from being activated.

Based on 3.4.2 minor version document


inc/classes/admin/settings/class-page.php
```diff
@@ -870,6 +870,14 @@ private function media_section() {
 			]
 		);
 
+		$rocket_maybe_disable_lazyload_plugins = '';
+		if ( rocket_maybe_disable_lazyload() ) {
+			$rocket_maybe_disable_lazyload_plugins = __( 'Autoptimize', 'rocket' ) . ', ';
+		}
+
+		$rocket_maybe_disable_lazyload_plugins = apply_filters( 'rocket_maybe_disable_lazyload_helper', $rocket_maybe_disable_lazyload_plugins );
+		$rocket_maybe_disable_lazyload_plugins = rtrim( $rocket_maybe_disable_lazyload_plugins, ', ' );
```
Screenfeed: Instead of adding `', '` at the end of `$rocket_maybe_disable_lazyload_plugins` and then trimming, I have a proposal:
Make `$rocket_maybe_disable_lazyload_plugins` an array and push plugin names into it. Then use this magic function: [wp_sprintf_l()](https://wpseek.com/function/wp_sprintf_l/).
You get the final string like this:
```php
$rocket_maybe_disable_lazyload_plugins = wp_sprintf_l( '%l', $rocket_maybe_disable_lazyload_plugins );
```

Depending on the number of items, it will output:
- 1: `plugin 1`,
- 2: `plugin 1 and plugin 2`,
- 3+: `plugin 1, plugin 2, and plugin 3`.

It’s a nice way to localize lists without headache, moreover arrays are easier to manipulate :)
What do you think?

inc/classes/admin/settings/class-page.php
```diff
@@ -870,6 +870,14 @@ private function media_section() {
 			]
 		);
 
+		$rocket_maybe_disable_lazyload_plugins = '';
+		if ( rocket_maybe_disable_lazyload() ) {
+			$rocket_maybe_disable_lazyload_plugins = __( 'Autoptimize', 'rocket' ) . ', ';
+		}
+
+		$rocket_maybe_disable_lazyload_plugins = apply_filters( 'rocket_maybe_disable_lazyload_helper', $rocket_maybe_disable_lazyload_plugins );
```
Screenfeed: A DocBlock is missing for this filter.

inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
```diff
@@ -0,0 +1,76 @@
...
+		}
+
+		$lazy_load_option = get_option( WP_SMUSH_PREFIX . 'settings' );
+		$lazyload         = isset( $lazy_load_option['lazy_load'] ) ? $lazy_load_option['lazy_load'] : false;
+
+		if ( ! $lazyload ) {
+			return $rocket_maybe_disable_lazyload_plugins;
+		}
+
+		if ( is_plugin_active( 'wp-smushit/wp-smush.php' ) && ! empty( $lazyload ) ) {
```
Screenfeed: `&& ! empty( $lazyload )` is useless since it cannot be empty (tested line 66).
Also, seeing how this method is used, this `is_plugin_active( 'wp-smushit/wp-smush.php' )` test seems useless: if `WP_SMUSH_VERSION` is defined, then the plugin is active. It could be useful on a page where the plugin could have been disabled earlier (the constant would still be defined, but the plugin deactivated), is it the case here?

inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
```diff
@@ -0,0 +1,76 @@
...
+	 * Disable WP Rocket lazyload fields if WP Smush lazyload is enabled
+	 *
+	 * @since  3.4.2
+	 * @author Soponar Cristina
+	 *
+	 * @return string
+	 */
+	public function is_smush_lazyload_active( $rocket_maybe_disable_lazyload_plugins ) {
+		if ( ! defined( 'WP_SMUSH_VERSION' ) ) {
+			return;
```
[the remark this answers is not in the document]
crystinutzaa: Correct. Removed it

inc/classes/admin/settings/class-page.php
crystinutzaa: Nice magic. It's much more clear now.

inc/classes/admin/settings/class-page.php
crystinutzaa: Added docblock

inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
crystinutzaa: Indeed is useless. Thank you for pointing this out. Removed it.

tests/Unit/ThirdParty/Plugins/Smush.php
```diff
@@ -0,0 +1,86 @@
...
+    public function testShouldDisableWPRocketLazyLoad()
+    {
+        $subscriber  = new Smush_Subscriber();
+
+        Functions\expect( 'get_option' )
+            ->once() // called once
+            ->andReturn( [ 'lazy_load' => true ] );
+
+        Functions\when( '__' )
+            ->justReturn( 'Smush, ' );
```
Tabrisrp: This should not be needed with Brain Monkey, the library takes care of this kind of WP functions automatically

tests/Unit/ThirdParty/Plugins/Smush.php
```diff
@@ -0,0 +1,86 @@
...
+
+        $this->assertNotEmpty( $subscriber->is_smush_lazyload_active( [] ) );
```
Tabrisrp:
```suggestion
        $this->assertContains( 'Smush' $subscriber->is_smush_lazyload_active( [] ) );
```

`assertContains` can be used to test if an array contains a value

inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
```diff
@@ -0,0 +1,68 @@
...
+		];
+	}
+
+	/**
+	 * Disable WP Rocket lazyload when activating WP Smush and values are already in the database.
+	 *
+	 * @since  3.4.2
+	 * @author Soponar Cristina
+	 */
+	public function smush_maybe_deactivate_lazyload() {
```
Tabrisrp:
```suggestion
	public function maybe_deactivate_rocket_lazyload() {
```

I think it's better to convey the goal of the method

tests/Unit/ThirdParty/Plugins/Smush.php
```diff
@@ -0,0 +1,86 @@
...
+     */
+    public function testShouldNotMaybeDeactivateLazyload()
+    {
+        $subscriber  = new Smush_Subscriber();
+
+        Functions\expect( 'get_option' )
+            ->once() // called once
+            ->andReturn();
+
+        $this->assertEmpty( $subscriber->smush_maybe_deactivate_lazyload() );
```
Tabrisrp: the method doesn't return anything, so this assertion is not useful here.

Instead what can be done is to pass Brain Monkey `expect` as tests.

At the start of your class, you need to had the following: `use \Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration;`

And in this method, you can do:
```
        Functions\expect( 'update_rocket_option' )
            ->never();
```

I would also add another test to check for the case when we should disable WPR lazyload, using the same logic, but with `update_rocket_option` called once with an argument of `0`:

```
Functions\expect('update_rocket_option')
  ->once()
  ->with('0');
```

tests/Unit/ThirdParty/Plugins/Smush.php
Screenfeed: I think a comma is missing in your suggestion @Tabrisrp.

tests/Unit/ThirdParty/Plugins/Smush.php
crystinutzaa: If I remove this I receive this error:  `"__" is not defined nor mocked in this test.`
 https://www.screencast.com/t/ciaLlNinxk

tests/Unit/ThirdParty/Plugins/Smush.php
crystinutzaa: Changed to: `$this->assertContains( 'Smush, ', $subscriber->is_smush_lazyload_active( [] ) );`

tests/Unit/ThirdParty/Plugins/Smush.php
Tabrisrp: Ah right, in fact it's Greg who wrote something to handle this. At the start of the method, you can use `$this->mockCommonWpFunctions();` and it will handle it for you 👍

PR #2042, opened 2019-10-16
Fixes #2041 Prevent double rewrite of srcset with relative URL

### LINKED ISSUE #2041: CDN - Incorrect path rewriting for relative paths in a srcset attribute
CDN feature doesn’t recognize relative paths and apply the CNAME twice on an empty domain in a srcset attribute and it results in broken URLs.

`/images/images.webp` is rewritten like this `https://cdn.cname.comhttps//cdn.cname.com/images/images.webp`

Ref: https://wp-media.slack.com/archives/C43T1AYMQ/p1571230717054300

Tabrisrp:
- Prevent double rewrite of srcset sources with relative URL
- Rewrite relative URL in subfolders
- Update unit tests

RegEx test: https://regex101.com/r/uOgjun/13

### DIFF
--- a/tests/Fixtures/CDN/rewrite.html
+++ b/tests/Fixtures/CDN/rewrite.html
@@ -334,13 +334,13 @@ img.emoji {
 	
 	<header id="masthead" class="site-header" role="banner" style="">
 
-		<div class="col-full">		<a class="skip-link screen-reader-text" href="#site-navigation">Aller à la navigation</a>
+		<div class="col-full">      <a class="skip-link screen-reader-text" href="#site-navigation">Aller à la navigation</a>
 		<a class="skip-link screen-reader-text" href="#content">Aller au contenu</a>
 				<div class="site-branding">
-			<div class="beta site-title"><a href="http://example.org/" rel="home">tests</a></div><p class="site-description">Just another WordPress site</p>		</div>
+			<div class="beta site-title"><a href="http://example.org/" rel="home">tests</a></div><p class="site-description">Just another WordPress site</p>        </div>
 					<nav class="secondary-navigation" role="navigation" aria-label="Navigation secondaire">
 							</nav><!-- #site-navigation -->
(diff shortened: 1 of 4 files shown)


tests/Fixtures/CDN/rewrite.html
```diff
@@ -515,19 +515,31 @@ <h1 class="entry-title">Page Image Alignment</h1>		</header><!-- .entry-header -
 <figure class="post-thumbnail">
 	<img width="1568" height="1046" src="http://cdn.example.org/wp-content/uploads/2018/03/sticker-mule-189122-unsplash-1568x1046.jpg" class="attachment-post-thumbnail size-post-thumbnail wp-post-image" alt="" srcset="http://example.org/wp-content/uploads/2018/03/sticker-mule-189122-unsplash-1568x1046.jpg 1568w, http://example.org/wp-content/uploads/2018/03/sticker-mule-189122-unsplash-300x200.jpg 300w, http://example.org/wp-content/uploads/2018/03/sticker-mule-189122-unsplash-768x513.jpg 768w, http://example.org/wp-content/uploads/2018/03/sticker-mule-189122-unsplash-1024x683.jpg 1024w" sizes="(max-width: 1568px) 100vw, 1568px" />
 </figure>
+<picture>
+	<source srcset="http://cdn.example.org/images/site/logo/logo-acommeassure.webp" type="image/webp">
+	<source srcset="/images/site/logo/logo-acommeassure.png 100w" type="image/png">
```
crystinutzaa: Shouldn't this line be:
`<source srcset="http://cdn.example.org/images/site/logo/logo-acommeassure.png 100w" type="image/png">`

Tabrisrp: No, the method will not match this one because of the ` 100w` in the `srcset`. But it matches the first one because there is no width definer.

This `srcset` will be taken care of by `rewrite_srcset()` in the end.

crystinutzaa: Ok. Got it. Thanks for the clarification

PR #2045, opened 2019-10-16
Fixed Cloudflare validation account matches domain

crystinutzaa:
Fixed Cloudflare account matches domain

I input CF credentials from another domain on another of my sites which is not running on Cloudflare. WP Rocket reported no issues and gave a successful cache purge message: https://jmp.sh/t5NMeEP
Official Cloudflare plugin detects the issue:
https://jmp.sh/6eUTkXm


inc/functions/cloudflare.php
```diff
@@ -41,15 +41,41 @@ function rocket_is_api_keys_valid_cloudflare( $cf_email, $cf_api_key, $cf_zone_i
...
 
 		if ( ! isset( $cf_zone->success ) || empty( $cf_zone->success ) ) {
 			return new WP_Error( 'cloudflare_invalid_auth', $cf_zone->errors[0]->message );
 		}
 
 		if ( true === $cf_zone->success ) {
-		    return true;
+			$zone_found = false;
+
+			foreach ( $cf_zone->result as $zone ) {
```
Tabrisrp: Is there a risk of the `result` property being undefined or empty? It might be good to check for that before using it in the loop

inc/functions/cloudflare.php
```diff
@@ -41,15 +41,41 @@ function rocket_is_api_keys_valid_cloudflare( $cf_email, $cf_api_key, $cf_zone_i
...
+				if ( get_site_url() === $zone->name ) {
```
Tabrisrp: I think `get_home_url()` is more appropriate here, the values between `get_home_url()` and `get_site_url()` can be different

inc/functions/cloudflare.php
```diff
@@ -41,15 +41,41 @@ function rocket_is_api_keys_valid_cloudflare( $cf_email, $cf_api_key, $cf_zone_i
...
+					$zone_found = true;
+				}
+			}
+
+			if ( ! $zone_found ) {
+				$msg = sprintf(
+					// translators: %s = WP Rocket plugin name.
+					__( 'It looks that you domain is not provisioned with Cloudflare.', 'rocket' ),
+					WP_ROCKET_PLUGIN_NAME
+				);
```
Tabrisrp: You use `sprintf`, but there is no corresponding token `%s` in the string. It could be:

`__( '%s: It looks like your domain is not provisioned with Cloudflare.', 'rocket' )`

inc/functions/cloudflare.php
```diff
@@ -41,15 +41,41 @@ function rocket_is_api_keys_valid_cloudflare( $cf_email, $cf_api_key, $cf_zone_i
...
+
+				$msg .= ' ' . sprintf(
+					/* translators: %1$s = opening link; %2$s = closing link */
+					__( 'Read the %1$sdocumentation%2$s for further guidance.', 'rocket' ),
+					/* translators: Documentation exists in EN, DE, FR, ES, IT; use loaclised URL if applicable */
```
Tabrisrp:
```suggestion
					// translators: Documentation exists in EN, DE, FR, ES, IT; use localised URL if applicable.
```

inc/functions/cloudflare.php
arunbasillal: Please see what the official Cloudflare plugin uses.

inc/functions/cloudflare.php
arunbasillal: @crystinutzaa There is a minor typo in `you domain` in the original message. (It's fixed in what Remy suggested)

I have asked @webtrainingwheels for feedback on the message. Please take a look at the doc when you make this change - https://docs.google.com/spreadsheets/d/1v0MJi-m7qtGOjqctA946JL1GdZqXNM1svEACxQcsago/edit#gid=0

arunbasillal: @crystinutzaa To make this easier, this is what Lucy suggested: 

```
**WP Rocket**: It looks like your domain is not set up on Cloudflare. Read the documentation for further guidance.
```

inc/functions/cloudflare.php
crystinutzaa: Cloudflare uses this function: https://www.screencast.com/t/3YHnvKTkqBE

inc/functions/cloudflare.php
```diff
@@ -41,15 +41,44 @@ function rocket_is_api_keys_valid_cloudflare( $cf_email, $cf_api_key, $cf_zone_i
...
 		if ( true === $cf_zone->success ) {
-		    return true;
+			$zone_found = false;
+			$site_url   = get_site_url();
+
+			if ( function_exists( 'domain_mapping_siteurl' ) ) {
+				$site_url = domain_mapping_siteurl( $site_url );
+			}
+
+			if ( isset( $cf_zone->result ) && ! empty( $cf_zone->result ) ) {
```
Screenfeed: This is enough:

```php
if ( ! empty( $cf_zone->result ) ) {
```

inc/functions/cloudflare.php
```diff
@@ -41,15 +41,44 @@ function rocket_is_api_keys_valid_cloudflare( $cf_email, $cf_api_key, $cf_zone_i
...
+				foreach ( $cf_zone->result as $zone ) {
+					if ( strtolower( $site_url ) === $zone->name ) {
+						$zone_found = true;
+					}
```
Screenfeed:
```php
$zone_found = true;
break;
```

inc/functions/cloudflare.php
```diff
@@ -41,15 +41,44 @@ function rocket_is_api_keys_valid_cloudflare( $cf_email, $cf_api_key, $cf_zone_i
...
+			$zone_found = false;
+			$site_url   = get_site_url();
+
+			if ( function_exists( 'domain_mapping_siteurl' ) ) {
+				$site_url = domain_mapping_siteurl( $site_url );
+			}
+
+			if ( isset( $cf_zone->result ) && ! empty( $cf_zone->result ) ) {
+				foreach ( $cf_zone->result as $zone ) {
+					if ( strtolower( $site_url ) === $zone->name ) {
```
Screenfeed: This `===` feels too simple to break:
- URL schemes may differ (http vs https), or be absent.
- Trailing slash wants you dead.

A small normalization helper may help:

```php
$url = strtolower( $url );
$url = preg_replace( '@^https?:@', '', $url );
$url = trim( $url, '/' );
```

crystinutzaa: What do you think about using `parse_url` and compare only the host?
`parse_url('http://abc.net/', PHP_URL_HOST);`

`$site_url` - is Wordpress URL
`$zone->name` - is coming as www.abc.com or abc.com

Tabrisrp:
Just some conflicts to fix before being able to merge :)

inc/functions/cloudflare.php
Screenfeed: If the host is enough, yes, no problem.
(sorry didn't see your answer)

PR #2063, opened 2019-10-23
Fixes #1936 - Cache cleaned when post slug / permalink is changed

### LINKED ISSUE #1936: Cache not clearing when post permalink (slug) changes
we have a post with the slug: `this-is-a-random-post`
now if we change the post slug to: `this-is-updated-post`
then the cache files of `this-is-a-random-post` is not cleared. The directory along with html files exists on cache directory.

Expected behaviour: `this-is-a-random-post` cache directory should be cleared because it's now obsolete. 

**Why fix it?**
Because it's breaking wordpress inbuilt feature: 301 redirect from `old-slug` to `new-slug`.
when a user lands on the `old-slug`, wp-rocket shows the old cache files (because it exists),
if the cache file wasn't there, wordpress would have made a 301 redirect to new slug.

crystinutzaa:
Cache cleaned when post slug / permalink is changed - cache cleaned for old post slug / permalink

Based on 3.4.2 Doc

### DIFF
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -1,6 +1,6 @@
 <?php
 defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
-
+use WP_Rocket\Logger\Logger;
 // Launch hooks that deletes all the cache domain.
 add_action( 'switch_theme', 'rocket_clean_domain' );  // When user change theme.
 add_action( 'user_register', 'rocket_clean_domain' );  // When a user is added.
@@ -527,3 +527,19 @@ function rocket_clean_cache_theme_update( $wp_upgrader, $hook_extra ) {
 
 	rocket_clean_domain();
 }
+
+/**
+ * Purge WP Rocket cache on Slug / Permalink change.
+ *
+ * @since  3.4.2
+ * @author Soponar Cristina
+ *
+ * @param int   $post_id   The post ID.
+ * @param array $post_data Array of unslashed post data.
+ */
+function rocket_clean_cache_on_slug_change( $post_id, $post_data) {
+	if ( get_post_field( 'post_name', $post_id ) != $post_data['post_name'] ) {
+        rocket_clean_files( get_the_permalink( $post_id ) );
+    }
+}
+add_action( 'pre_post_update', 'rocket_clean_cache_on_slug_change', 10, 2 );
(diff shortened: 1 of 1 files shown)


inc/common/purge.php
```diff
@@ -527,3 +527,19 @@ function rocket_clean_cache_theme_update( $wp_upgrader, $hook_extra ) {
...
+/**
+ * Purge WP Rocket cache on Slug / Permalink change.
+ *
+ * @since  3.4.2
+ * @author Soponar Cristina
+ *
+ * @param int   $post_id   The post ID.
+ * @param array $post_data Array of unslashed post data.
+ */
+function rocket_clean_cache_on_slug_change( $post_id, $post_data) {
```
Tabrisrp:
```suggestion
function rocket_clean_post_cache_on_slug_change( $post_id, $post_data ) {
```

I think adding `post` in the function name would be better to clearly identify it's clearing only this post cache

inc/common/purge.php
```diff
@@ -527,3 +527,19 @@ function rocket_clean_cache_theme_update( $wp_upgrader, $hook_extra ) {
...
+	if ( get_post_field( 'post_name', $post_id ) != $post_data['post_name'] ) {
```
Tabrisrp:
```suggestion
	if ( get_post_field( 'post_name', $post_id ) !== $post_data['post_name'] ) {
```

Per our coding standards, we always use strict comparison

inc/common/purge.php
```diff
@@ -527,3 +527,19 @@ function rocket_clean_cache_theme_update( $wp_upgrader, $hook_extra ) {
...
+        rocket_clean_files( get_the_permalink( $post_id ) );
+    }
+}
+add_action( 'pre_post_update', 'rocket_clean_cache_on_slug_change', 10, 2 );
```
Tabrisrp:
```suggestion
add_action( 'pre_post_update', 'rocket_clean_post_cache_on_slug_change', 10, 2 );
```

inc/common/purge.php
```diff
@@ -1,6 +1,6 @@
 <?php
 defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
-
+use WP_Rocket\Logger\Logger;
```
Tabrisrp: Is the Logger used in this file currently?

Screenfeed: It doesn’t seem so.

crystinutzaa: Nope. Forgot to remove it.

PR #2066, opened 2019-10-24
Fixes #1912 Add webfont.min.js from defer js

### LINKED ISSUE #1912: JupiterX + Typekit + Deferred JS
Using the JupiterX theme you can integrate your Adobe Fonts/Typekit project. 
Activating this option adds the following code to the site: 

`<script type='text/javascript' src='http://example.com/wp-content/themes/jupiterx/lib/admin/assets/lib/webfont/webfont.min.js?ver=1.6.26'></script>`
`<script type='text/javascript'>
WebFont.load({
typekit: {
id:'xxxxxxx'
}
});
</script>`

When only _Deferred JS_ (ie. without Combine JS) is active, there is an error and we need to exclude this file: 
`/wp-content/themes/jupiterx/lib/admin/assets/lib/webfont/webfont.min.js`

Related ticket: 
https://secure.helpscout.net/conversation/936664111/120329?folderId=377611

Maybe we could do it automatically?

crystinutzaa:
Using JupiterX theme you can integrate your Adobe Fonts / Typekit:
`<script type='text/javascript' src='http://example.com/wp-content/themes/jupiterx/lib/admin/assets/lib/webfont/webfont.min.js?ver=1.6.26'></script>`
`<script type='text/javascript'> WebFont.load({ typekit: { id:'xxxxxxx' } }); </script>`

Remove `/wp-content/themes/jupiterx/lib/admin/assets/lib/webfont/webfont.min.js` from defer JS in order to prevent undefined WebFont.load.

Based on 3.4.2 Doc

### DIFF
--- a/inc/functions/options.php
+++ b/inc/functions/options.php
@@ -412,6 +412,7 @@ function get_rocket_exclude_defer_js() {
 		'www.uplaunch.com',
 		'google.com/recaptcha',
 		'widget.reviews.co.uk',
+		'jupiterx/lib/admin/assets/lib/webfont/webfont.min.js'
 	];
 
 	if ( get_rocket_option( 'defer_all_js', 0 ) && get_rocket_option( 'defer_all_js_safe', 0 ) ) {
--- a/inc/functions/options.php
+++ b/inc/functions/options.php
@@ -412,7 +412,7 @@ function get_rocket_exclude_defer_js() {
 		'www.uplaunch.com',
 		'google.com/recaptcha',
 		'widget.reviews.co.uk',
-		'jupiterx/lib/admin/assets/lib/webfont/webfont.min.js'
+		'lib/admin/assets/lib/webfont/webfont.min.js'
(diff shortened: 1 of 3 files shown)


inc/functions/options.php
```diff
@@ -412,6 +412,7 @@ function get_rocket_exclude_defer_js() {
 		'www.uplaunch.com',
 		'google.com/recaptcha',
 		'widget.reviews.co.uk',
+		'jupiterx/lib/admin/assets/lib/webfont/webfont.min.js'
```
Tabrisrp: The risk here is that someone might have renamed the folder name for the theme, and so the exclusion doesn't work.

The quick solution is to remove the folder name from the path.

crystinutzaa: Basically to keep only: `lib/admin/assets/lib/webfont/webfont.min.js`

And Unit tests are failing . Should I correct those?

PR #2074, opened 2019-10-25
Fix #2073 - PHP Fatal Error if curl is disabled - reverted curl validation and CF validation from get_rocket_cloudflare_api_instance()

### LINKED ISSUE #2073: 3.4.1 - PHP Fatal Error if curl is disabled
After the change in #2045, the check for curl functions was removed from the `get_rocket_cloudflare_api_instance()` function.

This function is used in `rocket_get_cloudflare_ips()`, which is used in `rocket_set_real_ip_cloudflare()`.

If curl is disabled on the server, this will create a PHP fatal error when calling the `Cloudflare\Api()` method to create the Cloudflare instance with the library.

The fix is to add back the check for curl in `get_rocket_cloudflare_api_instance()` to be safe.

This should not be impacting that much of our users, since curl is a pretty common library to have, but it's an issue for those who don't, because it's breaking their site.

crystinutzaa:
Reverted Curl validation ad Cloudflare email & API validation in get_rocket_cloudflare_api_instance() function.

Why in some CF functions get_rocket_cloudflare_api_instance() is used instead of $GLOBALS['rocket_cloudflare'] = get_rocket_cloudflare_instance(); ?

I think that in 3.4.2 we should change all CF functions to use the $GLOBALS['rocket_cloudflare'] which contains all validations if there are no other reasons not to use this.


Tabrisrp:
`get_rocket_cloudflare_api_instance()` is used in in the function fetching CF IPs because fetching IPs doesn't require the zone data, only the API Email/Key

PR #2078, opened 2019-10-28
Fix #2077 Cloudflare domain validation

### LINKED ISSUE #2077: 3.4.1 - False positives for Cloudflare domain validation error messages
We discovered a case where it is possible to have false positives in the validation for the Cloudflare add-on. This prevents clearing of Cloudflare cache. 

The error message will say: 

```
WP Rocket: It looks like your domain is not set up on Cloudflare. Read the documentation for further guidance.
```

Happens because of the mismatch in the values here - https://github.com/wp-media/wp-rocket/blob/1a7cde69e787ee54492d82cd650465c2d653ba7d/inc/functions/cloudflare.php#L75

Related ticket - https://secure.helpscout.net/conversation/989956337/128450/

Related conversations:
- https://wp-media.slack.com/archives/CBJ2KQ0MU/p1572027591126000
- https://wp-media.slack.com/archives/CBJ2KQ0MU/p1572278184142800?thread_ts=1572277458.140500&cid=CBJ2KQ0MU

crystinutzaa:
Fix Cloudflare domain validation with strpos instead of strict comparison

### DIFF
--- a/inc/functions/cloudflare.php
+++ b/inc/functions/cloudflare.php
@@ -72,7 +72,7 @@ function rocket_is_api_keys_valid_cloudflare( $cf_email, $cf_api_key, $cf_zone_i
 
 			if ( ! empty( $cf_zone->result ) ) {
 				$parsed_url = wp_parse_url( $site_url );
-				if ( strtolower( $parsed_url['host'] ) === $cf_zone->result->name ) {
+				if ( strpos( strtolower( $parsed_url['host'] ), $cf_zone->result->name ) >= 0 ) {
 					$zone_found = true;
 				}
 			}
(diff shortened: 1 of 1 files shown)


inc/functions/cloudflare.php
```diff
@@ -72,7 +72,7 @@ function rocket_is_api_keys_valid_cloudflare( $cf_email, $cf_api_key, $cf_zone_i
 
 			if ( ! empty( $cf_zone->result ) ) {
 				$parsed_url = wp_parse_url( $site_url );
-				if ( strtolower( $parsed_url['host'] ) === $cf_zone->result->name ) {
+				if ( strpos( strtolower( $parsed_url['host'] ), $cf_zone->result->name ) >= 0 ) {
```
Tabrisrp:
```suggestion
				if ( false !== strpos( strtolower( $parsed_url['host'] ), $cf_zone->result->name ) ) {
```

We usually do the comparison with `strpos` like this, I'm not sure there is any actual benefit, but for the sake of consistency with the rest of the code let's do it the same everywhere :)

PR #2081, opened 2019-10-29
Cloudflare validation API done only when CF inputs are changed

crystinutzaa:
- The upgrade routine will stop setting cache_ssl as true, if it's already true. This save of settings during upgrade was causing Cloudflare validation during update. This makes everything more efficient.
- Cloudflare validation will only happen if the API key / email / zone ID is changed. So it isn't validated every time other settings are saved. More efficient this way.

### DIFF
--- a/inc/admin/options.php
+++ b/inc/admin/options.php
@@ -88,6 +88,21 @@ function rocket_after_save_options( $oldvalue, $value ) {
 		rocket_clean_cache_busting();
 	}
 
+	if ( ! empty( $_POST ) &&
+			( ( isset( $oldvalue['cloudflare_email'], $value['cloudflare_email'] ) && $oldvalue['cloudflare_email'] !== $value['cloudflare_email'] ) ||
+			( isset( $oldvalue['cloudflare_api_key'], $value['cloudflare_api_key'] ) && $oldvalue['cloudflare_api_key'] !== $value['cloudflare_api_key'] ) ||
+			( isset( $oldvalue['cloudflare_zone_id'], $value['cloudflare_zone_id'] ) && $oldvalue['cloudflare_zone_id'] !== $value['cloudflare_zone_id'] ) )
+			) {
+		// Check Cloudflare input data and display error message.
+		if ( get_rocket_option( 'do_cloudflare' ) && function_exists( 'rocket_is_api_keys_valid_cloudflare' ) ) {
+			$is_api_keys_valid_cloudflare = rocket_is_api_keys_valid_cloudflare( $value['cloudflare_email'], $value['cloudflare_api_key'], $value['cloudflare_zone_id'], false );
+			if ( is_wp_error( $is_api_keys_valid_cloudflare ) ) {
+				$cloudflare_error_message = $is_api_keys_valid_cloudflare->get_error_message();
+				add_settings_error( 'general', 'cloudflare_api_key_invalid', __( 'WP Rocket: ', 'rocket' ) . '</strong>' . $cloudflare_error_message . '<strong>', 'error' );
+			}
+		}
+	}
+
 	// Update CloudFlare Development Mode.
 	$cloudflare_update_result = array();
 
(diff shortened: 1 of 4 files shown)


inc/functions/cloudflare.php
```diff
@@ -51,10 +51,24 @@ function rocket_is_api_keys_valid_cloudflare( $cf_email, $cf_api_key, $cf_zone_i
 
 	try {
 		$cf_api_instance = new Cloudflare\Api( $cf_email, $cf_api_key );
-		$cf_user         = $cf_api_instance->get( 'user/' );
 		$cf_zone         = $cf_api_instance->get( 'zones/' . $cf_zone_id );
 
 		if ( ! isset( $cf_zone->success ) || empty( $cf_zone->success ) ) {
+			foreach ( $cf_zone->errors as $error ) {
+				if ( $error->code === 6003 ) {
```
Tabrisrp: can this property be undefined?

crystinutzaa: Based on CF docs, nope. An error object will contain an integer  `code` field and a `message`

https://api.cloudflare.com/#getting-started-responses

inc/admin/options.php
```diff
@@ -88,6 +88,21 @@ function rocket_after_save_options( $oldvalue, $value ) {
 		rocket_clean_cache_busting();
 	}
 
+	if ( ! empty( $_POST ) &&
```
Screenfeed: Some issues here I think.
- Case 1:
```php
$oldvalue = [ 'cloudflare_email' => 'foo', 'cloudflare_key' => 'bar', 'cloudflare_id' => 'baz' ];
$value    = [ 'cloudflare_email' => '', 'cloudflare_key' => '', 'cloudflare_id' => '' ];
```
If any entry of `$value` is set but empty while those in `$oldvalue` are not, you get in the `if` statement. Then `rocket_is_api_keys_valid_cloudflare()` is fed with empty values. I don't know what will happen but it probably is not a good thing.
I imagine this will happen each time the user empty the fields, it will display an error message from Cloudflare.

- Case 2:
```php
$oldvalue = [ 'cloudflare_email' => 'foo', 'cloudflare_key' => 'bar', 'cloudflare_id' => 'baz' ];
$value    = [ 'cloudflare_email' => 'bar' ];
```
If any entry of `$value` differs from `$oldvalue` while any other is not set, you also get in. Then php notices are triggered.

- This might be useful at some point:
```php
$has_values = ! empty( $value['cloudflare_email'] ) && ! empty( $value['cloudflare_api_key'] ) && ! empty( $value['cloudflare_zone_id'] );

if ( ! empty( $_POST ) && $has_values && (
	empty( $oldvalue['cloudflare_email'] ) || ( isset( $oldvalue['cloudflare_email'] ) && $oldvalue['cloudflare_email'] !== $value['cloudflare_email'] ) ||
	...
```

crystinutzaa: Empty values are handled in `rocket_is_api_keys_valid_cloudflare()` function and will display proper error messages. It will not end up to CF API until all values are filled.

https://github.com/wp-media/wp-rocket/blob/cc93d257fb625a439f76c1e1b31dadd7a201f912/inc/functions/cloudflare.php#L21-L46

Tabrisrp: @Screenfeed For case 2, the values are sanitized in the `sanitize_callback` function before going through this function, making sure they are always set even if empty in the form:

https://github.com/wp-media/wp-rocket/blob/1e915944d0bf9226e49a3f6944a7f88b33c8fc1e/inc/classes/admin/settings/class-settings.php#L359

Screenfeed: OK, thanks for your answers.
So I guess this means `isset( $oldvalue['cloudflare_email'], $value['cloudflare_email'] )` can be reduced to `isset( $oldvalue['cloudflare_email'] )`, right?
But then, the case where it is not set is not handled, which would be:
```php
! isset( $oldvalue['cloudflare_email'] ) && $value['cloudflare_email'] ||
isset( $oldvalue['cloudflare_email'] ) && $oldvalue['cloudflare_email'] !== $value['cloudflare_email'] ||
```
And for better readability:
```php
$oldvalue['cloudflare_email'] = ! empty( $oldvalue['cloudflare_email'] ) ? $oldvalue['cloudflare_email'] : '';
...
if ( ...
	$oldvalue['cloudflare_email'] !== $value['cloudflare_email'] ||
	...
```

Tabrisrp: In theory it's never not set either, but you know how it is 😅 

I like the tertiary operator before the condition for the 3 `oldvalue`, that way we should be safe

PR #2085, opened 2019-10-30
Fixes #306 Allow CDN exclusions to work with Minify/Remove Query Strings

### LINKED ISSUE #306: CDN File Exclusions Not Applied When Remove Query Strings Enabled
When excluding a file from the CDN, the original URL is required (like with minification), but when Remove Query Strings option is enabled, the busted version is served from the CDN. It's not practical to use the busted version for exclusions, since it will change based on the version number. 

Expected behaviour would be that the original URL is provided and is still excluded when Remove Query Strings option is enabled.

Tabrisrp:
Current priorities of functions hooked on `rocket_buffer` prevent the CDN exclusions from working correctly if minify CSS/JS and/or Remove Query Strings options are enabled.

Updating the order allows the exclusions to work, with CDN being applied first, and the above optimizations after.

Other priorities are updated too, to keep everything working in the right order:
- WebP needs to be after HTML minification
- Lazyload before CDN
- Defer JS after minify/combine JS
- Critical CSS insertion after minify/combine CSS and combine Google fonts
- Remove Query Strings after the 2 above
- Async CSS after the 3 above

### DIFF
--- a/inc/classes/subscriber/Media/class-webp-subscriber.php
+++ b/inc/classes/subscriber/Media/class-webp-subscriber.php
@@ -106,17 +106,17 @@ class Webp_Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'rocket_buffer'                                 => [ 'convert_to_webp', 30 ],
-			'rocket_cache_webp_setting_field'               => [
+			'rocket_buffer'                            => [ 'convert_to_webp', 16 ],
+			'rocket_cache_webp_setting_field'          => [
 				[ 'maybe_disable_setting_field' ],
 				[ 'webp_section_description' ],
 			],
-			'rocket_disable_webp_cache'                     => 'maybe_disable_webp_cache',
-			'rocket_third_party_webp_change'                => 'sync_webp_cache_with_third_party_plugins',
-			'rocket_homepage_preload_url_request_args'      => 'add_accept_header',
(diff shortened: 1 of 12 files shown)


inc/classes/subscriber/Media/class-webp-subscriber.php
```diff
@@ -106,17 +106,17 @@ public function __construct( Options_Data $options_data, Options $options_api, C
...
-			'rocket_third_party_webp_change'                => 'sync_webp_cache_with_third_party_plugins',
-			'rocket_homepage_preload_url_request_args'      => 'add_accept_header',
+			'rocket_disable_webp_cache'                => 'maybe_disable_webp_cache',
+			'rocket_third_party_webp_change'           => 'sync_webp_cache_with_third_party_plugins',
+			'rocket_homepage_preload_url_request_args' => 'add_accept_header',
 			'rocket_preload_after_purge_cache_request_args' => 'add_accept_header',
-			'rocket_preload_url_request_args'               => 'add_accept_header',
-			'rocket_partial_preload_url_request_args'       => 'add_accept_header',
+			'rocket_preload_url_request_args'          => 'add_accept_header',
+			'rocket_partial_preload_url_request_args'  => 'add_accept_header',
```
Screenfeed: Alignment issue.

Tabrisrp: Believe it or not, PHP CS is happy with this, and was reporting an error before 🤷‍♂

PR #2089, opened 2019-11-01
Fixed #1985 - Detect and report when </html>, </body> and wp_footer() are not available.

### LINKED ISSUE #1985: Detect and report when closing </html>, </body> and wp_footer() aren't available
Since WP Rocket relies on the `wp_footer` action and the closing `</html>` to work correctly, it would be ideal if we can report it the customer. 

Remy said we could use `did_action` to see if the action `wp_footer` was done. For `</html>` we can check the buffer directly. 

If these are not present and if the features that use them are activated (lazyload uses wp_footer for example), show an admin warning requesting the customer to fix it. 

Related discussion: https://github.com/wp-media/wp-rocket/issues/1883#issuecomment-518240417

crystinutzaa:
Detect and report when closing </html>, </body> and wp_footer() are not available.

I created a new hook in before Optimization process starts, which is hooked in "template_redirect" hook. It seems this hook is called only from template calls and I checked it out with REST API and with FEED and even if caching is enabled for these 2 it works fine (no false positive because this hook is not called)

### DIFF
--- a/inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
+++ b/inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
@@ -55,7 +55,7 @@ class Detect_Missing_Tags_Subscriber implements Subscriber_Interface {
 			return;
 		}
 
-		$transient    = get_transient( 'rocket_missing_tags' );
+		$transient    = get_transient( 'notice_missing_tags' );
 		$transient    = is_array( $transient ) ? $transient : [];
 		$missing_tags = array_unique( array_merge( $transient, $missing_tags ) );
 
@@ -63,7 +63,13 @@ class Detect_Missing_Tags_Subscriber implements Subscriber_Interface {
 			return;
 		}
 
-		set_transient( 'rocket_missing_tags', $missing_tags, HOUR_IN_SECONDS );
+		// Prevent saving the transient if the notice is dismissed.
+		$boxes = get_user_meta( get_current_user_id(), 'rocket_boxes', true );
+		if ( in_array( 'notice_missing_tags', (array) $boxes, true ) ) {
+			return;
+		}
+
+		set_transient( 'notice_missing_tags', $missing_tags, HOUR_IN_SECONDS );
 	}
 
 	/**
@@ -75,16 +81,16 @@ class Detect_Missing_Tags_Subscriber implements Subscriber_Interface {
 	public function notice_missing_tags() {
 		$screen = get_current_screen();
 
-		if ( ! current_user_can( 'rocket_manage_options' ) ) {
+		if ( ! current_user_can( 'rocket_manage_options' ) || ( 'settings_page_wprocket' !== $screen->id ) ) {
 			return;
 		}
 
-		if ( 'settings_page_wprocket' !== $screen->id ) {
+		$boxes = get_user_meta( get_current_user_id(), 'rocket_boxes', true );
+		if ( in_array( __FUNCTION__, (array) $boxes, true ) ) {
 			return;
 		}
 
-		$notice = get_transient( 'rocket_missing_tags' );
-
+		$notice = get_transient( 'notice_missing_tags' );
 		if ( empty( $notice ) || ! is_array( $notice ) ) {
 			return;
 		}
@@ -110,8 +116,10 @@ class Detect_Missing_Tags_Subscriber implements Subscriber_Interface {
 
 		\rocket_notice_html(
 			[
-				'status'  => 'info',
-				'message' => $msg,
+				'status'         => 'info',
+				'dismissible'    => '',
+				'message'        => $msg,
+				'dismiss_button' => __FUNCTION__,
 			]
 		);
 	}
--- a/inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
+++ b/inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
@@ -69,6 +69,10 @@ class Detect_Missing_Tags_Subscriber implements Subscriber_Interface {
 			return;
 		}
 
+		if ( 'settings_page_wprocket' !== $screen->id ) {
+			return;
+		}
+
 		$notice = get_transient( 'rocket_missing_tags' );
 
 		if ( $notice ) {
--- a/inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
+++ b/inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
@@ -17,7 +17,7 @@ class Detect_Missing_Tags_Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'admin_notices'                      => 'notice_missing_tags',
+			'admin_notices'                      => 'rocket_notice_missing_tags',
 			'rocket_before_maybe_process_buffer' => 'maybe_missing_tags',
 		];
 	}
@@ -55,7 +55,7 @@ class Detect_Missing_Tags_Subscriber implements Subscriber_Interface {
 			return;
 		}
 
-		$transient    = get_transient( 'notice_missing_tags' );
+		$transient    = get_transient( 'rocket_notice_missing_tags' );
 		$transient    = is_array( $transient ) ? $transient : [];
 		$missing_tags = array_unique( array_merge( $transient, $missing_tags ) );
 
@@ -65,11 +65,11 @@ class Detect_Missing_Tags_Subscriber implements Subscriber_Interface {
 
 		// Prevent saving the transient if the notice is dismissed.
 		$boxes = get_user_meta( get_current_user_id(), 'rocket_boxes', true );
-		if ( in_array( 'notice_missing_tags', (array) $boxes, true ) ) {
+		if ( in_array( 'rocket_notice_missing_tags', (array) $boxes, true ) ) {
 			return;
 		}
 
-		set_transient( 'notice_missing_tags', $missing_tags, HOUR_IN_SECONDS );
+		set_transient( 'rocket_notice_missing_tags', $missing_tags, HOUR_IN_SECONDS );
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	/**
+	 * This notice is displayed if there is a missing required tag or function: </html>, </body> or wp_footer()
+	 *
+	 * @since  3.4.2
+	 * @author Soponar Cristina
+	 */
+	public function notice_missing_tags() {
+		$screen = get_current_screen();
```
crystinutzaa: This notice should appear only in WP Rocket Settings page?

Tabrisrp: yes

tests/Unit/Subscriber/Tools/TestDetectMissingTags.php
```diff
@@ -0,0 +1,108 @@
...
+			->once()
+			->andReturn( 0 );
+
+		Functions\expect( 'set_transient' )
+			->once()
+			->with('rocket_missing_tags', false, HOUR_IN_SECONDS);
+
+		$missing_tag->maybe_missing_tags( $html );
+
+		$this->assertTrue( true ); // Prevent "risky" warning.
```
Tabrisrp: You should not need that line when using  `MockeryPHPUnitIntegration`

inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
```diff
@@ -0,0 +1,102 @@
...
+			$missing_tags[] = esc_html__( '</body>', 'rocket' );
+			Logger::debug( 'Not found closing </body> tag.', [ 'maybe_missing_tags' ] );
+		}
+
+		if ( did_action( 'wp_footer' ) === 0 ) {
+			$missing_tags[] = __( 'wp_footer()', 'rocket' );
+			Logger::debug( 'Did action did not run wp_footer() function.', [ 'maybe_missing_tags' ] );
+		}
+
+		set_transient( 'rocket_missing_tags', wp_sprintf_l( '%l', $missing_tags ), HOUR_IN_SECONDS );
```
Screenfeed: Some concerns here:
- We should set the transient only if we detected missing tags, otherwise a page where nothing is missing will empty the transient (that may not be empty until now). Ex: visit a page where `</html>` is missing, visit a page where it isn't: no admin notice. Am I right?
- This will run on every page, so it may be a good idea to set the transient only when needed, when we have something new to record.
- What if page 1 is missing `</body>` and page 2 is missing `</html>`? Assuming they are visited in that order, only `</html>` will be shown in the admin notice.

Suggestion:
```suggestion
		if ( ! $missing_tags ) {
			return;
		}

		$transient    = get_transient( 'rocket_missing_tags' );
		$transient    = is_array( $transient ) ? $transient : [];
		$missing_tags = array_unique( array_merge( $transient, $missing_tags ) );

		if ( count( $transient ) === count( $missing_tags ) ) {
			return;
		}

		set_transient( 'rocket_missing_tags', $missing_tags, HOUR_IN_SECONDS );
```

We still have one caveat: the user doesn't know which pages are lacking which tags, but I guess it would be awful to display a notice listing hundreds of URLs ^^' (no shit Sherlock!).

inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
```diff
@@ -0,0 +1,102 @@
...
+			return;
+		}
+
+		if ( 'settings_page_wprocket' !== $screen->id ) {
+			return;
+		}
+
+		$notice = get_transient( 'rocket_missing_tags' );
+
+		if ( empty( $notice ) ) {
```
Screenfeed: Following my previous comment:
```suggestion
		if ( empty( $notice ) || ! is_array( $notice ) ) {
```

inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
```diff
@@ -0,0 +1,102 @@
...
+			return;
+		}
+
+		$msg  = '<b>' . __( 'WP Rocket: ', 'rocket' ) . '</b>';
+		$msg .= sprintf(
+			/* translators: %1$s = missing tags; */
+			esc_html__( 'Failed to detect the following requirement(s) in your theme: closing %1$s.', 'rocket' ),
+			// translators: Documentation exists in EN, FR.
+			$notice
```
Screenfeed: Following my previous comment:
```suggestion
			wp_sprintf_l( '%l', $notice )
```

inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
```diff
@@ -0,0 +1,102 @@
...
+		$notice = get_transient( 'rocket_missing_tags' );
+
+		if ( empty( $notice ) ) {
+			return;
+		}
+
+		$msg  = '<b>' . __( 'WP Rocket: ', 'rocket' ) . '</b>';
+		$msg .= sprintf(
+			/* translators: %1$s = missing tags; */
+			esc_html__( 'Failed to detect the following requirement(s) in your theme: closing %1$s.', 'rocket' ),
```
Screenfeed: Following my previous comment, this can be improved that way:
```suggestion
			esc_html( _n( 'Failed to detect the following requirement in your theme: closing %1$s.', 'Failed to detect the following requirements in your theme: closing %1$s.', count( $notice ), 'rocket' ) ),
```
I'm not sure of the last part, `closing %1$s`, it feels a bit weird to me 🤔

inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
```diff
@@ -0,0 +1,102 @@
...
+
+		if ( 'settings_page_wprocket' !== $screen->id ) {
+			return;
+		}
+
+		$notice = get_transient( 'rocket_missing_tags' );
+
+		if ( empty( $notice ) ) {
+			return;
+		}
```
Screenfeed: Now that we have an array, what about wrapping the values with `<code>` tags? It also allows to escape the values, they were not until now.
```suggestion
		}

		foreach ( $notice as $i => $tag ) {
			$notice[ $i ] = '<code>' . esc_html( $tag ) . '</code>';
		}
```

inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
crystinutzaa: You are correct, it will be displayed only for a certain page. So if a type of page misses </body> and one misses </html> only the last visited page will display the error. 
Indeed, displaying error messages for all pages is not an option, because there will be a lot of URLs in there. 

Suggestion: Probably include the separate log files and create a log file where all pages which have missing tags are logged and display the URL to this log file in the notice?

Screenfeed: I first thought about storing the URLs in the transient, but it may use a lot of space in the DB.
Using a dedicated log file, why not, it’s probably the best solution so far. I think users must be able to deactivate it via filter, because of free disk space + read/write access time and resources (it's triggered on each page).

arunbasillal: Why do we need dedicated log file when we are already logging this into debug log? 

In 99% of the cases, most websites miss these because their footer.php doesn't have it. One page having this and the others not having it is really rare. But possible.

Tabrisrp: I'm wondering if this improvement should be discussed some more before going further, to be sure we provide at the same time:
- The most helpful data possible
- The less obstrusively & obnoxiously possible
- Ways for the users to solve the issue themselves

Let's remember that we want this notice to be meaningful for the users, and give them the information for them to fix the issue directly. Adding stuff to the debug log doesn't help prevent a support ticket at the moment.

arunbasillal: You are right. I am confused at the direction this is headed to. Let me open a slack thread to see if we are all on the same page.

crystinutzaa: Yes, this task has gone wild :)

Basically, I was thinking to provide the user the dedicated log file with all missing URLs and missing hooks in the notice. I was thinking of having the separate log file and when this issue appears, ask the client to enable debug logging and in the notice to provide them the URL to the dedicated debug file where they can see the exact missing tag from the exact URL.

Probably in this way the client will have all necessary information on how to solve by themselves the issue.

inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
```diff
@@ -55,15 +55,21 @@ public function maybe_missing_tags( $html ) {
 			return;
 		}
 
-		$transient    = get_transient( 'rocket_missing_tags' );
+		$transient    = get_transient( 'notice_missing_tags' );
```
Tabrisrp: the transient name should be prefixed with `rocket_`

Screenfeed:
The transient is still never deleted :/

tests/Unit/Subscriber/Tools/TestDetectMissingTags.php
```diff
@@ -0,0 +1,124 @@
...
+			->once()
+			->andReturn( [] );
+
+		Functions\expect( 'get_user_meta' )
+			->once()
+			->andReturn( [] );
+
+		Functions\expect( 'get_current_user_id' )
+			->once()
+			->andReturn( 1 );
```
Tabrisrp: You could use `Functions\when` instead of `expect` here, because we are not that interested in the calls and values returned for those functions in the test.

What we really want to validate is if `did_action` was called, and if the transient is correctly set, which we do with the above and below calls to `expect`

tests/Unit/Subscriber/Tools/TestDetectMissingTags.php
```diff
@@ -0,0 +1,124 @@
...
```
Tabrisrp: Same here, you can use `when` instead to patch these functions

crystinutzaa:
@Screenfeed the transient is deleted when the user clicks on `Dismiss this notice.` link.

This is triggered in the function: `rocket_dismiss_boxes()` : $args['box'] is the transient name and function: `rocket_notice_missing_tags`
https://github.com/wp-media/wp-rocket/blob/281eb493f44e873a9685f379cec701e8ebcf3532/inc/admin/admin.php#L129-L156

Screenfeed:
:o

PR #2097, opened 2019-11-05
Remove "Cheatin' uh?" error message

ranss:
Minor changes to the code.

### DIFF
--- a/wp-rocket.php
+++ b/wp-rocket.php
@@ -15,7 +15,7 @@
  * Copyright 2013-2019 WP Rocket
  * */
 
-defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
+defined( 'ABSPATH' ) || exit;
 
 // Rocket defines.
 define( 'WP_ROCKET_VERSION',               '3.4.1.2' );
--- a/wp-rocket.php
+++ b/wp-rocket.php
@@ -15,7 +15,7 @@
  * Copyright 2013-2019 WP Rocket
  * */
 
-defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
+defined( 'ABSPATH' ) || exit;
 
 // Rocket defines.
 define( 'WP_ROCKET_VERSION',               '3.4.1.2' );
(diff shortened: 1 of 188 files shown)


arunbasillal:
@ranss I am curious. What was wrong with `Cheatin' uh?`?

Tabrisrp:
Some Related WordPress ticket and changes in 4.9.5:
- https://core.trac.wordpress.org/ticket/38332
- https://core.trac.wordpress.org/changeset/42811

Pretty much any reference to this `Cheatin'` message has been removed from WP core now, and doesn't provide anything useful, so just exiting in this case is valid.

It doesn't have an impact outside of this, so I agree with the change.

wp-rocket.php
```diff
@@ -29,7 +29,7 @@
 define( 'WP_ROCKET_WEB_VALID',             WP_ROCKET_WEB_MAIN . 'valid_key.php' );
 define( 'WP_ROCKET_WEB_INFO',              WP_ROCKET_WEB_MAIN . 'plugin_information.php' );
 define( 'WP_ROCKET_FILE',                  __FILE__ );
-define( 'WP_ROCKET_PATH',                  realpath( plugin_dir_path( WP_ROCKET_FILE ) ) . '/' );
+define( 'WP_ROCKET_PATH',                  realpath( dirname( WP_ROCKET_FILE ) ) . '/' );
```
Tabrisrp: What is the idea behind changing this line?

arunbasillal: Probably because `plugin_dir_path` already comes with a trailing slash?

hellofromtonya: @arunbasillal True. It does:

```php
function plugin_dir_path( $file ) {
	return trailingslashit( dirname( $file ) );
}
```

hellofromtonya: I'd prefer this type of change though to be in a separate PR, as it's not part of replacing `cheating, uh` work.

What do you think @Tabrisrp?

hellofromtonya: @ranss I reverted commit 5c43c3dda5d3eec5586196d8a8a9f62a7368f2cd. Why? The code changes are not related to `cheatin' uh` replacement. Instead, those types of changes should be submitted as a separate PR.

ranss:
> 
> 
> @ranss Thank you for contributing to WP Rocket! We appreciate your contribution +1

You are welcome @hellofromtonya 
Glad to constribute :)

hellofromtonya:
@Tabrisrp A few notes for you when reviewing this PR:

- I reverted (removed) the code changes not related to swapping out the "cheatin' uh" code.
- I removed all "added" `defined( 'ABSPATH' ) || exit;` that are not currently in the codebase. Why? To minimize any unknown impacts.

PR #2102, opened 2019-11-11
Fixed false positive for missing tags when there is a redirect

crystinutzaa:
Fixed false positive reports for missing </html>, </body> or wp_footer() when there is a redirect and the content of the page is empty or less than 255 chars.

### DIFF
--- a/inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
+++ b/inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
@@ -35,24 +35,48 @@ class Detect_Missing_Tags_Subscriber implements Subscriber_Interface {
 		if ( strlen( $html ) <= 255 ) {
 			return;
 		}
-		Logger::info( 'START Detect Missing closing tags ( <html>, </body> or wp_footer() )', [ 'URI' => $this->get_raw_request_uri() ] );
+		Logger::info(
+			'START Detect Missing closing tags ( <html>, </body> or wp_footer() )',
+			[
+				'maybe_missing_tags',
+				'URI' => $this->get_raw_request_uri(),
+			]
+		);
 
 		// Remove all comments before testing tags. If </html> or </body> tags are commented this will identify it as a missing tag.
 		$html         = preg_replace( '/<!--([\\s\\S]*?)-->/', '', $html );
 		$missing_tags = [];
 		if ( false === strpos( $html, '</html>' ) ) {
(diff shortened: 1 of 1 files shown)


inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
```diff
@@ -31,24 +31,28 @@ public static function get_subscribed_events() {
 	 * @param string $html HTML content.
 	 */
 	public function maybe_missing_tags( $html ) {
-		Logger::info( 'START Detect_Missing_Tags_Subscriber - maybe_missing_tags ', [ 'maybe_missing_tags' ] );
+		// If there is a redirect the content is empty and can display a false positive notice.
+		if ( strlen( $html ) <= 255 ) {
+			return;
+		}
+		Logger::info( 'START Detect Missing closing tags ( <html>, </body> or wp_footer() )', [ 'URI' => $this->get_raw_request_uri() ] );
```
Screenfeed: We're not passing the `'maybe_missing_tags'` context anymore?

inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
```diff
@@ -123,4 +127,24 @@ public function rocket_notice_missing_tags() {
...
+	public function get_raw_request_uri() {
+		if ( ! isset( $_SERVER['REQUEST_URI'] ) ) {
+			return '';
+		}
+
+		if ( '' === $_SERVER['REQUEST_URI'] ) {
+			return '';
+		}
+
+		return '/' . ltrim( $_SERVER['REQUEST_URI'], '/' );
```
Screenfeed: I found nothing about escaping within logs, so I guess we should do it here. Also, `wp_unslash()`.
```suggestion
		return '/' . esc_html( ltrim( wp_unslash( $_SERVER['REQUEST_URI'] ), '/' ) );
```

inc/classes/subscriber/Tools/class-detect-missing-tags-subscriber.php
crystinutzaa: Nope, I taught in this case it's much more relevant to display the exact URL where there are missing tags instead of a plain context. 
What do you think?

Screenfeed: Why "instead"? We can have both. The context is useful when searching the log.

PR #2104, opened 2019-11-12
Fix #2103 Array to string conversion in mobile subscriber

### LINKED ISSUE #2103: 3.4.2-alpha1 - PHP Notice: Array to string conversion in /inc/classes/subscriber/third-party/plugins/class-mobile-subscriber.php on line 224
When `3.4.2-alpha1` is installed and I rollback to 3.3.7, I get the following notice:

```
Notice: Array to string conversion in /home/wpquicki/mega/wp-content/plugins/wp-rocket/inc/classes/subscriber/third-party/plugins/class-mobile-subscriber.php on line 224
```

You can see this issue here: http://mega.wp-rocket.me/wp-admin/options-general.php?page=wprocket#tools

crystinutzaa:
Notice: Array to string conversion in /home/wpquicki/mega/wp-content/plugins/wp-rocket/inc/classes/subscriber/third-party/plugins/class-mobile-subscriber.php on line 224

When 3.4.2-alpha1 is installed and I rollback to 3.3.7, I get the following notice:

### DIFF
--- a/inc/classes/subscriber/third-party/plugins/class-mobile-subscriber.php
+++ b/inc/classes/subscriber/third-party/plugins/class-mobile-subscriber.php
@@ -221,6 +221,7 @@ class Mobile_Subscriber implements Subscriber_Interface {
 	 */
 	public function maybe_update_mobile_cache_activation( $value, $old_value ) {
 		$plugins   = static::get_mobile_plugins();
+		$plugins   = array_keys( $plugins );
 		$value     = array_intersect( $plugins, (array) $value );
 		$old_value = array_intersect( $plugins, (array) $old_value );
 


inc/classes/subscriber/third-party/plugins/class-mobile-subscriber.php
```diff
@@ -221,6 +221,7 @@ public function delete_site_option_callback( $option, $network_id ) {
 	 */
 	public function maybe_update_mobile_cache_activation( $value, $old_value ) {
 		$plugins   = static::get_mobile_plugins();
+		$plugins   = array_keys( $plugins );
```
Screenfeed: Nice try, but no ^^'
https://3v4l.org/1t2FA

I think this is what happens here:
https://3v4l.org/ieCod

crystinutzaa: The array you passed is not the correct one. Basically this one will be used:
https://3v4l.org/Y0IFQ

Screenfeed: Right, I was looking at the old version. Apologies.

PR #2108, opened 2019-11-15
Rewrite Cloudflare API

crystinutzaa:
Rewrite CloudFlare code

Needs Unit Tests to be added - work in progress


inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+	 * @since 2.8.21 Get the zone ID from the options
+	 * @since 2.8.18 Add try/catch to prevent fatal error Uncaugh Exception
+	 * @since 2.8.16 Update to Cloudflare API v4
+	 * @since 2.5
+	 *
+	 * @return Object Cloudflare instance & zone_id if credentials are correct, WP_Error otherwise
+	 */
+	public static function get_instance() {
+		$cf_email             = get_rocket_option( 'cloudflare_email', null );
+		$cf_api_key           = ( defined( 'WP_ROCKET_CF_API_KEY' ) ) ? WP_ROCKET_CF_API_KEY : get_rocket_option( 'cloudflare_api_key', null );
```
Screenfeed: ಠ_ಠ

```suggestion
		$cf_api_key           = defined( 'WP_ROCKET_CF_API_KEY' ) ? WP_ROCKET_CF_API_KEY : get_rocket_option( 'cloudflare_api_key', null );
```

Also few other cases later.

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+	 * Get a Cloudflare\Api instance & the zone_id corresponding to the domain
+	 *
+	 * @since 2.8.21 Get the zone ID from the options
+	 * @since 2.8.18 Add try/catch to prevent fatal error Uncaugh Exception
+	 * @since 2.8.16 Update to Cloudflare API v4
+	 * @since 2.5
+	 *
+	 * @return Object Cloudflare instance & zone_id if credentials are correct, WP_Error otherwise
+	 */
+	public static function get_instance() {
```
Screenfeed: `get_instance()` is generally used to get the class instance, so it's a bit confusing here. What do you think of renaming to:

```suggestion
	public static function get_cloudflare_instance() {
```

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+	/**
+	 * Get all the current Cloudflare settings for a given domain.
+	 *
+	 * @since 2.8.16 Update to Cloudflare API v4
+	 * @since 2.5
+	 *
+	 * @return mixed bool|Array Array of Cloudflare settings, false if any error connection to Cloudflare
+	 */
+	public static function get_settings() {
+		$cf_instance = self::get_instance();
```
Screenfeed: Then, and for all the others:

```suggestion
		$cf_instance = self::get_cloudflare_instance();
```

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+	 * @return mixed Object|String Mode value if the update is successful, WP_Error otherwise
+	 */
+	public static function set_devmode( $mode ) {
+		$cf_instance = self::get_instance();
+
+		if ( is_wp_error( $cf_instance ) ) {
+			return $cf_instance;
+		}
+
+		if ( (int) 0 === $mode ) {
```
Screenfeed:
```suggestion
		if ( 0 === (int) $mode ) {
```

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+			$value = 'off';
+		} elseif ( (int) 1 === $mode ) {
```
Screenfeed:
```suggestion
		} else {
```

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+	 * @since 2.8.21 Save IPs in a transient to prevent calling the API everytime
+	 * @since 2.8.16
+	 *
+	 * @author Remy Perona
+	 *
+	 * @return Object Result of API request if successful, default CF IPs otherwise
+	 */
+	public function get_cloudflare_ips() {
+		$cf_ips = get_transient( 'rocket_cloudflare_ips' );
+		if ( false === $cf_ips ) {
```
Screenfeed: A pyramid level that can be flatten :)

```php
if ( false !== $cf_ips ) {
	return $cf_ips;
}
```

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+			$cf_email        = get_rocket_option( 'cloudflare_email', null );
+			$cf_api_key      = ( defined( 'WP_ROCKET_CF_API_KEY' ) ) ? WP_ROCKET_CF_API_KEY : get_rocket_option( 'cloudflare_api_key', null );
+			$cf_api_instance = new \Cloudflare\Api( $cf_email, $cf_api_key );
+			$cf_api_instance->setCurlOption( CURLOPT_USERAGENT, self::CF_USER_AGENT );
+
+			try {
+				$cf_ips_instance = new \Cloudflare\IPs( $cf_api_instance );
+				$cf_ips          = $cf_ips_instance->ips();
+
+				if ( ! isset( $cf_ips->success ) || ! $cf_ips->success ) {
```
Screenfeed:
```suggestion
				if ( empty( $cf_ips->success ) ) {
```

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
+<?php
+namespace WP_Rocket\Subscriber\Tools;
```
Tabrisrp: The subscriber should be located in `WP_Rocket\Subscriber\Addons\Cloudflare`

Also note that on branch-3.5, there is already a Cloudflare subscriber created, so we don't want to duplicate code and functionalities

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+	 */
+	public function __construct( Options $options) {
+		$this->options = $options;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public static function get_subscribed_events() {
+		if ( ! get_rocket_option( 'do_cloudflare' ) ) {
```
Tabrisrp: I believe it would be better to do this check via a private method like `is_enabled()` called in each methods hooked instead of the method used to subscribed the events, to improve the separation of concerns, and also remove coupling with procedural code

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+	}
+
+	/**
+	 * Automatically set Cloudflare development mode value to off after 3 hours to reflect Cloudflare behaviour
+	 *
+	 * @since 2.9
+	 * @author Remy Perona
+	 */
+	public function deactivate_devmode() {
+		$options                       = $this->options->get( WP_ROCKET_SLUG );
```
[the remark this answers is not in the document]
Tabrisrp: A better approach would be to pass the `WP_Rocket\Admin\Options` instance as a dependency for the class, and then use it here like this:

```
$this->options->set( 'cloudflare_devmode', 'off' );
$this->options_api->set( 'settings', $this->options->get_options();
```
It's simpler and abstract the interaction with the WordPress options API away from the class.

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+		set_transient( get_current_user_id() . '_cloudflare_purge_result', $cf_purge_result );
+	}
+
+	/**
+	 * Purge CloudFlare cache
+	 *
+	 * @since 2.5
+	 */
+	public function do_purge_cloudflare() {
+		if ( ! isset( $_GET['_wpnonce'] ) || ! wp_verify_nonce( $_GET['_wpnonce'], 'rocket_purge_cloudflare' ) ) {
```
Tabrisrp:
```suggestion
		if ( ! isset( $_GET['_wpnonce'] ) || ! wp_verify_nonce( sanitize_key( $_GET['_wpnonce'] ), 'rocket_purge_cloudflare' ) ) {
```

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+			$cf_purge_result = [
+				'result'  => 'success',
+				'message' => __( '<strong>WP Rocket:</strong> Cloudflare cache successfully purged.', 'rocket' ),
+			];
+		}
+
+		set_transient( get_current_user_id() . '_cloudflare_purge_result', $cf_purge_result );
+
+		wp_safe_redirect( esc_url_raw( wp_get_referer() ) );
+		die();
```
Tabrisrp:
```suggestion
		wp_die();
```

That makes the method testable, as we can mock `wp_die()`, but we can't do it for `die()`

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+	 * @since 2.8.16 Uses CloudFlare API v4 to get CloudFlare IPs
+	 * @since 2.5.4
+	 * @source cloudflare.php - https://wordpress.org/plugins/cloudflare/
+	 */
+	public function set_real_ip() {
+		$is_cf = ( isset( $_SERVER['HTTP_CF_CONNECTING_IP'] ) ) ? true : false;
+
+		if ( ! $is_cf ) {
+			return;
+		}
```
Tabrisrp: the `$is_cf` variable doesn't seem to be used anytime after in the method, so this can be simplified:
```
if ( ! isset( $_SERVER['HTTP_CF_CONNECTING_IP'] ) ) {
			return;
		}
```

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+						break;
+					}
+				}
+			}
+		}
+
+		// Let people know that the CF WP plugin is turned on.
+		if ( ! headers_sent() ) {
+			header( 'X-CF-Powered-By: WP Rocket ' . WP_ROCKET_VERSION );
+		}
```
Tabrisrp: I'm questioning if we should keep that part, does it bring any value for anyone? Not sure the support team uses it @wp-media/supportrocket

inc/classes/subscriber/Tools/class-cloudflare-subscriber.php
```diff
@@ -0,0 +1,888 @@
...
+	}
+
+	/**
+	 * This notice is displayed after purging the CloudFlare cache
+	 *
+	 * @since 2.9
+	 * @author Remy Perona
+	 */
+	public function maybe_print_notice() {
```
Tabrisrp:
```suggestion
	public function maybe_display_purge_notice() {
```

crystinutzaa:
Merged branch-3.5 in the current branch
Updated the code based on Remy
- `WP_Rocket\Addons\Cloudflare\Cloudflare`
-` WP_Rocket\Subscriber\Addons\Cloudflare\CloudflareSubscriber`
Changed Base branch to branch-3.5

* Needs Unit Tests for both new classes ( & fix current failing unit tests)

crystinutzaa:
@Screenfeed those are PRs from branch-3.5 which I merged in this current branch and changed the PR base branch to 3.5.
The files which I changed are:
 inc/admin/options.php
inc/admin/ui/notices.php 
inc/classes/Addons/Cloudflare/Cloudflare.php 
inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php 
inc/classes/ServiceProvider/class-addons-subscribers.php 
 inc/deprecated/deprecated.php 
 inc/deprecated/3.4.php 
inc/main.php

Tabrisrp:
To clear up the list of files changed on the PR, you can also merge the `develop` branch into the `branch-3.5`. It will give a better view of the changes to check

inc/classes/Addons/Cloudflare/Cloudflare.php
```diff
@@ -0,0 +1,611 @@
...
+		if ( ! function_exists( 'curl_init' ) || ! function_exists( 'curl_exec' ) ) {
+			return new \WP_Error( 'curl_disabled', __( 'Curl is disabled on your server. Please ask your host to enable it. This is required for the Cloudflare Add-on to work correctly.', 'rocket' ) );
+		}
+
+		if ( empty( $cf_email ) || empty( $cf_api_key ) ) {
+			return new \WP_Error(
+				'cloudflare_credentials_empty',
+				sprintf(
+					/* translators: %1$s = opening link; %2$s = closing link */
+					__( 'Cloudflare email, API key and Zone ID are not set. Read the %1$sdocumentation%2$s for further guidance.', 'rocket' ),
```
Tabrisrp:
```suggestion
					__( 'Cloudflare email and/or API key are not set. Read the %1$sdocumentation%2$s for further guidance.', 'rocket' ),
```

inc/classes/Addons/Cloudflare/Cloudflare.php
```diff
@@ -0,0 +1,611 @@
+<?php
+namespace WP_Rocket\Addons\Cloudflare;
+
+use WP_Rocket\Admin\Options_Data;
+
```
Tabrisrp:
```
use Cloudflare\Api;
use Cloudflare\Zone\Cache;
use Cloudflare\Zone\PageRules;
use Cloudflare\Zone\Settings;
use Cloudflare\IPs;
```
Importing the namespaces here makes the code easier to read and update when we need to create the instances of the classes later on

inc/classes/Addons/Cloudflare/Cloudflare.php
```diff
@@ -0,0 +1,611 @@
...
+			$is_api_keys_valid_cf = $this->is_api_keys_valid( $cf_email, $cf_api_key, $cf_zone_id );
+			set_transient( 'rocket_cloudflare_is_api_keys_valid', $is_api_keys_valid_cf, 2 * WEEK_IN_SECONDS );
+		}
+
+		if ( is_wp_error( $is_api_keys_valid_cf ) ) {
+			return $is_api_keys_valid_cf;
+		}
+
+		$cf_api_instance = new \Cloudflare\Api( $cf_email, $cf_api_key );
+		$cf_api_instance->setCurlOption( CURLOPT_USERAGENT, self::CF_USER_AGENT );
```
Tabrisrp: Smart idea to use that to pass the custom user agent, did you check that it's working correctly with the library?

inc/classes/Addons/Cloudflare/Cloudflare.php
```diff
@@ -0,0 +1,611 @@
...
+	const CF_USER_AGENT = 'wp-rocket/' . WP_ROCKET_VERSION;
+
+	/**
+	 * Constructor
+	 *
+	 * @param Options_Data $options WP Rocket options instance.
+	 */
+	public function __construct( Options_Data $options ) {
+		$this->options        = $options;
+		$this->cloudflare_api = $this->get_cloudflare_instance();
```
Tabrisrp: If  you do that at the creation of the instance, you have no check to make sure the CF addon is enabled. That would send API requests and display the error for no reason.

inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
```diff
@@ -90,4 +123,418 @@ private function should_filter_varnish() {
...
+			$cf_purge_result = [
+				'result'  => 'success',
+				'message' => __( '<strong>WP Rocket:</strong> Cloudflare cache successfully purged.', 'rocket' ),
+			];
+		}
+
+		set_transient( get_current_user_id() . '_cloudflare_purge_result', $cf_purge_result );
+
+		wp_safe_redirect( esc_url_raw( wp_get_referer() ) );
+		die();
```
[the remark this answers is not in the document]
crystinutzaa: `wp_die();` is not going to work with `wp_safe_redirect()`. Instead of the redirect an empty error page will be displayed and this is not the desired result. 
eg. with `wp_die();` - https://www.screencast.com/t/lFFacV6QU6w
eg. without any die or wp_die it works fine: so I think I can remove totally the die from this part

inc/classes/Addons/Cloudflare/Cloudflare.php
crystinutzaa: Yes, tested and is working just fine.

inc/classes/Addons/Cloudflare/Cloudflare.php
crystinutzaa: It shouldn't display any errors because everything is behind the do_cloudflare check ( https://www.screencast.com/t/BEwiVPEz1 ) . The validation API credentials is made on the constructor and saved in a transient. The transient is deleted / updated in 2 situations and expires in 2 weeks:
1. When Cloudflare addon is disabled / enabled it will delete the transient 
2. When the CF credentials are changed it will revalidate the credentials.
Not sure if I'm missing something in here :)

Tabrisrp: The check in the constructor will be good yes

inc/classes/Addons/Cloudflare/Cloudflare.php
Tabrisrp: great that will please Cloudflare

inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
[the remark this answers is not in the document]
Screenfeed: For security reasons, `die();` must not be removed.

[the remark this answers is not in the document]
Tabrisrp: Indeed it's important to keep it. In that case, the alternative is to keep `die();`, and move the  rest of the method's code to a different method, which will be testable.

inc/classes/Addons/Cloudflare/CloudflareFacade.php
```diff
@@ -0,0 +1,260 @@
...
+	 * @since 3.5
+	 * @author Soponar Cristina
+	 *
+	 * @param Cloudflare\Api            $api        - Cloudflare API Instance.
+	 * @param Cloudflare\Zone\Cache     $cache      - Cloudflare Cache Instance.
+	 * @param Cloudflare\Zone\PageRules $page_rules - Cloudflare Page Rules Instance.
+	 * @param Cloudflare\Zone\Settings  $settings   - Cloudflare Settings Instance.
+	 * @param Cloudflare\IPs            $ips        - Cloudflare IPS Instance.
+	 */
+	public function __construct( $api, $cache, $page_rules, $settings, $ips ) {
```
Screenfeed: That’s a lot of arguments. I wonder if we should use only one array `$args`. Same goes for `set_api_credentials()`.
What do you think?

inc/classes/ServiceProvider/class-addons-subscribers.php
```diff
@@ -55,7 +61,18 @@ public function register() {
 		$this->getContainer()->share( 'varnish_subscriber', 'WP_Rocket\Subscriber\Addons\Varnish\VarnishSubscriber' )
 			->withArgument( $this->getContainer()->get( 'varnish' ) )
 			->withArgument( $this->getContainer()->get( 'options' ) );
+		$this->getContainer()->add( 'cloudflare_facade', 'WP_Rocket\Addons\Cloudflare\CloudflareFacade' )
```
Screenfeed: Any reason not to `->share()` `'cloudflare_facade'` and `'cloudflare'`?

inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
```diff
@@ -90,4 +123,412 @@ private function should_filter_varnish() {
...
+	/**
+	 * Set Real IP from CloudFlare
+	 *
+	 * @since 2.8.16 Uses CloudFlare API v4 to get CloudFlare IPs
+	 * @since 2.5.4
+	 * @source cloudflare.php - https://wordpress.org/plugins/cloudflare/
+	 */
+	public function set_real_ip() {
+		// only run this logic if the REMOTE_ADDR is populated, to avoid causing notices in CLI mode.
+		if ( ! isset( $_SERVER['HTTP_CF_CONNECTING_IP'] ) || ! isset( $_SERVER['REMOTE_ADDR'] ) ) {
```
Screenfeed: Optional, can be simplified:
```suggestion
		if ( ! isset( $_SERVER['HTTP_CF_CONNECTING_IP'], $_SERVER['REMOTE_ADDR'] ) ) {
```

inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
```diff
@@ -90,4 +123,412 @@ private function should_filter_varnish() {
...
+				];
+			}
+		}
+
+		// Update CloudFlare settings.
+		if ( ! empty( $value['do_cloudflare'] ) && isset( $old_value['cloudflare_auto_settings'], $value['cloudflare_auto_settings'] ) && (int) $old_value['cloudflare_auto_settings'] !== (int) $value['cloudflare_auto_settings'] ) {
+			$cf_old_settings = explode( ',', $value['cloudflare_old_settings'] );
+
+			// Set Cache Level to Aggressive.
+			$cf_cache_level        = ( isset( $cf_old_settings[0] ) && 0 === $value['cloudflare_auto_settings'] ) ? 'basic' : 'aggressive';
```
Screenfeed:
```suggestion
			$cf_cache_level        = isset( $cf_old_settings[0] ) && 0 === $value['cloudflare_auto_settings'] ? 'basic' : 'aggressive';
```

inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
```diff
@@ -90,4 +123,412 @@ private function should_filter_varnish() {
...
+
+				$cloudflare_update_result[] = [
+					'result'  => 'success',
+					// translators: %s is the caching level returned by the CloudFlare API.
+					'message' => '<strong>' . __( 'WP Rocket: ', 'rocket' ) . '</strong>' . sprintf( __( 'Cloudflare cache level set to %s', 'rocket' ), $cf_cache_level_return ),
+				];
+			}
+
+			// Active Minification for HTML, CSS & JS.
+			$cf_minify        = ( isset( $cf_old_settings[1] ) && 0 === $value['cloudflare_auto_settings'] ) ? $cf_old_settings[1] : 'on';
```
Screenfeed:
```suggestion
			$cf_minify        = isset( $cf_old_settings[1] ) && 0 === $value['cloudflare_auto_settings'] ? $cf_old_settings[1] : 'on';
```

inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
```diff
@@ -90,4 +123,412 @@ private function should_filter_varnish() {
...
+			} else {
+				$cloudflare_update_result[] = [
+					'result'  => 'success',
+					// translators: %s is the message returned by the CloudFlare API.
+					'message' => '<strong>' . __( 'WP Rocket: ', 'rocket' ) . '</strong>' . sprintf( __( 'Cloudflare minification %s', 'rocket' ), $cf_minify_return ),
+				];
+			}
+
+			// Deactivate Rocket Loader to prevent conflicts.
+			$cf_rocket_loader        = ( isset( $cf_old_settings[2] ) && 0 === $value['cloudflare_auto_settings'] ) ? $cf_old_settings[2] : 'off';
```
Screenfeed:
```suggestion
			$cf_rocket_loader        = isset( $cf_old_settings[2] ) && 0 === $value['cloudflare_auto_settings'] ? $cf_old_settings[2] : 'off';
```

inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
```diff
@@ -90,4 +123,412 @@ private function should_filter_varnish() {
...
+			} else {
+				$cloudflare_update_result[] = [
+					'result'  => 'success',
+					// translators: %s is the message returned by the CloudFlare API.
+					'message' => '<strong>' . __( 'WP Rocket: ', 'rocket' ) . '</strong>' . sprintf( __( 'Cloudflare rocket loader %s', 'rocket' ), $cf_rocket_loader_return ),
+				];
+			}
+
+			// Set Browser cache to 1 year.
+			$cf_browser_cache_ttl    = ( isset( $cf_old_settings[3] ) && 0 === $value['cloudflare_auto_settings'] ) ? $cf_old_settings[3] : '31536000';
```
Screenfeed:
```suggestion
			$cf_browser_cache_ttl    = isset( $cf_old_settings[3] ) && 0 === $value['cloudflare_auto_settings'] ? $cf_old_settings[3] : '31536000';
```

inc/deprecated/deprecated.php
```diff
@@ -50,7 +50,6 @@ function rocket_sanitize_cookie( $cookie ) {
 	 */
 	function set_rocket_cloudflare_async( $cf_rocket_loader ) {
 		_deprecated_function( __FUNCTION__, '2.8.16', 'set_rocket_cloudflare_rocket_loader()' );
-		return set_rocket_cloudflare_rocket_loader( $cf_rocket_loader );
```
Screenfeed: Why remove it?

inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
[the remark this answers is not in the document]
crystinutzaa: Done

inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
```diff
@@ -90,4 +123,418 @@ private function should_filter_varnish() {
...
+					}
+				}
+			}
+		}
+
+		// Let people know that the CF WP plugin is turned on.
+		if ( ! headers_sent() ) {
+			header( 'X-CF-Powered-By: WP Rocket ' . WP_ROCKET_VERSION );
+		}
+	}
```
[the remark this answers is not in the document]
crystinutzaa: Removed

inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
```diff
@@ -90,4 +123,418 @@ private function should_filter_varnish() {
...
+				foreach ( $cf_ip_ranges as $range ) {
+					if ( rocket_ipv6_in_range( $ipv6, $range ) ) {
+						if ( $_SERVER['HTTP_CF_CONNECTING_IP'] ) {
+							$_SERVER['REMOTE_ADDR'] = $_SERVER['HTTP_CF_CONNECTING_IP'];
+						}
+						break;
+					}
+				}
+			}
+		}
```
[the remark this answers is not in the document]
crystinutzaa: Refactored everything in this function.

inc/classes/Addons/Cloudflare/CloudflareFacade.php
crystinutzaa: Yes, it can be switched to $args. @Tabrisrp what do you think? Should be ok to switch it?

inc/classes/ServiceProvider/class-addons-subscribers.php
crystinutzaa: These 2 are simple classes which are used as parameters in Cloudflare & CloudflareSubscriber. I don't think it should be shared because are used only inside these classes.

inc/deprecated/deprecated.php
crystinutzaa: > You should remove these lines - @Tabrisrp commented on `deprecated.php` file.

Screenfeed: OK thanks Cristina. Then, @Tabrisrp, why remove it? ^^

Tabrisrp: When I did the review it was showing the line twice, now it's not anymore. Git doing git things  I guess.

So yeah in fact we need to keep it, sorry about that

crystinutzaa: There are 2 functions in there: `set_rocket_cloudflare_async()` and `set_rocket_cloudflare_cache_lvl()`. Should those call the new functions from CloudflareSubscriber.php?
`Cloudflare_Subscriber::set_cache_level()` 
`Cloudflare_Subscriber::set_rocket_loader()`

Also for all deprecated functions, it will throw an error because `$GLOBALS['rocket_cloudflare']` is not set anymore. How should this be handled?
https://github.com/wp-media/wp-rocket/blob/7723e789d158d2bf00ee5da49ef9158061e158eb/inc/functions/cloudflare.php#L195

Screenfeed: This is something we never discussed actually, or I'm not remembering it. So that might be the good opportunity :)

I see 3 ways to handle the function behavior:
- keep the function as it was: its behavior doesn't change. We need to be careful with the functions it uses internally though, they may change in the future.
- replace its content by the new function it is replaced by (this sentence doesn't make sense ^^): its behavior may not change at first, but the new function may change in the future, maybe by adding a parameter to it for example, or simply by changing its internal behavior. If this happens, things will break if the deprecated function is still used.
- empty the function, so it does nothing anymore: things that were working with this function will not anymore, but will not "break".

Personally I prefer the first case, it feels safer imho.

About `$GLOBALS['rocket_cloudflare']`, and global vars in general. This one seems tricky to me. Since global vars are public, we shouldn't remove them in theory. On the other hand, keeping them will result in useless process. I did a quick search but couldn't find a way to lazyload it.

Maybe we could add something like that at the beginning of the functions that use the global var:
```php
if ( ! isset( $GLOBALS['rocket_cloudflare'] ) ) {
    $GLOBALS['rocket_cloudflare'] = get_rocket_cloudflare_instance();
}
```

What do you think about all this?

crystinutzaa: For the `$GLOBALS['rocket_cloudflare'] = get_rocket_cloudflare_instance();` I would suggest to totally remove it because this was the cause of too many hits to Cloudflare API. So this GLOBAL should not be used at all from my point of view.
But at this point, the deprecated methods will show an error because of the missing GLOBAL so I think we should discuss the approach for deprecations.

Screenfeed: OK, thanks, I was not aware of this. Indeed, we should discuss all that next week.

inc/classes/Addons/Cloudflare/CloudflareFacade.php
```diff
@@ -0,0 +1,260 @@
...
+	 * @since  3.5
+	 * @author Soponar Cristina
+	 *
+	 * @param string $email      - Cloudflare Email.
+	 * @param string $api_key    - Cloudflare API Key.
+	 * @param string $zone_id    - Cloudflare Zone ID.
+	 * @param string $user_agent - Cloudflare User Agent.
+	 * @return void
+	 */
+	public function set_api_credentials( $email, $api_key, $zone_id, $user_agent ) {
```
Tabrisrp: You could set the `user_agent` value directly in the method, no need to pass it as a parameter.

inc/classes/ServiceProvider/class-addons-subscribers.php
Tabrisrp: Yes `add` is fine here, I use `share` for the subscribers because I want to be able to access the same instance of the class in case we need to remove hooks for example

inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
```diff
@@ -90,4 +123,397 @@ private function should_filter_varnish() {
...
+
+	/**
+	 * Purge CloudFlare cache
+	 *
+	 * @since 2.5
+	 */
+	public function purge_cache_no_die() {
+		if ( ! isset( $_GET['_wpnonce'] ) || ! wp_verify_nonce( sanitize_key( $_GET['_wpnonce'] ), 'rocket_purge_cloudflare' ) ) {
+			wp_nonce_ays( '' );
+		}
```
Tabrisrp: I would keep this part into the original `purg_cache` method

inc/classes/Addons/Cloudflare/Cloudflare.php
```diff
@@ -0,0 +1,636 @@
...
+
+	/**
+	 * Get default Cloudflare IPs.
+	 *
+	 * @since  3.5
+	 * @author Soponar Cristina
+	 *
+	 * @return Object Default Cloudflare connecting IPs.
+	 */
+	public function get_default_ips() {
```
Tabrisrp:
```suggestion
	private function get_default_ips() {
```

It's only used internally, no need for it to be public

inc/deprecated/deprecated.php
Tabrisrp: I don't think any of our cloudflare functions were used externally by other plugins or anything, so the deprecation is there to prevent any error after the update mainly.

My opinion is to remove any call to the globals, and make the functions who used it to do nothing, only make sure they don't trigger any error.

crystinutzaa: So, if I remove the global from each deprecated function, then I should remove the full code inside the function because it will throw an notice with undefined:
https://github.com/wp-media/wp-rocket/blob/38e9239a402d5064b52c4e63af76e95532fb2f0d/inc/deprecated/3.5.php#L244-L254

So the solution in this case is to keep only the `_deprecated_function` in action for each deprecated cloudflare function?

Tabrisrp: For me that's fine yes

crystinutzaa: Removed all function content for all deprecated cloudflare functions which contained `$GLOBALS['rocket_cloudflare']; `

PR #2120, opened 2019-11-21
Pantheon Write Problem: Update code with custom path for CONFIG & CACHE file

spinxdigital:
Pantheon has protected path and this plugin can't create file on protected path.
So we did path variable as custom global constant so it can change from wp-config.

With that it will help to put cache, config file to custom location like inside upload folder.

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -95,7 +95,7 @@ function rocket_generate_advanced_cache_file() {
 	}
 	$done = true;
 
-	rocket_put_content( WP_CONTENT_DIR . '/advanced-cache.php', get_rocket_advanced_cache_file() );
+	rocket_put_content( WP_ROCKET_ADVANCED_CACHE_FILE, get_rocket_advanced_cache_file() );
 }
 
 /**
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -95,7 +95,7 @@ function rocket_generate_advanced_cache_file() {
 	}
 	$done = true;
 
-	rocket_put_content( WP_ROCKET_ADVANCED_CACHE_FILE, get_rocket_advanced_cache_file() );
+	rocket_put_content( WP_CONTENT_DIR . '/advanced-cache.php', get_rocket_advanced_cache_file() );
 }
 
 /**
--- a/uninstall.php
+++ b/uninstall.php
@@ -2,7 +2,11 @@
 defined( 'WP_UNINSTALL_PLUGIN' ) || die( 'Cheatin&#8217; uh?' );
 
 if ( ! defined( 'WP_ROCKET_CACHE_ROOT_PATH' ) ) {
-	define( 'WP_ROCKET_CACHE_ROOT_PATH', WP_ROCKET_CACHE_ROOT_PATH );
+	define( 'WP_ROCKET_CACHE_ROOT_PATH', WP_CONTENT_DIR . '/cache/' );
+}
+
+if ( ! defined( 'WP_ROCKET_CONFIG_PATH' ) ) {
+	define( 'WP_ROCKET_CONFIG_PATH', WP_CONTENT_DIR . '/wp-rocket-config/' );
 }
 
 // Delete all transients.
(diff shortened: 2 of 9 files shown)


spinxdigital:
@crystinutzaa  Can you please merge my commit? 
Also not sure why it's failed on last one.

uninstall.php
```diff
@@ -2,7 +2,7 @@
 defined( 'WP_UNINSTALL_PLUGIN' ) || die( 'Cheatin&#8217; uh?' );
 
 if ( ! defined( 'WP_ROCKET_CACHE_ROOT_PATH' ) ) {
-	define( 'WP_ROCKET_CACHE_ROOT_PATH', WP_CONTENT_DIR . '/cache/' );
+	define( 'WP_ROCKET_CACHE_ROOT_PATH', WP_ROCKET_CACHE_ROOT_PATH );
 }
```
Tabrisrp: that doesn't really make sense: if the constant is not defined, you can't assign a value that doesn't exist

spinxdigital: Fixed constant code issue.

spinxdigital:
Thank you for detailed review. @Tabrisrp 

I have fixed code constant error and revert back code about advanced-cache.php location.

Also found much batter solution with change full path to relative path which user can chick-in into source controller and should work for all dev, staging and live environment. 
But I am sure you know much batter test cases that if that will broke anything else or not.

inc/functions/files.php
```diff
@@ -21,8 +21,8 @@ function get_rocket_advanced_cache_file() {
 
 	// Include the Mobile Detect class if we have to create a different caching file for mobile.
 	if ( is_rocket_generate_caching_mobile_files() ) {
-		$buffer .= "if ( file_exists( '" . WP_ROCKET_VENDORS_PATH . "classes/class-rocket-mobile-detect.php' ) && ! class_exists( 'Rocket_Mobile_Detect' ) ) {\n";
-		$buffer .= "\tinclude_once '" . WP_ROCKET_VENDORS_PATH . "classes/class-rocket-mobile-detect.php';\n";
+		$buffer .= "if ( file_exists( WP_CONTENT_DIR . '/plugins/wp-rocket/inc/vendors/classes/class-rocket-mobile-detect.php' ) && ! class_exists( 'Rocket_Mobile_Detect' ) ) {\n";
+		$buffer .= "\tinclude_once WP_CONTENT_DIR . '/plugins/wp-rocket/inc/vendors/classes/class-rocket-mobile-detect.php';\n";
```
Tabrisrp: What's the reasoning behind changing those paths? The constant `WP_ROCKET_VENDORS_PATH` is perfectly fine here

inc/functions/files.php
```diff
@@ -48,19 +48,21 @@ function get_rocket_advanced_cache_file() {
...
 
+	$cache_dir_path = defined( \'WP_ROCKET_CACHE_ROOT_PATH\' ) ? WP_ROCKET_CACHE_ROOT_PATH : WP_CONTENT_DIR . \'/cache/wp-rocket/\';
 	( new \WP_Rocket\Buffer\Cache(
 		new \WP_Rocket\Buffer\Tests(
 			$rocket_config_class
 		),
 		$rocket_config_class,
 		[
-			\'cache_dir_path\' => \'' . WP_ROCKET_CACHE_PATH . '\',
+			\'cache_dir_path\' => $cache_dir_path,
```
Tabrisrp: These changes don't seem necessary to me either

inc/process-autoloader.php
```diff
@@ -2,7 +2,7 @@
 
 spl_autoload_register(
 	function( $class ) {
-		$rocket_path    = WP_ROCKET_PATH;
+		$rocket_path    = WP_CONTENT_DIR . '\plugins\wp-rocket/';
```
Tabrisrp: same here

spinxdigital:
@Tabrisrp 

I thought I can make path relative compare to absolute server path.
Like this path see on that file. `/srv/bindings/7580cf3199e348fca7cf1120d0c481a8/code/web/app/plugins/wp-rocket/` 
But I though I like previous version idea where path was like this. So it might use same file to different to dev, test and live server. But I think user can activate plugin on different environment and plugin will create that file.  
```
$rocket_cache_path = WP_CONTENT_DIR . '/uploads/cache/wp-rocket/';
$rocket_config_path = WP_CONTENT_DIR . '/uploads/wp-rocket-config/';
```

PR #2123, opened 2019-11-21
Fix: prevent false positive on missing tag notice

crystinutzaa:
Fix: prevent false positive notices on missing tags </body>, </html> and wp_footer()


Screenfeed:
The base branch is 3.4.2.

PR #2125, opened 2019-11-22
Add notice for WP Optimize incompatibility

WordPresseur:
WP Optimize is now a cache plugin and it triggers the advanced-cache.php notice if it’s activated before WP Rocket.

### DIFF
--- a/inc/admin/ui/notices.php
+++ b/inc/admin/ui/notices.php
@@ -126,6 +126,7 @@ function rocket_plugins_to_deactivate() {
 		'force-gzip'                                 => 'force-gzip/force-gzip.php',
 		'enable-gzip-compression'                    => 'enable-gzip-compression/enable-gzip-compression.php',
 		'leverage-browser-caching'                   => 'leverage-browser-caching/leverage-browser-caching.php',
+		'wp-optimize'				     => 'wp-optimize/wp-optimize.php'
 	];
 
 	if ( get_rocket_option( 'lazyload' ) ) {


remyperona (approved):

alfonso100:
Related tickets: 
https://secure.helpscout.net/conversation/1018452934/133918
https://secure.helpscout.net/conversation/1019845508/134302
https://secure.helpscout.net/conversation/1014889018/132934
https://secure.helpscout.net/conversation/1015502104/133077
https://secure.helpscout.net/conversation/1013688957/132750

GeekPress:
Can we do it only if their cache option is turned on?

Their database feature as actually more options than ours. Some users could just use WP-Optimize (as before the cache feature was implemented) for the database feature and not caching.

It could be frustrating for users to not be able to use anymore WP-Optimize database feature, because of an option in this plugin they don't activate.

webtrainingwheels:
@GeekPress This issue is occurring pretty often, could we address this in an upcoming release? It should have a custom error message though, not the same as all other incompatible plugins.

arunbasillal:
@GeekPress I tried out WP Optimize to see if having both the plugins can benefit users. 

> Their database feature as actually more options than ours. Some users could just use WP-Optimize (as before the cache feature was implemented) for the database feature and not caching.

While this is true, most of the extra features are potentially "one-time" tasks and are not available for schedule. 

These include:
- Removing unused tables
- These: 
image

A user insisting to use these can disable WP Rocket and use WP-Optimize for these. I don't see any other reason why a user would want to use WP Rocket and WP-Optimize together. 

> Should we directly put WP-Optimize on global incompatible plugin like WP Super Cache / W3 Total Cache, etc...?

My suggestion is to go with this. Easier for us and lesser development overhead in the future.

PR #2137, opened 2019-11-28
Fix #2130 - Cloudflare purge notice displayed on content update

### LINKED ISSUE #2130: 3.4.2: Cloudflare purge notice displayed on content update
When updating content we don't normally show a notice that the WP Rocket cache has been purged. But with the new Cloudflare integration re: Cache Everything, we are showing a notice, when using the Classic Editor (didn't see it when using Gutenberg), that the CF cache has been purged: 
https://recordit.co/2oWn9LA9j2

crystinutzaa:
Removed Cloudflare purge notice displayed on content update


inc/functions/cloudflare.php
```diff
@@ -740,21 +740,5 @@ function rocket_auto_purge_cloudflare_by_url( $post, $purge_urls, $lang ) {
 
 	// Purge CloudFlare.
 	$cf_purge = rocket_purge_cloudflare_by_url( $post, $purge_urls, $lang );
```
Screenfeed: No need to assign the result to `$cf_purge` then.
```suggestion
	rocket_purge_cloudflare_by_url( $post, $purge_urls, $lang );
```

Issue #1323, opened 2018-11-09
The Facebook Pixel add-on breaks the Facebook Comments plugin


vmanthos:
When enabled, the Facebook Pixel add-on breaks the **[Facebook Comments Plugin](https://developers.facebook.com/docs/plugins/comments/)**.

**To Reproduce**
1. Go to https://www.tediado.com.br/11/improvavel-escolinha-improvavel-20/?nocache
2. There is a comment box there (https://vgy.me/kAd0ow.jpg) 
3. Go to https://www.tediado.com.br/11/improvavel-escolinha-improvavel-20/
4. The comment box is gone

It seems that when we replace the `https:\/\/connect.facebook.net\/pt_BR\/sdk.js`, we change the scheme:
`http:\/\/connect.facebook.net\/pt_BR\/sdk.js`

Check the diff here:
https://www.diffchecker.com/9D4c9PgZ

It's taken from the file downloaded directly from Facebook (on the left) and the one that we produce (on the right).

You can download the files here:
http://bit.ly/2Db4xee

**Related Ticket**
https://secure.helpscout.net/conversation/704026920/85516?folderId=2135277

Screenfeed:
Sorry but I couldn’t reproduce the issue, the file I get has the https.
On a side note, instead of having `"revision": "4520341"` on the same line, I get `"revision":"4521801"`, so it seems it’s more recent. Since we don’t modify this file content, I'm guessing this revision `4520341` was delivered with this bug but not the new release.
I’d advise to do a new test on the customer’ site (by removing all the old files first).

webtrainingwheels:
another case https://secure.helpscout.net/conversation/759292826/93728?folderId=377611

alfonso100:
Another case: https://secure.helpscout.net/conversation/779026599/96657?folderId=2683093

viobru:
Another case: https://secure.helpscout.net/conversation/954924180/122979/

crystinutzaa:
This is not going to work because it needs:
`window.fbAsyncInit = function fbAsyncInit () {FB.init()})}`

However I noticed that this type of URL will not be busted:
`<script async defer crossorigin="anonymous" src="https://connect.facebook.net/en_US/sdk.js#xfbml=1&version=v5.0&appId=323009824493514&autoLogAppEvents=1"></script>`

Only this will get busted:
```
<script>(function(d, s, id) {
  var js, fjs = d.getElementsByTagName(s)[0];
[...]

Issue #1326, opened 2018-11-09
The Facebook Pixel add-on breaks the Facebook Like/Share plugin


vmanthos:
When enabled, the Facebook Pixel add-on breaks the [Facebook Like/Share Plugin](https://developers.facebook.com/docs/plugins#like-share-send).

**To Reproduce**

1. Go to http://wealthweekly.com/contest?nocache. There is a Facebook Like/Share box there (https://vgy.me/UmTONB.jpg), which is embedded using an `iframe`.
2. Go to http://wealthweekly.com/contest/
The Like/Share box is gone.

We replace only the `revision:` value in comparison to the sdk.js that we fetch from Facebook. Check the diff here:
https://www.diffchecker.com/FT5VIg3v

**Related Ticket**
https://secure.helpscout.net/conversation/704013562/85513?folderId=2135277

crystinutzaa:
Fixed with v3.4.3 release

alfonso100:
This issue was fixed and closed, but I've just encountered a similar case here with WPR 3.5.5.1: 
https://secure.helpscout.net/conversation/1178785091/168787/

If the Facebook Add-on is enabled, this likebox doesn't work: https://jmp.sh/dUFCT6O

Saved the JS files in the ticket for reference

GeekPress:
@crystinutzaa Can you have a look to this?
@alfonso100 can you ask a FTP access? It will be easer for @crystinutzaa if we can try directly on the customer website.

Thanks =)

crystinutzaa:
@GeekPress sure. @alfonso100 please contact me when we have access and I will take a closer look

piotrbak:
Related ticket:
https://secure.helpscout.net/conversation/1233430729/182036?folderId=2415573

GeekPress:
Closing this issue as Facebook Pixels add-on will be removed in WP Rocket 3.9

PR #2144, opened 2019-12-02
Fixes: #1323,#1326 - Facebook Pixel Addon need to run FB.init()

crystinutzaa:
Add FB.init() when Facebook Pizel Addon is enabled:

```
<script>
  window.fbAsyncInit = function() {
    FB.init({
      appId            : 'your-app-id',
      autoLogAppEvents : true,
      xfbml            : true,
      version          : 'v5.0'
    });
  };
</script>
```

### DIFF
--- a/inc/classes/busting/class-facebook-sdk.php
+++ b/inc/classes/busting/class-facebook-sdk.php
@@ -100,6 +100,12 @@ class Facebook_SDK extends Abstract_Busting {
 
 		$html      = str_replace( $tag, $replace_tag, $html );
 		$file_path = $this->get_busting_file_path( $locale );
+		$xfbml     = $this->get_xfbml_from_url( $tag );
+		$app_id    = $this->get_appId_from_url( $tag );
+		$version   = $this->get_version_from_url( $tag );
+
+		// Add FB async init.
+		$html .= '<script>window.fbAsyncInit = function fbAsyncInit () {FB.init({appId: \'' . $app_id . '\',xfbml: ' . $xfbml . ',version: \'' . $version . '\'})}</script>';
 
 		$this->is_replaced = true;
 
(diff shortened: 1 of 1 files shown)


inc/classes/busting/class-facebook-sdk.php
```diff
@@ -100,6 +100,12 @@ public function replace_url( $html ) {
 
 		$html      = str_replace( $tag, $replace_tag, $html );
 		$file_path = $this->get_busting_file_path( $locale );
+		$xfbml     = $this->get_xfbml_from_url( $tag );
+		$app_id    = $this->get_appId_from_url( $tag );
+		$version   = $this->get_version_from_url( $tag );
+
+		// Add FB async init.
+		$html .= '<script>window.fbAsyncInit = function fbAsyncInit () {FB.init({appId: \'' . $app_id . '\',xfbml: ' . $xfbml . ',version: \'' . $version . '\'})}</script>';
```
Tabrisrp: The script should be inserted at a specific place in the HTML markup, like just before the `</body>` tag.

Also, we're potentially inserting variables with a false value, so the script will not work correctly in this case. We should only insert it if all values are ok.

inc/classes/busting/class-facebook-sdk.php
```diff
@@ -455,6 +461,66 @@ private function get_locale_from_url( $url ) {
...
+	 */
+	private function get_version_from_url( $url ) {
+		$pattern = '@//connect\.facebook\.net/(?<locale>[a-zA-Z_-]+)/sdk\.js#xfbml=(?<xfbml>[0-9]+)&version=(?<version>[a-zA-Z0-9.]+)&appId=(?<appId>[0-9]+)@i';
+
+		if ( ! preg_match( $pattern, $url, $matches ) ) {
+			return false;
+		}
+
+		return $matches['version'];
+	}
```
Tabrisrp: I don't think repeating almost the same method three times is the way to go. Mixed returned values are also something to avoid, as it makes the output less predictable.

I think this should all be a single method, including the inline script generation. And the inline script would be returned only if all values were matched.

inc/classes/busting/class-facebook-sdk.php
```diff
@@ -455,6 +461,66 @@ private function get_locale_from_url( $url ) {
...
+	 *
+	 * @since  3.4.3
+	 * @access private
+	 * @author Soponar Cristina
+	 *
+	 * @param  string $url Any string containing the URL to bust.
+	 * @return string|bool The XFBML on success. False on failure.
+	 */
+	private function get_xfbml_from_url( $url ) {
+		$pattern = '@//connect\.facebook\.net/(?<locale>[a-zA-Z_-]+)/sdk\.js#xfbml=(?<xfbml>[0-9]+)&version=(?<version>[a-zA-Z0-9.]+)&appId=(?<appId>[0-9]+)@i';
```
Screenfeed: This will fail if `xfbml`, `version`, and `appId` are not in that order.
```suggestion
		$pattern = '@//connect\.facebook\.net/(?<locale>[a-zA-Z_-]+)/sdk\.js#(?:.+&)?xfbml=(?<xfbml>[0-9]+)@i';
```

inc/classes/busting/class-facebook-sdk.php
```diff
@@ -455,6 +461,66 @@ private function get_locale_from_url( $url ) {
...
+	 *
+	 * @since  3.4.3
+	 * @access private
+	 * @author Soponar Cristina
+	 *
+	 * @param  string $url Any string containing the URL to bust.
+	 * @return string|bool The appId on success. False on failure.
+	 */
+	private function get_appId_from_url( $url ) {
+		$pattern = '@//connect\.facebook\.net/(?<locale>[a-zA-Z_-]+)/sdk\.js#xfbml=(?<xfbml>[0-9]+)&version=(?<version>[a-zA-Z0-9.]+)&appId=(?<appId>[0-9]+)@i';
```
Screenfeed: Same here.
```suggestion
		$pattern = '@//connect\.facebook\.net/(?<locale>[a-zA-Z_-]+)/sdk\.js#(?:.+&)?appId=(?<appId>[0-9]+)@i';
```

inc/classes/busting/class-facebook-sdk.php
```diff
@@ -455,6 +461,66 @@ private function get_locale_from_url( $url ) {
...
+	 *
+	 * @since  3.4.3
+	 * @access private
+	 * @author Soponar Cristina
+	 *
+	 * @param  string $url Any string containing the URL to bust.
+	 * @return string|bool The version on success. False on failure.
+	 */
+	private function get_version_from_url( $url ) {
+		$pattern = '@//connect\.facebook\.net/(?<locale>[a-zA-Z_-]+)/sdk\.js#xfbml=(?<xfbml>[0-9]+)&version=(?<version>[a-zA-Z0-9.]+)&appId=(?<appId>[0-9]+)@i';
```
Screenfeed: And here.
```suggestion
		$pattern = '@//connect\.facebook\.net/(?<locale>[a-zA-Z_-]+)/sdk\.js#(?:.+&)?version=(?<version>[a-zA-Z0-9.]+)@i';
```
And this will also solve Remy’s "repeating almost the same method" issue ^^.

crystinutzaa: Thank you @Screenfeed . I was looking into something like this. You are a lifesaver!

inc/classes/busting/class-facebook-sdk.php
crystinutzaa: I think only APPID needs to be validated. In rest I can set default values for xfbml & version. 
And where this should be placed?

inc/classes/busting/class-facebook-sdk.php
```diff
@@ -100,6 +100,15 @@ public function replace_url( $html ) {
 
 		$html      = str_replace( $tag, $replace_tag, $html );
 		$file_path = $this->get_busting_file_path( $locale );
+		$xfbml     = $this->get_xfbml_from_url( $tag ); // Default value should be set to false.
+		$app_id    = $this->get_appId_from_url( $tag ); // APP_ID is the only required value.
+		$version   = false === $this->get_version_from_url( $tag ) ? 'v5.0' : $this->get_version_from_url( $tag ); // If version is not available set it to the latest: v.5.0.
```
Tabrisrp: Small enhancement: first assign the return of `$this->get_version_from_url( $tag )` to a variable, and then do the ternary operator check with that variable. That way you do only one method call, not 2

Screenfeed:
What will happen if the tag is replaced L101 but the script is not added L111?

Screenfeed:
Well, nothing less than before I guess.

crystinutzaa:
@Screenfeed - what will happen... the error will persist. Basically FB Comments, Like & Share will not be visible because FB API is not initialized.

PR #2147, opened 2019-12-03
Fixes #1945 cache race condition with multiple servers

### LINKED ISSUE #1945: Cache race condition with shared filesystem for multiple servers
**Describe the bug**
On a setup with multiple servers reading the same shared filesystem for cached files, it can happen that one server is requesting the cache file while another is still writing it, causing the first one to return a blank/incomplete page.

This is happening because we start writing directly in the final cache file destination.

As a solution, instead we can first write to a temp file which can then be renamed to the final name once complete.

Tabrisrp:
Prevent blank page when multiple servers are using the same shared filesystem. Instead of directly writing in the final cache file, we first write in a temp file, and rename it when it's done.

Also fixes an undetected issue with WebP and Firefox version > 66

### DIFF
--- a/inc/classes/Buffer/class-cache.php
+++ b/inc/classes/Buffer/class-cache.php
@@ -352,7 +352,14 @@ class Cache extends Abstract_Buffer {
 		rocket_direct_filesystem()->move( $temp_filepath, $cache_filepath );
 
 		if ( function_exists( 'gzencode' ) ) {
-			rocket_put_content( $temp_gzip_filepath, gzencode( $content, apply_filters( 'rocket_gzencode_level_compression', 3 ) ) );
+			/**
+			 * Filters the Gzip compression level to use for the cache file
+			 *
+			 * @param int $compression_level Compression level between 0 and 9.
+			 */
+			$compression_level = apply_filters( 'rocket_gzencode_level_compression', 3 );
+
+			rocket_put_content( $temp_gzip_filepath, gzencode( $content, $compression_level ) );
 			rocket_direct_filesystem()->move( $temp_gzip_filepath, $gzip_filepath );
 		}
 	}
--- a/inc/classes/Buffer/class-cache.php
+++ b/inc/classes/Buffer/class-cache.php
@@ -305,13 +305,7 @@ class Cache extends Abstract_Buffer {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


inc/classes/Buffer/class-cache.php
```diff
@@ -334,6 +328,35 @@ public function maybe_process_buffer( $buffer ) {
...
+		if ( rocket_direct_filesystem()->exists( $temp_filepath ) ) {
+			return;
+		}
+
+		// Save the cache file.
+		rocket_put_content( $temp_filepath, $content );
+		rocket_direct_filesystem()->move( $temp_filepath, $cache_filepath );
+
+		if ( function_exists( 'gzencode' ) ) {
+			rocket_put_content( $temp_gzip_filepath, gzencode( $content, apply_filters( 'rocket_gzencode_level_compression', 3 ) ) );
```
Screenfeed: Missing a DocBlock here for the filter.

inc/classes/subscriber/Media/class-webp-subscriber.php
```diff
@@ -154,7 +154,15 @@ public function convert_to_webp( $html ) {
 		}
 
 		if ( ! $http_accept || false === strpos( $http_accept, 'webp' ) ) {
-			return $html;
+			$user_agent = isset( $this->server['HTTP_USER_AGENT'] ) ? $this->server['HTTP_USER_AGENT'] : '';
+
+			if ( $user_agent && preg_match( '#Firefox/(?<version>[0-9]{2})#i', $this->server['HTTP_USER_AGENT'], $matches ) ) {
```
Screenfeed: With the current pace of version increment, I think it is safer to allow versions > 99.
```suggestion
			if ( $user_agent && preg_match( '#Firefox/(?<version>[0-9]{2,})#i', $this->server['HTTP_USER_AGENT'], $matches ) ) {
```

inc/classes/Buffer/class-cache.php
Tabrisrp: I added it but I didn't add `@since` and `@author` because it's so old I have no idea

Screenfeed: Sure 👍

PR #2150, opened 2019-12-04
Fixed #2124 - running rocket_clean_post() when saving drafts

### LINKED ISSUE #2124: rocket_clean_post() running when saving drafts
The `rockec_clean_post()` function is run after saving drafts of posts:
https://github.com/wp-media/wp-rocket/blob/281eb493f44e873a9685f379cec701e8ebcf3532/inc/common/purge.php#L56

This means that everything that is hooked to `after_rocket_clean_post` will also run.

For example, if someone is using Cloudflare's "Cache Everything" the `rocket_auto_purge_cloudflare_by_url()` function will run, increasing the hits to Cloudflare's API.

**Proposed solution:**

Add a check to see if a post is **draft** after the following line:
https://github.com/wp-media/wp-rocket/blob/281eb493f44e873a9685f379cec701e8ebcf3532/inc/common/purge.php#L56

```
if( 'draft' ==  get_post_status( $post_id ) ){
		return;
}
```

**Related ticket:**
https://secure.helpscout.net/conversation/1011323712/132260?folderId=2135277

crystinutzaa:
Fixes running rocket_clean_post() when saving drafts.

Cleaning rocket cache when a post status is changed from Publish to Draft.
There could be other statuses which could need to be excluded (eg. inherit)

### DIFF
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -43,7 +43,7 @@ add_filter( 'widget_update_callback', 'rocket_widget_update_callback' );
  *
  * @param int     $post_id The post ID.
  * @param WP_Post $post    WP_Post object.
- * @return array           Array with all URLs which needs to be purged.
+ * @return array           Array with all URLs which need to be purged.
  */
 function rocket_get_purge_urls( $post_id, $post ) {
 	$purge_urls = [];
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -36,58 +36,17 @@ function rocket_widget_update_callback( $instance ) {
 add_filter( 'widget_update_callback', 'rocket_widget_update_callback' );
 
 /**
- * Update cache when a post is updated or commented
+ * Get post purge urls.
  *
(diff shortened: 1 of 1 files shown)


inc/common/purge.php
```diff
@@ -36,59 +36,18 @@ function rocket_widget_update_callback( $instance ) {
...
- * @since 1.3.0 Purge all parents of the post and the author page
- * @since 1.2.2 Add wp_trash_post and delete_post to purge cache when a post is trashed or deleted
- * @since 1.1.3 Use clean_post_cache instead of transition_post_status, transition_comment_status and preprocess_comment
- * @since 1.0
+ * @since 3.4.3
+ * @author Soponar Cristina
  *
  * @param int     $post_id The post ID.
  * @param WP_Post $post    WP_Post object.
+ * @return array           Array with all URLs which needs to be purged.
```
Screenfeed: Typo ^^
```suggestion
 * @return array           Array with all URLs which need to be purged.
```

PR #2155, opened 2019-12-04
Fix/identify cloudflare cache everything rule

crystinutzaa:
Identify Cloudflare Cache everything rule correctly.

Were cases when it was a false positive

### DIFF
--- a/inc/functions/cloudflare.php
+++ b/inc/functions/cloudflare.php
@@ -659,8 +659,7 @@ function rocket_cf_has_page_rule( $action_value ) {
 		}
 
 		$cf_page_rule_arr = wp_json_encode( $cf_page_rule );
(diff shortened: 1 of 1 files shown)


inc/functions/cloudflare.php
```diff
@@ -658,8 +658,8 @@ function rocket_cf_has_page_rule( $action_value ) {
 			throw new Exception( $errors );
 		}
 
-		$cf_page_rule_arr = (array) $cf_page_rule;
-		return in_array( $action_value, $cf_page_rule_arr );
+		$cf_page_rule_arr = wp_json_encode( $cf_page_rule );
+		return preg_match( '/' . $action_value . '/', $cf_page_rule_arr );
```
Screenfeed:
```suggestion
		return preg_match( '/' . preg_quote( $action_value, '/' ) . '/', $cf_page_rule_arr );
```

PR #2168, opened 2019-12-09
Fix #2136 - Cache isn't cleared according to cache lifespan if that's set to < 1 hour

### LINKED ISSUE #2136: Cache isn't cleared according to cache lifespan if that's set to < 1 hour
With the new implementation of purging "expired" cache files, if the Cache Lifespan is set to a value less than 1 hour, that won't be respected.

This is because of the `rocket_purge_time_event` firing every hour.  

For example, if the cache lifespan is set to 30 minutes, you'd expect the files to be cleared twice in an hour. However, this will only happen once.

A side effect is that the majority of files will be cleared since it's most likely they will have long-expired.

**Proposed Solution:**

If the cache lifespan is lower than 60 minutes, match the schedule of the cron event.

crystinutzaa:
Cache isn't cleared according to cache lifespan if that's set to < 1 hour

Proposed Solution:

If the cache lifespan is lower than 60 minutes, match the schedule of the cron event.

Based on v3.4.3 doc


Screenfeed:
How about creating a `get_interval()` private method, instead of duplicating code?

inc/classes/subscriber/Cache/class-expired-cache-purge-subscriber.php
```diff
@@ -88,8 +107,34 @@ public static function get_subscribed_events() {
...
+		}
+	}
+
+	/**
+	 * Gets the interval when the scheduled clean cache purge needs to run.
+	 * If Minutes option is selected, then the interval will be set to minutes.
+	 * If Hours / Days options are selected, then it will be set to 1 hour.
+	 *
+	 * @since  3.4.3
+	 * @access public
```
Tabrisrp:
```suggestion
	 * @access private
```

Screenfeed: #copyPasteDrama

inc/classes/subscriber/Cache/class-expired-cache-purge-subscriber.php
```diff
@@ -70,9 +70,28 @@ public static function get_subscribed_events() {
...
+	 * @author Soponar Cristina
+	 *
+	 * @param array $schedules An array of non-default cron schedules.
+	 */
+	public function custom_cron_schedule( $schedules ) {
+		$interval                                        = $this->get_interval();
+		$schedules['rocket_expired_cache_cron_interval'] = [
+			'interval' => $interval,
```
Screenfeed: Simpler:

```php
			'interval' => $this->get_interval(),
```

PR #2172, opened 2019-12-11
Fix clean rocket post cache on slug change when post name is empty & status is not published

### LINKED ISSUE #1936: Cache not clearing when post permalink (slug) changes

crystinutzaa:
Fix clean rocket post cache when old slug is empty and old post_status is draft , pending or auto-draft (there is no cache to be cleaned in this situation).

Fixes #1936 [per discussion here](https://github.com/wp-media/wp-rocket/issues/1936#issuecomment-561142237).

Test Cases:

- with WooCommerce Follow-up Emails plugin
In order to test it go to: Tools > Scheduled Actions > Pending and run any of the 2 actions listed
- Publishing a new post causes the whole cache to be cleared.

TODO:

[X] Test locally with 2 test cases (see above)
[X] Identify needed post statuses
[X] Unit test
[ ] Integration test - @hellofromtonya This one will be in a new PR once a few updates are made to the integration bootstrapping process.

### DIFF
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -652,9 +652,7 @@ function rocket_clean_cache_theme_update( $wp_upgrader, $hook_extra ) {
  */
 function rocket_clean_post_cache_on_slug_change( $post_id, $post_data ) {
 	// Bail out if the post status is draft, pending or auto-draft.
-	if ( 'draft' === get_post_field( 'post_status', $post_id ) ||
-			'pending' === get_post_field( 'post_status', $post_id ) ||
-			'auto-draft' === get_post_field( 'post_status', $post_id ) ) {
+	if ( in_array( get_post_field( 'post_status', $post_id ), [ 'draft', 'pending', 'auto-draft' ], true ) ) {
 		return;
 	}
 	// Bail out if the slug hasn't changed.
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -651,8 +651,20 @@ function rocket_clean_cache_theme_update( $wp_upgrader, $hook_extra ) {
  * @param array $post_data Array of unslashed post data.
  */
 function rocket_clean_post_cache_on_slug_change( $post_id, $post_data ) {
-	if ( get_post_field( 'post_name', $post_id ) !== $post_data['post_name'] ) {
-        rocket_clean_files( get_the_permalink( $post_id ) );
-    }
+	// Bail out if the post status is draft, pending or auto-draft.
+	if ( 'draft' === get_post_field( 'post_status', $post_id ) ||
+			'pending' === get_post_field( 'post_status', $post_id ) ||
+			'auto-draft' === get_post_field( 'post_status', $post_id ) ) {
+		return;
+	}
+	// Bail out if the slug hasn't changed.
+	if ( get_post_field( 'post_name', $post_id ) === $post_data['post_name'] ) {
+		return;
+	}
+	// Bail out if the old slug has changed, but is empty.
+	if ( empty( get_post_field( 'post_name', $post_id ) ) ) {
+		return;
+	}
+	rocket_clean_files( get_the_permalink( $post_id ) );
 }
 add_action( 'pre_post_update', 'rocket_clean_post_cache_on_slug_change', 10, 2 );
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -655,14 +655,15 @@ function rocket_clean_post_cache_on_slug_change( $post_id, $post_data ) {
 	if ( in_array( get_post_field( 'post_status', $post_id ), [ 'draft', 'pending', 'auto-draft' ], true ) ) {
 		return;
 	}
+	$post_name = get_post_field( 'post_name', $post_id );
 	// Bail out if the slug hasn't changed.
-	if ( get_post_field( 'post_name', $post_id ) === $post_data['post_name'] ) {
+	if ( $post_name === $post_data['post_name'] ) {
 		return;
 	}
 	// Bail out if the old slug has changed, but is empty.
-	if ( empty( get_post_field( 'post_name', $post_id ) ) ) {
+	if ( empty( $post_name ) ) {
 		return;
 	}
 	rocket_clean_files( get_the_permalink( $post_id ) );
 }
-add_action( 'pre_post_update', 'rocket_clean_post_cache_on_slug_change', 10, 2 );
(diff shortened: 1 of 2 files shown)


inc/common/purge.php
```diff
@@ -651,8 +651,18 @@ function rocket_clean_cache_theme_update( $wp_upgrader, $hook_extra ) {
...
 function rocket_clean_post_cache_on_slug_change( $post_id, $post_data ) {
-	if ( get_post_field( 'post_name', $post_id ) !== $post_data['post_name'] ) {
-        rocket_clean_files( get_the_permalink( $post_id ) );
-    }
+	// Bail out if the post status is draft, pending or auto-draft.
+	if ( in_array( get_post_field( 'post_status', $post_id ), [ 'draft', 'pending', 'auto-draft' ], true ) ) {
+		return;
+	}
+	// Bail out if the slug hasn't changed.
+	if ( get_post_field( 'post_name', $post_id ) === $post_data['post_name'] ) {
```
hellofromtonya: @crystinutzaa `get_post_field()` can be an expensive operation as it goes through several different filters and processes. I'd suggest caching the post name in a local variable and then reusing it in both instance:

```php
$post_name = get_post_field( 'post_name', $post_id );

if ( $post_name === $post_data['post_name'] ) {
```

and same change on line 663.

inc/common/purge.php
```diff
@@ -651,8 +651,18 @@ function rocket_clean_cache_theme_update( $wp_upgrader, $hook_extra ) {
...
+		return;
+	}
+	// Bail out if the old slug has changed, but is empty.
+	if ( empty( get_post_field( 'post_name', $post_id ) ) ) {
+		return;
+	}
+	rocket_clean_files( get_the_permalink( $post_id ) );
 }
 add_action( 'pre_post_update', 'rocket_clean_post_cache_on_slug_change', 10, 2 );
```
hellofromtonya: @crystinutzaa 2 things:

1. Let's make our callback fire as late as possible by changing it from `10` to `PHP_INT_MAX`.  Why? By changing the priority number to the highest number possible, we are registering our callback in the last queue. That means ours will called after everything other 3rd party callback that has a priority number less than ours.
2. Plus, add an empty line at end of this file, ie for WPCS.

inc/common/purge.php
```diff
@@ -651,8 +651,18 @@ function rocket_clean_cache_theme_update( $wp_upgrader, $hook_extra ) {
  * @param array $post_data Array of unslashed post data.
  */
 function rocket_clean_post_cache_on_slug_change( $post_id, $post_data ) {
-	if ( get_post_field( 'post_name', $post_id ) !== $post_data['post_name'] ) {
-        rocket_clean_files( get_the_permalink( $post_id ) );
-    }
+	// Bail out if the post status is draft, pending or auto-draft.
+	if ( in_array( get_post_field( 'post_status', $post_id ), [ 'draft', 'pending', 'auto-draft' ], true ) ) {
```
hellofromtonya: I'm wondering if we shouldn't switch this around to check for the `post_status` that we want instead of what we don't want. 

What do I mean?

When do we want to trigger cleaning the post cache? There are [8 major post statuses](https://wordpress.org/support/article/post-status/): `publish`, `future`, `draft`, `pending`, `private`, `trash`, `auto-draft`, `inherit`, and also custom. _The custom one is the one that worries me._ 

What do you think @crystinutzaa?

crystinutzaa: The `custom` status is the real problem in here. My first idea was to bail out all posts which do not have `publish` post status. However I'm not sure if there could be any plugins which creates custom post statuses which are cached and should be cleaned on slug changes. So I've chosen the save way to exclude only `draft`, `auto-draft` and `pending`. Exactly as in `rocket_clean_post` function but with `pending` included:
https://github.com/wp-media/wp-rocket/blob/7723e789d158d2bf00ee5da49ef9158061e158eb/inc/common/purge.php#L81-L84

Probably there could be other exclusions for post statuses. 
Basically, this should check if the old post status was NOT published the old cache should not be cleaned.

hellofromtonya: >However I'm not sure if there could be any plugins which creates custom post statuses which are cached and should be cleaned on slug changes.

That's an interesting and very valid point. We don't know. A custom post status could be valid for caching or it may not be. We don't know. 

Hmm, this may be a larger discussion, ie beyond the scope of this PR. Let's put a pin in this particular topic. I'll open a separate ticket where we can discuss.

hellofromtonya:
@crystinutzaa How did you test that this PR fixes the 2 problems identified?

crystinutzaa:
There are 2 cases in here:
1. with  WooCommerce Follow-up Emails plugin
In order to test it go to:  Tools > Scheduled Actions > Pending and run any of the 2 actions listed
2. Publishing a new post causes the whole cache to be cleared.

hellofromtonya:
> There are 2 cases in here:
> 
> 1. with  WooCommerce Follow-up Emails plugin
>    In order to test it go to:  Tools > Scheduled Actions > Pending and run any of the 2 actions listed
> 2. Publishing a new post causes the whole cache to be cleared.

@crystinutzaa And you tested both cases. Right? What were the results?

crystinutzaa:
Yes, both are tested with this fix and is working fine in both cases.
1. WooCommerce Follow-up Emails plugin is prevented with this condition:

``
if ( empty( $post_name ) ) {
		return;
	}
``

2. Publishing a new post is prevented with the post_status condition
``if ( in_array( get_post_field( 'post_status', $post_id ), [ 'draft', 'pending', 'auto-draft' ], true ) ) {
		return;
	}``

inc/common/purge.php
Screenfeed: In case you need something similar someday, I made [this function](https://github.com/wp-media/imagify-plugin/blob/master/inc/functions/attachments.php#L59-L97) long ago in Imagify to get the public post statuses related to attachments. It uses [`get_post_stati()`](https://wpseek.com/function/get_post_stati/), that is handy to fetch statuses according our criteria.

PR #2173, opened 2019-12-11
Fixes #2139 Update logic on user actions purge

### LINKED ISSUE #2139: Stop clearing the entire cache when a user is added / updated / deleted

By default, we clear the entire cache when a user is created, updated and deleted. On some websites like e-commerce websites, it could be a problem.

Each time a new order is created by a new customer, it means the entire cache will be automatically cleared.

In fact it doesn't make sense to have this behavior. Until the customer isn't logged in, there is no info related to them on the default caching file version. So we don't have a valid reason to clear the entire cache for logged out visitors.

Don't clear the entire cache when a user is added / updated / deleted.

Clear the user cache if the "User Cache" option is turned on when a user is updated / deleted.

Tabrisrp:
- Removed the global purge on user creation/update/deletion
- Delete the cache for a specific user when it's updated/deleted, only if User cache option is active, and common logged-in users cache is not

This should improve cache lifetime on websites with frequent user creation (eCommerce especially), while still retaining the necessary cleaning of the user cache on sites with user-specific content like memberships or BuddyPress-based ones.

Fixes #2139

### DIFF
--- a/inc/classes/subscriber/Cache/PurgeActionsSubscriber.php
+++ b/inc/classes/subscriber/Cache/PurgeActionsSubscriber.php
@@ -1,0 +1,77 @@
+<?php
+namespace WP_Rocket\Subscriber\Cache;
+
+use WP_Rocket\Event_Management\Subscriber_Interface;
+use WP_Rocket\Admin\Options_Data;
+
+/**
+ * Subscriber for the cache purge actions
+ *
+ * @since 3.5
+ * @author Remy Perona
+ */
+class PurgeActionsSubscriber implements Subscriber_Interface {
+	/**
+	 * WP Rocket options instance
+	 *
+	 * @var Options_Data
+	 */
+	private $options;
+
+	/**
+	 * Constructor
+	 *
+	 * @param Options_Data $options WP Rocket options instance.
+	 */
+	public function __construct( Options_Data $options ) {
+		$this->options = $options;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public static function get_subscribed_events() {
+		return [
+			'profile_update' => 'purge_user_cache',
+			'deleted_user'   => 'purge_user_cache',
+		];
+	}
+
+	/**
+	 * Purges the cache of the corresponding user
+	 *
+	 * @since 3.5
+	 * @author Remy Perona
+	 *
+	 * @param int $user_id User ID.
+	 * @return void
+	 */
+	public function purge_user_cache( $user_id ) {
+		if ( ! $this->should_purge_user_cache() ) {
+			return;
+		}
+
+		rocket_clean_user( $user_id );
+	}
+
+	/**
+	 * Checks if the user cache should be purged
+	 *
+	 * @since 3.5
+	 * @author Remy Perona
+	 *
+	 * @return boolean
+	 */
+	private function should_purge_user_cache() {
+		if ( ! $this->options->get( 'cache_logged_user', 0 ) ) {
+			return false;
+		}
+
+		// This filter is documented in /inc/functions/files.php.
+		if ( apply_filters( 'rocket_common_cache_logged_users', false ) ) {
+			return false;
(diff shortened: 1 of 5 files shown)


tests/Unit/Subscriber/Cache/PurgeActionsSubscriber/TestPurgeUserCache.php
```diff
@@ -0,0 +1,80 @@
...
+namespace WP_Rocket\Tests\Unit\Subscriber\Cache\PurgeActionsSubscriber;
+
+use WP_Rocket\Subscriber\Cache\PurgeActionsSubscriber;
+use WP_Rocket\Tests\Unit\TestCase;
+use Brain\Monkey\Functions;
+use Brain\Monkey\Filters;
+
+/**
+ * @coversDefaultClass \WP_Rocket\Subscriber\Cache\PurgeActionsSubscriber
+ */
```
hellofromtonya: Add the `@group` tag here to apply it to all test methods in this class, ie instead of having it on each method.

tests/Unit/Subscriber/Cache/PurgeActionsSubscriber/TestPurgeUserCache.php
```diff
@@ -0,0 +1,80 @@
...
+class TestPurgeUserCache extends TestCase {
+	/**
+	 * @covers::purge_user_cache
+	 * @group purge_actions
+	 */
+	public function testShouldReturnNullWhenUserCacheDisabled() {
+		$options = $this->createMock('WP_Rocket\Admin\Options_Data');
+        $map     = [
```
hellofromtonya: @Tabrisrp Please auto run the code formatter to align code in this file for WPCS.

tests/Unit/Subscriber/Cache/PurgeActionsSubscriber/TestPurgeUserCache.php
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * @coversDefaultClass \WP_Rocket\Subscriber\Cache\PurgeActionsSubscriber
+ */
+class TestPurgeUserCache extends TestCase {
+	/**
+	 * @covers::purge_user_cache
+	 * @group purge_actions
+	 */
+	public function testShouldReturnNullWhenUserCacheDisabled() {
```
hellofromtonya: The method always returns `null` no matter what happens. I’d suggest renaming the test to something like: `testShouldNotPurgeUserCacheWhenUserCacheDisabled`. With this naming technique, it tells us what the expected behavior is when something happens.

tests/Unit/Subscriber/Cache/PurgeActionsSubscriber/TestPurgeUserCache.php
```diff
@@ -0,0 +1,80 @@
...
+
+		$purge = new PurgeActionsSubscriber( $options );
+		$this->assertNull($purge->purge_user_cache( 1 ));
+	}
+
+	/**
+	 * @covers::purge_user_cache
+	 * @group purge_actions
+	 */
+	public function testShouldReturnNullWhenCommonUserCache() {
```
hellofromtonya: Rename here to something like `testShouldNotPurgeUserCacheWhen...` and specify the when condition.

inc/classes/subscriber/Cache/PurgeActionsSubscriber.php
```diff
@@ -0,0 +1,77 @@
...
+	public function __construct( Options_Data $options ) {
+		$this->options = $options;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public static function get_subscribed_events() {
+		return [
+			'profile_update' => 'purge_user_cache',
```
hellofromtonya: Great job 👍 - only runs when user is updated or deleted.

PR #2179, opened 2019-12-12
Improves the integration test with reflection, @group, and overload valid key.

hellofromtonya:
This PR improves primarily the integration test suite by:

1. Adding reflection to provide granular control when you need to access a private or protected property or method within a test. Very handy.
2. Adds `@group` for targeted test runs.
3. Overloads the `rocket_valid_key()` to ensure all files are loaded when the plugin boots up for testing.
4. Fixes `TestIsSpeedTool` test using reflection to ensure `Config::$config_dir_path` is `null` when doing the test.

### DIFF
--- a/tests/Integration/Buffer/Tests/TestIsSpeedTool.php
+++ b/tests/Integration/Buffer/Tests/TestIsSpeedTool.php
@@ -5,6 +5,9 @@ use WP_Rocket\Tests\Integration\TestCase;
 use WP_Rocket\Buffer\Tests;
 use WP_Rocket\Buffer\Config;
 
+/**
+ * @group Buffer
+ */
 class TestIsSpeedTool extends TestCase {
      /**
      * @covers ::is_speed_tool
--- a/tests/Integration/Buffer/Tests/TestIsSpeedTool.php
+++ b/tests/Integration/Buffer/Tests/TestIsSpeedTool.php
@@ -11,7 +11,12 @@ class TestIsSpeedTool extends TestCase {
      * @author Remy Perona
      */
     public function testShouldReturnTrueWhenLighthouse() {
-        $config = new Config(
+    	// Grab the current Config::$config_dir_path value. We'll restore it when we're done.
+	    $config_dir_path = $this->get_reflective_property( 'config_dir_path', 'WP_Rocket\Buffer\Config' );
+	    // Set the Config::$config_dir_path value to `null`.
+	    $this->set_reflective_property( null, 'config_dir_path', 'WP_Rocket\Buffer\Config' );
+
+	    $config = new Config(
             [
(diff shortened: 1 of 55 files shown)


tests/Integration/bootstrap.php
```diff
@@ -52,6 +52,10 @@ function bootstrap_integration_suite( $wp_tests_dir ) {
 	tests_add_filter(
 		'muplugins_loaded',
 		function() {
+			// Overload the license key for testing.
+			\Patchwork\redefine( 'rocket_valid_key', '__return_true' );
```
hellofromtonya: @Tabrisrp @Screenfeed @crystinutzaa This code is important to ensure all files are loading in `inc/main.php` when the `'plugins_loaded'` event fires, ie when the plugin boots up. 

Why do we need to do it here? Why not within the Test Case or as needed within a test itself?

The loading happens _before_ the tests run. It's during the boot up of WordPress and the plugin. So we need to overload the function here to ensure those files load as they would for our customers.

tests/Integration/Buffer/Tests/TestIsSpeedTool.php
```diff
@@ -5,13 +5,21 @@
...
      /**
      * @covers ::is_speed_tool
      * @author Remy Perona
      */
     public function testShouldReturnTrueWhenLighthouse() {
-        $config = new Config(
+    	// Grab the current Config::$config_dir_path value. We'll restore it when we're done.
+	    $config_dir_path = $this->get_reflective_property( 'config_dir_path', 'WP_Rocket\Buffer\Config' );
+	    // Set the Config::$config_dir_path value to `null`.
+	    $this->set_reflective_property( null, 'config_dir_path', 'WP_Rocket\Buffer\Config' );
```
hellofromtonya: When the plugin boots up, the `Config::$config_dir_path` is set statically. When we go to run this test, it's not using the configuration below, but rather what was set on boot up. Okay, that's normal for production. 

How do we handle it here? With reflection.

The trait I added gives us the ability to access private and protected properties. Aha. That's really handy in cases like this one.

Here:

1. Grab the value that's been set in the static property.
2. Set the property to `null`, which is what we need for this test.
3. Run the test.
4. Restore the original value.

Tabrisrp: Does it mean that the test was looking fine before, but in fact it was not really reliable?

hellofromtonya: Yes.

It was working until:

- the plugin loaded as it does in production
- we built another integration test that preloaded the config. If the new test ran first, then the speed tool test would fail

Using the above strategy, we eliminated any race/timing issues and ensured the state is what we need specifically for the context of this one test and edge case.

PR #2180, opened 2019-12-13
Fixed: clean expired cache scheduled event when cron unit is set to minutes

crystinutzaa:
Cleaning the scheduled event from clean expired cache when cron unit is set to minutes.
For hours & days the scheduled event should be set to 1 hour, but for minutes should be set to those exact minutes.
I am wondering if the scheduled event should be cleaned also when switching from minutes to hour or days?

### DIFF
--- a/inc/classes/subscriber/Cache/class-expired-cache-purge-subscriber.php
+++ b/inc/classes/subscriber/Cache/class-expired-cache-purge-subscriber.php
@@ -67,11 +67,36 @@ class Expired_Cache_Purge_Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'init'                => 'schedule_event',
-			'rocket_deactivation' => 'unschedule_event',
-			static::EVENT_NAME    => 'purge_expired_files',
-			'cron_schedules'      => 'custom_cron_schedule',
+			'init'                            => 'schedule_event',
+			'rocket_deactivation'             => 'unschedule_event',
+			static::EVENT_NAME                => 'purge_expired_files',
+			'cron_schedules'                  => 'custom_cron_schedule',
+			'update_option_' . WP_ROCKET_SLUG => [ 'clean_expired_cache_scheduled_event', 10, 2 ],
 		];
+	}
+
+	/**
+	 * Clean expired cache scheduled event when Lifespan is changed to minutes.
+	 *
+	 * @since  3.4.3
+	 * @author Soponar Cristina
+	 *
+	 * @param array $old_value An array of previous values for the settings.
+	 * @param array $value     An array of submitted values for the settings.
+	 */
+	public function clean_expired_cache_scheduled_event( $old_value, $value ) {
+		if ( empty( $value['purge_cron_unit'] ) ) {
+			return;
+		}
+		// If Lifespan unit is not changed to minutes do not reschedule (for hours & days event should be scheduled at 1 hour).
+		if ( 'MINUTE_IN_SECONDS' !== $value['purge_cron_unit'] ) {
+			return;
+		}
+		// If previous Lifespan unit was minutes do not reschedule.
+		if ( $old_value['purge_cron_unit'] === $value['purge_cron_unit'] ) {
+			return;
+		}
+		$this->unschedule_event();
 	}
 
 	/**
--- a/tests/Integration/Subscriber/ExpiredCachePurgeSubscriber/TestCleanCacheScheduledEvent.php
+++ b/tests/Integration/Subscriber/ExpiredCachePurgeSubscriber/TestCleanCacheScheduledEvent.php
@@ -1,0 +1,95 @@
+<?php
+namespace WP_Rocket\Tests\Integration\Subscriber\ExpiredCachePurgeSubscriber;
+
+use PHPUnit\Framework\TestCase;
+use WP_Rocket\Admin\Options_Data;
+use WP_Rocket\Admin\Options;
+use WP_Rocket\Cache\Expired_Cache_Purge;
+use WP_Rocket\Subscriber\Cache\Expired_Cache_Purge_Subscriber;
+
+class TestCleanCacheScheduledEvent extends TestCase {
+	public function testShouldNotCleanScheduledEventWhenValueIsDifferenThanMinutes() {
+		update_option(
+			'wp_rocket_settings',
+			[
+				'purge_cron_interval' => 10,
+				'purge_cron_unit'     => 'HOUR_IN_SECONDS',
+			]
+		);
+
+		$old_value = [
+			'purge_cron_interval' => 10,
+			'purge_cron_unit'     => 'HOUR_IN_SECONDS',
+		];
+		$value    = [
+			'purge_cron_interval' => 10,
+			'purge_cron_unit'     => 'HOUR_IN_SECONDS',
+		];
+		$options                        = new Options_Data( (new Options( 'wp_rocket_'))->get( 'settings' ) );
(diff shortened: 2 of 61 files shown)


inc/classes/subscriber/Cache/class-expired-cache-purge-subscriber.php
```diff
@@ -67,13 +67,38 @@ public function __construct( Options_Data $options, Expired_Cache_Purge $purge )
...
 		return [
-			'init'                => 'schedule_event',
-			'rocket_deactivation' => 'unschedule_event',
-			static::EVENT_NAME    => 'purge_expired_files',
-			'cron_schedules'      => 'custom_cron_schedule',
+			'init'                            => 'schedule_event',
+			'rocket_deactivation'             => 'unschedule_event',
+			static::EVENT_NAME                => 'purge_expired_files',
+			'cron_schedules'                  => 'custom_cron_schedule',
+			'update_option_' . WP_ROCKET_SLUG => [ 'clean_expired_cache_scheduled_event', 10, 2 ],
```
hellofromtonya: @crystinutzaa For timing, does it matter when this callback fires in relationship to any 3rd party or other callbacks? 

What do I mean?

Do we need it to run:

- before everything else, ie run first? If yes, then the priority would be `0`.
- after everything else, ie run last? If yes, then the priority would be `PHP_INT_MAX`.

If it doesn't matter (ie there's not a potential of timing), then a priority of `10` is fine.

What do you think?

tests/Integration/Subscriber/ExpiredCachePurgeSubscriber/TestCleanCacheScheduledEvent.php
```diff
@@ -0,0 +1,95 @@
+<?php
+namespace WP_Rocket\Tests\Integration\Subscriber\ExpiredCachePurgeSubscriber;
+
+use PHPUnit\Framework\TestCase;
```
hellofromtonya: Change to [`WP_Rocket\Tests\Integration\TestCase`](https://github.com/wp-media/wp-rocket/blob/develop/tests/Integration/TestCase.php), which requires `develop` to be merged into this branch.

tests/Integration/Subscriber/ExpiredCachePurgeSubscriber/TestCleanCacheScheduledEvent.php
```diff
@@ -0,0 +1,95 @@
...
+		];
+		$value    = [
+			'purge_cron_interval' => 10,
+			'purge_cron_unit'     => 'HOUR_IN_SECONDS',
+		];
+		$options                        = new Options_Data( (new Options( 'wp_rocket_'))->get( 'settings' ) );
+		$expired_cache_purge_subscriber = new Expired_Cache_Purge_Subscriber( $options, new Expired_Cache_Purge( WP_ROCKET_CACHE_PATH ) );
+
+		$expired_cache_purge_subscriber->schedule_event();
+		$expired_cache_purge_subscriber->clean_expired_cache_scheduled_event( $old_value, $value );
```
hellofromtonya: For an integration test, we want to validate that this method is a callback that fires automatically when the option is updated. Instead of directly invoking this method, update the option again and valid the expected behavior.

tests/Integration/Subscriber/ExpiredCachePurgeSubscriber/TestCleanCacheScheduledEvent.php
```diff
@@ -0,0 +1,95 @@
+<?php
+namespace WP_Rocket\Tests\Integration\Subscriber\ExpiredCachePurgeSubscriber;
+
+use PHPUnit\Framework\TestCase;
+use WP_Rocket\Admin\Options_Data;
+use WP_Rocket\Admin\Options;
+use WP_Rocket\Cache\Expired_Cache_Purge;
+use WP_Rocket\Subscriber\Cache\Expired_Cache_Purge_Subscriber;
+
+class TestCleanCacheScheduledEvent extends TestCase {
```
hellofromtonya: Add `@group Subscriber` to the class' DocBlock for targeted test running.

tests/Integration/Subscriber/ExpiredCachePurgeSubscriber/TestCleanCacheScheduledEvent.php
```diff
@@ -0,0 +1,95 @@
...
+		];
+		$value    = [
+			'purge_cron_interval' => 10,
+			'purge_cron_unit'     => 'MINUTE_IN_SECONDS',
+		];
+		$options                        = new Options_Data( (new Options( 'wp_rocket_'))->get( 'settings' ) );
+		$expired_cache_purge_subscriber = new Expired_Cache_Purge_Subscriber( $options, new Expired_Cache_Purge( WP_ROCKET_CACHE_PATH ) );
+
+		$expired_cache_purge_subscriber->schedule_event();
+		$expired_cache_purge_subscriber->clean_expired_cache_scheduled_event( $old_value, $value );
```
hellofromtonya: Same note here as above, ie update the option for the values needed for this test case.

tests/Integration/Subscriber/ExpiredCachePurgeSubscriber/TestCleanCacheScheduledEvent.php
```diff
@@ -0,0 +1,95 @@
...
+		$old_value = [
+			'purge_cron_interval' => 10,
+			'purge_cron_unit'     => 'HOUR_IN_SECONDS',
+		];
+		$value    = [
+			'purge_cron_interval' => 10,
+			'purge_cron_unit'     => 'MINUTE_IN_SECONDS',
+		];
+		$options                        = new Options_Data( (new Options( 'wp_rocket_'))->get( 'settings' ) );
+		$expired_cache_purge_subscriber = new Expired_Cache_Purge_Subscriber( $options, new Expired_Cache_Purge( WP_ROCKET_CACHE_PATH ) );
```
hellofromtonya: Same note here as above, ie update the option for the values needed for this test case.

inc/classes/subscriber/Cache/class-expired-cache-purge-subscriber.php
crystinutzaa: @hellofromtonya this should not have any impact to any 3rd party or other callbacks. Basically it will simply reschedule the deletion of WP Rocket cache to the new lifespan.

tests/Integration/Subscriber/ExpiredCachePurgeSubscriber/TestCleanCacheScheduledEvent.php
```diff
@@ -0,0 +1,80 @@
...
+		);
+
+		$this->assertTrue( true ); // Prevent "risky" warning.
+	}
+
+	public function testShouldCleanScheduledEventWhenMinutesAndOldValueIsHours() {
+		Functions\expect( 'wp_clear_scheduled_hook' )
+			->once()
+			->with( Expired_Cache_Purge_Subscriber::EVENT_NAME )
+			->andReturnNull(); // No need to run it.
```
hellofromtonya: @crystinutzaa Just to make sure it's called for this test, I checked the arg with `with`. And to prevent it from running, I added `andReturnNull()`, ie we don't need `wp_clear_scheduled_hook` to run; rather, we just want to know that we made it this far to validate this test works as expected.

hellofromtonya:
@crystinutzaa As we discussed in Slack, the cherry picks needed to be done in reverse order as 6d39bba3478f290245ec5577cefc2339a85a9ab0 one depends on b58f5a047d8f26c3d7e5d4d4bf13fae08abadb57. Sorry I didn't specify that when we talked early. I went ahead and rolled it, reversed the order, and force pushed.

PR #2181, opened 2019-12-13
Added Home and Feed URLs to Cloudflare clean cache API call

crystinutzaa:
Cleaning Cloudflare cache when a post cache is cleaned needed also the home URL and the feeds URLs

## TODO
- [x] Potentially remove notices (because there is a report for some notice which still appears :)). Waiting on feedback from Lucy/Arun.

### DIFF
--- a/inc/functions/cloudflare.php
+++ b/inc/functions/cloudflare.php
@@ -691,21 +691,6 @@ function rocket_auto_purge_cloudflare() {
 
 	// Purge CloudFlare.
 	$cf_purge = rocket_purge_cloudflare();
-
-	if ( is_wp_error( $cf_purge ) ) {
-		$cf_purge_result = [
-			'result'  => 'error',
-			// translators: %s = CloudFare API return message.
-			'message' => sprintf( __( 'Cloudflare cache purge error: %s', 'rocket' ), $cf_purge->get_error_message() ),
-		];
-	} else {
-		$cf_purge_result = [
-			'result'  => 'success',
-			'message' => __( 'Cloudflare cache successfully purged', 'rocket' ),
-		];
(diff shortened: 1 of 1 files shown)


inc/functions/cloudflare.php
```diff
@@ -737,6 +737,22 @@ function rocket_auto_purge_cloudflare_by_url( $post, $purge_urls, $lang ) {
...
+	$feed_urls    = [];
+	$feed_urls[]  = get_feed_link();
+	$feed_urls[]  = get_feed_link( 'comments_' );
+
+	/**
+	 * Filter the home feeds urls
+	 *
+	 * @since 2.7
+	 * @param array     $urls The urls of the home feeds.
+	*/
```
Tabrisrp: since the filter already exists somewhere else, you can reference it that way instead:
`// this filter is documented in inc/functions/files.php`

crystinutzaa: Applied the same change in v3.5 refactored code.

inc/functions/cloudflare.php
```diff
@@ -737,6 +737,17 @@ function rocket_auto_purge_cloudflare_by_url( $post, $purge_urls, $lang ) {
 		return;
 	}
 
+	// Add home URL and feeds URLs to Cloudflare clean cache URLs list.
+	$home_url     = get_rocket_i18n_home_url( $lang );
+	$purge_urls[] = $home_url;
```
hellofromtonya: As `$home_url` is not used elsewhere in this function, we can combine these 2 lines of code:

```php
$purge_urls[] = get_rocket_i18n_home_url( $lang );
```

PR #2183, opened 2019-12-13
Fixed: Cloudflare Auto Purge By Url add home and feed urls to purge urls list

crystinutzaa:
Cleaning Cloudflare cache when a post cache is cleaned needed also the home URL and the feeds URLs

## TODO
- [x] Potentially remove notices (because there is a report for some notice which still appears :)). Waiting on feedback from Lucy/Arun.

### DIFF
--- a/inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
+++ b/inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
@@ -203,12 +203,7 @@ class CloudflareSubscriber implements Subscriber_Interface {
 		$feed_urls[]  = get_feed_link();
 		$feed_urls[]  = get_feed_link( 'comments_' );
 
-		/**
-		 * Filter the home feeds urls
-		 *
-		 * @since 2.7
-		 * @param array     $urls The urls of the home feeds.
-		*/
+		// this filter is documented in inc/functions/files.php.
 		$feed_urls  = apply_filters( 'rocket_clean_home_feeds', $feed_urls );
 		$purge_urls = array_unique( array_merge( $purge_urls, $feed_urls ) );
 
--- a/inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
+++ b/inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
@@ -196,6 +196,21 @@ class CloudflareSubscriber implements Subscriber_Interface {
 		if ( is_wp_error( $cf_cache_everything ) || ! $cf_cache_everything ) {
 			return;
 		}
+		// Add home URL and feeds URLs to Cloudflare clean cache URLs list.
+		$home_url     = get_rocket_i18n_home_url( $lang );
+		$purge_urls[] = $home_url;
+		$feed_urls    = [];
+		$feed_urls[]  = get_feed_link();
+		$feed_urls[]  = get_feed_link( 'comments_' );
+
+		/**
+		 * Filter the home feeds urls
+		 *
+		 * @since 2.7
+		 * @param array     $urls The urls of the home feeds.
+		*/
+		$feed_urls  = apply_filters( 'rocket_clean_home_feeds', $feed_urls );
+		$purge_urls = array_unique( array_merge( $purge_urls, $feed_urls ) );
 
 		// Purge CloudFlare.
 		$cf_purge = $this->cloudflare->purge_by_url( $post, $purge_urls, $lang );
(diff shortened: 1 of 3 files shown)


tests/Unit/Subscriber/Addons/CloudflareSubscriber/TestAutoPurgeByUrl.php
```diff
@@ -75,6 +75,8 @@ public function testShouldAutoPurgeByUrlWhenCacheEverythingButReturnError() {
 		$mocks = $this->getConstructorMocks( 1, '', '', '' );
 		Functions\when( 'current_user_can' )->justReturn( true );
 		Functions\when( 'get_current_user_id' )->justReturn( 1 );
+		Functions\when( 'get_rocket_i18n_home_url' )->justReturn( 'http://local.com/' );
+		Functions\when( 'get_feed_link' )->justReturn( 'http://local.com/feed/' );
```
Tabrisrp: Using `example.org` as the placeholder domain would be better:
- `local.com` is a real URL
- `example.org` is a domain registered specifically for testing purpose
- `example.org` is the default domain used by the WP test suite for integration test

tests/Unit/Subscriber/Addons/CloudflareSubscriber/TestAutoPurgeByUrl.php
```diff
@@ -95,6 +97,8 @@ public function testShouldAutoPurgeByUrlWhenCacheEverything() {
 		Functions\when( 'current_user_can' )->justReturn( true );
 		Functions\when( 'get_current_user_id' )->justReturn( 1 );
 		Functions\when( 'is_wp_error' )->justReturn( false );
+		Functions\when( 'get_rocket_i18n_home_url' )->justReturn( 'http://local.com/' );
+		Functions\when( 'get_feed_link' )->justReturn( 'http://local.com/feed/' );
```
Tabrisrp: Same changes to make here

inc/classes/subscriber/Addons/Cloudflare/CloudflareSubscriber.php
```diff
@@ -196,6 +196,16 @@ public function auto_purge_by_url( $post, $purge_urls, $lang ) {
 		if ( is_wp_error( $cf_cache_everything ) || ! $cf_cache_everything ) {
 			return;
 		}
+		// Add home URL and feeds URLs to Cloudflare clean cache URLs list.
+		$home_url     = get_rocket_i18n_home_url( $lang );
+		$purge_urls[] = $home_url;
```
hellofromtonya: As `$home_url` is not used elsewhere in this function, we can combine these 2 lines of code:

```php
$purge_urls[] = get_rocket_i18n_home_url( $lang );
```

tests/Unit/Subscriber/Addons/CloudflareSubscriber/TestAutoPurgeByUrl.php
hellofromtonya: We expect both of these functions to be invoked, ie both will run as part of this test. Let's change `when` (which means it may or may not run) to `expect` (meaning we "expect" it to run `once`, `with` these arguments, and then do something when done.

tests/Unit/Subscriber/Addons/CloudflareSubscriber/TestAutoPurgeByUrl.php
```diff
@@ -95,11 +99,14 @@ public function testShouldAutoPurgeByUrlWhenCacheEverything() {
...
+		Functions\expect( 'get_rocket_i18n_home_url' )->once()->with( '' )->andReturn( 'http://example.org/' );
+		Functions\expect( 'get_feed_link' )->twice()->andReturn( 'http://example.org/feed/', 'http://example.org/feed/comments' );
 
-		$wp_error   = \Mockery::mock( \WP_Error::class );
 		$cloudflare = \Mockery::mock( \WP_Rocket\Addons\Cloudflare\Cloudflare::class );
-		$cloudflare->shouldReceive('has_page_rule')->andReturn( true );
-		$cloudflare->shouldReceive('purge_by_url')->andReturn( true );
+		$cloudflare->shouldReceive('has_page_rule')->with( 'cache_everything' )->andReturn( true );
+		$cloudflare->shouldReceive('purge_by_url')
+		           ->with( 1, [ '/hello-world', 'http://example.org/', 'http://example.org/feed/', 'http://example.org/feed/comments' ], '' )
```
hellofromtonya: @crystinutzaa I added argument checks to further validate the behavior. Notice how we can test what we expect to be in `$purge_urls` using this approach.

crystinutzaa: @hellofromtonya Indeed the test is way much clear and `$purge_urls` can be tested.

PR #2193, opened 2019-12-16
Fix #2192 - Disabling Opcache using php.ini doesn't hide our "Clear OPcache" menu entry

### LINKED ISSUE #2192: Disabling Opcache using php.ini doesn't hide our "Clear OPcache" menu entry
When disabling OPcache using a **php.ini** file with the following line:

```
opcache.enable = Off 
```
it seems that when opcache is disabled this way, the function `opcache_reset` is still available. 
We use this function to check if opcache is available or not, and display the menu link:
https://github.com/wp-media/wp-rocket/blob/17bbc04a0f1e925ecd9e4f3e1c8f06a2e1037db3/views/settings/page-sections/dashboard.php#L152

So, for these cases, the **Clear Opcache** option will still be visible at the user admin panel, and when clicked it will produce the error: **OPcache Purgue failed**. 

**Related ticket:**
https://secure.helpscout.net/conversation/1017198190/133573?folderId=2683093

**Slack thread:**
https://wp-media.slack.com/archives/C08N8J6VC/p1576087910073600

crystinutzaa:
When disabling OPcache using a **php.ini** file with the following line:

```
opcache.enable = Off 
```
it seems that when opcache is disabled this way, the function `opcache_reset` is still available. 
We use this function to check if opcache is available or not, and display the menu link:
https://github.com/wp-media/wp-rocket/blob/17bbc04a0f1e925ecd9e4f3e1c8f06a2e1037db3/views/settings/page-sections/dashboard.php#L152

So, for these cases, the **Clear Opcache** option will still be visible at the user admin panel, and when clicked it will produce the error: **OPcache Purgue failed**. 

Reference Issue Ticket #2192


Screenfeed:
If opcache is disabled that way, shouldn’t `ini_get( 'opcache.enable' )` return `Off`? If this is the case, `if ( 'Off' ) {` will pass.

hellofromtonya:
>If opcache is disabled that way, shouldn’t ini_get( 'opcache.enable' ) return Off? If this is the case, if ( 'Off' ) { will pass.

Excellent catch @Screenfeed! Well done 🥇 . 

You're correct. This code needs to explicitly check for what value it needs.

Screenfeed:
Then this should do the trick:
```php
$opcache_disabled = ! $opcache_enabled || strtolower( (string) $opcache_enabled ) === 'off';
```

crystinutzaa:
> If opcache is disabled that way, shouldn’t `ini_get( 'opcache.enable' )` return `Off`? If this is the case, `if ( 'Off' ) {` will pass.

In PHP documentation related to ini_get there is this note:

> A boolean ini value of off will be returned as an empty string or "0" while a boolean ini value of on will be returned as "1".

inc/common/admin-bar.php
```diff
@@ -174,13 +174,14 @@ function rocket_admin_bar( $wp_admin_bar ) {
 		/**
 		 * Purge OPCache content if OPcache is active.
 		 */
+		$opcache_enabled  = ini_get( 'opcache.enable' );
```
hellofromtonya: [Per the PHP Manual for `ini_get()`](https://www.php.net/manual/en/function.ini-get.php#refsect1-function.ini-get-notes):
>Note: When querying boolean values
>A boolean ini value of off will be returned as an empty string or "0" while a boolean ini value of on will be returned as "1". The function can also return the literal string of INI value.

What does this mean?

For values, it returns:

- Enabled: `'1'` or `'On'` => returns `'1'` of type `string`
- Disabled:  `''`, `'0'`, or `'Off'` => returns `''` (empty string) of type `string`

Notice the type returned is a `string` and not a `boolean`.

In PHP 5 for conditional expressions, we're used to PHP type casting a `string` type into a `boolean`. However, with PHP 7, we can now be strict and declarative with our data types. For example, here we could ensure `$opcache_enabled` has a `boolean` data type.

Why? What's the point? Our variables represent data or a value. By setting the data type, we are being more _purposeful_ and _intentional_ in our code. It becomes more understandable and readable for us humans. And in some cases, it's a tiny bit more performant as PHP doesn't have to do the type casting when evaluating the value/data within a conditional expression.

Here and in each cause where we expect a boolean value:

```php
$opcache_enabled  = (bool) ini_get( 'opcache.enable' );
```



How can 

A better approach here is to `ini_get( 'opcache.enable' )`

hellofromtonya: What if the ini setting is invalid and not what's expected? What if there's a typo such as ``10`` when it should be `'1'`? Then using `(bool)` would return `true`. Whoopsie.

How do we guard against typos or incorrect enable/disable values in the php.ini? 

We can use [`filter_var`](https://www.php.net/manual/en/function.filter-var.php) with the [`FILTER_VALIDATE_BOOLEAN` filter](https://www.php.net/manual/en/filter.filters.validate.php):

```php
$opcache_enabled = filter_var( ini_get( 'opcache.enable' ), FILTER_VALIDATE_BOOLEAN );
```
What happens here?

- Gets the setting value
- Filters it through the `FILTER_VALIDATE_BOOLEAN` filter
      - Returns `TRUE` if the setting is `'1'`, '`true'`, `'on'`, or `'yes'`
      - Else, returns 'FALSE` (for boolean and non-boolean values)

PR #2194, opened 2019-12-17
[WIP] Prevent multiple calls to Cloudflare IPs API call

crystinutzaa:
Prevent WP Rocket to make multiple calls to Cloudflare IPs API call.
If the call to /ips did not received success then an Exception was thrown. 
Instead it should return the default ips from Cloudflare and set the transient.

### DIFF
--- a/inc/functions/cloudflare.php
+++ b/inc/functions/cloudflare.php
@@ -518,53 +518,68 @@ function rocket_get_cloudflare_ips() {
 	}
 
 	$cf_ips = get_transient( 'rocket_cloudflare_ips' );
+
 	if ( false === $cf_ips ) {
 		try {
 			$cf_ips_instance = new Cloudflare\IPs( $cf_instance );
 			$cf_ips          = $cf_ips_instance->ips();
-
 			if ( ! isset( $cf_ips->success ) || ! $cf_ips->success ) {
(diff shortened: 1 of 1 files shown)


inc/functions/cloudflare.php
```diff
@@ -518,48 +518,18 @@ function rocket_get_cloudflare_ips() {
...
 	$cf_ips = get_transient( 'rocket_cloudflare_ips' );
+
 	if ( false === $cf_ips ) {
 		try {
 			$cf_ips_instance = new Cloudflare\IPs( $cf_instance );
 			$cf_ips          = $cf_ips_instance->ips();
-
 			if ( ! isset( $cf_ips->success ) || ! $cf_ips->success ) {
-				throw new Exception( 'Error connecting to Cloudflare' );
+				$cf_ips = rocket_get_default_cloudflare_ips();
```
hellofromtonya: >Recapping @crystinutzaa and my conversation on a Zoom call:

This new code has the same end behavior as the original.

When we don't get a success back from CF: here's what happens:
- get the defaults (object)
- set the transient
- return the object

The only difference is: no exception is explicitly thrown.

We both think this isn't contributing to the multiple calls to the CF API problem.

PR #2197, opened 2019-12-17
[WIP] Add phpstan to our testing suite

Tabrisrp:
First step of the work: initial setup
- Require the necessary packages
- Add a composer script to run the analyze command
- Initialize a config file (based on https://github.com/szepeviktor/phpstan-wordpress/blob/master/example/phpstan.neon.dist)
- Initialize a bootstrap file

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,9 @@
 		"dealerdirect/phpcodesniffer-composer-installer": "^0.5.0",
 		"mikey179/vfsstream": "^1.6",
 		"phpcompatibility/phpcompatibility-wp": "^2.0",
+		"phpstan/phpstan": "^0.12.3",
 		"phpunit/phpunit": "^5.7 || ^7",
+		"szepeviktor/phpstan-wordpress": "^0.5.0",
 		"wp-coding-standards/wpcs": "^2"
 	},
 	"autoload": {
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -46,7 +46,9 @@
 		"dealerdirect/phpcodesniffer-composer-installer": "^0.5.0",
 		"mikey179/vfsstream": "^1.6",
 		"phpcompatibility/phpcompatibility-wp": "^2.0",
+		"phpstan/phpstan": "^0.12.3",
```
hellofromtonya: Note: `phpstan/phpstan` and `szepeviktor/phpstan-wordpress` require a minimum of PHP 7.1. The Travis builds are failing for 7.0 and 5.6 due to this.

Tabrisrp: Indeed I noticed that yesterday. We would not need it to run on all builds anyway, do you have an idea on how to work around that?

hellofromtonya: I don't know of a way to do it here. Instead, we'd handle it with Travis, i.e. either in a separate bash script or directly in existing yaml file.

Tabrisrp: For now I removed the need to install phpstan on Travis, since we're not yet ready to have it as part of our CI. In the future we can dedicate a build to run it

Tabrisrp:
Closing in favour of #2236

PR #2199, opened 2019-12-17
Upgrades test suites to use our phpunit package & virtual filesystem.

hellofromtonya:
Upgrades the test suite to:

- Use [`wp-media/phpunit`](https://github.com/wp-media/phpunit).
- Use the virtual filesystem included in `wp-media/phpunit`.
- Use `rocket_has_constant` and `rocket_get_constant` to never define a constant in the tests and allow mocking in the tests.

## TODO

- [x] Add `wp-media/phpunit` 
- [x] Update `wp-media/phpunit` to released version or `dev-master` once the [Virtual Filesystem PR 1](https://github.com/wp-media/phpunit/pull/1) is merged into `master`
- [x] Change `.travis.yml` back to run `composer run-tests` once the integration tests are upgraded.
- [x] Upgrade the unit tests.
- [x] Change `Unit/Cache/TestPurgeExpiredFiles.php` to be a unit test, i.e. mock all dependencies.
- [x] Upgrade the integration tests.


tests/Integration/Subscriber/ExpiredCachePurgeSubscriber/TestCleanCacheScheduledEvent.php
```diff
@@ -27,8 +30,6 @@ public function testShouldNotCleanScheduledEventWhenValuesAreTheSame() {
 				'purge_cron_unit'     => 'HOUR_IN_SECONDS',
 			]
 		);
-
-		$this->assertTrue( true ); // Prevent "risky" warning.
```
Screenfeed: Why remove this?

hellofromtonya: It's not testing anything. [`Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration`](https://github.com/wp-media/phpunit/blob/master/Integration/TestCase.php#L12), included in the base `TestCase`, eliminates needing the "risky" warning.

Screenfeed: Removing it won't trigger an error saying that the test does no assert?

hellofromtonya: Nope. Why? There is an assertion in the test.

```php
Functions\expect( 'wp_clear_scheduled_hook' )->never();
```

Screenfeed: I've seen several times where this would not be considered as an assertion, and the error would trigger.
Maybe I did something wrong at the time, or maybe it was a bug in an old version of BrainMonkey/Mockery (I'm not sure but I think I remember reading something about this). Also, I don't think I am the author of this, so I'm guessing I'm not the only one experiencing this 🤔 
Anyway, let's keep them removed, and see if there are error messages.
Thanks Tonya.

hellofromtonya: Before the `Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration` was not used in the base Integration Test Case. Without it, an error would trigger with the "risky" warning.

Screenfeed: OK, understood, thanks.

PR #2201, opened 2019-12-18
Fixed: clean expired cache scheduled event when cron unit is set to minutes

crystinutzaa:
Cleaning the scheduled event from clean expired cache when cron unit is set to minutes.
For hours & days the scheduled event should be set to 1 hour, but for minutes should be set to those exact minutes.

My latest changes got removed from the last PR #2180

Arun pointed out that when changing from Minutes to Minutes and if the interval is changes the scheduled event should be cleaned. Added the code + Integration tests for it.

### DIFF
--- a/inc/classes/subscriber/Cache/class-expired-cache-purge-subscriber.php
+++ b/inc/classes/subscriber/Cache/class-expired-cache-purge-subscriber.php
@@ -88,12 +88,31 @@ class Expired_Cache_Purge_Subscriber implements Subscriber_Interface {
 		if ( empty( $value['purge_cron_unit'] ) ) {
 			return;
 		}
-		// If Lifespan unit is not changed to minutes do not reschedule (for hours & days event should be scheduled at 1 hour).
-		if ( 'MINUTE_IN_SECONDS' !== $value['purge_cron_unit'] ) {
+
+		$unit_list = [ 'HOUR_IN_SECONDS', 'DAY_IN_SECONDS' ];
+		// Bail out if the cron unit is changed from hours to days.
+		if ( $old_value['purge_cron_unit'] === $value['purge_cron_unit'] &&
+				in_array( $old_value['purge_cron_unit'], $unit_list, true ) &&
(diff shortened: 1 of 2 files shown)


arunbasillal:
Related to #2168

inc/classes/subscriber/Cache/class-expired-cache-purge-subscriber.php
```diff
@@ -88,12 +88,31 @@ public function clean_expired_cache_scheduled_event( $old_value, $value ) {
...
 			return;
 		}
-		// If Lifespan unit is not changed to minutes do not reschedule (for hours & days event should be scheduled at 1 hour).
-		if ( 'MINUTE_IN_SECONDS' !== $value['purge_cron_unit'] ) {
+
+		$unit_list = [ 'HOUR_IN_SECONDS', 'DAY_IN_SECONDS' ];
+		// Bail out if the cron unit is changed from hours to days.
+		if ( $old_value['purge_cron_unit'] === $value['purge_cron_unit'] &&
+				in_array( $old_value['purge_cron_unit'], $unit_list, true ) &&
+				in_array( $value['purge_cron_unit'], $unit_list, true ) ) {
```
Screenfeed: One of those `in_array()` is useless.

crystinutzaa: Why do you think is useless?

Screenfeed: If A = B and A is in C, then B is in C.

crystinutzaa: Indeed, this condition should be totally removed because the next ones prevent this.
I will make a PR with origin:develop branch

Screenfeed: Do like you want, it doesn't hurt, it can stay here until there is something else to do.

PR #2208, opened 2019-12-24
Fix  #2149 - Error 403 on admin-ajax requests and WooCommerce nonces

### LINKED ISSUE #2149: Error 403 on admin-ajax requests and WooCommerce nonces
This is a copy of what @vmanthos reported here: https://hub.wp-media.me/wp-rocket/knowledge-base/issues/error-403-on-admin-ajax-requests-and-woocommerce-nonces/

In some cases, on WooCommerce websites with plugins, e.g. WooCommerce Mailchimp Discount, that use AJAX requests (admin-ajax.php), those fail with an error 403.

This is related to nonces and the way, when WooCommerce is installed, they are produced.

The issue has to do with the fact that WooCommerce uses the following filter, which is defined in the wp_create_nonce() function:

`$uid = apply_filters( 'nonce_user_logged_out', $uid, $action );`

When a user is logged out, WooCommerce sets the user ID for the non-logged in user to be the session ID, if one is set.

When WP Rocket generates the cache files, nonces are generated without any UserID (uid).

So, they are different than the ones WooCommerce creates. That’s why, when check_ajax_referer() is used in a plugin or theme, it fails during nonce validation.

The issue is also described here:

- https://github.com/CalderaWP/Caldera-Forms/issues/894
- https://codeclimate.com/github/skyverge/wc-plugin-framework/woocommerce/utilities/class-sv-wp-background-job-handler.php/source (lines 153-156)

crystinutzaa:
In some cases, on WooCommerce websites with plugins, e.g. WooCommerce Mailchimp Discount, that use AJAX requests (admin-ajax.php), those fail with an error 403.

This is related to nonces and the way, when WooCommerce is installed, they are produced.

The issue has to do with the fact that WooCommerce uses the following filter, which is defined in the wp_create_nonce() function:

$uid = apply_filters( 'nonce_user_logged_out', $uid, $action );

When a user is logged out, WooCommerce sets the user ID for the non-logged in user to a random one generated with Password Hash.

We identified 2 nonces till now:
- "Load more" AJAX functionality of the Newspaper theme is affected.
- WooCommerce MailChimp Discount.

This should be restricted to WooCommerce only, but in Integration Tests I receive an error from patchwork:
`Patchwork\Exceptions\NotUserDefined: Please include {"redefinable-internals": ["class_exists"]} in your patchwork.json.`
I am restricting with:
```		
 if ( ! class_exists( 'WooCommerce' ) ) {
     return [];
}
```
And in Integration Test I am using this one:
`Functions\expect( 'class_exists' )->once()->with( 'WooCommerce' )->andReturn( true );`


Tabrisrp:
We already have a subscriber for WooCommerce, you could use this once instead of creating a new class

crystinutzaa:
Indeed this can be moved into the existent WooCommerce Subscriber. The function `get_nonce_actions()` will be ongoing work. In this case shouldn't be better to reside in it's own class?

Also for Integration tests how I can bypass the condition: `if ( ! class_exists( 'WooCommerce' ) ) {` ?

I think we need a patchwork.json file and add the function: `class_exists`

hellofromtonya:
>I think we need a patchwork.json file and add the function: class_exists

We can't patch PHP language constructs or functions.

>Also for Integration tests how I can bypass the condition: if ( ! class_exists( 'WooCommerce' ) ) { ?

For integration tests, we want to include external package dependencies to ensure our code runs and behaves as expected when _fully integrated_.  How? Check out this [PR on our WC Rocket private repo](https://github.com/wp-media/woocommerce-rocket/pull/2).

inc/classes/subscriber/third-party/plugins/ecommerce/class-woocommerce-subscriber.php
```diff
@@ -418,4 +419,40 @@ public function exclude_product_attributes_cpcss( $excluded_taxonomies ) {
 
 		return array_merge( $excluded_taxonomies, wc_get_attribute_taxonomy_names() );
 	}
+
+	/**
+	 * Set $uid to 0 for certain nonce actions.
+	 *
+	 * @since  3.5.1
+	 * @access public
```
Tabrisrp: `@access`is not needed

inc/classes/subscriber/third-party/plugins/ecommerce/class-woocommerce-subscriber.php
```diff
@@ -418,4 +419,40 @@ public function exclude_product_attributes_cpcss( $excluded_taxonomies ) {
...
+			$uid = 0;
+		}
+		return $uid;
+	}
+
+	/**
+	 * List with nonce actions which needs to revert the $uid.
+	 *
+	 * @since  3.5.1
+	 * @access private
```
Tabrisrp: same here

inc/classes/subscriber/third-party/plugins/ecommerce/class-woocommerce-subscriber.php
```diff
@@ -418,4 +419,40 @@ public function exclude_product_attributes_cpcss( $excluded_taxonomies ) {
...
+	 * @author Soponar Cristina
+	 *
+	 * @return array $nonce_actions List with all nonce actions.
+	 */
+	private function get_nonce_actions() {
+		$nonce_actions = [
+			'wcmd-subscribe-secret', // WooCommerce MailChimp Discount.
+			'td-block', // "Load more" AJAX functionality of the Newspaper theme.
+		];
+		return $nonce_actions;
```
Tabrisrp: you can use the return directly on the array declaration, no need to create a variable here

inc/classes/subscriber/third-party/plugins/ecommerce/class-woocommerce-subscriber.php
```diff
@@ -418,4 +419,40 @@ public function exclude_product_attributes_cpcss( $excluded_taxonomies ) {
...
+	 * @param string $action The nonce action.
+	 *
+	 * @return int $uid      ID of the nonce-owning user.
+	 */
+	public function maybe_revert_uid_for_nonce_actions( $uid, $action ) {
+		if ( $uid && 0 !== $uid && $action && in_array( $action, $this->get_nonce_actions(), true ) ) {
+			$uid = 0;
+		}
+		return $uid;
+	}
```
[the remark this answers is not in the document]
crystinutzaa: For `$user_id`, WooCommerce generates the `$customer_id`/ `$uid` like this:
```
$hasher      = new PasswordHash( 8, false );
$customer_id = md5( $hasher->get_random_bytes( 32 ) );
```

So the `$user_id` in this case can be a string and with Woo will be a string, so the first condition  should be modified to:
```
if ( empty( $user_id ) || 0 === $user_id ) {
	return $user_id;
}
```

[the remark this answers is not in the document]
hellofromtonya: Interesting. So WooCommerce, and possibly other customer callbacks hooked into that same filter, are not necessarily using the primary key from the database as the user's ID. Then yes, you're right, i.e. we do need to check for different types of user/customer IDs.

inc/classes/subscriber/third-party/plugins/ecommerce/class-woocommerce-subscriber.php
```diff
@@ -421,38 +421,42 @@ public function exclude_product_attributes_cpcss( $excluded_taxonomies ) {
 	}
 
 	/**
-	 * Set $uid to 0 for certain nonce actions.
+	 * Set $user_id to 0 for certain nonce actions.
```
hellofromtonya: Also needs a "Why" explanation after the method's description.

>Change 6
>This callback is for a very specific edge case. @crystinutzaa let's give a reason why this method exists by adding it to the method's DocBlock. Notice that I wrote "Need an explanation here." Why do we include this? To help our future selves know why this method is important.

tests/Integration/Subscriber/ThirdParty/Plugins/Ecommerce/WooCommerce/TestNonceUserLoggedOut.php
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * @group Subscriber_TestNonce
+ */
+class TestNonceUserLoggedOut extends TestCase {
+	public function setUp() {
+		parent::setUp();
+
+		$this->handler = new \WC_Session_Handler();
+	}
+
```
hellofromtonya: Let's add another test for a non-WooCommerce handling test:

```php
	public function testShouldReturnGivenUserIdWhenActionNotInList() {
		wp_set_current_user( $this->factory->user->create() );
		$nonce = wp_create_nonce( __METHOD__ );

		$verified_nonce_failed_count = did_action( 'wp_verify_nonce_failed' );

		// Test when a user is logged in.
		$this->assertSame( 1, wp_verify_nonce( $nonce, __METHOD__ ) );
		$this->assertEquals( $verified_nonce_failed_count, did_action( 'wp_verify_nonce_failed' ) );

		// Test when there isn't a user logged in.
		wp_set_current_user( 0 );
		$this->assertFalse( wp_verify_nonce( $nonce, __METHOD__ ) );
		$this->assertEquals( $verified_nonce_failed_count + 1, did_action( 'wp_verify_nonce_failed' ) );
	}
```

tests/Integration/Subscriber/ThirdParty/Plugins/Ecommerce/WooCommerce/TestNonceUserLoggedOut.php
```diff
@@ -0,0 +1,55 @@
+<?php
+namespace WP_Rocket\Tests\Integration\ThirdParty\Plugins\Ecommerce\WooCommerce;
+use WP_Rocket\Tests\Integration\TestCase;
+use Brain\Monkey\Functions;
```
hellofromtonya: Brain Monkey isn't used. Let's remove this import.

inc/classes/subscriber/third-party/plugins/ecommerce/class-woocommerce-subscriber.php
hellofromtonya: Resolved with 2c1886ff35d242d6cd78a5caf918de3c2dd7cc14.

tests/Integration/Subscriber/ThirdParty/Plugins/Ecommerce/WooCommerce/TestNonceUserLoggedOut.php
hellofromtonya: Resolved with 2c1886ff35d242d6cd78a5caf918de3c2dd7cc14

PR #2214, opened 2019-12-26
Exclude Klaviyo JS file

GeekPress:
Klaviyo has a JS file to use their API. To work, this file must not be combined. A check is required to use their JS API and Klaviyo is checking that this file is loaded on the website.

If we combine the Klaviyo JS file, then Klaviyo isn't able to find it and can't validate the account to use the JS API.

e.g we had to exclude the Klaviyo JS file on our own website to be validated.

I decided to make a PR as we had our first ticket issue about that few days ago:
https://secure.helpscout.net/conversation/1036927375/137413/


hellofromtonya:
@Tabrisrp Should this PR be targeted at `branch-3.5` instead of `develop`?

GeekPress:
> I think it should be in the `get_excluded_external_file_path()` method, not in this one

Indeed! Sorry for the mistake 😢

PR #2218, opened 2019-12-27
Fix #2107 CDN srcset rewrites

### LINKED ISSUE #2107: CDN srcset rewrites
When rewriting `srcset` elements and one of the sizes is repeated, we will print wrong output (CDN URL will be multiplied) due to this change:
https://github.com/wp-media/wp-rocket/blob/389f7462cd9cc8424ab0b3ab88acea4db6884cd9/inc/classes/CDN/CDN.php#L71

To be more specific, we will be looking for `branding-magazine-news.png` in the already changed path.

Not sure if we should handle this kind of situation, as this doesn't seem to be correct markup:
http://snippi.com/s/6w69x4l

Look for: `branding-magazine-news.png 1024w`

Related ticket:
https://secure.helpscout.net/conversation/999401292/130285?folderId=2415573

crystinutzaa:
Fixes multiple CDN url replacement on srcset rewrites.

Applied array_unique() after I made a trim() on the sources array.

### DIFF
--- a/inc/classes/CDN/CDN.php
+++ b/inc/classes/CDN/CDN.php
@@ -61,11 +61,18 @@ class CDN {
 		if ( ! preg_match_all( $pattern, $html, $srcsets, PREG_SET_ORDER ) ) {
 			return $html;
 		}
-
 		foreach ( $srcsets as $srcset ) {
 			$sources    = explode( ',', $srcset['sources'] );
+			$sources    =
+				array_unique(
+					array_map(
+						function( $source ) {
+							return trim( $source );
+						},
+						$sources
+					)
+				);
 			$cdn_srcset = $srcset['sources'];
-
 			foreach ( $sources as $source ) {
 				$url        = \preg_split( '#\s+#', trim( $source ) );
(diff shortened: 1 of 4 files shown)


inc/classes/CDN/CDN.php
```diff
@@ -61,11 +61,18 @@ public function rewrite_srcset( $html ) {
...
 			$sources    = explode( ',', $srcset['sources'] );
+			$sources    =
+				array_unique(
+					array_map(
+						function( $source ) {
+							return trim( $source );
+						},
+						$sources
+					)
+				);
```
hellofromtonya: This code can be simplified into a single line:

```php
$sources  = array_unique( array_map( 'trim', $sources ) );
```

PHP looks up `trim`. As it's a callback, it executes that function. [See `array_map` in the PHP manual](https://www.php.net/manual/en/function.array-map.php) for more info.

Why do this?

- It's more readable.
- It'll run slightly faster and take a little less memory. Why? PHP doesn't have to create the closure in memory.

PR #2219, opened 2019-12-27
Fix #2142 - Partial Preloading is not working after cache lifespan expires

### LINKED ISSUE #2142: Partial preloading is not working after cache lifespan expires
Partial preloading, when the cache lifespan expires and permalinks have a trailing `/`, doesn't create cached files. When clearing a post's cache it works as expected. The same goes if the permalinks are without a `/`.

The `preload_after_automatic_cache_purge( $deleted )` function runs after the cache lifespan has expired:
https://github.com/wp-media/wp-rocket/blob/7723e789d158d2bf00ee5da49ef9158061e158eb/inc/classes/subscriber/preload/class-partial-preload-subscriber.php#L123-L142

The `$deleted` array is produced here:
https://github.com/wp-media/wp-rocket/blob/7723e789d158d2bf00ee5da49ef9158061e158eb/inc/classes/Cache/class-expired-cache-purge.php#L89-L146

It looks like this:
```JSON
Array
(
    [0] => Array
        (
            [home_url] => http://mega.wp-rocket.me/training
            [home_path] => /home/wpquicki/mega/training/wp-content/cache/wp-rocket/mega.wp-rocket.me/training
            [logged_in] => 
            [files] => Array
                (
                    [0] => /home/wpquicki/mega/training/wp-content/cache/wp-rocket/mega.wp-rocket.me/training/eum-animi-inventore-labore-quidem-facere
                    [1] => /home/wpquicki/mega/training/wp-content/cache/wp-rocket/mega.wp-rocket.me/training/id-quis-occaecati-ducimus-nemo-repudiandae
                    [2] => /home/wpquicki/mega/training/wp-content/cache/wp-rocket/mega.wp-rocket.me/training/category/uncategorized
                )
        )
)
```
The `files` don't have a trailing `/` and none is added by the `preload_after_automatic_cache_purge( )` function:
https://github.com/wp-media/wp-rocket/blob/7723e789d158d2bf00ee5da49ef9158061e158eb/inc/classes/Cache/class-expired-cache-purge.php#L128

A URL without a trailing `/` will cause a redirect `301`. That fails here:
https://github.com/wp-media/wp-rocket/blob/7723e789d158d2bf00ee5da49ef9158061e158eb/inc/classes/Buffer/class-tests.php#L324-L328

So, no cache is created for the `$deleted` files.

**Related ticket:**
https://secure.helpscout.net/conversation/1012444753/132479?folderId=2135277

_PS: It can also be tested on one of our testing sites._

crystinutzaa:
Partial preloading is not working after cache lifespan expires based on trailing /

### DIFF
--- a/inc/classes/subscriber/preload/class-partial-preload-subscriber.php
+++ b/inc/classes/subscriber/preload/class-partial-preload-subscriber.php
@@ -135,8 +135,10 @@ class Partial_Preload_Subscriber implements Subscriber_Interface {
 					// URL with query string.
 					$file_path = preg_replace( '/#/', '?', $file_path, 1 );
 				} else {
-					$permalink_structure = get_option( 'permalink_structure' );
-					$file_path           = rtrim( $file_path, '/' ) . ( '/' === substr( $permalink_structure, -1 ) ? '/' : '' );
+					$file_path = untrailingslashit( $file_path );
+					if ( '/' === substr( get_option( 'permalink_structure' ), -1 ) ) {
+						$file_path .= '/';
+					}
 				}
 
 				$this->urls[] = str_replace( $data['home_path'], $data['home_url'], $file_path );
--- a/inc/classes/subscriber/preload/class-partial-preload-subscriber.php
+++ b/inc/classes/subscriber/preload/class-partial-preload-subscriber.php
@@ -134,6 +134,9 @@ class Partial_Preload_Subscriber implements Subscriber_Interface {
 				if ( strpos( $file_path, '#' ) ) {
 					// URL with query string.
 					$file_path = preg_replace( '/#/', '?', $file_path, 1 );
+				} else {
+					$permalink_structure = get_option( 'permalink_structure' );
+					$file_path           = rtrim( $file_path, '/' ) . ( '/' === substr( $permalink_structure, -1 ) ? '/' : '' );
 				}
 
 				$this->urls[] = str_replace( $data['home_path'], $data['home_url'], $file_path );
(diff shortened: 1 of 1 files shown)


inc/classes/subscriber/preload/class-partial-preload-subscriber.php
```diff
@@ -134,6 +134,9 @@ public function preload_after_automatic_cache_purge( $deleted ) {
 				if ( strpos( $file_path, '#' ) ) {
 					// URL with query string.
 					$file_path = preg_replace( '/#/', '?', $file_path, 1 );
+				} else {
+					$permalink_structure = get_option( 'permalink_structure' );
+					$file_path           = rtrim( $file_path, '/' ) . ( '/' === substr( $permalink_structure, -1 ) ? '/' : '' );
```
Tabrisrp:
```suggestion
					$file_path           = untrailingslashit( $file_path ) . ( '/' === substr( $permalink_structure, -1 ) ? '/' : '' );
```

hellofromtonya: To improve upon @Tabrisrp's suggestion, I'd suggest the following:

1. Remove this line of code `$permalink_structure = get_option( 'permalink_structure' );`
2. Only do the concatenation when needed, instead of each time. 

Why? 

- The code will execute a few nanoseconds faster. Why?
    - It doesn't have to create the variable in memory in PHP's symtable.
    - It doesn't have to lookup the value in memory that's wired to the `$permalink_structure` variable.
    - It doesn't concatenate an empty string to the end of the `$file_path`.
- It's more readable and understandable.
    - The option tells us what it represents and what will be returned from `get_option`.
    - The conditional tells us: "Hey, if the permalink structure ends with a `/`, then let's add that to the end of the file path."

```php
$file_path = untrailingslashit( $file_path );
if ( '/' === substr(  get_option( 'permalink_structure' ), -1 ) {
    $file_path .= '/';
}
```

PR #2221, opened 2019-12-30
Fix #2159 - Don't clear the cache when preload option is changed

### LINKED ISSUE #2159: Don't clear the cache when preload option is changed
Currently (tested with 3.4.2.2), when you turn on or off the preload option, the cache is cleared, but there's no benefit to or need for doing that. 

Slack: https://wp-media.slack.com/archives/C43T1AYMQ/p1575513615021100

crystinutzaa:
Fix #2159 - Don't clear the cache when preload option is changed

When you turn on or off the preload option, the cache is cleared, but there's no benefit to or need for doing that.

### DIFF
--- a/inc/admin/options.php
+++ b/inc/admin/options.php
@@ -54,7 +54,6 @@ function rocket_after_save_options( $oldvalue, $value ) {
 		'sucury_waf_cache_sync'       => true,
 		'sucury_waf_api_key'          => true,
 		'manual_preload'              => true,
-		'sitemap_preload'             => true,
 	];
 
 	// Create 2 arrays to compare.
--- a/inc/admin/options.php
+++ b/inc/admin/options.php
@@ -53,6 +53,8 @@ function rocket_after_save_options( $oldvalue, $value ) {
 		'analytics_enabled'           => true,
 		'sucury_waf_cache_sync'       => true,
 		'sucury_waf_api_key'          => true,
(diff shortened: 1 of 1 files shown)


inc/admin/options.php
```diff
@@ -53,6 +53,8 @@ function rocket_after_save_options( $oldvalue, $value ) {
 		'analytics_enabled'           => true,
 		'sucury_waf_cache_sync'       => true,
 		'sucury_waf_api_key'          => true,
+		'manual_preload'              => true,
+		'sitemap_preload'             => true,
```
Tabrisrp: `sitemap_preload` is already on line 29 above

inc/admin/options.php
```diff
@@ -53,6 +53,7 @@ function rocket_after_save_options( $oldvalue, $value ) {
 		'analytics_enabled'           => true,
 		'sucury_waf_cache_sync'       => true,
 		'sucury_waf_api_key'          => true,
+		'manual_preload'              => true,
```
hellofromtonya: @crystinutzaa Were you able to confirm that `'manual_preload'` is the root cause for this problem?

crystinutzaa: @hellofromtonya Yes, I confirmed the `manual_preload` is the root cause of this issue. Because is not excluded , then each time the option is changed it will clean WP Rocket cache.

PR #2223, opened 2019-12-30
Fix #2185 - PHP 7.4 notice - Trying to access array offset on value of type null, in class-cache-dynamic-resource.php

### LINKED ISSUE #2185: PHP 7.4 notice - Trying to access array offset on value of type null, in class-cache-dynamic-resource.php
On Kinsta, this website is showing this notice after upgraded to PHP7.4

`Notice: Trying to access array offset on value of type null in /www/xxxx/public/wp-content/plugins/wp-rocket/inc/classes/optimization/class-cache-dynamic-resource.php on line 170`

This is the specific line:
 https://github.com/wp-media/wp-rocket/blob/7723e789d158d2bf00ee5da49ef9158061e158eb/inc/classes/optimization/class-cache-dynamic-resource.php#L170

Ticket:
https://secure.helpscout.net/conversation/1028206508/136026?folderId=2683093

crystinutzaa:
Fixes PHP 7.4 notice 

```
Notice: Trying to access array offset on value of type null in /public/wp-content/plugins/wp-rocket/inc/classes/optimization/class-cache-dynamic-resource.php on line 170
```

Reference Issue #2185

### DIFF
--- a/inc/classes/optimization/class-cache-dynamic-resource.php
+++ b/inc/classes/optimization/class-cache-dynamic-resource.php
@@ -167,7 +167,7 @@ class Cache_Dynamic_Resource extends Abstract_Optimization {
 	public function is_excluded_file( $src ) {
 		$file = get_rocket_parse_url( $src );
 
-		if ( is_array( $file ) && ! empty( $file['path'] ) && ! preg_match( '#\.php$#', $file['path'] ) ) {
+		if ( isset( $file['path'] ) && ! preg_match( '#\.php$#', $file['path'] ) ) {
 			return true;
 		}
 
@@ -179,15 +179,13 @@ class Cache_Dynamic_Resource extends Abstract_Optimization {
 			return true;
 		}
 
-		if ( is_array( $file ) && ! empty( $file['query'] ) ) {
-			$file['query'] = remove_query_arg( 'ver', $file['query'] );
-
-			if ( $file['query'] ) {
-				return true;
-			}
+		if ( ! isset( $file['query'] ) ) {
+			return false;
 		}
 
-		return false;
+		$file['query'] = remove_query_arg( 'ver', $file['query'] );
+
+		return (bool) $file['query'];
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


inc/classes/optimization/class-cache-dynamic-resource.php
```diff
@@ -167,7 +167,7 @@ public function is_allowed() {
 	public function is_excluded_file( $src ) {
 		$file = get_rocket_parse_url( $src );
 
-		if ( ! preg_match( '#\.php$#', $file['path'] ) ) {
+		if ( is_array( $file ) && ! empty( $file['path'] ) && ! preg_match( '#\.php$#', $file['path'] ) ) {
```
hellofromtonya: `isset( $file['path'] )` does the following checks for us:

- `$file` is an array <- inferred because we are checking for a key.
- Key `'path'` is defined
- `$file['path'` is not `null`

inc/classes/optimization/class-cache-dynamic-resource.php
```diff
@@ -179,10 +179,12 @@ public function is_excluded_file( $src ) {
 			return true;
 		}
 
-		$file['query'] = remove_query_arg( 'ver', $file['query'] );
+		if ( is_array( $file ) && ! empty( $file['query'] ) ) {
```
hellofromtonya: Here we can use `isset( $file['query'] )` instead. But we can also simplify this code a little bit more to eliminate by converting this into a guard clause instead. If it's not set, return false; else, remove the `ver` query arg and recheck.


```php
if ( ! isset( $file['query'] ) ) {
	return false;
}

$file['query'] = remove_query_arg( 'ver', $file['query'] );
// ... reset of the code here.
```

inc/classes/optimization/class-cache-dynamic-resource.php
```diff
@@ -179,10 +179,12 @@ public function is_excluded_file( $src ) {
...
+			$file['query'] = remove_query_arg( 'ver', $file['query'] );
 
-		if ( $file['query'] ) {
-			return true;
+			if ( $file['query'] ) {
```
hellofromtonya: As this is the last conditional check in the method, we can eliminate the hard coded boolean return statements. How? Return the boolean state from the conditional expression. 

[See this quick tip](https://knowthecode.io/smelly-code-ifreturn-truefalse-code-pattern) for more info.

PR #2229, opened 2020-01-03
Fixed #1550 Update Wordpress Compatibility

### LINKED ISSUE #1550: Update WordPress compatibility
We display this information to customers : 

```
You have version 3.1.4 installed. Update to 3.2.5. View version 3.2.5 details.
Compatibility with WordPress 4.9.9: Unknown
Compatibility with WordPress 5.0.3: Unknown
```

We would need to update the compatibility to the latest version of WordPress.
We can use that : https://github.com/YahnisElsts/plugin-update-checker/blob/d98d6bc6caac4b46b61186447a947323a466a021/Puc/v4p5/Plugin/UpdateChecker.php#L94

crystinutzaa:
Updated Wordpress Compatibility

I added a new constant: `WP_ROCKET_WP_VERSION_TESTED` which includes the latest version of Wordpress and how it displays like this: https://www.screencast.com/t/y3gCQ5ZOpG6R

This can also be fetched from update API or directly from Wordpress with `get_core_updates()` function. 

Reference Issue #1550 

**Which should be the best approach?**

### DIFF
--- a/inc/classes/subscriber/Plugin/class-updater-subscriber.php
+++ b/inc/classes/subscriber/Plugin/class-updater-subscriber.php
@@ -366,6 +366,7 @@ class Updater_Subscriber implements Subscriber_Interface {
 		$obj->new_version = $match['user_version'];
 		$obj->url         = $this->vendor_url;
 		$obj->package     = $match['package'];
+		$obj->tested      = WP_ROCKET_WP_VERSION_TESTED;
 
 		if ( $this->icons && ! empty( $this->icons['1x'] ) ) {
 			$obj->icons = $this->icons;
--- a/wp-rocket.php
+++ b/wp-rocket.php
@@ -20,6 +20,7 @@ defined( 'ABSPATH' ) || die( 'Cheatin&#8217; uh?' );
 // Rocket defines.
 define( 'WP_ROCKET_VERSION',               '3.5-alpha1' );
 define( 'WP_ROCKET_WP_VERSION',            '4.9' );
+define( 'WP_ROCKET_WP_VERSION_TESTED',     '5.3.2' );
 define( 'WP_ROCKET_PHP_VERSION',           '5.6' );
 define( 'WP_ROCKET_PRIVATE_KEY',           false );
 define( 'WP_ROCKET_SLUG',                  'wp_rocket_settings' );


GeekPress:
I will suggest to use a dynamic value. 

By doing it like now, we will have to make an update each time a new WordPress version will be released.

We have an endpoint where we can retrieve the tested info => https://wp-rocket.me/stat/1.0/wp-rocket/

hellofromtonya:
>I will suggest to use a dynamic value.
>
>By doing it like now, we will have to make an update each time a new WordPress version will be released.
>
>We have an endpoint where we can retrieve the tested info => https://wp-rocket.me/stat/1.0/wp-rocket/

While I like the idea of pulling minimum requirements from the endpoint @GeekPress shared, that endpoint contains information about the _**latest released version.**_ 

Think about previously released versions. If we dynamically pull this information in as a core function of Rocket, then every released version that has that code in it will have the same tested version. But that's not true. Right?

What do I mean?

We tested a release against a baseline of versions for its dependencies. Those versions will change in the future with more releases. So we can conclude that previous released versions of Rocket will have different dependency and tested version requirements.

IMO each released version of the Rocket plugin needs to clear identify what it needs to run (including what versions it's been tested against). Yes, it's a manual configuration, but that's by design. Why? We are indicating what a specific version of Rocket needs to run and what it was tested against. That's a hardcoded setting within each version of Rocket, just as the version of Rocket is also hardcoded.

hellofromtonya:
>Which should be the best approach?

The tested at version parameter needs to _hardcoded_, not dynamically pulled from an API ([see my comments above](https://github.com/wp-media/wp-rocket/pull/2229#issuecomment-570624530)).

For this PR, I suggest staying with the current code design of hardcoding versions into defined constants.

GeekPress:
>  If we dynamically pull this information in as a core function of Rocket, then every released version that has that code in it will have the same tested version. But that's not true. Right?

Indeed, good catch ;)

Tabrisrp:
Is it necessary to add an additional global constant for this? The value is used in one place, it could be set as a class constant instead, or the value directly assigned to the object property.

hellofromtonya:
>Is it necessary to add an additional global constant for this? The value is used in one place, it could be set as a class constant instead, or the value directly assigned to the object property.

I'd suggest that we keep all hardcoded parameters in one place. 

Why? To make it easier for us to manage changes for each version released. It also reminds us that when we change Rocket's version number we need to also consider WordPress' version and what version we tested it at.

crystinutzaa:
Thank you all for this review. Indeed @hellofromtonya has the best argument why to not use the API :)
Also I would not want to keep it as a class constant or hardcoded parameter because that file can be easily missed in the next upgrades.

PR #2231, opened 2020-01-06
Fixed: decreased integration tests running time

crystinutzaa:
Integration tests take a long time to run due to :
```
/**
     * @runInSeparateProcess
     * @preserveGlobalState disabled
     */
```
I inverted the @runInSeparateProcess to be available only for the function which defines the constant and it decreased the running time to 1.1s all integration tests (before it was about 9minutes)

### DIFF
--- a/tests/Integration/Subscriber/CDNSubscriber/TestRewrite.php
+++ b/tests/Integration/Subscriber/CDNSubscriber/TestRewrite.php
@@ -6,6 +6,7 @@ use WP_Rocket\Admin\Options_Data;
 use WP_Rocket\Admin\Options;
 use WP_Rocket\CDN\CDN;
 use WP_Rocket\Subscriber\CDN\CDNSubscriber;
+use Brain\Monkey;
 
 /**
  * @group Subscriber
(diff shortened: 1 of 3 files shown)


tests/Integration/Subscriber/CDNSubscriber/TestRewrite.php
```diff
@@ -64,6 +64,10 @@ public function testShouldReturnOriginalWhenCDNDisabled() {
         );
     }
 
+	/**
+     * @runInSeparateProcess
+     * @preserveGlobalState disabled
+     */
```
hellofromtonya: @crystinutzaa Why does this test need to run in a separate process? is it because the constant `DONOTROCKETOPTIMIZE` is being defined? If yes, [take a look at this commit in v3.5](https://github.com/wp-media/wp-rocket/pull/2196/commits/b885ec6185cd5bfb52e23148f18065a95daae2b2). It'll let us _mock_ constants for testing instead of actually defining them in our tests.

crystinutzaa: @hellofromtonya yes, running in separate process is related to constants. If I remember well all tests which need to run in separete process are related to constants, so this will be great. In this situation should I merge this commit in develop branch and fix all tests from develop which have run in separateprocess?

hellofromtonya: By [cherry picking this commit](https://github.com/wp-media/wp-rocket/pull/2196/commits/b885ec6185cd5bfb52e23148f18065a95daae2b2), we can mock the constant itself and not have to run in a separate process. Then we can merge this PR into 3.4.4.

PR #2247, opened 2020-01-13
Stop processing when "0", instead of "1", links are found.

### LINKED ISSUE #2241: The "display" parameter isn't added when there is just one Google Font
When there are more than one Google Fonts on a page, WP Rocket will add the `&display` parameter:
https://github.com/wp-media/wp-rocket/blob/dfb5e1cf4fe9486b528d6ca9eaf33605acc7bceb/inc/classes/optimization/CSS/class-combine-google-fonts.php#L153-L165

However, if there is just one this parameter is not added. The processing stops here:
https://github.com/wp-media/wp-rocket/blob/dfb5e1cf4fe9486b528d6ca9eaf33605acc7bceb/inc/classes/optimization/CSS/class-combine-google-fonts.php#L93

Although this is not a big issue, since the code is there, we should add the parameter.
Changing the previous to this will work:
```PHP
		if ( count( $matches ) == 0 ) {
```
**Steps to reproduce:**

1. Add one Google Font on a page.
2. Enable the **Combine Google Fonts** feature.
3. The `display` parameter isn't appended to the URL that WP Rocket "produces".

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

vmanthos:
Fixes #2241.

### DIFF
--- a/inc/classes/optimization/CSS/class-combine-google-fonts.php
+++ b/inc/classes/optimization/CSS/class-combine-google-fonts.php
@@ -90,7 +90,7 @@ class Combine_Google_Fonts extends Abstract_Optimization {
 	protected function find( $pattern, $html ) {
 		preg_match_all( '/' . $pattern . '/Umsi', $html, $matches, PREG_SET_ORDER );
 
-		if ( count( $matches ) <= 1 ) {
+		if ( count( $matches ) == 0 ) {
 			return false;
 		}
 


Tabrisrp:
I'm going to close this PR as we establised in #2241 that the work needed is bigger than that 👍

PR #2251, opened 2020-01-15
Fixes 2584: Normalizes directory separators for expired cache for IIS.

### LINKED ISSUE #2584: Expired cache preload doesn't work on IIS
**Describe the bug**
Expired cache preload doesn't work on IIS due to the slashes and backslashes usage in the paths.

That's because we treat the paths as logged in ones here:
https://github.com/wp-media/wp-rocket/blob/00cf7a1c5e0b2e7354b0aa76ca86603ee4420eae/inc/Engine/Preload/PartialPreloadSubscriber.php#L134

The `$data['logged_in']` is true, because the paths don't match here:
https://github.com/wp-media/wp-rocket/blob/e40b824350a165cf52f2285a681d5b6796ed5555/inc/classes/Cache/class-expired-cache-purge.php#L131

The values:
`Dir path: D:\inetpub\vhosts\domain.com\blog.domain.com/wp-content/cache/wp-rocket\blog.domain.com`

`Cache path: D:\inetpub\vhosts\monteleone.it\blog.domain.com/wp-content/cache/wp-rocket/blog.domain.com`

Please note the slash after `wp-rocket`. 

**To Reproduce**
1. Go to the WP Rocket dashboard hosted on IIS
2. Set cache lifespan to 10 minutes
3. Preload cache
4. After 10 minutes confirm that it's not preloaded

**Expected behavior**
It's expected to preload the cache correctly

**Additional context**
I don't know if there's any other feature that could be affected by this problem.

Related ticket:
https://secure.helpscout.net/conversation/1141100768/159114?folderId=2415573

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

crystinutzaa:
IIS directory separator is \ and this is breaking the partial preload system.
The first issue is related to logged checkup in here:
https://github.com/wp-media/wp-rocket/blob/57940fa791b42ee3ad96f648e376b01542d5822e/inc/classes/Cache/class-expired-cache-purge.php#L131

$dir_path will always be different than $this->cache_path so that partial preload will not work due to logged always true.
Wordpress default directory separator is `/` and 
`$dir_path     = $item->getPathname();` will get the value with `\`

The second cause is in here:
https://github.com/wp-media/wp-rocket/blob/57940fa791b42ee3ad96f648e376b01542d5822e/inc/classes/subscriber/preload/class-partial-preload-subscriber.php#L144

The URLs will have both \/ inside eg:
`https:://example.org\/category\uncategorized/`

## TODO

- [x] Move Expired Cache files into new architecture
- [x] Merge `develop` with new normalize function
- [x] Normalize paths in the `ExpiredCachePurge`
- [ ] Normalize paths in `PartialPreloadSubscriber`
- [ ] Live test on customer sites reporting this issue
- [ ] QA


hellofromtonya:
@crystinutzaa Needs tests with it to ensure the changes work and don't break anything.

Tabrisrp:
You have a WP function, wp_normalize_path, to handle this kind of issues: https://developer.wordpress.org/reference/functions/wp_normalize_path/

inc/classes/subscriber/preload/class-partial-preload-subscriber.php
```diff
@@ -140,8 +140,10 @@ public function preload_after_automatic_cache_purge( $deleted ) {
 						$file_path .= '/';
 					}
 				}
-
-				$this->urls[] = str_replace( $data['home_path'], $data['home_url'], $file_path );
+				$file_path = str_replace( $data['home_path'], '', $file_path );
```
Screenfeed: Good luck when `$data['home_path']` is equal to `/`, or when it is equal to `/foo` and `$file_path` is equal to `/foo/foobar/file.html` :s 
I think `preg_replace()` + `preg_quote()` will save us. Something like this (to test):
```php
$data['home_path'] = trailingslashit( wp_normalize_path( $data['home_path'] ) );
$file_path         = wp_normalize_path( $file_path );
$file_path         = preg_replace( '@^' . preg_quote( $data['home_path'], '@' ) . '@', '', $file_path );
$this->urls[]      = $data['home_url'] . $file_path;
```

Screenfeed (changes requested):

Screenfeed (approved):

hellofromtonya (changes requested):
Blocking the merge of this PR. Why? We need to figure out how to handle both files and directories for both Linux and IIS.

PR #2273, opened 2020-01-22
Unit/integration tests for rocket_delete_licence_data_file()

Tabrisrp:
- Update the function to use `rocket_get_constant()`
- Update the unit test to follow our new organization and reflect the change above
- Add integration test for the function


tests/Integration/Functions/Options/rocketDeleteLicenceDataFile.php
```diff
@@ -0,0 +1,40 @@
+<?php
```
hellofromtonya: The file should be located in `tests/Integration/Functions/`. `Options` would if it were a method on a class, where the class name is `Options`. 

Why?

All of the functions exist in the same namespace, which right now is the global namespace. They belong to the folder structure. Though they live in a particular file, they don't belong to that file per se. What do I mean? We could move the function to a different file without impacting its ability to run. Breaking it up into files helps us to organize common functionality around a single reason to exist. That's for organizational purposed.

A benefit of having all of the functions in the `Functions` folder is we can see a list of all functions to know the coverage.

Why are method test files within a folder named after the class? 2 reasons:

1. The method belongs to that class. It's bound to it. A function on the other hand is not. It exists outside of a class.
2. Name conflicts. Methods can have the same name across multiple classes. For example, we can have a method called `init` that exists within multiple classes. But we couldn't have `init.php` all within one folder for all of the classes in that folder. Right? So instead we put method test files within a folder named after its class. It bundles it together for that class.

Does that make sense?

tests/Integration/Functions/Options/rocketDeleteLicenceDataFile.php
```diff
@@ -0,0 +1,40 @@
...
+		Functions\when( 'rocket_get_constant' )
+			->justReturn( $this->path->url() . '/' );
+
+		$this->assertTrue( $this->path->hasChild( 'licence-data.php' ) );
+
+		rocket_delete_licence_data_file();
+
+		$this->assertFalse( $this->path->hasChild( 'licence-data.php' ) );
+	}
+}
```
hellofromtonya: Needs an empty line as the line of the file.

tests/Unit/Functions/Options/rocketDeleteLicenceDataFile.php
```diff
@@ -0,0 +1,57 @@
+<?php
```
hellofromtonya: Test file should be in `tests/Unit/Functions` instead. See explanation above.

tests/Unit/Functions/Options/rocketDeleteLicenceDataFile.php
```diff
@@ -0,0 +1,57 @@
...
+ * @group Options
+ */
+class Test_RocketDeleteLicenceDataFile extends TestCase {
+	private $path;
+	private $mock_fs;
+
+	public function setUp() {
+		parent::setUp();
+
+		require WP_ROCKET_PLUGIN_ROOT . 'inc/functions/options.php';
```
hellofromtonya: `require_once` instead. Why? `setUp()` runs for each test method in the class. Right now, there's only 1 test. Later we may have more. What happens if `setUp()` runs more than once? Fatal error. Switching it to `require_once` resolves that problem.

tests/Unit/Functions/Options/rocketDeleteLicenceDataFile.php
```diff
@@ -0,0 +1,57 @@
...
+			return $this->mock_fs;
+		});
+
+		$this->assertTrue( $this->path->hasChild( 'licence-data.php' ) );
+
+		rocket_delete_licence_data_file();
+
+		$this->assertFalse( $this->path->hasChild( 'licence-data.php' ) );
+	}
+}
```
hellofromtonya: Empty line at the end of the file.

PR #2276, opened 2020-01-22
Add API Credential Handling to Tests.

hellofromtonya:
Adds the ability to load private API credentials from:

- local machine's env var
- local machine's config file, i.e. located in `tests/env/local/`.
- via Travis in an encrypted env var

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 /node_modules/
 /vendor/*
+/tests/env/local/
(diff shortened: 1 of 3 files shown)


Screenfeed:
@hellofromtonya Question: any particular reason to use static property and method instead of instantiated ones?

hellofromtonya:
@Screenfeed The static property allows us to set the value once for all of the tests. The static method was intended to load the vars once and then return them upon request. Why? Speed up the tests. But looking at it, I didn't achieve the "load once and reuse" idea.

PR #2277, opened 2020-01-22
Tests for PurgeActionsSubscriber::purge_user_cache

Tabrisrp:
- Update unit tests to follow new organization
- Add integration tests


tests/Integration/Subscriber/Cache/PurgeActionsSubscriber/purgeUserCache.php
```diff
@@ -0,0 +1,58 @@
...
+	use \Mockery\Adapter\Phpunit\MockeryPHPUnitIntegration;
+
+	/**
+	 * Test should not purge user cache when user cache is disabled
+	 */
+	public function testShouldNotPurgeUserCacheWhenUserCacheDisabled() {
+		add_filter( 'pre_get_rocket_option_cache_logged_user', '__return_false' );
+
+		Functions\expect( 'rocket_clean_user' )
+			->never();
```
hellofromtonya: Instead of mocking the function, in an integrated tests, we want to validate that the user's cache is not purged when firing the `'delete_user'` event.

Why?

It validates the real expected behavior: when user cache is disabled (which happens by the filter), don't delete the user cache.

How?
Add a user cache. Then fire the event. Then check that the user's cache did not change.

tests/Integration/Subscriber/Cache/PurgeActionsSubscriber/purgeUserCache.php
```diff
@@ -0,0 +1,58 @@
...
+
+	/**
+	 * Test should not purge user cache when commonuser cache is enabled
+	 */
+	public function testShoulNotPurgeUserCacheWhenCommonUserCacheEnabled() {
+		add_filter( 'pre_get_rocket_option_cache_logged_user', '__return_true' );
+		add_filter( 'rocket_common_cache_logged_users', '__return_true' );
+
+		Functions\expect( 'rocket_clean_user' )
+			->never();
```
hellofromtonya: Same thing here. 

1. Add a user cache. 
2. Fire the event.
3. Check that the user's cache is the same (i.e. not purged).

tests/Integration/Subscriber/Cache/PurgeActionsSubscriber/purgeUserCache.php
```diff
@@ -0,0 +1,58 @@
...
+	 * Test should purge cache for user ID = 1
+	 */
+	public function testShouldPurgeCacheForUserID1() {
+		add_filter( 'pre_get_rocket_option_cache_logged_user', '__return_true' );
+
+		Functions\expect( 'rocket_clean_user' )
+			->once()
+			->with( 1 );
+
+		do_action( 'delete_user', 1 );
```
hellofromtonya: Same thing here. 

1. Add a user cache. 
2. Fire the event.
3. Check that the user's cache is purged.

tests/Unit/Subscriber/Cache/PurgeActionsSubscriber/purgeUserCache.php
```diff
@@ -0,0 +1,78 @@
...
+				'cache_logged_user',
+				0,
+				0,
+			],
+		];
+
+		$options->method( 'get' )->will( $this->returnValueMap( $map ) );
+
+		$purge = new PurgeActionsSubscriber( $options );
+		$this->assertNull( $purge->purge_user_cache( 1 ) );
```
hellofromtonya: `$this->assertNull()` doesn't tell us anything. Why? The method always returns `null` whether purging or not. Instead, use BrainMonkey to mock that `rocket_clean_user()` never runs.

```php
Functions\expect( 'purge_user_cache' )->never();

$purge = new PurgeActionsSubscriber( $options );
$purge->purge_user_cache( 1 );
```

tests/Unit/Subscriber/Cache/PurgeActionsSubscriber/purgeUserCache.php
```diff
@@ -0,0 +1,78 @@
...
+		];
+
+		$options->method( 'get' )->will( $this->returnValueMap( $map ) );
+
+		Filters\expectApplied( 'rocket_common_cache_logged_users' )
+		->once()
+		->andReturn( true );
+
+		$purge = new PurgeActionsSubscriber( $options );
+		$this->assertNull( $purge->purge_user_cache( 1 ) );
```
hellofromtonya: Same thing here.

tests/Integration/Subscriber/Cache/PurgeActionsSubscriber/purgeUserCache.php
Tabrisrp: That was my original idea, but then here was my thoughts:
- I want to validate the behaviour of the `purge_user_cache()` specifically
- If I add a user cache, I'm doing more than that, I'm validating `rocket_clean_user()` as a side-effect

So I opted for mocking `rocket_clean_user()` there

hellofromtonya: In a unit test, you are testing the isolated behavior of `purge_user_cache()`. Here you are testing to integrated behavior. Its job is to purge or not purge. The implementation of how we purge might change over time, meaning that `rocket_clean_user()` might be changed out to a different implementation. But the behavior and intent of the method stays the same: purge the user's cache.

PR #2296, opened 2020-02-05
Exclude combination of jQuery served from c0.wp.com

vmanthos:
The exclusion is done in a dynamic way, taking into consideration WordPress's version. `c0.wp.com`.

I'll close the older PR about this:
https://github.com/wp-media/wp-rocket/pull/1869

### DIFF
--- a/inc/classes/optimization/JS/class-combine.php
+++ b/inc/classes/optimization/JS/class-combine.php
@@ -731,6 +731,7 @@ class Combine extends Abstract_JS_Optimization {
 			'ck.page',
 			'cdn.jsdelivr.net/gh/AmauriC/',
 			'static.klaviyo.com/onsite/js/klaviyo.js',
+			'c0.wp.com/c/' . get_bloginfo( 'version' ) . '/wp-includes/js/jquery/jquery.js',
 		];
 
 		$excluded_external = array_merge( $defaults, $this->options->get( 'exclude_js', [] ) );


alfonso100:
Another case here: 
https://secure.helpscout.net/conversation/1107497866/151190?folderId=273761

webtrainingwheels:
I believe this requires a bit more validation before we automatically exclude jQuery for _all_ Jetpack site accelerator users. I have a test site where I have Combine JS active with Jetpack Site Accelerator managing the static files, and there is no problem. 
And I think we'd have a lot more tickets about this if it happened in every case.

vmanthos:
@webtrainingwheels 

>  I believe_ this requires a bit more validation before we automatically exclude jQuery for all Jetpack site accelerator users

As you know, we do this already for the jQuery that's included with WordPress, when someone enables the **Safe mode for jQuery**.

> I have a test site where I have Combine JS active with Jetpack Site Accelerator managing the static files, and there is no problem.

Does your test site have inline jQuery-depended JavaScript?
Please ping me in Slack, so I can check this.

> And I think we'd have a lot more tickets about this if it happened in every case.

I've come across sites where there is a `jQuery not defined` error, not necessarily related to WP Rocket, where the customer doesn't have a clue about it because the broken functionality isn't something obvious.

In the same spirit as with the WordPress-included jQuery, we have to make our best to avoid this.

alfonso100:
I have been able to confirm this issue on this ticket: 
https://secure.helpscout.net/conversation/1107497866/151190?folderId=2683093

When **Jetpack's site accelerator** is active, and the **Speed up static file load times** option is enabled
AND
When WP Rocket **Combine JS** is active, **Defer JS** is enabled, **Safe mode for jQuery** is active too
AND 
There is some **inline jQuery** dependant script in the HTML.

These conditions are triggering the `jQuery is not defined` error, because the c0.wp.com/c/5.3.2/wp-includes/js/jquery/jquery.js script is combined and deferred. 

Captura de pantalla 2020-03-31 a las 8 49 58

jQuery is combined and deferred. the inline JS is triggering the error.
Screenshot at mar 31 08-57-44

webtrainingwheels:
My comment on this issue is based on this understanding: 
This PR does one thing: _When Combine JS is active, exclude Jetpack jQuery from being combined._ 
From what I can see, it is not related to or dependent on detecting if Deferred JS Safe Mode is enabled, nor if there is  inline jQuery-depended JavaScript. 

Is my understanding wrong? If so, it could all be a moot point :)
If I'm not wrong, read on......

We do not automatically exclude default jQuery when Combine JS is enabled, so why do we have to do it for Jetpack?

We already exclude Jetpack jQuery from being deferred when Load JS Deferred with Safe Mode is active. In fact we currently seem to have a separate problem related to this which I'll create another issue for. 

But for this PR, if we exclude Jetpack jQuery for everyone using Jetpack site accelerator and Combine JS alone (meaning they don't also have deferred JS w/ safe mode active), we unnecessarily create some render-blocking.

vmanthos:
> We do not automatically exclude default jQuery when Combine JS is enabled, so why do we have to do it for Jetpack?

I see your point here and you are totally correct, as always! 😊

The exclusion should work only if the **Defer JavaScript files** is enabled.

I see that you have already created a GitHub issue (#2493) about that. 
I'm going to close this PR.

webtrainingwheels:
@arunbasillal please check the conclusion of the discussion in #2296 , that issue turned out to be a bit different and led to the creation of #2493 instead.

PR #2297, opened 2020-02-05
Disable WP 5.4 core lazyload when WPR images lazyload is enabled

Tabrisrp:
The feature is not merged yet in core, but to be ahead, we decide to handle this in the plugin using the provided filter `wp_lazy_loading_enabled`.

We will disable core lazyload when our images lazyload is enabled in WP Rocket settings.

### DIFF
--- a/inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
+++ b/inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
@@ -437,11 +437,11 @@ class Lazyload_Subscriber implements Subscriber_Interface {
 	 * @return bool
 	 */
 	public function maybe_disable_core_lazyload( $value ) {
-		if ( $this->can_lazyload_images() ) {
-			return false;
+		if ( false === $value ) {
+			return $value;
 		}
 
-		return $value;
+		return ! (bool) $this->can_lazyload_images();
 	}
 
 	/**
--- a/inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
+++ b/inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
@@ -89,16 +89,17 @@ class Lazyload_Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'wp_footer'            => [
+			'wp_footer'               => [
 				[ 'insert_lazyload_script', PHP_INT_MAX ],
 				[ 'insert_youtube_thumbnail_script', PHP_INT_MAX ],
 			],
-			'wp_head'              => [ 'insert_nojs_style', PHP_INT_MAX ],
-			'wp_enqueue_scripts'   => [ 'insert_youtube_thumbnail_style', PHP_INT_MAX ],
-			'rocket_buffer'        => [ 'lazyload', 18 ],
-			'rocket_lazyload_html' => 'lazyload_responsive',
-			'init'                 => 'lazyload_smilies',
-			'wp'                   => 'deactivate_lazyload_on_specific_posts',
+			'wp_head'                 => [ 'insert_nojs_style', PHP_INT_MAX ],
+			'wp_enqueue_scripts'      => [ 'insert_youtube_thumbnail_style', PHP_INT_MAX ],
+			'rocket_buffer'           => [ 'lazyload', 18 ],
+			'rocket_lazyload_html'    => 'lazyload_responsive',
+			'init'                    => 'lazyload_smilies',
+			'wp'                      => 'deactivate_lazyload_on_specific_posts',
+			'wp_lazy_loading_enabled' => 'maybe_disable_core_lazyload',
 		];
 	}
 
@@ -424,6 +425,22 @@ class Lazyload_Subscriber implements Subscriber_Interface {
 		if ( is_rocket_post_excluded_option( 'lazyload_iframes' ) ) {
 			add_filter( 'do_rocket_lazyload_iframes', '__return_false' );
 		}
+	}
+
+	/**
+	 * Disable WP core lazyload if our images lazyload is active
+	 *
+	 * @since 3.5
+	 * @author Remy Perona
+	 *
+	 * @return bool
+	 */
+	public function maybe_disable_core_lazyload() {
+		if ( $this->can_lazyload_images() ) {
+			return false;
+		}
+
+		return true;
 	}
 
 	/**
--- a/inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
+++ b/inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
@@ -433,14 +433,15 @@ class Lazyload_Subscriber implements Subscriber_Interface {
 	 * @since 3.5
 	 * @author Remy Perona
 	 *
+	 * @param bool $value Current value for the enabling variable.
 	 * @return bool
 	 */
-	public function maybe_disable_core_lazyload() {
+	public function maybe_disable_core_lazyload( $value ) {
 		if ( $this->can_lazyload_images() ) {
 			return false;
 		}
 
-		return true;
+		return $value;
 	}
 
 	/**


inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
```diff
@@ -426,6 +427,22 @@ public function deactivate_lazyload_on_specific_posts() {
...
+	 * @author Remy Perona
+	 *
+	 * @return bool
+	 */
+	public function maybe_disable_core_lazyload() {
+		if ( $this->can_lazyload_images() ) {
+			return false;
+		}
+
+		return true;
```
hellofromtonya: This code can be simplified to return the inverse value of the conditional expression:

```php
return ! (bool) $this->can_lazyload_images();
```

inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
```diff
@@ -426,6 +427,22 @@ public function deactivate_lazyload_on_specific_posts() {
...
 
+	/**
+	 * Disable WP core lazyload if our images lazyload is active
+	 *
+	 * @since 3.5
+	 * @author Remy Perona
+	 *
+	 * @return bool
+	 */
+	public function maybe_disable_core_lazyload() {
```
hellofromtonya: I think this method name is misleading. Why? We are returning `false` to disable it and `true` to enable it. When I first read this method name, I thought the inverse. To keep it crystal clear, I'd suggest changing the name to:

```php
public function maybe_enable_wp_native_lazyload() {
```

Think about it. If `true`, it enables WP native lazyload. That happens when ours cannot lazyload images, i.e. when `can_lazyload_images() === true`.

Tabrisrp: The native lazyload will be enabled by default in core (default value for the filter is true). That's why I felt naming the method like this made more sense: we want to disable core lazyload when our lazyload is active

inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
Screenfeed: @hellofromtonya Is there a purpose to this `(bool)`?

inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
hellofromtonya: Let's walk through what the code is doing and see if the callback _tells_ us the expected behavior:

- WordPress fires a filter called `'wp_lazy_loading_enabled'` and passes a `true` value to turn on and enable this feature.
- Our callback is fired.
- What does it mean when our callback returns `true`? Does `true` === "disable"? No, it means "enable" native core.

By convention, we think about a `true` state as doing the what the method tells us its behavior is and what will happen when it's invoked. So in this case, the expected behavior is not to disable, but to enable, where a `true` being returned means "enable" it.

inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
hellofromtonya: The method `can_lazyload_images()` returns a value through a filter. Therefore, it can return mixed values if some code is hooked into it and returns say a `1` or '0'` or something else. the (bool) here ensures we have a boolean value that is being inverted.

Screenfeed: Isn't the `!` already insuring it? The boolean cast seems redundant to me.

hellofromtonya: With PHP 7, we want to be strict with our data types. Here we are expecting a boolean and then we invert it. But what if we get a string or an integer. Yes, PHP can type juggle, but with PHP 7 we can strive to be very strict with what data type we expect.

Does that make sense?

Screenfeed: Still not. Is there a case where this doesn't return a boolean?
```php
return ! $this->can_lazyload_images();
```
For me, `!` will always cast the result as a boolean, am I wrong?

hellofromtonya: Yes. It's running through a filter. So any callback can return any value and data type.

```php
	/**
	 * Checks if we can lazyload images.
	 *
	 * @since 3.3
	 *
	 * @return boolean
	 */
	private function can_lazyload_images() {
		if ( ! $this->options->get( 'lazyload' ) ) {
			return false;
		}

		/**
		 * Filters the lazyload application on images
		 *
		 * @since 2.0
		 * @author Remy Perona
		 *
		 * @param bool $do_rocket_lazyload True to apply lazyload, false otherwise.
		 */
		return apply_filters( 'do_rocket_lazyload', true ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
	}
```

The way to ensure we always get a boolean is to put the `(bool)` in that method like this:

```php
return (bool) apply_filters( 'do_rocket_lazyload', true ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
```

Screenfeed: I think we don't understand each other.
I agree that a value which is filtered can return any type of data, and that we must cast it as boolean to insure we get the right format, I do it all the time. What I mean is that `!` is already doing this job, so there is no need to add another `(bool)` cast behind it.
This:
```php
return ! (bool) $this->can_lazyload_images();
```
Does exactly the same thing than this:
```php
return ! $this->can_lazyload_images();
```
Here, `(bool)` is useless.

hellofromtonya: >What I mean is that ! is already doing this job, so there is no need to add another (bool) cast behind it.

When the value come back from the method is not a boolean, then yes, `!` is type juggling that value into a boolean and then inverting it.

For example, let's say `$this->can_lazyload_images();` returns a string data type with a value of `'some value'`. What's happening here?

1. With `(bool)` we are type casting to ensure it's a boolean. We are being strict with our intent of the data type.
2. Then the `!` inverts the boolean.

What happens if we don't have the type casting to boolean data type? In PHP 5 (since PHP Is a weakly typed language) we are used to letting these operators do the type juggling to convert it for us into whatever the data type is supposed to be for the operation to continue.

With PHP 7 though, we are moving more and more away from weakly typed where possible and striving to be more purposeful about our data types.

So to answer your question:

1. Yes, if we stick with weakly typed, then yes, PHP will convert whatever is returned from the method into a boolean for us, i.e. without having to use explicitly tell PHP we want a boolean.
2. However, if we want to be more explicit and strict, then we tell PHP (and ourselves) what the data type should be.

hellofromtonya: A better solution here is to ensure a method always returns a known data type. In this case, `can_lazyload_images()` should return a boolean. Then there's no need for explicitly doing type casting like I showed above.

Screenfeed: OK, now I understand the purpose of `(bool)`: is not to prevent type juggling on the method’s output, but on `!`.
Wow, I don't think I like this, but it's another matter.
Thanks for the explanation.

hellofromtonya: You're welcome. Thanks for discussing it with me. Always enjoy a good technical discussion!

inc/classes/subscriber/Optimization/class-lazyload-subscriber.php
Tabrisrp: I updated the method a bit, to take into account the possibility that something else changed the default value before us (we're hooking on the default priority 10, so that's possible).

Instead of returning true/false, I'm returning false if our lazyload is active as that's the behavior we need, but if it's not, I'm returning the value passed to the callback, which can be eithere true or false, but we don't have to care about it.

This also makes the method name more appropriate, since now we disable if we need, but do nothing if we don't.

What do you think?

hellofromtonya: >to take into account the possibility that something else changed the default value before us (we're hooking on the default priority 10, so that's possible).

@Tabrisrp That's a very good point. What if it's already set to `false`? Then we don't need to check if ours is enabled or not. Right? If the given value is `false`, bail out and return it. Else, check if ours is enabled.

PR #2306, opened 2020-02-11
Closes #2465 Compress static assets with gzip

### LINKED ISSUE #2465: Gzip compression for static files
## Why are we doing this feature?
WP Rocket enables gzip compression on activation. Right now we generate the gzip versions of the files on the fly. That is, the file is gzipped when a browser requests it. For huge websites with a lot of traffic, this is inefficient and the file can be gzipped once and served multiple times. 

The impact is minimal for a huge portion of our customer base with smaller websites. But we can do this to be more efficient and improve performance wherever possible. 

## User expectations and experience 
Nothing changes for the end user, the feature will be invisible and under the hood. Smaller websites shouldn't expect any noticeable performance gain either. 

For websites with a lot of traffic might notice lower CPU usage. 

## How will we do this feature? 
- We already gzip compress HTML files. We will do the same for CSS and JS. 
- On Apache, rewrite rules to be added to accommodate for this. (Examples: [here](https://httpd.apache.org/docs/2.4/mod/mod_deflate.html#precompressed) and [here](https://royal.pingdom.com/can-gzip-compression-really-improve-web-performance/))

**Technical overview**
- Update the write_file() method in WP_Rocket\Optimization\Abstract_Optimization to also save a gz file for the same content, with the extension .css.gz or .js.gz
- Add new method to functions/htaccess.php, to generate the htaccess code to serve compressed assets directly, based on the rewrite rules above
- Update functions cleaning the minify/cache busting files in functions/files.php, to also clean the .gz files

This represent a couple days of work, but relies on the first note point below to make sure we can add proper automated tests for it.  

**Notes**
- @everyone: This only applies to files that are minified/combined/cache busted by us. This means that:
- If none of the options above are used, CSS/JS files will **not** be statically compressed
- If only minify is checked, only CSS/JS files minified by us will be statically compressed. Files which already have .min in their filename will be **not** statically compressed
- If combine is checked, the generated CSS/JS files will be staticly compressed
- CSS/JS files without a version query strings will **not** be statically compressed
- Any CSS/JS file excluded from the options above will **not** be statically compressed  

## Risks
- Since we will be generating more static files, the cache_dir_size issue could be an issue. 
- This shouldn't stop us from implementing this though. gzipped file will be small anyways. Plus we need to fix the original issue some other way anyways. 

## Testing
- Load a website on a gzip compatible browser. Make sure there are no 404 or such errors
- Test scenarios where CSS and JS combination is enabled and disabled.

Tabrisrp:
Closes #2465 

**To-Do**
- [x] Save a gzip version of the minified files
- [x] Update `rocket_clean_minify` to delete the gzip files too
- [x] Update `.htaccess` to use the gzip files if they exist
- [x] Unit tests
- [x] Integration tests


tests/Unit/FilesystemTestCase.php
```diff
@@ -7,43 +7,71 @@
 
 abstract class FilesystemTestCase extends VirtualFilesystemTestCase {
 	protected $wprocket_structure = [
```
hellofromtonya: @Tabrisrp Change this to `$structure`. Then it will override the [base structure found here]()https://github.com/wp-media/phpunit/blob/master/Unit/VirtualFilesystemTestCase.php#L22-L30.

tests/Unit/FilesystemTestCase.php
```diff
@@ -7,43 +7,71 @@
...
+					],
+				],
+			]
+		]
 	];
 	protected $cache_path;
 
 	public function setUp() {
-		$this->structure['wp-rocket'] = $this->wprocket_structure;
+		$this->structure['cache'] = $this->wprocket_structure;
```
hellofromtonya: @Tabrisrp Delete this line of code. It's not needed anymore. The parent `setUp()` method will take care of instantiating the filesystem for you. [See it here](https://github.com/wp-media/phpunit/blob/master/Unit/VirtualFilesystemTestCase.php#L53).

tests/Unit/FilesystemTestCase.php
```diff
@@ -6,44 +6,71 @@
...
 	protected $cache_path;
 
 	public function setUp() {
-		$this->structure['wp-rocket'] = $this->wprocket_structure;
 		parent::setUp();
 
 		// Redefine rocket_direct_filesystem() to use the virtual filesystem.
 		Functions\when( 'rocket_direct_filesystem' )->justReturn( $this->filesystem );
-		$this->cache_path = $this->filesystem->getUrl( 'wp-rocket' );
+		$this->cache_path = $this->filesystem->getUrl( 'cache' );
```
hellofromtonya: @Tabrisrp This isn't be used anywhere in the tests. Is it needed?

tests/Unit/Functions/rocketCleanCacheBusting.php
```diff
@@ -0,0 +1,67 @@
...
+use WP_Rocket\Tests\Unit\FilesystemTestCase;
+use Brain\Monkey\Functions;
+
+/**
+ * @covers rocket_clean_cache_busting()
+ * @group Functions
+ * @group Files
+ */
+class Test_RocketCleanCacheBusting extends FilesystemTestCase {
+	protected $rootVirtualDir = 'cache';
```
hellofromtonya: @Tabrisrp `$rootVirtualDir` is set to `'cache'` in the base `VirtualFilesystemTestCase` here https://github.com/wp-media/phpunit/blob/master/Unit/VirtualFilesystemTestCase.php#L15

tests/Unit/Functions/rocketCleanMinify.php
```diff
@@ -0,0 +1,57 @@
...
+use WP_Rocket\Tests\Unit\FilesystemTestCase;
+use Brain\Monkey\Functions;
+
+/**
+ * @covers rocket_clean_minify()
+ * @group Functions
+ * @group Files
+ */
+class Test_RocketCleanMinify extends FilesystemTestCase {
+	protected $rootVirtualDir = 'cache';
```
hellofromtonya: @Tabrisrp `$rootVirtualDir` is set to 'cache' in the base VirtualFilesystemTestCase here https://github.com/wp-media/phpunit/blob/master/Unit/VirtualFilesystemTestCase.php#L15

tests/Unit/bootstrap.php
```diff
@@ -15,7 +15,8 @@
  */
 function load_original_functions_before_mocking() {
 	$originals = [
-		'rocket_get_constant' => require_once WP_ROCKET_PLUGIN_ROOT . 'inc/constants.php',
+		'rocket_get_constant'      => require_once WP_ROCKET_PLUGIN_ROOT . 'inc/constants.php',
+		'rocket_direct_filesystem' => require_once WP_ROCKET_PLUGIN_ROOT . 'inc/functions/files.php'
```
hellofromtonya: @Tabrisrp We don't want to load the file here. Rather, we want a path to the file. It's then loaded within the `foreach` if the function doesn't already exist in memory.

tests/Unit/FilesystemTestCase.php
Tabrisrp: Currently no, this was in the file you initially wrote so I kept it

hellofromtonya: Let's go ahead and delete this. I did add it to the `wp-media/phpunit` base test case, just in case we ever need it. https://github.com/wp-media/phpunit/pull/3

tests/Unit/bootstrap.php
Tabrisrp: I don't understand, I replicated what you did before for `rocket_get_constant`, with `rocket_direct_filesystem `. Else I was getting an error `Cannot redeclare function`

tests/Integration/Functions/rocketGetCompressedAssets.php
```diff
@@ -0,0 +1,79 @@
...
+	];
+
+	public function setUp() {
+		parent::setUp();
+
+		global $is_apache;
+
+		$is_apache = true;
+
+		Functions\when( 'rocket_valid_key' )->justReturn( true );
```
hellofromtonya: @Tabrisrp This function is redefined in the bootstrap file and set to `true`. Did it not work for you?

Tabrisrp: During debugging, I found out that it was returning true on the first call, and other calls were still returning false. So I had to force the value in the test here

hellofromtonya: Good to know. Thank you.

tests/Integration/Subscriber/Optimization/MinifyCssSubscriber/process.php
```diff
@@ -0,0 +1,57 @@
...
+		Functions\expect( 'rocket_get_constant' )
+			->once()
+			->with( 'WP_ROCKET_MINIFY_CACHE_PATH' )
+			->andReturn( 'wp-content/cache/min/' );
+
+		add_filter( 'pre_get_rocket_option_minify_css', '__return_true' );
+		add_filter( 'pre_get_rocket_option_minify_concatenate_css', '__return_true' );
+		add_filter( 'pre_get_rocket_option_minify_css_key', function() {
+			return '123456';
+		} );
```
Screenfeed: These are not reset after the test.

```php
$key_callback = function() {
	return '123456';
};
add_filter( 'pre_get_rocket_option_minify_css', [ $this, 'return_true' ] );
add_filter( 'pre_get_rocket_option_minify_concatenate_css', [ $this, 'return_true' ] );
add_filter( 'pre_get_rocket_option_minify_css_key', $key_callback );

// ...

remove_filter( 'pre_get_rocket_option_minify_css', [ $this, 'return_true' ] );
remove_filter( 'pre_get_rocket_option_minify_concatenate_css', [ $this, 'return_true' ] );
remove_filter( 'pre_get_rocket_option_minify_css_key', $key_callback );
remove_filter( 'rocket_url_to_path', $callback );
```

tests/Integration/Subscriber/Optimization/MinifyCssSubscriber/process.php
```diff
@@ -0,0 +1,64 @@
...
+
+		$html = file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/Optimization/CSS/original.html');
+
+		apply_filters( 'rocket_buffer', $html );
+
+		$this->assertTrue( $this->filesystem->exists( 'cache/min/1/3e1ae3552abbc9a53ab11fc7766f5559.css' ) );
+		$this->assertTrue( $this->filesystem->exists( 'cache/min/1/3e1ae3552abbc9a53ab11fc7766f5559.css.gz' ) );
+
+		remove_filter( 'pre_get_rocket_option_minify_css', '__return_true' );
+		remove_filter( 'pre_get_rocket_option_minify_concatenate_css', '__return_true' );
```
Screenfeed: In [PHPUnit\Integration\TestCase](https://github.com/wp-media/phpunit/commit/4d2ee892c7ab13fd58bd5535b948b658283b40cd) we introduced short methods like `return_true()` for these cases. We should use them instead of WP's `__return_true()`, because when doing `remove_filter()` we could remove a filter added by something else (3rd party compatibility for example).
By using an instantiated method we are sure to remove only the one we added.
```php
		add_filter( 'pre_get_rocket_option_minify_css', [ $this, 'return_true' ] );
		add_filter( 'pre_get_rocket_option_minify_concatenate_css', [ $this, 'return_true' ] );
		// ...
		remove_filter( 'pre_get_rocket_option_minify_css', [ $this, 'return_true' ] );
		remove_filter( 'pre_get_rocket_option_minify_concatenate_css', [ $this, 'return_true' ] );
```

tests/Integration/inc/Functions/rocketCleanCacheBusting.php
```diff
@@ -0,0 +1,71 @@
+<?php
+namespace WP_Rocket\Tests\Integration\inc\Functions;
```
Screenfeed: To fit the plugin’s folders structure, the folder and the namespace for `Functions` should be lowercase.
```suggestion
namespace WP_Rocket\Tests\Integration\inc\functions;
```

Maybe it's a job for another PR though, since it's the same for all tests related to functions.

tests/Integration/Subscriber/Optimization/MinifyCssSubscriber/process.php
```diff
@@ -0,0 +1,64 @@
...
+
+use WP_Rocket\Tests\Integration\FilesystemTestCase;
+use Brain\Monkey\Functions;
+
+/**
+ * @covers \WP_Rocket\Subscriber\Optimization\Minify_CSS_Subscriber::process()
+ * @group Optimization
+ */
+class Test_Process extends FilesystemTestCase {
+	protected $rootVirtualDir = 'wp-content';
```
crystinutzaa: Alignment issue?

GeekPress:
@Tabrisrp I created the official « issue » about this enhancement:
https://github.com/wp-media/wp-rocket/issues/2465

tests/Integration/inc/Engine/Optimization/Minify/CSS/Subscriber/process.php
```diff
@@ -24,9 +25,13 @@ public function testShouldMinifyCSS( $original, $expected, $settings ) {
...
 		$this->assertSame(
-			$this->format_the_html( $expected ),
+			$this->format_the_html( $expected['html'] ),
 			$this->format_the_html( apply_filters( 'rocket_buffer', $original ) )
-		);
+        );
+
+        foreach( $expected['files'] as $file ) {
+            $this->assertTrue( $this->filesystem->exists( $file ) );
+        }
```
Screenfeed, line 34:
```suggestion
		);

		foreach( $expected['files'] as $file ) {
			$this->assertTrue( $this->filesystem->exists( $file ) );
		}
```

inc/Engine/Optimization/Minify/CSS/Combine.php
```diff
@@ -154,7 +154,7 @@ function( $style ) use ( &$files ) {
 	 * @param array  $to_remove An array of CSS tags to remove.
 	 * @return string
 	 */
-	public function insert_combined_css( $html, $css_url, array $to_remove ) {
+	protected function insert_combined_css( $html, $css_url, array $to_remove ) {
```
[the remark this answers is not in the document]
Tabrisrp: This method was made public in the first place to be testable, it was never really meant to be used outside of the class itself, and it doesn't make sense to use it on its own either.

I can revert back to `public` if we're really worried about someone using it

tests/Unit/inc/Engine/Optimization/Minify/JS/Combine/optimize.php
```diff
@@ -74,5 +74,8 @@ public function testShouldCombineJS( $original, $minified, $cdn_hosts, $cdn_url,
 			$this->format_the_html( $minified ),
 			$this->format_the_html( $this->combine->optimize( $original ) )
 		);
+
+		$this->assertTrue( $this->filesystem->exists( 'wordpress/wp-content/cache/min/1/900b339c19ff5a927b3311bf5ddb4dfd.js' ) );
```
crystinutzaa, line 78: The js filename could be moved in the Fixtures. What do you think?

tests/Unit/inc/Engine/Optimization/Minify/JS/Minify/optimize.php
```diff
@@ -63,5 +63,10 @@ public function testShouldMinifyJS( $original, $minified, $cdn_hosts, $cdn_url,
 			$minified,
 			$this->minify->optimize( $original )
 		);
+
+		$this->assertTrue( $this->filesystem->exists( 'wordpress/wp-content/cache/min/1/wp-content/themes/twentytwenty/assets/script-09b5ce74889313bd51265ef983880c47.js' ) );
```
crystinutzaa, line 67: The same in here, the JS Minify files can be moved in Fixtures in an array. What do you think?

tests/Unit/inc/Engine/Optimization/Minify/CSS/Minify/optimize.php
```diff
@@ -55,5 +55,10 @@ public function testShouldMinifyCSS( $original, $minified, $cdn_host, $cdn_url,
 			$minified,
 			$this->minify->optimize( $original )
 		);
+
+		$this->assertTrue( $this->filesystem->exists( 'wordpress/wp-content/cache/min/1/wp-content/themes/twentytwenty/style-dcd1a95e5d432b5d300d7c2f216d7150.css' ) );
```
crystinutzaa, line 59: Move the files in fixtures and use a loop to parse and make the assertTrue. What do you think?

tests/Unit/inc/Engine/Optimization/Minify/CSS/Combine/optimize.php
```diff
@@ -63,5 +63,8 @@ public function testShouldCombineCSS( $original, $combined, $cdn_host, $cdn_url,
 			$combined,
 			$this->combine->optimize( $original )
 		);
+
+		$this->assertTrue( $this->filesystem->exists( 'wordpress/wp-content/cache/min/1/0bca286481748a69cdbe9a6695015ec9.css' ) );
```
crystinutzaa, line 67: Move CSS files in fixtures and use a loop to parse those and make the assertTrue. What do you think?

tests/Unit/inc/Engine/Optimization/Minify/CSS/Combine/optimize.php
```diff
@@ -60,8 +60,15 @@ public function testShouldCombineCSS( $original, $combined, $cdn_host, $cdn_url,
...
-			$combined,
+			$expected['html'],
 			$this->combine->optimize( $original )
 		);
+
+		foreach ( $expected['files'] as $file ) {
+			$this->assertTrue( $this->filesystem->exists( $file ) );
+		}
+
+		$this->assertTrue( $this->filesystem->exists( 'wordpress/wp-content/cache/min/1/0bca286481748a69cdbe9a6695015ec9.css' ) );
```
crystinutzaa, line 71: These 2 assertions should be removed, right?

Tabrisrp: You're right I forgot to delete them, just pushed the commit for it

Tabrisrp:
@wp-media/dev-qa ready for QA

arunbasillal:
Starting QA

arunbasillal:
Tests:

- Deactivated and reactivated WP Rocket. The new rules were added to `.htaccess` 👍 
- Enabled CSS and JS minification. A .gz version of the minified file was generated alongside the normal version 👍 
- Enabled CSS and JS combination. A .gz version of the combined file was generated alongside the normal version 👍 

But I don't think the files are being served. Here is a test: https://youtu.be/YGoKrjg1crc
@Tabrisrp please note ^

Tabrisrp:
@arunbasillal I installed the changes from this branch on http://mega.wp-rocket.me/benchmark/

And set a no gzip directive in the `htaccess` to prevent dynamic compression. You can check that the gz files are served there.

arunbasillal:
@Tabrisrp Thanks. I had already tested this by disabling the gzip mod. Details on the comment right after this one: https://github.com/wp-media/wp-rocket/pull/2306#issuecomment-614002648. All good from my end 👍 

Looks like this PR is waiting on @hellofromtonya's review for a merge?

PR #2307, opened 2020-02-12
Update the list of nonce actions which need to revert the $uid

vmanthos:
**Related ticket:** https://secure.helpscout.net/conversation/1076171005/144793?folderId=273766


vmanthos:
I merged this PR to #2479.

PR #2308, opened 2020-02-12
Mobile preload

Screenfeed:
When preloading, also preload for mobile by sending a mobile user agent.
Mobile preload is enabled only when the 3 following options are enabled: preload (`manual_preload`), mobile caching (`cache_mobile`), and separate cache for mobile devices (`do_caching_mobile_files`). It works for both homepage preload and sitemap preload.
The feature can be disabled by using the filter `rocket_mobile_preload_enabled`.

About sitemap:
Some sitemaps can be dedicated to mobile pages ([see here](https://robots-txt.com/sitemaps/sitemap-mobile/)). In this case, a mobile header will be sent even if mobile preload is not enabled. This changes from the previous behavior, as mobile sitemaps were not detected so far.

For the occasion an abstract class `WP_Rocket\Preload\Process` is created to reunite common parts for partial and full process.

Other notes:
- Unrelated commits, to make tests work: 0138d25d1549ad5f7e9ddecce7043d332b641b4b, 5304e7c75772f7a1566a508431cade66a05fe404.
- Tagged for WPR 3.6.


hellofromtonya:
@Screenfeed I merged in the new test suite changes from `develop` and updated to the newest version of `wp-media/phpunit`. I'm curious to see if it solves the local test runs on your machine.

Screenfeed:
Yep, fixed 👍

inc/classes/preload/class-abstract-process.php
```diff
@@ -0,0 +1,213 @@
...
+	 *     The item to preload: an array containing the following values.
+	 *
+	 *     @type string $url    The URL to preload.
+	 *     @type bool   $mobile True when we want to send a "mobile" user agent with the request. Optional.
+	 * }
+	 * @return string
+	 */
+	public function get_item_user_agent( $item = [] ) {
+		if ( $item['mobile'] ) {
+			return 'WP Rocket/Preload iPhone';
```
hellofromtonya: Why iPhone?

Screenfeed: Because it’s the [first value in the mobile devices](https://github.com/wp-media/wp-rocket/blob/a37e46f58259e38dbf895061536551763f8aace7/inc/vendors/classes/class-rocket-mobile-detect.php#L172).

Screenfeed: To clarify (sorry I wasn't clear enough), this is the library WPR uses to detect mobile devices. It is used to separate mobile and non mobile caches, so I use the same source for my tests.

inc/classes/preload/class-sitemap.php
```diff
@@ -211,54 +212,100 @@ public function process_sitemap( $sitemap_url, $urls = [] ) {
...
-			return [];
+			return $urls;
 		}
 
 		$url_count        = count( $xml->url );
 		$sitemap_children = count( $xml->sitemap );
 
 		if ( $url_count > 0 ) {
+			$mobile_preload = $this->preload_process->is_mobile_preload_enabled();
+			$mobile_suffix  = '##wpm-mobile##';
```
Tabrisrp: I'm thinking you could define a class constant in the parent class for the mobile suffix, to improve maintainability

tests/Unit/Preload/Sitemap/runPreload.php
```diff
@@ -0,0 +1,110 @@
...
+
+	public function testShouldNotPreloadWhenNoUrls() {
+		$preload_process = $this->createMock( Full_Process::class );
+
+		Actions\expectDone( 'before_run_rocket_sitemap_preload' )->never();
+
+		// No URLs.
+		( new Sitemap( $preload_process ) )->run_preload( [] );
+
+		$this->assertTrue( true );
```
Tabrisrp: You should not need this line

inc/classes/preload/class-abstract-preload.php
```diff
@@ -55,4 +66,39 @@ public function cancel_preload() {
...
+	 * @return string
+	 */
+	protected function get_url_identifier( $url ) {
+		if ( ! isset( $this->cache_query_strings ) ) {
+			$this->cache_query_strings = array_fill_keys( get_rocket_cache_query_string(), '' );
+
+			ksort( $this->cache_query_strings );
+		}
+
+		$path  = (array) wp_parse_url( $url );
```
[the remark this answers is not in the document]
Screenfeed: Yes.

tests/Integration/inc/classes/preload/Homepage/preload.php
```diff
@@ -0,0 +1,146 @@
...
+					break;
+				case 'https://example.com/foobar/':
+					$file_name = 'foobar';
+					break;
+				case 'https://example.com/category/barbaz/':
+					$file_name = 'category-barbaz';
+					break;
+				default:
+					return new \WP_Error( 'wrong-url', 'Wrong URL', [ $url ] );
+			}
```
hellofromtonya: @Screenfeed We have a virtual filesystem now that can be used to build up the cache directory and files. Instead of faking the remote get, instead use the virtual filesystem, save the fixture content into each mobile cache file, and then let the code run as it normally would.

Why? Faking here might mask a problem. By using the virtual filesystem, you are adding cached mobile files and allowing the code to run normally.

tests/Integration/inc/classes/preload/Sitemap/runPreload.php
```diff
@@ -0,0 +1,244 @@
...
+					$path .= 'sitemap.xml';
+					break;
+				case 'https://example.com/sitemap-mobile.xml':
+					$path .= 'sitemap-mobile.xml';
+					break;
+				default:
+					return new \WP_Error( 'wrong-url', 'Wrong URL', [ $url ] );
+			}
+
+			if ( ! \file_exists( $path ) ) {
```
hellofromtonya: No need for the `\` as `file_exists()` is a PHP native function that is in global space. PHP will know how to find it.

tests/Integration/inc/classes/preload/Sitemap/runPreload.php
```diff
@@ -0,0 +1,244 @@
...
+			switch ( $url ) {
+				case 'https://example.com/sitemap.xml':
+					$path .= 'sitemap.xml';
+					break;
+				case 'https://example.com/sitemap-mobile.xml':
+					$path .= 'sitemap-mobile.xml';
+					break;
+				default:
+					return new \WP_Error( 'wrong-url', 'Wrong URL', [ $url ] );
+			}
```
hellofromtonya: Same here. Use the virtual fileystem instead. Don't fake the request. Let the code run as it normally would and then validate the behavior.

tests/Integration/inc/classes/preload/Sitemap/runPreload.php
```diff
@@ -0,0 +1,244 @@
...
+
+			if ( ! \file_exists( $path ) ) {
+				return new \WP_Error( 'file-not-found', 'File not found', [ basename( $path ) ] );
```
hellofromtonya: Import `WP_Error` instead.

tests/Integration/inc/classes/preload/Sitemap/runPreload.php
```diff
@@ -0,0 +1,244 @@
...
+					$path .= 'sitemap.xml';
+					break;
+				// sitemap-mobile.xml will return an error, and will trigger get_urls().
+				default:
+					return new \WP_Error( 'wrong-url', 'Wrong URL', [ $url ] );
+			}
+
+			if ( ! \file_exists( $path ) ) {
+				return new \WP_Error( 'file-not-found', 'File not found', [ basename( $path ) ] );
+			}
```
hellofromtonya: Same here. Use the virtual filesystem instead.

tests/Integration/inc/classes/preload/Sitemap/runPreload.php
```diff
@@ -0,0 +1,244 @@
...
+
+			return [
+				'headers'       => null, // Requests_Utility_CaseInsensitiveDictionary object.
+				'body'          => \file_get_contents( $path ),
```
hellofromtonya: No need for the global namespace `\` on `file_get_contents()`.

tests/Unit/inc/classes/preload/Sitemap/runPreload.php
```diff
@@ -0,0 +1,107 @@
...
+
+		// Stubs for $this->process_sitemap().
+		Functions\when( 'esc_url_raw' )->returnArg();
+		Functions\when( 'wp_remote_get' )->alias( function( $url, $args = [] ) {
+			switch ( $url ) {
+				case 'https://example.com/sitemap.xml':
+					return [ 'body' => \file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/Preload/Sitemap/sitemap.xml' ) ];
+				case 'https://example.com/sitemap-mobile.xml':
+					return [ 'body' => \file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/Preload/Sitemap/sitemap-mobile.xml' ) ];
+			}
```
hellofromtonya: No need for `\` for `file_get_contents()`.

tests/Integration/inc/classes/preload/Homepage/preload.php
Screenfeed: @hellofromtonya OK for the filesystem, but what about the request itself? As I expected, I get `cURL error 7`.

tests/Integration/inc/classes/preload/Full_Process/isMobilePreloadEnabled.php
```diff
@@ -0,0 +1,94 @@
+<?php
+namespace WP_Rocket\Tests\Integration\inc\classes\preload\Full_Process;
+
+use WPMedia\PHPUnit\Integration\TestCase;
```
hellofromtonya: Same thing here.

Tabrisrp:
Base branch needs to be changed to 3.5, and all `@since` to 3.5 too before merging

Screenfeed:
39 conflicts, live tests to do, I guess we're still far from releasing.

hellofromtonya:
@Screenfeed Would it be easier to create a new branch built off of `branch-3.5` and move the code from here to it? No merge conflicts then to deal with. And that new PR can reference this one for history.

Screenfeed:
> Base branch needs to be changed to 3.5, and all `@since` to 3.5 too before merging

@Tabrisrp There were no `@since 3.6` left already, see 72887df.

Screenfeed:
Great, tests are broken now.

Screenfeed:
Tested on `smashingcoding`:
- tested with separate mobile caching (mobile cache files are created) and without (they are not created).
- tested with homepage only and with sitemap.
Only one remark: if mobile caching is not enabled, a "mobile only" sitemap will still be preloaded. However I don’t think this should block the release, since mobile sitemaps were not supported until now (there were no distinction from classic sitemaps), and not supposed to be.

Tabrisrp:
Technically it's all good, it works as expected.

We do have an issue with the pages preloaded counter, because when the mobile preload is enabled, it's showing double the number of pages.

This is happening because desktop & mobile are treated as different pages. The inaccuracy is likely to cause support tickets.

hellofromtonya:
>We do have an issue with the pages preloaded counter, because when the mobile preload is enabled, it's showing double the number of pages.
>
>This is happening because desktop & mobile are treated as different pages. The inaccuracy is likely to cause support tickets.

@Screenfeed Are you working on this 👆 ?

Screenfeed:
I'm not. Should I?

hellofromtonya:
@Screenfeed Yes, please.

Screenfeed:
Ok, don't expect it right away, the counter is already inaccurate, I need to change a few things.

hellofromtonya:
Since the counter was inaccurate _before_ this feature, @Screenfeed let’s do it as a separate issue and PR. This PR then is approved and ready for merge into 3.5.

PR #2309, opened 2020-02-12
Excluded Dynamic Inline JS

piotrbak:
Excluded Dynamic Inline JS:
wp_post_blocks_vars.listed_posts=[

https://www.diffchecker.com/nXuEVYpQ


piotrbak:
Moved it here:
https://github.com/wp-media/wp-rocket/pull/2843

PR #2317, opened 2020-02-14
Fix #2245 - Rename GZIP caching file with .gz instead of _gzip for nginx

### LINKED ISSUE #2245: Rename GZIP caching file with .gz instead of _gzip
On NGINX server, it seems that serving our caching file with the extension `.gz` instead of `_gzip` helps to deliver our caching files faster.

@arunbasillal @hellofromtonya @Tabrisrp Is it possible to do some benchmark to validate that?

Related issue:
https://github.com/wp-media/wp-rocket/issues/1213

Related conversations:
https://secure.helpscout.net/conversation/1047434108/139419/
https://secure.helpscout.net/conversation/1053040835/140458?folderId=273761

crystinutzaa:
Rename GZIP caching file with .gz instead of _gzip for nginx


Tabrisrp:
I don't like much the idea of maintaining 2 different naming conventions depending on the server type. It's making the code more complicated, and it will likely bring more confusion to our users and support

hellofromtonya:
@Tabrisrp Apache doesn't like the .gz naming convention. It's popping a download modal instead of loading the file. Have we found a solution yet for that problem?

Tabrisrp:
I did have a working solution for Apache and the rewrite rules issue. What I don't know is:
- if it's the optimal solution
- if there is any side-effects to this solution

That's why I think we should take more time to refine & tests to get to a point where the solution looks & works great.

arunbasillal:
Ideally we want to have the simplest solution with just one naming convention irrespective of the webserver. 

However, if that is too risky, these are some things we can consider: 

- Ignore NGINX completely and suit things only for Apache. NGINX needs custom config based on what we know right now. 
- Have two separate file extensions based on webserver. Our primary goal is performance. User confusion is expected, but that should be secondary. 

Either way, let's not rush things, take our time with it, discuss this and see what we can come up with.

inc/common/purge.php
```diff
@@ -83,10 +83,10 @@ function rocket_get_purge_urls( $post_id, $post ) {
 			if ( is_ssl() ) {
 				$filename .= '-https';
 			}
-
+			global $is_nginx;
```
Screenfeed: Not used later? (same observation for other places)

crystinutzaa: Thank you! I wanted to removed those, but I forgot on the way

Screenfeed (approved):

dillix:
This change can break sites which use https://github.com/SatelliteWP/rocket-nginx

PR #2319, opened 2020-02-14
RocketCDN code improvements

Tabrisrp:
- Add method to preserve the authorization token
- Add label for cname field
- Remove prefix on rocketCDN JS functions
- Various code improvements


inc/classes/subscriber/CDN/RocketCDN/DataManagerSubscriber.php
```diff
@@ -79,6 +79,10 @@ public function update_user_token() {
 			return;
 		}
 
+		if ( ! is_string( $_POST['value'] ) ) {
+			wp_send_json_error( 'invalid_token' );
```
Screenfeed: No `return;` this time?

inc/classes/subscriber/CDN/RocketCDN/DataManagerSubscriber.php
```diff
@@ -50,7 +50,7 @@ public static function get_subscribed_events() {
 			'wp_ajax_rocketcdn_process_set'    => 'set_process_status',
 			'wp_ajax_rocketcdn_process_status' => 'get_process_status',
 			self::CRON_EVENT                   => 'maybe_disable_cdn',
-			'update_option_' . WP_ROCKET_SLUG  => [ 'maybe_update_api_status', 12, 2 ],
+			'update_option_' . rocket_get_constant( 'WP_ROCKET_SLUG' ) => [ 'maybe_update_api_status', 12, 2 ],
```
Screenfeed: Now the signs are not aligned anymore :s

inc/classes/CDN/RocketCDN/APIClient.php
```diff
@@ -278,4 +278,41 @@ public function update_website_status( $is_active ) {
...
+	 * @return array
+	 */
+	public function preserve_authorization_token( $args, $url ) {
+		if ( strpos( $url, self::ROCKETCDN_API ) === false ) {
+			return $args;
+		}
+
+		if ( empty( $args['headers']['Authorization'] ) ) {
+			return $args;
+		}
```
Screenfeed: So if a plugin completely removes the header...
Shouldn't we add it back anyway?

inc/classes/CDN/RocketCDN/APIClient.php
```diff
@@ -78,7 +78,7 @@ private function get_remote_subscription_data() {
 		}
 
 		$data = json_decode( $data, true );
-		$data = array_intersect_key( $data, $default );
+		$data = array_intersect_key( (array) $data, $default );
```
Screenfeed: To make sure to not trigger a notice when doing `if ( 'running' === $subscription_data['subscription_status'] ) {`:
```suggestion
		$data = array_intersect_key( (array) $data, $default );
		$data = array_merge( $default, $data );
```

inc/classes/subscriber/CDN/RocketCDN/DataManagerSubscriber.php
Tabrisrp: The sniffers are not triggered on that one, I think the structure of the key throw them off

PR #2324, opened 2020-02-18
Improves test bootstrapping

hellofromtonya:
This PR does the following:

- Changes the composer test scripts to run `phpunit` directly
- Changes to make Rocket first and then it loads `wp-media/phpunit` bootstrap files
- Solves the integration test database error problem. Credit @crystinutzaa 
- Moves WooCommerce tests to a separate group and composer script

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -88,13 +88,13 @@
 		"wp-media/cloudflare": "self.version"
 	},
 	"scripts": {
-		"test-unit": "\"vendor/bin/wpmedia-phpunit\" unit path=tests/Unit",
-		"test-integration": "\"vendor/bin/wpmedia-phpunit\" integration path=tests/Integration --exclude-group AdminOnly",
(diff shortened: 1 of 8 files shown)


hellofromtonya:
@Tabrisrp This PR fixes a couple of weird problems we were getting in Travis with the database errors, vars, etc. Weird stuff. 

The db errors were due to WooCommerce, which I think you first said. Cristina fixed that. But then instead of loading Woo for all tests, we only need it for certain tests. So I made it a separate group. That'll speed things up.

Also, you'll notice that the composer scripts are different. That's on purpose. Why?

1. For now, we'll need a way to exclude multiple groups or even possibly load a different phpunit.xml file. For example, the Cloudflare addon will need its own in order to set `do_cloudflare` to `true` and have it load up in the container and event manager.
3. This will allow us to run separate processes too if we need them. Later, we can add that capability to our phpunit library.

PR #2326, opened 2020-02-18
Hide RocketCDN on local/staging sites

Tabrisrp:
We don't want users to signup for RocketCDN on localhost or staging sites. This PR adds:
- a function `rocket_is_live_site()` to check if the current URL corresponds to localhost or known staging domains
- Checks in method displaying content related to RocketCDN on the settings page
- Updated unit & integration tests

**To-Do**
- [x] Add new tests for when `rocket_is_live_site()` is false
- [x] Test on live staging site. Done. Tested on our Dev live staging site. Works as expected.

### DIFF
--- a/inc/functions/api.php
+++ b/inc/functions/api.php
@@ -73,3 +73,75 @@ function get_rocket_cdn_cnames( $zone = 'all' ) { // phpcs:ignore WordPress.Nami
 
 	return $hosts;
 }
+
+/**
+ * Check if the current URL is for a live site (not local, not staging)
+ *
+ * @since 3.5
+ * @author Remy Perona
+ *
+ * @return bool True if live, false otherwise.
+ */
+function rocket_is_live_site() {
+	if ( rocket_get_constant( 'WP_ROCKET_DEBUG' ) ) {
+		return true;
+	}
+
+	$host = wp_parse_url( home_url(), PHP_URL_HOST );
+
+	if ( ! $host ) {
+		return false;
+	}
+
+	$localhost = [
+		'127.0.0.1',
+		'localhost',
+	];
+
+	if ( in_array( $host, $localhost, true ) ) {
+		return false;
+	}
+
+	$tld           = pathinfo( $host, PATHINFO_EXTENSION );
+	$excluded_tlds = [
+		'localhost',
+		'local',
+		'dev',
+		'test',
+		'docksal',
+	];
+
+	if ( in_array( $tld, $excluded_tlds, true ) ) {
+		return false;
+	}
+
+	if ( '.dev.cc' === substr( $host, -7 ) ) {
+		return false;
+	}
+
+	if ( '.lndo.site' === substr( $host, -10 ) ) {
+		return false;
+	}
+
+	$staging = [
+		'.wpengine.com',
+		'.pantheonsite.io',
+		'.flywheelsites.com',
+		'.flywheelstaging.com',
+		'.kinsta.com',
+		'.kinsta.cloud',
+		'.cloudwaysapps.com',
+		'.azurewebsites.net',
+		'.wpserveur.net',
+		'-liquidwebsites.com',
+		'.myftpupload.com',
+	];
+
+	foreach ( $staging as $domain ) {
+		if ( strpos( $host, $domain ) ) {
+			return false;
+		}
+	}
+
+	return true;
+}
--- a/inc/functions/api.php
+++ b/inc/functions/api.php
@@ -137,8 +137,8 @@ function rocket_is_live_site() {
 		'.myftpupload.com',
 	];
 
-	foreach ( $staging as $domain ) {
-		if ( strpos( $host, $domain ) ) {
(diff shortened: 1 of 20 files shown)


inc/functions/api.php
```diff
@@ -73,3 +73,75 @@ function get_rocket_cdn_cnames( $zone = 'all' ) { // phpcs:ignore WordPress.Nami
...
+/**
+ * Check if the current URL is for a live site (not local, not staging)
+ *
+ * @since 3.5
+ * @author Remy Perona
+ *
+ * @return bool True if live, false otherwise.
+ */
+function rocket_is_live_site() {
+	if ( rocket_get_constant( 'WP_ROCKET_DEBUG' ) ) {
```
hellofromtonya: Why does this constant assure us that it's a live site and not local or staging?

Tabrisrp: It doesn't, but we need to see everything when this constant is set, as it's the one we use for debugging.

It's also a constant that is supposed to be used only for a short period of time (debug time), not be kept on forever.

I felt it could fit there in those conditions.

hellofromtonya: If a dev or customer adds that to their staging or local site, then it'll short circuit and give them access.

What if we add an "include" list of domains that can use this constant, such as our staging and dev sites? And we need a way for our CI pipeline to run it too.

hellofromtonya: Or maybe I should ask this:

What is the risk if a dev or customer adds the constant to their staging or local site?

Tabrisrp: We might need to temporarily use this constant on customers website to do debugging, so an allowlist would restrict us too much.

The risk is that they will see the RocketCDN banners and call to action, so they could potentially try to subscribe for those sites if they go through all the process on purpose.

hellofromtonya: Both are true. Let's monitor it and see if we need adjust.

inc/functions/api.php
```diff
@@ -73,3 +73,75 @@ function get_rocket_cdn_cnames( $zone = 'all' ) { // phpcs:ignore WordPress.Nami
...
+		'.flywheelsites.com',
+		'.flywheelstaging.com',
+		'.kinsta.com',
+		'.kinsta.cloud',
+		'.cloudwaysapps.com',
+		'.azurewebsites.net',
+		'.wpserveur.net',
+		'-liquidwebsites.com',
+		'.myftpupload.com',
+	];
```
hellofromtonya: What about cases where a customer _needs it run_ on their staging environment? For example, some sites use staging to validate changes against performance, regression, and other metrics. How might they do this? I'm thinking this might be more of a product question.

hellofromtonya: Discussion in slack:

1. For now, we'll hide RocketCDN on staging as you have it here.
2. We'll monitor customer requests to determine if there's demand or need for this on staging.

inc/functions/api.php
```diff
@@ -73,3 +73,75 @@ function get_rocket_cdn_cnames( $zone = 'all' ) { // phpcs:ignore WordPress.Nami
...
+
+	foreach ( $staging as $domain ) {
+		if ( strpos( $host, $domain ) ) {
```
hellofromtonya:
```php
if ( strpos( $host, $domain ) !== false ) {
```

👆 this indicates the `$domain` was found in the `$host`.

Tabrisrp: maybe `$domain` is not the best name for the variable here, we're looking if the partial of the host in the array list is contained inside `$host`.

hellofromtonya: So maybe `$partial_host` instead.

Tabrisrp:
Re-requesting reviews after latest changes

tests/Integration/Subscriber/CDN/RocketCDN/NoticesSubscriber/displayRocketcdnCta.php
```diff
@@ -18,9 +18,52 @@ private function getActualHtml() {
...
+					</div>
+				</div>
+			</section>
+			<div class="wpr-rocketcdn-cta-footer">
+				<a href="https://go.wp-rocket.me/rocket-cdn" target="_blank" rel="noopener noreferrer">Learn more about RocketCDN</a>
+			</div>
+			<button class="wpr-rocketcdn-cta-close--no-promo" id="wpr-rocketcdn-close-cta">
+				<span class="screen-reader-text">Reduce this banner</span>
+			</button>
+		</div>' );
```
Screenfeed: Shouldn't this html live in a fixture file?

hellofromtonya: If it's reusable for other tests, then yes. A fixture gives us the ability to pull in a fixture for multiple tests. 

If no, then it's okay to have it directly in the test itself. Why? One advantage is: you can quickly see the HTML that should be generated, i.e. the expected behavior.

PR #2331, opened 2020-02-20
Fixes Cloudflare fatal errors on upgrade.

hellofromtonya:
This PR fixes the fatal errors QA got during 3.5 upgrade. Changes:

1. Updates Rocket to the latest version of https://github.com/wp-media/cloudflare/commit/87eb63a5d21e39a2bcb29b8c0f406f414c593da8
    - Makes `/ips` endpoint not need API credentials
    - Removes `"do_cloudflare"` from the package
2. Does not load the addon when `do_cloudflare` is not enabled, meaning the customer has not turned on the addon.
3. Moves the addons service provider into the `inc/Addon` module.

## TODO

- [x] Ensure it passes phpcs
- [x] Delete the `'rocket_cloudflare_is_api_keys_valid'` transient when first turning on the addon.
- [x] Live test


hellofromtonya:
Results of Live Test:

- Installed with no issues.
- No issues when turning the addon on and off over and over again.
- No issues when adding or removing API credentials.
- Turned on Development mode. Worked and registered in the CF dashboard too.
Screen Shot 2020-02-20 at 4 15 33 PM

- Turned off Development mode
Screen Shot 2020-02-20 at 4 16 07 PM

- No issues when leaving development mode on, then turning off the addon, and then turning it back on again.
- Credentials stay in the database and populate when turning off and one the addon. No problems.
- No errors in the server log.

inc/Addon/ServiceProvider.php
```diff
@@ -66,6 +62,23 @@ public function register() {
...
+	}
+
+	/**
+	 * Adds Cloudflare Addon into the Container when the addon is enabled.
+	 *
+	 * @since 3.5
+	 */
+	protected function addon_cloudflare() {
+		// If the addon is not enabled, delete the transient and bail out. Don't load the addon.
+		if ( ! get_rocket_option( 'do_cloudflare' ) ) {
```
Tabrisrp: Wouldn't it be faster to use the Options_Data instance that is in the container, instead of using `get_rocket_option` which will create a new instance?

inc/classes/class-plugin.php
```diff
@@ -154,13 +154,16 @@ function() {
...
 			'capabilities_subscriber',
 			'varnish_subscriber',
-			'cloudflare_subscriber',
 			'rocketcdn_rest_subscriber',
 			'detect_missing_tags_subscriber',
 			'purge_actions_subscriber',
 		];
 
-		if ( \rocket_valid_key() ) {
+		if ( get_rocket_option( 'do_cloudflare' ) ) {
```
Tabrisrp: same here

hellofromtonya:
More Results from Live Testing:

1. Delete the credentials and saved.
2. Turned off the addon.
3. Refreshed the page.
4. Turned it the addon back on.
5. Turned on Development mode. The toggle shows "ON", but:
     - It did not turn it on in Cloudflare. <- this is the correct behavior
     - there's no indication why it says "ON" but it's not
6. Saved the settings.
7. and got these error notices:

Screen Shot 2020-02-20 at 5 11 53 PM

But notice that the toggle is still "ON" even though it isn't.

inc/Addon/ServiceProvider.php
hellofromtonya: Yes, you're right. I'll change that.

hellofromtonya: @Tabrisrp After looking at how we've got options wired into the container, it's not a singleton. Each time we do `$this->getContainer()->get( 'options' )`, it creates a new instance of `Options_Data`. The behavior is the same as using `get_rocket_option()` with the exception that it doesn't have to go through the container itself.

To test this for yourself, run this code:

```php
$container = apply_filters( 'rocket_container', '' );
var_dump( spl_object_id( $container->get( 'options' ) ) );
var_dump( spl_object_id( $container->get( 'options' ) ) );
```

You'll get a different object ID from each.

Therefore, in this case, `get_rocket_option()` will actually be a teeny tiny bit faster.

inc/classes/class-plugin.php
hellofromtonya: See my notes above. It's actually a little faster with this function.

inc/Addon/ServiceProvider.php
Tabrisrp: Ah right, I forgot it was not returning the same instance each time on this. Does it make sense though? We don't really need a new instance each time we get this.

It's nitpicking at this point since it won't make a big difference. My feeling is that it feels better not mixing container code with some rocket specific code we have when we could have container code only there. This can be looked at later too.

Tabrisrp: Ah right, I forgot it was not returning the same instance each time on this. Does it make sense though? We don't really need a new instance each time we get this.

It's nitpicking at this point since it won't make a big difference. My feeling is that it feels better not mixing container code with some rocket specific code we have when we could have container code only there. This can be looked at later too.

PR #2339, opened 2020-02-22
Adds a filter to rocket_is_live_site() for dev TLDs to be live prod site.

hellofromtonya:
This PR adds a filter `'rocket_tld_is_live_prod_site'` for `.dev` and `.dev.cc` TLDs. 

- By default, these TLDs are considered development purposes only. 
- A customer or dev can register a `__return_true` callback to indicate that their `.dev` or `.dev.cc` TLD is the real live PROD website (and not for local or development site).

Why? We currently have over 4,500 customers with dev TLDs. This filter gives them a way to use RocketCDN.

### DIFF
--- a/inc/functions/api.php
+++ b/inc/functions/api.php
@@ -106,7 +106,6 @@ function rocket_is_live_site() {
 	$excluded_tlds = [
 		'localhost',
 		'local',
-		'dev',
 		'test',
 		'docksal',
 	];
@@ -115,8 +114,15 @@ function rocket_is_live_site() {
 		return false;
 	}
 
-	if ( '.dev.cc' === substr( $host, -7 ) ) {
-		return false;
+	if ( 'dev' === $tld || '.dev.cc' === substr( $host, -7 ) ) {
+		/**
+		 * Indicates if this website's .dev TLD is the real live production website, i.e. not staging or local dev.
+		 *
+		 * @since 3.5
+		 *
+		 * @param bool True indicates .dev is the real live PROD website.
+		 */
+		return (bool) apply_filters( 'rocket_tld_is_live_prod_site', false );
 	}
 
 	if ( '.lndo.site' === substr( $host, -10 ) ) {
--- a/tests/Integration/Functions/rocketIsLiveSite.php
+++ b/tests/Integration/Functions/rocketIsLiveSite.php
@@ -10,6 +10,7 @@ use Brain\Monkey\Functions;
  * @group API
  */
 class Test_RocketIsLiveSite extends TestCase {
+
 	public function testShouldReturnTrueWhenWPROCKETDEBUG() {
 		Functions\when( 'rocket_get_constant' )->justReturn( true );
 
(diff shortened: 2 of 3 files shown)


arunbasillal:
Instead of adding a specific filter for `.dev` and `.dev.cc` why not adding make `$excluded_tlds` filterable? This could be more future proof and would cover every case. 

What do you think?

hellofromtonya:
> Instead of adding a specific filter for .dev and .dev.cc why not adding make $excluded_tlds filterable? This could be more future proof and would cover every case.

@arunbasillal I think we need to be cautious when adding filters for all customers when only a small segment needs it.

Why?

A filter adds execution/processing time. Although it’s tiny fractions of a second, each we add slows down how fast Rocket runs.

Each piece of code, including filters, should be evaluated to determine how it impacts those who need it and those who don’t. When we can, we should target it directly to those who do need it without impacting those who don’t.

Back to this specific filter:

We identified that some customers who have these specific TLDs (.dev) might want access to RocketCDN. This PR only impacts that specific population of customers, ie 0.5%.

That means 99.5% of our customers will not see the tiny increase of processing time.

If we add a TLD exclusion list filter, then 100% of customers are impacted.

hellofromtonya:
## Live Testing Results

URL: knowthecode.dev

Results when installing the plugin without using the `'rocket_tld_is_live_prod_site'` filter:

Screen Shot 2020-02-24 at 6 46 37 AM

RocketCDN is not showing.

Results when adding this filter in a separate plugin:

```php
add_filter( 'rocket_tld_is_live_prod_site', '__return_true' );
```

RocketCDN is showing.

Screen Shot 2020-02-24 at 6 47 57 AM

Live testing works.

arunbasillal:
@hellofromtonya Thank you for the explanation. It makes sense. 

This is something I have had to consider in the past and didn't figure out how to benchmark this. 

- In our current implementation, every customer has to go through the `if` statement. 
- The `if` statements should be many times faster than `apply_filters`. Even with the `substr` evaluation that is part of the `if` statement. 
- However, (for future) can we do a test to see what the actual difference is? I have seen such tests done by recording the `time()` and running the tests 10 or 20 times to get an average. 

The reason I am pushing this: 

- Once we release this, we are in a way obligated to support `rocket_tld_is_live_prod_site` for a while. 
- As we add more `$excluded_tlds` and if more of those needs filters in the future, we will be adding more `if` statements and the complexity / maintainability of the code will go up. 
- Knowing the performance loss will help us estimate if it's worth it.

GeekPress:
@hellofromtonya What is the reason to include `.dev.cc`? I checked and this domain can't be buy, only `.dev` are valid domain since March 2019.

GeekPress:
In the mean time, if `.dev` is now a public domain, it should not be considered as is local website anymore.

Why? Now, `.dev` domain aren't counted during the license validation. If you have a Single license, it means that you can install unlimited `.dev` which is an issue ^^

PR #2340, opened 2020-02-24
Overwrite destination cache file if it already exists

Tabrisrp:
If the destination file defined in `move()` already exists, the method bails out early, and the temp file remains.

By using the 3rd parameter, an overwrite of the existing file is forced, preventing the issue.


hellofromtonya:
@Tabrisrp Are there tests? Did you do a live test to ensure it works as intended?

Tabrisrp:
We don't have test for this method yet because it was written at a time where we didn't have a virtual filesystem to work with.

I tested this change on my website, I didn't see any remaining `_temp` file

PR #2342, opened 2020-02-25
Fix the number of preloaded pages in the admin notice

### LINKED ISSUE #2341: Preload count in admin notice is inaccurate
The number of pages being preloaded does not match the number displayed in the admin notice.

Screenfeed:
About the way the pages are count:
- Do not count mobile pages.
- Count only pages successfully preloaded (request not blocked internally). We can't know about the request status (200) since the requests are non blocking, we have no answer.
- For the homepage preload, take also the homepage itself into account: since it is requested to grab its URLs, it is preloaded at the same time.
- Use 2 transients for each "source" (mobile and sitemap): this prevents the second source to reset the previous source’s count. So there is no `'rocket_preload_running'` transient anymore, but 2: `'rocket_homepage_preload_running'` and `'rocket_sitemap_preload_running'`. The admin notice shows the addition of both. Note: we could use a single transient containing an array of counts, but async calls + same (cached) storage source are never a good idea.

Tweak: the abstract class `WP_Rocket\Preload\Process` now has a `cancel_process()`, so we're sure it always has this method, even if we're extending an old version of `WP_Background_Process`.

Since we are so close of the beta version:
- No unit nor integration tests.
- With the way preload is built, I couldn't find a proper way to separate concerns (preload vs process), it’s quick and dirty.

Fixes #2341.

### DIFF
--- a/tests/Integration/inc/classes/preload/PreloadTestCase.php
+++ b/tests/Integration/inc/classes/preload/PreloadTestCase.php
@@ -1,16 +1,19 @@
 <?php
-namespace WP_Rocket\Tests\Integration;
 
-use WPMedia\PHPUnit\Integration\TestCase;
+namespace WP_Rocket\Tests\Integration\inc\classes\preload;
+
 use WP_Rocket\Tests\Integration\Fixtures\Preload\Process_Wrapper;
+use WPMedia\PHPUnit\Integration\TestCase;
 
-abstract class PreloadTestCase extends TestCase {
-	protected $site_url           = 'https://smashingcoding.com';
-	protected $identifier         = 'rocket_preload';
+class PreloadTestCase extends TestCase {
+	protected $site_url = 'https://smashingcoding.com';
+	protected $identifier = 'rocket_preload';
 	protected $option_hook_prefix = 'pre_get_rocket_option_';
 	protected $preloadErrorsTransient;
 	protected $preloadRunningTransient;
 	protected $process;
+	protected $setUpFilters = false;
+	protected $tearDownFilters = false;
 
 	public static function setUpBeforeClass() {
 		parent::setUpBeforeClass();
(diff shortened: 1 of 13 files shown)


hellofromtonya:
Live tested on knowthecode.dev. Count matches the cache filestructure and content. Count is correct.

Screen Shot 2020-02-24 at 8 48 19 PM

hellofromtonya:
@Tabrisrp Can you live test and note what you find?

crystinutzaa:
Tested on my local with the following configuration:

1. Initial run (No cache for mobile): 18pages - OK
2. Activate Mobile cache + separate mobile cache: 18 pages - OK
3. Added Sitemap Preload: 18pages - OK
4. Activated Cache Query strings + added in sitemap: 21pages (3 query strings) - OK
5. Never Cache URLs (added 2 urls): 19pages - OK
6. Set 2 pages to Drafts: OK

tests/Integration/inc/classes/preload/PreloadTestCase.php
```diff
@@ -1,16 +1,19 @@
...
+namespace WP_Rocket\Tests\Integration\inc\classes\preload;
+
 use WP_Rocket\Tests\Integration\Fixtures\Preload\Process_Wrapper;
+use WPMedia\PHPUnit\Integration\TestCase;
 
-abstract class PreloadTestCase extends TestCase {
-	protected $site_url           = 'https://smashingcoding.com';
-	protected $identifier         = 'rocket_preload';
+class PreloadTestCase extends TestCase {
+	protected $site_url = 'https://smashingcoding.com';
```
crystinutzaa: Alignment issue

crystinutzaa: I fixed and commited the small alignment issues which I spotted.

Tabrisrp:
Tested on my personal site, where I was getting incorrect count results with the previous changes.

Now it's all good, getting the expected number for the default (homepage) preload, and for the sitemap preload

PR #2380, opened 2020-03-05
Fixed Declared function at compile time, moved bailout condition before require

crystinutzaa:
Fixes:
Fatal error: Cannot redeclare rocket_remove_partial_purge_hooks() (previously declared in /wp-content/plugins/wp-rocket/inc/3rd-party/hosting/kinsta.php:107) in /wp-content/plugins/wp-rocket/inc/3rd-party/hosting/presslabs.php on line 72

In this case a function name (`rocket_remove_partial_purge_hooks`) is used in 2 placed and PHP functions defined inside IF conditions are declared at runtime and functions declared outside if conditions are declared at compile time. So the bailout will not work in this case and it worked in the previous version (it was inside if condition).

Solution:
Move the bailout condition before require. 

This will decrease also a little bit the number of files loaded by WP Rocket. 

https://secure.helpscout.net/conversation/1101455078/149590?folderId=3083222
Tested on the client staging server and is working fine.

### DIFF
--- a/inc/3rd-party/hosting/presslabs.php
+++ b/inc/3rd-party/hosting/presslabs.php
@@ -2,10 +2,6 @@
 
 defined( 'ABSPATH' ) || exit;
 
-if ( ! defined( 'PL_INSTANCE_REF' ) || ! class_exists( '\Presslabs\Cache\CacheHandler' ) ) {
-	return;
-}
-
 if ( ! file_exists( WP_CONTENT_DIR . '/advanced-cache.php' ) ) {
 	return;
 }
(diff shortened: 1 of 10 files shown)


inc/3rd-party/hosting/presslabs.php
```diff
@@ -2,10 +2,6 @@
 
 defined( 'ABSPATH' ) || exit;
 
-if ( ! defined( 'PL_INSTANCE_REF' ) || ! class_exists( '\Presslabs\Cache\CacheHandler' ) ) {
-	return;
-}
-
 if ( ! file_exists( WP_CONTENT_DIR . '/advanced-cache.php' ) ) {
```
hellofromtonya: @crystinutzaa What about this one? It was part of the entire flow of the file too. Right? Only load the file if this file also exists. I'm reading it correctly?

crystinutzaa: Please recheck. I missed it

PR #2387, opened 2020-03-05
Update priorities to prevent wrong insertion order

Tabrisrp:
Combine CSS and Google Fonts need to be inserted after the CPCSS in the page.

Reference:

- [HS Ticket](https://secure.helpscout.net/conversation/1101451982/149588?folderId=273768)
- Related to PR #2239 

Notes on the issue from Slack thread:

>the combined CSS process is happening on priority 26, so after the critical CSS insertion (19), and they both use the <title> as an anchor, so it ends up above instead of below

- test on customer website: changed priority for combined CSS and Google fonts to 16 and 18 respectively, display is correct ✅ 
- did the same test on my local site ✅

### DIFF
--- a/inc/classes/subscriber/Optimization/class-combine-google-fonts-subscriber.php
+++ b/inc/classes/subscriber/Optimization/class-combine-google-fonts-subscriber.php
@@ -21,7 +21,7 @@ class Combine_Google_Fonts_Subscriber extends Minify_Subscriber {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'rocket_buffer' => [ 'process', 28 ],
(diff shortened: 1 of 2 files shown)


hellofromtonya:
@Tabrisrp Can you add detail please:

1.how this resolves the issue
2. how you confirmed it is definitely resolved it

Why? It'll help us in the future if we need to backtrace an issue or why we changed it.

Thanks!

arunbasillal:
- Recreated the issue with 3.5.0.1 - https://jmp.sh/FsmytyG
- Tested this with the new version - https://jmp.sh/5mJRbmf

Looks good 👍

PR #2391, opened 2020-03-06
Fix #2328 Combine Google Fonts causing 404 error in specific conditions

### LINKED ISSUE #2328: Combine Google Fonts causing 404 error in specific conditions
Porto theme is adding Google Fonts using unexpected (incorrect) markup:
`//fonts.googleapis.com/css?display=swap&#038;family=Open+Sans:200,300,400,700,800,600,|Shadows+Into+Light:200,300,400,700,800,600,|Oswald:200,300,400,700,800,600,|&#038;subset=cyrillic,cyrillic-ext,greek,greek-ext,khmer,latin,latin-ext,vietnamese`

Note the `|` after the definition of the last, Oswald font.

Correct markup:
image

When our combination feature is enabled, we're not removing this `|` but appending another `|` with font definition. It leads to the 404 error, this is how it looks like after combination:
image

Their support team claims that it should be fixed on our end, advising the customer to deactivate our plugin. 

Related ticket:
https://secure.helpscout.net/conversation/1082860619/146039?folderId=2415573

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
**Identify the root cause** ✅ 
This is the line:
https://github.com/wp-media/wp-rocket/blob/a37e46f58259e38dbf895061536551763f8aace7/inc/classes/optimization/CSS/class-combine-google-fonts.php#L131-L132

In Porto theme, the last font adds a | 
When we combine Porto theme fonts with other google fonts, this will break the format because will contain 2 ||.

**Scope a solution** ✅ 
We need to parse the list of $this->fonts and remove from the end any `|` or `%7C`, so the double | will never appear while we implode this array.

**TODO**
- [x] Unit tests
- [x] Integration tests
- [ ] QA live tests


tests/Fixtures/inc/classes/optimization/CSS/Combine_Google_Fonts/optimize.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+// phpcs:disable WordPress.WP.EnqueuedResources.NonEnqueuedStylesheet
+
+return [
+
```
hellofromtonya: @crystinutzaa Let's add one more test data set that doesn't have the ending `|` character, i.e. just to make sure it runs with and without it.

inc/classes/optimization/CSS/class-combine-google-fonts.php
```diff
@@ -121,20 +123,22 @@ protected function parse( $matches ) {
 				return;
 			}
 
-			$font = wp_parse_args( $query );
-
+			$font        = wp_parse_args( $query );
+			$font_family = $font['family'];
```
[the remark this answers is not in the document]
crystinutzaa: It will match all links like this:
`<link(?:\s+(?:(?!href\s*=\s*)[^>])+)?(?:\s+href\s*=\s*([\'"])((?:https?:)?\/\/fonts\.googleapis\.com\/css(?:(?!\1).)+)\1)(?:\s+[^>]*)?>`

eg: `https://fonts.googleapis.com/css?`

If the matched URL has no query string, it will bailout and no google fonts will be combined.
However if it has query strings, I added a check for family and combined only those which have family included. The rest will be removed totally from HTML because are not valid (test: // Without Subsets & malformed URL (no family, but wrong query string) ).

inc/classes/optimization/CSS/class-combine-google-fonts.php
```diff
@@ -113,28 +113,34 @@ protected function find( $pattern, $html ) {
...
+				$subsets_array[] = rawurlencode( htmlentities( $font['subset'] ) );
 			}
 		}
 
 		// Concatenate fonts tag.
-		$this->subsets = ! empty( $this->subsets ) ? '&subset=' . implode( ',', array_filter( array_unique( $this->subsets ) ) ) : '';
-		$this->fonts   = implode( '|', array_filter( array_unique( $this->fonts ) ) );
+		$this->subsets = ! empty( $subsets_array ) ? '&subset=' . implode( ',', array_filter( array_unique( $subsets_array ) ) ) : '';
+		$this->fonts   = implode( '|', array_filter( array_unique( $fonts_array ) ) );
 		$this->fonts   = str_replace( '|', '%7C', $this->fonts );
```
hellofromtonya: Add a "not empty" check here to only do this work _when_ we have fonts. Else, set `$this->fonts = ''`.

```php
if ( ! empty( $fonts_array ) {
	$this->fonts   = implode( '|', array_filter( array_unique( $fonts_array ) ) );
	$this->fonts   = str_replace( '|', '%7C', $this->fonts );
} else {
	$this->fonts = '';
}
```

hellofromtonya:
This PR is ready for QA. @arunbasillal @piotrbak or @vmanthos please self-assign yourself to do this PR as part of 3.5.1 release. Thanks.

inc/classes/optimization/CSS/class-combine-google-fonts.php
```diff
@@ -113,29 +113,39 @@ protected function find( $pattern, $html ) {
...
-		$this->subsets = ! empty( $this->subsets ) ? '&subset=' . implode( ',', array_filter( array_unique( $this->subsets ) ) ) : '';
-		$this->fonts   = implode( '|', array_filter( array_unique( $this->fonts ) ) );
-		$this->fonts   = str_replace( '|', '%7C', $this->fonts );
+		$this->subsets = ! empty( $subsets_array ) ? '&subset=' . implode( ',', array_filter( array_unique( $subsets_array ) ) ) : '';
+		if ( ! empty( $fonts_array ) ) {
+			$this->fonts = implode( '|', array_filter( array_unique( $fonts_array ) ) );
+			$this->fonts = str_replace( '|', '%7C', $this->fonts );
+		} else {
+			$this->fonts = '';
+		}
```
Tabrisrp: is this `else` block even needed? The `fonts` property already has a default empty string value

hellofromtonya: @Tabrisrp It is set to a default of an empty string on object creation. However, what happens when the object is called more than once? This `else` ensures it set to either state when it's done.

inc/classes/optimization/CSS/class-combine-google-fonts.php
```diff
@@ -113,29 +113,39 @@ protected function find( $pattern, $html ) {
...
 		}
 
 		// Concatenate fonts tag.
-		$this->subsets = ! empty( $this->subsets ) ? '&subset=' . implode( ',', array_filter( array_unique( $this->subsets ) ) ) : '';
-		$this->fonts   = implode( '|', array_filter( array_unique( $this->fonts ) ) );
-		$this->fonts   = str_replace( '|', '%7C', $this->fonts );
+		$this->subsets = ! empty( $subsets_array ) ? '&subset=' . implode( ',', array_filter( array_unique( $subsets_array ) ) ) : '';
+		if ( ! empty( $fonts_array ) ) {
+			$this->fonts = implode( '|', array_filter( array_unique( $fonts_array ) ) );
+			$this->fonts = str_replace( '|', '%7C', $this->fonts );
```
hellofromtonya: @crystinutzaa Couldn't we implode using `'%7C'` instead of `'|'` to combine these 2 lines of code into one?

crystinutzaa: %7C = rawurlencode( | ), so I think it can be combined into a single implode in there without any str_replace.
I will give it a try and commit if it seems fine

hellofromtonya: Cool. If yes, go ahead and use a ternary to be consistent with subsets line above it.

inc/classes/optimization/CSS/class-combine-google-fonts.php
```diff
@@ -113,29 +113,34 @@ protected function find( $pattern, $html ) {
...
+				$subsets_array[] = rawurlencode( htmlentities( $font['subset'] ) );
 			}
 		}
 
 		// Concatenate fonts tag.
-		$this->subsets = ! empty( $this->subsets ) ? '&subset=' . implode( ',', array_filter( array_unique( $this->subsets ) ) ) : '';
-		$this->fonts   = implode( '|', array_filter( array_unique( $this->fonts ) ) );
-		$this->fonts   = str_replace( '|', '%7C', $this->fonts );
+		$this->subsets = ! empty( $subsets_array ) ? '&subset=' . implode( ',', array_filter( array_unique( $subsets_array ) ) ) : '';
+		$this->fonts   = ! empty( $fonts_array ) ? implode( '%7C', array_filter( array_unique( $fonts_array ) ) ) : '';
```
hellofromtonya: This code is now more performant: runs faster and uses less memory. Plus, it fits the subsets pattern, making it more readable. Good job @crystinutzaa 🥇

PR #2393, opened 2020-03-06
Fixes #2341: Preload count in admin notice is inaccurate

### LINKED ISSUE #2341: Preload count in admin notice is inaccurate

Screenfeed:
This fixes again the number of preloaded pages displayed in the admin notice.
The notice was displaying the homepage preload count only, missing the sitemap preload count.

At some point we will need to rethink some things in this `Preload_Subscriber` class, as it should deal only with homepage preload (as the `Sitemap_Preload_Subscriber` deals with sitemap preload), but ends up also dealing with the admin notice, and so, with the sitemap preload.
Imho, this class should be the reunion of all preload types, and a class `Homepage_Preload_Subscriber` (similar to `Sitemap_Preload_Subscriber`, and both using a common interface) should be created. The hidden advantage of this class would be its name being more understandable (it was a challenge for me to understand the role of each class, especially since the roles are mixed in this one).

Closes #2341.

### DIFF
--- a/inc/classes/ServiceProvider/class-preload-subscribers.php
+++ b/inc/classes/ServiceProvider/class-preload-subscribers.php
@@ -47,6 +47,7 @@ class Preload_Subscribers extends AbstractServiceProvider {
 			->withArgument( $this->getContainer()->get( 'full_preload_process' ) );
 		$this->getContainer()->share( 'preload_subscriber', 'WP_Rocket\Subscriber\Preload\Preload_Subscriber' )
 			->withArgument( $this->getContainer()->get( 'homepage_preload' ) )
+			->withArgument( $this->getContainer()->get( 'sitemap_preload' ) )
 			->withArgument( $this->getContainer()->get( 'options' ) );
 		$this->getContainer()->share( 'sitemap_preload_subscriber', 'WP_Rocket\Subscriber\Preload\Sitemap_Preload_Subscriber' )
 			->withArgument( $this->getContainer()->get( 'sitemap_preload' ) )
--- a/inc/classes/subscriber/preload/class-preload-subscriber.php
+++ b/inc/classes/subscriber/preload/class-preload-subscriber.php
@@ -5,6 +5,7 @@ use WP_Rocket\Logger\Logger;
 use WP_Rocket\Event_Management\Subscriber_Interface;
 use WP_Rocket\Admin\Options_Data;
 use WP_Rocket\Preload\Homepage;
+use WP_Rocket\Preload\Sitemap;
 
 /**
  * Preload Subscriber
@@ -24,6 +25,16 @@ class Preload_Subscriber implements Subscriber_Interface {
 	private $homepage_preloader;
 
 	/**
+	 * Sitemap Preload instance.
+	 *
+	 * @since  3.5.0.2
+	 * @author Grégory Viguier
+	 *
+	 * @var Sitemap
+	 */
+	private $sitemap_preloader;
+
+	/**
 	 * WP Rocket Options instance.
 	 *
 	 * @since 3.2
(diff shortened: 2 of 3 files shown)


hellofromtonya:
Team

Once this one gets through code review and QA, don't merge it yet. 

Why?

We are monitoring tickets to determined if we need to do a `3.5.0.3` hotfix or save this one for `3.5.1`.

hellofromtonya:
@piotrbak As you were working on identifying the root cause, I think it'll be good to have you do the QA on it. I'm adding you as a reviewer. Go ahead and do your tests/checks. Thanks.

inc/classes/subscriber/preload/class-preload-subscriber.php
```diff
@@ -40,10 +51,12 @@ class Preload_Subscriber implements Subscriber_Interface {
 	 * @author Remy Perona
 	 *
 	 * @param Homepage     $homepage_preloader Homepage Preload instance.
+	 * @param Sitemap      $sitemap_preloader  Sitemap Preload instance.
 	 * @param Options_Data $options            WP Rocket Options instance.
 	 */
-	public function __construct( Homepage $homepage_preloader, Options_Data $options ) {
+	public function __construct( Homepage $homepage_preloader, Sitemap $sitemap_preloader, Options_Data $options ) {
 		$this->homepage_preloader = $homepage_preloader;
+		$this->sitemap_preloader  = $sitemap_preloader;
```
hellofromtonya: Let's talk about this solution:

1. We are creating a new instance of the `Sitemap` when we inject it into the homepage's subscriber.
2. We are mixing concerns as these are different.
3. And we're only doing it because we need to add them together for the counter.

I propose that we look at create a different approach to handle the Preload Count notice separately. Let it pull from homepage, sitemap, or both.

What do you think @Screenfeed @Tabrisrp ?

hellofromtonya: From our Slack discussion, we agreed to simplify this:

1. Don't inject the Sitemap Preloader.
2. Combine the preload transients into one shared preload count transient.
3. Get the transient directly using `get_transient()` to refresh the admin notice.

Screenfeed: So after more discussion, we keep the two transients because they are needed as explained in #2342: to prevent sitemap preload to reset homepage’s counter, and because homepage and sitemap preload are actually working asynchronously.

inc/classes/ServiceProvider/class-preload-subscribers.php
```diff
@@ -47,6 +47,7 @@ public function register() {
 			->withArgument( $this->getContainer()->get( 'full_preload_process' ) );
 		$this->getContainer()->share( 'preload_subscriber', 'WP_Rocket\Subscriber\Preload\Preload_Subscriber' )
 			->withArgument( $this->getContainer()->get( 'homepage_preload' ) )
+			->withArgument( $this->getContainer()->get( 'sitemap_preload' ) )
```
hellofromtonya: Remove this. We don't need to inject the sitemap preloader.

inc/classes/subscriber/preload/class-preload-subscriber.php
```diff
@@ -5,6 +5,7 @@
 use WP_Rocket\Event_Management\Subscriber_Interface;
 use WP_Rocket\Admin\Options_Data;
 use WP_Rocket\Preload\Homepage;
+use WP_Rocket\Preload\Sitemap;
```
hellofromtonya: Not needed.

inc/classes/subscriber/preload/class-preload-subscriber.php
hellofromtonya: Let's remove these changes. We don't need the sitemap preloader in this subscriber.

inc/classes/subscriber/preload/class-preload-subscriber.php
```diff
@@ -270,13 +283,15 @@ public function notice_preload_running() {
 			return;
 		}
 
-		$running = $this->homepage_preloader->get_number_of_preloaded_items();
+		$homepage_count = $this->homepage_preloader->get_number_of_preloaded_items();
+		$sitemap_count  = $this->sitemap_preloader->get_number_of_preloaded_items();
```
hellofromtonya: Use the transients directly instead to simplify this. Then there's no need to inject the sitemap preloader.

Screenfeed:
With #2396 coming, I guess I will delete these branch and pull request to prevent conflicts and restart from a clean slate.

Screenfeed:
#2402 replaces this PR.

PR #2396, opened 2020-03-06
Bundle Preload files into a Module

hellofromtonya:
The files for Preload are distributed through the plugin. This PR bundles them together by:

- Pulling all of them into one module, i.e. `inc/Engine/Preload`
- Changes the namespaces to match
- Renames the class files to PSR-4, i.e. to match the class names

## Why

This PR is part of our Dev initiative to rearchitect Rocket's file structure into modular, reusable packages and modules. It complies with our coding standards.

Benefits:

- Easier to see the relationships between the files
- Easier to work with, navigate, refactor, troubleshoot, etc. as everything is bundled together
- PSR-4 gives us a modern PHP approach to naming and fits seamlessly into computer PSR-4 autoloader

## TODO

- [x] Live test -- tested on knowthecode.dev. Works.
- [x] QA


inc/Engine/Preload/preload.php
```diff
@@ -2,6 +2,10 @@
```
Tabrisrp: This file should be in inc/API/preload.php instead I believe

inc/main.php
```diff
@@ -44,7 +44,7 @@ function rocket_init() {
 	require WP_ROCKET_FUNCTIONS_PATH . 'files.php';
 	require WP_ROCKET_FUNCTIONS_PATH . 'posts.php';
 	require WP_ROCKET_FUNCTIONS_PATH . 'admin.php';
-	require WP_ROCKET_FUNCTIONS_PATH . 'preload.php';
+	require WP_ROCKET_INC_PATH . '/Engine/Preload/preload.php';
```
Tabrisrp:
```suggestion
	require WP_ROCKET_INC_PATH . '/API/preload.php';
```

hellofromtonya: @Tabrisrp Are we using these functions to power parts of Rocket's Engine, i.e. the sourcecode internal to it? Or are they for external themes/plugins/etc to use to trigger things within Rocket?

hellofromtonya: Looks like external stuff. I think you're right. I'll move now.

inc/Engine/Preload/preload.php
hellofromtonya: Resolved here https://github.com/wp-media/wp-rocket/pull/2396/commits/8675a6bc19165f5173f4f6d11bbd6e13bcfc005c

inc/main.php
hellofromtonya: Resolved here https://github.com/wp-media/wp-rocket/pull/2396/commits/8675a6bc19165f5173f4f6d11bbd6e13bcfc005c

hellofromtonya:
Hey @piotrbak @arunbasillal or @vmanthos, this PR is ready for QA. 

It's built off of 3.5.0.2. No feature changes. No bug fixes. It's re-architecting the Preload files to bundle them together as one module. For QA, please make sure that everything still works as it does in 3.5.0.2. 

Thank you!!

vmanthos:
Everything is working the same way as in the version without the bundle.

I've discovered that when updating the content of a post, partial preloading creates only the desktop version and not both the desktop and mobile version as I'd expect.

To reproduce:
1. Enable "Separate cache for mobile devices".
2. Update a post.
3. Check the post's folder.

hellofromtonya:
>I've discovered that when updating the content of a post, partial preloading creates only the desktop version and not both the desktop and mobile version as I'd expect.

@vmanthos Was this issue introduced in this PR? Or is a Mobile Preload issue? If a Mobile Preload issue and not introduced in this PR, please open a new issue. We'll want to address it separately.

hellofromtonya:
Also @vmanthos if the issue you noted was not introduced in this PR and everything works as expected, please approve this PR. Thanks.

vmanthos:
It's a separate issue related to **Mobile Preload**.
I'm already in the process of creating a new issue about this.

hellofromtonya:
Good job @vmanthos identifying this unexpected behavior! 🥇

Screenfeed:
@hellofromtonya This occurs to me now because of the conversation with Arun: since we are changing the namespaces for this module, shouldn't we take the opportunity to also remove the underscores from file names and classes?

hellofromtonya:
@Screenfeed Well that's a good point. You're right. We should. 

I'm thinking, we might want to keep `WP_Rocket` for now though, but change everything else. What do you think?

Screenfeed:
Yes, sure.

hellofromtonya:
Would you like to handle changing this in a separate PR or as part of your counter PR? @Screenfeed

Screenfeed:
Two PRs mean twice more QA. So I'll bundle it in my counter PR.

Screenfeed:
Also, should we create aliases for the old classes, to link to the new ones?

hellofromtonya:
>Also, should we create aliases for the old classes, to link to the new ones?

Good idea for backwards compatibility. Are you thinking of adding those into deprecated 3.5?

Screenfeed:
It seems to be a good place indeed.

PR #2400, opened 2020-03-09
Fixes 2376 Prevent PHP Fatal Error in process.php

### LINKED ISSUE #2376: Fatal Error: Call to undefined function wp_unslash() on 3.5
Original edited by Remy.

**Describe the bug**
During 3.5 development, we updated all the files to pass the coding standards. This included the `process.php` file, which is deprecated but still present for backward compatibility.

The file is not supposed to be used on WP Rocket installations since 3.3, but some installations seems to not have updated the `advanced-cache.php` file correctly since (even though that was part of the upgrade routine). So in some cases, it's still used currently.

We skipped the part where `process.php` can't have WP core functions in it, since WP is not loaded yet when the file is executed. So some functions like `wp_unslash()` made their way into this file, causing the fatal error.

**To Reproduce**
- Have an installation with the old version of the `advanced-cache.php` file
- Update to 3.5
- The fatal error should occur

**Expected behavior**
There should be no fatal error there, and `advanced-cache.php` should be updated with the correct content.

**Additional context**
This impacts a limited pool of installations that are still using the old version of the `advanced-cache.php` file for whatever reason. But it still creates support tickets.

For the support team, the fix for now is to delete the `advanced-cache.php` file so that it can be recreated with the correct content.

Tabrisrp:
- Remove calls to `wp_unslash()`, WP is not fully loaded at that point
- Regenerate the `advanced-cache.php` file on next plugin update

Closes #2376 

## TODO

- [x] Review the [Acceptance Criteria](https://github.com/wp-media/wp-rocket/issues/2376#issuecomment-596566555)
- [x] Review the ["What if" scenarios](https://github.com/wp-media/wp-rocket/issues/2376#issuecomment-596566555)
- [x] Unit tests
- [x] Integration tests
- [x] QA

### DIFF
--- a/inc/admin/upgrader.php
+++ b/inc/admin/upgrader.php
@@ -254,8 +254,6 @@ add_action( 'wp_rocket_first_install', 'rocket_first_install' );
  */
 function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 	if ( version_compare( $actual_version, '2.4.1', '<' ) ) {
-		// Regenerate advanced-cache.php file.
-		rocket_generate_advanced_cache_file();
 		delete_transient( 'rocket_ask_for_update' );
 	}
 
@@ -270,13 +268,9 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 
 		update_option( WP_ROCKET_SLUG, $options );
 
-		// Regenerate advanced-cache.php file.
-		rocket_generate_advanced_cache_file();
 	}
 
 	if ( version_compare( $actual_version, '2.7', '<' ) ) {
-		// Regenerate advanced-cache.php file.
-		rocket_generate_advanced_cache_file();
 
 		// Regenerate config file.
 		rocket_generate_config_file();
@@ -334,7 +328,6 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 		rocket_clean_domain();
 		rocket_clean_minify();
 		rocket_clean_cache_busting();
-		rocket_generate_advanced_cache_file();
 	}
 
 	if ( version_compare( $actual_version, '3.0.3', '<' ) ) {
@@ -347,10 +340,6 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 
 	if ( version_compare( $actual_version, '3.1.1', '<' ) ) {
 		rocket_generate_config_file();
-	}
-
-	if ( version_compare( $actual_version, '3.1.4', '<' ) ) {
-		rocket_generate_advanced_cache_file();
 	}
 
 	if ( version_compare( $actual_version, '3.2', '<' ) ) {
@@ -366,7 +355,6 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 
 		update_option( WP_ROCKET_SLUG, $options );
 		rocket_generate_config_file();
-		rocket_generate_advanced_cache_file();
 
 		// Create a .htaccess file in the log folder.
 		$handler = Logger::get_stream_handler();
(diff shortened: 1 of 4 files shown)


inc/admin/upgrader.php
```diff
@@ -254,8 +254,6 @@ function rocket_first_install() {
  */
```
hellofromtonya: @Tabrisrp Should this file move into `inc/Engine/Admin`???

inc/admin/upgrader.php
```diff
@@ -270,13 +268,9 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 
 		update_option( WP_ROCKET_SLUG, $options );
```
hellofromtonya: Delete this empty line.

inc/admin/upgrader.php
```diff
@@ -270,13 +268,9 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
...
 
-		// Regenerate advanced-cache.php file.
-		rocket_generate_advanced_cache_file();
 	}
 
 	if ( version_compare( $actual_version, '2.7', '<' ) ) {
-		// Regenerate advanced-cache.php file.
-		rocket_generate_advanced_cache_file();
```
hellofromtonya: Delete this empty line.

inc/admin/upgrader.php
Tabrisrp: Do we want to mix old and new code in  the new structure? And also classes vs procedural? We didn't discuss those yet

hellofromtonya: Not everything will change necessarily from procedural to OOP. In some cases, we want to have functions available.  But your point is valid.

Let's think about: Is this upgrader a candidate for refactoring into a class? If yes, then we can schedule that refactoring work for later. If no, then we can retain it's structure but relocate it into the Engine or API. 

What do you think?

Tabrisrp: I believe the whole file could get some benefits by being refactored in a class. It's all internal stuff, so it wouldn't be any value for the API part though.

hellofromtonya: Okay cool. Let's look at this as a separate task. Thanks @Tabrisrp

hellofromtonya:
This PR is ready for Dev QA. @arunbasillal @piotrbak or @vmanthos please self-assign to do the QA. Thanks.

hellofromtonya:
@Tabrisrp Don't merge this PR once it's approved through QA. Let's keep it parked here until we know if it we need a 3.5.0.3 hot fix.

Tabrisrp:
You can use the following code sample for the old `advanced-cache.php` file which loads `process.php`

```
<?php
	defined( 'ABSPATH' ) || exit;

	define( 'WP_ROCKET_ADVANCED_CACHE', true );

	$rocket_cache_path = '/app/public/wp-content/cache/';

	$rocket_config_path = WP_CONTENT_DIR . '/wp-rocket-config/';

	if ( file_exists( '/app/public/wp-content/plugins/wp-rocket/inc/vendors/classes/class-rocket-mobile-detect.php' ) && ! class_exists( 'Rocket_Mobile_Detect' ) ) {
	include_once '/app/public/wp-content/plugins/wp-rocket/inc/vendors/classes/class-rocket-mobile-detect.php';
}

	if ( file_exists( '/app/public/wp-content/plugins/wp-rocket/inc/front/process.php' ) ) {
		include '/app/public/wp-content/plugins/wp-rocket/inc/front/process.php';
	} else {
	define( 'WP_ROCKET_ADVANCED_CACHE_PROBLEM', true );
}
```

hellofromtonya:
@Tabrisrp Are there automated tests for this function? If no, please create them.

hellofromtonya:
@Tabrisrp Does this PR meet the [Acceptance Criteria](https://github.com/wp-media/wp-rocket/issues/2376#issuecomment-596566555)? Does it account for the ["What if" scenarios](https://github.com/wp-media/wp-rocket/issues/2376#issuecomment-596566555)? Please make sure "Yes" to both questions.

hellofromtonya:
This PR is urgent and ready for QA. @arunbasillal @piotrbak or @vmanthos please prioritize this one and do the QA. We want to do a hot fix 3.5.0.3 release.

PR #2408, opened 2020-03-11
Prevent cache from being recreated on each page load when there is no matching images

### LINKED ISSUE #2280: WebP: Cache recreated on every load, when there aren't any images on a page
The cache is recreated on every page load when:
* The WebP feature is enabled
* There aren't any images at all, or any supported images on a page.
A background image added e.g. with Elementor, doesn't count as a supported image.

We check for images on a page here:
https://github.com/wp-media/wp-rocket/blob/a37e46f58259e38dbf895061536551763f8aace7/inc/classes/subscriber/Media/class-webp-subscriber.php#L170-L171

If there aren't any, the comment about the existence or not of WebP is not added:
https://github.com/wp-media/wp-rocket/blob/a37e46f58259e38dbf895061536551763f8aace7/inc/classes/subscriber/Media/class-webp-subscriber.php#L199-L222

That's being used to determine the creation of the `.no-webp` file:
https://github.com/wp-media/wp-rocket/blob/a37e46f58259e38dbf895061536551763f8aace7/inc/classes/Buffer/class-cache.php#L284-L306

In this case, the file is not created which leads to the issue.

_Note: This issue isn't related to issues #2156 or #2158._

**Related ticket**
https://secure.helpscout.net/conversation/1059666026/141735/

**Steps to reproduce:**
1. Enable WP Rocket's WebP feature in the Media tab.
2. Create a page where there aren't any images at all.
3. Check the page's cache folder (/cache/wp-rocket) and note the file's timestamp
4. Visit the page in a browser and refresh it.
5. Check the timestamp again. It will change after every refresh of the page.

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

Tabrisrp:
Closes #2280 

When the page contained no images, or there was no extensions/attributes matching, the cache was recreated on each page load when visiting with WebP compatible browsers.

This happened because in those cases, the WebP comment was not added to the returned HTML.

By adding the comment, the `.no-webp` file is correctly created, and WebP compatible browsers won't trigger a regeneration each time.

- [x] Unit tests
- [x] Integration tests
- [ ] QA live tests


tests/Integration/inc/classes/subscriber/Media/WebpSubscriber/convertToWebp.php
```diff
@@ -0,0 +1,36 @@
...
+			}
+		);
+
+		$this->assertSame( $expected, apply_filters( 'rocket_buffer', $original ) );
+
+		remove_filter( 'pre_get_rocket_option_cache_webp', [ $this, 'return_1' ] );
+	}
+
+	public function noMatchProvider() {
+		return $this->getTestData( __DIR__, 'convert-to-webp-nomatch' );
```
hellofromtonya: @Tabrisrp How about an integration test for when it should convert into webp? Can you combine both test data sets into one to load all of it into the data provider? In that way, we are testing the behavior with different states to check what happens.

inc/classes/subscriber/Media/class-webp-subscriber.php
```diff
@@ -163,14 +163,14 @@ public function convert_to_webp( $html ) {
...
+			return $html . '<!-- Rocket no webp -->';
 		}
 
 		$extensions      = implode( '|', $extensions );
 		$attribute_names = implode( '|', $attribute_names );
 
 		if ( ! preg_match_all( '@["\'\s](?<name>(?:data-(?:[a-z0-9_-]+-)?)?(?:' . $attribute_names . '))\s*=\s*["\']\s*(?<value>(?:https?:/)?/[^"\']+\.(?:' . $extensions . ')[^"\']*?)\s*["\']@is', $html, $attributes, PREG_SET_ORDER ) ) {
-			return $html;
+			return $html . '<!-- Rocket no webp -->';
 		}
```
hellofromtonya: Is there a risk of it being processed more than once? If yes, could an edge case be that the HTML already has one of these comments appended to it, i.e. `<!-- Rocket no webp -->` or `<!-- Rocket has webp -->`?

hellofromtonya: And on lines 134 and 139, do we need to do anything there? In those 2 cases, we are bailing out. Do we need to add the `<!-- Rocket no webp -->` to those too?

tests/Integration/inc/classes/subscriber/Media/WebpSubscriber/convertToWebp.php
Tabrisrp: The integration test for successful conversion to WebP will require to use the virtual filesystem.

I kept it simple by checking the issue at hand specifically here, I can see if it's doable to check for this in addition to it, but it's possible I will get into blockers

hellofromtonya: You did it right: write the test for this specific issue. If the other state is doable, cool let's do it. Else, we can do it when we touch that code.

inc/classes/subscriber/Media/class-webp-subscriber.php
Tabrisrp: Lines 134 and 139 are fine, because we didn't get into the browser compatibility check yet in this, and we have the necessary checks on the cache side to not have any issue.

As far as I know it doesn't happen that the methods hooked on `rocket_buffer` are executed more than once per page.

hellofromtonya: >Lines 134 and 139 are fine, because we didn't get into the browser compatibility check yet in this, and we have the necessary checks on the cache side to not have any issue.

What happens to the cache when the page contains no images + using a WebP compatible browser +:
- `cache_webp` is off?
- or some callback is hooked into `rocket_disable_webp_cache` short-circuits the method? 

is the cache recreated for that web page?

Tabrisrp: No in those cases, all browsers will be looking for the cache file without the `webp` suffix, so it's all good.

hellofromtonya:
This PR is ready for Dev QA. @arunbasillal @piotrbak or @vmanthos -- please self-assign and do the QA on this one. Thanks.

arunbasillal:
Starting QA on this one

arunbasillal:
- I was able to recreate the issue with current `master branch` (3.5.0.3)
- Switching to `2280-no-webp` branch solved the issue. `.no-webp` file was created correctly 
- Cache was not recreated when the page was refreshed. 
- On a page with images, but no .webp images, normal cache is generated and the `.no-webp` file. Cache was not recreated on page refresh. 
- Added a `.webp` conversion for an image in the image's folder. Visited the page in Chrome, `index-webp.html` was generated. Visited the page in Internet Explorer, `index.html` was generated. 

All good 👍

PR #2413, opened 2020-03-12
Remove scheme from URLs entered in the DNS prefetch field

### LINKED ISSUE #2255: Proposal: Fix incorrectly entered prefetches
I see incorrect prefetch entries all the time: 

wrong_prefetch

Could we automatically fix them?

**Steps to Reproduce:** 

- In `Preload` tab > `Prefetch DNS Requests`
- Add an incorrect prefetch entry like `http://example.com` (or `https://example.com`)
- This should be automatically corrected to `//example.com` (i.e. no protocol) when a user clicks `Save Settings`

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

Tabrisrp:
Closes #2255 

Add a method to handle sanitization/validation of the DNS Prefetch field values to format the URLs correctly following this pattern : `//example.org`

- [x] Dev live test
- [x] Unit tests
- [x] Integration tests
- [x] QA live test


inc/classes/admin/settings/class-settings.php
```diff
@@ -496,4 +490,42 @@ function( $error ) {
...
+			$value = explode( "\n", $value );
+		}
+
+		$value = array_filter( array_map( 'trim', $value ) );
+
+		if ( empty( $value ) ) {
+			return [];
+		}
+
+		$value = array_filter( array_map( 'esc_url_raw', $value ) );
```
hellofromtonya: @Tabrisrp This works fine. But it's less performant as it takes more memory and cycles. Why?

Each time `array_map` runs, it has to iterate through all of the values. Plus it creates a new array in memory. So it's Big O for time-complexity is N and space-complexity is also N.

Same with `array_filter`. 

Think about that. There are 2 `array_maps` and `array_filters`. That's 4 times iterating through all of the values and creating 4 new arrays in memory.

We can improve this by using a `foreach`:

```php
$urls = [];
foreach ( $values as $index => $url ) {
	$url = trim( $url );
	if ( empty( $url ) ) {
		continue;
	}
	$url = esc_url_raw( $url );
	if ( empty( $url ) ) {
		continue;
	}
	$urls[] = $url;
}

if ( empty( $urls ) ) {
	return [];
}
```

Notice:

1. It only iterates through the values 1x instead of 4x.
2. It creates 1 array instead of 4. Plus, the array is sized for only the values we need.
3. It's still readable and understandable.

inc/classes/admin/settings/class-settings.php
```diff
@@ -496,4 +490,42 @@ function( $error ) {
...
+	/**
+	 * Sanitizes the DNS Prefetch sub-option value
+	 *
+	 * @since 3.5.1
+	 * @author Remy Perona
+	 *
+	 * @param array $input Array of values for the WP Rocket settings option.
+	 * @return array Sanitized array for the DNS Prefetch sub-option
+	 */
+	private function sanitize_dns_prefetch( $input ) {
```
hellofromtonya: Please type hint here.

hellofromtonya:
This PR is ready for QA. @arunbasillal @piotrbak or @vmanthos -- one of you please self-assign. Thanks!

arunbasillal:
Starting QA for this now

arunbasillal:
- Recreated the issue with current `master` branch
- Switched to `fix/2255-prefetch-scheme` branch

- Input data:

```
http://domain.com
https://domain2.com
http://domain3.com/
https://domain4.com/
http://domain5.com/example
http://domain6.com/example/
//domain7.com
http:///domain8.com
ftp://domain9.com
```

Output was: 

```
//domain.com
//domain2.com
//domain3.com
//domain4.com
//domain5.com
//domain6.com
//domain7.com
//
//domain9.com
```

- `http:///domain8.com` (triple slashes) is being converted to `//`
^ This is an edge case and invalid input (https://github.com/wp-media/wp-rocket/issues/2255#issuecomment-595680830). However, if we can handle it better rather than converting it into `//` that would be more ideal. Pretty sure this is an unplanned byproduct right now.

Tabrisrp:
`http:///domain8.com` is not a valid URL to be parsed, because of the triple slashes, so the function will return an empty string instead of the expected parts.

How do we want to approach there:

- Remove the entry?
- Try to format it correctly?

The first option is simpler with the drawback of making the entry disappear after submission.

The second is more complex, as correctly validating an URL is quite complicated currently without external libraries. But it would preserve the entry.

BTW, this kinda applies to incorrect strings like if a user enter `testtest` for example. This will be considered ok even though it’s not, But during development I considered the scope of the issue + how far we should go to fix user input mistakes, and I felt like we have to put some limits there

arunbasillal:
In my perspective we have these options: 

- Correct invalid input
- Leave invalid input as it is
- Remove invalid input

I don't think we have to go the extra step to correct invalid entries. As Remy described, its quite complicated. 

How complicated is it to remove invalid input? 
^ That is what I would prefer. What do you think @GeekPress ?

hellofromtonya:
If a URL is malformed, then it's not resolvable. We shouldn't be correcting it IMO.

What should we do? I think we should exclude it.

What about notification? How would the customer know that it's malformed and excluded? Some options might include:

- Logging
- Storing and then displaying the issues to be manually resolved

Tabrisrp:
> How complicated is it to remove invalid input?
It really depends on the case:
- `http:///domain8.com`: easy, `parse_url` doesn't like it, so it returns `null`, and we can catch that
- `testtest` or any similar string: without some form of URL validation, it's hard. Both `esc_url_raw` and `parse_url` will consider this a valid input. And we can't ignore entries not starting with `//`, because that would be excluding `domain.com` which is fine

GeekPress:
We have 4 solutions:

1. Don't touch it and keep `http:///domain8.com` like it was before the change
2. Keep it with the new behavior `//` 
3. Keep it with the new behavior `//` and make a notice to warn about the issue
4. Auto-correct to `//domain8.com`

What will be the best user experience?

4.
Why?
It's a part of our plugin DNA and how we are thinking plugin at WP Media: make it easiest as possible. 

If we can avoid extra steps to our users and make auto-correction when needed, we make our plugin easer to use.

I also prefer to have 1. compare to 2. 
Why?
On 1. the user doesn't know there is an issue.
On 2. the user will notice the change, try to paste again what he wrote and don't understand why each time he can't save what they want.

I don't like 3. has we are going to put another notices to the user. We are actually trying to remove the notices we can display.

If 4. is easy and fast to implement, I vote for it.

Tabrisrp:
You also have 5: remove the entry completely, and display a notice for the incorrect entry, like we do for all exclusion fields currently when an incorrect RegEx is added.

This is a system we already have and we can easily plug on

arunbasillal:
Thanks @Tabrisrp 

@hellofromtonya Your feedback reveals an important aspect of removing invalid input - about communicating this to the customers. 

However, this being an edge case, should we work on it right now as part of this update? If we could leave the invalid input as it is, eventually we will know what the invalid inputs are (just like this issue in the first place) and if they are too many, we can think about accommodating all of them. 

What do you think?

arunbasillal:
Sorry I missed the other replies. 

Based on Jonathan's reply, 4 is the best of course. If that's feasible, then sure. But if it's too much work to accommodate an edge case (that we presume without real data), then I feel it's outside the scope of this improvement. 

The problem with option 5 (https://github.com/wp-media/wp-rocket/pull/2413#issuecomment-599530906) is that, the customer will not know that it disappeared, especially if there is a long list of domains.

Tabrisrp:
> The problem with option 5 (#2413 (comment)) is that, the customer will not know that it disappeared, especially if there is a long list of domains.

We can totally display which entries were removed in the error message.

As I said previously, auto-correcting **is** hard. It's neither fast, nor easy. We will have to use external libraries to perform domains validation for example. It's making the plugin more complex, add more code. I want to make sure we balance this vs the edge cases we try to fix.

hellofromtonya:
What if we simplify this to achieve Option 4 for malformed protocols only (for now). If we do that, then we don't need an external package. Rather, we can isolate it down to `http://` and `https://`.

What do you think @Tabrisrp and @GeekPress?

Tabrisrp:
If we focus on the protocol only, it would not be too hard to implement

hellofromtonya:
>If we focus on the protocol only, it would not be too hard to implement

@GeekPress What do you think?

Tabrisrp:
@arunbasillal @GeekPress Can we have a final confirmation on what we go with?

GeekPress:
I'm not sure to understand the solution and what will be the result ^^

Tabrisrp:
**Proposed solution**
Automatically fix incorrect protocol like `http:///` to prevent the edge case Arun found during testing.

GeekPress:
@Tabrisrp Yes we can do the fix like that :)

Tabrisrp:
@vmanthos @piotrbak If one of you can do the QA for the last case I fixed, we could merge this PR. test case : `http:///domain8.com`

@crystinutzaa @Screenfeed @hellofromtonya If you can re-review this for this commit: https://github.com/wp-media/wp-rocket/pull/2413/commits/f217d33c6e3440007da58d9c9786d8c86b08800d

vmanthos:
Starting QA.

PR #2417, opened 2020-03-12
Fixes #2260: WP Smush's lazyload breaks WP Rocket's <iframe> lazyload

### LINKED ISSUE #2260: WP Smush's lazyload breaks WP Rocket's <iframe> lazyload
In WP Smush version [3.4.0](https://i.vgy.me/sAppHK.jpg), they introduced `iframe` lazy loading. When that is enabled along with WP Rocket's `iframe` lazyload feature, the `iframe` isn't displayed.

We have a compatibility feature where we disable our own lazyload feature for images when we detect WP Smush's lazyload:

https://github.com/wp-media/wp-rocket/blob/7723e789d158d2bf00ee5da49ef9158061e158eb/inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php

We don't do that for `iframe` lazyload. We may have to extend our compatibility code to include this.

**Related ticket:** https://secure.helpscout.net/conversation/1056062225/141011?folderId=2135277

**Steps to Reproduce:** 

- Install WP Smush (3.4.0 or later) - https://wordpress.org/plugins/wp-smushit/
- Add an iframe to a page.
- Enable LazyLoad for iframe in WP Smush and WP Rocket.
- Visit the page with the iframe. The iframe will now not be displayed. 
- Disable WP Rocket's LazyLoad for iframe. 
- Visit the page with the iframe. The iframe should be visible now. 

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

Screenfeed:
Depending on Smush lazyload settings, disable WP Rocket lazyload for images and/or iframes.

Closes #2260.


inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
```diff
@@ -24,9 +26,14 @@ public static function get_subscribed_events() {
 			];
 		}
 
+		$prefix = rocket_get_constant( 'WP_SMUSH_PREFIX', 'wp-smush-' );
```
Tabrisrp: this is used here and on line 104 too, maybe we can set a property in the constructor to hold this value?

tests/Integration/SmushSubscriberTestCase.php
```diff
@@ -0,0 +1,55 @@
+<?php
```
Tabrisrp: This TestCase could be in the same folder as the tests for Smush as it's a very specific use case

Screenfeed: Agree, but it will trigger a "Cannot redeclare class" error then.

inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
Screenfeed: The benefit seems very low in terms of performance, it will require to do this on both places:
```php
if ( ! isset( self::$prefix ) ) {
    self::$prefix = rocket_get_constant( 'WP_SMUSH_PREFIX', 'wp-smush-' );
}
```

tests/Integration/SmushSubscriberTestCase.php
Screenfeed: Surprisingly, no error ¯\(°_o)/¯

inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
Tabrisrp: no you're right, and since `get_subscribed_events` is static that doesn't work anyway

inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
```diff
@@ -37,34 +44,88 @@ public static function get_subscribed_events() {
...
+			'iframes' => false,
+		];
+
+		if ( ! class_exists( '\\Smush\\Core\\Settings' ) ) {
+			return $enabled;
+		}
+
+		$settings = Settings::get_instance();
+
+		if ( ! $settings->get( 'lazy_load' ) ) {
```
hellofromtonya: These 2 lines can be combined:

```php
if ( ! Settings::get_instance()->get( 'lazy_load' ) ) {
```

Why? 

- It's less memory -- no need to create a memory loc in the PHP symtable for `$settings`
- Faster -- it takes processing time to:
      - Add a user-defined symbol (i.e. `$settings`) into the PHP symtable
      - Bind that to the object's memory location in the stack
      - And then look up the value for `$settings` on line 100.

inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
```diff
@@ -37,34 +44,88 @@ public static function get_subscribed_events() {
...
+			// Whitelist image formats.
+			[
+				'jpeg' => false,
+				'png'  => false,
+				'gif'  => false,
+				'svg'  => false,
+			]
+		);
+
+		if ( $image_formats ) {
```
hellofromtonya: This is an array. Let's not treat it like a boolean. Check if it's not empty for the conditional.

tests/Integration/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/SmushSubscriberTestCase.php
```diff
@@ -0,0 +1,55 @@
...
+	protected $smush;
+	protected $smush_settings_option_name;
+	protected $smush_settings;
+	protected $smush_lazy_option_name;
+	protected $smush_lazy;
+
+	public function setUp() {
+		parent::setUp();
+
+		$this->subscriber                 = new Smush_Subscriber();
```
hellofromtonya: Pull the subscriber out of the container by using the `apply_filter`. You'll want now that we are injecting options into it.

tests/Integration/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/maybeDeactivateRocketLazyload.php
```diff
@@ -0,0 +1,148 @@
...
+class Test_MaybeDeactivateRocketLazyload extends SmushSubscriberTestCase {
+	private $option_hook_prefix = 'pre_get_rocket_option_';
+	private $rocket_settings;
+
+	public function setUp() {
+		parent::setUp();
+
+		remove_all_filters( $this->option_hook_prefix . 'lazyload' );
+		remove_all_filters( $this->option_hook_prefix . 'lazyload_iframes' );
+		remove_all_filters( 'update_option_wp_rocket_settings' );
```
hellofromtonya: Need to "remember" what is registered to each and then restore at `tearDown()`. Else, it could impact other integration tests.

tests/Integration/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/maybeDeactivateRocketLazyload.php
```diff
@@ -0,0 +1,148 @@
...
+			[
+				'lazyload'         => 1,
+				'lazyload_iframes' => 1,
+			 ]
+		);
+
+		$this->subscriber->maybe_deactivate_rocket_lazyload();
+
+		$this->assertFalse( empty( get_rocket_option( 'lazyload' ) ) );//
+		$this->assertFalse( empty( get_rocket_option( 'lazyload_iframes' ) ) );
```
hellofromtonya: - Use [`assertNotEmpty`](https://phpunit.readthedocs.io/en/7.3/assertions.html#assertempty) here instead.
- Instead of using `get_rocket_option`, create a helper method and pull from the database. Why? What if for some reason `get_rocket_option` is not working as expected. By pulling directly from the database we can ensure the state is what we expected for this test.

tests/Integration/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/maybeDeactivateRocketLazyload.php
```diff
@@ -0,0 +1,148 @@
...
+			[
+				'lazyload'         => 0,
+				'lazyload_iframes' => 0,
+			 ]
+		);
+
+		$this->subscriber->maybe_deactivate_rocket_lazyload();
+
+		$this->assertTrue( empty( get_rocket_option( 'lazyload' ) ) );
+		$this->assertTrue( empty( get_rocket_option( 'lazyload_iframes' ) ) );
```
hellofromtonya: - Use [`assertEmpty`](https://phpunit.readthedocs.io/en/7.3/assertions.html#assertempty) here instead.
- Same about `get_rocket_option()`.

tests/Integration/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/maybeDeactivateRocketLazyload.php
```diff
@@ -0,0 +1,148 @@
...
+			[
+				'lazyload'         => 1,
+				'lazyload_iframes' => 1,
+			 ]
+		);
+
+		$this->subscriber->maybe_deactivate_rocket_lazyload();
+
+		$this->assertTrue( empty( get_rocket_option( 'lazyload' ) ) );
+		$this->assertFalse( empty( get_rocket_option( 'lazyload_iframes' ) ) );//
```
hellofromtonya: Same here.

tests/Integration/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/maybeDeactivateRocketLazyload.php
```diff
@@ -0,0 +1,148 @@
...
+			[
+				'lazyload'         => 1,
+				'lazyload_iframes' => 1,
+			 ]
+		);
+
+		$this->subscriber->maybe_deactivate_rocket_lazyload();
+
+		$this->assertFalse( empty( get_rocket_option( 'lazyload' ) ) );
+		$this->assertTrue( empty( get_rocket_option( 'lazyload_iframes' ) ) );
```
hellofromtonya: Same here.

tests/Integration/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/maybeDeactivateRocketLazyload.php
```diff
@@ -0,0 +1,148 @@
...
+			[
+				'lazyload'         => 1,
+				'lazyload_iframes' => 1,
+			 ]
+		);
+
+		$this->subscriber->maybe_deactivate_rocket_lazyload();
+
+		$this->assertTrue( empty( get_rocket_option( 'lazyload' ) ) );
+		$this->assertTrue( empty( get_rocket_option( 'lazyload_iframes' ) ) );
```
hellofromtonya: Same here.

tests/Unit/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/SmushSubscriberTestCase.php
```diff
@@ -0,0 +1,47 @@
...
+		$this->subscriber = new Smush_Subscriber();
+	}
+
+	protected function mock_is_smush_lazyload_enabled( $lazyload_enabled, array $lazyload_formats ) {
+		$settings = Mockery::mock( 'alias:\\Smush\\Core\\Settings' );
+		$settings
+			->shouldReceive( 'get_instance' )
+			->andReturnUsing(
+				function() use ( $lazyload_enabled, $lazyload_formats ) {
+					$settings = Mockery::mock( '\\WP_Rocket\\Tests\\Fixtures\\ThirdPartyPlugins\\Smush\\Core\\Settings' ); // Don't look for me, I don’t exist.
```
hellofromtonya: 2 options here:

1. Use single `\` instead of double.
2. Or import the class namespace and use `Settings::class`.

tests/Unit/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/maybeDeactivateRocketLazyload.php
```diff
@@ -0,0 +1,136 @@
...
+			->alias(
+				function( $option ) {
+					switch ( $option ) {
+						case 'lazyload':
+						case 'lazyload_iframes':
+							return true;
+					}
+					return false;
+				}
+			);
```
hellofromtonya: This won't be needed anymore as we are removing `get_rocket_option` and injecting the option instance instead.

tests/Unit/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/maybeDeactivateRocketLazyload.php
```diff
@@ -0,0 +1,136 @@
...
```
hellofromtonya: Same here. You'll need to set up the option states instead.

tests/Unit/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/maybeDeactivateRocketLazyload.php
```diff
@@ -0,0 +1,136 @@
...
+		Functions\expect( 'update_rocket_option' )
+			->once()
+			->with( 'lazyload', 0 )
+			->never()
+			->with( 'lazyload_iframes', 0 );
```
hellofromtonya: Here too.

tests/Unit/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/maybeDeactivateRocketLazyload.php
```diff
@@ -0,0 +1,136 @@
...
+							return true;
+					}
+					return false;
+				}
+			);
+		Functions\expect( 'update_rocket_option' )
+			->once()
+			->with( 'lazyload', 0 )
+			->once()
+			->with( 'lazyload_iframes', 0 );
```
hellofromtonya: Both of these too.

tests/Unit/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/maybeDeactivateRocketLazyload.php
```diff
@@ -0,0 +1,136 @@
...
+							return true;
+					}
+					return false;
+				}
+			);
+		Functions\expect( 'update_rocket_option' )
+			->never()
+			->with( 'lazyload', 0 )
+			->once()
+			->with( 'lazyload_iframes', 0 );
```
hellofromtonya: Both of these too.

inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
```diff
@@ -24,47 +41,127 @@ public static function get_subscribed_events() {
...
+		$enabled = [
+			'images'  => false,
+			'iframes' => false,
+		];
+
+		if ( ! class_exists( '\Smush\Core\Settings' ) ) {
+			return $enabled;
+		}
+
+		if ( ! Settings::get_instance()->get( 'lazy_load' ) ) {
```
Tabrisrp: Shouldn't we also add a `method_exist` check, we never know what they could decide to change in the class in the future?

inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
```diff
@@ -24,47 +41,127 @@ public static function get_subscribed_events() {
...
+			return $enabled;
+		}
+
+		$prefix  = rocket_get_constant( 'WP_SMUSH_PREFIX', 'wp-smush-' );
+		$formats = Settings::get_instance()->get_setting( $prefix . 'lazy_load' );
```
Tabrisrp: Here too, safe guard as much as we can

inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
```diff
@@ -24,47 +41,127 @@ public static function get_subscribed_events() {
...
+		$formats = ! empty( $formats['format'] ) && is_array( $formats['format'] ) ? array_filter( $formats['format'] ) : [];
+
+		$image_formats = array_intersect_key(
+			$formats,
+			// Whitelist image formats.
```
Tabrisrp:
```suggestion
			// Allowlist image formats.
```

tests/Unit/inc/classes/subscriber/third-party/plugins/Smush_Subscriber/maybeDeactivateRocketLazyload.php
```diff
@@ -0,0 +1,123 @@
+<?php
+namespace WP_Rocket\Tests\Unit\inc\classes\third_party\plugins\Smush_Subscriber;
+
+//use Brain\Monkey\Functions;
```
Tabrisrp:
```suggestion
```

Screenfeed:
@Tabrisrp 
>`isSmushLazyloadActive`: you can use data providers to simplify the tests code, and make it easier to add more tests cases

What do you mean? Which part can be simplified?

>`maybeDeactivateRocketLazyload`: I feel the assertions/expectations are too hidden currently, maybe renaming `setSubscriber` to something else would make it clearer?

What do you think of `setAssertsAndSubscriber`?

Tabrisrp:
Instead of calling `mock_is_smush_lazyload_enabled` multiple times with different values, you could have a dataProvider for the 2 parameters of the method `$enabled` and `values` for example.

That way you would have only one method call, and with the dataProvider, it will automatically test all cases.

> What do you think of `setAssertsAndSubscriber`?
Yes I think it's better

Screenfeed:
Ah, I understand now. I'll take a look at this, thanks for the clarification.

Screenfeed:
One small issue in this case though: the subscriber will be instantiated for each data set instead of once for the whole set.

Tabrisrp:
@Screenfeed looking good with the use of data providers 👍 

inc/classes/subscriber/third-party/plugins/class-smush-subscriber.php
```diff
@@ -1,6 +1,9 @@
 <?php
```
hellofromtonya: @Screenfeed and @Tabrisrp Since we are touching this file, what do you think about move it to `inc/ThirdParty/Plugins/Smush/Subscriber.php` and renaming the class? We could also move any other smush files into this location. Thoughts?

Tabrisrp: I agree

Screenfeed: I will use this architecture. However, there is a sub-folder in `Plugins` currently, for the `AMP` plugin: `inc/ThirdParty/Plugins/Optimization/AMP.php`. Maybe we should harmonize.

Screenfeed: Done.

arunbasillal:
While we do have #2439 to improve the UI, can we include grey-ing out of the option with this update? The same way we do for `LazyLoad for images`? 

If not, as Piotr mentioned, users will be able to enable the option again and have the original issue again. 

#2439 is more about improving the messaging and handling the repetitive notices to customers.

Tabrisrp:
This would push this out of 3.5.2, so it's basically product team call at this point.

hellofromtonya:
If we wish to do the UI changes, I'd suggest moving this PR to 3.5.3. 

@arunbasillal and @GeekPress what would you like to do with this particular PR/fix?

arunbasillal:
Imo, greying out of the UI is mandatory to call this issue fixed. Sorry, the AC wasn't more specific (updated it). 

Over to @GeekPress for the final decision.

GeekPress:
This issue has been rescheduled for 3.5.3 and sprint 3.

Screenfeed:
This is the "new" UI: the checkbox for iframes is disabled if Smush’s option is enabled.
Here we see that loazyload for images is disabled by Avada and lazyload for iframes is disabled by Smush.

Capture d'écran 2020-04-27 19 05 39

inc/classes/admin/settings/class-page.php
```diff
@@ -888,21 +888,57 @@ private function media_section() {
...
+		$disable_iframes_lazyload = [];
+
+		if ( rocket_avada_maybe_disable_lazyload() ) {
+			$disable_images_lazyload[] = __( 'Avada', 'rocket' );
+		}
+
 		if ( rocket_maybe_disable_lazyload() ) {
-			$rocket_maybe_disable_lazyload_plugins[] = __( 'Autoptimize', 'rocket' );
+			$disable_images_lazyload[]  = __( 'Autoptimize', 'rocket' );
+			$disable_iframes_lazyload[] = __( 'Autoptimize', 'rocket' );
```
Tabrisrp, line 900:
```suggestion
```
I double-checked and Autoptimize doesn't have iframes lazyload, so this can be removed

inc/classes/admin/settings/class-page.php
```diff
@@ -888,21 +888,57 @@ private function media_section() {
...
-		 * @param array Will return the array with all plugin names which should disable LazyLoad
+		 * @param array $disable_images_lazyload Will return the array with all plugin names which should disable LazyLoad
+		 */
+		$disable_images_lazyload = apply_filters( 'rocket_maybe_disable_lazyload_helper', $disable_images_lazyload );
+		$disable_images_lazyload = array_unique( array_filter( (array) $disable_images_lazyload ) );
+
+		if ( $disable_images_lazyload ) {
+			foreach ( $disable_images_lazyload as $i => $plugin_name ) {
+				$disable_images_lazyload[ $i ] = sprintf( '<strong>%s</strong>', $plugin_name );
+			}
```
Tabrisrp, line 917: could use a small `array_map` instead?

inc/classes/admin/settings/class-page.php
```diff
@@ -888,21 +888,57 @@ private function media_section() {
...
 		 */
-		$rocket_maybe_disable_lazyload_plugins = apply_filters( 'rocket_maybe_disable_lazyload_helper', $rocket_maybe_disable_lazyload_plugins );
-		$rocket_maybe_disable_lazyload_plugins = wp_sprintf_l( '%l', $rocket_maybe_disable_lazyload_plugins );
+		$disable_iframes_lazyload = apply_filters( 'rocket_maybe_disable_iframes_lazyload_helper', $disable_iframes_lazyload );
+		$disable_iframes_lazyload = array_unique( array_filter( (array) $disable_iframes_lazyload ) );
+
+		if ( $disable_iframes_lazyload ) {
+			foreach ( $disable_iframes_lazyload as $i => $plugin_name ) {
+				$disable_iframes_lazyload[ $i ] = sprintf( '<strong>%s</strong>', $plugin_name );
+			}
```
Tabrisrp, line 933: same here, `array_map`

inc/classes/admin/settings/class-page.php
```diff
@@ -888,21 +888,57 @@ private function media_section() {
...
 		 * @since  3.4.2
 		 * @author Soponar Cristina
 		 *
-		 * @param array Will return the array with all plugin names which should disable LazyLoad
+		 * @param array $disable_images_lazyload Will return the array with all plugin names which should disable LazyLoad
+		 */
+		$disable_images_lazyload = apply_filters( 'rocket_maybe_disable_lazyload_helper', $disable_images_lazyload );
+		$disable_images_lazyload = array_unique( array_filter( (array) $disable_images_lazyload ) );
+
+		if ( $disable_images_lazyload ) {
```
Tabrisrp, line 914:
```suggestion
		if ( ! empty( $disable_images_lazyload ) ) {
```

inc/classes/admin/settings/class-page.php
```diff
@@ -888,21 +888,57 @@ private function media_section() {
...
+		 * @since 3.5.5
+		 *
+		 * @param array $disable_iframes_lazyload Will return the array with all plugin names which should disable LazyLoad
 		 */
-		$rocket_maybe_disable_lazyload_plugins = apply_filters( 'rocket_maybe_disable_lazyload_helper', $rocket_maybe_disable_lazyload_plugins );
-		$rocket_maybe_disable_lazyload_plugins = wp_sprintf_l( '%l', $rocket_maybe_disable_lazyload_plugins );
+		$disable_iframes_lazyload = apply_filters( 'rocket_maybe_disable_iframes_lazyload_helper', $disable_iframes_lazyload );
+		$disable_iframes_lazyload = array_unique( array_filter( (array) $disable_iframes_lazyload ) );
+
+		if ( $disable_iframes_lazyload ) {
```
Tabrisrp, line 930:
```suggestion
		if ( ! empty( $disable_iframes_lazyload ) ) {
```

inc/classes/admin/settings/class-page.php
```diff
@@ -888,21 +888,46 @@ private function media_section() {
 			]
 		);
 
-		$rocket_maybe_disable_lazyload_plugins = [];
+		$disable_images_lazyload  = [];
```
hellofromtonya, line 891: We touched this file. It should be moved into our new architecture, renamed, and the old deprecated.

inc/classes/admin/settings/class-page.php
```diff
@@ -888,21 +888,46 @@ private function media_section() {
...
+		 * @param array $disable_iframes_lazyload Will return the array with all plugin names which should disable LazyLoad
 		 */
-		$rocket_maybe_disable_lazyload_plugins = apply_filters( 'rocket_maybe_disable_lazyload_helper', $rocket_maybe_disable_lazyload_plugins );
-		$rocket_maybe_disable_lazyload_plugins = wp_sprintf_l( '%l', $rocket_maybe_disable_lazyload_plugins );
+		$disable_iframes_lazyload = apply_filters( 'rocket_maybe_disable_iframes_lazyload_helper', $disable_iframes_lazyload );
+		$disable_iframes_lazyload = array_unique( array_filter( (array) $disable_iframes_lazyload ) );
+		$disable_iframes_lazyload = array_map( 'sprintf', array_fill( 0, count( $disable_iframes_lazyload ), '<strong>%s</strong>' ), $disable_iframes_lazyload );
+
+		$disable_lazyload = array_merge( $disable_images_lazyload, $disable_iframes_lazyload );
+		$disable_lazyload = array_unique( $disable_lazyload );
```
hellofromtonya, line 926: There are a few considerations in this code that we should discuss:

1. What if the filters return an empty value? Then `array_filter` and `array_unique` are not needed. As these are expensive array functions, skipping them when there's no value is more performant.
2. The design pattern repeats. Move the `array_filter`, `array_unique`, `array_fill`, and `array_map` to a private method and return the value for each.

inc/classes/admin/settings/class-page.php
```diff
@@ -888,21 +888,46 @@ private function media_section() {
...
+
+		$disable_lazyload         = wp_sprintf_l( '%l', $disable_lazyload );
+		$disable_images_lazyload  = wp_sprintf_l( '%l', $disable_images_lazyload );
+		$disable_iframes_lazyload = wp_sprintf_l( '%l', $disable_iframes_lazyload );
```
hellofromtonya, line 930: Why do we need to run both of these through `wp_sprintf_l()`. I don't see where we are rendering the values in each below. What am I missing?

Screenfeed: `$disable_iframes_lazyload` is not anymore indeed. `$disable_images_lazyload` is used [here](https://github.com/wp-media/wp-rocket/pull/2417/files#diff-e277836ca30970aac97804f2ca6ffbaeR1002).

hellofromtonya:
@Tabrisrp Please re-review this one. As we touched the `Admin\Settings\Page` class, it's service provider, and subscriber, I went ahead and moved all of the files for `Admin\Settings\` into the new architecture. Now everything is together instead of distributed. Check and make sure I didn't miss anything in the moves.

inc/Engine/Admin/Settings/Page.php
```diff
@@ -904,11 +865,10 @@ private function media_section() {
 		 * Lazyload Helper filter which disables WPR lazyload functionality for images.
 		 *
 		 * @since  3.4.2
-		 * @author Soponar Cristina
 		 *
 		 * @param array $disable_images_lazyload Will return the array with all plugin names which should disable LazyLoad
 		 */
-		$disable_images_lazyload = apply_filters( 'rocket_maybe_disable_lazyload_helper', $disable_images_lazyload );
+		$disable_images_lazyload = (array) apply_filters( 'rocket_maybe_disable_lazyload_helper', $disable_images_lazyload );
```
Screenfeed, line 871: Casting to array is not needed since `sanitize_and_format_list()`’s first test is `is_array()`.

hellofromtonya: However, if in the future, we modify the code to do more processing with this variable, this code makes sense it's the data type we want.

The better approach is to set the data type where it has the potential to change. That way, we always know it's set to what we expect throughout our code.

Screenfeed: Yes, it makes sense.

hellofromtonya:
>@hellofromtonya There is a fatal error now, thrown by the container, I'm investigating.

@Screenfeed Find it. I missed updating the namespace to the ServiceProvider in the `Plugin`. Whoopsie. Pushing shortly.

Screenfeed:
@hellofromtonya Nice, thanks for finding it 👍

inc/deprecated/3.6.php
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * Class aliases.
+ */
+class_alias( '\WP_Rocket\ThirdParty\Plugins\Smush', '\WP_Rocket\Subscriber\Third_Party\Plugins\Smush_Subscriber' );
+class_alias( '\WP_Rocket\Admin\Settings\Page', '\WP_Rocket\Engine\Admin\Settings\Page' );
+class_alias( '\WP_Rocket\Admin\Settings\Render', '\WP_Rocket\Engine\Admin\Settings\Render' );
+class_alias( '\WP_Rocket\Admin\Settings\Settings', '\WP_Rocket\Engine\Admin\Settings\Settings' );
+class_alias( '\WP_Rocket\ServiceProvider\Settings', '\WP_Rocket\Engine\Admin\Settings\ServiceProvider' );
+class_alias( '\WP_Rocket\Subscriber\Admin\Settings\Page_Subscriber', '\WP_Rocket\Engine\Admin\Settings\Subscriber' );
```
Tabrisrp, line 14: - since it's a PR going into a minor 3.5 version, it should be in the 3.5 deprecated file, not 3.6
- Shouldn't it use a double `\\` instead of a simple? Or are both fine?

inc/Engine/Admin/Settings/Subscriber.php
```diff
@@ -1,18 +1,18 @@
...
 use WP_Rocket\Event_Management\Subscriber_Interface;
-use WP_Rocket\Admin\Settings\Page;
 
 defined( 'ABSPATH' ) || exit;
 
 /**
- * WP Rocket settings page subscriber
+ * WP Rocket settings page subscriber.
  *
+ * @since 3.6 Moves into the new architecture.
```
Tabrisrp, line 12:
```suggestion
 * @since 3.5.5 Moves into the new architecture.
```

Possibly it's not even needed

inc/Engine/Admin/Settings/Settings.php
```diff
@@ -1,17 +1,18 @@
...
-namespace WP_Rocket\Admin\Settings;
+namespace WP_Rocket\Engine\Admin\Settings;
 
 use WP_Rocket\Subscriber\Third_Party\Plugins\Security\Sucuri_Subscriber;
 
 /**
- * Settings class
+ * Settings class.
+ *
+ * @since 3.6 Moves into the new architecture.
```
Tabrisrp, line 9:
```suggestion
 * @since 3.5.5 Moves into the new architecture.
```

inc/Engine/Admin/Settings/ServiceProvider.php
```diff
@@ -1,15 +1,15 @@
...
-namespace WP_Rocket\ServiceProvider;
+namespace WP_Rocket\Engine\Admin\Settings;
 
 use League\Container\ServiceProvider\AbstractServiceProvider;
 
 /**
- * Service provider for the WP Rocket settings
+ * Service provider for the WP Rocket settings.
  *
+ * @since 3.6 Moves into the new architecture.
```
Tabrisrp, line 9:
```suggestion
 * @since 3.5.5 Moves into the new architecture.
```

inc/Engine/Admin/Settings/Render.php
```diff
@@ -8,35 +8,32 @@
 /**
  * Handle rendering of HTML content for the settings page.
  *
+ * @since 3.6 Moves into the new architecture.
```
Tabrisrp, line 11:
```suggestion
 * @since 3.5.5 Moves into the new architecture.
```

inc/Engine/Admin/Settings/Page.php
```diff
@@ -1,92 +1,84 @@
...
 
 use WP_Rocket\Admin\Database\Optimization;
 use WP_Rocket\Engine\Admin\Beacon\Beacon;
 use WP_Rocket\Interfaces\Render_Interface;
 
 /**
- * Registers the admin page and WP Rocket settings
+ * Registers the admin page and WP Rocket settings.
  *
+ * @since 3.6 Moves into the new architecture.
```
Tabrisrp, line 11:
```suggestion
 * @since 3.5.5 Moves into the new architecture.
```

inc/deprecated/3.6.php
hellofromtonya: Darn it. Totally missed that it's based to `develop`. Doh. I'll fix it up.

A simple `\` is all that's needed. Think of how we do it in the ServiceProviders. So to answer your question: both work.

inc/deprecated/3.5.php
```diff
@@ -6,6 +6,11 @@
 /**
  * Class aliases.
  */
+class_alias( '\WP_Rocket\Admin\Settings\Page', '\WP_Rocket\Engine\Admin\Settings\Page' );
+class_alias( '\WP_Rocket\Admin\Settings\Render', '\WP_Rocket\Engine\Admin\Settings\Render' );
+class_alias( '\WP_Rocket\Admin\Settings\Settings', '\WP_Rocket\Engine\Admin\Settings\Settings' );
+class_alias( '\WP_Rocket\ServiceProvider\Settings', '\WP_Rocket\Engine\Admin\Settings\ServiceProvider' );
+class_alias( '\WP_Rocket\Subscriber\Admin\Settings\Page_Subscriber', '\WP_Rocket\Engine\Admin\Settings\Subscriber' );
 class_alias( '\\WP_Rocket\\Engine\\Preload\\AbstractPreload', '\\WP_Rocket\\Preload\\Abstract_Preload' );
```
hellofromtonya, line 14: The class aliases that were already in this file are backwards:

```php
class_alias( 'the original class', 'the new class' );
```

[See the PHP manual](https://www.php.net/manual/en/function.class-alias.php). But let's fix these in a separate PR.

inc/deprecated/3.6.php
hellofromtonya: Fixed in 577a1bda3500efc262104bc024abf99c76f543b0.

hellofromtonya:
@wp-media/dev-qa @arunbasillal Please repeat the QA on this one. Changes since you last did QA on it:

- moved all of the files for the `Admin\Settings` module into the new architecture. They used to be in multiple folders
- added class aliases for the original files (i.e. in the 3.5 deprecated file) to provide backwards compatibility should a 3rd party use one of the moved files

Nothing in the Smush code was changed since you last did QA on it.

Screenfeed:
I'm surprised that the tests are showing a green light with that much warnings :(
[`class_alias()` was used the right way.](https://github.com/wp-media/wp-rocket/pull/2596#issuecomment-621498874)

arunbasillal:
@hellofromtonya I am seeing these warnings - http://snippi.com/raw/1t1p8ks
I thought it was my local, so I deleted my vendor folder and installed again. Everything persists.

Tabrisrp:
Indeed, Travis is showing ok because the tests themselves pass, but there is PHP warnings because of the `class_alias` change. It needs to be reverted

hellofromtonya:
@arunbasillal The class_alias problem (caused by me yesterday when I lost my mind 🙄) is now fixed.

Credit to @Screenfeed

PR #2420, opened 2020-03-12
Use filter_var instead of filter_input to validate REMOTE_ADDR

### LINKED ISSUE #2421: filter_input() function returns NULL instead of the correct value with FastCgi
**Describe the bug**
Website with Cloudflare flexible SSL. After update to 3.5, too many redirects issue happens in the browser.

`filter_input` is returning a NULL value instead of the IP when we check for it in the Cloudflare flexible SSL code: https://github.com/wp-media/wp-rocket/blob/a45f7625f202c775b080efa4e7ab18fc7791df37/inc/common/cloudflare-flexible-ssl.php#L35

There seems to be a bug with filter_input and FastCGI, so that would impact some setups: https://www.php.net/manual/en/function.filter-input.php#115960

**Expected behavior**
Website should be loading as expected.

**Solution**
Using filter_var with the global SERVER variable directly solves the issue. Confirmed on the customer website.

**HS ticket**
https://secure.helpscout.net/conversation/1106736317/150693?folderId=637040

remyperona:
Customer website was in a redirect loop after update to 3.5.

NOTE (added by Tonya): The code in question is from 3.4.1.

After digging, I found out that `filter_input` was returning a `NULL` value instead of the IP when we check for it in the Cloudflare flexible SSL code.

There seems to be a bug with `filter_input` and FastCGI, so that would impact some setups: https://www.php.net/manual/en/function.filter-input.php#115960

Using `filter_var` with the global SERVER variable directly solves the issue. Confirmed on the customer website.

Closes #2421

### DIFF
--- a/inc/common/cloudflare-flexible-ssl.php
+++ b/inc/common/cloudflare-flexible-ssl.php
@@ -31,7 +31,7 @@ function rocket_is_cloudflare() {
  * @return bool
  */
 function rocket_is_cf_ip() {
-	if ( ! isset( $_SERVER['REMOTE_ADDR'] ) ) {
+	if ( empty( $_SERVER['REMOTE_ADDR'] ) ) {
 		return false;
 	}
 
--- a/inc/common/cloudflare-flexible-ssl.php
+++ b/inc/common/cloudflare-flexible-ssl.php
@@ -38,7 +38,7 @@ function rocket_is_cf_ip() {
 	// Store original remote address in $original_ip.
 	$original_ip = filter_var( wp_unslash( $_SERVER['REMOTE_ADDR'] ), FILTER_VALIDATE_IP );
 
-	if ( ! isset( $original_ip ) ) {
+	if ( empty( $original_ip ) ) {
 		return false;
 	}
 
--- a/inc/common/cloudflare-flexible-ssl.php
+++ b/inc/common/cloudflare-flexible-ssl.php
@@ -31,8 +31,13 @@ function rocket_is_cloudflare() {
  * @return bool
  */
 function rocket_is_cf_ip() {
+	if ( ! isset( $_SERVER['REMOTE_ADDR'] ) ) {
+		return false;
+	}
+
 	// Store original remote address in $original_ip.
-	$original_ip = filter_input( INPUT_SERVER, 'REMOTE_ADDR', FILTER_VALIDATE_IP );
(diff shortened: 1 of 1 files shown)


inc/common/cloudflare-flexible-ssl.php
```diff
@@ -31,8 +31,13 @@ function rocket_is_cloudflare() {
...
 function rocket_is_cf_ip() {
+	if ( ! isset( $_SERVER['REMOTE_ADDR'] ) ) {
+		return false;
+	}
+
 	// Store original remote address in $original_ip.
-	$original_ip = filter_input( INPUT_SERVER, 'REMOTE_ADDR', FILTER_VALIDATE_IP );
+	$original_ip = filter_var( wp_unslash( $_SERVER['REMOTE_ADDR'] ), FILTER_VALIDATE_IP );
+
 	if ( ! isset( $original_ip ) ) {
```
Screenfeed: `filter_var()` doesn't return `null`, but `false`.
```suggestion
	if ( empty( $original_ip ) ) {
```

Screenfeed (changes requested):

Screenfeed (approved):

crystinutzaa (approved):

hellofromtonya:
@Tabrisrp We use `filter_input()` throughout the plugin. Is this the only place where it impacted the issue?

hellofromtonya (approved):

hellofromtonya:
This is ready for QA. @arunbasillal @piotrbak @vmanthos

hellofromtonya:
This PR is a bandaid. We are switching from a PHP specific filtering handler to one for vars. Why? There's something in the customer's stack or .ini configuration that is interfering. A bandaid of switching to `filter_var` provides the customer relief for whatever is wrong in his stack.

My assessment here is this:

- This is a bandaid.
- `filter_input()` is more secure and built for this instance.
- Rocket is not influencing or contributing to the problem. It's not the root cause.
- Safe and secure enough:
     - This code is not saving to the database, writing to a file, or rendering to the browser. Therefore, the implementation in this PR is safe and secure enough for the intent of how this works.
- I'll keep working with the PHP core team to find out why and what the right approach.

hellofromtonya (changes requested):
This PR is blocked from merging until:

1. We can reproduce the problem.
2. We have confidence that Rocket’s code is the root cause of the problem.
3. We have confidence that this PR won’t negatively impact any other customers.

PR #2451, opened 2020-03-18
[WIP] Fixes Too Few Arguments in pre_post_update hook

crystinutzaa:
Fixes Fatal Error: Too few arguments on pre_post_update hook calls.

It seems the second parameter is not defined while saving products in drafts. I tried to recreate it on my local (WP + WOO) but I could not recreate it, so I assume is a plugin affecting this one.

Ticket:
https://secure.helpscout.net/conversation/1088587705/147100?folderId=3083222

### DIFF
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -283,7 +283,7 @@ add_action( 'wp_update_comment_count', 'rocket_clean_post' );
  * @param int   $post_id   The post ID.
  * @param array $post_data Array of unslashed post data.
  */
-function rocket_clean_post_cache_on_status_change( $post_id, $post_data ) {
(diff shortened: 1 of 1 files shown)


hellofromtonya:
Closing this PR as the problem is due to a 3rd party plugin, theme, or must-use, where it has a field that is not prefixed and is called "update". When that field runs through `sanitize_post_field`, it fires a filter inside of that function that calls `"pre_post_{$field}"`. 

As `'pre_post_update'` is a defined WordPress Core action hook event, it is expected to behave [per Codex](https://developer.wordpress.org/reference/hooks/pre_post_update/) and the Core source code. 

But in this case, a 3rd party file is setting on that behavior. It impacts on many many plugins and themes, such as WooCommerce and others.

PR #2459, opened 2020-03-19
Fix #2216 Update AMP compatibility with ?amp

### LINKED ISSUE #2216: Update AMP compatibility with ?amp
The official AMP plugin is now using `?amp` as the default for URLs instead of  `/amp/`

ref: https://wordpress.org/support/topic/change-amp-to-amp-and-amp-home-page-not-working/#post-11636178

This means WP Rocket won't cache amp pages at all.

**Steps to reproduce**

1. Install AMP plugin - https://wordpress.org/plugins/amp/
2. Go to Amp > General
3. Choose Transitional template mode
4. Now for each page on your site the AMP equivalent is available at example.com/?amp - this isn't cached

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
Closes #2216 

Identify the root cause ✅
Transitional mode uses the ?amp query string, which is not cached by default by WP Rocket.

Scope a solution ✅
We can update our AMP compatibility code to add the amp query string to the cached query strings automatically, when transitional mode is enabled:

Moved code to OOP and Subscriber. 
Used `get_rocket_option_cache_query_strings` hook instead of activate / deactivate / update option events from AMP plugin.

Estimate the effort ✅
Effort: [S]

Most of the effort will be spent on writing automated tests.

TO DO:
- [x] Unit Tests
- [x] IntegrationTests
- [ ] QA tests


inc/classes/subscriber/third-party/plugins/Mobile/class-amp-subscriber.php
```diff
@@ -0,0 +1,82 @@
...
+	 * @since  3.5.2
+	 * @author Soponar Cristina
+	 * @inheritDoc
+	 */
+	public static function get_subscribed_events() {
+		$events = [];
+
+		if ( function_exists( 'is_amp_endpoint' ) ) {
+			$events['wp']                                    = 'disable_options_on_amp';
+			$events['get_rocket_option_cache_query_strings'] = 'is_amp_compatible_callback';
```
Tabrisrp:
```suggestion
			$events['rocket_cache_query_strings'] = 'is_amp_compatible_callback';
```

composer.json
```diff
@@ -81,17 +88,19 @@
 	},
 	"scripts": {
 		"test-unit": "\"vendor/bin/phpunit\" --testsuite unit --colors=always --configuration tests/Unit/phpunit.xml.dist",
-		"test-integration": "\"vendor/bin/phpunit\" --testsuite integration --colors=always --configuration tests/Integration/phpunit.xml.dist --exclude-group AdminOnly,WithWoo,DoCloudflare",
+		"test-integration": "\"vendor/bin/phpunit\" --testsuite integration --colors=always --configuration tests/Integration/phpunit.xml.dist --exclude-group AdminOnly,WithWoo,WithAmp,DoCloudflare",
 		"test-integration-adminonly": "\"vendor/bin/phpunit\" --testsuite integration --colors=always --configuration tests/Integration/phpunit.xml.dist --group AdminOnly",
 		"test-integration-docloudflare": "\"vendor/bin/phpunit\" --testsuite integration --colors=always --configuration tests/Integration/phpunit.xml.dist --group DoCloudflare",
 		"test-integration-withwoo": "\"vendor/bin/phpunit\" --testsuite integration --colors=always --configuration tests/Integration/phpunit.xml.dist --group WithWoo",
+		"test-integration-withamp": "\"vendor/bin/phpunit\" --testsuite integration --colors=always --configuration tests/Integration/phpunit.xml.dist --group WithAmp",
```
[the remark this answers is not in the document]
crystinutzaa: On test-integration WithAmp is on --exclude-group so that WithAMP will not run.

`

inc/classes/subscriber/third-party/plugins/Mobile/class-amp-subscriber.php
```diff
@@ -0,0 +1,82 @@
...
+
+		remove_filter( 'wp_resource_hints', 'rocket_dns_prefetch', 10, 2 );
+		add_filter( 'do_rocket_lazyload', '__return_false' );
+		unset( $wp_filter['rocket_buffer'] );
+
+		// this filter is documented in inc/front/protocol.php.
+		$do_rocket_protocol_rewrite = apply_filters( 'do_rocket_protocol_rewrite', false ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
+
+		if ( ( get_rocket_option( 'do_cloudflare', 0 ) && get_rocket_option( 'cloudflare_protocol_rewrite', 0 ) || $do_rocket_protocol_rewrite ) ) {
+			remove_filter( 'rocket_buffer', 'rocket_protocol_rewrite', PHP_INT_MAX );
```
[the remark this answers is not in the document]
crystinutzaa: This is the old function, which I did not revisit the code at all. I simply modified `get_rocket_option` into `$this->options->get` .

inc/classes/subscriber/third-party/plugins/Mobile/class-amp-subscriber.php
crystinutzaa: Thank you for this suggestion

crystinutzaa:
@Remy please check if this is correct: I moved the class to namespace: `WP_Rocket\Subscriber\Third_Party\Plugins\Optimization` ->path: `inc/classes/subscriber/third-party/plugins/Optimization`

> There should also be a callback on update_option_amp-options, to update the query strings & config file when the option is changed

There is no need to add any callback on update_option_amp-options because the operation is reversed: on `rocket_cache_query_strings` filter it checks if AMP plugin is enabled and if is in transitional mode. If yes, it adds amp query string, so updating AMP configuration does not require any callback.

inc/classes/subscriber/third-party/plugins/Optimization/class-amp-subscriber.php
```diff
@@ -0,0 +1,100 @@
...
+
+		remove_filter( 'wp_resource_hints', 'rocket_dns_prefetch', 10, 2 );
+		add_filter( 'do_rocket_lazyload', '__return_false' );
+		unset( $wp_filter['rocket_buffer'] );
+
+		// this filter is documented in inc/front/protocol.php.
+		$do_rocket_protocol_rewrite = apply_filters( 'do_rocket_protocol_rewrite', false ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
+
+		if ( ( $this->options->get( 'do_cloudflare', 0 ) && $this->options->get( 'cloudflare_protocol_rewrite', 0 ) || $do_rocket_protocol_rewrite ) ) {
+			remove_filter( 'rocket_buffer', 'rocket_protocol_rewrite', PHP_INT_MAX );
```
Tabrisrp:
```suggestion
```

Anything on `rocket_buffer` is already removed before by the `unset`

remy:
Yep LGTM 👍 (don't know what this is)

crystinutzaa:
@Tabrisrp I totally forgot that this change needs to be updated in config file. I added hooks which will call: rocket_generate_config_file() on activate, deactivate & update options.

Tabrisrp:
@crystinutzaa Looks great 👍 I did a small change in this commit: https://github.com/wp-media/wp-rocket/pull/2459/commits/18e486bdde7582fd6f6dc1c359437650376920ee

Set the Options_Data mock in the setUp method, as well as the AMP instantiation. Then on individual tests you can still use those without having to reinstantiate each time, it still works. That way it's a bit faster and cleaner.

inc/classes/ServiceProvider/class-third-party-subscribers.php
```diff
@@ -33,6 +33,7 @@ class Third_Party_Subscribers extends AbstractServiceProvider {
 		'ewww_webp_subscriber',
 		'optimus_webp_subscriber',
 		'bigcommerce_subscriber',
+		'amp_subscriber',
```
hellofromtonya: Since we are touching this file, let's move it into `inc/ThirdParty/ServiceProvider.php` and rename the class to  `ServiceProvider`.

inc/classes/ServiceProvider/class-third-party-subscribers.php
```diff
@@ -61,5 +62,7 @@ public function register() {
 			->withArgument( $this->getContainer()->get( 'options' ) );
 		$this->getContainer()->share( 'optimus_webp_subscriber', 'WP_Rocket\Subscriber\Third_Party\Plugins\Images\Webp\Optimus_Subscriber' );
 		$this->getContainer()->share( 'bigcommerce_subscriber', 'WP_Rocket\Subscriber\Third_Party\Plugins\Ecommerce\BigCommerce_Subscriber' );
+		$this->getContainer()->share( 'amp_subscriber', 'WP_Rocket\ThirdParty\Plugins\Optimization\AMP' )
+			->withArgument( $this->getContainer()->get( 'options' ) );
```
hellofromtonya: Each of these subscribers is using a separate instance of options. That's less performant.

Since we're touching this file, let's make it more performant. Create one instance and then use that variable in each of these:

```php
$options = $this->getContainer()->get( 'options' );
```

inc/ThirdParty/Plugins/Optimization/AMP.php
```diff
@@ -0,0 +1,112 @@
...
+		global $wp_filter;
+
+		remove_filter( 'wp_resource_hints', 'rocket_dns_prefetch', 10, 2 );
+		add_filter( 'do_rocket_lazyload', '__return_false' );
+		unset( $wp_filter['rocket_buffer'] );
+
+		// this filter is documented in inc/front/protocol.php.
+		$do_rocket_protocol_rewrite = apply_filters( 'do_rocket_protocol_rewrite', false ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
+
+		if ( ( $this->options->get( 'do_cloudflare', 0 ) && $this->options->get( 'cloudflare_protocol_rewrite', 0 ) || $do_rocket_protocol_rewrite ) ) {
```
hellofromtonya: To be more performant, we should:

- do the `apply_filters` within the conditional
- don't assign it to a variable <- Why? The variable is not used except for the conditional

I'd suggest deleting line 105 and changing the conditional to this:

```php
if ( 
    (bool) $this->options->get( 'do_cloudflare', 0 ) 
    && 
    (bool) $this->options->get( 'cloudflare_protocol_rewrite', 0 )
    || 
    (bool) apply_filters( 'do_rocket_protocol_rewrite', false ) 
) {
```

Question: @crystinutzaa Check the inside `( )`, i.e. as it's double wrapping the entire conditional expression. Did you intend to set priority to part of the conditional?

inc/ThirdParty/Plugins/Optimization/AMP.php
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	/**
+	 * Add compatibility with AMP query string by adding it as a cached query string.
+	 *
+	 * @since  3.5.2
+	 * @author Soponar Cristina
+	 *
+	 * @param array $value WP Rocket cache_query_strings value.
+	 */
```
hellofromtonya: Add `@return`

tests/Integration/inc/ThirdParty/Plugins/Optimization/AMP/TestCase.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace WP_Rocket\Tests\Integration\inc\ThirdParty\Plugins\Optimization\AMP;
+
+use WPMedia\PHPUnit\Integration\TestCase as BaseTestCase;
+
+abstract class TestCase extends BaseTestCase {
+	protected function setSettings( $option, $value ) {
+		\AMP_Options_Manager::update_option( $option, $value);
```
hellofromtonya: Import `AMP_Options_Manager` instead of `\`.

inc/ThirdParty/Plugins/Optimization/AMP.php
Tabrisrp: I updated the conditional, what we want:
- `do_cloudflare` must be `true`
- either `cloudflare_protocol_rewrite` is true or `do_protocol_rewrite` is true

tests/Integration/inc/ThirdParty/Plugins/Optimization/AMP/TestCase.php
Tabrisrp: done in https://github.com/wp-media/wp-rocket/pull/2459/commits/17739b5ec01a9c81159965e0ccbbc297f367d950

inc/ThirdParty/Plugins/Optimization/AMP.php
Tabrisrp: done in https://github.com/wp-media/wp-rocket/pull/2459/commits/e69dcc1dcbe18a6976e86fa40b35c8175b8c956c

inc/classes/ServiceProvider/class-third-party-subscribers.php
Tabrisrp: done in https://github.com/wp-media/wp-rocket/pull/2459/commits/218a81f02ceb1fce826e9c5c11519975f4174a47

inc/classes/ServiceProvider/class-third-party-subscribers.php
Tabrisrp: done in https://github.com/wp-media/wp-rocket/pull/2459/commits/218a81f02ceb1fce826e9c5c11519975f4174a47

crystinutzaa:
Thank you @Tabrisrp 💯 🥇

inc/ThirdParty/Plugins/Optimization/AMP.php
```diff
@@ -0,0 +1,116 @@
...
+		add_filter( 'do_rocket_lazyload', '__return_false' );
+		unset( $wp_filter['rocket_buffer'] );
+
+		if (
+			(bool) $this->options->get( 'do_cloudflare', 0 )
+			&& (
+				(bool) $this->options->get( 'cloudflare_protocol_rewrite', 0 )
+				// this filter is documented in inc/front/protocol.php.
+				|| (bool) apply_filters( 'do_rocket_protocol_rewrite', false ) // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
+				)
```
hellofromtonya: @Tabrisrp The new parentheses here changes the order of how this conditional is processed. 

Before the conditional expression was `true` _when_ either:

- Both options were enabled
- OR the filter returned `true`

Now it's `true` _when_:

- the `do_cloudflare` is enabled
- AND either:
    - the `cloudflare_protocol_rewrite` is enabled
    - OR the filter returned `true`

Do we want to change this behavior? Is this how it's supposed to behave, i.e. meaning we are fixing a potential side effect problem.

Tabrisrp: yes in fact the correct behavior is the new one, that was an old mistake that never got caught before.

The filter is supposed to be an alternative for the option only, not for both.

hellofromtonya: That makes sense. But wanted to make sure.

Cool -- we fixed a problem in the code through refactoring. Love it!

hellofromtonya:
@piotrbak I merged this PR before you were able to do the QA. Darn it. Please go ahead and do the QA on this one for us. If there are issues, then we'll open a new PR to fix them.

piotrbak:
Starting QA

hellofromtonya:
@crystinutzaa and @Tabrisrp See @piotrbak notes above. Let's discuss what needs to happen.

Tabrisrp:
@piotrbak Did you have the cache for WebP enabled when doing the testing? I suspect the AMP plugin of stripping the webp comment, causing the issue where the cache is regenerated on each request (like when a different plugin is applying HTML minification).

In Reader mode, I noticed that I have URLs with `/amp/` for posts, but `?amp` for pages. Is it the same for you?

piotrbak:
@Tabrisrp No, this specific one was disabled. However, you're right, the cache of `?amp` gets recreated each page load (that applies for `Transitional mode`) and it keeps happening with regular pages as soon as the rewrite rules are removed (the site is on Cloudflare and `$_SEVER['HTTPS']` is not set).

However, for `Reader mode`, the cache is not being generated at all. Please check this video:
https://www.loom.com/share/600545bd36fa413684cb8ab65dfcc21e

I'm going to test it in a different environment now.

piotrbak:
@Tabrisrp I'm not able to get the cached version in `Reader mode` when the query string is present also on my local website and on mega.

Tabrisrp:
When testing I was able to get the `@debug` in transitional mode when WebP cache is disabled.

For the `Reader mode`, I have URLs with `/amp/` for posts, but URLs with `?amp` for the rest.

@crystinutzaa We need to add the `amp` query strings to the list also for Reader mode to account for that

crystinutzaa:
@piotrbak & @Tabrisrp good catch. I missed the Reader Mode.
Made a new PR adding the Reader mode PR #2488

PR #2460, opened 2020-03-19
Closes #2399: Partial Preloading doesn't create a mobile file

### LINKED ISSUE #2399: Partial Preloading doesn't create a mobile file in 3.5
When updating a post, partial preloading will update the cached file.

**Expected behavior**

In 3.5, I'd expect partial preloading would create cached files for both desktop and mobile when the following options are enabled:
1. **Preload**
2. **Separate cache for mobile devices**

**To Reproduce**
1. Make sure the previously mentioned options are enabled.
2. Update a post.
3. Check the post's folder in `/cache/wp-rocket/`.
4. Only the desktop version will be there.

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

Screenfeed:
This allows to preload (also) the mobile version of the URLs after:
- clearing a post cache,
- clearing a term cache,
- clearing automatically expired cache files.

Closes #2399.

- [ ] QA tests

### DIFF
--- a/tests/Integration/inc/Engine/Preload/PartialPreloadSubscriber/maybeDispatch.php
+++ b/tests/Integration/inc/Engine/Preload/PartialPreloadSubscriber/maybeDispatch.php
@@ -62,9 +62,6 @@ class Test_MaybeDispatch extends TestCase {
 		do_action( 'after_rocket_clean_post', $post, $purge_urls, 'en' );
 		do_action( 'shutdown' );
 
-		//$this->subscriber->preload_after_clean_post( $post, $purge_urls, 'en' );
-		//$this->subscriber->maybe_dispatch();
-
 		$key = $this->process_wrapper->getGeneratedKey();
 
 		$this->assertNotNull( $key );
--- a/tests/Integration/inc/Engine/Preload/PartialPreloadSubscriber/maybeDispatch.php
+++ b/tests/Integration/inc/Engine/Preload/PartialPreloadSubscriber/maybeDispatch.php
@@ -1,0 +1,112 @@
+<?php
+
+namespace WP_Rocket\Tests\Integration\inc\Engine\Preload\PartialPreloadSubscriber;
+
+use WP_Rocket\Tests\Fixtures\inc\Engine\Preload\PartialProcess_Wrapper;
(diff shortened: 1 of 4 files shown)


tests/Integration/inc/Engine/Preload/PartialPreloadSubscriber/maybeDispatch.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+namespace WP_Rocket\Tests\Integration\inc\Engine\Preload\PartialPreloadSubscriber;
+
+use WP_Rocket\Tests\Fixtures\inc\Engine\Preload\PartialProcess_Wrapper;
+use WPMedia\PHPUnit\Integration\TestCase;
+
+/**
+ * @covers \WP_Rocket\Engine\Preload\PartialPreloadSubscriber::maybeDispatch
```
Tabrisrp:
```suggestion
 * @covers \WP_Rocket\Engine\Preload\PartialPreloadSubscriber::maybe_dispatch
```

Tabrisrp:

Partial preloading shoud now generate the mobile version of the cache when preload and mobile cache are enabled.

piotrbak:
Starting QA

hellofromtonya:
>However, it seems that we're clearing too many files when there are relations parent-child in categories. Moreover, updating the category description clears its cache but doesn't preload it. I'll investigate it asap.

@piotrbak 

- Did you discover what's happening here?
- Does this happen as a result of this PR, i.e. meaning it doesn't happen without this PR?

piotrbak:
@hellofromtonya 

> Did you discover what's happening here?
Not yet, when it comes to the categories cache clearing, I asked for advice in `l2` channel.

> Does this happen as a result of this PR, i.e. meaning it doesn't happen without this PR?
I was able to determine today that it's not related to this specific PR.

hellofromtonya:
>I was able to determine today that it's not related to this specific PR.

@piotrbak Thank you. I'll go ahead and merge this PR. 

If there's an issue related to categories cache clearing, then please open a new GH issue. Thanks.

PR #2468, opened 2020-03-23
Fix #2464 - CPCSS Allow <style> tag in fallback input field

### LINKED ISSUE #2464: [3.5] Fallback CPCSS doesn't save when <style> tags are present
In 3.4.4 you could paste fallback critical path CSS with opening and closing <style> tags. We removed those tags automatically. 

In 3.5.1 if you paste in CSS with the tags, nothing is saved. 

Sample CSS from the customer: snippi.com/raw/pdcnoum

**To Reproduce**
1. On 3.5.1 paste the above CSS into the Fallback CSS field
2. Save
3. The field is empty after saving
4. Compare same behaviour on 3.4.4 where the style tags are correctly removed and the remaining CSS is saved .

**Expected behavior**
The CSS should be stripped of the opening and closing <style> tags and the rest should be saved.

**Screenshots**
https://recordit.co/JNVcgAiROj

Customer can manually remove the style tags to successfully save the CSS in the meantime.

Ticket: https://secure.helpscout.net/conversation/1114940745/152804?folderId=377611

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
Closes #2468.

**Identify the root cause** ✅ 
The root cause is related to the order of the operations on this line:
https://github.com/wp-media/wp-rocket/blob/26e4ef22eb690577b5c46f180927af8e6e807742/inc/classes/admin/settings/class-settings.php#L315

**Scope a solution** ✅ 
The solution will be to first remove the <style> </style> tags and then to run wp_strip_all_tags:

`$input['critical_css'] = ! empty( $input['critical_css'] ) ? wp_strip_all_tags( str_replace( [ '<style>', '</style>' ], '', $input['critical_css'] ), [ "\'", '\"' ] ) : '';`

**Estimate the effort** ✅ 
Effort [XS]

- [x] Automated tests
- [x] QA tests

### DIFF
--- a/inc/classes/admin/settings/class-settings.php
+++ b/inc/classes/admin/settings/class-settings.php
@@ -312,7 +312,7 @@ class Settings {
 		$input['async_css'] = ! empty( $input['async_css'] ) ? 1 : 0;
 
 		// Option: Critical CSS.
-		$input['critical_css'] = ! empty( $input['critical_css'] ) ? str_replace( [ '<style>', '</style>' ], '', wp_strip_all_tags( $input['critical_css'], [ "\'", '\"' ] ) ) : '';
+		$input['critical_css'] = ! empty( $input['critical_css'] ) ? wp_strip_all_tags( str_replace( [ '<style>', '</style>' ], '', $input['critical_css'] ), [ "\'", '\"' ] ) : '';
 
 		// Database options.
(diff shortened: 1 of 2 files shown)


Tabrisrp:
@vmanthos @piotrbak This PR is ready for QA

Issue #2406, opened 2020-03-11
Get rid of the Remove Query Strings option


GeekPress:
## **What are we proposing to do?**

With this update we are hoping to remove the option from WP Rocket UI: Remove query strings from static resources (File Optimization tab)

How our current option looks like: 

![](https://productboard-files.s3.amazonaws.com/production/attachments/976768d27fe808b09e29ba8df10460fb88a290aa1bc3e6aa1e7a50991f6aabde/image.png)

## **Why are we doing this?**

Ressource: <https://sirv.com/help/resources/remove-query-strings-from-static-resources/>

Here is a GTmetrix test comparison: 

- With query strings: <https://gtmetrix.com/reports/mega.wp-rocket.me/lXacNRNS>
- Without query strings: <https://gtmetrix.com/reports/mega.wp-rocket.me/MCZq51yj>

The scores are exactly the same. GTmetrix no longer recommends this nor takes into account query strings for the grades. 

## **Does query strings affect performance?**

Here are the headers for a second visit to the same page: 

![](https://productboard-files.s3.amazonaws.com/production/attachments/8345278a0a61c0b69fe0b8ead06f81f402e98220cd880363c5737e491767df7b/image.png)

Modern browsers and proxies cache based on headers and not based on the presence of query strings. 

## **User expectations and experience. How will we do this feature?**
From version 3.6, all users will not have this option in WP Rocket UI. 

Both existing and new users will have the same experience. Query strings will come back. But as we know from our research, this isn't going to affect performance. 

Pros:

- Simpler approach
- Easier to implement and maintain

Cons: 

- Higher risk since we don't know what edge cases we are missing.

### Sub-Cards
- [x] [3rd party part](https://github.com/wp-media/wp-rocket/issues/2436)
- [x] [Admin misc part](https://github.com/wp-media/wp-rocket/issues/2433)
- [x] [Facebook & Google Tracking part](https://github.com/wp-media/wp-rocket/issues/2438)
- [x] [Front part](https://github.com/wp-media/wp-rocket/issues/2434)
- [x] [Simple Custom CSS part](https://github.com/wp-media/wp-rocket/issues/2437)
- [x] [Common part](https://github.com/wp-media/wp-rocket/issues/2435)
- [x] [Settings part](https://github.com/wp-media/wp-rocket/issues/2432)
- [x] [Option management part](https://github.com/wp-media/wp-rocket/issues/2431)

## **Risks**

[...]

PR #2469, opened 2020-03-23
Closes #2433 Remove Query Strings: admin misc changes

### LINKED ISSUE #2433: Remove Query Strings Option Removal: Misc Admin Part (2/8)
Sub-card for the epic #2406

This is for misc admin files.

### inc/admin/ui/notices.php
- [x] Deprecate the `rocket_warning_busting_cache_dir_permissions()` and remove the related `add_action`
### inc/classes/subscriber/admin/Deactivation/class-deactivation-intent-subscriber.php
- [ ] Write tests for `activate_safe_mode`
- [x] `activate_safe_mode` method: remove the `remove_query_strings` entry

crystinutzaa:
Sub-card for the epic #2406

This is for misc admin files.

### inc/admin/ui/notices.php
- [x] Deprecate the `rocket_warning_busting_cache_dir_permissions()` and remove the related `add_action`
### inc/classes/subscriber/admin/Deactivation/class-deactivation-intent-subscriber.php
- [x] Write tests for `activate_safe_mode`
- [x] Rename and move class to `inc/Engine/Admin/Deactivation/DeactivationIntent.php`
- [x] `activate_safe_mode` method: remove the `remove_query_strings` entry

### Live tests
- [ ] QA tests


GeekPress:
@Tabrisrp @crystinutzaa By deleting the notice about busting folder, what about our Google Tracking and Facebook Pixels add-on which are using this folder too? Do our add-on have customer notices?

crystinutzaa:
@GeekPress good question. However it seems that the notice was available only when Remove Query Strings was enabled: `&& ( get_rocket_option( 'remove_query_strings', false ) )`

@Tabrisrp should we move this notice in Busting Factory? Is there already a task for it?

Tabrisrp:
The notice was indeed only displayed if the remove query strings option was enabled. We have the choice to:
- Preserve the current behaviour, i.e. no warning when the FB/Google tracking add-ons are active
- Add notices to the Facebook/Google tracking subscribers as an enhancement of the current behaviour

If we decide to go with the second option, this can be done on the sub-task corresponding to those add-ons. It will need to be added as an additional task there.

GeekPress:
@Tabrisrp Thanks for the analyze.

Let's go with the second solution.

Why?
The « Remove Query Option » is activated by 70% of our users. 

When the « Google Tracking Add-on » is activated, 80% of these users also activated the « Remove Query String Option ».

In most of cases, we can expect these add-on working thanks to the « Remove Query String Option » and the warning we can have when this option is turned on in case of permission issues.

Tabrisrp:
👌 I added 2 new items on the to-do list for the FB & Google tracking sub-task to take care of this.

@crystinutzaa You can go ahead for this sub-task and keep as planned, other changes will be handled in another PR.

Tabrisrp:
@piotrbak This is ready for QA. Things to test:
- You should no longer have a notice about write permissions for the busting folder when it's not writable
- Activating safe mode should ignore the remove query strings option value now

inc/classes/ServiceProvider/class-admin-subscribers.php
```diff
@@ -41,7 +41,7 @@ public function register() {
 			->withArgument( $this->getContainer()->get( 'settings_page' ) );
 		$this->getContainer()->add( 'deactivation_intent_render', 'WP_Rocket\Admin\Deactivation\Render' )
 			->withArgument( $this->getContainer()->get( 'template_path' ) . '/deactivation-intent' );
-		$this->getContainer()->share( 'deactivation_intent_subscriber', 'WP_Rocket\Subscriber\Admin\Deactivation\Deactivation_Intent_Subscriber' )
+		$this->getContainer()->share( 'deactivation_intent_subscriber', 'WP_Rocket\Engine\Admin\Deactivation\DeactivationIntent' )
```
hellofromtonya: As we are touching this file too, let's move and rename it:

- file: `inc/Engine/Admin/ServiceProvider.php`
- class name: `ServiceProvider`.

tests/Unit/inc/Engine/Admin/Deactivation/DeactivationIntent/activateSafeMode.php
```diff
@@ -0,0 +1,69 @@
...
+	private $deactivation;
+	private $options;
+	private $options_api;
+
+	public function setUp() {
+		parent::setUp();
+
+		$this->options_api  = $this->createMock( 'WP_Rocket\Admin\Options' );
+		$this->options      = $this->createMock( 'WP_Rocket\Admin\Options_Data' );
+		$this->deactivation = new DeactivationIntent( $this->createMock( 'WP_Rocket\Admin\Deactivation\Render' ), $this->options_api, $this->options );
```
hellofromtonya: Replace `$this->createMock` with `Mockery::mock` and then update mock assertions.

tests/Unit/inc/Engine/Admin/Deactivation/DeactivationIntent/activateSafeMode.php
```diff
@@ -0,0 +1,69 @@
...
+
+		$this->options->expects( $this->once() )
+			->method( 'set_values' )
+			->with( $options );
+		$this->options->expects( $this->once() )
+			->method( 'get_options' )
+			->willReturn( $options );
+		$this->options_api->expects( $this->once() )
+			->method( 'set' )
+			->with( 'settings', $options );
```
hellofromtonya:
```suggestion
		$this->options->shouldReceive( 'set_values' )
			->once()
			->with( $options );
		$this->options->shouldReceive( 'get_options' )
			->once()
			->andReturn( $options );
		$this->options_api->shouldReceive( 'set' )
			->once()
			->with( 'settings', $options );
```

PR #2392, opened 2020-03-06
Fixes #2278 Compatibility with CDN URL containing a path

remyperona:
Closes #2278.

This issue has multiple factors going into play:
- This line is removing the `path` part of the CDN URL(s) added in the settings: https://github.com/wp-media/wp-rocket/blob/8cf329ca6f1e7ad9145c80499cfd4be98e702e3b/inc/classes/subscriber/CDN/CDNSubscriber.php#L145
- Fixing the above makes this conditional fail. It is checking the asset file `host` vs the known `hosts` values which include the CDN URL(s). So right now it won't correctly match: https://github.com/wp-media/wp-rocket/blob/8cf329ca6f1e7ad9145c80499cfd4be98e702e3b/inc/classes/optimization/class-abstract-optimization.php#L61
- The function `rocket_url_to_path()` needs to be updated too to keep things working: https://github.com/wp-media/wp-rocket/blob/8cf329ca6f1e7ad9145c80499cfd4be98e702e3b/inc/functions/formatting.php#L467

This means it impacts:
- Cache dynamic resources
- Remove Query Strings
- Minify/Combine CSS
- Minify/Combine JS

**Solution** ✅ 
Solving the issue means fixing the 3 points above, and also making sure there is no side-effects.

- Fixing first point: don't use the `host` only, but both `host` and `path`
- Fixing second point: instead of using the file `host` only, check if the there is a match between the hosts list and the full URL of the file
- Fixing third point: same approach as second point

### Progress
**Write tests to validate current state**
- [x] Unit/Integration Tests `WP_Rocket\Subscriber\CDN\CDNSubscriber::get_cdn_hosts`
- [x] Unit Test `WP_Rocket\Optimization\Cache_Dynamic_Resource::is_excluded_file`
- [x] Integration Test `WP_Rocket\Subscriber\Optimization\Cache_Dynamic_Resource_Subscriber::cache_dynamic_resource`
- [x] Unit Test `WP_Rocket\Optimization\Remove_Query_String::remove_query_strings_css`
- [x] Unit Test `WP_Rocket\Optimization\Remove_Query_String::remove_query_strings_js`
- [x] Integration Test `WP_Rocket\Subscriber\Optimization\Remove_Query_String_Subscriber::process`
- [x]  `Cristina`  Unit Test `WP_Rocket\Optimization\CSS\Minify::optimize`
- [x]  `Cristina`  Unit Test `WP_Rocket\Optimization\CSS\Combine::optimize`
[...]

PR #2470, opened 2020-03-23
Improves virtual filesystem for integration tests.

hellofromtonya:
This PR improves our virtual filesystem (vfs) for our integration tests to match Rocket's filesystem structure in the `wp-content` directory.

Changes include:

- During bootstrapping before the plugin launches, it sets up the `wp-content` virtual filesystem structure
- Defines the `WP_ROCKET_CACHE_ROOT_PATH` and `WP_ROCKET_CACHE_ROOT_URL` constants to point to the vfs
- Moves structure definitions to Fixtures.
- Updates all existing integration tests.
- Makes all vfs structures consistent.

Note: We no longer need to wrap constants `WP_ROCKET_CACHE_ROOT_PATH` and `WP_ROCKET_CACHE_ROOT_URL` in `rocket_get_constant()`. Why? With this PR, these constants are set to the virtual filesystem. That means we no longer need to mock these particular constants.


tests/ArrayTrait.php
```diff
@@ -0,0 +1,105 @@
+<?php
+
+namespace WP_Rocket\Tests;
+
+trait ArrayTrait {
```
hellofromtonya: @Tabrisrp This trait gives us some handy array functionality for working with the filesystem, config, etc. Do you recognize the intent of this code? It’s inspired by older versions of Laravel.

tests/Fixtures/inc/classes/Cache/Expired_Cache_Purge/purgeExpiredFiles.php
```diff
@@ -0,0 +1,109 @@
+<?php
+
+
+return [
+	'vfs_dir'   => 'cache/wp-rocket/',
+
+	// Virtual filesystem structure.
+	'structure' => [
```
hellofromtonya: Each test class adds the vfs structure in the Fixture file. Only configure what the test needs, but make sure it dits within the default wp-content structure.

tests/GlobTrait.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+namespace WP_Rocket\Tests;
+
+trait GlobTrait {
```
hellofromtonya: vfsStream doesn’t work with `glob`. This trait exists to encapsulate replacement functionality for `glob` within the tests.

tests/Integration/FilesystemTestCase.php
```diff
@@ -3,73 +3,182 @@
...
+
+		return $this->merged_structure;
+	}
+
+	/**
+	 * Gets the default virtual wp-content directory filesystem structure.
+	 *
+	 * @return array
+	 */
+	private function getDefaultVfs() {
```
hellofromtonya: Notice how this array matches the default `wp-content` directory structure. It provides consistency for our entire test suite while ensuring it matches what Rocket does within the real env.

tests/GlobTrait.php
```diff
@@ -0,0 +1,26 @@
...
+
+	public function deleteDomainCallback( $root ) {
+		$root = rtrim( $root, '*' );
+		$this->delete_files( $root );
```
Screenfeed: Shouldn't it be `deleteFiles()`?
```suggestion
		$this->deleteFiles( $root );
```

tests/GlobTrait.php
```diff
@@ -0,0 +1,26 @@
...
+
+	/**
+	 * Recursively deletes all the files in the given virtual directory.
+	 *
+	 * @param string $dir Virtual directory absolute path.
+	 */
+	protected function deleteFiles( $dir ) {
+		foreach ( $this->scandir( $dir ) as $item ) {
+			if ( $this->filesystem->is_dir( $item ) ) {
+				$this->delete_files( $item );
```
Screenfeed: And here too?
```suggestion
				$this->deleteFiles( $item );
```

hellofromtonya: Good catch @Screenfeed. Committed your suggestion https://github.com/wp-media/wp-rocket/pull/2470/commits/c5c8f18c1da98d08b1617ce23032dbaa2ffe87ec

tests/GlobTrait.php
hellofromtonya: Good catch @Screenfeed. Committed your suggestion https://github.com/wp-media/wp-rocket/pull/2470/commits/c5c8f18c1da98d08b1617ce23032dbaa2ffe87ec

tests/Fixtures/inc/classes/Cache/Expired_Cache_Purge/purgeExpiredFiles.php
Tabrisrp: I believe we will need to set the file structure from the root instead of `wp-content`. WP Rocket can read and write to files outside of `wp-content`, like `wp-config.php`, `.htaccess`, and read files in `wp-includes` for example

tests/GlobTrait.php
Tabrisrp: In #2392, I replaced the use of `glob` by SPL iterators instead, since `glob` doesn't work with virtual filesystem. It's also fixing a known issue on some hosts that don't have every flags of `glob` active. What do you think?

tests/Fixtures/inc/classes/Cache/Expired_Cache_Purge/purgeExpiredFiles.php
hellofromtonya: Yes, that's a very good point. I'll update it.

tests/GlobTrait.php
hellofromtonya: @Tabrisrp will PR 2392 replace all instances of `glob`?

Tabrisrp: only one, so we will still need the trait for now 👍

hellofromtonya: @Tabrisrp I moved the trait to `wp-media/phpunit`. As we discussed, once we replace `glob()`, then this trait can be deleted and the code in the test removed.

tests/Fixtures/inc/classes/Cache/Expired_Cache_Purge/purgeExpiredFiles.php
hellofromtonya: Done in https://github.com/wp-media/wp-rocket/pull/2470/commits/da1f9c888de5d65eac93b57127f77df12cdd6410

hellofromtonya:
@Tabrisrp I moved the vfs test case set up and `ArrayTrait` to [`wp-media/phpunit`](https://github.com/wp-media/phpunit/pull/16). We need to merge our PHPUnit PR 16 first and do a new release. Then Travis will pass here.

licence-data.php
```diff
@@ -1,18 +0,0 @@
-<?php
```
Tabrisrp: This file should not be removed

hellofromtonya: Hmm, this is going to be a continuous problem with our commits. It gets deleted when live testing.

Added back in https://github.com/wp-media/wp-rocket/pull/2470/commits/338decfad8bc2418078c750b2eace051f9da9676

tests/Unit/inc/classes/subscriber/CDN/RocketCDN/AdminPageSubscriber/displayRocketcdnStatus.php
```diff
@@ -3,137 +3,167 @@
...
+//		<div>
+//			<span class="wpr-title3">Next Billing Date</span>
+//			<span class="wpr-infoAccount wpr-isValid">2020-01-01</span>
+//		</div>
+//	</div>
+//</div>
+//HTML;
+//
+//		$this->assertSame( $this->format_the_html( $expected ), $this->getActualHtml() );
+//	}
```
Tabrisrp: should this be preserved?

hellofromtonya: Whoopsie.... Fixed here https://github.com/wp-media/wp-rocket/pull/2470/commits/8209a02ecb018d7fd1a1b6d5688cbd1fd3efd88a.

PR #2474, opened 2020-03-25
Closes #2434: Remove Query Strings Option Removal, Front part (4/8)

### LINKED ISSUE #2434: Remove Query Strings Option Removal: Front part (4/8)
Sub-card for the epic #2406

This is for files managing the remove query strings option itself.

#### inc/classes/optimization/class-remove-query-string.php
- [x] Remove file
#### inc/classes/subscriber/Optimization/class-remove-query-string-subscriber.php
- [x] Remove file
#### inc/classes/class-plugin.php
- [x] `load()` method: remove `remove_query_string_subscriber` entry
#### inc/classes/ServiceProvider/class-optimization-subscribers.php
- [x] Remove references to the remove query strings classes

Screenfeed:
Sub-card for the epic #2406.

This is for files managing the remove query strings option itself.

#### inc/classes/optimization/class-remove-query-string.php
- [x] Remove file
#### inc/classes/subscriber/Optimization/class-remove-query-string-subscriber.php
- [x] Remove file
#### inc/classes/class-plugin.php
- [x] `load()` method: remove `remove_query_string_subscriber` entry
#### inc/classes/ServiceProvider/class-optimization-subscribers.php
- [x] Remove references to the remove query strings classes

Closes #2434.


Screenfeed:
I'm not sure. We don't do this for the functions so far, but we may have good reasons to do it for classes.
_(thinking as I write, throwing ideas...)_
- What we do so far is to split the deprecated code by versions, maybe we could do the same for the classes, like a `inc/deprecated/3.6/classes/` folder containing classes and (maybe) functions. Which would still allow us to create sub-folders if we want.
- Without sub-folders, at some point we may have a lot of classes in the `classes` folder, like we had a lot of functions in the `deprecated.php` file, so it might be a good idea to have sub-folders, whatever system we use. We also may have filename conflicts.
- On the other hand, splitting like `inc/deprecated/3.6/classes/` will make looking for a specific class a bit more difficult (we must know in which "version folder" it is to find it), but nothing that a file search can solve.
- We could also do something like this to make it easier to find: `inc/deprecated/3.6/classes/Subscriber_Optimization_remove-query-string-subscriber.php` (but it's ugly as Hell).

How about this?
```
inc/deprecated/3.6/classes/subscriber/Optimization/class-remove-query-string-subscriber.php
inc/deprecated/3.6/functions.php
```

Ping @Tabrisrp @hellofromtonya.

Tabrisrp:
I don't think it's necessary to have the version number for which the class has been deprecated in the filestructure. We have this information in the class itself.

We split the deprecated functions into files because the main one was becoming too big and too hard to read. This is not really something we will face with classes since they stay in their own files.

I think we don't need to preserve the `classes` inside the filestructure either, but preserving the rest can be fine.

inc/deprecated/classes/class-remove-query-string-subscriber.php
```diff
@@ -5,12 +5,16 @@
...
+ * Hooks into WordPress to remove query strings for static files.
  *
- * @since 3.1
+ * @since  3.1
+ * @since  3.6 Deprecated.
  * @author Remy Perona
+ * @deprecated
  */
 class Remove_Query_String_Subscriber extends Minify_Subscriber {
+	use \WP_Rocket\Traits\DeprecatedClass;
```
Tabrisrp: You can import the trait before, and then `use DeprecatedClass;`

inc/classes/traits/trait-deprecated-class.php
```diff
@@ -0,0 +1,111 @@
+<?php
```
hellofromtonya: @Tabrisrp `classes` will eventually go away. This trait should likely go into the `inc/deprecated` folder. What do you think?

Tabrisrp: Good question, in a way it makes sense to have it in the `deprecated` folder since it's only used for deprecated classes. We will have to remember that the trait itself is not deprecated though. 

If it goes into the `inc/deprecated` folder, it will need to be autoloaded by the PSR-4 autoloader, so the filepath/filename needs to be changed to match the namespace/class name.

hellofromtonya: @Screenfeed Let's do this:

- move this trait to `inc/deprecated`
- rename the file and trait to `DeprecatedClassTrait`
- change it's namespace to `WP_Rocket\deprecated`
- and then update where it's used.

Thanks!

Screenfeed: Done in ade1bb7.

inc/deprecated/DeprecatedClassTrait.php
```diff
@@ -0,0 +1,111 @@
...
+			 * Without replacement.
+			 */
+			call_user_func(
+				'trigger_error',
+				sprintf(
+					/* translators: 1: PHP class name, 2: version number. */
+					__( 'The called class %1$s is <strong>deprecated</strong> since version %2$s!', 'rocket' ),
+					'<code>' . self::class . '</code>',
+					'<strong>' . $version . '</strong>'
+				)
```
hellofromtonya, line 78: There's redundancy here. Instead, create the message. Then do `call_user_func( 'trigger_error', $message );`.

inc/deprecated/DeprecatedClassTrait.php
```diff
@@ -0,0 +1,111 @@
...
+		/**
+		 * Without replacement.
+		 */
+		call_user_func(
+			'trigger_error',
+			sprintf(
+				'The called class %1$s is <strong>deprecated</strong> since version %2$s!',
+				'<code>' . self::class . '</code>',
+				'<strong>' . $version . '</strong>'
+			)
```
hellofromtonya, line 108: Same here. Let's reduce the redundancy.

inc/deprecated/DeprecatedClassTrait.php
```diff
@@ -0,0 +1,111 @@
...
+
+			/**
+			 * Without replacement.
+			 */
+			call_user_func(
+				'trigger_error',
+				sprintf(
+					/* translators: 1: PHP class name, 2: version number. */
+					__( 'The called class %1$s is <strong>deprecated</strong> since version %2$s!', 'rocket' ),
+					'<code>' . self::class . '</code>',
```
hellofromtonya, line 76: Change `self::class` to `static::class` for late binding usage.

inc/deprecated/DeprecatedClassTrait.php
```diff
@@ -0,0 +1,111 @@
...
+			if ( ! empty( $replacement ) ) {
+				/**
+				 * With replacement.
+				 */
+				call_user_func(
+					'trigger_error',
+					sprintf(
+						/* translators: 1: PHP class name, 2: version number, 3: replacement class name. */
+						__( 'The called class %1$s is <strong>deprecated</strong> since version %2$s! Use %3$s instead.', 'rocket' ),
+						'<code>' . self::class . '</code>',
```
hellofromtonya, line 60: Change `self::class` to `static::class` for late binding usage.

inc/deprecated/DeprecatedClassTrait.php
```diff
@@ -0,0 +1,111 @@
...
+
+		if ( ! empty( $replacement ) ) {
+			/**
+			 * With replacement.
+			 */
+			call_user_func(
+				'trigger_error',
+				sprintf(
+					'The called class %1$s is <strong>deprecated</strong> since version %2$s! Use %3$s instead.',
+					'<code>' . self::class . '</code>',
```
hellofromtonya, line 91: Change `self::class` to `static::class` for late binding usage.

inc/deprecated/DeprecatedClassTrait.php
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		/**
+		 * Without replacement.
+		 */
+		call_user_func(
+			'trigger_error',
+			sprintf(
+				'The called class %1$s is <strong>deprecated</strong> since version %2$s!',
+				'<code>' . self::class . '</code>',
```
hellofromtonya, line 106: Change `self::class` to `static::class` for late binding usage.

inc/deprecated/DeprecatedClassTrait.php
Screenfeed: That's the reason I used `self` instead of `static` actually: the class being deprecated is the one using this trait, not necessarily the one extending it imho. And if the class extending the deprecated one must also be deprecated, then use this trait in it. In this case we have the choice.
Or, we could also say that a class that extends a deprecated class is automatically considered deprecated, which I can totally understand.
What do you think @hellofromtonya @crystinutzaa @Tabrisrp?

Tabrisrp: That's an interesting question. IMO, any class extending a deprecated class should be considered deprecated too, or updated to not extend it (extend another one).

inc/Optimization/ServiceProvider.php
```diff
@@ -56,12 +56,8 @@ public function register() {
 			->withArgument( $this->getContainer()->get( 'buffer_optimization' ) );
 		$this->getContainer()->add( 'cache_dynamic_resource', 'WP_Rocket\Optimization\Cache_Dynamic_Resource' )
 			->withArgument( $this->getContainer()->get( 'options' ) )
```
Tabrisrp, line 58: We can do `$options = $this->getContainer()->get( 'options' )` once, and then use `$options` everywhere for a little perf improvement

inc/deprecated/DeprecatedClassTrait.php
Screenfeed: >Change `self::class` to `static::class` for late binding usage.

>any class extending a deprecated class should be considered deprecated too

OK, let's go with this then? @crystinutzaa Anything against it?

crystinutzaa: Nope. I love the idea that any class which extends the deprecated class will is automatically considered deprecated.

Screenfeed: OK then. Thank you all :)

Screenfeed: Done in a37f1d6.

inc/deprecated/DeprecatedClassTrait.php
Screenfeed: Done in a37f1d6.

inc/deprecated/DeprecatedClassTrait.php
Screenfeed: Done in a37f1d6.

Screenfeed:
@hellofromtonya Ready for review.

inc/deprecated/3.6.php
```diff
@@ -1,9 +1,13 @@
...
 <?php
 // phpcs:ignoreFile
 
 defined( 'ABSPATH' ) || exit;
 
+/**
+ * Class aliases.
+ */
+class_alias( '\WP_Rocket\Engine\Optimization\ServiceProvider', '\WP_Rocket\ServiceProvider\Optimization_Subscribers' );
+
```
hellofromtonya, line 10: We also need to load the deprecated the class files here too. 

Why? Composer won't autoload them. Why? Their namespace will not match their new `inc/deprecated/` filesystem location (and we don't want to change their namespace).

Screenfeed: @hellofromtonya Having these classes in [composer’s classmap](https://github.com/wp-media/wp-rocket/pull/2474/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R63) seems to be enough, they are autoloaded:
```
	"autoload": {
		"classmap": [
			"inc/classes",
			"inc/deprecated",
			"inc/vendors/classes"
		],
```

hellofromtonya: Let's do don't add it to Composer's autoload. Why? In the future, we want to add a way to disable loading deprecated files, i.e. to lower memory consumption when a website doesn't need the older stuff.

Screenfeed: Done in https://github.com/wp-media/wp-rocket/pull/2474/commits/ce06cb0a809f7d52c4ede8ca27a6040666319801.

inc/deprecated/3.6.php
```diff
@@ -1,9 +1,20 @@
...
 // phpcs:ignoreFile
 
 defined( 'ABSPATH' ) || exit;
 
+/**
+ * Require deprecated classes.
+ */
+require WP_ROCKET_DEPRECATED_PATH . 'DeprecatedClassTrait.php';
+require WP_ROCKET_DEPRECATED_PATH . 'Engine/Optimization/QueryString/Remove.php';
+require WP_ROCKET_DEPRECATED_PATH . 'Engine/Optimization/QueryString/RemoveSubscriber.php';
```
hellofromtonya, line 11: 1. Let's use `require_once` to avoid fatal errors.
2. Use `__DIR__` instead of the constant. Why? The files are in this folder.

Screenfeed: Done in https://github.com/wp-media/wp-rocket/pull/2474/commits/8840398e6c675a49784f63380ad98dfc12bfe571.

hellofromtonya:
@arunbasillal @piotrbak @vmanthos This PR is ready for QA.

arunbasillal:
Starting QA

Screenfeed:
@arunbasillal This part doesn't affect the UI indeed.

PR #2476, opened 2020-03-25
Update cleanup in uninstall.php

### LINKED ISSUE #2352: On plugin deletion, cleanup is incomplete
When you delete WP Rocket from the plugins screen in WordPress, it should completely cleanup after itself. 
However, these 2 folders are not deleted: 
- `cache/wp-rocket` (the cache files are removed but the folder itself and the htaccess within it are not)
- `cache/critical-css` - folder and the css files remain after the plugin has been deleted

Additionally it seems some database entries are left. After deleting the plugin and then re-installing, I saw this notice:
WP_Rocket_‹_WordPress_Tutorials_—_WordPress_-_Mozilla_Firefox
I would not expect that to happen on a fresh install. 

How to reproduce: 

- Install WP Rocket
- Enable Optimize CSS Delivery
- Currently you'll likely have to click "regenerate cpcss" to create the CSS files
- Make sure some CSS files are created
- Delete WP Rocket from the plugins screen

I haven't done a more thorough test to see if other files are leftover (I think /min/ is correctly deleted), I just discovered these while testing the other CPCSS generation issue :)
I'm able to reproduce this on 2 of my own sites, on different servers.

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

Tabrisrp:
Closes #2352 

- [x] Update existing cleanup to use arrays and `foreach` loops to perform the deletion instead of calling the same functions multiple times
- [x] Add missing options/transients/events to the arrays
- [x] Replace use of `glob` by `scandir` instead
- [x] Delete `cache/critical-css`
- [x] Create new class to handle uninstall
- [x] Instantiate new class in `uninstall.php` and call `uninstall()` method
- [x] Write tests


inc/classes/wp-rocket-uninstall.php
```diff
@@ -0,0 +1,212 @@
...
+	 *
+	 * @since 3.5.2
+	 * @author Remy Perona
+	 *
+	 * @return void
+	 */
+	private function delete_plugin_data() {
+		foreach ( $this->transients as $transient ) {
+			delete_transient( $transient );
+		}
```
Screenfeed: What about using `array_map()`?
```suggestion
		array_map( 'delete_transient', $this->transients );
```

inc/classes/wp-rocket-uninstall.php
```diff
@@ -0,0 +1,212 @@
...
+
+		delete_site_transient( 'wp_rocket_update_data' );
+
+		foreach ( $this->options as $option ) {
+			delete_option( $option );
+		}
```
Screenfeed:
```suggestion
		array_map( 'delete_option', $this->options );
```

inc/classes/wp-rocket-uninstall.php
```diff
@@ -0,0 +1,212 @@
...
+
+		// Delete all user meta related to WP Rocket.
+		delete_metadata( 'user', '', 'rocket_boxes', '', true );
+
+		foreach ( $this->events as $event ) {
+			wp_clear_scheduled_hook( $event );
+		}
```
Screenfeed:
```suggestion
		array_map( 'wp_clear_scheduled_hook', $this->events );
```

inc/classes/wp-rocket-uninstall.php
```diff
@@ -0,0 +1,212 @@
...
+	private function delete( $file ) {
+		if ( ! is_dir( $file ) ) {
+			@unlink( $file );
+			return;
+		}
+
+		$items = @scandir( $file );
+
+		if ( ! $items ) {
+			return;
```
Screenfeed: This would leave an empty folder.
```suggestion
			@rmdir( $file );
			return;
```

tests/Integration/inc/classes/WP_Rocket_Uninstall/uninstall.php
```diff
@@ -0,0 +1,153 @@
...
+			delete_option( $option);
+		}
+
+		foreach ( $this->transients as $transient ) {
+			delete_transient( $transient);
+		}
+
+		foreach( $this->events as $event ) {
+			wp_clear_scheduled_hook( $event );
+		}
```
Screenfeed:
```suggestion
		array_map( 'delete_option', $option );
		array_map( 'delete_transient', $transient );
		array_map( 'wp_clear_scheduled_hook', $event );
```

Tabrisrp:
Yes I was wondering when writing that, `foreach` or `array_map`? Not sure if there is a definite better answer there seeing how small the loops are.

@hellofromtonya do you see one or the other being clearly better, in terms of performance and readibility?

hellofromtonya:
>Yes I was wondering when writing that, foreach or array_map? Not sure if there is a definite better answer there seeing how small the loops are.
>
>@hellofromtonya do you see one or the other being clearly better, in terms of performance and readibility?

Don't use `array_map`. Why? It walks through the even array, but then returns an array. That return adds extra processing and memory.

[`array_walk`](https://www.php.net/manual/en/function.array-walk.php) provides a one-liner approach. It's similar to `foreach`, where it walks through each element in the array. It's similar to `array_map` in that each element is passed to a callback. But unlike `array_map`, it does not return an array.

Great question BTW!

inc/classes/wp-rocket-uninstall.php
```diff
@@ -0,0 +1,208 @@
...
+	 *
+	 * @param string $file Path to file or directory.
+	 */
+	private function delete( $file ) {
+		if ( ! is_dir( $file ) ) {
+			@unlink( $file );
+			return;
+		}
+
+		$items = @scandir( $file );
```
hellofromtonya: @Tabrisrp Why `scandir()` instead of SPL iterators? I know I originally had this code in the vfs. But I changed it to this https://github.com/wp-media/phpunit/blob/master/VirtualFilesystemDirect.php#L308

Tabrisrp: I copied what you used before in fact, that's why.

I updated the code to use the SPL iterators instead

inc/classes/wp-rocket-uninstall.php
```diff
@@ -0,0 +1,195 @@
+<?php
```
hellofromtonya: Move this into `Engine` as `classes` will eventually be gone.

inc/classes/wp-rocket-uninstall.php
```diff
@@ -0,0 +1,195 @@
...
+				new RecursiveDirectoryIterator( $file, FilesystemIterator::SKIP_DOTS ),
+				RecursiveIteratorIterator::CHILD_FIRST
+			);
+		} catch ( UnexpectedValueException $e ) {
+			return;
+		}
+
+		foreach ( $iterator as $item ) {
+			if ( is_dir( $item ) ) {
+				@rmdir( $item );
```
hellofromtonya: Add phpcs rule exception as we want to suppress here.

inc/classes/wp-rocket-uninstall.php
```diff
@@ -0,0 +1,195 @@
...
+
+				continue;
+			}
+
+			@unlink( $item );
```
hellofromtonya: Add phpcs rule exception as we want to suppress here.

inc/classes/wp-rocket-uninstall.php
```diff
@@ -0,0 +1,195 @@
...
+		}
+
+		@rmdir( $file );
```
hellofromtonya: Add phpcs rule exception as we want to suppress here.

inc/classes/wp-rocket-uninstall.php
```diff
@@ -0,0 +1,195 @@
...
+			$iterator = new RecursiveIteratorIterator(
+				new RecursiveDirectoryIterator( $file, FilesystemIterator::SKIP_DOTS ),
+				RecursiveIteratorIterator::CHILD_FIRST
+			);
+		} catch ( UnexpectedValueException $e ) {
+			return;
+		}
+
+		foreach ( $iterator as $item ) {
+			if ( is_dir( $item ) ) {
```
hellofromtonya: Use `$item->isDir()` instead.

tests/Integration/inc/classes/WP_Rocket_Uninstall/uninstall.php
```diff
@@ -0,0 +1,113 @@
...
+
+		$this->uninstall = new WP_Rocket_Uninstall( $this->filesystem->getUrl( 'wp-content/cache'), $this->filesystem->getUrl( 'wp-content/wp-rocket-config' ) );
+	}
+
+	public function tearDown() {
+		array_walk( $this->options, 'delete_option' );
+		array_walk( $this->transients, 'delete_transient' );
+
+		foreach ( $this->events as $event ) {
```
hellofromtonya: Why not `array_walk` for consistency?

inc/classes/wp-rocket-uninstall.php
```diff
@@ -0,0 +1,195 @@
...
+	 * Recursively deletes files and directories
+	 *
+	 * @since 3.5.2
+	 * @author Remy Perona
+	 *
+	 * @param string $file Path to file or directory.
+	 */
+	private function delete( $file ) {
+		if ( ! is_dir( $file ) ) {
+			@unlink( $file );
```
hellofromtonya: Add phpcs rule exception as we want to suppress here.

inc/classes/wp-rocket-uninstall.php
```diff
@@ -0,0 +1,195 @@
...
+	private function delete_plugin_data() {
+		delete_site_transient( 'wp_rocket_update_data' );
+
+		// Delete all user meta related to WP Rocket.
+		delete_metadata( 'user', '', 'rocket_boxes', '', true );
+
+		array_walk( $this->transients, 'delete_transient' );
+		array_walk( $this->options, 'delete_option' );
+
+		foreach ( $this->events as $event ) {
```
hellofromtonya: Why not `array_walk` instead for consistency?

Tabrisrp: `array_walk` was generating an error with `wp_clear_schedule_hook` because the callback for `array_walk` receives 2 parameters: the value, and the key

`wp_clear_scheduled_hook was called with an argument that is <strong>deprecated</strong> since version 3.0.0! This argument has changed to an array to match the behavior of the other cron functions.`

inc/classes/wp-rocket-uninstall.php
Tabrisrp: Fixed https://github.com/wp-media/wp-rocket/pull/2476/commits/f5180d716d294f8d4122c6486ae28621546262f5

inc/classes/wp-rocket-uninstall.php
Tabrisrp: Fixed https://github.com/wp-media/wp-rocket/pull/2476/commits/f5180d716d294f8d4122c6486ae28621546262f5

inc/classes/wp-rocket-uninstall.php
Tabrisrp: Fixed https://github.com/wp-media/wp-rocket/pull/2476/commits/f5180d716d294f8d4122c6486ae28621546262f5

inc/classes/wp-rocket-uninstall.php
Tabrisrp: Fixed https://github.com/wp-media/wp-rocket/pull/2476/commits/f5180d716d294f8d4122c6486ae28621546262f5

inc/classes/wp-rocket-uninstall.php
Tabrisrp: Fixed https://github.com/wp-media/wp-rocket/pull/2476/commits/f5180d716d294f8d4122c6486ae28621546262f5

inc/classes/wp-rocket-uninstall.php
Tabrisrp: Fixed https://github.com/wp-media/wp-rocket/pull/2476/commits/f5180d716d294f8d4122c6486ae28621546262f5

hellofromtonya:
@piotrbak This PR is ready for QA.

piotrbak:
Starting QA

PR #2481, opened 2020-03-29
Optimizes, guards, and adds tests for get_rocket_post_dates_urls()

hellofromtonya:
This PR does the following:

- Adds integration and unit tests _before_ the function was optimized, i.e. safety net.
- Optimizes `get_rocket_post_dates_urls` by:
    - Running the year and month link getters 1x instead of 3x each
    - Returns the filtered urls instead of assigning and then returning the assignment.
- Guards from when `get_the_time()` returns `false`
- Ensures the function always returns an array by type casting what is returned from the filter. Why? We can't trust that callbacks will return an array.

## TODO

- [x] Unit tests
- [x] Integration tests
- [x] QA


crystinutzaa (approved):

remyperona (approved):

hellofromtonya:
This one is ready for QA @arunbasillal @piotrbak @vmanthos

arunbasillal:
Starting QA

arunbasillal:
- Compared the output of `get_rocket_post_dates_urls` with the current master branch. 
- Tested by Updating a post, editing a post cache and preloading. 

All works as it is right now 👍

arunbasillal (approved):

PR #2483, opened 2020-03-29
Tests: Adds tests for rocket_is_plugin_active() and rocket_is_plugin_active_for_network()

hellofromtonya:
Adds unit and integration tests for both `rocket_is_plugin_active()` and `rocket_is_plugin_active_for_network()`.

Additional changes:

- Adds `@return` to the functions' DocBlocks
- Minor micro-optimizations
- Minor reformatting
- Excludes the `Multisite` group from the `test-integration` script

### DIFF
--- a/inc/functions/options.php
+++ b/inc/functions/options.php
@@ -61,6 +61,8 @@ function rocket_is_plugin_active( $plugin ) {
  * @source wp-admin/includes/plugin.php
  *
  * @param string $plugin Plugin folder/main file.
+ *
+ * @return bool true if multisite and plugin is active for network; else, false.
  */
 function rocket_is_plugin_active_for_network( $plugin ) {
 	if ( ! is_multisite() ) {
--- a/inc/functions/options.php
+++ b/inc/functions/options.php
@@ -48,9 +48,15 @@ function update_rocket_option( $key, $value ) { // phpcs:ignore WordPress.Naming
  * @source wp-admin/includes/plugin.php
  *
  * @param string $plugin Plugin folder/main file.
+ *
+ * @return boolean true when plugin is active; else false.
  */
 function rocket_is_plugin_active( $plugin ) {
-	return in_array( $plugin, (array) get_option( 'active_plugins', [] ), true ) || rocket_is_plugin_active_for_network( $plugin );
+	return (
+		in_array( $plugin, (array) get_option( 'active_plugins', [] ), true )
+		||
+		rocket_is_plugin_active_for_network( $plugin )
+	);
 }
 
 /**
--- a/inc/functions/options.php
+++ b/inc/functions/options.php
@@ -76,7 +76,7 @@ function rocket_is_plugin_active_for_network( $plugin ) {
 	}
 
 	$plugins = get_site_option( 'active_sitewide_plugins' );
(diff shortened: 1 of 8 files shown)


inc/functions/options.php
```diff
@@ -61,18 +67,16 @@ function rocket_is_plugin_active( $plugin ) {
...
 		return false;
 	}
 
 	$plugins = get_site_option( 'active_sitewide_plugins' );
-	if ( isset( $plugins[ $plugin ] ) ) {
-		return true;
-	}
-
-	return false;
+	return ( isset( $plugins[ $plugin ] ) );
```
Tabrisrp: are the wrapping parenthesis required?

inc/functions/options.php
```diff
@@ -48,9 +48,15 @@ function update_rocket_option( $key, $value ) { // phpcs:ignore WordPress.Naming
...
+ *
+ * @return boolean true when plugin is active; else false.
  */
 function rocket_is_plugin_active( $plugin ) {
-	return in_array( $plugin, (array) get_option( 'active_plugins', [] ), true ) || rocket_is_plugin_active_for_network( $plugin );
+	return (
+		in_array( $plugin, (array) get_option( 'active_plugins', [] ), true )
+		||
+		rocket_is_plugin_active_for_network( $plugin )
+	);
```
Tabrisrp: are the wrapping parenthesis required?

inc/functions/options.php
hellofromtonya: They are not. I have a habit of adding them to make it clear that it's a conditional expression. But definitely not required.

What do you think @Tabrisrp? Do you find them unnecessary?

inc/functions/options.php
hellofromtonya: Same answer as above. We are returning the result of a conditional expression. 

They are unnecessary. I use the `( )` to show "conditional expression" intent and a boolean will be returned.

inc/functions/options.php
Tabrisrp: I understand, we have to be sure to be consistent with this though if we start doing that. We need to agree as a team with that practice and everyone will be doing it.

hellofromtonya: This one I think is too much. I removed it https://github.com/wp-media/wp-rocket/pull/2483/commits/befc799b10763ff51e6d5c80e2f0839aa8896434

inc/functions/options.php
hellofromtonya: Let's discuss our preferred standard for returning the result of a conditional expression. This code is a great example.

This code is a conditional expression. It returns the result of `in_array()` OR `rocket_is_plugin_active_for_network()`.

```php
return (
		in_array( $plugin, (array) get_option( 'active_plugins', [] ), true )
		||
		rocket_is_plugin_active_for_network( $plugin )
	);
```

For me, I find this formatting to be very clear and understandable.  Why?

- By wrapping it in `( )`, it tells us it's a conditional expression. 
- It looks just like an `if` conditional, i.e. making it easily recognizable and understandable.
- With the conditional operator on its own line, we can quickly see that it's an OR expression.

@Tabrisrp and @crystinutzaa What do you think? Do you like or not like this formatting and design pattern as part of our coding standards? Let's discuss.

PR #2484, opened 2020-03-29
Adds unit tests for rocket_has_i18n().

hellofromtonya:
Adds unit tests for `rocket_has_i18n()`.

No source code changes. No QA is needed.

### DIFF
--- a/tests/Unit/inc/Addon/Varnish/Subscriber/cleanHome.php
+++ b/tests/Unit/inc/Addon/Varnish/Subscriber/cleanHome.php
@@ -16,6 +16,12 @@ use WP_Rocket\Addon\Varnish\Varnish;
  */
 class Test_CleanHome extends TestCase {
 
+	public static function setUpBeforeClass() {
+		parent::setUpBeforeClass();
+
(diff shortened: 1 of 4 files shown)


tests/Unit/inc/Addon/Varnish/Subscriber/cleanHome.php
```diff
@@ -16,6 +16,12 @@
  */
 class Test_CleanHome extends TestCase {
 
+	public static function setUpBeforeClass() {
+		parent::setUpBeforeClass();
+
+		require_once WP_ROCKET_PLUGIN_ROOT . 'inc/functions/i18n.php';
```
crystinutzaa: @hellofromtonya may I ask you why you added require_once in here? Will this be used in this Unit test when we will revisit this code?

hellofromtonya: @crystinutzaa Multiple test class files are loading this same file. With `require_once`, PHP checks if the file has already been loaded into memory. If yes, it doesn't load it again.

PR #2485, opened 2020-03-29
Tests: Adds tests for get_rocket_sample_permalink()

hellofromtonya:
Adds unit and integration tests for `get_rocket_sample_permalink()`.

QA is not needed for this as source code was not changed. Tests only.

### DIFF
--- a/tests/Unit/inc/functions/getRocketSamplePermalink.php
+++ b/tests/Unit/inc/functions/getRocketSamplePermalink.php
@@ -1,0 +1,113 @@
+<?php
+
+namespace WP_Rocket\Tests\Unit\inc\functions;
+
+use Brain\Monkey\Functions;
+use WPMedia\PHPUnit\Unit\TestCase;
+
+/**
+ * @covers ::get_rocket_sample_permalink
+ * @group Functions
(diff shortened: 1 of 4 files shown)


tests/Unit/inc/functions/getRocketSamplePermalink.php
```diff
@@ -0,0 +1,113 @@
...
+	public static function setUpBeforeClass() {
+		parent::setUpBeforeClass();
+
+		require_once WP_ROCKET_PLUGIN_ROOT . 'inc/functions/posts.php';
+	}
+
+	public function tearDown() {
+		parent::tearDown();
+
+		remove_filter( 'editable_slug', [ $this, 'editable_slug_cb' ] );
```
Tabrisrp: Stray line from copy/paste I think :)

hellofromtonya: You're right. Removed the `tearDown()` method here https://github.com/wp-media/wp-rocket/pull/2485/commits/23fcefe8c3193109574aa7cfd8a47f28d876ca49

PR #2488, opened 2020-03-30
Fix #2216 - Add Reader mode in query strings to provide compatibility with AMP

### LINKED ISSUE #2216: Update AMP compatibility with ?amp

crystinutzaa:
AMP plugin compatibility with Reader Mode.

It seems that in Reader Mode  I have URLs with /amp/ for posts, but URLs with ?amp for the rest.

### DIFF
--- a/inc/ThirdParty/Plugins/Optimization/AMP.php
+++ b/inc/ThirdParty/Plugins/Optimization/AMP.php
@@ -57,7 +57,7 @@ class AMP implements Subscriber_Interface {
 	 * @since  3.5.2
 	 * @author Soponar Cristina
 	 */
-	public function generate_config_file() {
+	public function generate_config_file( $val ) {
 		rocket_generate_config_file();
 	}
 
--- a/inc/ThirdParty/Plugins/Optimization/AMP.php
+++ b/inc/ThirdParty/Plugins/Optimization/AMP.php
@@ -57,7 +57,7 @@ class AMP implements Subscriber_Interface {
 	 * @since  3.5.2
 	 * @author Soponar Cristina
 	 */
-	public function generate_config_file( $val ) {
+	public function generate_config_file() {
 		rocket_generate_config_file();
 	}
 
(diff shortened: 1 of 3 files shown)


hellofromtonya:
This PR is ready for QA. @piotrbak can you take this one? It addresses the problem you found.

hellofromtonya:
>however, I think we should clear (preload?) cache when switching between Reader/Transitional mode, since the layout changes then.
>
>We're serving outdated AMP cache after the switch.

@arunbasillal 👆 What do you think?

If yes, then we'll need to open a new issue/product feature and PR to address this.

arunbasillal:
@hellofromtonya and @piotrbak I think it's best to keep things simple and leave it as it is right now. 

**My reasoning:** 
- Clearing the cache is required for every user who switches the modes.
- However, I am guessing that switching the modes is only something that they do while setting up the AMP plugin and not something that is done on a regular basis.
- Even if the customer forgets to clear the cache, the cache will be cleared at the end of lifespan (10 hours). So the mistake will only last for 10 hours in most cases.
- Until we have customer requests asking for this feature, my proposal is to keep things simple and leave it as it is right now.

Discussion in product channel - https://wp-media.slack.com/archives/G8L4EE8N9/p1585813311020900

PR #2497, opened 2020-03-31
Closes #2437 Simple Custom CSS plugin OOP and remove query strings (5/8)

### LINKED ISSUE #2437: Remove Query Strings Option Removal: Simple Custom CSS plugin compatibility update part (5/8)
Sub-card for the epic #2406

#### inc/3rd-party/plugins/simple-custom-css.php
- [x] Migrate the code to `inc/ThirdParty`, use OOP
- [x] Update and test code accordingly to make sure the compatibility still works

crystinutzaa:
Sub-card for the epic #2406

#### inc/3rd-party/plugins/simple-custom-css.php
- [x] Migrate the code to `inc/ThirdParty`, use OOP
- [x] Update and test code accordingly to make sure the compatibility still works


inc/ThirdParty/Plugins/SimpleCustomCss.php
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	/**
+	 * Caches SCCSS code & remove the default enqueued URL
+	 *
+	 * @since 2.9
+	 * @author Remy Perona
+	 */
+	public function cache_sccss() {
+		$sccss = rocket_get_cache_busting_paths( 'sccss.css', 'css' );
```
Tabrisrp, line 40: You can deprecate this function and manage the paths directly in the class

inc/ThirdParty/Plugins/SimpleCustomCss.php
```diff
@@ -0,0 +1,90 @@
...
+
+		if ( ! rocket_direct_filesystem()->exists( $sccss['filepath'] ) ) {
+			$this->create_cache_file( $sccss['bustingpath'], $sccss['filepath'] );
+		}
+
+		// Bailout if the SCCSS file could not be created.
+		if ( ! rocket_direct_filesystem()->exists( $sccss['filepath'] ) ) {
+			return;
+		}
```
Tabrisrp, line 49: You can check that directly in the `create_cache_file` method and use a boolean return value for it

inc/ThirdParty/Plugins/SimpleCustomCss.php
```diff
@@ -0,0 +1,90 @@
...
+	/**
+	 * Deletes & recreates cache for SCCSS code
+	 *
+	 * @since 2.9
+	 * @author Remy Perona
+	 */
+	public function delete_cache_file() {
+		$sccss = rocket_get_cache_busting_paths( 'sccss.css', 'css' );
+
+		array_map( 'unlink', glob( $sccss['bustingpath'] . 'sccss*.css' ) );
```
Tabrisrp, line 64: Let's replace the use of `glob` by either `scandir` or SPL iterators

inc/deprecated/3.5.php
```diff
@@ -17,6 +17,90 @@ class_alias( '\\WP_Rocket\\Engine\\Preload\\ServiceProvider', '\\WP_Rocket\\Serv
...
+/**
+ * Caches SCCSS code & remove the default enqueued URL
+ *
+ * @since 2.9
+ * @deprecated 3.5.3
+ *
+ * @author Remy Perona
+ */
+function rocket_cache_sccss() {
+	_deprecated_function( __FUNCTION__ . '()', '3.5.3', '\WP_Rocket\ThirdParty\Plugins\SimpleCustomCss::cache_sccss()' );
```
Tabrisrp, line 29: Deprecated version should be 3.6

tests/Integration/inc/ThirdParty/Plugins/SimpleCustomCss/cacheSccss.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace WP_Rocket\Tests\Integration\inc\ThirdParty\Plugins\SimpleCustomCss;
+
+use Brain\Monkey\Functions;
```
Tabrisrp, line 5: You can remove the import it's not used in this file

tests/Integration/inc/ThirdParty/Plugins/SimpleCustomCss/deleteCacheFile.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace WP_Rocket\Tests\Integration\inc\ThirdParty\Plugins\SimpleCustomCss;
+
+use Brain\Monkey\Functions;
```
Tabrisrp, line 5: Same here

inc/ThirdParty/Plugins/SimpleCustomCss.php
```diff
@@ -0,0 +1,118 @@
...
+	 *
+	 * @param string $filename name of the cache busting file.
+	 * @return array Array of paths used for cache busting
+	 */
+	private function get_cache_busting_paths( $filename ) {
+		$blog_id                = get_current_blog_id();
+		$cache_busting_path     = rocket_get_constant( 'WP_ROCKET_CACHE_BUSTING_PATH' ) . $blog_id;
+		$filename               = rocket_realpath( rtrim( str_replace( [ ' ', '%20' ], '-', $filename ) ) );
+		$cache_busting_filepath = $cache_busting_path . $filename;
+		$cache_busting_url      = rocket_get_constant( 'WP_ROCKET_CACHE_BUSTING_URL' ) . $blog_id . $filename;
```
Tabrisrp, line 107: I don't think we need this method. The base busting path and busting URL could be setup in the constructor with injected arguments instead.

This will also reduce the number of arguments you need to pass to `create_cache_file`, since they will already be available as properties.

The line 105 was needed before when the function was used for multiple files,  but here we already know the filename is clean.

tests/Unit/inc/ThirdParty/Plugins/SimpleCustomCss/cacheSccss.php
```diff
@@ -19,25 +19,17 @@ public function setUp() {
 		parent::setUp();
 		$this->sccss = new SimpleCustomCss();
 
-		Functions\expect( 'rocket_has_constant' )->with( 'FS_CHMOD_DIR' )->andReturn( true );
-		Functions\expect( 'rocket_get_constant' )->with( 'FS_CHMOD_DIR' )->andReturn( 0755 & ~ umask() );
+		Functions\expect( 'rocket_get_constant' )->with( 'WP_ROCKET_CACHE_BUSTING_PATH' )->andReturn( 'wp-content/cache/busting/' );
+		Functions\expect( 'rocket_get_constant' )->with( 'WP_ROCKET_CACHE_BUSTING_URL' )->andReturn( 'http://example.org/wp-content/cache/busting/' );
```
Screenfeed, line 23: Just  a suggestion, this also can be written like this:
```suggestion
		Functions\expect( 'rocket_get_constant' )
		    ->with( 'WP_ROCKET_CACHE_BUSTING_PATH' )
		    ->andReturn( 'wp-content/cache/busting/' )
		    ->andAlsoExpectIt()
		    ->with( 'WP_ROCKET_CACHE_BUSTING_URL' )
		    ->andReturn( 'http://example.org/wp-content/cache/busting/' );
```

inc/ThirdParty/Plugins/SimpleCustomCss.php
```diff
@@ -0,0 +1,118 @@
...
+
+		/** This filter is documented in inc/functions/minify.php */
+		$cache_busting_url = apply_filters( 'rocket_css_url', $cache_busting_url );
```
Tabrisrp, line 110: You can use the filter in `cache_scss` when we enqueue the style instead

inc/ThirdParty/Plugins/SimpleCustomCss.php
```diff
@@ -58,33 +53,66 @@ public function cache_sccss() {
 	 * @since 2.9
 	 * @author Remy Perona
 	 */
-	public function delete_cache_file() {
-		$sccss = rocket_get_cache_busting_paths( 'sccss.css', 'css' );
-
-		array_map( 'unlink', glob( $sccss['bustingpath'] . 'sccss*.css' ) );
+	public function update_cache_file() {
```
Screenfeed, line 56: The method `delete_cache_file()` should be deprecated.

tests/Unit/inc/ThirdParty/Plugins/SimpleCustomCss/updateCacheFile.php
```diff
@@ -0,0 +1,49 @@
...
+class Test_DeleteCacheFile extends FilesystemTestCase {
+	protected $path_to_test_data = '/inc/ThirdParty/Plugins/SimpleCustomCss/updateCacheFile.php';
+	private $sccss;
+
+	public function setUp() {
+		parent::setUp();
+		$this->sccss = new SimpleCustomCss();
+
+		Functions\expect( 'rocket_get_constant' )->with( 'WP_ROCKET_CACHE_BUSTING_PATH' )->andReturn( 'wp-content/cache/busting/' );
+		Functions\expect( 'rocket_get_constant' )->with( 'WP_ROCKET_CACHE_BUSTING_URL' )->andReturn( 'http://example.org/wp-content/cache/busting/' );
```
Screenfeed, line 23: Here too, you can written this like this if you want:
```suggestion
		Functions\expect( 'rocket_get_constant' )
		    ->with( 'WP_ROCKET_CACHE_BUSTING_PATH' )
		    ->andReturn( 'wp-content/cache/busting/' )
		    ->andAlsoExpectIt()
		    ->with( 'WP_ROCKET_CACHE_BUSTING_URL' )
		    ->andReturn( 'http://example.org/wp-content/cache/busting/' );
```

inc/ThirdParty/Plugins/SimpleCustomCss.php
crystinutzaa: Originally it was : function rocket_delete_sccss_cache_file(), which is deprecated.
Will update the deprecated to:

Screenfeed: The method `delete_cache_file()` was public, it still needs to be deprecated.

crystinutzaa: The method `delete_cache_file()` was not in any production code, only in this PR. 
So I don't think it should be deprecated because I simply renamed it.
Opinions about this one?

Screenfeed: I don't understand, it is tagged `@since 2.9`.

crystinutzaa: Yes, because it was this function before: https://github.com/wp-media/wp-rocket/blob/927900bfdd59d25a6809eaa358b2357886af0220/inc/deprecated/3.6.php#L74

Screenfeed: OK, then the `@since` tag is misleading, it should be `3.6` imho, as this method doesn't exist before 3.6.

inc/deprecated/3.6.php
```diff
@@ -4,6 +4,127 @@
...
+ * @since 2.9
+ * @deprecated 3.6
+ * @author Remy Perona
+ *
+ * @param string $filename name of the cache busting file.
+ * @param string $extension file extension.
+ * @return array Array of paths used for cache busting
+ */
+function rocket_get_cache_busting_paths( $filename, $extension ) {
+	_deprecated_function( __FUNCTION__ . '()', '3.6', '\WP_Rocket\ThirdParty\Plugins\SimpleCustomCss::get_cache_busting_paths()' );
```
Tabrisrp, line 19:
```suggestion
	_deprecated_function( __FUNCTION__ . '()', '3.6' );
```

inc/deprecated/3.6.php
```diff
@@ -4,6 +4,127 @@
...
+ * @since 2.9
+ * @deprecated 3.6
+ *
+ * @author Remy Perona
+ *
+ * @param string $filename filename.
+ * @return string filename
+ */
+function rocket_sccss_cache_busting_filename( $filename ) {
+	_deprecated_function( __FUNCTION__ . '()', '3.6', '\WP_Rocket\ThirdParty\Plugins\SimpleCustomCss::cache_busting_filename()' );
```
Tabrisrp, line 95:
```suggestion
	_deprecated_function( __FUNCTION__ . '()', '3.6' );
```

hellofromtonya:
@arunbasillal This one is ready for QA.

arunbasillal:
Starting QA

crystinutzaa:
@arunbasillal please recheck. I missed it. 🙏

arunbasillal:
@crystinutzaa The path is corrected now, but it looks like this: `http://wp-rocket-dev.local/wp-content/cache/busting/1/wp-content/cache/busting/1/sccss-1586863241.css`

`/wp-content/cache/busting/1/` seems to be repeated. Is this expected?

crystinutzaa:
Yes, it seems is the desired file / folder path when Remove Query String option is enabled.

Simple Custom CSS plugin is a particular case and the CSS file is saved in /busting folder. The goal with this 3rd party compatibility is to prevent dynamic generation of the CSS on each request. So we save the CSS content inside a file in the busting folder. Then it will be treaten exactly as a CSS file from a third party plugin even if it resides in /cache/busting folder.

Tabrisrp:
Since the remove query strings option is going to be removed, this repetition is not an issue as it won't be possible to happen once everything on this epic is finished 👍

Issue #2377, opened 2020-03-05
CPCSS for specific pages (using metabox)


GeekPress:
## Why are we doing this feature?

The feature proposal is currently to add page specific CPCSS to any post type that is public and has an editor interface (i.e. the WP Rocket options metabox will be available). It's very common for a lot of publishers to have custom page layouts. We have many examples on our WP Rocket website (example: <https://wp-rocket.me/features>). 

Having page specific CPCSS will benefit websites with such custom layouts. 

## User expectations and experience 

Users will be able to generate custom CPCSS for their pages in one click. 

## How will we do this feature?

- In the beginning, the page uses the default CPCSS. At this point, we need only one button that says, "Generate CPCSS For This Page".
- On clicking it, CPCSS is generated and saved with other CPCSS CSS files as post_type-id.css (example page-19.css)
- "Generate CPCSS" now renamed to "Regenerate CPCSS". Click it will regenerate the CPCSS. 
- Below it a small link says "Disable custom CPCSS". This will delete the custom generated CPCSS and fall back to the common CPCSS for the post type. 

UI being discussed here: <https://wp-media.slack.com/archives/CFSCRLVGB/p1580933950000100>

Technical overview

Flowchart for admin/front: <https://drive.google.com/file/d/1K5vPc0_JURHWiWX7KujZAo1OhJczZbpj/view?usp=sharing>

  
**Code**
- [x] [UI](https://github.com/wp-media/wp-rocket/issues/2443)
- [x] [AJAX](https://github.com/wp-media/wp-rocket/issues/2444)
- [x] [CPCSS generation](https://github.com/wp-media/wp-rocket/issues/2445)
- [x] [CPCSS deletion](https://github.com/wp-media/wp-rocket/issues/2446)
- [x] [Update common CPCSS deletion](https://github.com/wp-media/wp-rocket/issues/2447)
- [x] [Front](https://github.com/wp-media/wp-rocket/issues/2448)

- New REST API endpoint for the CPCSS, with a subscriber class to register it, handle the validation & callbacks
- New JS script on the edit.php page to add the AJAX handlers
- New subscriber for the CPCSS metabox
- Update CPCSS subscriber to use the specific CPCSS when it exists
- Update clean method from CPCSS class, to ignore pages CPCSS
- Add automated tests for all the above
  
**Estimate**

~~2 weeks~~ 1 week of work with Tonya (doing the JS) and Remy doing the plugin work.

## Risks

Risk should be minimal because: 

[...]

PR #2508, opened 2020-04-02
Closes #2447: Update Common CPCSS deletion method (Part 1)

### LINKED ISSUE #2447: CPCSS for specific pages: update common CPCSS deletion method (1/6)
Sub-Card for epic #2377.

This card is for tasks related to the common CPCSS deletion method.

#### inc/classes/optimization/CSS/class-critical-css.php
- [x] Update `clean_critical_css()` to only delete CSS files in the root `cache/critical-css` folder
- [x] Write unit tests
- [x] Write integration tests

crystinutzaa:
Sub-Card for epic #2377.

This card is for tasks related to the common CPCSS deletion method.

#### inc/classes/optimization/CSS/class-critical-css.php
- [x] Update `clean_critical_css()` to only delete CSS files in the root `cache/critical-css` folder
- [x] Write unit tests
- [x] Write integration tests
- [x] Move into our new architecture
- [x] QA


inc/classes/optimization/CSS/class-critical-css.php
```diff
@@ -129,26 +129,19 @@ public function stop_generation() {
...
-		}
-
-		try {
-			$files = new \RecursiveIteratorIterator( $directory, \RecursiveIteratorIterator::CHILD_FIRST );
-		} catch ( \Exception $e ) {
-			// no logging yet.
-			return;
-		}
+		$filesystem = rocket_direct_filesystem();
+		$dir        = $filesystem->dirlist( $this->critical_css_path );
```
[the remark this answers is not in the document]
Tabrisrp: The old code was recursive for no real reason, all generic CPCSS files are at the root of the folder.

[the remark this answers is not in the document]
Screenfeed: OK, thanks :+1:

crystinutzaa:
From my point of view: `dirlist()` seems cleaner and easier to use. And it relies into Wordpress `WP_Filesystem_Direct` class. I don't find any reasons why to not use it and rely into FileSystemIterator .
Related to performance probably @hellofromtonya has much more insights

hellofromtonya:
Comment by @Tabrisrp 
>I'm wondering for clean_critical_css, if we should go with the use of dirlist, or with the FileSystemIterator SPL class, for consistency with the other clean methods we use elsewhere in the code.

Comment by @crystinutzaa 
>From my point of view: dirlist() seems cleaner and easier to use. And it relies into Wordpress WP_Filesystem_Direct class. I don't find any reasons why to not use it and rely into FileSystemIterator .

Use the SPL instead of `dirlist()` unless you need all the directory detail information. `dirlist()` is not performant. Why? It's gathering all the information such as permissions, group, owner, etc. for each and every file and directory into the given path. It returns an array with all that information.

Here, we need a list of the directories in order to iterate through them. Then check if it's a file or directory for the conditional work within the loop itself.

hellofromtonya:
On my TODO list is to build `RocketFilesystem` which will have method(s) for the SPL functionality we need. Then we can standardize and reduce the redundancy in our code.

inc/classes/optimization/CSS/class-critical-css.php
hellofromtonya, line 133: `dirlist()` is not performant nor is it needed for this implementation. Instead, use SPL.

inc/classes/subscriber/Optimization/class-critical-css-subscriber.php
```diff
@@ -2,7 +2,7 @@
 namespace WP_Rocket\Subscriber\Optimization;
```
Tabrisrp, line 2:
```suggestion
namespace WP_Rocket\Engine\Optimization\CSS\CriticalPath;
```
We can move & rename the subscriber too

phpcs.xml
```diff
@@ -78,13 +78,13 @@
 	<rule ref="WordPress.DB.DirectDatabaseQuery.DirectQuery">
 		<exclude-pattern>inc/classes/admin/Database/class-optimization-process.php</exclude-pattern>
 		<exclude-pattern>inc/classes/admin/Database/class-optimization.php</exclude-pattern>
-		<exclude-pattern>inc/classes/optimization/CSS/class-critical-css.php</exclude-pattern>
+		<exclude-pattern>inc/Engine/Optimization/CSS/CriticalCSS.php</exclude-pattern>
```
Tabrisrp, line 81:
```suggestion
		<exclude-pattern>inc/Engine/Optimization/CSS/CriticalPath/CriticalCSS.php</exclude-pattern>
```

phpcs.xml
```diff
@@ -78,13 +78,13 @@
...
 	</rule>
 
 	<rule ref="WordPress.DB.DirectDatabaseQuery.NoCaching">
 		<exclude-pattern>inc/classes/admin/Database/class-optimization-process.php</exclude-pattern>
 		<exclude-pattern>inc/classes/admin/Database/class-optimization.php</exclude-pattern>
-		<exclude-pattern>inc/classes/optimization/CSS/class-critical-css.php</exclude-pattern>
+		<exclude-pattern>inc/Engine/Optimization/CSS/CriticalCSS.php</exclude-pattern>
```
Tabrisrp, line 87:
```suggestion
		<exclude-pattern>inc/Engine/Optimization/CSS/CriticalPath/CriticalCSS.php</exclude-pattern>
```

phpcs.xml
```diff
@@ -94,7 +94,7 @@
 
 	<rule ref="WordPress.DB.PreparedSQL.InterpolatedNotPrepared">
 		<exclude-pattern>inc/classes/admin/Database/class-optimization-process.php</exclude-pattern>
-		<exclude-pattern>inc/classes/optimization/CSS/class-critical-css.php</exclude-pattern>
+		<exclude-pattern>inc/Engine/Optimization/CSS/CriticalCSS.php</exclude-pattern>
```
Tabrisrp, line 97:
```suggestion
		<exclude-pattern>inc/Engine/Optimization/CSS/CriticalPath/CriticalCSS.php</exclude-pattern>
```

inc/classes/ServiceProvider/class-common-subscribers.php
```diff
@@ -52,7 +52,7 @@ public function register() {
 			->withArgument( $this->getContainer()->get( 'db_optimization' ) )
 			->withArgument( $this->getContainer()->get( 'options' ) );
 		$this->getContainer()->add( 'critical_css_generation', 'WP_Rocket\Optimization\CSS\Critical_CSS_Generation' );
-		$this->getContainer()->add( 'critical_css', 'WP_Rocket\Optimization\CSS\Critical_CSS' )
+		$this->getContainer()->add( 'critical_css', 'WP_Rocket\Engine\Optimization\CSS\CriticalCSS' )
```
hellofromtonya, line 55: Shouldn't all of the critical CSS be in the optimization service provider? @Tabrisrp @crystinutzaa

Tabrisrp: It's currently in the common because the CriticalCSS subscriber has code for the admin and for the front at the same time (and some for both, for example for regeneration triggered from the admin bar).

And the Optimization service provider is loaded only when it's _not_ the admin, so it wouldn't work.

It's of course possible to change that if we feel it's necessary. We can either:
- Load the optimization service provider always, but I don't think it's great
- Split the Critical CSS subscriber in multiple parts, loaded in the appropriate place (one for admin, one for front, one for common)

hellofromtonya: Makes sense. @Tabrisrp We want to move both the common and optimization service providers into our new architecture. What do you think makes sense for bundling everything?

Tabrisrp: IMO, the critical CSS subscriber is currently doing too much (the whole class has high complexity, and 13 public methods)

So if we want to tackle that right now, what makes the most sense to me is breaking it into 3 parts.

Once this is done, we can move each ServiceProvider (Admin/Common/Optimization) into the new architecture.

hellofromtonya: Other ways to architecture it are:

1. Have a Critical CSS module that contains all of its files (self-contained) including its own `ServiceProvider`. 
2. Have an `Optimization` module where each submodule has all of its own files (self-contained) and then there's one `ServiceProvider` for the entire `Optimization` module.

Remember, the idea with the new architecture is to have complete modules and sub-modules.

@Tabrisrp What do you think?

Tabrisrp:
I like the option 1, especially since we’re going to add more classes with the new cpcss feature. 

I’m not sure to completely visualize option 2, do you have an example of how it would look?

hellofromtonya:
>I’m not sure to completely visualize option 2, do you have an example of how it would look?

Sure. Look at [`inc/Addon/ServiceProvider`](https://github.com/wp-media/wp-rocket/blob/master/inc/Addon/ServiceProvider.php). Notice how there's just one Service Provider for _all_ of the addons instead of one per addon.

So for `inc/Engine/Optimization`, there could be just one `ServiceProvider` that handles all of the modules within it. But if there's too many, then we may want to go deeper instead of it and break up the service providers by a logical grouping that makes sense for the modules each handles.

crystinutzaa:
I made the first round of moving including CriticalCSSGeneration and all tests and fixtures in the new structure.

For the second part, to split the Subscriber. I created the ServiceProvider for Optimization and moved Critical CSS in there. 

I like the idea to split the subscriber & Service provider in multiple parts and I think there will be a slight speed increase if we load only what we need in each section : 

> Split the Critical CSS subscriber in multiple parts, loaded in the appropriate place (one for admin, one for front, one for common)

However, how will this be managed. Should I create a folder inside `Engine\Optimization\CSS\CriticalPath`:
```
/admin
/common
/front
```
and create inside each a Subscriber?
And for ServiceProvider should we have a CommonServiceProvider, AdminServiceProvider, FrontServiceProvider? 

@hellofromtonya , @Tabrisrp What do you think? And how should I proceed on this PR?

inc/Engine/CriticalPath/CriticalCSSGeneration.php
```diff
@@ -11,7 +9,7 @@
  *
  * @see WP_Background_Process
  */
-class Critical_CSS_Generation extends \WP_Background_Process {
+class CriticalCSSGeneration extends \WP_Background_Process {
```
Tabrisrp, line 12: import `WP_Background_Process`

inc/Engine/CriticalPath/ServiceProvider.php
```diff
@@ -0,0 +1,44 @@
+<?php
+namespace WP_Rocket\Engine\CriticalPath;
+
+use League\Container\ServiceProvider\AbstractServiceProvider;
+use WP_Rocket\Admin\Options_Data;
```
Tabrisrp, line 5: This import is not necessary

inc/Engine/CriticalPath/ServiceProvider.php
```diff
@@ -0,0 +1,44 @@
...
+	 */
+	public function register() {
+		$options = $this->getContainer()->get( 'options' );
+
+		$this->getContainer()->add( 'critical_css_generation', 'WP_Rocket\Engine\CriticalPath\CriticalCSSGeneration' );
+		$this->getContainer()->add( 'critical_css', 'WP_Rocket\Engine\CriticalPath\CriticalCSS' )
+			->withArgument( $this->getContainer()->get( 'critical_css_generation' ) );
+		$this->getContainer()->share( 'critical_css_subscriber', 'WP_Rocket\Engine\CriticalPath\CriticalCSSSubscriber' )
+			->withArgument( $this->getContainer()->get( 'critical_css' ) )
+			->withArgument( $this->getContainer()->get( 'options' ) );
```
Tabrisrp, line 42:
```suggestion
			->withArgument( $options );
```

inc/Engine/CriticalPath/CriticalCSSGeneration.php
crystinutzaa: Thank you!

inc/Engine/CriticalPath/ServiceProvider.php
crystinutzaa: I removed the `$options` variable because it is used right now in a single place, so it does not need to be declared as a variable.

crystinutzaa:
@arunbasillal This one is ready for QA.

PR #2509, opened 2020-04-02
Make rocket_mkdir_p compatible with streams

Tabrisrp:
This makes the function compatible with VfsStream (and other stream URLs)


hellofromtonya:
@Tabrisrp I'll add some tests for this one as a baseline.

hellofromtonya:
@Tabrisrp Take a look at the changes I added:

- Adds `rocket_is_stream()` which is a copy of `wp_is_stream()`
- Replaces the first conditional with `rocket_is_stream()` to match `wp_mkdir_p()`
- Replaces `untrailingslashit()` with `rtrim`

BTW vfsStream registers itself to the PHP streams. 🎉

PR #2510, opened 2020-04-02
Replaces glob in rocket_clean_domain and makes rocket_rrmdir work with dir preserves.

Tabrisrp:
This PR replaces `glob` in `rocket_clean_domain()` with SPL iterators. Why?

- `glob` causes problems on some hosts
- it doesn't work with our virtual filesystem.

SPL iterators are a powerful, common, and popular way to traverse a filesystem. It's found in libraries/frameworks such as [Symfony](https://github.com/symfony/symfony/blob/1866a2a88c6c47ce64cd92a48cbc7a434eeec776/src/Symfony/Component/Filesystem/Filesystem.php) and Laravel.

Changes in this PR:

- Replaces `glob()` in `rocket_clean_domain` with SPL iterators
- Restores the original `rocket_rrmdir()` preserve language directories logic
- Standardizes the virtual filesystem structure
- Updates the `rocket_rrmdir()` test suites to use the new vfs structure and same testing strategy as `rocket_clean_domain()`

## TODO

- [x] i18n tests -- ✅ PR #2518 and PR #2529 
- [x]  `rocket_rrmdir()` replace `glob()` and add tests -- ✅ PR #2522 
- [x] Validate `glob()` behavior and compare to SPL behavior
- [x] Validate how the three i18n plugins work with rocket
     - ✅with the original code
     - ✅with this PR (works the same)
- [x] Use one vfs structure
- [x] Add integration tests
- [x] Make `rocket_rrmdir()` work with preserve languages
- [x] Add unit tests
- [x] Update `rocket_rrmdir()` tests
- [x] QA 👈 source code changed.


hellofromtonya:
@Tabrisrp I'll add tests to this one.

inc/functions/files.php
```diff
@@ -839,12 +855,15 @@ function rocket_clean_domain( $lang = '' ) {
...
-		// Delete cache domain files.
-		$dirs = glob( $root . '*', GLOB_NOSORT );
-		if ( $dirs ) {
-			foreach ( $dirs as $dir ) {
-				rocket_rrmdir( $dir, get_rocket_i18n_to_preserve( $lang ) );
+		try {
+			$files = new RegexIterator( $iterator, "/{$file['host']}*{$file['path']}/i" );
+
+			foreach ( $files as $file ) {
+				rocket_rrmdir( $file, get_rocket_i18n_to_preserve( $lang ) );
```
Screenfeed, line 862: We could do this before the first `foreach` loop on L837.
```php
$dirs_to_preserve = get_rocket_i18n_to_preserve( $lang );
```
And maybe also this, while at it:
```php
$url_no_dots = apply_filters( 'rocket_url_no_dots', false );
```

hellofromtonya:
Working on the tests for this PR now...

inc/functions/files.php
```diff
@@ -818,15 +818,34 @@ function rocket_clean_domain( $lang = '' ) {
...
+		// No logging yet.
+		return;
+	}
+
+	try {
+		$iterator = new RecursiveIteratorIterator( $cache, RecursiveIteratorIterator::CHILD_FIRST );
+	} catch ( Exception $e ) {
+		// No logging yet.
+		return;
+	}
```
hellofromtonya, line 835: @Tabrisrp This code is changing the behavior of what files/directories are passed to `rocket_rrmdir()` and the resultant deletion. 

With `glob()`:

1. It's passing the matching files/dirs from the given root, but not the child files/dirs within each level. It's not iterative.
2. It leaves the root directory(ies), i.e. deletes what is inside of each.

## What is this code doing?

It's recursively getting _all_ files/directories within the given cache and matching root pattern. **ALL** It's changing the behavior of this function and impacting performance when we run `rocket_rrmdir()`.

## How so?

Because it's recursively grabbing all levels of files/directories, it runs `rocket_rrmdir()` for each of those entries. 

## Why is that problematic? 

Let's look at what's happening in `rocket_rrmdir()`.

 Each time it's called:

- What's passed to it:
    - `glob()`: the entries inside of the root, but not the root itself.
   - this PR's code: the root + entries inside of it
         - Result: the root is also deleted.
- It fires the action `'before_rocket_rrmdir'`. 
     - With this code: same as `glob()` plus the number of the root files/directories. Result: Fired more times than `glob()`.
  - If it's a file, it's deleted and bails out.
  - If it's a directory:
       - it runs `glob` to get a list of all child entries at that level.
       - it filters the list to remove those that are set to "preserve"
       - And then it recursively iterates over the child entries returned from its `glob`.
       - So it's scanning the filesystem of each given directory. 
           - That's fine with the current behavior. 
           - With this new code, we're _rescanning the filesystem for each directory_. "Rescanning" is not performant.
  - For each directory, it fires `after_rocket_rrmdir`. <- directories only. 
      - Works the same for both `glob` and this new code.

Okay. So the code introduced in this PR has a negative impact on `rocket_rrmdir()` in terms of memory and processing speed.

## What can we do to preserve the current behavior, but get rid of `glob()`?

Replace both `try/catch` with:

```php
try {
	$iterator = new FilesystemIterator( $cache_path, FilesystemIterator::SKIP_DOTS );
} catch ( Exception $e ) {
	// No logging yet.
	return;
}
```

inc/functions/files.php
```diff
@@ -818,15 +818,34 @@ function rocket_clean_domain( $lang = '' ) {
...
 		$file = get_rocket_parse_url( $url );
 
-		/** This filter is documented in inc/front/htaccess.php */
-		if ( apply_filters( 'rocket_url_no_dots', false ) ) {
+		if ( $url_no_dots ) {
 			$file['host'] = str_replace( '.', '_', $file['host'] );
 		}
 
-		$root = WP_ROCKET_CACHE_PATH . $file['host'] . '*' . $file['path'];
+		$root = $cache_path . $file['host'] . $file['path'];
```
hellofromtonya, line 848: We've changed the value of `$root` that's being passed to any callbacks registered to both actions. But we are not acting upon `$root`, i.e. only passing it to both action events. 

Since we're not using it, should we retain the original for those callbacks using it? Then we won't break any 3rd party code.

inc/functions/files.php
Tabrisrp: You're right, thank you for catching that. I'm updating the code.

For later, do you think it would be more efficient to not rely on `rocket_rrmdir` for this function, and instead grab everything we need with the iterators, instead of the multiple `glob` calls?

> It leaves the root directory(ies), i.e. deletes what is inside of each.

This part is kind of an oversight from the original code in fact, the directories for each cache path _should_ be deleted when all their content is deleted too. But we can work on that later too.

inc/functions/files.php
Tabrisrp: I made this change because in fact the `*` is more of a bother than useful when we are hooking on `before_rocket_clean_domain` in other parts of the code.

As an example in here for our Varnish addon: https://github.com/wp-media/wp-rocket/blob/00cf7a1c5e0b2e7354b0aa76ca86603ee4420eae/inc/Addon/Varnish/Varnish.php#L114

I'm removing the `*` before using the value, else it wouldn't work correctly.

I thought it was a good opportunity to get rid of something that didn't serve us all this time. But I can revert it if we feel it's too much.

inc/functions/files.php
Tabrisrp: Done in https://github.com/wp-media/wp-rocket/pull/2510/commits/186cae0ddae49e17ecc2f3bb050003b07542015a

hellofromtonya: >For later, do you think it would be more efficient to not rely on `rocket_rrmdir` for this function, and instead grab everything we need with the iterators, instead of the multiple glob calls?

Yes, absolutely. When we refactor the filesystem functions into our own `Filesystem` implementation, then we can look at reducing the redundancy and optimizing everything. 

For example, these `try/catch` chunks of code will be held within the new `Filesystem` class. That'll reduce the code where we use it.

inc/functions/files.php
hellofromtonya: Hmm, interesting point. What do you think is best?

Tabrisrp: I know that internally, we don't have a use for the `*` anywhere when we use `before/after_rocket_clean_domain`.

I don't know if a 3rd party has/had a use for it. But we do know that it was causing an issue for some, like reported in this issue: https://github.com/wp-media/wp-rocket/issues/1527

inc/functions/files.php
```diff
@@ -854,7 +854,12 @@ function rocket_clean_domain( $lang = '' ) {
 		do_action( 'before_rocket_clean_domain', $root, $lang, $url ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
 
 		try {
-			$files = new RegexIterator( $iterator, "/{$file['host']}*{$file['path']}/i" );
+			$regex = sprintf( "/{$file['host']}*%s/i",
+				empty( $file['path'] )
```
Screenfeed, line 858: Shouldn’t the logic be reversed?
```suggestion
				'' !== trim( $file['path'], '/\\' )
```

hellofromtonya:
@arunbasillal This one is ready for QA. Please put this one at the top of your list as a few other open PRs are dependent upon it. Thanks.

arunbasillal:
@hellofromtonya Should I wait for @Screenfeed's review so as to avoid double QA if there are any suggested changes?

hellofromtonya:
@arunbasillal No, go ahead and do the QA.

PR #2522, opened 2020-04-06
rocket rrmdir() - Optimizes, replaces glob, and provides full code coverage

hellofromtonya:
`rocket_rrmdir()` is an original filesystem function that has not been optimized nor does it have tests. This PR gives it some love as it's a critical part of Rocket.

This PR does the following:

- 🏁 Adds tests
- 🚀 Optimizes by calling `rocket_direct_filesystem()` 1x instead of 11x
- 🚀 Checks if the given directory is in the list of preserves _before_ doing anything else
- 🚀 Optimizes the check for directory in preserves list
- 🏁 Replaces `glob()` with SPL

👉 Needed to move PR #2510 forward.

## TODO

- [x] Validate glob behavior
- [x] Add unit and integration tests
- [x] Optimize
- [x] Replace `glob`
- [ ] QA 👈 needed because the source code did change.

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1040,6 +1040,11 @@ function rocket_rrmdir( $dir, $dirs_to_preserve = [] ) {
 	$dir        = untrailingslashit( $dir );
 	$filesystem = rocket_direct_filesystem();
 
+	// Bail out if the given directory is in the list of directories to preserve.
+	if ( ! empty( $dirs_to_preserve ) && $filesystem->is_dir( $dir ) ) {
+		return;
+	}
+
 	/**
 	 * Fires before a file/directory cache is deleted
 	 *
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1037,7 +1037,8 @@ function rocket_clean_cache_dir() {
  * @return void
  */
 function rocket_rrmdir( $dir, $dirs_to_preserve = [] ) {
-	$dir = untrailingslashit( $dir );
+	$dir        = untrailingslashit( $dir );
+	$filesystem = rocket_direct_filesystem();
 
 	/**
 	 * Fires before a file/directory cache is deleted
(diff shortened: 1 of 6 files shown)


inc/functions/files.php
```diff
@@ -1031,72 +1031,82 @@ function rocket_clean_cache_dir() {
...
-	$dirs = glob( $dir . '/*', GLOB_NOSORT );
-	if ( $dirs ) {
+	// Get the directory entries.
+	try {
+		$entries = new FilesystemIterator( $dir, FilesystemIterator::SKIP_DOTS );
+	} catch ( Exception $e ) {
+		$entries = [];
+	}
+
+	foreach ( $entries as $entry ) {
```
Tabrisrp, line 1085: Shouldn't we check if `$entries` is empty, before the `foreach`? And if empty, delete the directory then bailout?

hellofromtonya: We are. If the `$entries` are empty, the `foreach` doesn't run and skips over all the code within the `foreach`. What happens next?

- It jumps down to line 1100 and deletes the directory.
- It then fires the `after_rocket_rrmdir` action event.
- It then exits the function. Done.

@Tabrisrp Do you think it shouldn't fire the action event?

Tabrisrp: I was under the impression that the `foreach` would try to run even with an empty array and create an error, but it's very likely my brain doesn't work this morning

hellofromtonya: Nope. If the array is empty, a `foreach` does not run.

Tabrisrp: A new day a new thing to learn

hellofromtonya: For PHP array functions, such as `foreach`, `array_walk`, etc., first check if there's anything to iterate over. If no, i.e. the array is empty, then these functions bail out. Pretty cool!

hellofromtonya:
@arunbasillal This PR is ready for QA.

arunbasillal:
Starting QA

PR #2524, opened 2020-04-07
Closes #2473 Add preconnect for Google Fonts

### LINKED ISSUE #2473: Preconnect requests resources from Google Fonts
If the « Combine Google Fonts » options is activated, we can preconnect requests resources from Google Fonts (fonts.gstatic.com).

Adding the preconnect resource hint will instruct the browser to preconnect to Google Fonts while it is loading the CSS, saving load time.

The solution is to automatically add this line in the `<head>` section:
`<link href="https://fonts.gstatic.com" crossorigin rel="preconnect" />`

**Backlog Grooming**
- [x] Scope a solution
- [x] Estimate the effort

Tabrisrp:
Closes #2473 

## TODO

- [x] Move & rename Combine Google Fonts & Subscriber to new architecture
- [x] Add `preconnect` method to Google Fonts Subscriber
- [x] Add unit & integration tests for `preconnect`
- [x] QA test


tests/Integration/inc/Engine/Optimization/GoogleFonts/Subscriber/preconnect.php
```diff
@@ -0,0 +1,33 @@
...
+        };
+
+		add_filter( 'pre_get_rocket_option_minify_google_fonts', $callback );
+
+		$this->assertSame(
+			$expected,
+			apply_filters( 'wp_resource_hints', $urls, $relation_type )
+        );
+
+        remove_filter( 'pre_get_rocket_option_minify_google_fonts', $callback );
```
hellofromtonya, line 27: @Tabrisrp Instead of using a closure as a callback, let's make it a method and then move the `remove_filter()` to the `tearDown()` method. 

Why? If the test fails, the `remove_filter()` will not run. But by moving it into the `tearDown()`, we make sure that it's unhooked after each test.

Tabrisrp: how can I pass the value from `$option_value` to the method in that case?

hellofromtonya: Set it to a private property in the test class. Then reset that property in `setUp()`.

Tabrisrp: Done in https://github.com/wp-media/wp-rocket/pull/2524/commits/befeef31c3b583bc0ef34ee67cf2c504a7e7cb56

Tabrisrp:
@arunbasillal this is ready for QA

Acceptance criteria:
- Enable the google fonts option
- On the front, you should see `<link href="https://fonts.gstatic.com" crossorigin rel="preconnect" />` in the head of the HTML

arunbasillal:
Starting QA on this one

GeekPress:
@arunbasillal We should not add if no fonts. Why? It's going to make a request from nothing and it's going to have a performance cost for no use.

arunbasillal:
@Tabrisrp Please note: https://github.com/wp-media/wp-rocket/pull/2524#issuecomment-613282866

Tabrisrp:
To do that we have to completely change the approach we took for development, because in the current context of the method, it's not possible to know if there is any font on the page.

And what about Google fonts loaded from CSS files? They would not be in the HTML, but they would still benefit from the `preconnect`.

GeekPress:
@Tabrisrp Ok let's keep the actual behavior. We plan to remove the UI option in 3.7 to automatically apply the Google Fonts Optimization. We will change the development approach during that change.

I guess we can consider the QA done :D

PR #2529, opened 2020-04-08
Tests: adds tests for get_rocket_i18n_home_url() and get_rocket_i18n_to_preserve()

hellofromtonya:
Adds unit and integration tests for:

- `get_rocket_i18n_to_preserve()`
- `get_rocket_i18n_home_url()`

This PR is needed for PR #2510 and others.

QA is not needed for this PR.

### DIFF
--- a/tests/Fixtures/inc/functions/getRocketI18nToPreserve.php
+++ b/tests/Fixtures/inc/functions/getRocketI18nToPreserve.php
@@ -1,0 +1,210 @@
+<?php
+
+$i18n_plugins = require WP_ROCKET_TESTS_FIXTURES_DIR . '/i18n/pluginsData.php';
+
+return [
+	'false_ShouldReturnEmptyArrayWhenNoLangGiven'   => [
+		'current_lang' => '',
+		'config'       => [
+			'i18n_plugin'     => false,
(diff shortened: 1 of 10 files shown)


tests/Fixtures/inc/functions/getRocketI18nToPreserve.php
```diff
@@ -0,0 +1,466 @@
...
+	'wpml_ShouldReturnEnAndDe'                                        => [
+		'current_lang' => 'fr',
+		'config'       => [
+			'data'            => $i18n_plugins['wpml'],
+			'i18n_plugin'     => 'wpml',
+			'rocket_has_i18n' => 'wpml',
+		],
+		'expected'     => [
+			'vfs://public/wp-content/cache/wp-rocket/example.org(.*)/',
+			'vfs://public/wp-content/cache/wp-rocket/example.org(.*)/',
```
hellofromtonya, line 115: @Tabrisrp This is the behavior we were talking about. When the i18n plugin, which in this case is WPML, returns the language as a query string, i.e. `?lang=fr`, currently this function doesn't handle it. We can fix this in a separate PR.

PR #2530, opened 2020-04-09
Closes #2430 - PHP Deprecated notice - base_convert() in ip_in_range.php and PHP7.4

### LINKED ISSUE #2430: PHP Deprecated : ip_in_range.php on line 108
Dear,

I'm receiving many errors in my debug.log file with the below description. There seem to be a deprecation in WP Rocket 3.5.0.4 under PHP 7.4.

[16-Mar-2020 17:24:39 UTC] PHP Deprecated:  Invalid characters passed for attempted conversion, these have been ignored in /home/customer/www/DOMAIN/public_html/wp-content/plugins/wp-rocket/inc/vendors/ip_in_range.php on line 108

Can you please fix it?
Thanks

**Steps to reproduce**
- PHP 7.4
- Website on Cloudflare
- IPv6

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
This is a bandaid for the current problem. 
As @Tabrisrp  said the proper solution would be to remove totally ip_in_range.php (which is really old - last modification in 2012 and is not maintained anymore) and move all the code in Cloudflare Addon.
I would recommend to embed this package and adapt it for our needs:

> I discovered that there is an official package from Cloudflare to handle IP checking and rewriting: https://github.com/cloudflare/cf-ip-rewrite

**Reproduce the problem** ✅ 
We would need a website with PHP 7.4, Cloudflare, and someone who uses IPv6 to reproduce ourselves. We have a good number of tickets to see that the issue is reproduced though.

**Identify the root cause** ✅
This is happening because starting with PHP 7.4, `base_convert()` sends a notice when invalid characters are used for its first parameter. Before, it was ignoring them silently.

**Scope a solution** ✅
A solution is to sanitize the value passed to `base_convert()` beforehand, to only preserve valid characters. Those are `[0-9a-fA-F]` since IPv6 uses a base 16 format (hexadecimal).

**Estimate the effort** ✅
effort: `[S]`, need tests for `rocket_ip2long6()`


hellofromtonya:
@crystinutzaa Isn't this file only used by our Cloudflare addon? If yes, shouldn't it be included as part of that addon?

Tabrisrp:
I don't think we should go forward with a bandaid fix like this.

I know the priority is set to high, but this issue has created (at the time of writing) 5 tickets to our support, all of them closed with the users not complaining further.

I believe it would be beneficial for us to take the time to handle this issue properly, by moving the code to the Cloudflare addon, and testing the package to replace this old code.

The package has its own test suite, which is a good thing for us. We can build upon that.

crystinutzaa:
Yes. If this is really urgent (because it had the priority high) it can be merged and will fix the issue.
But the long term solution will be to use that package and move everything in Cloudflare.
I think it needs to be scheduled because this one will be a [M] effort.
This file is used only in Cloudflare and some functions are used also outside Flexible SSL fix.

hellofromtonya:
Good points @Tabrisrp and @crystinutzaa. 

'd like to propose a compromise:

1. Fix the current issue.
2. Then open a new task to move this file into the CF repo and addon.

Why? As Cristina points out, there are other files that need to move to CF too. It makes sense to create a separate task and PR specifically for that work.

What do you both think @crystinutzaa and @Tabrisrp?

Tabrisrp:
If we all agree that's an acceptable compromise for the time being to make sure there is no notice on users websites, I'm ok with it

Tabrisrp:
@arunbasillal ready for QA

arunbasillal:
Starting QA. Also wondering how this can be tested 💭

crystinutzaa:
@arunbasillal this is a hard issue to be recreated on it's own. BUT it can be forced.
You need PHP 7.4, Cloudflare addon enabled. 
Even if you have a IP6 IP you need a malformed one, so I would force the IP to contain a "/"

In /inc/common/cloudflare-flexible-ssl.php on line: 34 I would force the IP to be:
`$original_ip = '2001:db8::/32';`

arunbasillal:
I am having trouble setting up PHP 7.4 on my local. Requested @piotrbak to handle this one instead - https://wp-media.slack.com/archives/GUKB44GNN/p1586871505008900

piotrbak:
@arunbasillal It can't be reproduced on the local, I've reached affected customers and asked about the possibility to test on their environment
https://wp-media.slack.com/archives/GUKB44GNN/p1586896385011400?thread_ts=1586871505.008900&cid=GUKB44GNN

piotrbak:
I wasn't able to reproduce the problem on my environment good enough to be sure that it indeed fixes the deprecated message. Related convo:
https://wp-media.slack.com/archives/GUKB44GNN/p1587072735020900?thread_ts=1586871505.008900&cid=GUKB44GNN

Asked customer for help in the meantime (not many of affected answered):
https://secure.helpscout.net/conversation/1121941586/154278?folderId=2415573

GeekPress:
The customer confirmed the update fixed the issue. QA done

PR #2531, opened 2020-04-09
Fixes `rocket rrmdir` directories to preserve checks

hellofromtonya:
PR #2522 broke the check for directories to preserve. How? The directories include regex pattern. For example, PR 2522 did not process:

```
[
	'vfs://public/wp-content/cache/wp-rocket/example.org(.*)/fr',
	'vfs://public/wp-content/cache/wp-rocket/example.org(.*)/de',
]
```

This PR does the following:

- Converts the array of directories to preserve into an OR string and then escapes the `/` for regex
- Adds a private internal function for checking if the given file/dir should be preserved or not
- Uses the original code as a guide

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1033,22 +1033,30 @@ function rocket_clean_cache_dir() {
  * @since 3.5.3 Bails if given dir should be preserved; replaces glob; optimizes.
  * @since 1.0
  *
- * @param string       $dir              File/Directory to delete.
- * @param array|string $dirs_to_preserve Optional. Dirs that should not be deleted.
+ * @param string      $dir              File/Directory to delete.
+ * @param array       $dirs_to_preserve Optional. Dirs that should not be deleted.
+ * @param string|null $preserve_pattern Optional. Regex pattern for directories to preserve.
  */
-function rocket_rrmdir( $dir, $dirs_to_preserve = '' ) {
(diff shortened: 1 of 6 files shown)


inc/functions/files.php
```diff
@@ -1050,8 +1057,8 @@ function rocket_rrmdir( $dir, array $dirs_to_preserve = [] ) {
 	 *
 	 * @since 1.1.0
 	 *
-	 * @param string $dir              File/Directory to delete.
-	 * @param array  $dirs_to_preserve Directories that should not be deleted.
+	 * @param string       $dir              File/Directory to delete.
+	 * @param array|string $dirs_to_preserve Directories that should not be deleted.
 	 */
 	do_action( 'before_rocket_rrmdir', $dir, $dirs_to_preserve ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
```
hellofromtonya, line 1063: @Tabrisrp Now that it's a regex pattern string (instead of an array), will it break something else?

Tabrisrp: I don't think so, the actions are not used in the plugin. The other function using `rocket_rrmdir` with the second argument is `rocket_clean_user`, which uses the same format as `rocket_clean_domain`, so that should be working too

hellofromtonya: Just to make sure, I reverted to keep that array and then created a new variable to hold the regex pattern. It's overkill. But at least it won't break anything that's hooked into that event. 2d33aff163d04f4e4c2160c420566b497a5af42d

PR #2532, opened 2020-04-09
Closes #2425 - Rewrite CSS files when CDN is enabled or CNAME changes

### LINKED ISSUE #2425: URLs in CSS files aren't rewritten as soon as the CDN is enabled
When the CDN feature is enabled, we rewrite URLs of static assets that are in the source code of the page and URLs in CSS files, e.g. of background images.

The first time one enables the CDN, the minified/combined files aren't cleared. So, the URLs in there are not rewritten and the images aren't downloaded from the CDN, but from the origin server.

If the cache is cleared, using the admin menu button, the URLs in the newly created files are properly rewritten.

**To Reproduce**

1. Add a `background-image:url(https://eaxmple.com/image.pnp)` in the theme's style.css file.
2. Enable WP Rocket's CSS minification feature.
3. Note down the filename of the CSS file.
4. Enable WP Rocket's CDN feature.
5. The filename from step 3 will still be the same and the image's URL will not have been rewritten to the CDN.
6. Clear WP Rocket's path and check the CSS filename. It will have changed, as expected, and the URL of the image will point to the CDN.

**Expected behavior**
Rewrite the CSS files URLs, as soon as the option is enabled.

**Possible Solution**
Clear the minified files when the CDN option is enabled/disabled.

**Additional context**
An edge case would be if someone disables the CDN and then cancel their subscription.
The CSS files URLs' will still point to the CDN and this will cause `errors 404`.

Clearing the cache will fix this, but we can avoid this in the first place.

**Related ticket**
https://secure.helpscout.net/conversation/1101999708/149719/

**Slack Thread**
https://wp-media.slack.com/archives/C43T1AYMQ/p1583602741148200?thread_ts=1583602533.148100

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
Cleans the CSS files also on CNAME change.

Introduces : **AdminSubscriber & AdminServiceProvider** 🎉 

**Reproduce the issue** ✅ 
Reproduced the issue on local

**Identify the root cause** ✅ 
These are the lines :
https://github.com/wp-media/wp-rocket/blob/fbb9380bcb79bc617ad5bfa27c513da1c44a5b78/inc/admin/options.php#L78-L87
Mainly this condition is the issue:
`( isset( $oldvalue['cdn'] ) && ! isset( $value['cdn'] ) || ! isset( $oldvalue['cdn'] ) && isset( $value['cdn'] )`

**Scope a solution** ✅ 
The solution will be fix that condition by switching to $oldvalue['cdn'] !== $value['cdn'].

TODO:
- [x] Unit Tests
- [x] Integration Tests
- [x] Unit Tests for regenerate_minify_css_key
- [x] Integration  Tests for regenerate_minify_css_key


inc/Engine/Optimization/Minify/CSS/AdminSubscriber.php
```diff
@@ -0,0 +1,48 @@
...
+	 */
+	public function clean_minify( $old_value, $value ) {
+		// Purge all minify cache files.
+		if ( ( isset( $old_value['minify_css'], $value['minify_css'] ) && $old_value['minify_css'] !== $value['minify_css'] )
+			||
+			( isset( $old_value['exclude_css'], $value['exclude_css'] ) && $old_value['exclude_css'] !== $value['exclude_css'] )
+			||
+			( isset( $old_value['cdn'], $value['cdn'] ) && $old_value['cdn'] !== $value['cdn'] )
+			||
+			( isset( $old_value['cdn_cnames'], $value['cdn_cnames'] ) && $old_value['cdn_cnames'] !== $value['cdn_cnames'] )
```
hellofromtonya, line 43: I think we should use `array_key_exists()` instead of `isset()`. 

Why? `isset` returns `false` when (a) the key does not exist _or_ (b) the value for that element is `null`. 

In this code, _the `isset` serves to validate that the key exists_. Then the `!==` checks _if_ the old and new values are different.

inc/admin/options.php
```diff
@@ -75,20 +75,23 @@ function rocket_after_save_options( $oldvalue, $value ) {
...
+		||
+		( ! isset( $oldvalue['cdn'] ) && isset( $value['cdn'] ) )
+	) {
 		rocket_clean_minify( 'js' );
 	}
 
 	// Purge all cache busting files.
 	// phpcs:ignore WordPress.Security.NonceVerification.Missing
-	if ( ! empty( $_POST ) && ( $oldvalue['remove_query_strings'] !== $value['remove_query_strings'] ) ) {
+	if ( ! empty( $_POST ) && ( isset( $oldvalue['remove_query_strings'], $value['remove_query_strings'] ) && $oldvalue['remove_query_strings'] !== $value['remove_query_strings'] ) ) {
```
hellofromtonya, line 94: use `array_key_exists()` instead of `isset`

inc/admin/options.php
```diff
@@ -75,20 +75,23 @@ function rocket_after_save_options( $oldvalue, $value ) {
...
 
-	// Purge all minify cache files.
 	// phpcs:ignore WordPress.Security.NonceVerification.Missing
-	if ( ! empty( $_POST ) && ( $oldvalue['minify_css'] !== $value['minify_css'] || $oldvalue['exclude_css'] !== $value['exclude_css'] ) || ( isset( $oldvalue['cdn'] ) && ! isset( $value['cdn'] ) || ! isset( $oldvalue['cdn'] ) && isset( $value['cdn'] ) ) ) {
-		rocket_clean_minify( 'css' );
-	}
-
-	// phpcs:ignore WordPress.Security.NonceVerification.Missing
-	if ( ! empty( $_POST ) && ( $oldvalue['minify_js'] !== $value['minify_js'] || $oldvalue['exclude_js'] !== $value['exclude_js'] ) || ( isset( $oldvalue['cdn'] ) && ! isset( $value['cdn'] ) || ! isset( $oldvalue['cdn'] ) && isset( $value['cdn'] ) ) ) {
+	if ( ! empty( $_POST )
```
hellofromtonya, line 79: `! empty( $_POST )` is used here and the next 2 conditional expressions. That's redundant and little bit slower. Instead, use it once and pull all 3 conditionals under it.

inc/admin/options.php
```diff
@@ -75,20 +75,23 @@ function rocket_after_save_options( $oldvalue, $value ) {
...
+		&&
+		( isset( $oldvalue['minify_js'], $value['minify_js'] ) && $oldvalue['minify_js'] !== $value['minify_js'] )
+		||
+		( isset( $oldvalue['exclude_js'], $value['exclude_js'] ) && $oldvalue['exclude_js'] !== $value['exclude_js'] )
+		||
+		( isset( $oldvalue['cdn'] ) && ! isset( $value['cdn'] ) )
+		||
+		( ! isset( $oldvalue['cdn'] ) && isset( $value['cdn'] ) )
```
hellofromtonya, line 87: use `array_key_exists()` instead of `isset`

inc/Engine/Optimization/Minify/CSS/AdminSubscriber.php
crystinutzaa: But what if: $old_value or $value are NULL. In this case with array_key_exists() a warning is displayed: `WARNING array_key_exists() expects parameter 2 to be array, null given` .
With array_key_exists() we need to check that also $value & $old_values are arrays.
I am pretty sure that under Integration tests these cases will occur.
So, do we really want to change the isset to array_key_exists?

inc/admin/options.php
Tabrisrp: I'm not sure why this check for `$_POST` was there in the first place (it's part of the very old code we have).

It is preventing the code inside the conditional from running if the update was not triggered from the settings page, but why would we want that? It should also be running if we call `update_option()` directly.

Does it make sense to you?

crystinutzaa: Yes. I confirm that `$_POST` is from old code in there and can be safely removed. I kept it because it was out of this task scope. However, this condition should be moved in the Admin Subscriber from Engine/Optimization/Minify/JS when it will be revisited. 

If we decide to remove the POST i recommend to move it now in it's AdminSubscriber and write Unit & Integration tests. It should not take too much time.

inc/Engine/Optimization/Minify/CSS/AdminSubscriber.php
hellofromtonya: That's a valid point. What if either is not an array? You can add a guard clause to check _or_ expect that both must be an array by type hinting.

crystinutzaa: I think a guard clause will do the trick in this case.
Will give it a try.

tests/Fixtures/inc/Engine/Optimization/Minify/CSS/AdminSubscriber/cleanMinify.php
```diff
@@ -0,0 +1,79 @@
...
+				'exclude_css' => [],
+				'cdn'         => false,
+				'cdn_cnames'  => [],
+			],
+			'should_run' => true,
+		],
+		'shouldCleanMinifyExcludeCSS'      => [
+			'value'     => [
+				'minify_css'  => true,
+				'exclude_css' => [ '.css_to_exclude' ],
```
Tabrisrp, line 54: This is not blocking, but it might become an unexpected issue later: this field expects relative path(s) to file(s), not CSS content

crystinutzaa: Fixed. Thank you for spotting this issue :)

Tabrisrp:
@wp-media/dev-qa ready for QA

arunbasillal:
QA test:

- Reproduced the issue using `master` branch 👍 
- On the fix branch, when I enable CDN, it didn't change the name of the minified file ❌ 

@crystinutzaa please note - https://youtu.be/qezRup787Q0

Cristina, not sure if this was already considered, but please consider a case where the CNAME field is already filled in but the option is disabled first and the user enables it - https://jmp.sh/akxOLUt
^ In this case too, the CSS files should be regenerated.

arunbasillal:
@Tabrisrp I believe you will be adding or modifying the tests since Cristina is on vacation. Please let me know when this is ready for QA

arunbasillal:
Retested this: 

- Enabling and Disabling CDN option regenerates the minifed file with a different name 👍 
- So does changing the CNAME 👍 
- Contents of the minified file is updated correctly 👍 

> Not sure if this was already considered, but please consider a case where the CNAME field is already filled in but the option is disabled first and the user enables it - https://jmp.sh/akxOLUt
^ This was tested and works as expected. Minifed file is regenerated 👍 

Discovered a super edge case. If the `Enable CDN` option is disabled and if you change and save the CNAME, the minified file is still regenerated unnecessarily. While this isn't a deal breaker, it's something we can improve since we know about it - https://jmp.sh/GjU97FQ (cc: @Tabrisrp)

hellofromtonya:
@Tabrisrp Please review the changes I did. Thanks

Tabrisrp:
@hellofromtonya Looks good, thank you for the additional improvements 👍

hellofromtonya:
@wp-media/dev-qa This PR has changes since the last QA. It also fixes the edge case that @arunbasillal identified. Please re-do the QA.

PR #2544, opened 2020-04-15
Fixes #2211 Prevent false positives on cron status notice

### LINKED ISSUE #2211: Fixing false positives on CRON status detection
This is a copy paste from here for better issue tracking and organisation - https://github.com/wp-media/wp-rocket/issues/1472#issuecomment-527544891

**With the 3.3.7 update, here is what we did:** 

- An admin notice is displayed if a scheduled event missed it's schedule time. 
- This will accommodate for WordPress installations that use server cron as well. 
- The notice displayed on WP Rocket dashboard can be dismissed and will not come back once dismissed. 
- This is basically what WordPress health check tool uses as well

**False positives**

At the time of writing, we have 9 false positives reported where customers saw this notice even when cron was working correctly. This could be because of an edge case. 

- Say an event is scheduled for 15:02
- Server cron is set to run every 5 minutes.
- It runs at 15:00
- Customer visits the dashboard at 15:03 and the notice will be displayed because the even scheduled for 15:02 is missed and will only be done at 15:05 when the server cron runs again. 

**Next Steps**

We will continue to monitor tickets about this and make a decision about whether to make changes in a future update.

**Steps to Replicate:** 

This is a tricky one to replicate. After multiple attempts, here is what worked: 

- Add `define('DISABLE_WP_CRON', 'true');` to `wp-config.php`
- Set a schedule for database optimization and save settings (by clicking `Optimize`). 
- The notice should show up immediately - https://jmp.sh/dWafjWY
- Here is proof of the issue being replicated - https://youtu.be/qqyzUJsRO4o

Now in my case, server CRON was not setup. But the notice doesn't wait for a ping from server cron to show up. 

**Related conversations:**

- https://wp-media.slack.com/archives/C08N8J6VC/p1567075962002300

**Proposed Idea:** 

- It looks like in the current approach, WP Rocket looks for a missed schedule. 
- If the theory in the section titled "False Positives" (above) is possible, then our current approach will obviously fail. 
- What if we add a buffer? (say 60 minutes) before determining if a schedule is missed or not?

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

Tabrisrp:
Fixes #2211 

Add a delay on the cron events check to prevent false positives on the cron status notice.

- [x] Create new class `Admin\Warnings`
- [x] Deprecate `rocket_warning_cron()`
- [x] Write unit tests for `missed_cron()`
- [x] Write integration tests for `missed_cron()`
- [ ] QA


inc/Engine/Admin/HealthCheck.php
```diff
@@ -0,0 +1,120 @@
...
+			return;
+		}
+
+		$events = [
+			'rocket_purge_time_event'                      => 'Scheduled Cache Purge',
+			'rocket_database_optimization_time_event'      => 'Scheduled Database Optimization',
+			'rocket_database_optimization_cron_interval'   => 'Database Optimization Process',
+			'rocket_preload_cron_interval'                 => 'Preload',
+			'rocket_critical_css_generation_cron_interval' => 'Critical Path CSS Generation Process',
+		];
```
Screenfeed, line 73: Shouldn’t we translate the descriptions?
```suggestion
		$events = [
			'rocket_purge_time_event'                      => __( 'Scheduled Cache Purge', 'rocket' ),
			'rocket_database_optimization_time_event'      => __( 'Scheduled Database Optimization', 'rocket' ),
			'rocket_database_optimization_cron_interval'   => __( 'Database Optimization Process', 'rocket' ),
			'rocket_preload_cron_interval'                 => _x( 'Preload', 'noun', 'rocket' ),
			'rocket_critical_css_generation_cron_interval' => __( 'Critical Path CSS Generation Process', 'rocket' ),
		];
```

inc/deprecated/3.5.php
```diff
@@ -792,3 +792,81 @@ function rocket_varnish_http_purge( $url ) {
...
+ * Display a warning notice if WP Rocket scheduled events are not running properly
+ *
+ * @since 3.5.4 deprecated
+ * @since 3.3.7
+ * @author Remy Perona
+ *
+ * @return void
+ */
+function rocket_warning_cron() {
+	_deprecated_function( __FUNCTION__ . '()', '3.5.4', 'WP_Rocket\Admin\Warnings::missed_cron()' );
```
Screenfeed, line 805:
```suggestion
	_deprecated_function( __FUNCTION__ . '()', '3.5.4', 'WP_Rocket\Engine\Admin\HealthCheck::missed_cron()' );
```

hellofromtonya:
@Tabrisrp Please review the changes I made.

inc/admin/ui/notices.php
```diff
@@ -987,7 +909,7 @@ function rocket_notice_html( $args ) {
 		<p>
 			<?php echo $args['action']; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>
 			<?php if ( $args['dismiss_button'] ) : ?>
-			<a class="rocket-dismiss" href="<?php echo esc_url( wp_nonce_url( admin_url( 'admin-post.php?action=rocket_ignore&box=' . $args['dismiss_button'] ), 'rocket_ignore_' . $args['dismiss_button'] ) ); ?>"><?php esc_html_e( 'Dismiss this notice.', 'rocket' ); ?></a>
+			<a class="rocket-dismiss" href="<?php echo wp_nonce_url( admin_url( 'admin-post.php?action=rocket_ignore&box=' . $args['dismiss_button'] ), 'rocket_ignore_' . $args['dismiss_button'] ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>"><?php esc_html_e( 'Dismiss this notice.', 'rocket' ); ?></a>
```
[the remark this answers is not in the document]
Tabrisrp: `wp_non_url` already returns the escaped URL with `esc_html()`, so I figured it's not necessary to escape twice

inc/Engine/Admin/HealthCheck.php
```diff
@@ -117,4 +96,36 @@ public function missed_cron() {
...
 	}
+
+	/**
+	 * Checks if health check should not run.
+	 *
+	 * @since 3.5.4
+	 *
+	 * @return bool true when should not do health check; else, false.
+	 */
+	protected function do_not_check() {
```
Tabrisrp, line 108: @hellofromtonya what do you think of using a name like `should_check` instead? And of course inverting the returns inside the method.

I tend to prefer avoiding names with a negation, as it makes it more difficult to read for me when you have to use it with `!` for example

inc/admin/ui/notices.php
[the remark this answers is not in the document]
hellofromtonya: You're correct. 👍 I should have checked Core first before asking. Doh.

inc/Engine/Admin/HealthCheck.php
hellofromtonya: @Tabrisrp That's fine. You'll need to invert the logic in the method.

Tabrisrp:
@wp-media/dev-qa Ready for QA

New delay is 5 minutes with default WP cron, 1 hour when WP cron is disabled

piotrbak:
Starting the QA

Issue #2543, opened 2020-04-15
Preload For Fonts


arunbasillal:
## What are we proposing to do? 

Font files that are requested in a CSS file or later in the page load can be downloaded faster by preloading them. With this update, we will provide the necessary tools for our users to accomplish this. 

## Why are we doing this? 

It is common practice to reference font files in CSS files. In this case, the browser has to download and parse the CSS files before it can download the fonts. Modern browsers can download multiple files in parallel. 

With this feature, our users will be able to “tell” the browser about these hidden fonts. The feature also informs the browser that files are indeed fonts. Modern browsers handle fonts very well 

This feature will also satisfy the “Preload key requests” suggestion of PageSpeed. 

![image4](https://user-images.githubusercontent.com/5751474/79350236-d07c2a00-7f54-11ea-9c05-1d1f9b297add.png)

## Benchmark Tests

Jonathan did benchmarks for this feature on https://wp-rocket.me/ with positive results. Here is the waterfall

**No Preload:** 

![no-preload_LI](https://user-images.githubusercontent.com/5751474/79349894-6bc0cf80-7f54-11ea-802e-5b3180c3e1eb.jpg)
Full scale: https://jmp.sh/A336ebO

**With Preload:** 

![with-font-preload_LI](https://user-images.githubusercontent.com/5751474/79349937-78452800-7f54-11ea-8d8a-01304c479288.jpg)
Full scale: https://jmp.sh/T8Gb6Ko

Note how the fonts take considerably lower time to load. 

Another example is depicted here: https://developers.google.com/web/tools/lighthouse/audits/preload
User expectations and experience. How will we do this feature?

## UI Proposal

Feature location: **Preload tab**
Section title: **Preload Fonts**
Section description: **Improves performance by helping browsers discover fonts in CSS files. More info** (Thanks Lucy https://github.com/wp-media/wp-rocket/issues/2543#issuecomment-615335738)

Text area title: **Fonts to preload**
Text area description: **Specify urls of the font files to be preloaded (one per line).**
Text area guideline (in green): **The domain part of the URL will be stripped automatically.
Allowed font extensions: ttf, otf, woff, woff2, svg.**

![preload fonts](https://user-images.githubusercontent.com/5751474/79944201-892bf700-8488-11ea-806f-df3e12b18806.png)

**A related change is also recommended for cache preloading**. Currently the section title simply says “Preload”. To avoid any confusion, let’s rename it to “**Preload Cache**”.

[...]

PR #2554, opened 2020-04-20
Preload for Fonts: UI (1/3)

Screenfeed:
Sub-cards for epic #2543.

This card is focused on the UI changes to make to add the option on the settings page.

#### Admin\Settings\Page
- [x] `preload_section()`: update `menu_description` in `add_page_section()`.
- [x] `preload_section()`: update `title` in `add_settings_sections()` for `preload_section`.
- [x] `preload_section()`: add new  item array  in `add_settings_fields()` for the new option.

#### Q&A
- [x] Q&A

Closes #2545.

PR #2555, opened 2020-04-20
Preload for fonts: option management (2/3)

### LINKED ISSUE #2546: Preload for fonts: option management (2/3)
Sub-cards for epic #2543.

This card is for the new option management tasks.

#### admin/upgrader.php
- [x] Add option `preload_fonts` to `rocket_first_install()`
- [x] Update tests

#### Admin\Settings\Settings
- [x] Add sanitization for the option in `sanitize_callback()`
- [x] Update tests

Screenfeed:
Sub-cards for epic #2543.

This card is for the new option management tasks.

#### admin/upgrader.php
- [x] Add option `preload_fonts` to `rocket_first_install()`
- [x] Update tests

#### Admin\Settings\Settings
- [x] Add sanitization for the option in `sanitize_callback()`
- [x] Update tests

#### Q&A
- [ ] Q&A

Closes #2546.


inc/classes/admin/settings/class-settings.php
```diff
@@ -37,6 +37,16 @@ class Settings {
...
 
+	/**
+	 * Host domain.
+	 *
+	 * @since 3.6
+	 * @see   $this->get_host()
+	 *
+	 * @var string|bool
+	 */
+	private $host;
```
hellofromtonya, line 48: We've touched this class. Let's move it to our new architecture, rename, and deprecate this class file.

inc/classes/admin/settings/class-settings.php
```diff
@@ -543,4 +565,69 @@ function( $url ) {
...
+
+		$ext     = strtolower( pathinfo( $file, PATHINFO_EXTENSION ) );
+		$formats = [
+			'eot',
+			'otf',
+			'ttf',
+			'svg',
+			'woff',
+			'woff2',
+		];
```
hellofromtonya, line 604: Move this array to a property. Why? Avoids creating the array each time this callback is invoked, which results in a micro-optimization for execution time and memory space.

inc/classes/admin/settings/class-settings.php
```diff
@@ -543,4 +565,69 @@ function( $url ) {
...
+	 * Get site’s host domain.
+	 *
+	 * @since 3.6
+	 *
+	 * @return string|bool The host. False on failure.
+	 */
+	private function get_host() {
+		if ( isset( $this->host ) ) {
+			return $this->host;
+		}
```
hellofromtonya, line 623: @Screenfeed What do you think of initializing the host property in the `__construct()`? Do you see any potential problems?

Why? Simplifies + micro-optimization. It'll remove the guard clause and allow the callback to use the property instead of calling this method each time.

tests/Integration/inc/admin/rocketFirstInstall.php
```diff
@@ -0,0 +1,114 @@
...
+			'minify_css_key'   => '',
+			'minify_js_key'    => '',
+		];
+
+		rocket_first_install();
+
+		$options = get_option( $this->option_name );
+
+		$this->assertIsArray( $options );
+		$this->assertTrue( ! empty( $options['secret_cache_key'] ) && ! empty( $options['minify_css_key'] ) && ! empty( $options['minify_js_key'] ) );
```
hellofromtonya, line 97: Make separate assertions:

```suggestion
		$this->assertArrayHasKey( 'secret_cache_key', $options );
		$this->assertNotEmpty( $options['secret_cache_key'] ); 
		$this->assertArrayHasKey( 'minify_css_key', $options );
		$this->assertNotEmpty( $options['minify_css_key'] ); 
		$this->assertArrayHasKey( 'minify_js_key', $options );
		$this->assertNotEmpty( $options['minify_js_key'] ); 
```

tests/Unit/inc/admin/rocketFirstInstall.php
```diff
@@ -0,0 +1,53 @@
...
+		Functions\expect( 'add_option' )
+			->once()
+			->with( 'wp_rocket_settings', $expected );
+		Functions\expect( 'rocket_dismiss_box' )
+			->once()
+			->with( 'rocket_warning_plugin_modification' );
+
+		rocket_first_install();
+
+		$this->assertTrue( Filters\applied( 'rocket_first_install_options' ) === 1 );
```
hellofromtonya, line 47:
```suggestion
		$this->assertSame( 1, Filters\applied( 'rocket_first_install_options' ) );
```

inc/classes/admin/settings/class-settings.php
Screenfeed: @hellofromtonya I did that for the same reason, micro-optimization.
Generally speaking, I try to never perform operations in the constructor: doing this in the constructor means we're doing this even if we don't need it. While doing this in the getter method is like lazy-loading the operation.
- In the constructor: the whole operation is done only once, but on every load, even if not needed.
- In the getter method: the whole operation is done only once again, only if needed, but requires one or several `isset()` when called.

Putting it in the constructor seems counter-productive to me but I may be wrong, you know better than me in this domain. What do you think?

inc/classes/admin/settings/class-settings.php
Screenfeed: Done in https://github.com/wp-media/wp-rocket/pull/2555/commits/ed57696b00a089b73997a90df3f396dab7e3b3e9.

inc/classes/admin/settings/class-settings.php
Screenfeed: Same file location issue than https://github.com/wp-media/wp-rocket/pull/2554#issuecomment-617475684. Let’s decide what to do.

Tabrisrp: `WP_Rocket\Engine\Settings\Settings`

Screenfeed: Done in https://github.com/wp-media/wp-rocket/pull/2555/commits/908182cf191f3dcef00e78e50f59212c7293fb00.

licence-data.php
```diff
@@ -1,18 +0,0 @@
-<?php
```
Tabrisrp, line 1: oopsie

inc/classes/admin/settings/class-settings.php
hellofromtonya: @Screenfeed Your assessment and reasoning is good and valid. It has to be weight against how often the method is called in its current structure. Then select the implementation which causes the less impact to performance (speed and memory). In this case, I suspect you are correct.

inc/Engine/Admin/Settings/Settings.php
```diff
@@ -350,6 +376,18 @@ public function sanitize_callback( $input ) {
...
 		}
 
+		// Option : fonts to preload.
+		if ( ! empty( $input['preload_fonts'] ) ) {
+			if ( ! is_array( $input['preload_fonts'] ) ) {
+				$input['preload_fonts'] = explode( "\n", $input['preload_fonts'] );
+			}
+			$input['preload_fonts'] = array_map( [ $this, 'sanitize_font' ], $input['preload_fonts'] );
+			$input['preload_fonts'] = array_filter( $input['preload_fonts'] );
+			$input['preload_fonts'] = array_unique( $input['preload_fonts'] );
```
Tabrisrp, line 386: I think it would be better if this part is in the `sanitize_font` callback, then we can do a ternary instead:

`$input['preload_fonts'] = ! empty( $input['preload_fonts'] ) ? $this->sanitize_font( $input['preload_fonts'] ) : [];`

Screenfeed: Done in https://github.com/wp-media/wp-rocket/pull/2555/commits/96953434ee3007a2a22f1127a3daae1bb4e1dc45.

Screenfeed:
Ready for a new review.

inc/classes/admin/settings/class-settings.php
Screenfeed: OK, thank you Tonya.

Tabrisrp:
@arunbasillal Ready for QA

Now you have the UI, and when entering data in the option, it should be saved and sanitized correctly to only preserve the correct values

Issue #2549, opened 2020-04-16
Excessive terms clearing during the partial purge


piotrbak:
**Describe the bug**
We're clearing taxonomy terms recursively during the partial purge.

**To Reproduce**
1. Assign a Woo product to the category A
2. Create categories B, C, D that are children of the category A
3. Make sure that the product is not present in the child categories
4. Update the product
5. We'll pass `domain.com/a/` to the `rocket_rrmdir` via this part of code:
https://github.com/wp-media/wp-rocket/blob/00cf7a1c5e0b2e7354b0aa76ca86603ee4420eae/inc/common/purge.php#L138
6. It will also clear  `domain.com/a/b/` `domain.com/a/c/` `domain.com/a/d/` and all further nested categories where the product is not present.

**Expected behavior**
Make sure that child terms are not cleared if the post is not added there.

**Additional context**
It could be a pretty big problem if products are sharing category permalinks like following:
`domain.com/product/` - Category base permalink
`domain.com/product/%product_cat%/` - Product permalink
The above will cause the purge of all products that are present in the passed category term. However, Remy mentioned that the above example is an edge case:
https://wp-media.slack.com/archives/C43T1AYMQ/p1585602235046800?thread_ts=1585337918.031600&cid=C43T1AYMQ

Related ticket:
https://secure.helpscout.net/conversation/1113270980/152492?folderId=2415573

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

Tabrisrp:
**Reproduce the issue** ✅ 
**Identify the root cause** ✅ 
Piotr handled those already 👍 

**Scope a solution** ❌
This one is tough, because the system is not currently designed to handle directories to preserve on that part.

We could only delete the files in the corresponding cache folder, but that would ignore any special folder like the one created when there is cached query strings, so that's not ok.

@hellofromtonya maybe we can discuss together how best to approach that

hellofromtonya:
Good idea @Tabrisrp. Let's pull this into Slack and figure out the best way to handle this.

Tabrisrp:
**Scope a solution** ✅ 
Encapsulate getting the term(s) URL(s) and the appropriate files to clean in a private function dedicated to that. 

[...]

PR #2556, opened 2020-04-20
Improve (no glob) and add tests for rocket_clean_files().

hellofromtonya:
This PR does the following for `rocket_clean_files()`:

- replaces `glob()` with SPL
- improves the function for speed and memory usage
- adds unit and integration tests

Multiple cleaning functions will use the same iterator patterns. To reduce redundancy and increase reuse, this PR:

- moves the iterators to private internal functions
- updates `rocket_clean_domain()` to use the private functions

These private functions will become private methods when refactor and rearchitect.

## TODO

- [x] Validate current behavior with test data
- [x] Add unit and integration tests
- [x] Replace `glob`
- [x] Optimize
- [x] Move redundant code to reusable private/internal functions for reuse
- [ ] QA


hellofromtonya:
@wp-media/dev-qa This PR is ready for QA. Please make sure that `rocket_clean_files()` and `rocket_clean_domain()` still work as they did before, i.e. there should no change to its behavior.

hellofromtonya:
>🐛 However, when I call rocket_clean_domain() directly via functions.php, then all the folders in the wp-content/cache folder is deleted. This isn't right because other plugins can use the same folder and we shouldn't be deleting them. Seems like a bug.

@arunbasillal Can you share the code of how you called `rocket_clean_domain()` directly in the theme's `functions.php`? Was it within a callback or just in the root of the file? And do you have any i18n plugins installed?

vmanthos:
@hellofromtonya 

I had a similar case, which I asked about in the Level 2 channel:
https://wp-media.slack.com/archives/C43T1AYMQ/p1587554387007900

hellofromtonya:
I think the problem is not in `rocket_clean_domain()`, but rather, it's in a callback hooked into `'before_rocket_clean_domain'` which is fired in `rocket_clean_domain()`.  

@arunbasillal to confirm, please do the following:

- comment out line of code https://github.com/wp-media/wp-rocket/blob/improve/rocket_clean_files/inc/functions/files.php#L853
- then re-run your test with `rocket_clean_domain()` in your theme's `functions.php`

hellofromtonya:
@piotrbak This PR is ready for QA.

piotrbak:
Starting the QA

hellofromtonya:
>However, there's one thing to consider @hellofromtonya. When passing URLs to the rocket_clean_files, we are purging everything recursively. 

Yes, this is correct. It cleans all of the files from what is given. As you note below, this PRs behavior matches current and release release’s behavior.

>If we pass the homepage, whole cache will be cleared, if we pass the taxonomy terms URL, all posts from this category will be cleared (if using default pretty permalink setup- example.com/category/post). This was also happening in the previous versions. It's similar situation to this one: #2549

@piotrbak Let’s pull the discussion either into Slack or an issue. It’ll be good to discuss.

PR #2563, opened 2020-04-21
Preload for Fonts: front (3/3)

### LINKED ISSUE #2547: Preload for Fonts: front (3/3)
Sub-card for epic #2543.

#### Engine\Preload
- [x] Add new class `Fonts` to `Engine\Preload`
- [x] Instantiate the class in `Engine\Preload\ServiceProvider`
- [x] Register the class alias in `Plugin::load()`

#### Engine\Preload\Fonts
- [x] Add method `preload_fonts` as a callback for `wp_head`
- [x] Write tests for `preload_fonts`

Screenfeed:
Sub-card for epic #2543.

#### Engine\Preload
- [x] Add new class `Fonts` to `Engine\Preload`
- [x] Instantiate the class in `Engine\Preload\ServiceProvider`
- [x] Register the class alias in `Plugin::load()`

#### Engine\Preload\Fonts
- [x] Add method `preload_fonts` as a callback for `wp_head`
- [x] Write tests for `preload_fonts`

#### Q&A
- [x] Q&A

Closes #2547.


inc/Engine/Preload/Fonts.php
```diff
@@ -0,0 +1,107 @@
...
+	/**
+	 * Return an array of events that this subscriber wants to listen to.
+	 *
+	 * @since  3.6
+	 *
+	 * @return array
+	 */
+	public static function get_subscribed_events() {
+		return [
+			'wp_head' => [ 'preload_fonts' ],
```
Tabrisrp, line 44: This one is on me, verry sorry. I should have thought about this before but:

I think we should use `rocket_buffer` instead of `wp_head`. Because we:
- want the links only inserted on cached page. It's easier to handle this by using `rocket_buffer` as the hook
- want to make sure the preload links are early in the head, ideally before the inline critical CSS and the combined CSS files. I believe the loading impact wouldn't be as good if they are not first

We can use the `</title>` closing tag as the anchor to insert them.

Priority of the callback should be `20` or above.

inc/Engine/Preload/Fonts.php
```diff
@@ -0,0 +1,107 @@
...
+		];
+	}
+
+	/**
+	 * Add the required <link/> tags used to preload fonts.
+	 *
+	 * @since 3.6
+	 */
+	public function preload_fonts() {
```
Tabrisrp, line 53: 2 things here:
- The preload links should be inserted only if the page is cached
- The links should use the CDN URL(s) if CDN is enabled

inc/Engine/Preload/Fonts.php
```diff
@@ -0,0 +1,107 @@
...
+
+		$file = trim( $file );
+
+		if ( empty( $file ) ) {
+			return false;
+		}
+
+		$ext     = strtolower( pathinfo( $file, PATHINFO_EXTENSION ) );
+		$formats = [
+			'eot',
```
Tabrisrp, line 93:
```suggestion
```

Consistency with the backend sanitize

inc/Engine/Preload/Fonts.php
GeekPress: > The preload links should be inserted only if the page is cached
Before to lake this change, please wait our confirmation, we are discussing about that => https://github.com/wp-media/wp-rocket/issues/2543#issuecomment-617788770

Screenfeed: >The links should use the CDN URL(s) if CDN is enabled

Done in https://github.com/wp-media/wp-rocket/pull/2563/commits/1c5bd1acc576a024e57dc1731579d290a77011be.

inc/Engine/Preload/Fonts.php
```diff
@@ -0,0 +1,106 @@
...
+	public function preload_fonts() {
+		$fonts = $this->options->get( 'preload_fonts', [] );
+		$fonts = array_map( [ $this, 'sanitize_font' ], $fonts );
+		$fonts = array_filter( $fonts );
+
+		if ( empty( $fonts ) ) {
+			return;
+		}
+
+		$home = untrailingslashit( get_option( 'home' ) );
```
Tabrisrp, line 62:
```suggestion
		$site_url = get_rocket_parse_url( site_url() );
		$home   = $site_url['scheme'] . '://' . $site_url['host']:
```

How about this instead?

I understand the idea of using `get_option( 'home' )` directly instead of `home_url()` because of the filter, but I also feel `site_url()` is a good way to retrieve what we want. This is also what is used in other parts of the plugin.

inc/Engine/Preload/Fonts.php
GeekPress: Whatever the preload tags for fonts are adding, it doesn't have an impact on their performance.

Here a test when the CSS files is loaded before the preload:
https://gtmetrix.com/reports/wp-rocket.me/abBweyMK

Another test with the CPCSS + CSS file loaded before the preload:
https://gtmetrix.com/reports/wp-rocket.me/nnDXH7q4

Then when the preload is added before the CSS file:
https://gtmetrix.com/reports/wp-rocket.me/VKIdXuLR

The loading time for fonts are always the same.

Why?
The CSS file is loaded, but the fonts aren't used yet.

Tabrisrp: Thank you for the tests, I wasn't sure that was the case or not 👍

inc/Engine/Preload/Fonts.php
Screenfeed: `get_option( 'home' )`, `site_url()`, `home_url()`, all three are used in the plugin and `home_url()` is the most used, but this must not be one of the criteria for deciding.

So here are my thoughts, you may have right about using `site_url()`:
1- Among other places, fonts may be located in themes or plugins.
2- `WP_PLUGIN_URL` from WP’s core:
```php
if ( ! defined( 'WP_CONTENT_URL' ) ) {
	define( 'WP_CONTENT_URL', get_option( 'siteurl' ) . '/wp-content' );
}

// ...

if ( ! defined( 'WP_PLUGIN_URL' ) ) {
	define( 'WP_PLUGIN_URL', WP_CONTENT_URL . '/plugins' );
}
```

So, `WP_PLUGIN_URL` uses `WP_CONTENT_URL`, which uses `get_option( 'siteurl' )`.
3- `get_site_url()` also uses `get_option( 'siteurl' )` obviously.

So I think `site_url()` is the safest one to use.

Screenfeed: Done in https://github.com/wp-media/wp-rocket/pull/2563/commits/242860e25cd7f02c2300a17bef508d3165438c24.

inc/Engine/Preload/Fonts.php
GeekPress: Please keep the actual behavior to also apply the loading time improvement on uncached version.

Screenfeed:
>I'm thinking we could have a filter somewhere in this for the preload fonts array, if we (or a 3rd party) want to programatically add fonts to this at some point

@Tabrisrp Agree. It would be the opportunity to create a single method that returns the extensions. The extensions list will be used in 3 places:
- `WP_Rocket\Engine\Admin\Settings\Page`.
- `WP_Rocket\Engine\Admin\Settings\Settings`.
- `WP_Rocket\Engine\Preload\Font`.

If we go this way, `WP_Rocket\Engine\Preload\Font` seems the best place. Then we could use the filter `'rocket_container'` to get the instance in the two other classes.
Any better idea?

inc/Engine/Preload/Fonts.php
Screenfeed: @Tabrisrp So I guess there is nothing to change here, we don’t need to use `rocket_buffer`.

inc/Engine/Preload/Fonts.php
```diff
@@ -0,0 +1,107 @@
...
+			return;
+		}
+
+		$base_url = get_rocket_parse_url( site_url() );
+		$base_url = "{$base_url['scheme']}://{$base_url['host']}";
+
+		foreach ( array_unique( $fonts ) as $font ) {
+			printf(
+				"\n<link rel=\"preload\" as=\"font\" href=\"%s\" crossorigin>",
+				esc_url( get_rocket_cdn_url( $base_url . $font ) )
```
Tabrisrp, line 68: I remember we talked about it before, but I don't remember in what context: what's our stance on using `get_rocket_cdn_url()` vs injecting the CDN class in the constructor and using the `rewrite_url()` method?

hellofromtonya: I'd vote for injecting an instance of the CDN into the constructor and then using its `rewrite_url()` method. 

Why? More performant.

- `get_rocket_cdn_url()` gets the container instance, then the CDN instance from the container, and then fires the `rewrite_url()` method. 
- By directly injecting the instance of the CDN, we skip getting the container and CDN instance over and over again as the `foreach` iterates the fonts.
- Benefits: less memory and execution processing time.

Screenfeed:
@Tabrisrp Oh, you meant the fonts directly. Sorry I misunderstood, I thought you were talking about the font formats.

Tabrisrp:
@arunbasillal Ready for QA. With this PR the full feature is available, backend and frontend.

When adding fonts in the options, they should be added in the `<head>`.

The fonts URLs should use the CDN URL(s) when the CDN option is enabled and cnames exist.

PR #2565, opened 2020-04-21
Tests and optimizes: rocket_add_url_protocol()

hellofromtonya:
`rocket_add_url_protocol()`:

- Adds unit tests
- Adds integration tests
- Optimizes by:
    - flipping http/https checks into guard clause
    - returning URL from `set_url_scheme()` instead of assigning to variable and then returning

Tests were done _before_ the code was optimized to ensure the changes did not change the behavior.

### DIFF
--- a/inc/functions/formatting.php
+++ b/inc/functions/formatting.php
@@ -259,22 +259,29 @@ function rocket_remove_url_protocol( $url, $no_dots = false ) {
 }
 
 /**
- * Add HTTP protocol to an url that does not have
+ * Add HTTP protocol to an url that does not have.
  *
  * @since 2.2.1
  *
  * @param string $url The URL to parse.
- * @return string $url The URL with protocol
+ *
+ * @return string $url The URL with protocol.
  */
 function rocket_add_url_protocol( $url ) {
-
-	if ( strpos( $url, 'http://' ) === false && strpos( $url, 'https://' ) === false ) {
-		if ( substr( $url, 0, 2 ) !== '//' ) {
-			$url = '//' . $url;
-		}
-		$url = set_url_scheme( $url );
+	// Bail out if the URL starts with http:// or https://.
+	if (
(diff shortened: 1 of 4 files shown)


inc/functions/formatting.php
```diff
@@ -259,22 +259,29 @@ function rocket_remove_url_protocol( $url, $no_dots = false ) {
 }
 
 /**
- * Add HTTP protocol to an url that does not have
+ * Add HTTP protocol to an url that does not have.
```
Tabrisrp, line 262:
```suggestion
 * Add HTTP protocol to an url that does not have it.
```

inc/functions/formatting.php
```diff
@@ -259,22 +259,29 @@ function rocket_remove_url_protocol( $url, $no_dots = false ) {
...
+ * @return string $url The URL with protocol.
  */
 function rocket_add_url_protocol( $url ) {
+	// Bail out if the URL starts with http:// or https://.
+	if (
+		strpos( $url, 'http://' ) !== false
+		||
+		strpos( $url, 'https://' ) !== false
+	) {
+		return $url;
```
Tabrisrp, line 277: The URL should still go through `set_url_scheme()` to make sure it uses the correct scheme depending on the value of `is_ssl()`

hellofromtonya: It doesn't currently. Currently, it only goes through `set_url_scheme()` to get the `http:` or `http:` protocol if the given URL doesn't have either.

hellofromtonya: @Tabrisrp Are you thinking this is a bug?

Tabrisrp: oh yes you're right, I read the original wrong, sorry about that!

I think it's ok that way in fact, looking at where it's used in our codebase, there is no issue with it.

PR #2571, opened 2020-04-22
Fixes `rocket_clean_domain()` regex for domain in absolute path

hellofromtonya:
This PR improves the regex and fixes a problem when the domain name is in the absolute path before `cache/wp-rocket`.

## Root cause

When a website's absolute path has the domain name in it, the old regex matched. For example, let's say the domain is `example.org` and it's absolute path is:

```
/home/foo/example.org/wp-content/cache/wp-rocket/
```
Aha, that's not what we want. We want to target only the filesystem entries in the wp-rocket cache.

@arunbasillal Noted the problem here https://github.com/wp-media/wp-rocket/pull/2556#pullrequestreview-397835764

and @vmanthos noted the problem here https://github.com/wp-media/wp-rocket/pull/2556#issuecomment-617753769

## Testing the changes

- Changes have been validated on 2 customer sites that reported this problem.


inc/functions/files.php
```diff
@@ -837,13 +838,13 @@ function rocket_clean_domain( $lang = '' ) {
 	}
 
 	foreach ( $urls as $url ) {
-		$file = get_rocket_parse_url( $url );
+		$parsed_url = get_rocket_parse_url( $url );
 
 		if ( $url_no_dots ) {
-			$file['host'] = str_replace( '.', '_', $file['host'] );
+			$parsed_url['host'] = str_replace( '.', '_', $file['host'] );
```
Tabrisrp, line 844:
```suggestion
			$parsed_url['host'] = str_replace( '.', '_', $parsed_url['host'] );
```

Tabrisrp:
@wp-media/dev-qa Ready for QA, as soon as possible

piotrbak:
Starting QA

PR #2574, opened 2020-04-23
Fixes regex and optimizes rocket_clean_minify()

hellofromtonya:
## Risk of potential bug 🐛 

It's possible for a given extension to be part of the absolute path to the minify cache directory. For example, it could be part of the user account or domain name, such as `kb.css.example.org` or account name `foo.css`. As we learned in PR #2571, we need to prefix the regex with absolute path to the `cache/min` directory to avoid the potential risk of this problem.

In addition, this PR:

- Optimizes the function:
     - Bails out if there's no extension to clean
     - Calls `rocket_direct_filesystem()` 1x instead of 5x
     - Only updates the extensions variable if a string was passed into the function
- Makes it consistent with the other clean functions
- Makes the tests consistent with the other clean functions
- Adds unit tests

## TODO

- [x] Make consistent
- [x] Optimize
- [x] Add unit tests
- [x] QA

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -430,9 +430,10 @@ function set_rocket_wp_cache_define( $turn_it_on ) { // phpcs:ignore WordPress.N
  */
 function rocket_clean_minify( $extensions = [ 'js', 'css' ] ) {
 	$extensions = is_string( $extensions ) ? (array) $extensions : $extensions;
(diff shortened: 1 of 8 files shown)


hellofromtonya:
@arunbasillal Please go ahead and do the QA on this one. It's SPL implementations are now consistent with `rocket_clean_domain()` and `rocket_clean_files()`. Therefore, I don't anticipate code changes once Remy is available to review it.

PR #2587, opened 2020-04-27
Tests: rocket_generate_advanced_cache_file()

hellofromtonya:
- Adds unit and integration tests for `rocket_generate_advanced_cache_file()`.
- Adds the means to test the function by resetting the static function variable when testing.


inc/functions/files.php
```diff
@@ -86,21 +86,26 @@ function get_rocket_advanced_cache_file() { // phpcs:ignore WordPress.NamingConv
...
 	if ( $done ) {
 		return;
 	}
 	$done = true;
 
-	rocket_put_content( WP_CONTENT_DIR . '/advanced-cache.php', get_rocket_advanced_cache_file() );
+	rocket_put_content(
+		rocket_get_constant( 'WP_CONTENT_DIR' ) . '/advanced-cache.php',
+		get_rocket_advanced_cache_file()
+	);
```
hellofromtonya, line 108: @Tabrisrp What if something goes wrong when getting the content and/or creating/writing the file? Should we return the results of `rocket_put_content()`? And shouldn't we wait to set `$done = true` until we get a positive indication that the file was generated? Or do we want to throw an error if the result is not successful?

Tabrisrp: We never really considered that. I'm not sure it came up as an issue either, since we have a check and notice to make sure the file is writeable.

Ideally we should check and warn if there is any error indeed, same for the `$done`, but wouldn't that be outside of the scope of this PR?

hellofromtonya: >Ideally we should check and warn if there is any error indeed, same for the $done, but wouldn't that be outside of the scope of this PR?

Yes, it is outside the scope of this PR. But as I was here in the function, I became curious if we need to modify it's behavior in a new PR.

PR #2589, opened 2020-04-28
Closes #2520 Rewrite both http & https versions of the URLs to the CDN URL

### LINKED ISSUE #2520: CDN rewriting doesn't work when WP_HOME/WP_SITE_URL scheme is different than the site's URLs
When the `WP_HOME` is using `http` but the site's URLs are using `https`, e.g. there is an htaccess redirect from `http` to `https` CDN rewriting doesn't take place.

This is because of the `str_replace()` in the `rewrite_url` method of the `CDN` class:
https://github.com/wp-media/wp-rocket/blob/00cf7a1c5e0b2e7354b0aa76ca86603ee4420eae/inc/classes/CDN/CDN.php#L120

**Example:**
http://sandbox.onlinephpfunctions.com/code/1521df9f18dd8b11edf02e77f9994656078f6539

**To Reproduce**

1. Make sure the `WP_HOME` and `WP_SITE_URL` are using `http`.
2. Enable `http` to `https` redirection in the  `.htaccess` file.
3. Enable the CDN option.
4. Check the website. Static assets will not be rewritten to point to the CDN.

**Expected behavior**
Static assets should be pointing to the CDN.

**Additional context**

Using the `rocket_is_ssl_website()` function before the `string_replace()` resolves the issue:
```PHP
if( rocket_is_ssl_website() ) {
    $home_parts['scheme'] = 'https';
}
```
**Related ticket:**
https://secure.helpscout.net/conversation/1116358022/153068/

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

Tabrisrp:
Closes #2520

This PR also fixes an issue where the assets are not correctly rewritten when using i18n plugins with a subdomain or different domain configuration.

- [x] Update classes & tests to new architecture
- [x] Update `rewrite_url` to replace both http & https
- [x] Update tests
- [x] QA


inc/Engine/CDN/ServiceProvider.php
```diff
@@ -0,0 +1,38 @@
...
+	 * Registers the services in the container
+	 *
+	 * @return void
+	 */
+	public function register() {
+		$this->getContainer()->share( 'cdn', 'WP_Rocket\Engine\CDN\CDN' )
+			->withArgument( $this->getContainer()->get( 'options' ) );
+		$this->getContainer()->share( 'cdn_subscriber', 'WP_Rocket\Engine\CDN\Subscriber' )
+			->withArgument( $this->getContainer()->get( 'options' ) )
+			->withArgument( $this->getContainer()->get( 'cdn' ) );
```
hellofromtonya, line 36:
```suggestion
		$options = $this->getContainer()->get( 'options' );
		
		$this->getContainer()->share( 'cdn', 'WP_Rocket\Engine\CDN\CDN' )
			->withArgument( $options );
		$this->getContainer()->share( 'cdn_subscriber', 'WP_Rocket\Engine\CDN\Subscriber' )
			->withArgument( $options )
			->withArgument( $this->getContainer()->get( 'cdn' ) );
```

inc/Engine/CDN/Subscriber.php
```diff
@@ -118,7 +112,7 @@ public function rewrite_css_properties( $content ) {
 			return $content;
 		}
 
-		if ( ! $this->options->get( 'cdn' ) ) {
+		if ( ! $this->options->get( 'cdn', 0 ) ) {
```
hellofromtonya, line 115: This pattern repeats in the class. Let's move it to a method to encapsulate the knowledge of how to get the `'cdn'` option and its default.

tests/Fixtures/inc/Engine/CDN/CDN/rewrite.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+return [
+	'testDefaultSiteURL' => [
+		'site_url' => 'http://example.org',
+		'original' => file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/Engine/CDN/CDN/HTML/siteURL/original.html' ),
+		'expected' =>file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/Engine/CDN/CDN/HTML/siteURL/rewrite.html' ),
```
hellofromtonya, line 7:
```suggestion
		'expected' => file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/Engine/CDN/CDN/HTML/siteURL/rewrite.html' ),
```

tests/Fixtures/inc/Engine/CDN/CDN/rewrite.php
```diff
@@ -0,0 +1,14 @@
...
+	],
+	'testSiteURLWithPath' => [
+		'site_url' => 'http://example.org/blog',
+		'original' => file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/Engine/CDN/CDN/HTML/siteURLWithPath/original.html' ),
+		'expected' =>file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/Engine/CDN/CDN/HTML/siteURLWithPath/rewrite.html' ),
```
hellofromtonya, line 12:
```suggestion
		'expected' => file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/Engine/CDN/CDN/HTML/siteURLWithPath/rewrite.html' ),
```

tests/Fixtures/inc/Engine/CDN/Subscriber/rewrite.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+return [
+	'testDefaultSiteURL' => [
+		'site_url' => 'http://example.org',
+		'original' => file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/Engine/CDN/Subscriber/HTML/siteURL/original.html' ),
+		'expected' =>file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/Engine/CDN/Subscriber/HTML/siteURL/rewrite.html' ),
```
hellofromtonya, line 7:
```suggestion
		'expected' => file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/Engine/CDN/Subscriber/HTML/siteURL/rewrite.html' ),
```

tests/Fixtures/inc/Engine/CDN/Subscriber/rewrite.php
```diff
@@ -0,0 +1,14 @@
...
+	],
+	'testSiteURLWithPath' => [
+		'site_url' => 'http://example.org/blog',
+		'original' => file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/Engine/CDN/Subscriber/HTML/siteURLWithPath/original.html' ),
+		'expected' =>file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/Engine/CDN/Subscriber/HTML/siteURLWithPath/rewrite.html' ),
```
hellofromtonya, line 12:
```suggestion
		'expected' => file_get_contents( WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/Engine/CDN/Subscriber/HTML/siteURLWithPath/rewrite.html' ),
```

tests/Integration/inc/Engine/CDN/Subscriber/rewrite.php
```diff
@@ -0,0 +1,81 @@
...
+		return $this->site_url;
+	}
+
+	public function setContentURL() {
+		return $this->content_url;
+	}
+
+	public function setIncludesURL() {
+		return $this->includes_url;
+	}
```
hellofromtonya, line 80: There are repeating methods in this bunch. Let's create an `TestCase` for the `CDN` and move these methods to it. Then we can share it amongst all the tests in this bunch.

Screenfeed:
Isn't it a bit risky to replace all `get_option( 'home' )` by `site_url()` like this?

Tabrisrp:
> Isn't it a bit risky to replace all get_option( 'home' ) by site_url() like this?

I did this to bring more consistency with the rest of the plugin.

On the 2 occurrences, only the `site_url()` host part is used. Can we think of scenarii where the host value from `home` and `site_url` are different?

inc/Engine/CDN/Subscriber.php
Tabrisrp: https://github.com/wp-media/wp-rocket/pull/2589/commits/0143b88bf6ffa55847b18305f7e4ce1ad0e409ac

Screenfeed:
>Can we think of scenarii where the host value from `home` and `site_url` are different?

Some multilingual plugin with a dedicated domain for each language. Also, maybe some host companies where the admin panel is not located at the same place from the front, but I'm not sure the host value changes in this case.
Anyway, these cases must be handled with a 3rd party compatibility layer I guess. But I'm just wondering if we will introduce a bug by doing this.

Tabrisrp:
> Some multilingual plugin with a dedicated domain for each language.

the plugin would filter `site_url` and return the appropriate domain, so the replacement would still work as expected, no? The same if the i18n plugin is used with subdomains configuration.

Screenfeed:
>the plugin would filter `site_url` and return the appropriate domain

Wouldn't it filter `home_url` instead?
For example that's what WPML does:
```php
public function add_global_hooks() {
	add_filter( 'home_url', array( $this, 'home_url_filter' ), - 10, 4 );
	// posts and pages links filters
	add_filter( 'post_link', array( $this, 'permalink_filter' ), 1, 2 );
	add_filter( 'post_type_link', array( $this, 'permalink_filter' ), 1, 2 );
	add_filter( 'wpml_filter_link', array( $this, 'permalink_filter' ), 1, 2 );
	add_filter( 'get_edit_post_link', array( $this, 'get_edit_post_link' ), 1, 3 );
}
```
But I'm not sure the custom domain would be applied here, it may be elsewhere in the plugin.

piotrbak:
Tested this PR with WPML subdomain site. HTTP/HTTPS rewrites worked correctly. In this setup replacing ` get_option( 'home' );` in `rewrite_url()` and `get_base_url()` with any of the following solved the problem:
- `site_url()`
- `home_url()`
- `get_home_url()`

inc/Engine/CDN/CDN.php
```diff
@@ -207,7 +205,7 @@ public function get_cdn_urls( $zones = 'all' ) {
 	 * @return string
 	 */
 	private function get_base_url() {
-		return '//' . wp_parse_url( get_option( 'home' ), PHP_URL_HOST );
+		return '//' . wp_parse_url( site_url(), PHP_URL_HOST );
```
hellofromtonya, line 208: Let's talk about `site_url()` vs. `home_url()` to replace `get_option( 'home' )`:

- WordPress' `home_url()` function grabs `get_option( 'home' ):
```php
	if ( empty( $blog_id ) || ! is_multisite() ) {
		$url = get_option( 'home' );
	} else {
		switch_to_blog( $blog_id );
		$url = get_option( 'home' );
		restore_current_blog();
	}
```
- Polylang, WPML, and qTranslate all use the filter `'home_url`' and none use `'site_url'`
- These same i18n plugins also use `home_url()`
- It seems all of them use `home_url()` and `'home_url'` filter for getting or setting the right language

@Tabrisrp It seems `home_url()` is more aligned with the original functionality than `site_url()`. 

Switching this code from `get_option( 'home' )` to `home_url()` is a good strategy to fire the filter which then triggers the language plugins and anything else that filters the home URL.

Tabrisrp: > Polylang, WPML, and qTranslate all use the filter 'home_url' and none use 'site_url'

That's not entirely true. Polylang definitely filters `site_url` (2 occurrences in the current code), WPML too since we tested it on a customer website above.

`site_url` is used in WP as the base URL for anything in `wp_includes|wp-content|wp-content/uploads`, so it made sense to me to use it. But in the end the `host` part of either `home_url()` or `site_url()` should be the same, so I don't mind switching to `home_url()` there. Going to make the change in a bit.

hellofromtonya:
>I did this to bring more consistency with the rest of the plugin.

@Tabrisrp This is a great point. Do you think we need a helper function to centralize how we grab the URL, i.e. one place where we can ensure the code is consistently handling this task? What do you think? Is that overkill or helpful?

Tabrisrp:
I can see `home_url()` and `site_url()` values that could be part of the "configuration" system that we plan on using to replace constants.

inc/classes/ServiceProvider/class-common-subscribers.php
```diff
@@ -63,11 +61,6 @@ public function register() {
 		$this->getContainer()->share( 'expired_cache_purge_subscriber', 'WP_Rocket\Subscriber\Cache\Expired_Cache_Purge_Subscriber' )
 			->withArgument( $this->getContainer()->get( 'options' ) )
```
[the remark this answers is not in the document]
Tabrisrp: https://github.com/wp-media/wp-rocket/pull/2589/commits/7a20212d504a974148db5532beab9467e214fdfe

inc/Engine/CDN/CDN.php
Tabrisrp: https://github.com/wp-media/wp-rocket/pull/2589/commits/b587bf701af10ed681825e8218d9f6f7a03ade54

hellofromtonya:
>I can see home_url() and site_url() values that could be part of the "configuration" system that we plan on using to replace constants.

Agreed. We'll tackle that when we move to configuration architecture in the future. Thanks.

inc/Engine/CDN/CDN.php
```diff
@@ -108,14 +104,17 @@ public function rewrite_url( $url ) {
 			return rocket_add_url_protocol( $cdn_url . '/' . ltrim( $url, '/' ) );
 		}
 
-		$home       = get_option( 'home' );
+		$home       = home_url();
 		$home_parts = wp_parse_url( $home );
```
hellofromtonya, line 108: The `$home` var isn't used past the next line of code. Right? As the function is named "home_url", it's clear what gets returned from it. Therefore, I think we can combine these 2 lines of code.

What are the advantages?

- Less memory
- Less processing
+ One less line of code to maintain

inc/Engine/CDN/CDN.php
```diff
@@ -207,7 +205,7 @@ public function get_cdn_urls( $zones = 'all' ) {
 	 * @return string
 	 */
 	private function get_base_url() {
-		return '//' . wp_parse_url( get_option( 'home' ), PHP_URL_HOST );
+		return '//' . wp_parse_url( home_url(), PHP_URL_HOST );
```
hellofromtonya, line 208: We are getting `home_url()` 2x in this class. That means it'll go through the filtering 2x. To make this more performant, let's move it to a private method and add a property. If the property is not populated, them get `home_url()`; else, return the property.

inc/classes/ServiceProvider/class-common-subscribers.php
```diff
@@ -46,36 +44,33 @@ class Common_Subscribers extends AbstractServiceProvider {
...
+			->withArgument( $options );
 		$this->getContainer()->add( 'critical_css_generation', 'WP_Rocket\Optimization\CSS\Critical_CSS_Generation' );
 		$this->getContainer()->add( 'critical_css', 'WP_Rocket\Optimization\CSS\Critical_CSS' )
 			->withArgument( $this->getContainer()->get( 'critical_css_generation' ) );
 		$this->getContainer()->share( 'critical_css_subscriber', 'WP_Rocket\Subscriber\Optimization\Critical_CSS_Subscriber' )
 			->withArgument( $this->getContainer()->get( 'critical_css' ) )
-			->withArgument( $this->getContainer()->get( 'options' ) );
+			->withArgument( $options );
 		$this->getContainer()->add( 'expired_cache_purge', 'WP_Rocket\Cache\Expired_Cache_Purge' )
 			->withArgument( WP_ROCKET_CACHE_PATH );
```
hellofromtonya, line 61: Let's go ahead and wrap this constant in `rocket_get_constant( 'WP_ROCKET_CACHE_PATH' );`. I'm seeing instances in our test scenarios where we may need to alter this value.

tests/Integration/inc/Engine/CDN/Subscriber/TestCase.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace WP_Rocket\Tests\Integration\inc\Engine\CDN\Subscriber;
+
+use WPMedia\PHPUnit\Integration\TestCase as BaseTestCase;
+
+class TestCase extends BaseTestCase {
```
hellofromtonya, line 7: Make this `abstract`.

tests/Integration/inc/Engine/CDN/Subscriber/rewrite.php
```diff
@@ -0,0 +1,60 @@
...
+		return $this->getTestData( __DIR__, 'rewrite' );
+	}
+
+	public function setContentURL() {
+		return $this->content_url;
+	}
+
+	public function setIncludesURL() {
+		return $this->includes_url;
+	}
```
hellofromtonya, line 59: Move these to the test case too, just in case we need them in the future for other tests.

tests/Unit/inc/Engine/CDN/CDN/rewrite.php
```diff
@@ -0,0 +1,45 @@
...
+			if ( strpos( $url, 'http://' ) !== false || strpos( $url, 'https://' ) !== false ) {
+				return $url;
+			}
+
+			if ( substr( $url, 0, 2 ) === '//' ) {
+				return 'http:' . $url;
+			}
+
+			return 'http://' . $url;
+		} );
```
hellofromtonya, line 32: Let's create a base test case for the module and move these stubs into there. It'll reduce the code in the individual tests.

Tabrisrp: they all depend on values passed from the fixtures to the test itself, that's why I didn't move them into the testCase.

hellofromtonya: Good point!

inc/classes/ServiceProvider/class-common-subscribers.php
Tabrisrp: https://github.com/wp-media/wp-rocket/pull/2589/commits/5fabd07e111b0cf8cfeb8036f4f4bdde1d6c3ec2

inc/Engine/CDN/CDN.php
Tabrisrp: https://github.com/wp-media/wp-rocket/pull/2589/commits/7c2ec6c8f0c776c7b86cf3a17d81daf192df446e

inc/Engine/CDN/CDN.php
Tabrisrp: https://github.com/wp-media/wp-rocket/pull/2589/commits/7c2ec6c8f0c776c7b86cf3a17d81daf192df446e

tests/Integration/inc/Engine/CDN/Subscriber/rewrite.php
Tabrisrp: https://github.com/wp-media/wp-rocket/pull/2589/commits/680c402d2ee1b0611609890244583b4b81a76e49

tests/Integration/inc/Engine/CDN/Subscriber/TestCase.php
Tabrisrp: https://github.com/wp-media/wp-rocket/pull/2589/commits/680c402d2ee1b0611609890244583b4b81a76e49

hellofromtonya:
@wp-media/dev-qa @piotrbak @arunbasillal This PR is ready for QA.

Tabrisrp:
It's a different issue than the one for this PR, so if it happens we can create a new issue and work on that.

PR #2590, opened 2020-04-29
Closes #2478 Preload fetches wrong WPML Urls

### LINKED ISSUE #2478: Preload fetches wrong WPML URLs when cache expires
**Describe the bug**
When WPML is being used in subdirectories, we're not able to preload language URLs after the cache expiration.

That's because we're passing URL with a double slash here:
https://github.com/wp-media/wp-rocket/blob/9b175725a5c568f27808875c041fc67cba264ae7/inc/Engine/Preload/PartialPreloadSubscriber.php#L149

`$data['home_path']` value: `/home/domain/public_html/wp-content/cache/wp-rocket/domain.com/da`

`$data['home_url']` value: `https://domain.com/da/`

`$file_path` value: `/home/domain/public_html/wp-content/cache/wp-rocket/domain.com/da/store/path-to-file/`

After `str_replace`:
`https://domain.com/da//store/path-to-file/`

**To Reproduce**
1. Use WPML subdirectory environment
2. Make sure that the subdirectory, language content is cached, preload is turned on, cache lifespan is set to the low value, so we don't have to wait
3. The cache should be purged but not preloaded

**Expected behavior**
The cache should be preloaded correctly. A possible solution is to make sure there are no double slashes in the URL.

Related ticket:
https://secure.helpscout.net/conversation/1107833503/151289?folderId=2415573

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
**Reproduce the issue** ✅ 
Reproduced on mega website.

**Identify the root cause** ✅ 
The root cause is related to `$data['home_path']` which has the missing / at the end, so the str_replace() will have double // .
`home_path` trims the `/` in here:
https://github.com/wp-media/wp-rocket/blob/0bdcd1d0f9897541af2b589731d04af166ba1504/inc/classes/Cache/class-expired-cache-purge.php#L117-L118

**Scope a solution** ✅ 
The solution is to add back the '/' in `home_path` if the permalink_structure has the trailing slash:
```
$file_path = untrailingslashit( $file_path );
$data['home_path'] = untrailingslashit( $data['home_path'] );
if ( '/' === substr( get_option( 'permalink_structure' ), -1 ) ) {
		$file_path .= '/';
		$data['home_path'] .= '/';
}
```

**Estimate the effort** ✅ 
Effort [S] - All the effort should go in writing Unit & Integration tests for `preload_after_automatic_cache_purge()` function

**To Do**
- [x] Unit tests for `preload_after_automatic_cache_purge()` function
- [x] Integration tests for `preload_after_automatic_cache_purge()` function
- [x] QA

### DIFF
--- a/tests/Integration/inc/Engine/Preload/PartialPreloadSubscriber/preloadAfterAutomaticCachePurge.php
+++ b/tests/Integration/inc/Engine/Preload/PartialPreloadSubscriber/preloadAfterAutomaticCachePurge.php
@@ -29,8 +29,8 @@ class Test_PreloadAfterAutomaticCachePurge extends TestCase {
 		parent::tearDown();
 
 		$this->subscriber = null;
-		$this->property->setAccessible( false );
 		$this->property->setValue( $this->subscriber, [] );
+		$this->property->setAccessible( false );
 	}
 
 	/**
--- a/tests/Integration/inc/Engine/Preload/PartialPreloadSubscriber/preloadAfterAutomaticCachePurge.php
+++ b/tests/Integration/inc/Engine/Preload/PartialPreloadSubscriber/preloadAfterAutomaticCachePurge.php
@@ -30,6 +30,9 @@ class Test_PreloadAfterAutomaticCachePurge extends TestCase {
 
 		$this->property->setValue( $this->subscriber, [] );
 		$this->property->setAccessible( false );
+
+		remove_filter( 'pre_get_rocket_option_manual_preload', [ $this, 'manual_preload_filter' ] );
+		remove_filter( 'pre_option_permalink_structure', [ $this, 'permalink_structure_filter' ] );
 	}
 
 	/**
@@ -54,9 +57,6 @@ class Test_PreloadAfterAutomaticCachePurge extends TestCase {
 		foreach ( $expected as $url ) {
 			$this->assertContains( $url, $this->urls );
 		}
-
-		remove_filter( 'pre_get_rocket_option_manual_preload', [ $this, 'manual_preload_filter' ] );
-		remove_filter( 'pre_option_permalink_structure', [ $this, 'permalink_structure_filter' ] );
 	}
 
 	public function manual_preload_filter() {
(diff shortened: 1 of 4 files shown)


tests/Unit/inc/Engine/Preload/PartialPreloadSubscriber/preloadAfterAutomaticCachePurge.php
```diff
@@ -0,0 +1,71 @@
...
+class Test_PreloadAfterAutomaticCachePurge extends TestCase {
+	private $urls = [];
+	private $options;
+	private $partial_process;
+	private $subscriber;
+	private $property;
+
+	public function setUp() {
+		$this->options         = $this->createMock( Options_Data::class );
+		$this->partial_process = $this->createMock( PartialProcess::class );
```
Tabrisrp, line 25: use `Mockery` for mocks, to be consistent across all our tests

tests/Integration/inc/Engine/Preload/PartialPreloadSubscriber/preloadAfterAutomaticCachePurge.php
```diff
@@ -0,0 +1,73 @@
...
+		if ( ! $expected ) {
+			$this->assertEmpty( $this->urls );
+		}
+
+		foreach ( $expected as $url ) {
+			$this->assertContains( $url, $this->urls );
+		}
+
+		remove_filter( 'pre_get_rocket_option_manual_preload', [ $this, 'manual_preload_filter' ] );
+		remove_filter( 'pre_option_permalink_structure', [ $this, 'permalink_structure_filter' ] );
```
hellofromtonya, line 59: Move the `remove_filter` to the `tearDown()`. Why? If the assertions fail, these 2 lines will not execute.

crystinutzaa:
@arunbasillal this one needs QA

piotrbak:
@arunbasillal Worked like expected on Mega. Values of `$data['home_path']`:
-Before fix
`/home/wpquicki/mega/wp-content/cache/wp-rocket/mega.wp-rocket.me/en`
-After fix
`/home/wpquicki/mega/wp-content/cache/wp-rocket/mega.wp-rocket.me/en/`

The preload fetched the correct URLs in all cases. Is there any chance you could check what's going on in your case?

PR #2592, opened 2020-04-29
Closes #2543 Preload for fonts

### LINKED ISSUE #2543: Preload For Fonts

Tabrisrp:
Closes epic #2543

This PR contains all the changes for the feature and completes the epic. Each sub-task has been reviewed and QA already.

**Missing**
- [x] Preserve the query string & hash in the URL in the option field
- [x] An URL with the CDN URL should be preserved when adding it in the option field
- [x] Documentation FR links for Beacon
- [x] QA


webtrainingwheels:
If the CDN option is activated, and the customer enters their font using their CDN url, eg.
`https://cdn.lucy.com/wp-content/themes/paperback/inc/fontawesome/fonts/fontawesome-webfont.woff2?v=4.4.0`

Currently we don't recognize it and remove the whole font url from the box when saving. I think it would be ideal to recognize the CDN CNAME and just remove the domain, as if it were the local domain.

webtrainingwheels:
Some fonts will have a query string eg: `wp-content/themes/paperback/inc/fontawesome/fonts/fontawesome-webfont.woff2?v=4.4.0`
Because the query string is stripped off, this creates 2 issues:
-  there's now 2 requests for the font in the Waterfall- one with the query string and one without: https://jmp.sh/RuRlH75
-   Chrome shows this warning: https://jmp.sh/1eK19KA

Tabrisrp:
@hellofromtonya @arunbasillal @GeekPress Please note the 2 comments above from Lucy.

> there's now 2 requests for the font in the Waterfall- one with the query string and one without: https://jmp.sh/RuRlH75

Is it only on GTmetrix, or also in the browser network?

> Chrome shows this warning

About this warning:
- it's also present in Safari

This means the font should be added to the preload field really only if it's used early in the page. I guess this will need to be in the doc about the feature?

webtrainingwheels:
@Tabrisrp 
> This means the font should be added to the preload field really only if it's used early in the page. I guess this will need to be in the doc about the feature?

To be clear, the Chrome warning only shows due to the query string issue. 
Preloading the font file correctly with the query string removes the Chrome warning.

Tabrisrp:
Thanks for the clarification 👍

webtrainingwheels:
The EN link for documentation will be: https://docs.wp-rocket.me/article/1317-preload-fonts
The article ID is: 5eab7729042863474d19f647

Tabrisrp:
PR is updated to fix the issues reported by Lucy above
- Preserve the query string/fragment when it's present in the original URL
- Preserve the entry if it uses an URL corresponding to one of the URL in the CDN cnames field

inc/Engine/Preload/Fonts.php
```diff
@@ -0,0 +1,136 @@
...
+		$fonts = (array) apply_filters( 'rocket_preload_fonts', $fonts );
+		$fonts = array_map( [ $this, 'sanitize_font' ], $fonts );
+		$fonts = array_filter( $fonts );
+
+		if ( empty( $fonts ) ) {
+			return;
+		}
+
+		$base_url = get_rocket_parse_url( home_url() );
+		$base_url = "{$base_url['scheme']}://{$base_url['host']}";
```
hellofromtonya, line 99: @Tabrisrp Do we get the base URL in other places in WPR? If yes, then I'm thinking we'll want to centralize this functionality to reuse everywhere. What do you think?

Tabrisrp: we have different need in different places in fact, sometimes we need only the `host`, sometimes the base URL, sometimes we need that but with the `site_url()` instead.

tests/Unit/inc/Engine/Admin/Settings/Settings/sanitizeCallback.php
```diff
@@ -19,7 +19,7 @@ class Test_SanitizeCallback extends TestCase {
 	public function setUp() {
 		parent::setUp();
 
-		$this->options  = $this->createMock( Options::class );
+		$this->options  = $this->createMock( Options_Data::class );
```
crystinutzaa, line 22: This should be changed to Mockery:mock(), right?

hellofromtonya: Yes it should.

Tabrisrp: https://github.com/wp-media/wp-rocket/pull/2592/commits/3a497ea3a2dcefb601fc4fc0ab9c785ddd8046e7

Tabrisrp:
@wp-media/dev-qa Ready for a new QA, fixing the 2 issues reported above by Lucy

Tabrisrp:
> Output: https://jmp.sh/7mLayVB 👍
Output in Frontend: These were not added to the front end ❌ (@Tabrisrp please note)

I don't understand, your image shows the output on the frontend, with the expected link?

hellofromtonya:

Tabrisrp:
These should be fixed now too

Tabrisrp:
@arunbasillal Edge case indeed, but I was able to change the code to accomodate that. Can you give it another try?

arunbasillal:
@Tabrisrp Thanks for the update. 

Right now, when I input `https://example.com/wp-content/themes/font.woff`, and when the CDN is set to `https://example.com/path`, the input is removed after sanitization. 

Can you please try the exact same input at your end and see what happens?

Tabrisrp:
Is `example.com` the site URL or a different URL? It's unclear.

If `https://example.com/path` is the CDN URL in the cnames, and you input `https://example.com/wp-content/themes/font.woff`, the input will be removed, because it doesn't match the CDN URL.

`example.com` !== `example.com/path`.

arunbasillal:
@Tabrisrp `example.com` is the CDN CNAME and NOT the website URL. 

> If https://example.com/path is the CDN URL in the cnames, and you input https://example.com/wp-content/themes/font.woff, the input will be removed, because it doesn't match the CDN URL.
> example.com !== example.com/path.

This was the edge case reported. And this is indeed covered by the update. So we are good 👍 

However, I think we have created a regression. 

This is the input I used:
```
#Local fonts with domain (wp-rocket-dev.local) is the local domain
http://wp-rocket-dev.local/wp-content/themes/theme/font.otf
http://wp-rocket-dev.local/wp-content/themes/theme/font.ttf
http://wp-rocket-dev.local/wp-content/themes/theme/font.svg
http://wp-rocket-dev.local/wp-content/themes/theme/font.woff
http://wp-rocket-dev.local/wp-content/themes/theme/font.woff2
http://wp-rocket-dev.local/wp-content/themes/theme/font.eot

#External domain
http://domain.com/wp-content/themes/theme/font-external.woff

#No domain
/wp-content/themes/theme/font-no-domain.woff

#Disallowed extension
/wp-content/themes/theme/font.eot

#Query String
wp-content/themes/paperback/inc/fontawesome/fonts/fontawesome-webfont.woff2?v=4.4.0

#Query string with leading slash
/wp-content/themes/paperback/inc/fontawesome/fonts/fontawesome-webfont.woff2?v=4.4.0

#Fragment
wp-content/themes/paperback/inc/fontawesome/fonts/fontawesome-webfont.woff2#Fragment

#CDN with Path (CDN CNAME set to https://example.com/path)
https://example.com/path/wp-content/themes/paperback/inc/fontawesome/fonts/fontawesome-cdn-path.woff2?v=path

#CDN without Path
https://example.com/wp-content/themes/paperback/inc/fontawesome/fonts/fontawesome-cdn-no-path.woff2?v=path
```

Output was: 

```
/wp-content/themes/theme/font-no-domain.woff
/wp-content/themes/paperback/inc/fontawesome/fonts/fontawesome-webfont.woff2?v=4.4.0
/wp-content/themes/paperback/inc/fontawesome/fonts/fontawesome-webfont.woff2#Fragment
/wp-content/themes/paperback/inc/fontawesome/fonts/fontawesome-cdn-path.woff2?v=path
```

Looks like `http://wp-rocket-dev.local` (the local site url) was treated as external domain. 
Also it recorded this in debug.log - http://snippi.com/raw/uvvl4tc when you try the input exactly as above including the #Titles

Tabrisrp:
The `#titles` should not be included in the input, using the `#` will make the sanitizer think it's part of a fragment URL.

I fixed the issue you noted above, it was a mistake from me in the logic

PR #2596, opened 2020-04-29
Fixes the 3.5 deprecated class aliases, i.e. backwards.

hellofromtonya:
This PR fixes the `class_alias` declarations in the 3.5 deprecated file. Why? The arguments are backwards and should be:

```php
class_alias( 'the original class', 'the new class' );
```

[See the PHP manual](https://www.php.net/manual/en/function.class-alias.php).

### DIFF
--- a/inc/deprecated/3.5.php
+++ b/inc/deprecated/3.5.php
@@ -6,18 +6,18 @@ defined( 'ABSPATH' ) || exit;
 /**
  * Class aliases.
  */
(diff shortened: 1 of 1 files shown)


Screenfeed:
@hellofromtonya I believe it's the opposite. I understand it that way:

```php
class_alias( 'the new current class', 'the old class' );
```

My tests:
The previous code: https://3v4l.org/5UO1J
The new code: https://3v4l.org/WKQUY

PR #2597, opened 2020-04-30
Fix unpredictable caching on TravisCI

szepeviktor:
...and `true` is a better `return 0`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ branches:
   - /branch-.*$/
 cache:
   directories:
-  - vendor
   - $HOME/.composer/cache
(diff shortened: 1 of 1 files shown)


szepeviktor:
`Test_GetSubscriptionData::testShouldReturnDefaultAndSetTransientWhenValidUserTokenButNoDataReceived` fails, it seems to be unrelated

szepeviktor:
This PR just received commits for .github/ISSUE_TEMPLATE/bug_report.md 😮

szepeviktor:
@Tabrisrp Now it is OK 🍏

PR #2603, opened 2020-05-01
Fixes #2170 Disable cache clear when a term is not public

### LINKED ISSUE #2170: Disable cache clearing when a term is not public
Currently, when a term is created/edited/deleted we clear the whole cache:
https://github.com/wp-media/wp-rocket/blob/7723e789d158d2bf00ee5da49ef9158061e158eb/inc/common/purge.php#L14-L16

There are cases, e.g. Yoast Premium, where a taxonomy is registered having `'public' => false`. In this case, it's not necessary to clear the cache.

This can be an issue on sites where terms are constantly created/changed.

**Steps to reproduce the issue**
1. Register a non-public taxonomy: http://snippi.com/raw/iqs1jwq
2. Check `/wp-content/cache/` and make sure there is cached content.
3. Create a post and add a term to the taxonomy created👆.
4. Publish the post.
5. The cache from step 2 will be cleared.

**Proposed solution:**

Check if a term is `public` before clearing the cache with `rocket_clean_domain()`.

**Related ticket:**
https://secure.helpscout.net/conversation/1027543602/135931/

Tabrisrp:
Fixes #2170 

- [x] Move `PurgeActionsSubscriber` to new architecture
- [x] Add method to purge the cache if term is public
- [x] Add service provider for the cache module
- [x] Remove callback for terms actions in `purge.php`
- [x] Write tests
- [x] QA


inc/Engine/Cache/PurgeActionsSubscriber.php
```diff
@@ -54,11 +55,44 @@ public function purge_user_cache( $user_id ) {
...
+	/**
+	 * Checks if the given taxonomy is public
+	 *
+	 * @param string $name Taxonomy name.
+	 * @return bool
+	 */
+	private function is_taxonomy_public( $name ) {
+		$taxonomy = get_taxonomy( $name );
+
+		if ( ! $taxonomy ) {
```
hellofromtonya, line 85: `get_taxonomy()` returns `WP_Taxonomy` or `false`. The guard clause here should check if `false` is returned. Right now, it's a boolean check that evaluates whatever data type is returned from `get_taxonomy()`. If an object of `WP_Taxonomy` is returned, PHP has to type juggle to do the boolean check. Instead, we can be a bit more clear and performant by checking:

```php
if ( false === $taxonomy ) {
```

When doing this, it avoids the type juggle and is a little faster.

inc/Engine/Cache/PurgeActionsSubscriber.php
```diff
@@ -54,11 +55,44 @@ public function purge_user_cache( $user_id ) {
...
+			return false;
+		}
+
+		return (bool) ( $taxonomy->public && $taxonomy->publicly_queryable );
```
hellofromtonya, line 89: `(bool)` is not needed here because the result of the `( ... )` conditional expression is a boolean data type. We only need to type cast _when_ it might not be that data type.

tests/Fixtures/inc/Engine/Cache/PurgeActionsSubscriber/maybePurgeCacheOnTermChange.php
```diff
@@ -0,0 +1,62 @@
...
+						'index.html_gzip' => '',
+					],
+					'example.org-wpmedia-594d03f6ae698691165999' => [
+						'index.html'      => '',
+						'index.html_gzip' => '',
+					],
+				],
+			],
+		],
+	],
```
hellofromtonya, line 22: Could you use the default structure instead? 

Why? I'm thinking it would be good to exercise our code in a broader sense of our filesystem structure. 

If yes, then you can delete the entire element as the vfs test case will automatically pull in the default structure.

Tabrisrp: I updated the tests to use the default structure & the new methods to perform the assertions. Can you check I used them correctly?

Tabrisrp:
@wp-media/dev-qa Ready for QA

PR #2604, opened 2020-05-01
[WIP] Benchmarking rocket_clean_files() RegexIterator performance

hellofromtonya:
Adds benchmarking code for `rocket_clean_files()` SPL `RegexIterator` performance compared to the `glob()` logic.

Compiled results are found in [this spreadsheet](https://docs.google.com/spreadsheets/d/1lHSzxnbFAjB3Ls08T-Jb4ImbDxA1YAhsNh5LQ3nlWpE/edit?usp=sharing) and individual csv results are in the results folder of this PR.

## Test Environment

- Tested with 30 URLs
- `glob()` and `_rocket_get_entries_regex()` will grab 31 entries which are then sent to `rocket_rrmdir()` for recursive removal
- Cache is created when running the tests
- Cache consists of:
     - 31 root entries:
          - Domain cache:   `example.org` 
          - 30 user caches: `example.org-user1-123456` through `example.org-user30-123456` 
     - Each cache has 30 entries `post1` through `post30`
     - There are no nested directories with the `postN` directories.
- Using the real hard drive instead of virtual to better match conditions customers see

## Observations

- regex for `RegexIterator` was matching the first part of the URLs `'path'`. 
For example, if the URL is `http://example.org/post1/`:
     - `glob()` matches `example.org*/post1/`
     - `RegexIterator` matched `example.org(.*)/post1`, which. resulted in "post1" and "post10" to "post19" for each root level entry. 👈 This is not the behavior we want.

- Fixed that problem by adding `\b` to the regex. Now `glob()` and `_rocket_get_entries_regex()` grab and return the same results

- `_rocket_get_entries_regex()` is faster than `glob()` by ~0.38ms
- When iterating through the returned entries, `glob()` is faster by ~31ms
- On average `glob()` is faster by:
      - Per URL = ~31ms
      - 30 URLs = ~1 sec

cc / @vmanthos @Tabrisrp @piotrbak


hellofromtonya:
After updating to 3.5.5 version of `rocket_clean_files()` and re-running the benchmarking.

Cleaning 90 URLS (1 main domain + 30 user caches):

- WPR 3.5.2 with `glob`: ~0.004s
- WPR 3.5.5 with SPL: ~0.004s

They run essentially in the same amount of time.

PR #2573, opened 2020-04-22
 Closes #2443 CPCSS for specific pages: UI & metabox (4/6)

Tabrisrp:
Closes #2443

Brings the UI changes to perform the generation/regeneration/deletion of the specific CPCSS for the current post in the metabox.

PR #2609, opened 2020-05-04
Closes #2444 - CPCSS AJAX handlers Part 5

### LINKED ISSUE #2444: CPCSS for specific pages: AJAX handlers (5/6)
Sub-Card for epic #2377.

This card is for tasks related to the AJAX handlers on  the `edit.php` screen to send the requests and receive the responses + display the success/error messages.

### Tasks
- [x] Enqueue new JS file on the `edit.php` screen. Needs to be compatible with both the classic editor and the block editor
#### Generation
- [x] Send POST request to REST API endpoint, with the ID of the page and the post type as a parameter + nonce
- [x] Receive JSON-formatted response from the endpoint `{success:true|false,message:"response"}`
- [x] Display success/error message depending on the response
#### Deletion
- [x] Send DELETE request to REST API endpoint, with the ID of the page and the post type as a parameter + nonce
- [x] Receive JSON-formatted response from the endpoint `{success:true|false,message:"response"}`
- [x] Display success/error message depending on the response

crystinutzaa:
Sub-Card for epic #2377.

This card is for tasks related to the AJAX handlers on  the `edit.php` screen to send the requests and receive the responses + display the success/error messages.

### Tasks
- [x] Enqueue new JS file on the `edit.php` screen. Needs to be compatible with both the classic editor and the block editor
#### Generation
- [x] Send POST request to REST API endpoint, with the ID of the page and the post type as a parameter + nonce
- [x] Receive JSON-formatted response from the endpoint `{success:true|false,message:"response"}`
- [x] Display success/error message depending on the response
#### Deletion
- [x] Send DELETE request to REST API endpoint, with the ID of the page and the post type as a parameter + nonce
- [x] Receive JSON-formatted response from the endpoint `{success:true|false,message:"response"}`
- [x] Display success/error message depending on the response

TO DO
- [x] Fix current Unit Tests
- [x] Fix current IntegrationTests
- [x] Write new Unit Tests for WP_Rocket\Engine\CriticalPath\AdminSubscriber::enqueue_admin_edit_script()
- [x] Write new IntegrationTests for WP_Rocket\Engine\CriticalPath\AdminSubscriber::enqueue_admin_edit_script()


assets/js/wpr-cpcss.js
```diff
@@ -0,0 +1,104 @@
+var $ = jQuery;
```
Tabrisrp, line 1: What's our take on using jQuery or not? Should we avoid it since it's an "of-the-past" framework? What happens if someone decides to unload it from the admin?

@hellofromtonya @crystinutzaa

hellofromtonya: We should eliminate the use of jQuery and standardize on vanilla JS or React.

crystinutzaa: Ok. So vanilla JS sounds good at this point?
I will convert the code today and get rid of all jQuery.

crystinutzaa: This is done. No more jQuery only Vanilla JS.

views/metabox/cpcss/container.php
```diff
@@ -16,15 +16,20 @@
...
+			<p><?php echo esc_html( $data['disabled_description'] ); ?></p>
+		<?php endif; ?>
 	</div>
 </div>
-<?php endif; ?>
+<script>
+	var cpcss_rest_url       = '<?php echo esc_url( $data['cpcss_rest_url'] ); ?>';
+	var cpcss_rest_nonce     = '<?php echo esc_html( $data['cpcss_rest_nonce'] ); ?>';
+	var cpcss_generate_btn   = '<?php esc_html_e( 'Generate Specific CPCSS', 'rocket' ); ?>';
+	var cpcss_regenerate_btn = '<?php esc_html_e( 'Regenerate specific CPCSS', 'rocket' ); ?>';
```
Screenfeed, line 34: If these contain a `'` character, we're screwed.
```suggestion
	var cpcss_rest_url       = '<?php echo esc_js( $data['cpcss_rest_url'] ); ?>';
	var cpcss_rest_nonce     = '<?php echo esc_js( $data['cpcss_rest_nonce'] ); ?>';
	var cpcss_generate_btn   = '<?php echo esc_js( __( 'Generate Specific CPCSS', 'rocket' ) ); ?>';
	var cpcss_regenerate_btn = '<?php echo esc_js( __( 'Regenerate specific CPCSS', 'rocket' ) ); ?>';
```

views/metabox/cpcss/generate.php
```diff
@@ -15,7 +15,7 @@
 defined( 'ABSPATH' ) || exit;
 
 ?>
-<p>
+<p class="cpcss_generate <?php echo ( $data['cpcss_exists'] ? 'hidden' : '' ); ?>">
```
Screenfeed, line 18:
```suggestion
<p class="cpcss_generate <?php echo ! empty( $data['cpcss_exists'] ) ? 'hidden' : ''; ?>">
```

views/metabox/cpcss/generate.php
```diff
@@ -25,8 +25,32 @@
 		);
 	?>
 </p>
-<div class="components-panel__row">
+<p class="cpcss_regenerate <?php echo ( ! $data['cpcss_exists'] ? 'hidden' : '' ); ?>">
```
Screenfeed, line 28:
```suggestion
<p class="cpcss_regenerate <?php echo empty( $data['cpcss_exists'] ) ? 'hidden' : ''; ?>">
```

views/metabox/cpcss/generate.php
```diff
@@ -25,8 +25,32 @@
...
+			if ( $data['cpcss_exists'] ) {
+				esc_html_e( 'Regenerate specific CPCSS', 'rocket' );
+			} else {
+				esc_html_e( 'Generate Specific CPCSS', 'rocket' );
+			}
+			?>
+		</span>
+	</button>
+</div>
+<div class="components-panel__row cpcss_regenerate <?php echo ( ! $data['cpcss_exists'] ? 'hidden' : '' ); ?>">
```
Screenfeed, line 52:
```suggestion
<div class="components-panel__row cpcss_regenerate <?php echo empty( $data['cpcss_exists'] ) ? 'hidden' : ''; ?>">
```

assets/js/wpr-cpcss.js
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		const cpcss_response = JSON.parse( this.responseText );
+		if ( cpcss_response.data.status !== 200 ) {
+			stopCPCSSGeneration( spinner );
+			cpcssNotice( cpcss_response.message, 'error' );
+			return;
+		}
+
+		if ( cpcss_response.data.status === 200 && cpcss_response.code !== 'cpcss_generation_pending' ) {
```
Screenfeed, line 36:
```suggestion
		if ( 200 === cpcss_response.data.status && 'cpcss_generation_pending' !== cpcss_response.code ) {
```

assets/js/wpr-cpcss.js
```diff
@@ -0,0 +1,112 @@
...
+	spinner.style.visibility = 'visible';
+
+	const xhttp              = new XMLHttpRequest();
+	xhttp.onreadystatechange = () => {
+		if ( this.readyState !== 4 || this.status !== 200 ) {
+			return;
+		}
+
+		const cpcss_response = JSON.parse( this.responseText );
+		if ( cpcss_response.data.status !== 200 ) {
```
Screenfeed, line 30:
```suggestion
		if ( 200 !== cpcss_response.data.status ) {
```

assets/js/wpr-cpcss.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+const deleteCPCSS = () => {
+	const xhttp              = new XMLHttpRequest();
+	xhttp.onreadystatechange = () => {
+		if ( this.readyState !== 4 | this.status !== 200 ) {
+			return;
+		}
+		const cpcss_response = JSON.parse( this.responseText );
+		if ( cpcss_response.data.status !== 200 ) {
```
Screenfeed, line 80:
```suggestion
		if ( 200 !== cpcss_response.data.status ) {
```

hellofromtonya:
@crystinutzaa I update the JS to ES6. Can you check that it still works the way? Also fix the conflicting files too. Thx

Screenfeed (changes requested):
Note:
We're displaying the API message directly (`cpcssNotice( cpcss_response.message, 'error' );` for example). At some point we will want the API to return a code (it may be already the case, I don't know), and use this code to select a corresponding local message. We can still have a message in English returned, that could be used as fallback (in case we add new messages to the API but the plugin is not updated with them yet).

Two reasons:
- Translatable: the message from the API is not, as far as I know.
- We don't display contents from outside (security reason).

Note inside the note (noteception):
We don't want the API to return a message that is already translated (by previously sending it a language code for example), because if some day we need to store this message in the database, it will then be displayed in this language to all users, which may use another language.

assets/js/wpr-cpcss.js
```diff
@@ -0,0 +1,116 @@
...
+var rocketCPCSSReGenerate   = document.querySelectorAll( '.cpcss_regenerate' );
+
+rocketDeleteCPCSSbtn.addEventListener( 'click', function( e ) {
+	e.preventDefault();
+	deleteCPCSS();
+});
+
+rocketGenerateCPCSSbtn.addEventListener( 'click', function( e ) {
+	e.preventDefault();
+	checkCPCSSGeneration();
```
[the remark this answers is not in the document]
crystinutzaa: I added it for Generate / Regenerate & Delete button. 
If is disabled, then the button is not clickable so it does not need to check this attribute.
Thanks for spotting this one 👍

inc/Engine/CriticalPath/RESTGenerate.php
```diff
@@ -0,0 +1,392 @@
...
+				// translators: %s = post URL.
+				sprintf( __( 'Critical CSS for %s in progress.', 'rocket' ), $post_url ),
+				200
+			);
+		}
+
+		if ( isset( $job_data->data->state, $job_data->data->critical_path ) && 'complete' === $job_data->data->state ) {
+			return $this->on_job_success( $post_id, $post_url, $post_type, $job_data->data->critical_path );
+		}
+	}
```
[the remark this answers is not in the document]
crystinutzaa: I assume that CPCSS API always have a `data->state` returned. Otherwise will return a `400 or 404` error. But we can return a generic error as a fallback.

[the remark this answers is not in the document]
hellofromtonya: Let's validate that assumption @crystinutzaa.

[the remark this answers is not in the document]
crystinutzaa: @hellofromtonya I think this Wiki will be enough:
https://github.com/wp-media/criticalpath-css-generator/wiki/API

views/metabox/cpcss/generate.php
```diff
@@ -15,7 +15,7 @@
 defined( 'ABSPATH' ) || exit;
```
Tabrisrp, line 16: We're checking `empty( $data['cpcss_exists'] )` multiple times in the file, let's do it once at the start of the file, then use that result after

remyperona (changes requested):

inc/Engine/CriticalPath/RESTGenerate.php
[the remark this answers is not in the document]
hellofromtonya: Agreed.

inc/Engine/CriticalPath/AdminSubscriber.php
```diff
@@ -65,8 +86,12 @@ public static function get_subscribed_events() {
 	 * @return void
 	 */
 	public function cpcss_section() {
+		global $post, $pagenow;
+
```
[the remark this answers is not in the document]
crystinutzaa: Do we need a variable? It is used only in a single place?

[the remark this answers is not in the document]
Tabrisrp: It's really for readability of the line below, I found it hard to understand with everything mixed in.

I also inverted the check, as testing for positive is easier to read than testing for negative.

views/metabox/cpcss/generate.php
```diff
@@ -25,8 +26,32 @@
...
+		);
+	?>
+</p>
+<div class="components-panel__row cpcss_generate cpcss_regenerate">
 	<button id="rocket-generate-post-cpss" class="button components-button is-secondary" <?php disabled( $data['disabled'] ); ?>>
-		<?php esc_html_e( 'Generate Specific CPCSS', 'rocket' ); ?>
+		<span class="spinner"></span>
+		<span class="rocket-generate-post-cpss-btn-txt">
+			<?php
+			if ( ! empty( $data['cpcss_exists'] ) ) {
```
Tabrisrp, line 44:
```suggestion
			if ( ! $rocket_cpcss_exists ) {
```

remyperona (changes requested):

views/metabox/cpcss/generate.php
crystinutzaa: Thank you! Missed it totally

inc/Engine/CriticalPath/RESTGenerate.php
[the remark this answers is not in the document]
Screenfeed: In that case, it can be simplified:
```php
	protected function check_cpcss_job_status( $cpcss_job_id, $post_id, $post_url, $post_type ) {
		$job_data = $this->get_critical_path( $cpcss_job_id );

		if ( in_array( (int) $job_data->status, [ 400, 404 ], true ) ) {
			return $this->on_job_error( $post_id, $post_url, $job_data );
		}

		if ( 'complete' !== $job_data->data->state ) {
			return $this->return_array_response(
				true,
				'cpcss_generation_pending',
				// translators: %s = post URL.
				sprintf( __( 'Critical CSS for %s in progress.', 'rocket' ), $post_url ),
				200
			);
		}

		return $this->on_job_success( $post_id, $post_url, $post_type, $job_data->data->critical_path );
	}
```

assets/js/wpr-cpcss.js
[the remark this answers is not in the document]
Screenfeed: Right, I didn't even know that disabled buttons don't trigger events attached to them. Now I know, thanks ^^'

inc/Engine/CriticalPath/RESTGenerate.php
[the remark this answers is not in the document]
crystinutzaa: Thank you 🙏

hellofromtonya (changes requested):

hellofromtonya (approved):

Screenfeed (approved):

remyperona (approved):

hellofromtonya:

arunbasillal (changes requested):
QA:

- There was a blank notice under the button in the new editor - https://youtu.be/Gc90gADSqbc?t=38 (@crystinutzaa please note)
- There is also an extra space to the right of the wording (where the spinner comes in later) - https://jmp.sh/e1DTAUY (Can this be removed?)
- Regenerate button has the extra space too - https://jmp.sh/UGUUdzc
- Regenerate and Revert buttons have different styling. Can these be the same please? - https://jmp.sh/UGUUdzc (Classic) https://jmp.sh/PLFA3I6 (New Editor). (It used to look like this https://jmp.sh/KoPONLM (Classic) https://jmp.sh/rvLl3Cp (New))

- Error message worked correctly on Local 👍 
- No blank notice in the classic editor - https://jmp.sh/wp66hC3 👍 
- On Clicking the `Revert` button, the specific CPCSS was deleted correctly. UI changed to https://jmp.sh/kgN0QKD (New) - https://jmp.sh/mCIiYC1 (Classic) 👍 

For reference, the approved UI is available here: https://github.com/wp-media/wp-rocket/pull/2573#pullrequestreview-405787464

crystinutzaa:
@hellofromtonya & @Tabrisrp please review the latest changes applied for UI issues reported by @arunbasillal 
https://github.com/wp-media/wp-rocket/pull/2609/commits/0f85eb438e22cf6e485c1384a35d1ded9da6ada4 & https://github.com/wp-media/wp-rocket/pull/2609/commits/f822c7dabac43e09cb0be39a62e6be6340d2c6c7


@arunbasillal  all issues are fixed now. Please retry.

hellofromtonya (approved):

views/metabox/cpcss/generate.php
```diff
@@ -38,7 +38,7 @@
 </p>
 <div class="components-panel__row cpcss_generate cpcss_regenerate">
 	<button id="rocket-generate-post-cpss" class="button components-button is-secondary" <?php disabled( $data['disabled'] ); ?>>
-		<span class="spinner"></span>
+		<span style="display: none;" class="spinner"></span>
```
Tabrisrp, line 41: why not use the `hidden` class and remove it when needed, instead of an inline style?

remyperona (changes requested):

views/metabox/cpcss/generate.php
crystinutzaa: If I will use the `hidden` class, then I will have to modify also the CSS file. This is because:
`.spinner` has the following rules:
```
display: inline-block;
visibility: hidden;
```
And it will take priority over `hidden` class rules. So I will have to add a new CSS rule which will take priority with display: none;. 

In the HTML files are we allowed to use inline CSS code? I am sure that is preffered not to use any CSS code inside the `template` files, however for this case what is the direction?

arunbasillal:
@crystinutzaa The UI looks okay and works perfectly on my Local 👍 

On Mega, I had trouble deleting the Specific CPCSS. Please check this post - http://mega.wp-rocket.me/wp-admin/post.php?post=9336&action=edit
i

crystinutzaa:
@arunbasillal it seems that something is blocking the access to the REST API endpoint.
Was the generation successfull and only the deletion failed?

arunbasillal:
@crystinutzaa The generation was successful and only the deletion failed. I have installed this branch on the mega site, please take a look directly - http://mega.wp-rocket.me/wp-admin/post.php?post=9336&action=edit 

You can replicate this on any post as well.

crystinutzaa:
@arunbasillal It seems that the method DELETE may be blocked by third party security applications.
On mega I assume this plugin: ManageWP - Worker is blocking our call.

To test it, I switched from DELETE to GET and it worked just fine.

@hellofromtonya @Tabrisrp  what should be the approach in this case? I would go on changing from DELETE to GET / POST with a delete param, so that we can bypass any security plugin which may be blocking all DELETE call.

Tabrisrp:
the `DELETE` method for HTTP calls is a legitimate method, and the recommended one for any action to delete things when using an HTTP API like the REST API.

The notice on error can display the reason, and our documentation explain what to do in that case.

Should we compromise the logic and semantic to workaround plugins that are blocking valid requests?

crystinutzaa:
@Tabrisrp really good point. This is my question too. 
However let's think about the end customer and how support may be impacted if the customers do not read the documentation. 
@hellofromtonya , @arunbasillal , @GeekPress ?

arunbasillal:
I believe this is the source of the issue: 

image

My vote is towards bypassing this in WP Rocket code rather than expecting users to fix it. Unless of course it causes bigger problems in the future.

Tabrisrp:
This plugin (bulletproof security) is going to also block requests from the core WP REST API with this code in the `.htaccess` 😕

I understand why they block the others, but this one doesn't make sense.

If we really want to change the method, maybe we can use the `PATCH` method instead? It's not ideal, but it's the closest one to `DELETE`.

I'm kind of worried of our take of "let's fix this in WP Rocket" all the time when there is a 3rd party issue though. We do this again and again, and it gets our codebase bigger and more complex as time go by.

GeekPress:
I'm agree with @Tabrisrp, let's go with the default behavior. If BPS is blocking the core WP REST API, it a choice from the user to block this. And then, to also block our CPCSS to properly work. We will indicate into our documentation the conflict which could happen with BPS and how they could fix it (maybe turn off an option into BPS).

GeekPress:
@crystinutzaa @hellofromtonya  You can consider the QA done. We've just validate that with @arunbasillal .

remyperona (approved):

PR #2617, opened 2020-05-05
Fixes "max execution time" when due to "path" matching

hellofromtonya:
This PR fixes the URL "path" matching by limiting it to exact word matches. We have customers who have experienced "max execution time" errors due to this specific problem.

How fixed? 

Adds `\b` to the `_rocket_get_entries_regex()` "path" regex.

This fix does not need QA. Why? @vmanthos has applied the change to an impacted customer and validated that it did fix the problem.

Note: The `max execution time` error may also be caused by other problems. This PR fixes the one bug we've validated thus far.

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1375,6 +1375,7 @@ function _rocket_get_cache_path_iterator( $cache_path ) { // phpcs:ignore WordPr
 /**
  * Gets the entries from the URL using RegexIterator.
  *
+ * @since  3.5.5 Limited regex search to exact "path" matches.
  * @since  3.5.4
  * @access private
  *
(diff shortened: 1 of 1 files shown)


hellofromtonya:
Instance of the problem and notes from @vmanthos:

>I had a ticket where the customer was getting a max_execution_time error when trying to delete a user. They didn't have the user cache enabled when I checked. 
>I applied the change in the RegEx [i.e. adding `\b` after the "path" in the regex] and the user was properly deleted.
>Here is the cache dir structure: https://i.vgy.me/eNAIeb.jpg
>ABSPATH : /var/www/html/

In this case, adding `\b` to ensure we exactly match a URL's "path" fixed the error for this customer.

hellofromtonya:
This solution is now invalid as we are taking a new approach.

PR #2622, opened 2020-05-06
Refactor CPCSS REST

engahmeds3ed:
# Refactor Engine CPCSS REST APIs files as follows:-

1. Create new class called APIClient which will be responsible for making API requests to wp-rocket cpcss endpoint jobs to do the following:- 
* Send generation request.
* Get job data.
* Check job status.
2. Create DataManager class which will be responsible for saving and deleting cpcss.
3. Create RESTWP abstract class which will be instantiated for posts, terms, home, ....etc.
4. Create RESTWPPost for handling posts.


inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+	 * Constant url for Critical Path API job.
+	 */
+	const API_URL = 'https://cpcss.wp-rocket.me/api/job/';
+
+	/**
+	 * Sends a generation request to the Critical Path API
+	 *
+	 * @since 3.6
+	 *
+	 * @param string $post_url The post URL.
```
Tabrisrp, line 23:
```suggestion
	 * @param string $url The URL to send a CPCSS generation request for.
```

The URL can be a post URL, but it can also be any kind of URL on a site, so it's better to be generic

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+	 * @return array
+	 */
+	public function send_generation_request( $post_url ) {
```
Tabrisrp, line 26:
```suggestion
	public function send_generation_request( $url ) {
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+		$response = wp_remote_post(
+			self::API_URL,
+			[
+				// This filter is documented in inc/Engine/CriticalPath/CriticalCSSGeneration.php.
+				'body' => apply_filters(
+					'rocket_cpcss_job_request',
+					[
+						'url' => $post_url,
```
Tabrisrp, line 34:
```suggestion
						'url' => $url,
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+					]
+				),
+			]
+		);
+
+		return $this->prepare_response( $response, $post_url );
```
Tabrisrp, line 40:
```suggestion
		return $this->prepare_response( $response, $url );
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	/**
+	 * Prepare the response to be returned.
+	 *
+	 * @param array|WP_Error $response The response or WP_Error on failure.
+	 * @param string         $post_url Url for the post to be checked.
```
Tabrisrp, line 47:
```suggestion
	 * @param string         $url URL to be checked.
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+	 * @return array|WP_Error
+	 * @since 3.6
+	 */
+	private function prepare_response( $response, $post_url ) {
```
Tabrisrp, line 51:
```suggestion
	private function prepare_response( $response, $url ) {
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+		$response_data        = $this->get_response_data( $response );
+		$response_status_code = $this->get_response_status( $response );
+		$succeeded            = $this->get_response_success( $response_status_code, $response_data );
+
+		if ( $succeeded ) {
+			return $response_data;
+		}else {
```
Tabrisrp, line 58: You don't need the `else` here, since if the first condition is successful, you `return`. It will stop the method execution.

It will make the code easier to read 👍

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+			$response_code    = $this->get_response_code( $response );
+			$response_message = $this->get_response_message( $response_status_code, $response_data, $post_url );
```
Tabrisrp, line 60:
```suggestion
			$response_message = $this->get_response_message( $response_status_code, $response_data, $url );
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+		return $this->prepare_response( $response, $post_url );
+	}
+
+	/**
+	 * Prepare the response to be returned.
+	 *
+	 * @param array|WP_Error $response The response or WP_Error on failure.
+	 * @param string         $post_url Url for the post to be checked.
+	 * @return array|WP_Error
+	 * @since 3.6
```
Tabrisrp, line 49: Add this before the `@param`

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+	 * @param array|WP_Error $response The response or WP_Error on failure.
+	 * @param null|int       $status Status code to overwrite the response status.
+	 * @return int|string status code|number of response.
+	 */
+	private function get_response_status( $response, $status = null ) {
+		if ( is_null( $status ) ) {
+			$status = wp_remote_retrieve_response_code( $response );
+		}
+
+		return $status;
```
Tabrisrp, line 109:
```suggestion
		return (int) $status;
```

Let's typehint the return to always return an `int`

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	/**
+	 * Get response message.
+	 *
+	 * @since 3.6
+	 *
+	 * @param int    $response_status_code Response status code.
+	 * @param array  $response_data Array of data returned from request.
+	 * @param string $item_url Url for the web page to be checked.
```
Tabrisrp, line 119:
```suggestion
	 * @param string $url URL to be checked.
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+	 * @return string
+	 */
+	private function get_response_message( $response_status_code, $response_data, $item_url ) {
```
Tabrisrp, line 122:
```suggestion
	private function get_response_message( $response_status_code, $response_data, $url ) {
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+		$message = '';
+
+		switch ( $response_status_code ) {
+			case 200:
+				if ( ! isset( $response_data->data->id ) ) {
+					$message .= sprintf(
+					// translators: %s = item URL.
+						__( 'Critical CSS for %1$s not generated. Error: The API returned an empty response.', 'rocket' ),
+						$item_url
```
Tabrisrp, line 131:
```suggestion
						$url
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+					);
+				}
+				break;
+			case 400:
+			case 440:
+			case 404:
+				// translators: %s = item URL.
+				$message .= sprintf( __( 'Critical CSS for %1$s not generated.', 'rocket' ), $item_url );
```
Tabrisrp, line 139:
```suggestion
				$message .= sprintf( __( 'Critical CSS for %1$s not generated.', 'rocket' ), $url );
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+				break;
+			default:
+				$message .= sprintf(
+				// translators: %s = post URL.
+					__( 'Critical CSS for %1$s not generated. Error: The API returned an invalid response code.', 'rocket' ),
+					$item_url
```
Tabrisrp, line 145:
```suggestion
					$url
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+		return 'cpcss_generation_failed';
+	}
+
+	/**
+	 * Get job details by calling API with job ID.
+	 *
+	 * @since 3.6
+	 *
+	 * @param string $job_id ID for the job to get details.
+	 * @param string $item_url URL for item to be used in error messages.
```
Tabrisrp, line 189:
```suggestion
	 * @param string $url URL to be used in error messages.
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+	 * @return mixed|WP_Error Details for job.
+	 */
+	public function get_job_details( $job_id, $item_url ) {
```
Tabrisrp, line 192:
```suggestion
	public function get_job_details( $job_id, $url ) {
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+		$response = wp_remote_get(
+			self::API_URL . "{$job_id}/"
+		);
+
+		return $this->prepare_job_details_response( $response, $item_url );
```
Tabrisrp, line 197:
```suggestion
		return $this->prepare_job_details_response( $response, $url );
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	/**
+	 * Prepares Job details response to be returned.
+	 *
+	 * @since 3.6
+	 *
+	 * @param array|WP_Error $response The response or WP_Error on failure.
+	 * @param string         $item_url URL for item to be used in error messages.
```
Tabrisrp, line 206:
```suggestion
	 * @param string         $url URL to be used in error messages.
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+	 * @return mixed|WP_Error
+	 */
+	private function prepare_job_details_response( $response, $item_url ) {
```
Tabrisrp, line 209:
```suggestion
	private function prepare_job_details_response( $response, $url ) {
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,234 @@
...
+		$response_data        = $this->get_response_data( $response );
+		$response_status_code = $this->get_response_status( $response, ( isset( $response_data->status ) ) ? $response_data->status : null );
+		$succeeded            = $this->get_response_success( $response_status_code, $response_data );
+
+		if ( $succeeded ) {
+			return $response_data;
+		}else {
```
Tabrisrp, line 216: Same here you don't need the `else`

inc/Engine/CriticalPath/APIClient.php
engahmeds3ed: Ok but this method wp_remote_retrieve_response_code returns either the number of status code or empty string and phpdoc for it contains that it returns 
```
@return int|string The response code as an integer. Empty string on incorrect parameter given.
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,230 @@
...
+		$response_code    = $this->get_response_code( $response );
+		$response_message = $this->get_response_message( $response_status_code, $response_data, $url );
+
+		if ( 200 === $response_status_code ) {
+			$response_status_code = 400;
+		}
+
+		return new WP_Error(
+			$response_code,
+			$response_message,
```
hellofromtonya, line 65: The response code and message getters can be called directly in the `new WP_Error` instead of first assigning them to vars. Why? 

- Readability and maintainability are the same by using the methods without the variables.
The method names are very clear as to what will be returned, i.e. code or message. 

- Variables not needed.
The variables are not being used anywhere else. Rather, they are only used as args for `WP_Error`.

- More performant. Why? Less memory and processing time. How? 
Let's explore how PHP internals work:
    - After `get_response_code()` runs, the code it returns is added into the memory stack.
    - Next, PHP adds the variable into the symtable (i.e. symbols table - a separate lookup in memory).
    - Next, PHP binds the variable in the symtable to the code's memory location in the stack.
    - Repeat for the message.
    - Then PHP looks up the variable in the symtable and then the code in memory to populate the first arg in `WP_Error`.
    - Repeat for the message and 2nd arg.

```suggestion
		if ( 200 === $response_status_code ) {
			$response_status_code = 400;
		}

		return new WP_Error(
			$this->get_response_code( $response ),
			$this->get_response_message( $response_status_code, $response_data, $url ),
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	/**
+	 * Get the status of response.
+	 *
+	 * @param int   $response_code Response code to check success or failure.
+	 * @param array $response_data Array of data returned from request.
+	 * @return bool success or failed.
+	 * @since 3.6
+	 */
```
hellofromtonya, line 79:
```suggestion
	 * Get the status of response.
	 *
	 * @since 3.6
	 *
	 * @param int   $response_code Response code to check success or failure.
	 * @param array $response_data Array of data returned from request.
	 * @return bool success or failed.
	 */
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,230 @@
...
+	private function get_response_success( $response_code, $response_data ) {
+		return (
+			200 === $response_code &&
+			! empty( $response_data ) &&
+			(
+				( isset( $response_data->status ) && 200 === $response_data->status ) ||
+				( isset( $response_data->data ) && isset( $response_data->data->id ) )
+			)
+		);
```
hellofromtonya, line 88: To help the conditional chain be more quickly understandable, format it like this:

```suggestion
		return (
			200 === $response_code 
			&&
			! empty( $response_data )
			&&
			(
				( 
					isset( $response_data->status ) 
					&& 
					200 === $response_data->status 
				) 
				||
				( 
					isset( $response_data->data ) 
					&& 
					isset( $response_data->data->id ) 
				)
			)
		);
```

Notice how you can quickly see the relationships.

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,230 @@
...
+	 * @param null|int       $status Status code to overwrite the response status.
+	 * @return int|string status code|number of response.
+	 */
+	private function get_response_status( $response, $status = null ) {
+		if ( is_null( $status ) ) {
+			$status = wp_remote_retrieve_response_code( $response );
+		}
+
+		return (int) $status;
+	}
```
hellofromtonya, line 106: Overwriting the variable is a little less performant here. Instead, make the conditional a guard clause to bail out but also type cast the status to an integer data type.

Notice with this design that if it's `null`, we can directly return from the function call and type casting. No need to assign the status code to the variable and then look it up again. Little less memory and processing time.

```suggestion
	private function get_response_status( $response, $status = null ) {
		if ( ! is_null( $status ) ) {
			return (int) $status;
		}

		return (int) wp_remote_retrieve_response_code( $response );
	}
```

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,230 @@
...
+
+	/**
+	 * Get response status code/number.
+	 *
+	 * @since 3.6
+	 *
+	 * @param array|WP_Error $response The response or WP_Error on failure.
+	 * @param null|int       $status Status code to overwrite the response status.
+	 * @return int|string status code|number of response.
```
hellofromtonya, line 98: The returned value will be an integer only.

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,230 @@
...
+		return (int) $status;
+	}
+
+	/**
+	 * Get response message.
+	 *
+	 * @since 3.6
+	 *
+	 * @param int    $response_status_code Response status code.
+	 * @param array  $response_data Array of data returned from request.
```
hellofromtonya, line 114: The response data is shown as an object below. Can it be an array or object? If it's an object, what is it's data type, i.e. instance of ______? Add that to the `@param`.

inc/Engine/CriticalPath/APIClient.php
```diff
@@ -0,0 +1,230 @@
...
+			$response_status_code = 400;
+		}
+
+		return new WP_Error(
+			$response_code,
+			$response_message,
+			[
+				'status' => $response_status_code,
+			]
+		);
```
hellofromtonya, line 227: This code is exactly the same as the code in `prepare_response()` method above. Refactor into a private method and then invoke it in both methods. Why? DRY.

inc/Engine/CriticalPath/DataManager.php
```diff
@@ -0,0 +1,135 @@
...
+	 */
+	private $critical_css_path;
+
+	/**
+	 * DataManager constructor, adjust the critical css path for posts.
+	 *
+	 * @param string $critical_css_path path for main critical css folder.
+	 */
+	public function __construct( $critical_css_path ) {
+		$this->critical_css_path = $critical_css_path . get_current_blog_id() . DIRECTORY_SEPARATOR;
```
hellofromtonya, line 26: Add a `$filesystem` property and set it here in the constructor.

```php
$this->filesystem = rocket_direct_filesystem();
```

Why? It's being used over and over again. Each time we invoke `rocket_direct_filesystem()`, it takes memory and time to instantiate a new `WP_Direct_Filesystem` object. We can invoke once, store in a property in the object, and then use the property throughout the object.

Note: When we build a `Filesystem` module, then we can inject it as a dependency. We could inject _now_ in the ServiceProvider.

inc/Engine/CriticalPath/DataManager.php
```diff
@@ -0,0 +1,135 @@
...
+	 * Save CPCSS into file.
+	 *
+	 * @since 3.6
+	 *
+	 * @param string $path Path for cpcss file related to this web page.
+	 * @param string $cpcss CPCSS code to be saved.
+	 * @return bool
+	 */
+	public function save_cpcss( $path, $cpcss ) {
+		if ( ! rocket_direct_filesystem()->is_dir( $this->critical_css_path ) ) {
```
hellofromtonya, line 39:
```suggestion
		if ( ! $this->filesystem->is_dir( $this->critical_css_path ) ) {
```

inc/Engine/CriticalPath/DataManager.php
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	/**
+	 * Delete critical css file by path.
+	 *
+	 * @param string $path critical css file path to be deleted.
+	 * @return bool|WP_Error
+	 */
+	public function delete_cpcss( $path ) {
+		$filesystem = rocket_direct_filesystem();
```
hellofromtonya, line 55: Not needed anymore. Please delete.

inc/Engine/CriticalPath/DataManager.php
```diff
@@ -0,0 +1,135 @@
...
+
+		$full_path = $this->critical_css_path . $path;
+
+		if ( ! $filesystem->exists( $full_path ) ) {
```
hellofromtonya, line 59:
```suggestion
		if ( ! $this->filesystem->exists( $full_path ) ) {
```

inc/Engine/CriticalPath/DataManager.php
```diff
@@ -0,0 +1,135 @@
...
+			return new WP_Error(
+				'cpcss_not_exists',
+				__( 'Critical CSS file does not exist', 'rocket' ),
+				[
+					'status' => 400,
+				]
+			);
+		}
+
+		if ( ! $filesystem->delete( $full_path ) ) {
```
hellofromtonya, line 69:
```suggestion
		if ( ! $this->filesystem->delete( $full_path ) ) {
```

inc/Engine/CriticalPath/APIClient.php
engahmeds3ed: @hellofromtonya you are right and thanks for illustration, I will do that for response code but I can't do that for the message because of:-
- I need to call `get_response_message` before normalizing the value of `$response_status_code`.
- This helps me in the case of the response code is 200 but the response body is empty so here I want to pass the code 200 and inside the method `get_response_message` I can check if the code is 200 and the response body is empty get the corresponding message but the response code in this case should return 400 so I'm changing it after getting the message.

inc/Engine/CriticalPath/APIClient.php
engahmeds3ed: I merged two methods into one.

inc/Engine/CriticalPath/DataManager.php
engahmeds3ed: I think for now we can't do that because of the following:-
1. Plugin tries to call load method here https://github.com/wp-media/wp-rocket/blob/4f40095d616ad3e3f485d2580e037e731b5bb3a6/inc/main.php#L39-L40
2. Which loads the service provider for cpcss https://github.com/wp-media/wp-rocket/blob/253905f6ad3207b32a182ec67410e75e8b2f3b7a/inc/classes/class-plugin.php#L70
3. And inside the service provider it tries to get an instance of DataManager class which inside constructor call the function `rocket_direct_filesystem`
https://github.com/wp-media/wp-rocket/blob/253905f6ad3207b32a182ec67410e75e8b2f3b7a/inc/Engine/CriticalPath/ServiceProvider.php#L51
4. Then after that it tries to load files.php which contains the definition for the function `rocket_direct_filesystem` https://github.com/wp-media/wp-rocket/blob/4f40095d616ad3e3f485d2580e037e731b5bb3a6/inc/main.php#L44

So it will fail with `Fatal error: Uncaught Error: Call to undefined function`

hellofromtonya: We can. We need to move where `files.php` is being loaded. We want to load that right after composer's autoloader is loaded. Then it's available everyone. Let me create a PR for that to get it into `develop`. Then we can merge it here.

engahmeds3ed: Great, now we can :+1:

inc/Engine/CriticalPath/RESTWP.php
```diff
@@ -0,0 +1,480 @@
...
+		register_rest_route(
+			self::ROUTE_NAMESPACE,
+			'cpcss/' . $this->route_namespace . '/(?P<id>[\d]+)',
+			[
+				'methods'             => 'DELETE',
+				'callback'            => [ $this, 'delete' ],
+				'permission_callback' => [ $this, 'check_permissions' ],
+			]
+		);
+	}
```
Tabrisrp, line 89: This should be part of `RESTWPPost` instead, it's specific to the posts generation. Let's keep `RESTWP` as generic as possible

inc/Engine/CriticalPath/RESTWP.php
```diff
@@ -0,0 +1,480 @@
...
+			// get item url.
+			$item_url = $this->get_url( $item_id );
+			$timeout  = ( isset( $request['timeout'] ) && ! empty( $request['timeout'] ) );
+
+			$output = $this->process_generate( $item_url, $item_id, $timeout );
+		}else {
+			$output = $this->return_error( $validated );
+		}
+
+		return rest_ensure_response( $output );
```
Tabrisrp, line 126:
```suggestion
		if ( is_wp_error( $validated ) ) {
			return rest_ensure_response( $this->return_error( $validated ) );
		}

                // get item url.
                $item_url  = $this->get_url( $item_id );
	        $timeout  = ( isset( $request['timeout'] ) && ! empty( $request['timeout'] ) );
		$output   = $this->process_generate( $item_url, $item_id, $timeout );

		return rest_ensure_response( $output );
```

Return early when we can, most of the time an `else` is not needed. Here by inverting the conditional logic, we can return early when there is an error, and process the rest after the conditional if not.

inc/Engine/CriticalPath/RESTWP.php
```diff
@@ -0,0 +1,480 @@
...
+	 * Send Generation first request.
+	 *
+	 * @since 3.6
+	 *
+	 * @param string $item_url Url for item to send the generation request for.
+	 * @param int    $item_id ID for item to send the generation request for.
+	 * @return array
+	 */
+	private function send_generation_request( $item_url, $item_id ) {
+		$output = null;
```
Tabrisrp, line 166: what is this line for?

inc/Engine/CriticalPath/RESTWP.php
```diff
@@ -0,0 +1,480 @@
...
+	 * @return array|mixed|WP_Error
+	 */
+	private function get_cpcss_job_details( $job_id, $item_url ) {
+		$job_details = $this->api_client->get_job_details( $job_id, $item_url );
+
+		if ( is_wp_error( $job_details ) ) {
+			return $this->return_error( $job_details );
+		}else {
+			return $job_details;
+		}
```
Tabrisrp, line 200:
```suggestion
		}

		return $job_details;

```

Here the `else is not needed either, you already returned in the conditional if it was true

inc/Engine/CriticalPath/RESTWP.php
```diff
@@ -0,0 +1,480 @@
...
+
+		// validate item.
+		$validated = $this->validate_item_for_delete( $item_id );
+		if ( ! is_wp_error( $validated ) ) {
+			$output = $this->process_delete( $item_id );
+		}else {
+			$output = $this->return_error( $validated );
+		}
+
+		return rest_ensure_response( $output );
```
Tabrisrp, line 376:
```suggestion
		if ( is_wp_error( $validated ) ) {
			rest_ensure_response( $this->return_error( $validated ) );
		}

                $output = $this->process_delete( $item_id );
		return rest_ensure_response( $output );
```

tests/Integration/inc/Engine/CriticalPath/APIClient/sendGenerationRequest.php
```diff
@@ -0,0 +1,75 @@
...
+use WP_Error;
+
+/**
+ * @covers \WP_Rocket\Engine\CriticalPath\APIClient::send_generation_request
+ * @group CriticalPath
+ * @group  vfs
+ */
+class Test_SendGenerationRequest extends FilesystemTestCase {
+
+	protected $path_to_test_data = '/inc/Engine/CriticalPath/APIClient/sendGenerationRequest.php';
```
hellofromtonya, line 16: Is the filesystem used in the `APIClient::send_generation_request() method or its dependencies? 

If no:

1. Switch this test case to `TestCase`
2. Remove the property `$path_to_test_data`
3. Change the data provider to `configTestData`
4. Remove the vfs and `test_data` from the test data fixture, i.e. make it an array of test data

tests/Integration/inc/Engine/CriticalPath/APIClient/sendGenerationRequest.php
```diff
@@ -0,0 +1,75 @@
...
+			->atMost()
+			->times( 1 )
+			->with( 'postRequest' )
+			->andReturn( $response_code );
+
+		Functions\expect( 'wp_remote_retrieve_body' )
+			->atMost()
+			->times( 1 )
+			->with( 'postRequest' )
+			->andReturn( $response_body );
```
hellofromtonya, line 49: Instead of `atMost, be specific to validate the expected behavior with these mocks. [Use this method as a guide.](https://github.com/wp-media/wp-rocket/blob/feature/cpcss-refactor/tests/Unit/inc/Engine/CriticalPath/APIClient/getJobDetails.php#L41)

tests/Unit/inc/Engine/CriticalPath/APIClient/sendGenerationRequest.php
```diff
@@ -0,0 +1,82 @@
...
+use WP_Rocket\Tests\Unit\FilesystemTestCase;
+use Brain\Monkey\Functions;
+use WP_Error;
+
+/**
+ * @covers \WP_Rocket\Engine\CriticalPath\APIClient::send_generation_request
+ * @group CriticalPath
+ * @group  vfs
+ */
+class Test_SendGenerationRequest extends FilesystemTestCase {
```
hellofromtonya, line 14: Same comments in this test class as its integration test.

inc/Engine/CriticalPath/CPCSSService.php
```diff
@@ -0,0 +1,256 @@
...
+
+		// For pending job status.
+		if ( isset( $job_state ) && 'complete' !== $job_state ) {
+			return $this->on_job_pending( $item_url );
+		}
+
+		// For successful job status.
+		if (
+			isset( $job_state, $job_details->data->critical_path ) &&
+			'complete' === $job_state
```
hellofromtonya, line 140: As our standard, let's build these types of long conditions like this:

```suggestion
			isset( $job_state, $job_details->data->critical_path ) 
			&&
			'complete' === $job_state
```

Why? It's easier to read. You see that's an AND and each line is a conditional.

inc/Engine/CriticalPath/RESTCSSSubscriber.php
```diff
@@ -0,0 +1,55 @@
...
+	private $rest_post_manager;
+
+	/**
+	 * RESTCSSSubscriber constructor.
+	 *
+	 * @param RESTWPPost $rest_post_manager Post manager instance.
+	 */
+	public function __construct( RESTWPPost $rest_post_manager ) {
+		$this->rest_post_manager = $rest_post_manager;
+	}
```
hellofromtonya, line 28: This implementation is too restrictive as it's binds only the `RESTWPPost` object to this subscriber. Instead, let's use an interface and change the naming here:

```suggestion
	/**
	 * RESTCSSSubscriber constructor.
	 *
	 * @param RESTWPInterface $rest_manager Post manager instance.
	 */
	public function __construct( RESTWPInterface $rest_manager ) {
		$this->rest_manager = $rest_manager;
	}
```

inc/Engine/CriticalPath/RESTWPPost.php
```diff
@@ -0,0 +1,102 @@
...
+namespace WP_Rocket\Engine\CriticalPath;
+
+use WP_Error;
+
+/**
+ * Class RESTWPPost
+ *
+ * @package WP_Rocket\Engine\CriticalPath
+ */
+class RESTWPPost extends RESTWP {
```
hellofromtonya, line 12:
```suggestion
class RESTWPPost extends RESTWP implements RESTWPInterface {
```

PR #2623, opened 2020-05-06
Fix #2618 for rocket_clean_files() by running RegexIterator once.

### LINKED ISSUE #2618: "max execution time" errors in 3.5.4 version of rocket_clean_files()
**Describe the bug**
Customers are reporting `max execution time` errors in the 3.5.4 version of `rocket_clean_files()`. 

**To Reproduce**
So far, we've not been able to reproduce the problem.

**Additional context**
In version 3.5.4, we upgraded `rocket_clean_files()` to use PHP SPL iterators instead of `glob()`. 

We are working with L2 to identify when this error occurs, it's contributors, and what may be causing it. More information will be coming as we learn more.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

hellofromtonya:
This PR does the following:

- Redesigns the processing strategy for `rocket_clean_files()` (details below)
- Optimizes its processing by running only the tasks that need to be run, i.e. instead of running them for every given URL.

Results:

- Processing time: `glob()` (<= WPR 3.5.3) === this version with SPL instead of `glob`.
- It also provides scaffolding to further optimize all of the cleaning functionality.

## Background for Context

The previous design used `glob()` (<= v 3.5.3). In WPR 3.5.4, we replaced `glob` with SPL `RegexIterator`. Both of these approaches crawl the filesystem using a wildcard or regex. Why? We need to grab the domain + any user caches for the given URLs to be cleaned. 

For example, let's say the domain is `example.org` and there are 3 user caches for this domain. The cache structure might look like this:

```
cache
	wp-rocket
		example.org
			// cache files/dirs
		example.org-user1-123456
			// cache files/dirs
		example.org-user2-123456
			// cache files/dirs
		example.org-user3-123456
			// cache files/dirs
```

Using a wildcard/regex method allows us to grab all both the domain and all of the user caches. 

But this is an expensive operation. We saw the impact of this when we switched to SPL `RegexIterator`. 

Why? The original design did the wildcard/regex crawl on each and every given URL. When there were 30, 50, 100 URLs given, it took way too long to process these, resulting in timeout errors for some customers. For others, it was visually taking longer to process.

## New Design

The new strategy does the following:

- Only runs the wildcard/regex crawl once for each given URL domain host. 
- It stores these directories, which will be in root of `cache/wp-content/`, in memory.
     - Note: Later we will optimize this further to store in db or via mapping module.
- It then runs each URL with each of these root level cache directories without crawling the filesystem. 
     - If the URL's file/directory does not exist in the cache, it skips and goes to the next one.
     - If it's a directory, it runs `rocket_rrmdir()`.
     - Else, it's a file. It deletes it without running `rocket_rrmdir()`.

## TODO

- [x] Optimization: Add new design
- [x] Optimization: Replace recursive directory and iterator in `_rocket_get_cache_dirs()`
- [x] Optimization: Allow the instance of the filesystem to be passed into `rocket_clean_files()` and `rocket_rrmdir()`
- [x] Optimization: Don't apply `*` after domain's host.
- [x] Add more the test scenarios
- [x] Record benchmarks 🚀 
- [x] Run live tests on our dev test server 🚀 
- [x] Have @vmanthos or @piotrbak  add the changes to customers who are reporting the problem. Validate the design fixes the problem and runs as fast as the WPR 3.5.2.
- [x] QA

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1463,11 +1463,6 @@ function _rocket_get_cache_dirs( $url_host, $cache_path = '', $hard_reset = fals
 			new FilesystemIterator( $cache_path )
 		);
 	} catch ( Exception $e ) {
-		if ( isset( $GLOBALS['debug_fs'] ) ) {
-			echo "\n Iterator error: {$e->getMessage()} \n"; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-		}
-
-		// No logging yet.
 		return [];
 	}
 
@@ -1476,20 +1471,12 @@ function _rocket_get_cache_dirs( $url_host, $cache_path = '', $hard_reset = fals
 	try {
 		$entries = new RegexIterator( $iterator, $regex );
 	} catch ( Exception $e ) {
-		if ( isset( $GLOBALS['debug_fs'] ) ) {
-			echo "\n RegexIterator error: {$e->getMessage()} \n"; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-		}
-
 		return [];
 	}
 
 	$domain_dirs[ $url_host ] = [];
 	foreach ( $entries as $entry ) {
 		$domain_dirs[ $url_host ][] = $entry->getPathname();
-	}
-
-	if ( isset( $GLOBALS['debug_fs'] ) ) {
-		var_dump( $domain_dirs ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped, WordPress.PHP.DevelopmentFunctions.error_log_var_dump
 	}
 
 	return $domain_dirs[ $url_host ];
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1432,7 +1432,7 @@ function _rocket_get_entries_regex( Iterator $iterator, $url, $cache_path = '' )
 /**
  * Gets the directories for the given URL host from the cache/wp-rocket/ directory or stored memory.
  *
- * @since 3.5.5
+ * @since  3.5.5
  * @access private
  *
  * @param string $url_host   The URL's host.
@@ -1441,7 +1441,7 @@ function _rocket_get_entries_regex( Iterator $iterator, $url, $cache_path = '' )
  *
  * @return array
  */
-function _rocket_get_cache_dirs( $url_host, $cache_path = '', $hard_reset = false ) {
+function _rocket_get_cache_dirs( $url_host, $cache_path = '', $hard_reset = false ) { // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedFunctionFound
 	static $domain_dirs = [];
 
 	if ( true === $hard_reset ) {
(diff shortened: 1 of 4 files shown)


hellofromtonya:
**Benchmarks**

Total time to clean 90 File and Directory URLs:

| WPR Version | Seconds |
| --- | ----------- |
| This PR | 0.008s |
| 3.5.2 |  2.692s |

Total time to clean 30 File URLs:

| WPR Version | Seconds |
| --- | ----------- |
| This PR | 0.003s |
| 3.5.2 |  0.214s |

Total time to clean 30 Single Directory (each having 2 files) URLs:

| WPR Version | Seconds |
| --- | ----------- |
| This PR | 0.005s |
| 3.5.2 |  0.636s |

Total time to clean 30 Nested Directory (each having 2 subdirs and 6 files) URLs:

| WPR Version | Seconds |
| --- | ----------- |
| This PR | 0.003s |
| 3.5.2 |  2.045s |

vmanthos:
I've used the `files.php` of this PR on a website where moving posts to and from the bin lead to a "[mod_fcgid: read data timeout](https://support.plesk.com/hc/en-us/articles/213374769-Website-or-webmail-shows-500-internal-server-error-on-Plesk-mod-fcgid-read-data-timeout)" error in the log.

The issue was fixed! 🎉

**Ticket:** https://secure.helpscout.net/conversation/1155786902/162916?folderId=2135277**

inc/functions/files.php
```diff
@@ -1417,3 +1428,56 @@ function _rocket_get_entries_regex( Iterator $iterator, $url, $cache_path = '' )
...
+
+	try {
+		$iterator = new IteratorIterator(
+			new FilesystemIterator( $cache_path )
+		);
+	} catch ( Exception $e ) {
+		return [];
+	}
+
+	$regex = sprintf( '/%s%s(.*)/i', str_replace( '/', '\/', $cache_path ), $url_host );
```
Tabrisrp, line 1469:
```suggestion
	$regex = sprintf( '/%1$s%2$s(.*)/i', str_replace( '/', '\/', $cache_path ), $url_host );
```

What do you think?

hellofromtonya:
@wp-media/dev-qa @piotrbak @arunbasillal This PR is ready for QA.

piotrbak:
I'm testing it on our customers' sites now. Later, I'll move to testing it on environments created locally.

hellofromtonya:
**QA Update**

@vmanthos and @piotrbak added this PR to 4 customer sites who had the issue. Here are the cache sizes for these customers:

- ~ 60k cached files
- ~4k cached files
- ~250k cached files
- ~38k cached files

In all cases, the PR fixed the time-out issue and worked as expected.

hellofromtonya:
>@Tabrisrp We have a bug with multisite subdirectories. When clearing the cache of the Static Homepage of the main site, we'll clear the cache of all sites. It's happening with the older versions too though.

@piotrbak Can you open a new issue for this problem please? Thanks.

PR #2624, opened 2020-05-06
Adding two nonces to the array

piotrbak:
Adding nonces related to the Dokan


inc/ThirdParty/Plugins/Ecommerce/WooCommerceSubscriber.php
```diff
@@ -468,6 +468,8 @@ private function get_nonce_actions() {
 			'ajax_search_nonce', // xtra theme AJAX search.
 			'xtra_wishlist_content', // xtra theme wishlist feature.
 			'ajax-login-security', // OneSocial theme pop-up login.
+			'dokan_pageview', // Dokan related pageview
+			'dokan_report_abuse', // Dokan report abuse popup
```
hellofromtonya, line 472: @piotrbak Add periods (.) to the end of each comment. Our code standard checkers check for full-stops, ie meaning ends with a period.

piotrbak: Got it

PR #2625, opened 2020-05-07
Fixes #1747 Check paths early in advanced-cache.php

### LINKED ISSUE #1747: Detect if paths in advanced-cache.php are correct
Based on customer feedback: 

> WPR saves the absolute path of the WP installation in multiple lines of the file advanced-cache.php, for example:
> 
> if ( file_exists( 'public_html/blog/wp-content/plugins/wp-rocket/inc/vendors/classes/class-rocket-mobile-detect.php' ) && ! class_exists( 'Rocket_Mobile_Detect' ) ) {
> include_once 'public_html/blog/wp-content/plugins/wp-rocket/inc/vendors/classes/class-rocket-mobile-detect.php';
> }
> 
> We have a couple of incidents when the following happened:
> 1. we migrate a website to a different hosting provider
> 2. the website was working fine on the first examination
> 3. after a few days, the website homepage was just a blank page for non-logged users
> 4. the dashboard was working fine with no error message
> 4. nothing in error_log or on screen
> We found out that the problem was the old path in the advanced-cache.php file.

Would it be possible / useful to detect if the path is correct, and if not, then warn the user to deactivate / reactivate WP Rocket?

Not sure if there are many cases like this one, but it could be useful when migrating sites between environments (using automated migration tools for example), to avoid issues with incorrect paths.

Related HS ticket: 
https://secure.helpscout.net/conversation/866426504/109859?folderId=273761

Slack Thread: https://wp-media.slack.com/archives/C43T1AYMQ/p1559333955020300

Tabrisrp:
- [x] Create `AdvancedCache.php` class
- [x] Use `get_advanced_cache_content()` method in `rocket_generate_advanced_cache_file()`
- [x] Create a base `advanced-cache.php` file instead of having the output inside the function
- [x] Deprecate `get_rocket_advanced_cache_file`
- [x] Update base `advanced-cache.php` file to add early paths checks
- [x] Move related notices to the `AdvancedCache` class
- [x] Add callbacks for notices in `AdminSubscriber`
- [x] Write tests
- [x] QA


inc/Engine/Cache/AdvancedCache.php
```diff
@@ -0,0 +1,171 @@
...
+	 * @since 2.2
+	 *
+	 * @return void
+	 */
+	public function notice_content_not_ours() {
+		global $pagenow;
+
+		if (
+			'plugins.php' === $pagenow
+			&& isset( $_GET['activate'] ) // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
Screenfeed, line 128: What about this? No more phpcs comment.
```suggestion
			&& filter_has_var( INPUT_GET, 'activate' )
```

wp-rocket.php
```diff
@@ -33,6 +33,10 @@
 define( 'WP_ROCKET_PATH',                  realpath( plugin_dir_path( WP_ROCKET_FILE ) ) . '/' );
 define( 'WP_ROCKET_INC_PATH',              realpath( WP_ROCKET_PATH . 'inc/' ) . '/' );
 
+if ( ! defined( 'WP_ROCKET_VIEWS' ) ) {
+	define( 'WP_ROCKET_VIEWS', WP_ROCKET_PATH . 'views' );
```
Screenfeed, line 37: All of the other folder paths have a trailing slash, I think it should be the same for this one imho.

inc/Engine/Cache/AdvancedCache.php
```diff
@@ -0,0 +1,171 @@
...
+	 * Gets the content for the advanced-cache.php file
+	 *
+	 * @since 3.6
+	 *
+	 * @return string
+	 */
+	public function get_advanced_cache_content() {
+		$content = $this->filesystem->get_contents( $this->template_path . '/advanced-cache.php' );
+		$mobile  = is_rocket_generate_caching_mobile_files() ? '$2' : '';
+		$content = preg_replace( "/('{{MOBILE_CACHE}}';)(\X*)('{{\/MOBILE_CACHE}}';)/", $mobile, $content );
```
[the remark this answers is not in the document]
Tabrisrp: I tried with `.*` but it was not working, `\X*` does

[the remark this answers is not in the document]
Screenfeed: Oh, strange 🤔

hellofromtonya:
@Tabrisrp Let's talk about this tomorrow when we meet. I'd like to walkthrough it to ensure I understand.

inc/functions/files.php
```diff
@@ -1,96 +1,19 @@
...
- * Creates advanced-cache.php file.
+ * Creates the advanced-cache.php file.
  *
+ * @since 3.6 Uses AdvancedCache::get_advanced_cache_content().
  * @since 2.0
+ *
+ * @param AdvancedCache $advanced_cache Optional. Instance of the advanced cache handler.
  */
-function rocket_generate_advanced_cache_file() {
+function rocket_generate_advanced_cache_file( $advanced_cache = null ) {
```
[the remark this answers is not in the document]
hellofromtonya: The Plan:
Remy and I talked about this. This function and any others for advanced cache will be refactored later into the `AdvancedCache` file or module. Any functions that might be used by 3rd parties will be moved to the `inc/API` and wired to their new OOP versions, i.e. instead of deprecating the function.

Tabrisrp:

Things to test:
- Make sure the `advanced-cache.php` file is correctly generated on activation, when updating WP Rocket settings, and especially when migrating a site between different environments (like staging/live, or different local)
- Check the notices are correctly displayed if no permissions to write, or the file content is not containing our data

I've asked JB to re-enable our Pressable access, if it happens we will also need a QA on Pressable to confirm the notices don't display there.

inc/ThirdParty/ServiceProvider.php
```diff
@@ -73,6 +75,9 @@ public function register() {
 		$this->getContainer()->share( 'amp_subscriber', 'WP_Rocket\ThirdParty\Plugins\Optimization\AMP' )
 			->withArgument( $options )
 			->withArgument( $this->getContainer()->get( 'cdn_subscriber' ) );
+		$this->getContainer()->share( 'pressable_subscriber', 'WP_Rocket\ThirdParty\Hostings\Pressable' )
+			->withArgument( $this->getContainer()->get( 'admin_cache_subscriber' ) );
+		$this->getContainer()->share( 'litespeed_subscriber', 'WP_Rocket\Subscriber\Third_Party\Hostings\Litespeed_Subscriber' );
```
crystinutzaa, line 80: Shouldn't we also move LiteSpeed in the new structure?
If everybody is ok to keep it like this, then it's fine for me. However I am thinking if things will not mess up in this way. 🤔

hellofromtonya: @crystinutzaa I agree with you. We should. Let's do that as part of a separate PR once this one is merged. Maybe something @engahmeds3ed can work on????

crystinutzaa: @hellofromtonya & @Tabrisrp I approved it 🎉

arunbasillal:
QA: 

- Switched to the fix branch. Deactivated WP Rocket.
- Deleted `advanced-cache.php`. Activated WP Rocket. Got the following error. 

```
[20-May-2020 10:30:14 UTC] PHP Fatal error:  Uncaught Error: Call to a member function get() on null in /app/public/wp-content/plugins/wp-rocket/inc/functions/files.php:30
Stack trace:
#0 /app/public/wp-content/plugins/wp-rocket/inc/main.php(219): rocket_generate_advanced_cache_file()
#1 /app/public/wp-includes/class-wp-hook.php(287): rocket_activation(false)
#2 /app/public/wp-includes/class-wp-hook.php(311): WP_Hook->apply_filters('', Array)
#3 /app/public/wp-includes/plugin.php(478): WP_Hook->do_action(Array)
#4 /app/public/wp-admin/includes/plugin.php(690): do_action('activate_wp-roc...', false)
#5 /app/public/wp-admin/plugins.php(44): activate_plugin('wp-rocket/wp-ro...', 'http://wp-rocke...', false)
#6 {main}
  thrown in /app/public/wp-content/plugins/wp-rocket/inc/functions/files.php on line 30
```

Tabrisrp:
@hellofromtonya @crystinutzaa Arun caught this issue on activation. This is because on activation callback, the container is not set, so I passed the instance of `AdvancedCache` directly to `rocket_generate_advanced_cache_file()` in this commit: https://github.com/wp-media/wp-rocket/pull/2625/commits/e86123b3e136098f70594b9895ee1cc4e8633ee5

This is the only place in the plugin where the function is called outside of a context where the container is available.

@arunbasillal This should be working now, I tested on my local and activation was ok with this change.

hellofromtonya:
>@hellofromtonya @crystinutzaa Arun caught this issue on activation ☝️. This is because on activation callback, the container is not set, so I passed the instance of AdvancedCache directly to rocket_generate_advanced_cache_file() in this commit: e86123b

Thinking out-loud. 

Creating an instance of `AdvancedCache` requires knowledge of its dependencies.  Do we anticipate needing this in other places? If yes, we should move that knowledge to a helper function in the `inc/API` and let it create the instance of `AdvancedCache` and return it. If no, then it's a one-time use case.

arunbasillal:
@hellofromtonya Should I wait for the [discussion](https://github.com/wp-media/wp-rocket/pull/2625#issuecomment-631437213) to be over before I do the QA?

Tabrisrp:
@hellofromtonya I see it as a one-time use case, on other contexts the container is available and we can pull it directly from there.

On deactivation we don't need the `AdvancedCache` instance, as we directly write an empty value in the file.

hellofromtonya:
@arunbasillal You can proceed with QA. Thanks!

arunbasillal:
QA

- `advanced-cache.php` (ACP) is generated correctly on activation on local. 
- On mega site, I kept getting writing permission notice:
image
- No issues with the current master branch on mega. 

To replicate on Mega:
- uninstall current WP Rocket
- Make sure `advanced-cache.php` is deleted (I had to manually delete it)
- Install this branch. On activation you should see the notice. 

I recorded it here: https://youtu.be/GEHUwtOdJRM
If you look closely at [0:44](https://youtu.be/GEHUwtOdJRM?t=44), `advanced-cache.php` does have   some contents which is then later on cleared for some reason.

@Tabrisrp Please take a look.

Tabrisrp:
@arunbasillal Thank you, I found the issue (a missing separator in the path to the template file).

Fixed here, and also fixed on mega.

arunbasillal:
QA

On Local: 

- Simulated and upgrade. advanced-cache.php was rewritten on upgrade. 👍  
- Deleted rewrite rules from .htaccess. Cache was served correctly. 👍 
- advanced-cache.php updated corrected when Separate cache for mobile is enabled or disabled 👍 
- Manually edited `$rocket_path`, `$rocket_config_path` and `$rocket_cache_path` to wrong values. Refreshed WordPress admin and the path was reset correctly 👍 
- Deactivated WP Rocket and the file was emptied out correctly. 👍 

On mega site: 

- advanced-cache.php created correctly on activation. 👍 
- Updated correctly when separate cache for mobile was enabled. 👍 

- However, when file permission was set to 444, no notice was displayed - https://youtu.be/jm9TvmZLMNc
- Deleted contents of advanced-cache.php. No notice was displayed. 

@Tabrisrp please note ^

PR #2626, opened 2020-05-07
Closes #2471 - Clean WP Rocket cache when RocketCDN REST Endpoints are called

### LINKED ISSUE #2471: Cache is not cleaned when updating CDN options via REST API
**Describe the bug**
RocketCDN Rest API endpoints do not refresh WP Rocket cache while activating / deactivating RocketCDN. At disable the website will be broken because it still have the CDN URL in the cache files.

**Reproduce the problem** ✅ 
Issue reproduced

**Identify the root cause** ✅ 
it’s not auto-cleaning here because the callback function that does it in a regular context (ie from WP Rocket settings page), is only loaded when is_admin() is true, which won’t be for a REST request.

**Scope a solution** ✅ 
Add a call to `rocket_clean_domain()` in the enable/disable callbacks methods of `CDNOptionsManager`

**Estimate the effort** ✅ 
Effort [S]

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
**Describe the bug**
RocketCDN Rest API endpoints do not refresh WP Rocket cache while activating / deactivating RocketCDN. At disable the website will be broken because it still have the CDN URL in the cache files.

**Reproduce the problem** ✅ 
Issue reproduced

**Identify the root cause** ✅ 
it’s not auto-cleaning here because the callback function that does it in a regular context (ie from WP Rocket settings page), is only loaded when is_admin() is true, which won’t be for a REST request.

**Scope a solution** ✅ 
Add a call to `rocket_clean_domain()` in the enable/disable callbacks methods of `CDNOptionsManager`

**Estimate the effort** ✅ 
Effort [S]

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

TO DO:
- [x] Revisit Unit Tests
- [x] Revisit Integration Tests

### DIFF
--- a/inc/Engine/CDN/RocketCDN/CDNOptionsManager.php
+++ b/inc/Engine/CDN/RocketCDN/CDNOptionsManager.php
@@ -51,6 +51,7 @@ class CDNOptionsManager {
 		$this->options_api->set( 'settings', $this->options->get_options() );
 
 		delete_transient( 'rocketcdn_status' );
+		rocket_clean_domain();
 	}
 
 	/**
@@ -69,5 +70,6 @@ class CDNOptionsManager {
 
 		delete_option( 'rocketcdn_user_token' );
 		delete_transient( 'rocketcdn_status' );
+		rocket_clean_domain();
 	}
 }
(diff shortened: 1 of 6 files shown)


hellofromtonya:

PR #2627, opened 2020-05-08
Closes #2548 Enable CDN for images on AMP pages

### LINKED ISSUE #2548: Set the CDN to be active for images on AMP pages
**Describe the bug**
The CDN rewrite is not working on the AMP pages
Currently, the rewrite is not expected since we remove all optimizations in the AMP pages.

**To Reproduce**
1. Install the [AMP](https://wordpress.org/plugins/amp/) plugin and enable the CDN in WP Rocket
2. Enable the _Transitional_ template in _AMP -> General_ settings.
3. Visit an AMP version of a page
4. Check the source of the images in the developer tools (the CDN rewrite is not applied)

**Expected behavior**
The CDN rewrite could be active for images only.

**Screenshots**
image
_Full-size image: https://share.getcloudapp.com/qGudyY4v_ 

￼image
_Full-size image: https://share.getcloudapp.com/wbu0rv0r_
 
**Additional context**
Ticket: https://secure.helpscout.net/conversation/1137020355/158076/
Slack thread: https://wp-media.slack.com/archives/C08N8J6VC/p1587049067360300
Related plugin: https://wordpress.org/plugins/amp/

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
**Reproduce the issue** ✅ 
Identified this issue on localhost.

**Identify the root cause** ✅ 
The root cause is because rocket_buffer is removed totally from AMP compatibility.
https://github.com/wp-media/wp-rocket/blob/57003cff362df710d36acb49085115ea4d09e0eb/inc/ThirdParty/Plugins/Optimization/AMP.php#L103

**Scope a solution** ✅ 
The solution will be to add in AMP.php file CDN rewrite & rewrite_srcset from CDNSubscriber.

**Effort** ✅ 
Effort seems minimal , however I assume that Integration & Unit tests will take longer time. 
[S]

TO DO:
- [x] - Fix and refactor Current Unit tests
- [x] - Fix and refactor Current Integration tests
- [x] - Unit tests for method WP_Rocket\ThirdParty\Plugins\Optimization\AMP::rewrite_cdn() 
- [x] - Integration tests for method WP_Rocket\ThirdParty\Plugins\Optimization\AMP::rewrite_cdn()


inc/ThirdParty/Plugins/Optimization/AMP.php
```diff
@@ -115,4 +117,20 @@ public function disable_options_on_amp() {
...
+	/**
+	 * Rewrites URLs to the CDN URLs if allowed.
+	 * Rewrites URLs in srcset attributes if allowed.
+	 *
+	 * @since 3.5.5
+	 *
+	 * @param  string $html HTML content.
+	 * @return string       HTML content.
+	 */
+	public static function rewrite_cdn( $html ) {
```
hellofromtonya, line 130: Why `static`?

inc/ThirdParty/Plugins/Optimization/AMP.php
```diff
@@ -115,4 +117,20 @@ public function disable_options_on_amp() {
...
+		$container      = apply_filters( 'rocket_container', '' );
+		$cdn_subscriber = $container->get( 'cdn_subscriber' );
```
hellofromtonya, line 132: If the method is not `static`, then you could inject the CDN Subscriber. In doing so, you wouldn’t need to get the container and then a new instance of the `’cdn_subscriber’`. It would a little more performant and maintainable.

inc/ThirdParty/Plugins/Optimization/AMP.php
crystinutzaa: I used first the Static class method and then I converted it to instance method but I forgot to remove the static.

inc/ThirdParty/Plugins/Optimization/AMP.php
crystinutzaa: How to inject the CDN Subscriber?

crystinutzaa:
@Tabrisrp This was my first solution. To remove only those hooks which are not needed and keep only the `rewrite` and `rewrite_srcset`, however there are 2 issues:
1. The array which keeps the hooks (callbacks) is large and needs to be parsed each time the cache is created.
2. The  callback key starts with a random string: `0000000076cde1c40000000038116c32rewrite_srcset` 

Because of this the array needs to be parsed each time, grab the keys and check if it contains our desired key & unset those which are not needed.

The current solution unsets everything and then only calls those 2 functions which are needed.

@Tabrisrp & @hellofromtonya 
Do you see any other solution for this one? Any idea which solution is faster and how should we measure each solution?

inc/ThirdParty/Plugins/Optimization/AMP.php
hellofromtonya: @crystinutzaa You'll want to inject it into the constructor from the ServiceProvider that handles it.

Tabrisrp:
In fact it's not a random string, it's a hash of the object instance used to register the method callback.

So for example here, we can use the following to get the corresponding hash:
```
$container = apply_filters( 'rocket_container', '' );
$cdn_subscriber = $container->get( 'cdn_subscriber' );
$hash = spl_object_hash( $cdn_subscriber );
```

Then you build an array of callbacks to preserve:
```
$preserve = [
    "{$hash}rewrite" => 1,
    "{hash}rewrite_srcset" => 1,
];
```

And you can use `array_intersect_key` to only preserve those:
```
$wp_filter['rocket_buffer'] = array_intersect_key( $wp_filter['rocket_buffer'], $preserve
```

That way, if we ever need to preserve more filters for AMP, we can just change the keys of the `preserve` array, instead of having to create new duplicating methods each time.

crystinutzaa:
Thank you both for the suggestions. I found out that I need to add a filter to exclude all CSS and JS files from CDN on AMP and this was not working 100% (Ops! 😁 ).

I injected the subscriber in constructor.

@Tabrisrp related to your idea, I tried it, however is hard to work with it. The format of WP Hook is like this:
`$wp_filter['rocket_buffer']->callbacks[PRIORITY] = [ARRAY_OF_CALLBACK_FUNCTIONS];`

For this specific case, I need to go to callbacks and for each priority, filter and remove which is not needed.So I can keep was is needed. Even so, with the correct callbacks, it did not worked properly.
Probably i missed something, however it is harder to manage.

Are there any real benefits into modifying the logic in this case?

inc/ThirdParty/Plugins/Optimization/AMP.php
```diff
@@ -24,18 +31,18 @@ class AMP implements Subscriber_Interface {
 	/**
 	 * Constructor
 	 *
-	 * @param Options_Data $options WP Rocket Options instance.
+	 * @param Options_Data $options        WP Rocket Options instance.
+	 * @param Subscriber   $cdn_subscriber WP Rocket CDN Subscriber.
 	 */
-	public function __construct( Options_Data $options ) {
-		$this->options = $options;
+	public function __construct( Options_Data $options, $cdn_subscriber ) {
```
hellofromtonya, line 37: The subscriber implements an interface. We can type hint to the interface instead of the concrete like this:


```php

// code left out for brevity.

/**
 * Instance of CDN Subscriber.
 *
 * @var Subscriber_Interface
 */
private $cdn_subscriber;

/**
 * Constructor
 *
 * @param Options_Data $options WP Rocket Options instance.
 * @param Options_Data $options        WP Rocket Options instance.
 * @param Subscriber_Interface   $cdn_subscriber Instance of CDN Subscriber.
 */
public function __construct( Options_Data $options, Subscriber_Interface $cdn_subscriber ) {
```

Why? It allows us to swap out the concrete for testing or some future use case.

Tabrisrp:
I don't know, it still feels to me this is unoptimal and kinda "hacky", but if it's too complicated to work with the `$wp_filter` global, what about this:

```
unset( $wp_filter['rocket_buffer'] );
add_filter( 'rocket_buffer', [ $cdn_subscriber, 'rewrite' ] );
add_filter( 'rocket_buffer', [ $cdn_subscriber, 'rewrite_srcset ] );
```

The `AMP::rewrite_cdn()` method is not needed since you inject the subscriber, the callbacks are available already.

crystinutzaa:
@Tabrisrp would you like to give it a try with wp_filter? Probably I am doing something wrong.
Indeed it feels a little hacky right now.

hellofromtonya:

PR #2629, opened 2020-05-11
Applies the new clean strategy to `rocket_clean_domain()`.

### LINKED ISSUE #2618: "max execution time" errors in 3.5.4 version of rocket_clean_files()

hellofromtonya:
Applies the new clean strategy from PR #2623 to `rocket_clean_domain()`. 

Part of issue #2618 

The new strategy does the following:

- Only runs the wildcard/regex crawl once for each given URL domain host. 
- It stores these directories, which will be in root of `cache/wp-content/`, in memory.
     - Note: Later we will optimize this further to store in db or via mapping module.
- It then runs each URL with each of these root level cache directories without crawling the filesystem. 
     - If the URL's file/directory does not exist in the cache, it skips and goes to the next one.
     - If it's a directory, it runs `rocket_rrmdir()`.
     - Else, it's a file. It deletes it without running `rocket_rrmdir()`.

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1398,52 +1398,6 @@ function _rocket_get_cache_path_iterator( $cache_path ) { // phpcs:ignore WordPr
 }
 
 /**
- * Gets the entries from the URL using RegexIterator.
- *
- * @since  3.5.4
- * @access private
- *
- * @param Iterator     $iterator   Instance of the iterator.
- * @param string|array $url        URL or parsed URL to convert into filesystem path regex to get entries.
- * @param string       $cache_path Optional. Filesystem path to rocket's cache.
- *
- * @return array|RegexIterator when successful, returns iterator; else an empty array.
- */
-function _rocket_get_entries_regex( Iterator $iterator, $url, $cache_path = '' ) { // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedFunctionFound
-	if ( empty( $cache_path ) ) {
-		$cache_path = str_replace( '/', '\/', rocket_get_constant( 'WP_ROCKET_CACHE_PATH' ) );
-	}
-
-	$parsed_url = is_array( $url ) ? $url : get_rocket_parse_url( $url );
-	$host       = $cache_path . rtrim( $parsed_url['host'], '*' );
-
-	if ( '' !== $parsed_url['path'] && '/' !== $parsed_url['path'] ) {
-		$path = trim( $parsed_url['path'], '/' );
-
-		// Count the hierarchy to determine the depth.
-		$depth = substr_count( $path, '/' ) + 1;
-
-		// Prepare the paths' separator for regex.
-		if ( $depth > 1 ) {
-			$path = str_replace( '/', '\/', $path );
-		}
-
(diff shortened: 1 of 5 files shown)


remyperona (approved):

inc/functions/files.php
```diff
@@ -1132,7 +1146,7 @@ function rocket_rrmdir( $dir, array $dirs_to_preserve = [], $filesystem = null )
 		if ( ! $filesystem->is_dir( $entry ) ) {
 			$filesystem->delete( $entry );
 		} else {
-			rocket_rrmdir( $entry, $dirs_to_preserve );
+			rocket_rrmdir( $entry, $dirs_to_preserve, $filesystem );
```
hellofromtonya, line 1149: Micro-optimization: on recursive calls to `rocket_rrmdir()`, `rocket_direct_filesystem()` will not have to be called over and over again.

inc/functions/files.php
```diff
@@ -838,17 +841,18 @@ function rocket_clean_domain( $lang = '' ) {
 	 */
 	$urls = (array) apply_filters( 'rocket_clean_domain_urls', $urls, $lang );
 	$urls = array_filter( $urls );
```
engahmeds3ed, line 843: I think this is a small and maybe not effective change but wanted to highlight it.
I think looping on items to assure their values are better/faster than array_filter, I tried sample code to check and found that.
You can take this as reference: http://www.levijackson.net/are-array_-functions-faster-than-loops/

crystinutzaa (approved):

engahmeds3ed (approved):

hellofromtonya:

@piotrbak It applies the same strategy we did for `rocket_clean_files()`. 

Note: It does not fix the IIS/Windows directory separator problem; rather, that will be a separate PR.

Let me know if you have any questions.

inc/functions/files.php
hellofromtonya: @engahmeds3ed I remember this article. I had issues with how the benchmarking was done, though I didn't validate his findings through proper benchmarking.

What do I mean? Why did I take issues with it?

Benchmarking requires a large statistical sample of running the code over and over again and then taking the averages. Typically, we run these samples in 10k, 100k, and/or 1m cycles. 

Why? Each time we run it, the results vary. When dealing with fractions of a second, we need a large sample set to analyze and average.

But you raise a valuable point. We need to analyze and benchmark to optimize. Let's do that as part of our Filesystem module work. Sound good?

engahmeds3ed: Sounds great.

piotrbak (approved):
Worked like expected, also on IIS.

PR #2641, opened 2020-05-12
Closes #2619 Varnish cache not cleaned at automatic cache purge

### LINKED ISSUE #2619: Varnish cache is not deleted after expiration of Rocket's cache
**Describe the bug**
We are not clearing Varnish cache when purging expired cache:
https://github.com/wp-media/wp-rocket/blob/e40b824350a165cf52f2285a681d5b6796ed5555/inc/Addon/Varnish/Subscriber.php#L44-L46

**To Reproduce**
1. Enable Varnish Add-on
2. Set the Cache Lifespan to 10 minutes
3. Add error_log here:
https://github.com/wp-media/wp-rocket/blob/e40b824350a165cf52f2285a681d5b6796ed5555/inc/Addon/Varnish/Varnish.php#L116
http://snippi.com/s/q9danzj
4. Preload the cache
5. Update one post, just to be sure that we send the request to Varnish correctly
6. Wait a couple of minutes until the cache expiration event clears the cache and confirm that it hasn't sent the request to Varnish

**Expected behavior**
We should clear the Varnish cache just as ours, it can potentially cause broken layout.

**Additional context**
What about our compatibility with NGINX Helper + 3rd parties like hostings?

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
**Reproduce the issue** ✅ 
Identified this issue on localhost.

**Identify the root cause** ✅ 
The root cause is that Varnish Addons needs to be hooked in Expired Cache Purge event `rocket_after_automatic_cache_purge`.

**Scope a solution** ✅ 
The solution is to hook Varnish Addon into: `rocket_after_automatic_cache_purge` and clean Varnish cache for the deleted urls:
https://github.com/wp-media/wp-rocket/blob/d6c62d85364f422f4c9f575f5ee661a350856fa7/inc/classes/Cache/class-expired-cache-purge.php#L249

**Effort** ✅ 
Effort is minimal [S]

**TO DO**
- [x] - Unit tests
- [x] - Integration tests

### DIFF
--- a/inc/Addon/Varnish/Subscriber.php
+++ b/inc/Addon/Varnish/Subscriber.php
@@ -41,9 +41,10 @@ class Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'before_rocket_clean_domain' => [ 'clean_domain', 10, 3 ],
-			'before_rocket_clean_file'   => [ 'clean_file' ],
-			'before_rocket_clean_home'   => [ 'clean_home', 10, 2 ],
+			'before_rocket_clean_domain'         => [ 'clean_domain', 10, 3 ],
+			'before_rocket_clean_file'           => [ 'clean_file' ],
+			'before_rocket_clean_home'           => [ 'clean_home', 10, 2 ],
+			'rocket_after_automatic_cache_purge' => 'clean_urls',
 		];
 	}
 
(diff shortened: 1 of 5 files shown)


remyperona (approved):

inc/Addon/Varnish/Subscriber.php
```diff
@@ -116,4 +117,36 @@ public function clean_home( $root, $lang ) {
...
+		if ( ! $this->should_purge() ) {
+			return;
+		}
+
+		foreach ( $urls as $data ) {
+			foreach ( $data['files'] as $file_path ) {
+				if ( strpos( $file_path, '#' ) ) {
+					// URL with query string.
+					continue;
+				} else {
```
hellofromtonya, line 136: This `else` is not needed. Why? If the `if` conditional is true, the first `foreach` skips the rest of the code. Therefore, the `else` never fires.

inc/Addon/Varnish/Subscriber.php
```diff
@@ -116,4 +117,36 @@ public function clean_home( $root, $lang ) {
...
+					$data['home_path'] = untrailingslashit( $data['home_path'] );
+					$data['home_url']  = untrailingslashit( $data['home_url'] );
+					if ( '/' === substr( get_option( 'permalink_structure' ), -1 ) ) {
+						$file_path         .= '/';
+						$data['home_path'] .= '/';
+						$data['home_url']  .= '/';
+					}
+				}
+
+				$url = str_replace( $data['home_path'], $data['home_url'], $file_path );
```
hellofromtonya, line 147: Do we have a function that handles this path to url conversion? If yes, can we reuse it here?

hellofromtonya (changes requested):

inc/Addon/Varnish/Subscriber.php
Tabrisrp: we have one for URL -> path, but not for path -> URL yet

hellofromtonya: Okay. We will have both of these for reuse in the new Filesystem module. I'll add this code to my running list where we can simplify once that module is ready. Thanks.

Tabrisrp:
Re-reading the issue, did we confirm that this was causing problems for our users? 🤔

jazir555:
@Tabrisrp 

I can confirm I am having this issue actively right now. I have to manually purge varnish constantly. To be able to clear all the caches including varnish by clearing the wprocket cache would be great. Manually having to purge it is so tedious.

PR #2632, opened 2020-05-11
Fixes 2633: RegexIterator regex to use Windows dir separator when IIS/Windows

hellofromtonya:
This PR fixes #2633 by:

1. Checking if it's a Windows-based machine, i.e. via the constant `DIRECTORY_SEPARATOR`.
2. If yes:
    - Changes `/` into `\` for the `FilesystemIterator`
    - Prepares the directory separator for regex.

Note: Skips this switching code when using the virtual filesystem. 

## TODO 

- [x] Add unit tests.
- [x] Test in Parallels (VM for Windows on Mac). ✅ Works
- [x] Test on Windows laptop/desktop. ✅ Works
- [x] Test on customer's IIS servers, i.e. customers who have reported this problem. @vmanthos @piotrbak 
- [x] QA

PR #2643, opened 2020-05-12
Normalizes filesystem paths for IIS

### LINKED ISSUE #2642: Language directories not preserved when running rocket_clean_domain on IIS
**Describe the bug**
In a specific environment, language directories are not preserved when cleaning the domain, on both 3.5.4 version and after applying the new cleaning strategy:
https://github.com/wp-media/wp-rocket/pull/2629

It happens when the main language is set to be on a domain:
- domain.com
And other languages are set to subdirectories:
-domain.com/en/

I believe that it's because of the array of the paths that should be preserved, there are mixed slashes and backslashes:
`2020-05-12 18:05:03] URLs to preserve: Array
(
    [0] => D:\inetpub\vhosts\domain.com\httpdocs/wp-content/cache/wp-rocket/domain.com(.*)/en
    [1] => D:\inetpub\vhosts\domain.com\httpdocs/wp-content/cache/wp-rocket/domain.com(.*)/fr
    [2] => D:\inetpub\vhosts\domain.com\httpdocs/wp-content/cache/wp-rocket/domain.com(.*)/de
    [3] => D:\inetpub\vhosts\domain.com\httpdocs/wp-content/cache/wp-rocket/domain.com(.*)/ru
    [4] => D:\inetpub\vhosts\domain.com\httpdocs/wp-content/cache/wp-rocket/domain.com(.*)/es
)`

**To Reproduce**
1. On IIS environment
2. WPML with subdirectories enabled, the main language is set to be on the domain (no subdirectory)
3. Try to clean the main language only
4. Confirm that everything is purged

**Expected behavior**
Clear only the selected language

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

hellofromtonya:
This PR adds normalizing for a Windows/IIS filesystem (fs) to fix various problems in the filesystem that impacts i18n and Preload.

Closes #2642 

Changes include:

- Adds a private function `__rocket_is_windows_fs()` to determine if this filesystem is for Windows/IIS. Note: This is turned off when running our virtual filesystem in our automated test suites.
- Adds a private function `_rocket_normalize_path()` to encapsulate the normalizing process:
    - If Windows/IIS fs, converts all `/` directory separators into:
        - when character escaping => `\\\\` (for regex)
        - Else => `\\`
    - For non-Windows/IIS fs:
        - when character escaping => `\/` (for regex)
        - Else => returns original given path
- Adds a private function `_rocket_get_wp_rocket_cache_path()` to get and normalize the `WP_ROCKET_CACHE_PATH` constant. This function can be reused throughout Rocket for now, until we create a new Filesystem module.

## TODO

- [x] Add normalizing function
- [x] Add `WP_ROCKET_CACHE_PATH` normalizing getter
- [x] Update `_rocket_get_cache_dirs()`
- [x] Update `rocket_clean_files()`
- [x] Update `rocket_clean_domain()`
- [x] Update `get_rocket_i18n_to_preserve()` - which fixes #2642 
- [x] Add unit tests for the new private functions
- [ ] QA

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1426,7 +1426,6 @@ function _rocket_get_cache_dirs( $url_host, $cache_path = '', $hard_reset = fals
 		$cache_path = rocket_get_constant( 'WP_ROCKET_CACHE_PATH' );
 	}
 
-	// When Windows-based.
 	$is_windows = (
 		DIRECTORY_SEPARATOR === '\\'
 		&&
@@ -1469,4 +1468,66 @@ function _rocket_get_cache_dirs( $url_host, $cache_path = '', $hard_reset = fals
 	}
 
 	return $domain_dirs[ $url_host ];
+}
+
+/**
+ * Normalizes the given filesystem path:
+ *  - Windows/IIS-based servers: converts all directory separators to "\\" or, when escaping, to "\\\\".
+ *  - Linux-based servers: if $forced is true, uses wp_normalize_path(); else, returns the original path.
+ *
+ * @since  3.5.5
+ * @access private
+ *
+ * @param string $path   Filesystem path (file or directory) to normalize.
+ * @param bool   $escape Optional. When true, escapes the directory separator(s).
+ * @param bool   $force  Optional. When true, forces normalizing off non-Windows' paths.
+ *
+ * @return string
+ */
+function _rocket_normalize_path( $path, $escape = false, $force = false ) { // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedFunctionFound
+	if ( _rocket_is_windows_fs( $path ) ) {
+		return str_replace(
+			'/',
+			$escape ? '\\\\' : '\\',
+			$path
+		);
+	}
+
+	if ( $escape ) {
+		return str_replace( '/', '\/', $path );
+	}
+
+	if ( ! $force ) {
+		return $path;
+	}
+
+	return wp_normalize_path( $path );
+}
+
+/**
+ * Checks if the filesystem (fs) is for Windows/IIS server.
+ *
+ * @since  3.5.5
+ * @access private
+ *
+ * @param bool $hard_reset Optional. When true, resets the memoization.
+ *
+ * @return bool
+ */
+function _rocket_is_windows_fs( $hard_reset = false ) { // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedFunctionFound
+	static $is_windows = null;
+
+	if ( $hard_reset ) {
+		$is_windows = null;
+	}
+
+	if ( is_null( $is_windows ) ) {
+		$is_windows = (
+			DIRECTORY_SEPARATOR === '\\'
+			&&
+			! rocket_get_constant( 'WP_ROCKET_RUNNING_VFS', false )
+		);
+	}
+
+	return $is_windows;
 }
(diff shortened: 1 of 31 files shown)


hellofromtonya:
@crystinutzaa Please validate that these changes work on your Windows PC by:

- [ ] Edit and save a post - should run `rocket_clean_files()`
- [ ] Go to Settings > Rocket and then click on the "Clean cache" button.

hellofromtonya:
@Tabrisrp and @crystinutzaa This PR came out tests Piotr was running where he discovered IIS wasn't working for `get_rocket_i18n_to_preserve()`. 

This PR moves the IIS fixes from PR #2632 into private functions for reuse. It then uses them in 4 functions:

- `_rocket_get_cache_dirs()`
- `rocket_clean_files()`
- `rocket_clean_domain()`

After this one is merged, I'll then apply the same changes in PR #2251 to fix the expired cache preload for IIS (issue #2584).

inc/functions/files.php
```diff
@@ -622,7 +622,7 @@ function rocket_clean_files( $urls, $filesystem = null ) {
 
 	/** This filter is documented in inc/front/htaccess.php */
 	$url_no_dots = (bool) apply_filters( 'rocket_url_no_dots', false );
-	$cache_path  = rocket_get_constant( 'WP_ROCKET_CACHE_PATH' );
+	$cache_path  = _rocket_get_wp_rocket_cache_path();
```
crystinutzaa, line 625: Why you defined the functions with a leading _ ? Is there a benefit / purpose of using it like this?

crystinutzaa: Just saw the conversation is slack.

hellofromtonya: To capture the conversation here:

These are private functions which are for internal use only. We don’t want 3rd parties using them.

Benefits to using this naming convention and marking them private:

- Recognizable from other programming languages as “private” for internal use. It’s a common function naming convention in other languages. It’s effective because you stop and wonder why before using it.

- We won’t need to deprecate them later when we build the Filesystem module,

inc/functions/i18n.php
```diff
@@ -323,7 +323,8 @@ function get_rocket_i18n_to_preserve( $current_lang ) { // phpcs:ignore WordPres
 	foreach ( $langs as $lang ) {
 		$parse_url           = get_rocket_parse_url( get_rocket_i18n_home_url( $lang ) );
 		$langs_to_preserve[] = _rocket_normalize_path(
-			"{$cache_path}{$parse_url['host']}(.*)/" . trim( $parse_url['path'], '/' )
+			"{$cache_path}{$parse_url['host']}(.*)/" . trim( $parse_url['path'], '/' ),
+			true // escape directory separators for regex.
```
hellofromtonya, line 334: @Tabrisrp As this is a regex pattern, the directory separators need to be escaped. I made this change to ensure they are both normalized and escaped. Do you foresee any problems here? Why? The array it returns is different:

For Linux:

```
\/path\/to\/cache\/wp-rocket\/example.org(.*)\/fr
```

For IIS/Windows:

```
C:\\path\\to\\cache\\wp-rocket\\example.org(.*)\\fr
```

Tabrisrp: As far as I can think of, it should be fine. Do you think using a different delimiter than `/` for the regex could help simplify this?

hellofromtonya: >Do you think using a different delimiter than / for the regex could help simplify this?

I don't think it'll impact it and it's consistent with every where else in the code.

PR #2647, opened 2020-05-13
Closes #2595 Add license data to Beacon prefill

### LINKED ISSUE #2595: Add the license information to the Beacon
For the Beacon to bring the license information to HelpScout and be added into the "License" custom field based on what the customer has.

image

Tabrisrp:
Closes #2595 

Populate the `prefill` property with the customer license custom field value, to be passed to Helpscout and auto-populate it in new tickets sent from Beacon.

As reference the Beacon API doc on prefill: https://developer.helpscout.com/beacon-2/web/javascript-api/#beaconprefill-formobject

- [x] Add method to populate prefill data in Beacon script
- [x] Update view to add prefill data
- [x] Update tests
- [x] QA


Tabrisrp:
@hellofromtonya Can you check this commit https://github.com/wp-media/wp-rocket/pull/2647/commits/7a51a3374f17c203a753a5258b5c2bfbc70b0b11 and confirm that it's how we should do this for unit tests using `Abstract_Render`?

tests/Integration/inc/Engine/Admin/Beacon/Beacon/insertScript.php
```diff
@@ -26,6 +26,7 @@ public function tearDown() {
 		parent::tearDown();
 
 		set_current_screen( 'front' );
+		delete_transient( 'wp_rocket_customer_data' );
```
hellofromtonya, line 29: Could this transient already be set before entering into this test class? 

If yes or maybe, then you want to set it up to save and then restore the original. How? Add this property to class:

```php
protected static $transients = [
    'wp_rocket_customer_data' => null,
];
```

The `TestCase` handles the original save and restore stuff for you.

tests/Unit/inc/Engine/Admin/Beacon/Beacon/insertScript.php
```diff
@@ -28,32 +27,29 @@ public function setUp() {
 		parent::setUp();
 
 		$this->options = Mockery::mock( Options_Data::class );
-		$this->beacon  = new Beacon( $this->options, $this->filesystem->getUrl( 'wp-content/plugins/wp-rocket/views/settings' ) );
+		$this->beacon  = Mockery::mock( Beacon::class . '[insert_script]', [
```
hellofromtonya, line 30: Here you want `insert_script` to run normally. But you want to mock `generate` method. The way Mockery works is you are defining which method will be mocked, i.e. a partial mock. 

```suggestion
		$this->beacon  = Mockery::mock( Beacon::class . '[generate]', [
```

[Here's an example](https://github.com/wp-media/wp-rocket/blob/develop/tests/Unit/inc/Engine/CDN/RocketCDN/NoticesSubscriber/promoteRocketcdnNotice.php#L23-L30)

Why are mocking `generate()` method? It's a separate public method which is used as dependency within the `insert_script` method. With a unit test, we want to test only the target method/function, while mocking each dependency.

tests/Unit/inc/Engine/Admin/Beacon/Beacon/insertScript.php
```diff
@@ -68,9 +64,20 @@ public function testShouldReturnBeaconScript( $locale, $expected ) {
...
 			->andReturn( [] );
 
-		$this->assertSame(
-			$this->format_the_html( $expected ),
-			$this->getActualHtml()
-		);
+		$this->setUpGenerate( 'beacon', $expected['data'] );
+
+		$this->beacon->shouldReceive( 'insert_script' )
+			->once();
```
hellofromtonya, line 70: This mock is not needed. Why? We're directly invoking in the next lines of code. We know it's going to run once because we are invoking it.

views/settings/beacon.php
```diff
@@ -18,6 +19,7 @@
 <script type="text/javascript">window.Beacon('init', '<?php echo esc_js( $data['form_id'] ); ?>')</script>
 <script>window.Beacon("identify", <?php echo $data['identify']; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>);</script>
 <script>window.Beacon("session-data", <?php echo $data['session']; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>);</script>
+<script>window.Beacon("prefill", <?php echo $data['prefill']; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?>);</script>
 <script>window.addEventListener("hashchange", function () {
 	window.Beacon("suggest");
 }, false);</script>
```
hellofromtonya, line 25: General question: Why do we have separate `<script></script>` elements for each of these? Why not create one `<script>` with all of the JS needed???

Tabrisrp: Honestly I don't remember if there was a reason. I removed them and packed everything in one inline script

Tabrisrp:

To test this, you'll need to enable the custom fields in Beacon. To do that, in Helpscout, go to Manage > Beacons and select Beacon Support EN.

Then on Contact > Contact Form > Show custom fields, enable it, and check the License checkbox, save the changes.

Once that's done, You can reload the WP Rocket settings page, and you can submit a new ticket via Beacon

On the ticket, the License custom field should be showing your license level automatically.

arunbasillal:
@GeekPress / @Tabrisrp I reverted the changes I made to Beacon. Not sure when is the right time to enable it. Maybe once this version (3.6) is released, right?

Tabrisrp:
Once it's released yes, thank you for the QA

PR #2648, opened 2020-05-14
Closes #2635 Remove CPCSS styletag via JS

### LINKED ISSUE #2635: Once the page is loaded, remove the CPCSS style tag via JS
We don't remove our CPCSS once the page is loaded. It can cause layout issues if some JS scripts are added or updated CSS classes after applying the CPCSS.

We could remove the CPCSS style tag once the page is fully loaded.

**Additional context**
We already tried to implement this:
https://github.com/wp-media/wp-rocket/pull/1172/files

We decided to not implement because it cause an issue for one customer during the beta test, even if it helped to fix issues for more people. We will like to implement it as it should fix more issues than it should create.

crystinutzaa:
**Reproduce the issue** ✅ 
I can confirm that CPCSS is not removed from my localhost.

**Scope a solution** ✅ 
The solution will be to use 
https://developer.mozilla.org/en-US/docs/Web/API/Document/readyState
Add an event listener  to remove the CPCSS code.
- No need for readyState = complete

And for IE and Opera we should use:
attachEvent on window load. 

**Effort** ✅ 
Effort `[S]`


GeekPress:
@crystinutzaa What do you think of providing a minified version? 

```function a(){document.getElementById("rocket-critical-css").remove()}window.addEventListener?window.addEventListener("load",a):window.attachEvent&&window.attachEvent("onload",a);```

And if the `WP_SCRIPT_DEBUG` is set to true, we provide the unminified version?

crystinutzaa:
@GeekPress Yup. Sure.
Is still work in progress but will be minified at the end.

PR #2639, opened 2020-05-12
Closes #2377 CPCSS for specific pages

remyperona:
Closes #2377 

PR to close the epic.

👉 No code review is needed for this one as it's an epic and filled with PRs of user stories we've already reviewed and approved.

- [x] #2622 Refactor CPCSS code
- [x] QA

PR #2655, opened 2020-05-15
Add WP_ROCKET_VIEWS constant

Tabrisrp:
This is needed in multiple opened PRs:
- #2622
- #2625
- #2647

Add a `WP_ROCKET_VIEWS` constant, which is defined early in the integration bootstrap to use the Vfs structure.

Tests are updated for it to keep them working.


hellofromtonya:
@Tabrisrp Why do we need to mock the views? Shouldn't we be using the real view files and then capturing those in our tests?

Tabrisrp:
@hellofromtonya I agree that's what I would prefer, but what I encountered previously was that when using the `FilesystemTestCase` and the Vfs, the views would try to be loaded from the Vfs instead of the real filesystem during the tests.

So what we would need is to load the views from the real filesystem, but use the Vfs for anything else.

hellofromtonya:
@Tabrisrp Here's an example of using the real views:

Beacon:

- [Here's the fixture](https://github.com/wp-media/wp-rocket/blob/develop/tests/Fixtures/inc/Engine/Admin/Beacon/Beacon/insertScript.php)
- [Here's the integration test](https://github.com/wp-media/wp-rocket/blob/develop/tests/Integration/inc/Engine/Admin/Beacon/Beacon/insertScript.php)

In these, in the `develop` branch, the template path points to the real views and not the vfs.

You might notice the vfs structure in the test data fixture. That's only used for [the unit tests](https://github.com/wp-media/wp-rocket/blob/develop/tests/Unit/inc/Engine/Admin/Beacon/Beacon/insertScript.php#L16) and not the integration tests. Notice the test cases.

>what I encountered previously was that when using the FilesystemTestCase and the Vfs, the views would try to be loaded from the Vfs instead of the real filesystem during the tests.

Let's look at specific examples to figure out if we really need to do this PR.

Tabrisrp:
Yes I'm not sure this PR is needed anymore, we can update #2625, #2639 and #2647 with the latest develop changes to see how it goes

hellofromtonya:
@Tabrisrp I think that’s a good plan. Let’s pause this PR. Mark it as a draft. 

If there are issues after updating, ping me and we can work through each together.

Tabrisrp:
Closing this PR as it's not needed in the end.

Issue #2227, opened 2019-12-30
rocket_init() server callback taking 0.0672s


gevcen:
**Before submitting an issue please check that you’ve completed the following steps:**
Everything up-to-date.

**Describe the bug**
After investigation with my hosting provider to analyse the slow server response time on uncached pages of my site, we've found out that one rocket_init() callback in wp-rocket/inc/main.php:26 is taking 0.0672s, which is obviously a too high value.

**To Reproduce**
Execute a simulation of the loading times using the WordPress built-in command-line interface.

**Expected behavior**
As per the instruction of my hosting provider, any value above 0.00XX is too high and must be addressed.

**Screenshots**
![image](https://user-images.githubusercontent.com/51261962/71604673-ead7f680-2b63-11ea-80b4-9aa485b7430e.png)

**Additional context**
N/A

Tabrisrp:
@hellofromtonya I assigned this to you, to have a look at, as it's most likely the same thing as we had on Imagify. I don't think the issue is in `rocket_init` itself, but in the `load()` method of `WP_Rocket\Plugin`, where there is a couple of `array_merge` uses.

hellofromtonya:
@Tabrisrp This one is part of our Performance initiatives, i.e. to measure and optimize the time (PHP runtime/execution) and space (memory usage) of the plugin.

piotrbak:
Going to close it. The `array_merge` were reduced in the past. We haven't had any feedbacks about that problem since then.

PR #2656, opened 2020-05-15
Upgrades the Plugin by moving into the architecture & optimizing

hellofromtonya:
- Puts the `Plugin` into the root of `inc` as it is the starting point for all of the modules.
- Moves the `Plugin` class into the new architecture
- Renames the file from `class-plugin.php` to `Plugin.php`.
- Refactors to break up the `load` method by moving each part of the subscriber inits into separate methods. Easier to read. Encapsulates the pieces via grouping.
- Optimizes:
     - Instead of calling `get_rocket_option()`, use the `Option` instance directly. Why? The `Plugin` object is the master router of the plugin. It has access and control of the options api and data. An instance of `Options` already exists in `Plugin`. `get_rocket_option()` creates a new instance of it. That's unnecessary.
     - Stores instances on properties for reuse and potentially getters (future).
     - Runs `rocket_valid_key()` once instead of twice by storing into a property.


inc/Plugin.php
```diff
@@ -0,0 +1,276 @@
...
+
+namespace WP_Rocket;
+
+use Imagify_Partner;
+use League\Container\Container;
+use WP_Rocket\Event_Management\Event_Manager;
+use WP_Rocket\Admin\Options;
+use WP_Rocket\Admin\Options_Data;
+
+defined( 'ABSPATH' ) || exit;
```
Tabrisrp, line 11:
```suggestion
```

Tabrisrp:
You might want to also check #2227 while doing this change for the performance optimization part

PR #2657, opened 2020-05-15
Upgrades 3.6 test suite

hellofromtonya:
Upgrades tests from `branch-3.6`:

- Ensures tests restore the original state before leaving the test class, including: 
     - Original settings
     - Admin capability
     - Transients
     - Hooks
- Adds checks to ensure `gzencode()` function exists
- Remove the `whenRocketGetConstant()`. This method was renamed and added to the `FilesystemTestCase` as part of the `setUp()` for all filesystem tests
- Tests use `StubTrait` for constants
- Refactors
     - Adds test cases to reduce redundant code
     - Adds test data


tests/Integration/inc/Engine/Optimization/TestCase.php
```diff
@@ -1,14 +1,44 @@
...
+	}
+
+	public static function tearDownAfterClass() {
+		parent::setUpBeforeClass();
+
+		update_option( 'wp_rocket_settings', self::$original_settings );
+	}
+
+	public function setUp() {
+		$this->wp_content_dir = 'vfs://public/wp-content';
```
Tabrisrp, line 27: is this still necessary when using the FilesystemTestCase?

hellofromtonya: Nope. It was removed.

PR #2666, opened 2020-05-18
Closes #2665 Expired Cache preload doesn't fetch correct URLs

### LINKED ISSUE #2665: Expired cache preload doesn't fetch the correct URLs
**Describe the bug**
In 3.5.5 version we fixed the problem with expired cache preload for WPML. In this fix, we introduced another problem. When `$data['home_url']` doesn't contain `/`:
https://github.com/wp-media/wp-rocket/blob/c7a020448d6639ca3cb4cddf72195f9d9c18fbec/inc/Engine/Preload/PartialPreloadSubscriber.php#L151

We're not preloading the expired cache.

**To Reproduce**
1. Go to the WordPress site without WPML and other plugins manipulating home_url
2. Set Cache Lifespan to 10 minutes
3. Preload cache manually
4. Confirm that the expired cache is not preloaded

**Expected behavior**
We need to preload expired cache correctly

**Additional context**
The solution would be to make an additional check and append `/` if it's missing:
http://snippi.com/s/wpld6o2

We need to be sure that `$data['home_url']` doesn't contain `/` before appending it because we might introduce another bug.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
**Describe the bug**
In 3.5.5 version we fixed the problem with expired cache preload for WPML. In this fix, we introduced another problem. When `$data['home_url']` doesn't contain `/`:
https://github.com/wp-media/wp-rocket/blob/c7a020448d6639ca3cb4cddf72195f9d9c18fbec/inc/Engine/Preload/PartialPreloadSubscriber.php#L151

We're not preloading the expired cache.

**To Reproduce**
1. Go to the WordPress site without WPML and other plugins manipulating home_url
2. Set Cache Lifespan to 10 minutes
3. Preload cache manually
4. Confirm that the expired cache is not preloaded

**Expected behavior**
We need to preload expired cache correctly

**Additional context**
The solution would be to make an additional check and append `/` if it's missing:
http://snippi.com/s/wpld6o2

We need to be sure that `$data['home_url']` doesn't contain `/` before appending it because we might introduce another bug.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

### DIFF
--- a/inc/Engine/Preload/PartialPreloadSubscriber.php
+++ b/inc/Engine/Preload/PartialPreloadSubscriber.php
@@ -142,9 +142,11 @@ class PartialPreloadSubscriber implements Subscriber_Interface {
 				} else {
 					$file_path         = untrailingslashit( $file_path );
 					$data['home_path'] = untrailingslashit( $data['home_path'] );
+					$data['home_url']  = untrailingslashit( $data['home_url'] );
 					if ( '/' === substr( get_option( 'permalink_structure' ), -1 ) ) {
 						$file_path         .= '/';
 						$data['home_path'] .= '/';
+						$data['home_url']  .= '/';
 					}
 				}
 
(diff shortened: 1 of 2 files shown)


piotrbak:
Starting the QA

PR #2672, opened 2020-05-20
Closes #2612 CPCSS for mobile: generation (3/4)

### LINKED ISSUE #2612: CPCSS for mobile: generation part (3/4)
Sub-task for epic #2378 

### inc/Engine/CriticalCSSSubscriber.php
- [x] Update `generate_critical_css_on_activation()` to also trigger a generation when "Optimize CSS delivery" is enabled, "specific cache for mobile" is enabled and "async_css_mobile" = `1`
- [x] Update tests

### inc/Engine/CriticalCSS.php
- [x] Update `set_items()` to also add mobile version of each item if the "specific cache for mobile" option is enabled and "async_css_mobile"  = `1`
- [x] Add tests

### inc/Engine/CriticalCSSGeneration.php
- [x] Update `task()`to use the new handshake with the API introduced in #2512. Re-queue the task when it's still in progress instead of using a `while` loop and `sleep`. Also make sure the `mobile` parameter is passed along the POST request when needed
- [x] Add tests

### CPCSS for specific posts
- [x] Generate the mobile version when needed along the normal version
- [x] Update tests

remyperona:
Closes #2612

#### CriticalPath/AdminSubscriber
- [x] Unit tests failing for `enqueue_admin_edit_script()` => Tonya

#### inc/Engine/CriticalCSSSubscriber.php
- [x] Add new method `maybe_generate_cpcss_mobile()` to also trigger a generation when "Optimize CSS delivery" is enabled, "specific cache for mobile" is enabled and "async_css_mobile" = `1`
- [x] Update `init_critical_css_generation()` to change the CPCSS to generate depending on the options
- [x] Update `generate_critical_css_on_activation()` to change the CPCSS to generate depending on the options
- [x] Add unit & integration tests for `maybe_generate_cpcss_mobile()` => Ahmed
- [x] Add unit & integration tests for `init_critical_css_generation()` => Remy
- [x] Update unit & integration tests for `generate_critical_css_on_activation()` => Remy

#### inc/Engine/CriticalCSS.php
- [x] Update `set_items()` to also add mobile version of each item if the "specific cache for mobile" option is enabled and "async_css_mobile"  = `1`
- [x] Update `process_handler()` to set different items list depending on the type of CPCSS to generate
- [x] Add unit & integration tests for `process_handler()` => Ahmed
- [x] `clean_critical_css()` use an argument like we did for set_items(), to be able to do a partial clear only

#### inc/Engine/CriticalCSSGeneration.php
- [x] Update `task()`to use the new handshake with the API introduced in #2512. Re-queue the task when it's still in progress instead of using a `while` loop and `sleep`. Also make sure the `mobile` parameter is passed along the POST request when needed
- [x] Add unit & integration tests for `task()` => Remy

#### CPCSS for specific posts
- [x] Generate the mobile version when needed along the normal version
- [x] Delete the mobile version when needed along the normal version
- [x] Update tests => Ahmed and Cristina
- [x] Delete CPCSS for a post when the post is deleted => Cristina
- [x] See if we can display the notice on post differently depending on the result (if that’s not yet the case), by using the success/warning/error styles
Cristina - this was working just fine in Classic editor, but not in Block editor. Needed to add & remove some div class to make it work properly in Block Editor.
- [x] Don't generate CPCSS mobile for posts if the option "separate cache for mobile devices" is disabled => Ahmed

#### DataManager
- [x]  get_cache_job_id() - adds $is_mobile
- [x]  set_cache_job_id() - adds $is_mobile
- [x]  delete_cache_job_id() - adds $is_mobile
- [x] delete_cpcss() - adds $is_mobile 

#### RESTWPPOST
- [x] delete tests are failing => Tonya


views/metabox/cpcss/container.php
```diff
@@ -28,3 +28,12 @@
...
 </div>
+<?php if ( $data['async_css_mobile'] ) : ?>
+	<div id="cpcss_response_notice_mobile" class="components-notice is-notice is-warning hidden">
+		<div class="components-notice__content">
+		</div>
+	</div>
+<?php endif; ?>
+<script type="text/javascript">
+	let wprMobileCpcssEnabled = <?php echo esc_js( $data['async_css_mobile'] ); ?>;
+</script>
```
Tabrisrp, line 39: This should be added in the `wp_localize_script()` in `AdminSubscriber::admin_enqueue_script()` instead

crystinutzaa: Thank you! I was not aware about this till now. Modified 💯

inc/Engine/CriticalPath/RESTWP.php
```diff
@@ -109,10 +136,13 @@ public function generate( WP_REST_Request $request ) {
 
 		// get item url.
 		$item_url  = $this->get_url( $item_id );
-		$item_path = $this->get_path( $item_id );
 		$timeout   = ( isset( $request['timeout'] ) && ! empty( $request['timeout'] ) );
+		$item_path = $this->get_path( $item_id );
+		if ( $is_mobile ) {
+			$item_path = $this->get_path( $item_id, true );
+		}
```
Tabrisrp, line 143:
```suggestion
		$item_path = $this->get_path( $item_id, $is_mobile );
```

inc/Engine/CriticalPath/CriticalCSS.php
```diff
@@ -269,51 +277,75 @@ public function get_public_taxonomies() {
...
 		 * Filters the array containing the items to send to the critical CSS generator.
 		 *
 		 * @since  2.11.4
 		 *
-		 * @param array $this ->items Array containing the type/url pair for each item to send.
+		 * @param array $items Array containing the type/url pair for each item to send.
 		 */
 		$this->items = (array) apply_filters( 'rocket_cpcss_items', $this->items );
+
+		ksort( $this->items );
```
hellofromtonya, line 348: @Tabrisrp `ksort()` is slow. Why do we need this?

Tabrisrp: The goal is to re-order the array to group the 2 versions (default & mobile) for the same page together, instead of doing all the default ones first, then all the mobile ones after

hellofromtonya: Then shouldn't we only do it if we're also doing mobile, i.e. version === 'all'?

inc/Engine/CriticalPath/CriticalCSSGeneration.php
```diff
@@ -49,141 +54,36 @@ class CriticalCSSGeneration extends WP_Background_Process {
...
-						'url' => $item['url'],
-					],
-					$item
-				),
-			]
-		);
-
+		$timeout   = false;
 		$transient = get_transient( 'rocket_critical_css_generation_process_running' );
+		$mobile    = isset( $item['mobile'] ) ? true : false;
```
hellofromtonya, line 59: `isset` returns `true and `false. We don't need to explicitly return `true` or `false`.

inc/Engine/CriticalPath/DataManager.php
```diff
@@ -153,13 +178,14 @@ public function delete_cache_job_id( $item_url ) {
...
+	 * @param bool   $is_mobile Bool identifier for is_mobile CPCSS generation.
 	 *
 	 * @return string
 	 */
-	private function get_cache_key_from_url( $item_url ) {
+	private function get_cache_key_from_url( $item_url, $is_mobile = false ) {
 		$encoded_url = md5( $item_url );
 
-		return 'rocket_specific_cpcss_job_' . $encoded_url;
+		return 'rocket_specific_cpcss_job_' . $encoded_url . ( $is_mobile ? '_mobile' : '' );
```
hellofromtonya, line 189: Optimization opportunity:

It's faster to not concatenate an empty string. 

```php
private function get_cache_key_from_url( $item_url, $is_mobile = false ) {
	$encoded_url = md5( $item_url );

	if ( $is_mobile ) {
		$encoded_url .= '_mobile';
	}

	return 'rocket_specific_cpcss_job_' . $encoded_url;
}
```

inc/Engine/CriticalPath/ProcessorService.php
```diff
@@ -156,14 +163,20 @@ private function check_cpcss_job_status( $job_id, $item_path, $item_url ) {
...
-		$error = sprintf( __( 'Critical CSS for %1$s not generated.', 'rocket' ), $item_url );
+	private function on_job_error( $job_details, $item_url, $is_mobile = false ) {
+		$this->data_manager->delete_cache_job_id( $item_url, $is_mobile );
+
+		if ( $is_mobile ) {
+			// translators: %1$s = page URL.
+			$error = sprintf( __( 'Mobile Critical CSS for %1$s not generated.', 'rocket' ), $item_url );
+		} else {
+			// translators: %1$s = page URL.
+			$error .= sprintf( __( 'Critical CSS for %1$s not generated.', 'rocket' ), $item_url );
```
hellofromtonya, line 177: Shouldn't this assign instead of concatenate and assign?

inc/Engine/CriticalPath/CriticalCSS.php
Tabrisrp: you're right it's not needed for the other cases. Commit: https://github.com/wp-media/wp-rocket/commit/d04a9865962683ff3a85a390e457fba1a066775c

hellofromtonya (changes requested):

hellofromtonya (approved):

Tabrisrp:

Testing (considering that `async_css_mobile` is enabled):
- Enable "_optimize CSS delivery_" + "_separate cache for mobile_": default & mobile version of CPCSS should be generated
- "_Optimize CSS Delivery_" already enabled, enable "_separate cache for mobile_": mobile version of CPCSS should be generated
- Click on "_regenerate CPCSS_": appropriate version should be generated depending on the selected options (either only default, or default + mobile)
- On a post: click on generate, 2 files should be generated (default + mobile) if "_separate cache for mobile_" is enabled
- On a post: deleting the specific CPCSS should delete the 2 files

crystinutzaa (approved):

piotrbak (changes requested):
@Tabrisrp @hellofromtonya 

**The 2nd scenario failed:**
1. "Optimize CSS Delivery" already enabled
2. Enabling "Separate Cache for Mobile Devices"
3. Mobile Critical CSS is being generated
4. **Desktop Critical CSS is wiped out**

**Also, the 4th one is not working as expected:**
1. "Optimize CSS Delivery" enabled
2. "Separate Cache for Mobile Devices" disabled
3. Generating Critical CSS for post
4. Two versions of Critical CSS are generated

**Additional thoughts:**
1. Thrashing and removing post doesn't remove Critical CSS
2. We're showing the `yellow` popup for a success message. In most cases we use yellow when it's a warning:
image

3. The following scenario doesn't generate mobile Critical CSS:
- Enable "Optimize CSS Delivery" without "Separate Cache for Mobile Devices" enabled
- Files are generated, disable "Optimize CSS Delivery"
- Files are not removed
- Enable "Separate Cache for Mobile Devices"
- Enable "Optimize CSS Delivery"

4. Not sure if this can damage anything but:
- "Optimize CSS Delivery" enabled
- Open post editor
- Disable "Optimize CSS Delivery"
- We're still able to generate Critical CSS for a post correctly.

crystinutzaa:
@piotrbak where did you tested the color notice? Where I can see it. For this one:
` See if we can display the notice on post differently depending on the result (if that’s not yet the case), by using the success/warning/error styles` 
I am using these 2 classes for that notice: `notice-error` or `notice-success` depending on the reply from the REST API.

tests/Unit/inc/Engine/CriticalPath/CriticalCSSSubscriber/maybeGenerateCpcssMobile.php
```diff
@@ -33,7 +35,8 @@ public function setUp() {
 			$options,
 			null,
 		] );
-		$this->subscriber   = new CriticalCSSSubscriber( $this->critical_css, $options, null );
+		$this->processor_service = Mockery::mock( ProcessorService::class );
```
hellofromtonya, line 38: @crystinutzaa Are we mocking methods on the `ProcessorService`? If no, then we don't need to have this as a property on the test class.

inc/Engine/CriticalPath/RESTWP.php
```diff
@@ -100,7 +110,39 @@ public function check_permissions() {
...
+					new WP_Error(
+						'caching_mobile_separate_enabled',
+						__( 'Separate cache files for mobile devices option is enabled.', 'rocket' ),
+						[
+							'status' => 400,
+						]
+					)
+				)
+			);
+		}
```
Tabrisrp, line 144: I don't understand this part, is there something missing?

hellofromtonya (approved):

inc/Engine/CriticalPath/RESTWP.php
hellofromtonya: @Tabrisrp It's part of this commit https://github.com/wp-media/wp-rocket/commit/3a7adbfc5607f65455d91d981465a9fcfc94dbd6. What should this be?

Tabrisrp: it should be the opposite, we should prevent generation if the option is disabled

Tabrisrp: https://github.com/wp-media/wp-rocket/pull/2672/commits/726e28ac72a6d497cfec4e09e0f98e9059a5e939

Tabrisrp:
@piotrbak This is ready for a new QA.

- Scenario 2 & 4 should be fixed
- Deleting a post should also delete the related CPCSS (still preserved when trashing, in case the post is restored)
- Notices display is updated
- Mobile critical CSS should be correctly generated for your point 3
- No changes for point 4 as it's pretty edge case and doesn't break anything

tests/Unit/inc/Engine/CriticalPath/CriticalCSSSubscriber/maybeGenerateCpcssMobile.php
engahmeds3ed: I think this is resolved here: https://github.com/wp-media/wp-rocket/commit/3180e6571c892ccc5173782186766c1c0ea44b2e

piotrbak:
Starting the QA

piotrbak:
@wp-media/dev 
I have problems with the last scenario (testing on Cloudflare), I can't tell you if it's related to the Cloudflare itself or the changes made after my last tests. When trying to switch the post to the regular CPCSS, I'm getting 403 error:

I'm not able to test it in the same environment as yesterday since something is broken on Mega now.

My second concern is that I'm very rarely able to generate CPCSS for all post types. On my personal website, where I have 24 to generate, I wasn't able to do it even once. After some requests we're receiving:
`timeout. Please retry a little later.`
Maybe we can somehow slow down the process, in order not to make requests that fast? I'm not sure if this is the real culprit though.

@arunbasillal Could you test those things on your own?
On sites with and without Cloudflare:
- Generate CPCSS for a single post
- Try to revert it back to the regular CPCSS
- Is there a 403?

On the same sites, check if you're able to generate CPCSS for all post types without any problems (give it a couple of tries). Thanks!

crystinutzaa:
@piotrbak the 403 on Revert back to regular CPCSS is related to a security plugin (probably Cloudflare) which blocks the DELETE REST request.
As for generation, the request is sent only once & then it simply rechecks the result from the API. 
It can be something in the CPCSS API which causes this.

@Tabrisrp who can check why this happens?

hellofromtonya:
@piotrbak This PR is ready for re-QA. 

What changed? You found a time-out issue. That was being caused by us blasting the API for requests to check if it was done processing or not. We weren't waiting to give it time to get done.

In PR 2686, we added an async system that waits 3 seconds to ask the API if it's done.

Please redo your QA on this PR. Thanks!

arunbasillal:
Doing the QA again. Noticed this: 

cpcss

Some early feedback: 

- The mobile version is currently counted as a post type. This shouldn't be the case because mobile version isn't a post type. 
- The count seems to be wrong as you can see in the screenshot. We had something similar in the past, maybe it's related? - https://github.com/wp-media/wp-rocket/issues/1333
- We had agreed not to display separate notices for mobile, sorry if this was not communicated widely - https://wp-media.slack.com/archives/G8L4EE8N9/p1590096487031800?thread_ts=1590088503.020600&cid=G8L4EE8N9
- ^ separate notice for mobile can be confusing for customers. They will wonder why they are not seeing it (i.e. when separate cache for mobile is not enabled). 

Continuing with testing.

hellofromtonya:
@crystinutzaa See @arunbasillal's notes above about not having a mobile separate notices. Thoughts?

arunbasillal:
- Enable "optimize CSS delivery" + "separate cache for mobile": default & mobile version of CPCSS should be generated 👍 
- ^ Tested if the mobile version is actually used on mobile. 👍 
- ^ When mobile version was not available, desktop CPCSS was used as fallback 👍 

- "Optimize CSS Delivery" already enabled, enable "separate cache for mobile": mobile version of CPCSS should be generated 👍 
- ^Desktop CPCSS is not wiped out 👍 

- Click on "regenerate CPCSS": appropriate version should be generated depending on the selected options (either only default, or default + mobile) 👍 

- On a post: click on generate, 2 files should be generated (default + mobile) if "separate cache for mobile" is enabled 👍 
- ^ Tested if the mobile version is actually used on mobile. 👍 
- ^ When mobile version was not available, desktop CPCSS was used as fallback 👍 
- ^ the generated notice is green:
image
- ^ Tested with classic editor as well 👍 
- ^ When "Separate cache for mobile is disabled", only desktop version of specific CPCSS is generated. 👍 

- On a post: deleting the specific CPCSS should delete the 2 files 👍 
- ^ The notice is green
image
- ^ Tested with classic editor as well 👍 
- ^ Trashing a post deletes it CPCSS as well 👍 

Everything good so far 🚀 

**NOTES**

On the [mega site](http://mega.wp-rocket.me/wp-admin/options-general.php?page=wprocket#cache), I am seeing the case of the infinite loop (#1333): 
image

Something else regarding the position of the "Regenerate Critical Path CSS" item in the admin bar. 

- On this feature branch the admin bar item is the last one after "Support"
image

- This is how it used to be, even on version 3.6-alpha1:
image

- Why was the position moved? 

Tested this scenario that Piotr recorded: 
- Enable "Optimize CSS Delivery" without "Separate Cache for Mobile Devices" enabled
- Files are generated, disable "Optimize CSS Delivery"
- Files are not removed
- Enable "Separate Cache for Mobile Devices"
- Enable "Optimize CSS Delivery"

Result: CPCSS for mobile was deleted correctly. 
But the desktop CPCSS was deleted first and then regenerated. Is this necessary? Video: https://youtu.be/k1dyjQ3_TTs

hellofromtonya:
@crystinutzaa @Tabrisrp Please look at the Arun's notes and questions including:

- On the mega site, I am seeing the case of the infinite loop (#1333)
- Something else regarding the position of the "Regenerate Critical Path CSS" item in the admin bar.
- But the desktop CPCSS was deleted first and then regenerated. Is this necessary?
- Mobile notices

Are there actions we need to take?

Tabrisrp:
> On the mega site, I am seeing the case of the infinite loop (#1333)

Was it infinite, or did it stop eventually? I'm wondering if it's related to what we discussed last week, with the possible read/write of the transients at the same time by 2 different processes.

> Something else regarding the position of the "Regenerate Critical Path CSS" item in the admin bar.

This is because of the change I made to bring this admin bar item code into the module. The admin bar system doesn't contain a way to order sub-items like this, except by using hook priority. To preserve the existing ordering, I would have to re-do all the related admin bar code in the plugin to use appropriate priorities. This didn't fit into what we are doing here, so I went with the current approach, that is not perfect but preserve the functionality.

> But the desktop CPCSS was deleted first and then regenerated. Is this necessary?

This scenario considers that we want to generate both versions of the files, because "Optimize CSS Delivery" was not enabled before (even though the desktop files exists). So it's going to delete everything, and recreate them. We also have no way to know if the existing files are still valid or not, maybe they're old, or they changed plugins/themes in between changing the "Optimize CSS Delivery" option value?

> Mobile notices

We can skip adding them into the notice if it's the approach that the product team wants. It seems to be the original intent, but it was missed in our communication until now. Are we fine with users not knowing if a mobile version was not generated correctly?

hellofromtonya:
>Was it infinite, or did it stop eventually? I'm wondering if it's related to what we discussed last week, with the possible read/write of the transients at the same time by 2 different processes.

How can we validate the _**root cause**_ of this problem? @Tabrisrp

hellofromtonya:
For Mobile Notices:

>We can skip adding them into the notice if it's the approach that the product team wants. It seems to be the original intent, but it was missed in our communication until now. Are we fine with users not knowing if a mobile version was not generated correctly?

Yes, this is something Product wants:

>>We had agreed not to display separate notices for mobile, sorry if this was not communicated widely - https://wp-media.slack.com/archives/G8L4EE8N9/p1590096487031800?thread_ts=1590088503.020600&cid=G8L4EE8N9

Let's create a PR to remove the mobile notices and update the tests please. @crystinutzaa @Tabrisrp

Tabrisrp:
> Does this mean that eventually as we redo other aspects of the admin bar, we will have the original order in the future?

Yes, in the end we will

arunbasillal:
@Tabrisrp I had a chat with Jonathan regarding the position of "Regenerate CPCSS" button. Adding it in the end doesn't look good and can be confusing for customers. 

Jonathan said, `A code update shouldn’t change the current state / behavior of the plugin`. 

Shall we do one of these?
- Update everything in the admin bar so that the original order is preserved. Not sure how much time it needs. 
- Revert it to how it used to be. We can update everything related to this in a different sprint. 

cc: @hellofromtonya

hellofromtonya:
@arunbasillal @piotrbak This PR is ready for re-QA.

Things fixed from Arun's last QA:

- Only disables desktop notices (no mobile notices)
- Fixed the order of the admin bar menu items
- Fixed the count

arunbasillal:
- Notices doesn't show separate mobile CPCSS status 👍 
- Admin bar looks good 👍 
- Count looks good 👍 

Noticed an issue, to replicate:

- Make sure "Separate Cache for mobile" is disabled. 
- Enable Optimize CSS delivery. Let the generation complete. 
- Enable "Separate cache for mobile"
- This error happens. But file generated correctly. 
image
Video: https://youtu.be/fyfMEQv79x0

Another issue:

- When separate cache for mobile is enabled
- The complete notice says "complete" but in `blue` and asks customers to refresh 
image
- Then it completes with the same message
image

Maybe this is the time it takes to generate mobile CPCSS? Can this situation be avoided? 
Video (from 1:25s - https://youtu.be/qLwKGPwjSzI)

crystinutzaa:
@arunbasillal I fixed the first issue with the "unwanted 0 notice". Can you please recheck and let me know if is ok.

As for the second issue, the issue is with Mobile CPCSS. It takes time to complete so the notice will be displayed only when it finished totally. 
This is mandatory because we have 2 major actions on it:
```
                 /**
		 * Fires when the critical CSS generation process is complete.
		 *
		 * @since 2.11
		 */
		do_action( 'rocket_critical_css_generation_process_complete' );

		rocket_clean_domain();
```
Basically if we display the complete notice faster ( and the mobile did not finished) then rocket_clean_domain() will not run until the mobile is finished. So the customer will simply report that something if off if it does not wait / recheck in a while after the mobile generation finished too.
I don't think this notice should display the complete until mobile is generated too. Otherwise it may lead to a lot of confusion. 

@Tabrisrp & @hellofromtonya & @GeekPress ideas?

For that notice we can simply it and provide the following format. Basically the notice will contain both notices for Mobile & Desktop but with the details hidden. And we can provide a link / button (details) when clicked will enlarge and display the full list with each url. In this way the customer can be 100% informed about the CPCSS status.

https://www.screencast.com/t/2j7iIhHUc9Z

hellofromtonya:
The blue and then green messaging is due to non-mobile is done but mobile is still generating. Therefore, **_CPCSS generation is not done._**

CPCSS generation is done when both mobile and non-mobile are done for each path/URL.

Given that:

@crystinutzaa We'd need to change the behavior of the notice handler to only update the notice and count when both mobile and non-mobile are done for each path/URL.

By doing 👆, then the message will stay in sync and color match for the user's experience. The updating of the notice will be slower because we have 2 tasks running for each URL/path. But that's to be expected.

hellofromtonya:
@arunbasillal @GeekPress We are moving forward with [what I proposed above](https://github.com/wp-media/wp-rocket/pull/2672#issuecomment-638082701): 

- Notice and count for each URL/path only display _when_ both non-mobile and mobile CPCSS generation are done.

The team is brainstorming. We'll keep you in the loop.

arunbasillal:
- I retested the `unwanted 0 notice` and it is gone 👍 
- Mobile CPCSS is generated in the background 👍 

Thanks for the explanation @hellofromtonya. This is the ideal behavior. 

@crystinutzaa To answer your suggestions, the idea of not showing mobile specific CPCSS related notice is not to confuse customers. We generate it only when `separate cache for mobile` is enabled and having it on some websites and not having it on others (or when they change the `separate cache for mobile` setting) can be confusing. 

- The probability of desktop CPCSS generation being successful and mobile CPCSS generation to fail is low (this is an assumption of course). 
- Even if mobile CPCSS generation fails, the desktop CPCSS will be used as fallback. From our experience so far, for most responsive themes this is perfectly fine.

arunbasillal (approved):
Did a quick retest:

- Separate notice for mobile CPCSS is not displayed 👍 
- "unwanted 0 notice" stays fixed 👍 
- Count of post types and generated post types is consistent (no infinite or excess count) 👍 
- Notice and count for each URL/path only display when both non-mobile and mobile CPCSS generation are done. 👍 

Everything reported in the last two days looks good so far. I would love to repeat the QA completely from the beginning, but it's a long one and its too late for me today. I will do it on the alpha version. 

Thanks and congrats to everyone 🚀

PR #2673, opened 2020-05-21
Correctly generate CPCSS on Optimize CSS delivery activation

engahmeds3ed:
Solve the issue of checking critical css folder that has any files to bailout and stop generation.
## Solution

- [x] Check the critical css path if has files or not
- [x] Skip folders and don't check them

### DIFF
--- a/inc/Engine/CriticalPath/CriticalCSSSubscriber.php
+++ b/inc/Engine/CriticalPath/CriticalCSSSubscriber.php
@@ -168,17 +168,16 @@ class CriticalCSSSubscriber implements Subscriber_Interface {
 
 		try {
 			$critical_css_path_file_iterator = ( new FilesystemIterator( $critical_css_path, FilesystemIterator::SKIP_DOTS ) );
-			if ( $critical_css_path_file_iterator->valid() ) {
-				// Bail out if this folder has no files (not folders).
-				foreach ( $critical_css_path_file_iterator as $file ) {
-					if ( $file->isFile() ) {
-						return;
-					}
-				}
-			}
 		} catch ( UnexpectedValueException $e ) {
 			// Bail out when folder is invalid.
 			return;
+		}
+
+		// Bail out if this folder has no files (not folders).
+		foreach ( $critical_css_path_file_iterator as $file ) {
+			if ( $file->isFile() ) {
+				return;
+			}
 		}
 
 		// Generate the CPCSS files.
--- a/inc/Engine/CriticalPath/CriticalCSSSubscriber.php
+++ b/inc/Engine/CriticalPath/CriticalCSSSubscriber.php
@@ -173,7 +173,7 @@ class CriticalCSSSubscriber implements Subscriber_Interface {
 			return;
 		}
 
-		// Bail out if this folder has no files (not folders).
+		// Bail out if this folder has no files in it.
 		foreach ( $critical_css_path_file_iterator as $file ) {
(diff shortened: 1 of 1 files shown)


inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -167,9 +167,14 @@ public function generate_critical_css_on_activation( $old_value, $value ) {
...
-				return;
+			$critical_css_path_file_iterator = ( new FilesystemIterator( $critical_css_path, FilesystemIterator::SKIP_DOTS ) );
+			if ( $critical_css_path_file_iterator->valid() ) {
+				// Bail out if this folder has no files (not folders).
+				foreach ( $critical_css_path_file_iterator as $file ) {
+					if ( $file->isFile() ) {
+						return;
+					}
+				}
 			}
```
hellofromtonya, line 178: 1. The `->valid()` if conditional is no longer needed. Why? We are iterating through each entry via the `foreach`. The `valid()` method is handy when doing a `while()` loop.
2. Let's move the `foreach()` out of the `try`. Why? The `try` is there to catch errors when instantiating the iterator itself. It's not needed for the `foreach`. By moving it out of the `foreach`, it'll be more readable and consistent with our other SPL logic.

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -167,15 +167,19 @@ public function generate_critical_css_on_activation( $old_value, $value ) {
 		}
 
 		try {
-			if ( ( new FilesystemIterator( $critical_css_path, FilesystemIterator::SKIP_DOTS ) )->valid() ) {
-				// Bail out if the folder is not empty.
-				return;
-			}
+			$critical_css_path_file_iterator = ( new FilesystemIterator( $critical_css_path, FilesystemIterator::SKIP_DOTS ) );
```
Tabrisrp, line 170:
```suggestion
			$critical_css_path_file_iterator = new FilesystemIterator( $critical_css_path, FilesystemIterator::SKIP_DOTS );
```

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -167,15 +167,19 @@ public function generate_critical_css_on_activation( $old_value, $value ) {
...
 
+		// Bail out if this folder has no files (not folders).
+		foreach ( $critical_css_path_file_iterator as $file ) {
+			if ( $file->isFile() ) {
+				return;
+			}
+		}
+
 		// Generate the CPCSS files.
 		$this->critical_css->process_handler();
```
hellofromtonya, line 184: I don't think we want to bail out if we find a file. Rather, we want to bail out if there are no directories. 

@Tabrisrp is this right?

if yes, I'd suggest this code instead:

```suggestion
		// Bail out if there are no directories within the critical-css.
		foreach ( $iterator as $entry ) {
			if ( $entry->isDir() ) {
				// Generate the CPCSS files.
				return $this->critical_css->process_handler();;
			}
		}
```

Suggestion 2: Shrink the variable name down to `$iterator` to save memory.

Tabrisrp:

Testing:
- Have “Optimize CSS delivery” disabled
- Check the option
- Save the settings
- It should launch the CPCSS generation automatically (if there is no files in the `critical-css/1/` folder)

PR #2688, opened 2020-05-28
Add tests for function rocket_clean_home

engahmeds3ed:
- Add Unit and Integration tests for `rocket_clean_home` method which locates in `inc/functions/files.php`
- Refactor the code itself to use SPL iterators instead of glob function.
Why?
  - To be able to make the code testable.
  - To make the code faster as this function is called multiple times through the code.

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -667,7 +667,7 @@ function rocket_clean_home( $lang = '', $filesystem = null ) {
 		if ( ! $filesystem->exists( $domain_entry ) ) {
 			continue;
 		}
-		$iterator     = new DirectoryIterator( $domain_entry );
+		$iterator = new DirectoryIterator( $domain_entry );
 
 		// Delete homepage.
 		// Remove the hidden empty file for mobile detection on NGINX with the Rocket NGINX configuration.
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -680,7 +680,7 @@ function rocket_clean_home( $lang = '', $filesystem = null ) {
 
 		// Delete homepage pagination.
 		$pagination_dir = $domain_entry . DIRECTORY_SEPARATOR . $GLOBALS['wp_rewrite']->pagination_base;
-		rocket_rrmdir( $pagination_dir );
+		rocket_rrmdir( $pagination_dir,[], $filesystem );
 
 	}
 	/**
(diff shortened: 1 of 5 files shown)


inc/functions/files.php
```diff
@@ -656,38 +662,28 @@ function rocket_clean_home( $lang = '' ) {
...
+			continue;
 		}
-	}
-
-	// Remove the hidden empty file for mobile detection on NGINX with the Rocket NGINX configuration.
-	$nginx_mobile_detect_files = glob( $root . '/.mobile-active', GLOB_BRACE | GLOB_NOSORT );
-	if ( $nginx_mobile_detect_files ) {
-		foreach ( $nginx_mobile_detect_files as $nginx_mobile_detect_file ) { // no array map to use @.
-			rocket_direct_filesystem()->delete( $nginx_mobile_detect_file );
+		$iterator = new DirectoryIterator( $domain_entry );
```
Tabrisrp, line 670: You should wrap this in a `try/catch` because the constructor can throw an exception: https://www.php.net/manual/en/directoryiterator.construct.php

inc/functions/files.php
```diff
@@ -656,38 +662,28 @@ function rocket_clean_home( $lang = '' ) {
...
 		}
-	}
 
-	// Remove the hidden empty file for webp.
-	$nowebp_detect_files = glob( $root . '/.no-webp', GLOB_BRACE | GLOB_NOSORT );
-	if ( $nowebp_detect_files ) {
-		foreach ( $nowebp_detect_files as $nowebp_detect_file ) { // no array map to use @.
-			rocket_direct_filesystem()->delete( $nowebp_detect_file );
+		// Delete homepage pagination.
+		$pagination_dir = $domain_entry . DIRECTORY_SEPARATOR . $GLOBALS['wp_rewrite']->pagination_base;
```
Tabrisrp, line 682: I'm thinking we should use a `global $wp_rewrite` at the start of the function, and then use the following here:

```suggestion
		$pagination_dir = $domain_entry . DIRECTORY_SEPARATOR . $wp_rewrite->pagination_base;
```

iCaspar:
This PR has been ported to [module-filesystem, PR #1](https://github.com/wp-media/module-filesystem/pull/1)

PR #2693, opened 2020-05-31
Add tests for rocket_clean_cache_dir function

engahmeds3ed:
- [x] Add Unit and Integration tests for rocket_clean_cache_dir method which locates in inc/functions/files.php

- [x] Refactor the code itself to use SPL iterators instead of glob function.

  Why?
- To be able to make the code testable.
- To make the code faster as this function is called multiple times through the code.

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -979,10 +979,11 @@ function rocket_clean_cache_dir() {
 	*/
 	do_action( 'before_rocket_clean_cache_dir' ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
 
+	$filesystem = rocket_direct_filesystem();
+
 	// Delete all caching files.
-	$dirs = glob( WP_ROCKET_CACHE_PATH . '*', GLOB_NOSORT );
-	if ( $dirs ) {
-		foreach ( $dirs as $dir ) {
+	foreach ( _rocket_get_cache_dirs( '.' ) as $dir ) {
+		if ( $filesystem->is_dir( $dir ) ) {
 			rocket_rrmdir( $dir );
 		}
 	}
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -982,7 +982,8 @@ function rocket_clean_cache_dir() {
 	$filesystem = rocket_direct_filesystem();
 
 	// Delete all caching files/folders.
-	foreach ( _rocket_get_cache_dirs( '.' ) as $path ) {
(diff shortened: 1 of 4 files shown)


inc/functions/files.php
```diff
@@ -979,11 +979,14 @@ function rocket_clean_cache_dir() {
...
 
-	// Delete all caching files.
-	$dirs = glob( WP_ROCKET_CACHE_PATH . '*', GLOB_NOSORT );
-	if ( $dirs ) {
-		foreach ( $dirs as $dir ) {
-			rocket_rrmdir( $dir );
+	$filesystem = rocket_direct_filesystem();
+
+	// Delete all caching files/folders.
+	foreach ( _rocket_get_cache_dirs( '.' ) as $path ) {
```
Tabrisrp, line 985: `_rocket_get_cache_dirs()` expects an URL host as the first parameter, passing it a dot like this is strange.

How about using this instead?

```suggestion
	foreach ( _rocket_get_cache_path_iterator( _rocket_get_wp_rocket_cache_path() ) as $path ) {
```

inc/functions/files.php
```diff
@@ -979,11 +979,14 @@ function rocket_clean_cache_dir() {
...
+		if ( ! $filesystem->is_dir( $path ) ) {
+			$filesystem->delete( $path );
+		}else {
+			rocket_rrmdir( $path );
```
Tabrisrp, line 989:
```suggestion
		if ( ! $filesystem->is_dir( $path ) ) {
			$filesystem->delete( $path );
			continue;
		}

		rocket_rrmdir( $path );
```

engahmeds3ed: Really inside loops I don't like using continue and break as I think they are `hidden traps` and makes code not readable, what do you think?

hellofromtonya: The same can be said of `else`, as some think `else` blocks are evil.

A good rule of thumb is this: 

- If no other processing should happen after the `if` block, then bail out: `return`, `continue`, or `break`.

Why?  Nothing else should happen after this `if` is done. Therefore, adding a bail out makes it very clear that the work is done.

engahmeds3ed: Great, nice Idea.
Thanks @Tabrisrp @hellofromtonya .

iCaspar:
This PR was ported to [module-filesystem, PR #3](https://github.com/wp-media/module-filesystem/pull/3)

PR #2701, opened 2020-06-02
Update WP Rocket admin bar items order

Tabrisrp:
Add new hook `rocket_admin_bar_items` to preserve the order of the sub-items in the WP Rocket admin bar menu.

admin-bar


hellofromtonya:
@Tabrisrp Can you add a screenshot if the menu item order with this branch? It'll give us context and validation for the future.

Tabrisrp:
Added on the opening comment

PR #2718, opened 2020-06-04
Prevent php notice: Unavailable index in Beacon

Tabrisrp:
- Add `Unavailable` index to the array
- Update fixtures


engahmeds3ed:
@Tabrisrp it's a fast solution but I think we need not to save the transient till data comes from wp-rocket server as if that case happened, I need to wait 24 hours to the transient to be expired before checking the server again, right?

PR #2722, opened 2020-06-05
Adjust CPCSS messages to differentiate between custom and specific page types

engahmeds3ed:
Reverts CPCSS notice back per request from @GeekPress in [slack](https://wp-media.slack.com/archives/GUT7FLHF1/p1591288051020400):

>Can we put back the previous behavior: instead of the URL, having the CPT / Taxonomy slug
>
> Why? It’s going to be confused by just displaying that. We can think that the CPCSS is generated only for the pages displayed into the notice.
>
> The URL should be displayed only for custom CPCSS pages

Adjust CPCSS messages to check if this is custom page or a page type so that we can show the url inside the messages or the type itself.

## TODO

- [x] \WP_Rocket\Engine\CriticalPath\ProcessorService
- [x] \WP_Rocket\Engine\CriticalPath\DataManager
- [x] \WP_Rocket\Engine\CriticalPath\APIClient
- [x] process_generate method usages inside RESTWP
- [x] process_generate method usages inside wpBackgroundProcess task
- [x] process_generate method usages inside \WP_Rocket\Engine\CriticalPath\Admin\Admin Hearbeat
- [x] Unit tests
- [x] Integration tests
---------------------------------
- [x] Make CPCSS fail with descriptive message if external requests are blocked.

### DIFF
--- a/inc/Engine/CriticalPath/ProcessorService.php
+++ b/inc/Engine/CriticalPath/ProcessorService.php
@@ -40,18 +40,18 @@ class ProcessorService {
 	 *
 	 * @param string $item_url  URL for item to be used in error messages.
 	 * @param string $item_path Path for item to be processed.
-	 * @param array  $additional_parameter additional parameters for generation.
+	 * @param array  $additional_parameters additional parameters for generation.
 	 *
 	 * @return array|WP_Error
 	 */
-	public function process_generate( $item_url, $item_path, $additional_parameter = [
+	public function process_generate( $item_url, $item_path, $additional_parameters = [
 		'timeout'   => false,
 		'is_mobile' => false,
 		'item_type' => 'custom',
 	] ) {
-		$timeout   = isset( $additional_parameter['timeout'] ) ? $additional_parameter['timeout'] : false;
-		$is_mobile = isset( $additional_parameter['is_mobile'] ) ? $additional_parameter['is_mobile'] : false;
-		$item_type = isset( $additional_parameter['item_type'] ) ? $additional_parameter['item_type'] : 'custom';
+		$timeout   = isset( $additional_parameters['timeout'] ) ? $additional_parameters['timeout'] : false;
+		$is_mobile = isset( $additional_parameters['is_mobile'] ) ? $additional_parameters['is_mobile'] : false;
+		$item_type = isset( $additional_parameters['item_type'] ) ? $additional_parameters['item_type'] : 'custom';
 
 		// Ajax call requested a timeout.
 		if ( $timeout ) {
--- a/inc/Engine/CriticalPath/ProcessorService.php
+++ b/inc/Engine/CriticalPath/ProcessorService.php
@@ -44,27 +44,26 @@ class ProcessorService {
 	 *
 	 * @return array|WP_Error
 	 */
-	public function process_generate( $item_url, $item_path, $additional_parameters = [
-		'timeout'   => false,
-		'is_mobile' => false,
-		'item_type' => 'custom',
-	] ) {
-		$timeout   = isset( $additional_parameters['timeout'] ) ? $additional_parameters['timeout'] : false;
-		$is_mobile = isset( $additional_parameters['is_mobile'] ) ? $additional_parameters['is_mobile'] : false;
-		$item_type = isset( $additional_parameters['item_type'] ) ? $additional_parameters['item_type'] : 'custom';
+	public function process_generate( $item_url, $item_path, $additional_parameters = [] ) {
+		$defaults = [
+			'timeout'   => false,
+			'is_mobile' => false,
+			'item_type' => 'custom',
+		];
+		$args = array_merge( $defaults, $additional_parameters );
 
 		// Ajax call requested a timeout.
-		if ( $timeout ) {
-			return $this->process_timeout( $item_url, $is_mobile, $item_type );
+		if ( $args['timeout'] ) {
+			return $this->process_timeout( $item_url, $args['is_mobile'], $args['item_type'] );
 		}
 
-		$cpcss_job_id = $this->data_manager->get_cache_job_id( $item_url, $is_mobile );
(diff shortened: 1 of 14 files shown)


inc/Engine/CriticalPath/ProcessorService.php
```diff
@@ -40,24 +40,31 @@ public function __construct( DataManager $data_manager, APIClient $api_client )
...
 	 */
-	public function process_generate( $item_url, $item_path, $timeout = false, $is_mobile = false ) {
+	public function process_generate( $item_url, $item_path, $additional_parameters = [
+		'timeout'   => false,
+		'is_mobile' => false,
+		'item_type' => 'custom',
+	] ) {
+		$timeout   = isset( $additional_parameters['timeout'] ) ? $additional_parameters['timeout'] : false;
+		$is_mobile = isset( $additional_parameters['is_mobile'] ) ? $additional_parameters['is_mobile'] : false;
+		$item_type = isset( $additional_parameters['item_type'] ) ? $additional_parameters['item_type'] : 'custom';
```
Tabrisrp, line 54: Instead of doing it like this, we could use a `$defaults` array with the default value, and then use [wp_parse_args()](https://developer.wordpress.org/reference/functions/wp_parse_args/) to merge the parameter array with the default array:

`$args = wp_parse_args( $additional_parameters, $defaults );`

That way we always know that the index are all created with a value. What do you think?

remyperona (changes requested):

inc/Engine/CriticalPath/ProcessorService.php
engahmeds3ed: Amazing, totally right.
Adjusting it right now

hellofromtonya: Or an even faster way is to use `array_merge()`.  Why? You know you have an array that needs to be merged with a default array. `wp_parse_args()` is nice when merging different data types. But when it's an array, it uses `array_merge()`. For us, we can skip all the checks in `wp_parse_args()` and just use PHP's `array_merge()` directly. That will be faster and use a little less memory.

tests/Fixtures/inc/Engine/CriticalPath/ProcessorService/processGenerate.php
```diff
@@ -36,11 +36,12 @@
...
+				'type'               => 'post'
 			],
 			'expected' => [
 				'success' => false,
 				'code'    => 'cpcss_generation_timeout',
-				'message' => 'Critical CSS for http://example.org/?p=21 timeout. Please retry a little later.',
+				'message' => 'Critical CSS for post timeout. Please retry a little later.',
 				'data'    => [
 					'status' => 400,
 				],
```
hellofromtonya, line 47: Let's add other types to the fixture too to ensure it behaves as expected.

hellofromtonya (changes requested):

inc/Engine/CriticalPath/ProcessorService.php
engahmeds3ed: Yes I opened the implementation for `wp_parse_args` and it converts the first argument to array based on its type then apply `array_merge` on this argument array and defaults (second argument).
and mainly at unit testing I should mock it by using array_merge.
I imagined that using wordpress abstraction layer is good enough but I think on this case both are the same (from abstraction prespective).
Action: I will use `array_merge`

Tabrisrp: Agreed

tests/Fixtures/inc/Engine/CriticalPath/ProcessorService/processGenerate.php
engahmeds3ed: Done, now I added for homepage and category pages.
it needed some adjustments in tests and fixtures

hellofromtonya (approved):

remyperona (approved):

hellofromtonya:
@arunbasillal This PR is ready for QA.

arunbasillal:
Hello 👋 

This is how it used to be in WP Rocket 3.5.5.1
image

How it looks like in this feature branch
image

So far so good 👍 

To test a failure, I added `define( 'WP_HTTP_BLOCK_EXTERNAL', true );` in `wp-config.php`

3.5.5.1: 
image

feature branch:
image

Can we put back the old format of errors? (What do you think @GeekPress)? 
Our doc references this error as `Invalid response code` - https://docs.wp-rocket.me/article/1267-troubleshooting-critical-css-generation-issues

Did another test with feature branch on local. The error messages looks good: 
image

hellofromtonya:
@engahmeds3ed Can you take of restoring the old format of errors?

remyperona (approved):

engahmeds3ed:
@hellofromtonya Sure got it
@arunbasillal Could you please give it a try now, I adjusted it to be `Critical CSS for XXXX not generated. Error: XXXX`

arunbasillal:
Retested

On local:

image

image

On main mega site:

image

image

After adding `define( 'WP_HTTP_BLOCK_EXTERNAL', true );`: 

image

image

- One difference I notice here is that in `3.5.5.1` we used to say, `The API returned an invalid response code`
- The current error, `User has blocked requests through HTTP.` seems more appropriate. 

I recommend that we keep this and update [our doc](https://docs.wp-rocket.me/article/1267-troubleshooting-critical-css-generation-issues) to reflect the same. (cc @webtrainingwheels)

arunbasillal (approved):

webtrainingwheels:
Thanks @arunbasillal ! Will update the docs accordingly!

arunbasillal:
@webtrainingwheels 

- I was looking at the code and we still have `Invalid response code` part in the code. So let's preserve that. 
- I think the change we can do is to add a section for `User has blocked requests through HTTP` which happens when user has set `define( 'WP_HTTP_BLOCK_EXTERNAL', true );` via `wp-config.php` or a security plugin and if `167.114.226.142` isn't whitelisted. 

🙏

PR #2727, opened 2020-06-05
Fixed unwanted notice for mobile CPCSS ( early bailout if no desktop message )

crystinutzaa:
Early bailout if no desktop message from CPCSS in the transient


hellofromtonya:
@arunbasillal This one is ready for you to re-QA.

PR #2731, opened 2020-06-05
Exclude Gleam.io JS From Being Merged

ScottTravisHartley:
If you're using a gleam.io widget on a website and the combine js feature is enabled the widget will not render. This can be fixed by excluding the JS in the plugin options. (had to do this on a client site basically the widget never renders). This patch simply makes it permanent to save people time and to avoid issues.


ScottTravisHartley:
Full JavaScript file that is referenced for the embed widget. 
<script type="text/javascript" src="https://widget.gleamjs.io/e.js" async="true"></script>

PR #2740, opened 2020-06-09
Solve error 500 on WP Rocket activation - Opening comment tag in wp-config.php removed

### LINKED ISSUE #2537: 500 error on WP Rocket activation - Opening comment tag in wp-config.php removed
- Version 3.5.2

**Describe the bug**
When the opening comment tag is set on the **1st** line of the `wp-config.php` file, the WP Rocket activation will remove it causing a 500 error (parse error).

**To Reproduce**
1. Edit the 1st line of the `wp-config.php` file like this: http://snippi.com/s/pcr3v72.
Note that the opening comment tag of the default comment is placed on the 1st line instead of the 2nd line.
2. Activate WP Rocket. This will add the `define( 'WP_CACHE', true ); // Added by WP Rocket` line in the `wp-config.php` file.
3. See the **500 error** on the website.
4. The resulting `wp-config.php` file looks like this: http://snippi.com/s/t4a6o7o.

**Expected behavior**
Adding the `define( 'WP_CACHE', true ); // Added by WP Rocket` line in the `wp-config.php` file should not remove the opening comment tag.

**Additional context**
Only happens for the  `wp-config.php` files where default comment opening tag is placed on the 1st line instead of the 2nd line.
**Ticket**: https://secure.helpscout.net/conversation/1129985389/156210?folderId=2675957.
**Slack** thread: https://wp-media.slack.com/archives/C08N8J6VC/p1586811121269400.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

engahmeds3ed:
Closes #2537 
Here I will add after the php starter keyword `<?php` our constant `WP_CACHE` not to replace the full line with the constant.
That will solve the following issues:-
1. If there is a comment starter or a single line comment on the first line beside `<?php`, it won't be removed.
2. The default case is that `<?php` will be alone in the first line.

### DIFF
--- a/tests/Integration/inc/functions/setRocketWpCacheDefine.php
+++ b/tests/Integration/inc/functions/setRocketWpCacheDefine.php
@@ -20,8 +20,7 @@ class Test_SetRocketWpCacheDefine extends FilesystemTestCase {
 	protected $path_to_test_data = '/inc/functions/setRocketWpCacheDefine.php';
 	protected $config_file = '';
 
-	public function tearDown()
-	{
+	public function tearDown() {
 		remove_filter( 'rocket_wp_config_name', [ $this, 'setWpCacheFilePath' ] );
 		parent::tearDown();
 	}
@@ -43,7 +42,7 @@ class Test_SetRocketWpCacheDefine extends FilesystemTestCase {
 		$this->assertEquals( $expected, str_replace( "\r\n", "\n", $actual ) );
 	}
 
-	public function setWpCacheFilePath($file) {
+	public function setWpCacheFilePath( $file ) {
 		return $this->config_file;
 	}
 }
--- a/tests/Integration/inc/functions/setRocketWpCacheDefine.php
+++ b/tests/Integration/inc/functions/setRocketWpCacheDefine.php
@@ -1,9 +1,9 @@
 <?php
 
-namespace WP_Rocket\Tests\Unit\inc\functions;
+namespace WP_Rocket\Tests\Integration\inc\functions;
 
 use Brain\Monkey\Functions;
-use WP_Rocket\Tests\Unit\FilesystemTestCase;
+use WP_Rocket\Tests\Integration\FilesystemTestCase;
 
 /**
  * @covers ::set_rocket_wp_cache_define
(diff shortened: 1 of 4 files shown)


remyperona (approved):

tests/Integration/inc/functions/setRocketWpCacheDefine.php
```diff
@@ -0,0 +1,48 @@
...
+ */
+class Test_SetRocketWpCacheDefine extends FilesystemTestCase {
+	protected $path_to_test_data = '/inc/functions/setRocketWpCacheDefine.php';
+	protected $config_file = '';
+
+	public function tearDown()
+	{
+		remove_filter( 'rocket_wp_config_name', [$this, 'setWpCacheFilePath'] );
+		parent::tearDown();
+	}
```
hellofromtonya, line 27: Formatting

```suggestion
	public function tearDown() {
		remove_filter( 'rocket_wp_config_name', [ $this, 'setWpCacheFilePath' ] );
		parent::tearDown();
	}
```

tests/Integration/inc/functions/setRocketWpCacheDefine.php
```diff
@@ -0,0 +1,48 @@
...
+
+	/**
+	 * @dataProvider providerTestData
+	 */
+	public function testShouldAddWpCacheDefine( $config, $expected ) {
+		$this->config_file = $config['file'];
+		$config_file_full_path = $this->config['vfs_dir'] . $this->config_file . '.php';
+		add_filter( 'rocket_wp_config_name', [$this, 'setWpCacheFilePath'] );
```
hellofromtonya, line 35: Formatting

```suggestion
		add_filter( 'rocket_wp_config_name', [ $this, 'setWpCacheFilePath' ] );
```

hellofromtonya (changes requested):

hellofromtonya (approved):

hellofromtonya:

crystinutzaa (approved):

arunbasillal (changes requested):
- Recreated the issue using `master` branch
image

- Retested this in fix branch. Issue is fixed for the specific case
image

However, the value doesn't change on WP Rocket deactivation. 

I tested this with the normal state of `wp-config.php` and the same issue. WP Rocket adds `define('WP_CACHE', true); // Added by WP Rocket` but does not set it to false. 

Video: https://youtu.be/3mlRA4naQuI

engahmeds3ed:
Thanks @arunbasillal , checking now

engahmeds3ed:
@hellofromtonya @Tabrisrp @crystinutzaa @arunbasillal I adjusted the code to check if the constant is already found and the value to be changed, so I will leave it and if value changed, such as on deactivation, I will replace this line with the new line with the new value

remyperona (approved):

arunbasillal (approved):
- Retested this. Added correctly on activation 
image

- Updated correctly on deactivation
image

- Updated correctly on re-activation. 
- When `define('WP_CACHE', true);` was missing and WP Rocket was active, this was added back on refresh of admin. 

**Note:** A trivial thing I noticed is that if the `wp-config.php` is as normal, then there is an extra space added. 
image

This is not the case with current master branch. No extra space is added 
image

Don't think this will cause issues. I tried deactivating and reactivating WP Rocket many times, but new lines were not added on each activation and deactivation 👍 

I will leave it upto the Dev team to see if this should be changed. Good from my side.

PR #2755, opened 2020-06-10
Remove lazyload from search page

### LINKED ISSUE #2535: LazyLoad CSS and JS file not removed on search page
The usual WordPress search query page isn't cached by default. The URL is like this: 
`example.com/?s=query`

But if LazyLoad is active, the inline CSS and JS file are still there. 

**To Reproduce**
1. Enable LazyLoad
2. Perform a search, or visit a URL like: `example.com/?s=query`
3. View the page source. At the bottom you'll see the code above-mentioned code: 
https://jmp.sh/AEj7oAP

**Expected behavior**
It should not be there :)

**Additional context**
Very similar to this issue: https://github.com/wp-media/wp-rocket/issues/2523

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

engahmeds3ed:
Closes #2535 
Stop loading lazyload assets on search page.

### DIFF
--- a/inc/Engine/Media/LazyloadSubscriber.php
+++ b/inc/Engine/Media/LazyloadSubscriber.php
@@ -309,6 +309,8 @@ class LazyloadSubscriber implements Subscriber_Interface {
 	 * @return bool
 	 */
 	private function should_lazyload() {
+		// This filter is documented in inc/front/process.php.
+		$rocket_cache_search = apply_filters( 'rocket_cache_search', false );
 		if (
 			is_admin()
 			||
@@ -316,7 +318,7 @@ class LazyloadSubscriber implements Subscriber_Interface {
 			||
 			is_preview()
 			||
-			is_search()
+			( is_search() && ! $rocket_cache_search )
 			||
 			rocket_get_constant( 'REST_REQUEST', false )
 			||
--- a/inc/Engine/Media/LazyloadSubscriber.php
+++ b/inc/Engine/Media/LazyloadSubscriber.php
@@ -309,22 +309,31 @@ class LazyloadSubscriber implements Subscriber_Interface {
 	 * @return bool
 	 */
 	private function should_lazyload() {
-		// This filter is documented in inc/classes/Buffer/class-tests.php.
-		$rocket_cache_search = apply_filters( 'rocket_cache_search', false );
+		if (
+			rocket_get_constant( 'REST_REQUEST', false )
+			||
+			rocket_get_constant( 'DONOTLAZYLOAD', false )
+			||
+			rocket_get_constant( 'DONOTROCKETOPTIMIZE', false )
+		) {
+			return false;
+		}
+
 		if (
 			is_admin()
 			||
 			is_feed()
 			||
 			is_preview()
-			||
-			( is_search() && ! $rocket_cache_search )
-			||
-			rocket_get_constant( 'REST_REQUEST', false )
-			||
-			rocket_get_constant( 'DONOTLAZYLOAD', false )
-			||
-			rocket_get_constant( 'DONOTROCKETOPTIMIZE', false )
+		) {
+			return false;
+		}
+
+		// This filter is documented in inc/classes/Buffer/class-tests.php.
+		if (
+			is_search()
+			&&
+			! (bool) apply_filters( 'rocket_cache_search', false )
 		) {
 			return false;
 		}
--- a/inc/Engine/Media/LazyloadSubscriber.php
+++ b/inc/Engine/Media/LazyloadSubscriber.php
@@ -309,11 +309,21 @@ class LazyloadSubscriber implements Subscriber_Interface {
 	 * @return bool
 	 */
 	private function should_lazyload() {
-		if ( is_admin() || is_feed() || is_preview() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) || ( defined( 'DONOTLAZYLOAD' ) && DONOTLAZYLOAD ) ) {
-			return false;
-		}
-
-		if ( defined( 'DONOTROCKETOPTIMIZE' ) && DONOTROCKETOPTIMIZE ) {
+		if (
+			is_admin()
+			||
+			is_feed()
+			||
+			is_preview()
+			||
+			is_search()
+			||
+			( defined( 'REST_REQUEST' ) && REST_REQUEST )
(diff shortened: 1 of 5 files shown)


tests/Unit/inc/Engine/Media/LazyloadSubscriber/insertLazyloadScript.php
```diff
@@ -29,6 +29,7 @@ public function setUp() {
 		Functions\when( 'is_admin' )->justReturn( false );
 		Functions\when( 'is_feed' )->justReturn( false );
 		Functions\when( 'is_preview' )->justReturn( false );
+		Functions\when( 'is_search' )->justReturn( false );
```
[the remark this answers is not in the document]
engahmeds3ed: Refactored, now it's more dynamic and accepts more test cases

tests/Unit/inc/Engine/Media/LazyloadSubscriber/insertLazyloadScript.php
```diff
@@ -29,6 +29,7 @@ public function setUp() {
...
 		Functions\when( 'rocket_get_constant' )->justReturn( 'http://example.org/wp-content/plugins/wp-rocket/assets/' );
```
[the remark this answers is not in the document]
engahmeds3ed: I adjusted StubTrait to accept more dynamic constants now

inc/Engine/Media/LazyloadSubscriber.php
```diff
@@ -309,11 +309,21 @@ public function insert_youtube_thumbnail_style() {
...
+			||
+			is_preview()
+			||
+			is_search()
+			||
+			( defined( 'REST_REQUEST' ) && REST_REQUEST )
+			||
+			( defined( 'DONOTLAZYLOAD' ) && DONOTLAZYLOAD )
+			||
+			( defined( 'DONOTROCKETOPTIMIZE' ) && DONOTROCKETOPTIMIZE )
```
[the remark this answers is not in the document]
engahmeds3ed: Ok @Tabrisrp I did that now

inc/Engine/Media/LazyloadSubscriber.php
```diff
@@ -309,11 +309,23 @@ public function insert_youtube_thumbnail_style() {
...
+			is_preview()
+			||
+			( is_search() && ! $rocket_cache_search )
+			||
+			rocket_get_constant( 'REST_REQUEST', false )
+			||
+			rocket_get_constant( 'DONOTLAZYLOAD', false )
+			||
+			rocket_get_constant( 'DONOTROCKETOPTIMIZE', false )
+		) {
```
Tabrisrp, line 328: For cases where there is a lot of conditions to check, do we prefer splitting them into logical groups, or having them all in one place like this?

inc/Engine/Media/LazyloadSubscriber.php
```diff
@@ -309,11 +309,23 @@ public function insert_youtube_thumbnail_style() {
 	 * @return bool
 	 */
 	private function should_lazyload() {
-		if ( is_admin() || is_feed() || is_preview() || ( defined( 'REST_REQUEST' ) && REST_REQUEST ) || ( defined( 'DONOTLAZYLOAD' ) && DONOTLAZYLOAD ) ) {
-			return false;
-		}
-
-		if ( defined( 'DONOTROCKETOPTIMIZE' ) && DONOTROCKETOPTIMIZE ) {
+		// This filter is documented in inc/front/process.php.
```
Tabrisrp, line 312:
```suggestion
		// This filter is documented in inc/classes/Buffer/class-tests.php.
```

inc/Engine/Media/LazyloadSubscriber.php
hellofromtonya: It depends. Here there are a lot of them. What I often like to do is move them into a private method with a method name that _tells_ us what it's doing. Then the logic of the guard clause is encapsulated in its own method. What do you @Tabrisrp @engahmeds3ed think?

engahmeds3ed: You are right but this method do only two things:-
1. check these checks
2. check for the following fields if they are exist on $_GET `'fl_builder', 'et_fb', 'ct_builder'`
so it's simple function that almost do this check only.

hellofromtonya: Also the `apply_filters()` isn't necessary if `is_admin()` or `is_feed()` or `is_preview()` are true or if `is_search()` is `false`.

We can refactor this list of guard clauses to be more performant and readable.

```suggestion
		if ( 
			rocket_get_constant( 'REST_REQUEST', false )
			||
			rocket_get_constant( 'DONOTLAZYLOAD', false )
			||
			rocket_get_constant( 'DONOTROCKETOPTIMIZE', false )
		{
			return false;
		}

		if ( is_admin() || is_feed() || is_preview() {
			return false;
		}

		// This filter is documented in inc/classes/Buffer/class-tests.php.
		if ( is_search() && ! (bool) apply_filters( 'rocket_cache_search', false ) ) {
			return false;
		}
```

Let's break it down:

1. Check the constants first. Why? If one of those constants is `true`, we bail out quicker. Result: more performant, i.e. less memory and faster execution.
2. Only check the filter once all the other guard clauses pass. Why? More performant. Filters are slower and use more memory.

hellofromtonya: @Tabrisrp @engahmeds3ed What do you think of this approach 👆 ?

engahmeds3ed: Totally agree, this sequence order makes sense for me.

hellofromtonya: @Tabrisrp Any concerns?

Tabrisrp: I agree with this too

hellofromtonya: @engahmeds3ed Can you take care of this refactor for us?

engahmeds3ed: @hellofromtonya Sure

hellofromtonya:

PR #2758, opened 2020-06-10
Closes #2739 Clean post cache on Specific CPCSS generation or deletion

### LINKED ISSUE #2739: Clear post cache automatically when Specific CPCSS is generated or deleted
This relates to #2377 which is to be released in 3.6

Currently we [do not](https://wp-media.slack.com/archives/G8L4EE8N9/p1591647560002000) clear the cache of the post when specific CPCSS is generated or deleted. We could automate this. 

If preload is enabled, once the cache is deleted, the specific post could be preloaded as well. 

**To Reproduce**
- Enable Optimize CSS delivery. 
- Make sure a post is cached. 
- Generate specific CPCSS for this page. 
- The post cache is not deleted, nor the specific CPCSS is used. 
- Delete the cache of the post manually. The CPCSS will be updated with specific CPCSS. 

**Acceptance Criteria:**
- Post cache is cleared and preloaded (if preload is enabled) when Specific CPCSS is generated
- Post cache is cleared and preloaded when specific CPCSS is deleted. 

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
**Reproduce the issue** ✅ 
This is easy to be reproduces on local.

**Identify the root cause** ✅ 
The root cause is in the `RestWP.php`
https://github.com/wp-media/wp-rocket/blob/1455187900e01a22cc6ec8696511f1df01e18de7/inc/Engine/CriticalPath/RESTWP.php#L112
https://github.com/wp-media/wp-rocket/blob/1455187900e01a22cc6ec8696511f1df01e18de7/inc/Engine/CriticalPath/RESTWP.php#L223
In the initial specs of this one we forgot to include the cleaning of the cache for the specific page for which we regenerate the CPCSS.

**Scope a solution** ✅ 
The solution is to create a private method in `RESTWP.php` which cleans the cache based on post id. This method can be shared between delete() & generate() functions.

Preloading this page at this point is not an option because the code is already fragile and it may lead to other complications.

**Effort** ✅ 
Effort `[S]` . Unit tests and Integrations test will need to be updated to reflect this change

 TODO:
- [x] - Integration tests

### DIFF
--- a/inc/Engine/CriticalPath/RESTWP.php
+++ b/inc/Engine/CriticalPath/RESTWP.php
@@ -101,6 +101,17 @@ abstract class RESTWP implements RESTWPInterface {
 	}
 
 	/**
+	 * Clean post cache files on CPCSS generation or deletion.
+	 *
+	 * @since 3.6.1
+	 *
+	 * @param int $item_id ID for this item to get Url for.
+	 */
+	private function clean_post_cache( $item_id ) {
+		rocket_clean_files( $this->get_url( $item_id ) );
+	}
+
+	/**
 	 * Generates the CPCSS for the requested post ID.
 	 *
 	 * @since 3.6
@@ -159,6 +170,8 @@ abstract class RESTWP implements RESTWPInterface {
 				$this->return_error( $generated )
 			);
 		}
+
+		$this->clean_post_cache( $item_id );
 
 		return rest_ensure_response(
 			$this->return_success( $generated )
@@ -239,6 +252,8 @@ abstract class RESTWP implements RESTWPInterface {
 		if ( is_wp_error( $deleted ) ) {
 			return rest_ensure_response( $this->return_error( $deleted ) );
 		}
+
+		$this->clean_post_cache( $item_id );
 
 		return rest_ensure_response( $this->return_success( $deleted ) );
 	}
--- a/tests/Unit/inc/Engine/CriticalPath/RESTWPPost/generate.php
+++ b/tests/Unit/inc/Engine/CriticalPath/RESTWPPost/generate.php
@@ -75,7 +75,7 @@ class Test_Generate extends FilesystemTestCase {
 		$is_mobile                    = isset( $config['mobile'] )
 			? $config['mobile']
 			: false;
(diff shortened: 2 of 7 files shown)


inc/Engine/CriticalPath/RESTWP.php
```diff
@@ -100,6 +100,17 @@ public function check_permissions() {
...
 
+	/**
+	 * Clean post cache files on CPCSS generation or deletion.
+	 *
+	 * @since 3.6.1
+	 *
+	 * @param int $item_id ID for this item to get Url for.
+	 */
+	private function clean_post_cache( $item_id ) {
+		rocket_clean_files( $this->get_url( $item_id ) );
```
Tabrisrp, line 111: A reason to use `rocket_clean_files()` instead of `rocket_clean_post()` ?

crystinutzaa: Yes there is a reason. `rocket_clean_post()` will clean also Homepage & Post related links (like archives, adjacent posts, parent posts). 
However when the CPCSS for a specific page is regenerated it will affect only that specific page. So only the specific page needs a cache purge. 
Toughts?

Tabrisrp: That makes sense 👍

tests/Unit/inc/Engine/CriticalPath/RESTWPPost/generate.php
```diff
@@ -199,6 +199,15 @@ function( $post_id ) use ( $expected ) {
 			}
 		}
 
+		if ( $expected['success'] && 'cpcss_generation_successful' === $expected['code'] ) {
+			Functions\expect( 'rocket_clean_files' )
+				->atMost()
+				->times( 1 )
```
Tabrisrp, line 205: Is there a case where the function would not be called?

crystinutzaa: You are correct. In that situation exactly 1 time it should be called.
Updated the code to reflect this.

Tabrisrp:

The cache for the post should now be deleted when (re-)generating/deleting the CPCSS.

PR #2763, opened 2020-06-11
Fixes #2523 Prevent lazyload inline CSS insertion on AMP pages

### LINKED ISSUE #2523: LazyLoad Inline CSS isn't removed on pages generated by AMP plugin
**Describe the bug**
Since 1.50. AMP plugin introduced a new template mode: Transitional

When our LazyLoad is active, the inline CSS related to this feature isn't removed from the AMP page version.

It will cause an error during the AMP validation process

**To Reproduce**
1. Install AMP plugin
2. In settings select « Transitional » template mode. The AMP URL should use a query string like mashingcoding.com/2012/01/07/template-sticky/?amp
3. Activate all Lazyload options into WP Rocket
4. See the inline CSS in the source code

```
<style id='rocket-lazyload-inline-css' type='text/css'>
.rll-youtube-player{position:relative;padding-bottom:56.23%;height:0;overflow:hidden;max-width:100%;}.rll-youtube-player iframe{position:absolute;top:0;left:0;width:100%;height:100%;z-index:100;background:0 0}.rll-youtube-player img{bottom:0;display:block;left:0;margin:auto;max-width:100%;width:100%;position:absolute;right:0;top:0;border:none;height:auto;cursor:pointer;-webkit-transition:.4s all;-moz-transition:.4s all;transition:.4s all}.rll-youtube-player img:hover{-webkit-filter:brightness(75%)}.rll-youtube-player .play{height:72px;width:72px;left:50%;top:50%;margin-left:-36px;margin-top:-36px;position:absolute;background:url(http://smashingcoding.com/wp-content/plugins/wp-rocket/assets/img/youtube.png) no-repeat;cursor:pointer}.wp-has-aspect-ratio .rll-youtube-player{position:absolute;padding-bottom:0;width:100%;height:100%;top:0;bottom:0;left:0;right:0}
</style>
```
And the `<noscript>` tag which trigger the error during the validation process:
`<noscript><style id="rocket-lazyload-nojs-css">.rll-youtube-player, [data-lazy-src]{display:none !important;}</style></noscript>`

**Expected behavior**
No inline CSS related to LazyLoad on AMP pages when Transitional mode is selected.

**Screenshots**
https://d33v4339jhl8k0.cloudfront.net/inline/18928/f09382920cc782ff56d82f2c30d831ea76bae97c/76e2fbb098efe191837410b679ff62852b011295/image-1.png

**Additional context**
After deactivated the LazyLoad iframe, the `<noscript>` tag which contains a small inline CSS is still present.

This code is removed after turning off all LazyLoad options.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

Tabrisrp:
Fixes #2523

The issue is happening because lazyload for images is disabled on AMP pages, but not lazyload for iframes. So the inline CSS fallback is still inserted on the pages.

- [x] Use filter to disable lazyload for iframes on AMP pages
- [x] Update tests

### DIFF
--- a/inc/ThirdParty/Plugins/Optimization/AMP.php
+++ b/inc/ThirdParty/Plugins/Optimization/AMP.php
@@ -104,6 +104,7 @@ class AMP implements Subscriber_Interface {
 
 		remove_filter( 'wp_resource_hints', 'rocket_dns_prefetch', 10, 2 );
 		add_filter( 'do_rocket_lazyload', '__return_false' );
+		add_filter( 'do_rocket_lazyload_iframes', '__return_false' );
 		unset( $wp_filter['rocket_buffer'] );
 
 		$options = get_option( self::AMP_OPTIONS, [] );
--- a/tests/Integration/inc/ThirdParty/Plugins/Optimization/AMP/disableOptionsOnAmp.php
+++ b/tests/Integration/inc/ThirdParty/Plugins/Optimization/AMP/disableOptionsOnAmp.php
@@ -28,13 +28,14 @@ class Test_DisableOptionsOnAmp extends TestCase {
 	 */
 	public function testShouldDoExpected( $config, $expected ) {
 		if ( $expected[ 'bailout' ] ) {
-			$this->assertNotFalse( has_filter( 'wp_resource_hints', 'rocket_dns_prefetch') );
+			$this->assertNotFalse( has_filter( 'wp_resource_hints', 'rocket_dns_prefetch' ) );
 		} else {
 			global $wp_filter;
 
 			add_theme_support( AMP_Theme_Support::SLUG );
-			$this->assertNotFalse( has_filter( 'wp_resource_hints', 'rocket_dns_prefetch') );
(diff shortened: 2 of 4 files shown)


hellofromtonya:
@Tabrisrp Has this fix been applied to any of the customer websites with open support tickets for this problem? 

If no, can you work with Support to add this fix and validate that it indeed does fix the problem.

Tabrisrp:
@hellofromtonya Shouldn't we test with our QA first?

hellofromtonya:
@Tabrisrp We want to know if this is the right fix. Does it solve the problem for those customer who reported it? If we can check it there, awesome. If no, then yes, we can go through our internal QA and then verify after releasing if it resolves their issue.

What do you think?

piotrbak:
@Tabrisrp It worked as expected on the customer's website, the CSS is not present anymore.

hellofromtonya:
@Tabrisrp As this one has been validated on a customer's site, it should be good to merge.

PR #2764, opened 2020-06-12
Fixes #1315 Google Tag Manager detection fails when URL uses relative protocol

### LINKED ISSUE #1315: Google Tag Manager detection fails when URL uses relative protocol
The regex used in `class-google-tag-manager.php` doesn't detect script tags like this:

```
<script async src="...">
```

Unfortunately, that's the code I was given, and I assume it's the default code that GTM generates for end-users.
I presume this `async` issue is also true for other scripts that should be locally-copied.

Another issue I ran into is that when the script is loaded with a protocol-relative URL, the regex doesn't catch it either:

```
<script src="//www.googletagmanager.com/...">
```

Changing the URL to the standard format fixed it:

```
<script src="https://...">
```

However, I feel that this is difficult for non-technical users to figure out, and it's therefore a bug.

engahmeds3ed:
fixes #1315 
When google tracking addon is enabled, the google tag manager script url can be served locally even when using the relative protocol like that:
`<script src="//www.googletagmanager.com/...">`
By converting it to the full protocol
`<script src="https//www.googletagmanager.com/...">`


inc/classes/busting/class-google-analytics.php
```diff
@@ -40,7 +40,7 @@ class Google_Analytics extends Abstract_Busting {
 	 * @access protected
 	 * @author Grégory Viguier
 	 */
-	protected $filename = 'ga-%s.js';
+	protected $file_name_pattern = 'ga-%s.js';
```
hellofromtonya, line 43: As we touched this file, let's move it into the module and rename.

inc/classes/busting/trait-file-busting.php
```diff
@@ -85,7 +85,7 @@ public function delete() {
 	 * @return string|bool Version of the file. False if the file does not exist.
 	 */
 	protected function get_busting_version() {
-		if ( isset( $this->file_version ) ) {
+		if ( isset( $this->file_version ) && empty( $this->file_version ) ) {
```
hellofromtonya, line 88: Same here. We touched this file. Time to move it into the module and rename.

hellofromtonya, line 88: `empty()` is enough here as it's a combo of `isset` and not falsey.

Are we want to check that it's empty or that it's not empty?

inc/classes/busting/trait-file-busting.php
```diff
@@ -130,35 +129,19 @@ private function get_all_files() {
...
-		$pattern = \sprintf( $pattern, '(?<version>(?:[a-f0-9]{32}|local))' );
+		$entries = rocket_get_dir_files_by_regex( $dir_path, $pattern );
 
-		foreach ( $dir as $entry ) {
-			if ( 'f' !== $entry['type'] ) {
-				continue;
-			}
+		$list = [];
+		foreach ( $entries as $entry ) {
+			$file_name = $entry->getFilename();
```
hellofromtonya, line 141: For commonality, let's standardize on one word for `filename`. That fits well with PHP internals too.

inc/classes/busting/trait-file-busting.php
```diff
@@ -130,35 +129,19 @@ private function get_all_files() {
...
 
-			if ( \preg_match( '/^' . $pattern . '$/', $entry['name'], $matches ) ) {
-				$list[ $entry['name'] ] = $matches['version'];
-			}
+			preg_match( $pattern, $file_name, $file_details_match );
+			$list[ $file_name ] = $file_details_match[1];
```
hellofromtonya, line 144: What if there are no matches?

engahmeds3ed: it isn't the case here, at the following line
https://github.com/wp-media/wp-rocket/blob/0833d81e74efa7ea30a255f5b4adb878221779b0/inc/classes/busting/trait-file-busting.php#L137
we are searching for files with this pattern and then we are using preg_match with the same pattern to get the file version from the file name
For example:-
if we have a file named: `gtm-123456789abc.js` so the version is `123456789abc` and we got that from applying the pattern `gtm-([a-f0-9]{32}|local).js` on the filename.
So when reaching this line we must have one match.

inc/classes/busting/trait-file-busting.php
engahmeds3ed: You are totally right, I wanted to check if this file_version is not empty.
Solved now

inc/classes/busting/class-google-analytics.php
engahmeds3ed: Done.

inc/classes/busting/trait-file-busting.php
engahmeds3ed: Done.

inc/classes/busting/trait-file-busting.php
hellofromtonya: What you're saying then is the iterator will only return filesystem entries that have the file version.

Can we ever imagine a scenario where a match can't be made, where it fails for some reason?

I'm asking this because if that scenario, even a remote weird edge case, would cause a fatal error. If we could imagine an edge case, then we should guard against it to prevent the failure.

inc/Addon/Busting/FileBustingTrait.php
```diff
@@ -226,7 +207,7 @@ private function get_busting_file_name( $version ) {
 			return false;
 		}
 
-		return \sprintf( $this->filename, $version );
+		return sprintf( $this->file_name_pattern, $version );
```
Tabrisrp, line 210:
```suggestion
		return sprintf( $this->filename_pattern, $version );
```

inc/Addon/Busting/FileBustingTrait.php
```diff
@@ -253,16 +234,12 @@ private function get_busting_file_path( $version ) {
 	 *
 	 * @since  3.2.4
 	 * @access private
-	 * @author Grégory Viguier
 	 *
-	 * @param  string $file_name The file name.
+	 * @param  string $file_name_pattern The file name.
```
Tabrisrp, line 238:
```suggestion
	 * @param  string $filename_pattern The file name.
```

inc/Addon/Busting/FileBustingTrait.php
```diff
@@ -253,16 +234,12 @@ private function get_busting_file_path( $version ) {
...
 	 * @return string
 	 */
-	private function escape_file_name( $file_name ) {
-		$file_name = \explode( '%s', $file_name );
-		$file_name = \array_map( 'preg_quote', $file_name );
-
-		return \implode( '%s', $file_name );
+	private function escape_file_name( $file_name_pattern ) {
```
Tabrisrp, line 241:
```suggestion
	private function escape_file_name( $filename_pattern ) {
```

inc/Addon/Busting/FileBustingTrait.php
```diff
@@ -253,16 +234,12 @@ private function get_busting_file_path( $version ) {
...
+		return preg_quote( $file_name_pattern, '/' );
```
Tabrisrp, line 242:
```suggestion
		return preg_quote( $filename_pattern, '/' );
```

inc/Addon/GoogleTracking/GoogleAnalytics.php
```diff
@@ -40,7 +41,7 @@ class Google_Analytics extends Abstract_Busting {
 	 * @access protected
 	 * @author Grégory Viguier
 	 */
-	protected $filename = 'ga-%s.js';
+	protected $file_name_pattern = 'ga-%s.js';
```
Tabrisrp, line 44:
```suggestion
	protected $filename_pattern = 'ga-%s.js';
```

inc/Addon/GoogleTracking/GoogleTagManager.php
```diff
@@ -29,17 +29,15 @@ class Google_Tag_Manager extends Abstract_Busting {
 	 * @var    string
 	 * @since  3.2.4
 	 * @access protected
-	 * @author Grégory Viguier
 	 */
-	protected $filename = 'gtm-%s.js';
+	protected $file_name_pattern = 'gtm-%s.js';
```
Tabrisrp, line 33:
```suggestion
	protected $filename_pattern = 'gtm-%s.js';
```

inc/Addon/GoogleTracking/GoogleTagManager.php
```diff
@@ -68,18 +64,21 @@ class Google_Tag_Manager extends Abstract_Busting {
...
+	public function __construct( $busting_path, $busting_url, GoogleAnalytics $ga_busting, $filesystem = null ) {
 		$blog_id            = get_current_blog_id();
 		$this->busting_path = $busting_path . $blog_id . '/';
 		$this->busting_url  = $busting_url . $blog_id . '/';
 		$this->ga_busting   = $ga_busting;
-		$this->filesystem   = \rocket_direct_filesystem();
+		if ( is_null( $filesystem ) ) {
+			$filesystem = rocket_direct_filesystem();
+		}
+		$this->filesystem = $filesystem;
```
Tabrisrp, line 81:
```suggestion
		if ( is_null( $filesystem ) ) {
			$this->filesystem = rocket_direct_filesystem();
		} else {
		    $this->filesystem = $filesystem;
		}
```

remyperona (changes requested):

inc/classes/busting/trait-file-busting.php
engahmeds3ed: Ok, I did the check now.

inc/Addon/GoogleTracking/GoogleTagManager.php
engahmeds3ed: I think this from performance wise is the same, why do we need to adjust that?

Tabrisrp: How about:

`$this->filesystem = is_null( $filesystem ) ? rocket_direct_filesystem() : $filesystem;`

Makes the code more readable

inc/Addon/Busting/FileBustingTrait.php
```diff
@@ -455,7 +434,7 @@ private function get_file_contents( $file_path ) {
 	 */
 	private function get_remote_contents( $url ) {
 		try {
-			$response = \wp_remote_get( $url );
+			$response = wp_remote_get( $url );
 		} catch ( \Exception $e ) {
```
crystinutzaa, line 438:
```suggestion
		} catch ( Exception $e ) {
```

tests/Fixtures/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
```diff
@@ -0,0 +1,37 @@
...
+					'integration' => '<html><head><script async src="https://www.test.com/rocket.js"></script></head></html>',
+				]
+		],
+		'shouldReplaceUrlWithRelativeProtocol' => [
+			'config' => [
+				'url' => 'https://www.googletagmanager.com/gtag/js?id=UA-135617916-1',
+				'html' => '<html><head><script async src="//www.googletagmanager.com/gtag/js?id=UA-135617916-1"></script></head></html>',
+			],
+			'expected' => [
+				'unit' => '<html><head><script data-no-minify="1" async src="vfs://public/wp-content/cache/busting/1/gtm-88c587e9d2fdeb7ac5d4cdd9bd8d4af5.js"></script></head></html>',
```
crystinutzaa, line 32: Why do we need in expected both unit & integration? 
We can use a placeholder for the domain and / or wp-content  folderand replace it in test with the proper domain based on each test.
What do you think? Will this work?

tests/Integration/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
```diff
@@ -0,0 +1,37 @@
...
+			Functions\expect( 'wp_remote_get' )->once()->with( $url )->andReturn( 'request' );
+			Functions\expect( 'wp_remote_retrieve_body' )->once()->with( 'request' )->andReturn( 'Remote File contents here' );
+		}
+
+		$busting_path       = WP_ROCKET_CACHE_ROOT_PATH . 'busting/';
+		$busting_url        = WP_ROCKET_CACHE_ROOT_URL . 'busting/';
+		$google_tag_manager = new GoogleTagManager( $busting_path, $busting_url, new GoogleAnalytics( $busting_path, $busting_url ), $this->filesystem );
+		$actual             = $google_tag_manager->replace_url( $html );
+
+		$this->assertEquals($expected['integration'], $actual);
```
crystinutzaa, line 34:
```suggestion
		$this->assertEquals( $expected['integration'], $actual );
```

tests/Integration/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
```diff
@@ -0,0 +1,37 @@
...
+ */
+class Test_ReplaceUrl extends FilesystemTestCase {
+	protected $path_to_test_data = '/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php';
+
+	/**
+	 * @dataProvider providerTestData
+	 */
+	public function testShouldReplaceUrl( $config, $expected ) {
+		$html = isset( $config['html'] ) ? $config['html'] : '';
+		$url = isset( $config['url'] ) ? $config['url'] : '';
```
crystinutzaa, line 22:
```suggestion
		$url  = isset( $config['url'] ) ? $config['url'] : '';
```

tests/Integration/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
```diff
@@ -0,0 +1,37 @@
...
+
+		if ( !empty( $url ) ) {
```
crystinutzaa, line 24:
```suggestion
		if ( ! empty( $url ) ) {
```

tests/Unit/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+class Test_ReplaceUrl extends FilesystemTestCase {
+	protected $path_to_test_data = '/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php';
+
+	/**
+	 * @dataProvider providerTestData
+	 */
+	public function testShouldReplaceUrl( $config, $expected ) {
+		$html = isset( $config['html'] ) ? $config['html'] : '';
+		$url = isset( $config['url'] ) ? $config['url'] : '';
```
crystinutzaa, line 22:
```suggestion
		$url  = isset( $config['url'] ) ? $config['url'] : '';
```

tests/Unit/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
```diff
@@ -0,0 +1,40 @@
...
+		if ( !empty( $url ) ) {
+			Functions\expect( 'wp_remote_get' )->once()->with( $url )->andReturn( 'request' );
+			Functions\expect( 'wp_remote_retrieve_body' )->once()->with( 'request' )->andReturn( 'Remote File contents here' );
+		}
+
+
+		$busting_path       = WP_ROCKET_CACHE_ROOT_PATH . 'busting/';
+		$busting_url        = WP_ROCKET_CACHE_ROOT_URL . 'busting/';
+		$google_tag_manager = new GoogleTagManager( $busting_path, $busting_url, new GoogleAnalytics( $busting_path, $busting_url ), $this->filesystem );
+		$actual = $google_tag_manager->replace_url( $html );
```
crystinutzaa, line 35: Please align `=`  for code readability.

tests/Unit/inc/Addon/GoogleTracking/Subscriber/deleteTrackingCache.php
```diff
@@ -3,10 +3,10 @@
 namespace WP_Rocket\Tests\Unit\inc\Addon\GoogleTracking\Subscriber;
 
 use \Mockery;
```
crystinutzaa, line 5:
```suggestion
use Mockery;
```

inc/Addon/GoogleTracking/GoogleTagManager.php
engahmeds3ed: Ok, Nice from readability perspective

tests/Fixtures/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
engahmeds3ed: I thought that this will be easier for non developers to add fixture with full values, but your idea is totally right.

inc/functions/files.php
```diff
@@ -1317,6 +1317,32 @@ function rocket_find_wpconfig_path() {
...
+		return [];
+	}
+
+	try {
+		return new RegexIterator( $iterator, $regex );
+	} catch ( Exception $e ) {
+		return [];
+	}
+
+}
```
hellofromtonya, line 1344: 2 things:

1. Let's make this private, i.e. to be used only internally within Rocket. This will avoid us having to deprecate it later when it becomes a method in the filesystem module.
2. Why 2 try/catch? We could simplify by combining.

```suggestion
function _rocket_get_dir_files_by_regex( $dir, $regex ) {
	try {
		$iterator = new IteratorIterator(
			new FilesystemIterator( $dir )
		);

		return new RegexIterator( $iterator, $regex );
	} catch ( Exception $e ) {
		return [];
	}
}
```

inc/classes/busting/class-busting-factory.php
```diff
@@ -1,6 +1,9 @@
 <?php
```
hellofromtonya, line 1: We touched this file. Let's move this into new architecture.

hellofromtonya (changes requested):

inc/Addon/Busting/FileBustingTrait.php
```diff
@@ -130,34 +128,20 @@ private function get_all_files() {
...
-			return false;
-		}
-
-		if ( ! $dir ) {
-			return [];
-		}
+		$pattern = '/' . sprintf(
+			$this->escape_file_name( $this->filename_pattern ),
+			'([a-f0-9]{32}|local)'
+		) . '/';
```
hellofromtonya, line 134: Question: When does the `$this->filename_pattern` change?

Why?

This could be an expensive process. I'm wondering if the pattern should run only when `$this->filename_pattern is set or changed.

What do you think?

engahmeds3ed: this is a trait so the class that use this trait has an attribute `filename_pattern` so for example 

https://github.com/wp-media/wp-rocket/blob/62462b4e8e34623444454a39ff588cd973a7a363/inc/Addon/GoogleTracking/GoogleAnalytics.php#L44

https://github.com/wp-media/wp-rocket/blob/3be25639a3c18b3058f29c7bda9c9ffbdb3c0f9e/inc/Addon/GoogleTracking/GoogleTagManager.php#L33

so it's constant for each class of them.
and keep in mind that this operation happens only one time due to this condition

https://github.com/wp-media/wp-rocket/blob/dd4c4052b6db0d91f9c07a6bfbe968203de6eb37/inc/Addon/Busting/FileBustingTrait.php#L87-L89

so what do you think?

tests/Fixtures/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
hellofromtonya: @engahmeds3ed If there are no differences, keep it simple and use the same expected string for both test suites.

hellofromtonya (changes requested):

tests/Fixtures/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
engahmeds3ed: Ok

engahmeds3ed: @crystinutzaa now it's resolved, plz review and approve changes.

crystinutzaa (approved):

crystinutzaa:
@arunbasillal & @Davidacu this is ready for QA

Davidacu (changes requested):
[Case 1](https://www.notion.so/wpmedia/Should-load-Google-analytics-tag-when-url-is-absolute-76d90dd627fb4e4889a0c786c18b72d9)  -  loading Google analytics tag - Variant 1 - absolute path - **Passed**
[Case 2 ](https://www.notion.so/wpmedia/Should-load-Google-analytics-tag-when-url-is-relative-1c94307666924d01ac647311d96abc58) -  loading Google analytics tag - Variant 2 - relative path - **Passed**
[Case 3](https://www.notion.so/wpmedia/Should-load-Google-analytics-tag-when-using-async-70af1b8f392e4020a2ccef8f47f43baf) -  loading Google analytics tag - Variant 3 (Async) - **Failed** (URL is not replaced) http://sc.com.ly/show/CjPB2wyHqSi5nYkT4OrVFp701GZuasKc
[Case 4 ](https://www.notion.so/wpmedia/Should-load-Global-site-tag-when-url-is-absolute-e65bff27de314a5ab1f7793525ee72cc)- loading Global site tag - Variant 1 - absolute path - **Passed**
[Case 5](https://www.notion.so/wpmedia/Should-load-Global-site-tag-when-url-is-relative-3486ac23e8e4496a8f66c80ac1e4dd9f) - loading Global site tag - Variant 2 - relative path - **Passed**

@engahmeds3ed @Tabrisrp 
Test Case 3 failed, when the google analytics tag is loaded async as recommended [here](https://developers.google.com/analytics/devguides/collection/analyticsjs/)
we do not replace the URL with our cached version, this is not a regression, not sure if this case should be handled in scope of this PR, I though on bringing this up since the original reporter of #1315  mentioned an issue when using `async`

Tabrisrp:
@Davidacu You are right, when I tested during the grooming, I tested with `www.googletagmanager.com`, but didn't test with `www.google-analytics.com/analytics.js`, and this one is indeed not replaced.

I think the async way recommended didn't exist yet when we developed this feature.

This can be included in this PR though, the changes are not big to make it work, I did some tests on local and ended with those changes: https://quickdiff.net/?unique_id=24AA1A7B-C0EB-50BE-CE39-142434113B26

@engahmeds3ed Can you take care of the above changes to make this one case pass?

engahmeds3ed:
@Tabrisrp @Davidacu Checking now

engahmeds3ed:
@Tabrisrp I adjusted the remaining issue, could u please review now.

inc/Addon/GoogleTracking/GoogleAnalytics.php
```diff
@@ -215,15 +216,19 @@ public function get_url() {
 	 * @return string
 	 */
 	protected function find( $pattern, $html ) {
-		\preg_match_all( '/' . $pattern . '/Umsi', $html, $matches, PREG_SET_ORDER );
+		\preg_match_all( '/' . $pattern . '/Umi', $html, $matches, PREG_SET_ORDER );
```
Tabrisrp, line 219:
```suggestion
		preg_match_all( '/' . $pattern . '/Umi', $html, $matches, PREG_SET_ORDER );
```

inc/Addon/GoogleTracking/GoogleAnalytics.php
```diff
@@ -215,15 +216,19 @@ public function get_url() {
...
 
 		if ( ! $matches ) {
 			return false;
 		}
 
 		$matches = \array_map(
```
Tabrisrp, line 225:
```suggestion
		$matches = array_map(
```

inc/Addon/GoogleTracking/GoogleAnalytics.php
```diff
@@ -215,15 +216,19 @@ public function get_url() {
...
 			function( $match ) {
-				if ( false === \strpos( $match[1], 'GoogleAnalyticsObject' ) ) {
+				if ( isset( $match['attr'] ) && ! preg_match( '/src\s*=\s*[\'"]\s*(?:https?:)?\/\/www\.google-analytics\.com\/analytics\.js\s*[\'"]/i', $match['attr'] ) ) {
+					return;
+				}
+
+				if ( isset( $match['content'] ) && false === \strpos( $match['content'], 'GoogleAnalyticsObject' ) ) {
```
Tabrisrp, line 231:
```suggestion
				if ( isset( $match['content'] ) && false === strpos( $match['content'], 'GoogleAnalyticsObject' ) ) {
```

inc/functions/files.php
```diff
@@ -1195,6 +1195,30 @@ function rocket_get_filesystem_perms( $type ) {
 	return $perms[ $type ];
 }
 
+/**
+ * Gets Directory files matches regex.
+ *
+ * @since 3.6.1
```
Tabrisrp, line 1201:
```suggestion
 * @since 3.6.3
```

tests/Integration/inc/Addon/GoogleTracking/GoogleAnalytics/replaceUrl.php
```diff
@@ -0,0 +1,39 @@
...
+ * @covers \WP_Rocket\Addon\GoogleTracking\GoogleAnalytics::replace_url
+ * @group  Addon
+ * @group  GoogleTracking
+ */
+class Test_ReplaceUrl extends FilesystemTestCase {
+	protected $path_to_test_data = '/inc/Addon/GoogleTracking/GoogleAnalytics/replaceUrl.php';
+
+	public function tearDown() {
+		parent::tearDown();
+		remove_filter('pre_get_rocket_option_google_analytics_cache', '__return_true');
```
Tabrisrp, line 17:
```suggestion
		remove_filter( 'pre_get_rocket_option_google_analytics_cache', [ $this, 'return_true' ] );
```

That way, if `__return_true` is ever used on this filter somewhere else, we do not remove it completely during testing.

tests/Integration/inc/Addon/GoogleTracking/GoogleAnalytics/replaceUrl.php
```diff
@@ -0,0 +1,39 @@
...
+	public function testShouldReplaceUrl( $config, $expected ) {
+		$html = isset( $config['html'] ) ? $config['html'] : '';
+		$url  = isset( $config['url'] ) ? $config['url'] : '';
+
+		if ( ! empty( $url ) ) {
+			Functions\expect( 'wp_remote_get' )->once()->with( $url )->andReturn( 'request' );
+			Functions\expect( 'wp_remote_retrieve_body' )->once()->with( 'request' )->andReturn( 'Remote File contents here' );
+		}
+
+		add_filter('pre_get_rocket_option_google_analytics_cache', '__return_true');
```
Tabrisrp, line 32:
```suggestion
		add_filter( 'pre_get_rocket_option_google_analytics_cache', [ $this, 'return_true' ] );
```

tests/Integration/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
```diff
@@ -0,0 +1,41 @@
+<?php
+namespace WP_Rocket\Tests\Integration\inc\Addon\GoogleTracking\GoogleTagManager;
+
+use WP_Rocket\Addon\GoogleTracking\GoogleAnalytics;
+use WP_Rocket\Addon\GoogleTracking\GoogleTagManager;
```
Tabrisrp, line 5:
```suggestion
```

tests/Integration/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
```diff
@@ -0,0 +1,41 @@
...
+ * @covers \WP_Rocket\Addon\GoogleTracking\GoogleTagManager::replace_url
+ * @group  Addon
+ * @group  GoogleTracking
+ */
+class Test_ReplaceUrl extends FilesystemTestCase {
+	protected $path_to_test_data = '/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php';
+
+	public function tearDown() {
+		parent::tearDown();
+		remove_filter('pre_get_rocket_option_google_analytics_cache', '__return_true');
```
Tabrisrp, line 19:
```suggestion
		remove_filter( 'pre_get_rocket_option_google_analytics_cache', [ $this, 'return_true' ] );
```

tests/Integration/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
```diff
@@ -0,0 +1,41 @@
...
+	public function testShouldReplaceUrl( $config, $expected ) {
+		$html = isset( $config['html'] ) ? $config['html'] : '';
+		$url  = isset( $config['url'] ) ? $config['url'] : '';
+
+		if ( ! empty( $url ) ) {
+			Functions\expect( 'wp_remote_get' )->once()->with( $url )->andReturn( 'request' );
+			Functions\expect( 'wp_remote_retrieve_body' )->once()->with( 'request' )->andReturn( 'Remote File contents here' );
+		}
+
+		add_filter('pre_get_rocket_option_google_analytics_cache', '__return_true');
```
Tabrisrp, line 34:
```suggestion
		add_filter( 'pre_get_rocket_option_google_analytics_cache', [ $this, 'return_true'] );
```

tests/Unit/inc/Addon/GoogleTracking/GoogleAnalytics/replaceUrl.php
```diff
@@ -0,0 +1,39 @@
...
+	/**
+	 * @dataProvider providerTestData
+	 */
+	public function testShouldReplaceUrl( $config, $expected ) {
+		$html = isset( $config['html'] ) ? $config['html'] : '';
+		$url  = isset( $config['url'] ) ? $config['url']   : '';
+
+		Functions\when( 'get_current_blog_id' )->justReturn( 1 );
+
+		if ( !empty( $url ) ) {
```
Tabrisrp, line 25:
```suggestion
		if ( ! empty( $url ) ) {
```

tests/Unit/inc/Addon/GoogleTracking/GoogleTagManager/replaceUrl.php
```diff
@@ -0,0 +1,40 @@
...
+	/**
+	 * @dataProvider providerTestData
+	 */
+	public function testShouldReplaceUrl( $config, $expected ) {
+		$html = isset( $config['html'] ) ? $config['html'] : '';
+		$url  = isset( $config['url'] ) ? $config['url']   : '';
+
+		Functions\when( 'get_current_blog_id' )->justReturn( 1 );
+
+		if ( !empty( $url ) ) {
```
Tabrisrp, line 26:
```suggestion
		if ( ! empty( $url ) ) {
```

remyperona (changes requested):
Mainly formatting and small stuff to fix, after that it's good for another QA

engahmeds3ed:
@Davidacu @wp-media/qa 
Ready for review again, thanks.

remyperona (approved):

Davidacu (changes requested):
[Case 1](https://www.notion.so/wpmedia/Should-load-Google-analytics-tag-when-url-is-absolute-76d90dd627fb4e4889a0c786c18b72d9) - loading Google analytics tag - Variant 1 - absolute path - **Failed**, this time this one is failing, the replacement is not happening.
https://i.imgur.com/34onMT1.png
[Case 2](https://www.notion.so/wpmedia/Should-load-Google-analytics-tag-when-url-is-relative-1c94307666924d01ac647311d96abc58) - loading Google analytics tag - Variant 2 - relative path - **Failed**
replacement is not happening https://i.imgur.com/jJQ1hV1.png
[Case 3](https://www.notion.so/wpmedia/Should-load-Google-analytics-tag-when-using-async-70af1b8f392e4020a2ccef8f47f43baf) - loading Google analytics tag - Variant 3 (Async) - **Passed** 
https://i.imgur.com/bmS5NyW.png
[Case 4](https://www.notion.so/wpmedia/Should-load-Global-site-tag-when-url-is-absolute-e65bff27de314a5ab1f7793525ee72cc) - loading Global site tag - Variant 1 - absolute path - **Passed**
https://i.imgur.com/BidUnzW.png
[Case 5](https://www.notion.so/wpmedia/Should-load-Global-site-tag-when-url-is-relative-3486ac23e8e4496a8f66c80ac1e4dd9f) - loading Global site tag - Variant 2 - relative path - **Passed**
https://i.imgur.com/BidUnzW.png

3/5 test cases passed
2/5 test cases failed

It looks like now the cases that are failing are the ones for analytics.js , the non-async versions.
Please have a look. @engahmeds3ed @Tabrisrp

engahmeds3ed:
@Davidacu Last time round plz, many thanks for your patience

Davidacu (changes requested):
[Case 1](https://www.notion.so/wpmedia/Should-load-Google-analytics-tag-when-url-is-absolute-76d90dd627fb4e4889a0c786c18b72d9) - loading Google analytics tag - Variant 1 - absolute path - **Passed**
[Case 2](https://www.notion.so/wpmedia/Should-load-Google-analytics-tag-when-url-is-relative-1c94307666924d01ac647311d96abc58) - loading Google analytics tag - Variant 2 - relative path - **Passed**
[Case 3](https://www.notion.so/wpmedia/Should-load-Google-analytics-tag-when-using-async-70af1b8f392e4020a2ccef8f47f43baf) - loading Google analytics tag - Variant 3 (Async) - **Failed**
https://i.imgur.com/HeGXAcl.png
[Case 4 ](https://www.notion.so/wpmedia/Should-load-Global-site-tag-when-url-is-absolute-e65bff27de314a5ab1f7793525ee72cc)- loading Global site tag - Variant 1 - absolute path - **Passed**
[Case 5](https://www.notion.so/wpmedia/Should-load-Global-site-tag-when-url-is-relative-3486ac23e8e4496a8f66c80ac1e4dd9f) - loading Global site tag - Variant 2 - relative path - **Passed**

4/5 test cases passed
1/5 test case failed.

Looks like now the `async` version of analytics.js is not being replaced.

Davidacu (approved):
All cases passed QA in this round!! 🎉 
Thanks @engahmeds3ed !

Tabrisrp:
Good job!

@engahmeds3ed You can merge the PR & close the related issue 👍

PR #2766, opened 2020-06-12
Fixes #2502 WP_CACHE improvements

### LINKED ISSUE #2502: Don't add WP_CACHE in wp-config.php if the constant is already defined somewhere else
Currently, we don't have a way to skip adding the `WP_CACHE` in the `wp-config.php` file.

This could be useful if that file is under version control.

**Related ticket:** https://secure.helpscout.net/conversation/1124360039/154873?folderId=2135277
**Slack discussion:** https://wp-media.slack.com/archives/C43T1AYMQ/p1585745468055100

Tabrisrp:
There is a lot going on in this PR, so I'm going to summarize the changes:

- The functions `rocket_find_wpconfig_path()`, `set_rocket_wp_cache_define()`, `rocket_warning_wp_config_permissions()` and `rocket_maybe_set_wp_cache_define()` have been deprecated
- a new class `WPCache` is added, with methods corresponding to the above functions. The functions calls have been replaced by either a direct method call (when hooks can't be used), or by callbacks on hooks using the `AdminSubscriber` subscriber
- To fix issue #2502, the code in `WPCache::set_wp_cache_constant()` is updated to bailout if the constant is not found in `wp-config.php`, but is set somehwere else
- a new method `add_wp_cache_status_test()` is added, which is adding a test for the `WP_CACHE` constant value on the **Site Health** tools page
- Tests have been either updated or added for each method


hellofromtonya:

Tabrisrp:
@crystinutzaa @hellofromtonya @arunbasillal Re-requesting review after finding out yesterday that `WP_CACHE` is always set to `false` by WordPress if it has not been done before `wp-settings.php` is loaded.

Changes since previous review:
- Remove the conditional in `set_wp_cache_constant()` that was checking for the constant existence, since it's useless
- Add a new filter `rocket_set_wp_cache_constant` to allow advanced users to prevent WP Rocket from writing in the `wp-config.php`

For QA:
- WP Rocket should always write in `wp-config.php`, except if the filter value is set to `false`

Tabrisrp:
> Please note that I am fine with either approach. Customers who use non-standard WordPress installs can use the filter and are used to doing such things. But my questions is, if that is the case, should we do the other checks (i.e if WP_CACHE is true and such before writing into wp-config). Why not keep things simple and just add the filter and be done with things?

Yes in the end, the difference in term of functionality is the addition of this filter, to comply with the user request, and the addition of the Site Health Check.

The fact that WordPress defines `WP_CACHE` to `false` on its own if it has not been set before loading `wp-settings.php` prevents us from doing more, since during normal WP execution this constant will always be set, and we can't know _when_. Was it before `wp-config.php` by the user, or after by WP?

We can check for the value itself when it's for `true`, because we know it's not WordPress which did that. But when it's `false`, we don't.

But that PR also brings a lot of our code to the new architecture at the same time, and adds more code coverage, so it's a win-win anyway.

arunbasillal:
@hellofromtonya @Tabrisrp @GeekPress 

> Yes in the end, the difference in term of functionality is the addition of this filter, to comply with the user request, and the addition of the Site Health Check.

In that case, shall we remove the additional checks when WP_CACHE is true and keep things simple? Any customer who wants this can use the filter and control WP Rocket action. Unless the filter is used, we will simply assume it's default WordPress and proceed. 

What do you think?

Tabrisrp:
I don't see the benefit of removing them? The code is not complex currently, and doing them when we can is a better experience overall than no checks at all imo.

hellofromtonya:
>In that case, shall we remove the additional checks when WP_CACHE is true and keep things simple?

@arunbasillal What additional checks are you proposing we remove?

arunbasillal:
@hellofromtonya @Tabrisrp The idea to keep things simple was from a product and support perspective. 

Right now, if WP_CACHE is defined as true elsewhere, we don't add the constant to wp-config. But we do it if WP_CACHE is false. One additional thing to consider for support when WP_CACHE is not present in wp-config. 

On the other hand, if we simply follow the filter, all we have to do is check for the status of the filter. 

> I don't see the benefit of removing them? The code is not complex currently, and doing them when we can is a better experience overall than no checks at all imo.

I am guessing here, but there could be some cases where customers do not have to set the filter at all and we do not try to edit wp-config. That could be useful. 

Either way, this is rare. Since all the work is done, let's keep it how it is. I will proceed with the QA

Tabrisrp:
> If wp-config.php is writeable, WP Rocket writes define( 'WP_CACHE', false ); // Added by WP Rocket into it. 

This is unavoidable without the use of the filter, and that's why the filter was introduced.

My last commit prevents displaying the notice when the filter value is set to `false`, to fix what was reported in the last QA.

Tabrisrp:
@wp-media/qa This PR needs another round of QA

Tabrisrp:
Last commit fixes the deactivation notice. It's a different one specific for deactivation, I missed it.

@Davidacu If you can do the QA today for this specific part 👌

On deactivation, if `add_filter( 'rocket_set_wp_cache_constant', '__return_false' );` is added and wp-config.php is write protected, there should be no notice.

arunbasillal:
@Tabrisrp @Davidacu I still have the test environments ready. I will test the latest update and confirm in a few minutes 👍

PR #2770, opened 2020-06-12
Closes #2607 - Improve sub-feeds exclusions

### LINKED ISSUE #2607: Improve the sub-feeds exclusions
**Describe the bug**
WP Rocket automatically exclude the https://example.org/feed/ from being cached:
https://github.com/wp-media/wp-rocket/blob/00cf7a1c5e0b2e7354b0aa76ca86603ee4420eae/inc/functions/options.php#L234-L235

but some Podcast plugin like [https://wordpress.org/plugins/podlove-podcasting-plugin-for-wordpress/](https://wordpress.org/plugins/podlove-podcasting-plugin-for-wordpress/) are using this `/feed/` base to add URLs like `/feed/mp3/` which won’t be automatically excluded from being cached. WP Rocket will then change `MIME-types` from `content-type: application/rss+xml;` to `content-type: txt/html;`.

This is because the current pattern we use will only match for URLs ending with `/feed/` or `/feed`, if there is anything after, it won’t work.

**To Reproduce**
1. Install and configure this plugin [https://wordpress.org/plugins/podlove-podcasting-plugin-for-wordpress/](https://wordpress.org/plugins/podlove-podcasting-plugin-for-wordpress/)
2. Check the URL it will create
3. Test it in this tool [https://gf.dev/mime-sniffing-test](https://gf.dev/mime-sniffing-test)
4. See the `MIME-type` : `content-type: txt/html;`

**Expected behavior**
The `MIME-type` in should [https://gf.dev/mime-sniffing-test](https://gf.dev/mime-sniffing-test) have been `content-type: application/rss+xml;`

**Fix propostion**
https://regex101.com/r/Rp1rix/2

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
**Reproduce the issue** ✅ 
Identified the issue on my localhost.

**Identify the root cause** ✅ 
https://github.com/wp-media/wp-rocket/blob/00cf7a1c5e0b2e7354b0aa76ca86603ee4420eae/inc/functions/options.php#L234-L235

**Scope a solution** ✅ 
The feed url should include also all sub-urls (eg. `/feed/`, `/feed/mp3/`)
Regex which should work:
`$uris[] = '/(.+/)?' . $GLOBALS['wp_rewrite']->feed_base . '/?.+/?';`

**Effort** ✅ 
Effort `[S]` , I think Unit & Integration tests will need almost all the effort

To Do
- [x] - Unit tests
- [x] - Integration tests

### DIFF
--- a/inc/admin/upgrader.php
+++ b/inc/admin/upgrader.php
@@ -356,12 +356,6 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 		wp_safe_remote_get( esc_url( home_url() ) );
 	}
 
-	if ( version_compare( $actual_version, '3.3.2', '<' )
-		||
-		version_compare( $actual_version, '3.6.1', '<' ) ) {
-		rocket_generate_config_file();
-	}
-
 	if ( version_compare( $actual_version, '3.3.6', '<' ) ) {
 		delete_site_transient( 'update_wprocket' );
 		delete_site_transient( 'update_wprocket_response' );
@@ -392,6 +386,10 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 		rocket_clean_cache_busting();
 		rocket_clean_domain();
 		rocket_generate_advanced_cache_file();
+	}
+
+	if ( version_compare( $actual_version, '3.6.1', '<' ) ) {
+		rocket_generate_config_file();
 	}
 }
 add_action( 'wp_rocket_upgrade', 'rocket_new_upgrade', 10, 2 );
--- a/inc/admin/upgrader.php
+++ b/inc/admin/upgrader.php
@@ -356,7 +356,9 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 		wp_safe_remote_get( esc_url( home_url() ) );
 	}
 
-	if ( version_compare( $actual_version, '3.3.2', '<' ) ) {
+	if ( version_compare( $actual_version, '3.3.2', '<' )
+		||
+		version_compare( $actual_version, '3.6.1', '<' ) ) {
 		rocket_generate_config_file();
 	}
 
--- a/inc/functions/options.php
+++ b/inc/functions/options.php
@@ -203,7 +203,7 @@ function rocket_get_ignored_parameters() {
  */
 function get_rocket_cache_reject_uri( $force = false ) { // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
 	static $uris;
-	$wp_rewrite = $GLOBALS['wp_rewrite'];
+	global $wp_rewrite;
 
 	if ( $force ) {
 		$uris = null;
--- a/inc/functions/options.php
+++ b/inc/functions/options.php
@@ -232,7 +232,7 @@ function get_rocket_cache_reject_uri() { // phpcs:ignore WordPress.NamingConvent
 	}
 
 	// Exclude feeds.
-	$uris[] = '/(.+/)?' . $GLOBALS['wp_rewrite']->feed_base . '/?';
+	$uris[] = '/(.+/)?' . $GLOBALS['wp_rewrite']->feed_base . '/?.+/?';
 
 	// Exlude embedded URLs.
 	$uris[] = '/(?:.+/)?embed/';
--- a/inc/functions/options.php
+++ b/inc/functions/options.php
@@ -197,24 +197,26 @@ function rocket_get_ignored_parameters() {
  * @since 2.4.1 Auto-exclude WordPress REST API.
  * @since 2.0
  *
+ * @param bool $force Force the static uris to be reverted to null.
+ *
  * @return string A pipe separated list of rejected uri.
  */
-function get_rocket_cache_reject_uri() { // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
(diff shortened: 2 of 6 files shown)


inc/functions/options.php
```diff
@@ -197,24 +197,26 @@ function rocket_get_ignored_parameters() {
...
+	}
 	if ( $uris ) {
 		return $uris;
 	}
 
-	$uris      = get_rocket_option( 'cache_reject_uri', [] );
-	$uris      = is_array( $uris ) ? $uris : [];
-	$home_root = rocket_get_home_dirname();
+	$uris              = get_rocket_option( 'cache_reject_uri', [] );
+	$uris              = is_array( $uris ) ? $uris : [];
```
Tabrisrp, line 214:
```suggestion
	$uris              = (array) get_rocket_option( 'cache_reject_uri', [] );
```

inc/functions/options.php
```diff
@@ -232,7 +234,7 @@ function get_rocket_cache_reject_uri() { // phpcs:ignore WordPress.NamingConvent
 	}
 
 	// Exclude feeds.
-	$uris[] = '/(.+/)?' . $GLOBALS['wp_rewrite']->feed_base . '/?';
+	$uris[] = '/(.+/)?' . $GLOBALS['wp_rewrite']->feed_base . '/?.+/?';
```
Tabrisrp, line 237:
```suggestion
	$uris[] = '/(.+/)?' . $wp_rewrite->feed_base . '/?.+/?';
```

And add `global $wp_rewrite;` at the start of the function. This is for consistency with the rest of the code 👍

remyperona (changes requested):
You will also need to add a call to `rocket_generate_config_file()` in `rocket_new_upgrade()` to make sure the file is updated on plugin update to 3.6.1

inc/admin/upgrader.php
```diff
@@ -356,7 +356,9 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 		wp_safe_remote_get( esc_url( home_url() ) );
 	}
 
-	if ( version_compare( $actual_version, '3.3.2', '<' ) ) {
+	if ( version_compare( $actual_version, '3.3.2', '<' )
+		||
+		version_compare( $actual_version, '3.6.1', '<' ) ) {
```
Tabrisrp, line 361: We don't need to keep the first `version_compare()`, since the second will be executed anyway. You can copy this block and put it at the end of the function instead

inc/functions/options.php
```diff
@@ -197,24 +197,27 @@ function rocket_get_ignored_parameters() {
...
  * @since 2.0
  *
+ * @param bool $force Force the static uris to be reverted to null.
+ *
  * @return string A pipe separated list of rejected uri.
  */
-function get_rocket_cache_reject_uri() { // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
+function get_rocket_cache_reject_uri( $force = false ) { // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
 	static $uris;
+	$wp_rewrite = $GLOBALS['wp_rewrite'];
```
Tabrisrp, line 206:
```suggestion
	global $wp_rewrite;
```

`$wp_rewrite`  is a global set by WordPress directly, we can access it that way

remyperona (changes requested):

tests/Integration/inc/functions/getRocketCacheRejectUri.php
```diff
@@ -0,0 +1,44 @@
...
+class Test_GetRocketCacheRejectUri extends TestCase {
+
+	/**
+	 * @dataProvider providerTestData
+	 */
+	public function testShouldReturnExcludeDeferJSArray( $config, $expected ) {
+		$options = [
+			'cache_reject_uri' => $config['options']['cache_reject_uri'],
+		];
+		update_option( 'wp_rocket_settings', $options );
```
Tabrisrp, line 22: For the options part, we have the `SettingsTrait` available to make sure everything is saved/restored to original on setup/teardown

tests/Unit/inc/functions/getRocketCacheRejectUri.php
```diff
@@ -0,0 +1,41 @@
...
+ * @group Functions
+ * @group Options
+ */
+class Test_GetRocketCacheRejectUri extends TestCase {
+	/**
+	 * @dataProvider providerTestData
+	 */
+	public function testShouldGetRocketCacheRejectUri( $config, $expected ) {
+		$GLOBALS['wp_rewrite']            = new stdClass();
+		$GLOBALS['wp_rewrite']->feed_base = 'feed/';
```
Tabrisrp, line 19: To be sure we have no conflicts between tests, add a `tearDown()` method, which will call `unset( $GLOBALS['wp_rewrite'] )`

remyperona (changes requested):

remyperona (approved):

crystinutzaa:
@wp-media/dev-qa  this one is ready for QA

arunbasillal (approved):
- Recreated the issue with master branch 
image

- Switched to fix branch. Config file was updated 
image

- Feed is no longer cached. 👍 

- Nothing logged in `debug.log`

PR #2781, opened 2020-06-16
Fixes #2730 Remove unneeded notice related to advanced-cache.php

### LINKED ISSUE #2730: Improve admin notice for invalid content in advanced-cache.php
Currently we have two admin notices related to `advanced-cache.php`

- One is when `advanced-cache.php` cannot be written into.
- Another is when `advanced-cache.php` has the wrong content.

Here they are in the above mentioned order: 
image

**The first notice above can be re-used for both cases and we shall get rid of the second notice. Here is why:** 

- If the file is writable and if the contents are not WP Rocket's, then WP Rocket updates the content.
- If the file is not writable and the contents are correct, then we do not show any notice.
- If the file is not writable and if WP Rocket has to update the file (example when we enable caching for mobile), then we can show the notice with the updated content and leave it at that. 

The second notice doesn't serve any real purpose and can be removed. 

Note: Some customers / hosts prevent writing into `advanced-cache.php` as a security measure. We should show the notice only if the content is not what it is supposed to be (as we do now). 

**To Reproduce**
- Remove writing permissions to `advanced-cache.php` (set it to 444) and delete the contents of the file. 
- Enable or disable mobile caching. 
- The notice should show up. 

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

Tabrisrp:
### inc/Engine/Cache/AdvancedCache.php
- [x] Remove notice_content_not_ours() method
- [x] Remove tests

### inc/Engine/Cache/AdminCacheSubscriber.php
- [x] Remove notice_content_not_ours() callback method
- [x] Remove tests


Tabrisrp:
@wp-media/dev-qa Ready for QA

PR #2782, opened 2020-06-16
Remove/update instances of blacklist/whitelist in the code

Tabrisrp:
We didn't have many instances of them in our codebase

We still have one instance of `blacklist`, but it's in a conditional coming from an API response, so we need the response updated before we can also update on our side.

### DIFF
--- a/inc/functions/options.php
+++ b/inc/functions/options.php
@@ -573,7 +573,7 @@ function rocket_check_key() {
 
 		if ( 'USER_BLACKLISTED' === $body ) {
 			// Translators: %1$s = opening link tag, %2$s = closing link tag.
-			$message = __( 'License validation failed. This user account is blacklisted.', 'rocket' ) . '<br>' . sprintf( __( 'Please see %1$sthis guide%2$s for more info.', 'rocket' ), '<a href="https://docs.wp-rocket.me/article/100-resolving-problems-with-license-validation#errors" rel="noopener noreferrer" target=_"blank">', '</a>' );
+			$message = __( 'License validation failed. This user account is blocked.', 'rocket' ) . '<br>' . sprintf( __( 'Please see %1$sthis guide%2$s for more info.', 'rocket' ), '<a href="https://docs.wp-rocket.me/article/100-resolving-problems-with-license-validation#errors" rel="noopener noreferrer" target=_"blank">', '</a>' );
 			set_transient( 'rocket_check_key_errors', [ $message ] );
 
 			return $return;
--- a/inc/functions/options.php
+++ b/inc/functions/options.php
@@ -571,7 +571,7 @@ function rocket_check_key() {
 			return $return;
 		}
 
-		if ( 'USER_BLACKLISTED' === $body ) {
+		if ( 'USER_BLOCKED' === $body ) {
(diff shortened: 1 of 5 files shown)


iCaspar:
I noticed instances of `whitelist` in our `phpunit.xml.dist` files, but I'm not sure we can avoid those. (I thought someone mentioned Sebastian had changed, or was going to change, those but haven't found alternative terms that accomplish the same thing in that xml.)

Tabrisrp:
Indeed, I don't think we can change those yet, as we use older version of PHPUnit currently 👍

hellofromtonya:
>I noticed instances of whitelist in our phpunit.xml.dist files, but I'm not sure we can avoid those. (I thought someone mentioned Sebastian had changed, or was going to change, those but haven't found alternative terms that accomplish the same thing in that xml.)

@iCaspar Yes, Sebastian is [deprecating them in 9.3](https://twitter.com/s_bergmann/status/1269917026107248641) and then [removing them in v10.0](https://twitter.com/s_bergmann/status/1269917027822682112).

There are other strategies we can use to eliminate the need for it in our phpunit configuration. @iCaspar let's explore that together when work on our phpunit package module.

inc/functions/options.php
```diff
@@ -573,7 +573,7 @@ function rocket_check_key() {
 
 		if ( 'USER_BLACKLISTED' === $body ) {
```
hellofromtonya, line 574: Can we deprecate this value and then swap it out with `USER_DENYLIST` instead?

Tabrisrp: That's the one instance I was talking about in the OP, it's coming from here: https://github.com/wp-media/wp-rocket.me/blob/33b97937de54fd390c249b5785eee1fe717e1b60/valid_key.php#L41

So this needs to be updated before updating the plugin

hellofromtonya: That sounds like something @iCaspar could take care of in the other repo. And then circle back to fix it here.

iCaspar: Ok, sounds like a plan. @Tabrisrp , let's look at the other repo together briefly, and I'll get it started.

iCaspar:
Also https://github.com/wp-media/wp-rocket/blob/1a84bf720eeb8d4a7e7f8a477d09c044bb17f335/languages/rocket.pot#L2684

inc/functions/options.php
Tabrisrp: we'll need to think of a way to preserve backward compatibility too on that one, as older versions of the plugin will expect the current response, while newer versions will expect the new one

Tabrisrp:
Yes for all translations files, we have a semi-automatic setup in place: https://github.com/wp-media/wp-rocket/wiki/Managing-Plugin-Translations

inc/functions/options.php
iCaspar: @GeekPress agreed that the version check we discussed would take care of back-compat issues. We actually have 2 instances in the wp-rocket.me repo that return that response. Besides the `valid_key.php` there is also one here: https://github.com/wp-media/wp-rocket.me/blob/master/check_update.php#L83

He's updated the code on the other repo and taken care of the inernal checks on redis storage, etc. Looks like a go for 3.6.2

PR #2783, opened 2020-06-17
Closes #2493 Exclude external jQuery from combine when Defer Safe Mode is active

### LINKED ISSUE #2493: Externally hosted jQuery not excluded from defer when safe mode + combine js are active
**Describe the bug**
We automatically exclude Jetpack-hosted jQuery (from their Site Accelerator feature) from being deferred when Deferred JS with Safe Mode is active. 

However, when Combine JS is also active, this file is no longer excluded. 
This is not consistent with what happens when Jetpack Site Accelerator is OFF. 

When using default WordPress-included jQuery, if _both_ Deferred JS w/ Safe Mode AND Combine JS are active, this file will be excluded and not deferred. 

So I would expect the behaviour to be the same when it's an externally-hosted version of jQuery that we detect.

I haven't checked if Jetpack is the only affected external JS or if it also applies to the [other external hosts we detect:](https://github.com/wp-media/wp-rocket/issues/1196) 

**To Reproduce**
Requires a site with Jetpack enabled. 
1. Enable Combine JS and Deferred JS with Safe Mode
2. Observe that jQuery.js is excluded from Combine and is not deferred. 
3. Enable Jetpack Site Accelerator with _Speed up static file load times_
4. Clear the Cache
5. Observe that jQuery is no longer excluded, but is included in the Combined JS file.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
**Reproduce the problem** ✅ 
Reproduced on test installation

**Identify the root cause** ✅ 
In https://github.com/wp-media/wp-rocket/blob/00cf7a1c5e0b2e7354b0aa76ca86603ee4420eae/inc/classes/optimization/JS/class-abstract-js-optimization.php#L145

We only get the URL for the WP core jQuery, not for any externally hosted jQuery like we do for deferred JS.

**Scope a solution** ✅ 
Improve this method to also exclude known externally hosted jQuery, like from Jetpack, googleapis.com or cdnjs.cloudflare.com

**Effort** ✅ 
Effort `[S]`

### DIFF
--- a/inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
+++ b/inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
@@ -39,10 +39,10 @@ abstract class AbstractJSOptimization extends AbstractOptimization {
 	 */
 	protected function get_excluded_files() {
 		$excluded_files = $this->options->get( 'exclude_js', [] );
-		$jquery_url     = $this->get_jquery_url();
+		$jquery_urls    = $this->get_jquery_urls();
 
-		if ( $jquery_url ) {
-			$excluded_files[] = $jquery_url;
+		if ( ! empty( $jquery_urls ) ) {
+			$excluded_files = array_merge( $excluded_files, $jquery_urls );
 		}
 
 		/**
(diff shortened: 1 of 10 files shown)


inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
```diff
@@ -146,23 +146,21 @@ protected function get_minify_url( $filename, $original_url = '' ) {
...
-			return false;
+			return [];
 		}
-
-		if ( '' === wp_parse_url( $wp_scripts->registered['jquery-core']->src, PHP_URL_HOST ) ) {
-			return wp_parse_url( site_url( $wp_scripts->registered['jquery-core']->src, PHP_URL_PATH ) );
-		}
-
-		return $wp_scripts->registered['jquery-core']->src;
+		$jquery           = site_url( wp_scripts()->registered['jquery-core']->src );
```
Tabrisrp, line 157: - We already have the global `$wp_scripts` at the start of the method, is there a reason to use `wp_scripts()` instead?
- Did you test that the returned value from the `src` property is always a path, and not an absolute URL? Else the URL could be incorrect

inc/Engine/Optimization/Minify/JS/Combine.php
```diff
@@ -194,6 +193,13 @@ function( $script ) {
 							}
 						}
 
+						if ( ! empty( $this->jquery_urls ) ) {
+							$jquery_urls = implode( '|', $this->jquery_urls );
+							if ( preg_match( '#^(' . $jquery_urls . ')$#', rocket_remove_url_protocol( $matches['url'] ) ) ) {
+								return;
+							}
+						}
```
Tabrisrp, line 201: What about exclusion of the internal jquery? If the check is performed here only, it won't be excluded

crystinutzaa: Yup. This is responsible only for external jquery exclusion.

Internal jquery exclusion is handled in here: `is_minify_excluded_file()`

inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
crystinutzaa: removed the global and revisited the code. It's an url and I am using `rocket_clean_exclude_file()`  function to ged rid of the domain and grab only the path.

crystinutzaa:
@wp-media/dev-qa this one is ready for QA

Tabrisrp:
@arunbasillal Yes correct, those are the ones we currently automatically exclude

PR #2785, opened 2020-06-17
Closes #2689 Include 3rd party CSS into concatenation

### LINKED ISSUE #2689: Include 3rd Party CSS into concatenation

Currently WP Rocket does not process externally hosted CSS files during CSS combination. However CSS can be safely combined in the combined CSS file and that is the purpose of this enhancement. 

## Why are we doing this? 

The expected advantages of this enhancement will be: 

- Combining external CSS files into the combined CSS file will help reduce the number of HTTP requests. 

- If the external host doesn’t set proper browser caching rules, having the file combined locally can help cache the CSS locally in the browser. 

- Both of these will help with performance and help improve PageSpeed and GTmetrix scores. 

We can follow a similar pattern as we do with external JavaScript files. External CSS files can be automatically combined when a user chooses to combine CSS files. 

<p></p>

The text area in the UI for “Excluded CSS Files” can be modified to accept external domains or full urls of external CSS files so that users can exclude external CSS files from combination (similar to how we offer for external JS files). 

## Technical Specs

- For starters let’s only include files with .css file extension. 

- We already have the filter “rocket_exclude_css” to exclude CSS files from minification and combination. We will use this one to exclude external CSS too.

- We automatically apply this to all users on WP Rocket update. 

## Risks

The biggest change would be for customers with externally hosted CSS. They will be added to the combined CSS file when they update WP Rocket. We are not expecting issues because the CSS will be loaded from the combined CSS file instead and shouldn’t cause issues. 

<p></p>

Another risk we can anticipate is similar to the cache dir size we once had with combining JS files. External CSS might be dynamic and this could lead to a separate combined CSS file per each page. However, most of the issues related to JS was because of dynamic inline JS. This should be really rare for an externally hosted file to be dynamic. 

<p></p>

Note from Arun: I checked our ticket history and did some tests. I couldn’t see a case where either of the above cases might be an issue. I consulted our team mates for potential issues and the feedback suggests that this should be a safe feature. 

## Acceptance Criteria

- A website that loads external CSS. When combine CSS is enabled, the external file is combined into the combined CSS file. 

- If the CSS file doesn’t have a .css extension, the file is not combined. (Example Google Fonts)

- If the file is added as an excluded file in the UI or via the filter, the file is excluded from combination. 

- Document the new/updated filter

- Translation of the help text in the UI

- We have a doc for JS tited "[Excluding External JS from Concatenation](https://docs.wp-rocket.me/article/39-excluding-external-js-from-concatenation)". Should we create one like this for CSS?

## Development

inc/Engine/Optimization/Minify/CSS/Combine.php

Edit the function `optimize()` and instead of bailout the external files add logic which will bailout if the file is excluded or provide the content of the external file to be combined.

`exclude_css` sanitization

Edit the sanitization of `exclude_css` input field and allow external URL.

Tabrisrp:
Closes the epic #2689 

3rd party CSS is now included in our combine CSS process.


inc/Engine/Optimization/AssetsLocalCache.php
```diff
@@ -0,0 +1,103 @@
...
+	 * Gets the filepath of the local copy for the given URL
+	 *
+	 * @since 3.7
+	 *
+	 * @param string $url URL to get filepath for.
+	 * @return string
+	 */
+	protected function get_filepath( $url ) {
+		$parts    = wp_parse_url( $url );
+		$filename = $parts['host'] . str_replace( '/', '-', $parts['path'] );
```
hellofromtonya, line 78: Is it possible that we could get here and there's no path, i.e. `$parts['path']` is empty?

Tabrisrp: It's unlikely, and I don't think we ever had this reported before, but I've seen more unlikely things happening, so I added a guard clause: https://github.com/wp-media/wp-rocket/pull/2785/commits/812b7ae185b1ab4aa2a825edfeea4d449123e3ce

Tabrisrp:
@wp-media/dev-qa Ready for a final QA with all tasks merged

Tabrisrp:
>  I noticed that when CDN is enabled or disabled, the combined CSS file is regenerated. However the Combined JS file is the same.
> Is this expected? Is this needed? @Tabrisrp

Yes, it's to make sure URLs inside the CSS files are correctly rewritten to the CDN URL, or reverted back to the original, when the CDN option is enabled/disabled.

@crystinutzaa Note the above for the exclusion field

Tabrisrp:
@wp-media/qa This is ready for another QA, the above issue has been fixed

Tabrisrp:
1. That's normal for now, the documentation is going to come at a later date, and we'll update this link accordingly
2. After looking into it, I discovered that is was an edge case. The combined CSS filenames are based on a hash created using the URLs of all the CSS files to be combined on the page, before their content is retrieved (for performance).

I compared http://mega.wp-rocket.me/qa/ & http://mega.wp-rocket.me/qa/sample-page, which contains the 2 files that have the same content:
- on the front page, there is fake file `<link rel='stylesheet' id='invalid-resource'  href="http://afakedomain.com/notadirectory/notafile.css" />`, that will still be processed, and used to create the filename, but no content will be taken from it
- on the sample page, the fake file is not there, so the filename is different, but the content generated is the same

Apart from that, both pages have the same CSS, which explains why the combined CSS files have the same content, but not the same name.

On a normal environment, this is going to happen rarely or never though.

PR #2786, opened 2020-06-17
Fix #2658 update async css method

### LINKED ISSUE #2059: Duplicate ID caused by Optimize CSS Delivery option
- Made sure you’re on the latest version ✓
- Used the search feature to ensure that the bug hasn’t been reported before ✓

**Describe the bug**
Have developed a theme for a client and checked my theme with w3 validator. There are some Errors caused by a duplicate id for all Stylesheet and JavaScript Files.
Its came from the noscript at the bottom of the page, where i think you pushed it.

And the other warnings is coused by type attributes on css/js files. 
So there are unnecessary, maybe you can delete them.

**To Reproduce**
1. Activate minimize files
2. logout
3. check page with w3 validator https://validator.w3.org

**Expected behavior**
No w3c errors caused by wp-rocket? xD

**Screenshots**
grafik

**Backlog Grooming**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

### LINKED ISSUE #2658: Update our load CSS asynchronously method
We currently use [this technique](https://www.filamentgroup.com/lab/async-css.html#a-modern-approach) to load CSS asynchronously when Optimize CSS delivery is enabled:
- `rel="preload"` instead of `rel="stylesheet"`, with and `onload` event to switch the value of the `rel` attribute

This technique works fine, but `preload` support is still not full for browsers, so that requires us to also insert the `loadCSS` polyfill script in the page, and it might cause some issues when testing pages on PageSpeed for key requests.

The [new technique](https://www.filamentgroup.com/lab/load-css-simpler/) is an improvement, where the switch occurs on the `media` attribute, using by default the value of `print`, and switching it back to its original value with the same `onload` event.

As described in the article, the advantages are:
- Better browser support out of the box
- No need for polyfill script
- using `preload` will trigger fetches very early, often too early compared to what is needed
- the browser will load the stylesheet without delaying page rendering, asynchronously!

### LINKED ISSUE #2828: Removing rocket-critical-css can lead to "delayed" FOUC in Firefox
Tested with WP Rocket 3.6.1

In some cases removing the CPCSS can result in "delayed" FOUC. 
Please see these gifs to demonstrate the difference in loading behaviour:

3.5.5.1 - http://recordit.co/Wcil5yYpgH (no fouc)
3.6.1 - http://recordit.co/PzpB8spSpN (fouc )
3.6.1 modified to keep the CPCSS - http://recordit.co/ljGzTvjIGe (no fouc)

It's not happening on all sites, so I'm not yet sure of the exact steps to reproduce. 
Related ticket: https://secure.helpscout.net/conversation/1199425512/174160?folderId=377611
Slack conversation (Cristina is assisting): https://wp-media.slack.com/archives/C43T1AYMQ/p1593110340302700

This seems to be specific to Firefox

crystinutzaa:
Closes #2658 

Closes #2059

## TODO

- [x] Add the lean [DOM Module](https://github.com/wp-media/module-dom)
- [x] Replace the regex with `HTMLDocument`

For `AsyncCSS::modify_html()`:
- [x] Update the test data fixture
- [x] Add `AsyncCSS::modify_html` unit tests
- [x] Add `AsyncCSS::modify_html` integration tests

For the `CriticalCSSSubscriber::async_css()`:
- [x] Update the test data fixture
- [x] Update the unit tests

- [x] Modify XPath query to only grab `<link>` _when_ `rel="stylesheet"` AND `href=""` is not empty.
- [x] Inserts a `<link rel="preload" href=".." as="style">` element before each stylesheet element.
- [x] Remove `id` attribute from links in the `<noscript>`
- [x] Change `{{WP_ROCKET_CONDITIONAL}}` to a comment, i.e. `<!--{{WP_ROCKET_CONDITIONAL}}-->`
- [x] Move IE Conditionals Subscriber class into new architecture
- [x] Add tests for `IEConditionalSubscriber`


inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -687,7 +594,7 @@ public function async_css( $buffer ) {
 		 */
 		$css_pattern = apply_filters(
 			'rocket_async_css_regex_pattern',
-			'/(?=<link[^>]*\s(rel\s*=\s*[\'"]stylesheet["\']))<link[^>]*\shref\s*=\s*[\'"]([^\'"]+)[\'"](.*)>/iU'
+			'/(?=<link[^>]*\s(rel\s*=\s*[\'"]stylesheet["\']))<link[^>]*\shref\s*=\s*[\'"]([^\'"]+)[\'"]\s*(type\s*=\s*[\'"].*[\'"])\s*(media\s*=\s*[\'"].*[\'"])(.*)>/iU'
```
Tabrisrp, line 597: To improve readability, I would recommend to name the different capturing groups, by using the syntax `?<name>`

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -696,8 +603,19 @@ public function async_css( $buffer ) {
...
 
+	/**
+	 * Updates the HTML code by defering all CSS matches files.
+	 *
+	 * @param  array  $tags_match All matched css files.
+	 * @param  string $buffer     HTML code.
+	 * @return string             Modified HTML code with defer CSS.
+	 */
+	private function async_css_tag_update( $tags_match, $buffer ) {
+		$excluded_css = array_flip( get_rocket_exclude_async_css() );
```
Tabrisrp, line 617: `get_rocket_exclude_async_css()` deprecate the function, and add a method in `CriticalCSS` class to replace it. It's not used anywhere else

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -707,19 +625,34 @@ public function async_css( $buffer ) {
...
-			$preload = str_replace( 'stylesheet', 'preload', $tags_match[1][ $i ] );
-			$onload  = preg_replace( '~' . preg_quote( $tags_match[3][ $i ], '~' ) . '~iU', ' as="style" onload=""' . $tags_match[3][ $i ] . '>', $tags_match[3][ $i ] );
-			$tag     = str_replace( $tags_match[3][ $i ] . '>', $onload, $tag );
-			$tag     = str_replace( $tags_match[1][ $i ], $preload, $tag );
-			$tag     = str_replace( 'onload=""', 'onload="this.onload=null;this.rel=\'stylesheet\'"', $tag );
-			$buffer  = str_replace( $tags_match[0][ $i ], $tag, $buffer );
-
+			$tag        = str_replace( $tags_match[3][ $i ], 'as="style" onload="" ' . $tags_match[3][ $i ], $tag );
+			$tag        = str_replace( $tags_match[4][ $i ], 'media="print"', $tag );
+			$tag        = str_replace( 'onload=""', 'onload="this.media=\'all\'"', $tag );
```
Tabrisrp, line 630: What if the original `media` attribute value is not `all`? We should preserve the original value.

It's also possible the `media` attribute is not set. In that case, we should add it, and use `all` by default on the `onload`

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -707,19 +625,34 @@ public function async_css( $buffer ) {
...
 		return str_replace( '</body>', $noscripts . '</body>', $buffer );
 	}
 
+	/**
+	 * Conditions which will allow to defer loading of CSS files.
+	 *
+	 * @return bool True / False if async_css should be applied.
+	 */
+	private function maybe_async_css() {
+		if ( ( defined( 'DONOTROCKETOPTIMIZE' ) && DONOTROCKETOPTIMIZE ) || ( defined( 'DONOTASYNCCSS' ) && DONOTASYNCCSS ) ) {
```
Tabrisrp, line 644: use `rocket_get_constant()` instead of the `defined`

remyperona (changes requested):

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -687,39 +594,82 @@ public function async_css( $buffer ) {
...
 		$css_pattern = apply_filters(
 			'rocket_async_css_regex_pattern',
-			'/(?=<link[^>]*\s(rel\s*=\s*[\'"]stylesheet["\']))<link[^>]*\shref\s*=\s*[\'"]([^\'"]+)[\'"](.*)>/iU'
+			'/(?=<link[^>]*\s(rel\s*=\s*[\'"]stylesheet["\']))<link(?=[^<>]*?(?:(?<type>type\s*=\s*[\'"][^<>"\']*[\'"])|>))(?=[^<>]*?(?:(?<onload>onload\s*=\s*(.*))|>))(?=[^<>]*?(?:(?<media>media\s*=\s*[\'"][^<>"\']*[\'"])|>))(?=[^<>]*?(?:href\s*=\s*[\'"](?<url>.*?)[\'"]|>))(?=[^<>]*?(?:(?<rel>rel\s*=\s*[\'"]>[^<>"\']*)[\'"]|>))(?:.*?<\/\1>|[^<>]*>)/ix'
 		);
 
 		// Get all css files with this regex.
 		preg_match_all( $css_pattern, $buffer, $tags_match );
+
 		if ( ! isset( $tags_match[0] ) ) {
```
engahmeds3ed, line 603:
```suggestion
		if ( empty( $tags_match[0] ) ) {
```

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -687,39 +594,82 @@ public function async_css( $buffer ) {
...
 			}
 
-			$preload = str_replace( 'stylesheet', 'preload', $tags_match[1][ $i ] );
-			$onload  = preg_replace( '~' . preg_quote( $tags_match[3][ $i ], '~' ) . '~iU', ' as="style" onload=""' . $tags_match[3][ $i ] . '>', $tags_match[3][ $i ] );
-			$tag     = str_replace( $tags_match[3][ $i ] . '>', $onload, $tag );
-			$tag     = str_replace( $tags_match[1][ $i ], $preload, $tag );
-			$tag     = str_replace( 'onload=""', 'onload="this.onload=null;this.rel=\'stylesheet\'"', $tag );
-			$buffer  = str_replace( $tags_match[0][ $i ], $tag, $buffer );
+			$original_media = 'all';
+			$tag            = str_replace( $tags_match['type'][ $i ], ' as="style" ' . $tags_match['type'][ $i ] . ( empty( $tags_match['media'][ $i ] ) ? ' media="print"' : '' ) . ( empty( $tags_match['onload'][ $i ] ) ? ' onload=""' : '' ), $tag );
```
engahmeds3ed, line 630: Here if we have any other attribute like `id` would be removed I think this needs to be more generic from the regex till the replacement.
`I think this is corner case`

inc/deprecated/deprecated.php
```diff
@@ -676,11 +676,10 @@ function rocket_insert_critical_css() {
  *
  * @since 2.10
  * @deprecated 2.11
- * @see Rocket_Critical_CSS->insert_load_css()
  * @author Remy Perona
```
engahmeds3ed, line 679: remove `@author`

engahmeds3ed (changes requested):

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -687,39 +594,82 @@ public function async_css( $buffer ) {
...
+	 * @param  string $buffer     HTML code.
+	 * @return string             Modified HTML code with defer CSS.
+	 */
+	private function async_css_tag_update( $tags_match, $buffer ) {
+		$excluded_css = array_flip( $this->critical_css->get_exclude_async_css() );
+		$noscripts    = '';
 		foreach ( $tags_match[0] as $i => $tag ) {
 			// Strip query args.
-			$path = rocket_extract_url_component( $tags_match[2][ $i ], PHP_URL_PATH );
+			$path = rocket_extract_url_component( $tags_match['url'][ $i ], PHP_URL_PATH );
```
Tabrisrp, line 622:
```suggestion
			$path = wp_parse_url( $tags_match['url'][ $i ], PHP_URL_PATH );
```

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
Tabrisrp: This line is too complicated to read, with multiple ternary expression in it. Simplify it by splitting into variables to concatenate after

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -687,39 +594,82 @@ public function async_css( $buffer ) {
...
+				$original_media = $media_match['media'];
+				$tag            = str_replace( $tags_match['media'][ $i ], 'media="print"', $tag );
+			}
+
+			if ( ! empty( $tags_match['onload'][ $i ] ) ) {
+				$onload_delimiter = substr( $tags_match['onload'][ $i ], 7, 1 );
+				$onload_end       = strpos( $tags_match['onload'][ $i ], $onload_delimiter . ' ' );
+				if ( ! $onload_end ) {
+					$onload_end = strpos( $tags_match['onload'][ $i ], $onload_delimiter . '>' );
+				}
```
Tabrisrp, line 642: Can you explain what is happening in this part?

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -687,39 +594,82 @@ public function async_css( $buffer ) {
...
 			$noscripts .= '<noscript>' . $tags_match[0][ $i ] . '</noscript>';
 		}
 
 		return str_replace( '</body>', $noscripts . '</body>', $buffer );
 	}
 
+	/**
+	 * Conditions which will allow to defer loading of CSS files.
+	 *
+	 * @return bool True / False if async_css should be applied.
```
Tabrisrp, line 658:
```suggestion
	 * Checks if we should apply deferring of CSS files.
	 *
	 * @return bool True if we should, false otherwise.
```

remyperona (dismissed):

inc/Engine/CriticalPath/CriticalCSS.php
```diff
@@ -551,4 +551,27 @@ public function is_async_css_mobile() {
...
+		 * @since 2.10
+		 * @author Remy Perona
+		 *
+		 * @param array $exclude_async_css An array of URLs for the CSS files to be excluded.
+		 */
+		$exclude_async_css = (array) apply_filters( 'rocket_exclude_async_css', [] );
+		$exclude_async_css = array_filter( $exclude_async_css );
+		$exclude_async_css = array_flip( array_flip( $exclude_async_css ) );
+
+		return $exclude_async_css;
```
hellofromtonya, line 575: Let's return here instead of assigning and then returning.

```suggestion
		return array_flip( array_flip( $exclude_async_css ) );
```

inc/Engine/CriticalPath/CriticalCSS.php
```diff
@@ -551,4 +551,27 @@ public function is_async_css_mobile() {
...
+	public function get_exclude_async_css() {
+		/**
+		 * Filter list of async CSS files
+		 *
+		 * @since 2.10
+		 * @author Remy Perona
+		 *
+		 * @param array $exclude_async_css An array of URLs for the CSS files to be excluded.
+		 */
+		$exclude_async_css = (array) apply_filters( 'rocket_exclude_async_css', [] );
```
hellofromtonya, line 571: If it's an empty array that gets returned, bail out. Why? There's no need to do the more expensive array operations after it.

```suggestion
		$exclude_async_css = (array) apply_filters( 'rocket_exclude_async_css', [] );
		if ( empty( $exclude_async_css ) {
			return $exclude_async_css;
		}
```

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
crystinutzaa: The issue with `onload` is that it might contain the quotes: `"` or `'` . So in order to identify the content of onload the only reliable solution which I found is to identify the delimiter:
`onload="` or `onload='` . So the next step will be identify the ending onload delimiter. Which can be either `' ` / `" ` if there are other attributes after onload or onload is the last attribute in that case it is `'>` / `">`.
This will help me to extract the full length onload attribute.

$onload_end will give the length of the onload attribute.

If anyone has another solution I am opened to it.

Regex seemed useless in this case because I was unable to identify the full content of onload if it contained '" inside it.

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -687,39 +594,82 @@ public function async_css( $buffer ) {
...
-			$tag     = str_replace( $tags_match[3][ $i ] . '>', $onload, $tag );
-			$tag     = str_replace( $tags_match[1][ $i ], $preload, $tag );
-			$tag     = str_replace( 'onload=""', 'onload="this.onload=null;this.rel=\'stylesheet\'"', $tag );
-			$buffer  = str_replace( $tags_match[0][ $i ], $tag, $buffer );
+			$original_media = 'all';
+			$tag            = str_replace( $tags_match['type'][ $i ], ' as="style" ' . $tags_match['type'][ $i ] . ( empty( $tags_match['media'][ $i ] ) ? ' media="print"' : '' ) . ( empty( $tags_match['onload'][ $i ] ) ? ' onload=""' : '' ), $tag );
+			if ( ! empty( $tags_match['media'][ $i ] ) ) {
+				preg_match( '/media\s*=\s*[\'"](?<media>.*)[\'"]/ix', $tags_match['media'][ $i ], $media_match );
+				$original_media = $media_match['media'];
+				$tag            = str_replace( $tags_match['media'][ $i ], 'media="print"', $tag );
```
hellofromtonya, line 634: To help with readability and multiple lookups, let's store in a variable:

```suggestion
				$media_attr = $tags_match['media'][ $i ];
				preg_match( '/media\s*=\s*[\'"](?<media>.*)[\'"]/ix', $media_attr, $media_match );
				$original_media = $media_match['media'];
				$tag            = str_replace( $media_attr, 'media="print"', $tag );
```

inc/deprecated/deprecated.php
crystinutzaa: I don't think we should remove author on deprecated. Should we?

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -687,39 +594,82 @@ public function async_css( $buffer ) {
...
+		if ( rocket_get_constant( 'DONOTROCKETOPTIMIZE' ) || rocket_get_constant( 'DONOTASYNCCSS' ) ) {
+			return false;
+		}
+		if ( ! $this->options->get( 'async_css' ) ) {
+			return false;
+		}
+		if ( is_rocket_post_excluded_option( 'async_css' ) ) {
+			return false;
+		}
+		return true;
```
hellofromtonya, line 670: return the conditional instead of explicitly `false` or `true`.

```suggestion
		return ! is_rocket_post_excluded_option( 'async_css' );
```

hellofromtonya (changes requested):

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
crystinutzaa: @engahmeds3ed no other tags will be removed. Only those which are needed for this change are altered or totally removed. 
I added a new fixture which has an `id="css_id"` and it will be preserved.

engahmeds3ed: Yes you are right I tried it now with adding ID and worked

engahmeds3ed (dismissed):

tests/Fixtures/inc/Engine/CriticalPath/CriticalCSSSubscriber/asyncCss.php
```diff
@@ -0,0 +1,270 @@
...
+					'DONOTROCKETOPTIMIZE' => false,
+					'DONOTASYNCCSS'       => false
+				],
+				'options' => [
+					'async_css' => true
+				],
+				'exclude_options' => [
+					'async_css' => false
+				],
+				'html' => '<!doctype html><html lang="en-US"><head><meta charset="UTF-8" /><link rel="stylesheet" type="text/css" href="http://www.example.com/file1.css" media="print" onload="alert(\'Hi\')"></head><body>Content here</body></html>'
```
hellofromtonya, line 227: Question @crystinutzaa:

If the `<link>` already has `media="print"`, why do we need `onload="this.media='print'"? That attribute already exists in the element.

tests/Fixtures/inc/Engine/CriticalPath/CriticalCSSSubscriber/asyncCss.php
```diff
@@ -0,0 +1,270 @@
...
+			],
+			'expected' => [
+				'html' => '<!doctype html><html lang="en-US"><head><meta charset="UTF-8" /><link onload="this.media=\'print\'" rel="stylesheet"  as="style" type="text/css" href="http://www.example.com/file2.css" media="print"><link rel="stylesheet"  as="style" type="text/css" onload="this.media=\'all\'" href="//www.example.com/file3.css" media="print"><link rel="stylesheet"  as="style" type="text/css" href="http://www.example.com/file1.css" media="print" onload="this.media=\'print\'"></head><body>Content here<noscript><link onload="alert(\'Hi\')" rel="stylesheet" type="text/css" href="http://www.example.com/file2.css" media="print"></noscript><noscript><link rel="stylesheet" type="text/css" href="//www.example.com/file3.css" media="all"></noscript><noscript><link rel="stylesheet" type="text/css" href="http://www.example.com/file1.css" media="print" onload=\'alert(\'Hi\')\'></noscript></body></html>'
+			]
+		],
```
hellofromtonya, line 268: Missing the following test scenarios:

- The original `<link>` already has a `media='all'` attribute.
- The original `<link>` string has its quotation marks inverted, e.g.
```html
<link rel="stylesheet" type="text/css" href="http://www.example.com/file1.css" media='print' onload='this.as="style"'>
```
- Appending the new `onload` attribute value when one already exists by is different than `this.media` such as `onload=this.href=""` or a data attribute.

hellofromtonya (dismissed):

tests/Fixtures/inc/Engine/CriticalPath/CriticalCSSSubscriber/asyncCss.php
crystinutzaa: If there is already `media="print"` then there should be also an onload="this.media='all'". Otherwise the CSS will not work. 
In case the original media is print I can add a check and switch it to all.
What do you think?

crystinutzaa:
This PR will fix also this issue: https://github.com/wp-media/wp-rocket/issues/2828
The issue is happening in Firefox only due to the old way of handling the async CSS loading.
The new way with media = print / all is clearing this issue.

@hellofromtonya while adding the code (my old one without any DomDocument) I noticed an issue. If there are missing attributes on link, then this change will not work. I assume this will be fixed by DomDocument, however I want to bring in to your attention. 
This link example should be added in tests to make sure that will pass correctly:
`<link rel="stylesheet" href="" />`

hellofromtonya:
>If there are missing attributes on link, then this change will not work. I assume this will be fixed by `DomDocument`, however I want to bring in to your attention.

@crystinutzaa One of the benefits of `DomDocument` is that we are working directly with the attributes. If you want to work with the `class` attribute, you get it from the element's node and then work directly with it. You don't have to work with _all_ the attributes like you do when manipulating a HTML string.

Yes, it will handle it natively. Phew. 

But thank you for pointing out that problem too. We'll want to make sure we test against it.

hellofromtonya (approved):
Seems weird for me to do a code review.

hellofromtonya:
@crystinutzaa @engahmeds3ed @iCaspar This PR is done and ready for code review and testing on your machines:

## For Testing

1. There should be a CPCSS stylesheet in the `<head>`.
2. It's `<link>` which have these attributes:
   - `as="style"`
   - `rel="preload"`
   - `media="print"`
   - `onload="this.onload=null;this.media='all';this.rel='stylesheet'"`
3. There should be a `<noscript><link ...></noscript>` added just above the `</body>` where the original CSS stylesheet `<link>` element lives.

## For Code Review

### What's new?

- Moved the original `CriticalCSSSubscriber::async_css()` code to a new class `AsyncCSS`. 
- Added a new module: DOM Module with has 2 classes in it:
     - `Attribute` which is a helper class of static methods to make it easier to work with attributes.
     - `HTMLDocument` which extends PHP's `DOMDocument` to make it easier to work with it.

### How does it work?

When the `'rocket_buffer'` event fires, the `CriticalCSSSubscriber::async_css()` callback is invoked and the HTML string is passed to it.

The `AsyncCSS` class has a named constructor which handles creating the object. _and_ .
       - If it's okay to create the DOM, then it:
             - passes the HTML string to `HTMLDocument::from_html()` to create an instance of the DOM
             - returns an instance of `AsyncCSS`
      - else, `null` is returned

If `null` is returned, `CriticalCSSSubscriber::async_css()` returns the original HTML string untouched; else, it invokes `AsyncCSS::modify_html()`. Now the fun starts!

#### Modifying the HTML for deferring CSS files.

`AsyncCSS::modify_html()` handles all of the modifications. It uses a `query` from the DOM. 

>Note: This query string is strange, I know. It's using XPath. In the future, we'll encapsulate that later into DOM Module.

It processes the stylesheet `<link>` nodes _if_:
    1. It has an `href` attribute and it's not empty.
    2. It has an `rel="stylesheet"` attribute.
    3. It doesn't have any of the excluded CSS URLs, if there are any.

If there are any stylesheet `<link>` elements that match the above criteria, then each is processed in the `AsyncCSS::modify_css()` method. Notice what it does:

- Adds a `<noscript>` element to the end of webpage just above the `</body>`.
- Inserts the original stylesheet `<link>` element into the `<noscript>` element.
- Changes the `rel` attribute from `rel="stylesheet"` to `rel="preload"`.
- Adds the `as="style"` attribute
- Builds the `onload` attribute (more on this later)
- Sets the `media` attribute to `media="print"`.

##### Building the `onload` attribute value

Original `onload` attribute values are retained but validated and updated if necessary. This is done in the `AsyncCSS::get_onload_values()` method. For example:

Let's say we have this original stylesheet element:

```html
<link rel="stylesheet" type='text/css' href='https://example.org/file4.css' onload='  console.log("Hello");  ' media="screen">
```

What do you notice? It has an `onload` attribute already. It's a function that's passing a string. 

After running it through this method:

- the original element is copied to the `<noscript>` element `</body>`:

```html
<noscript>
	<link rel="stylesheet" type="text/css" href="https://example.org/file4.css" onload='  console.log("Hello");  ' media="screen">
</noscript>
```

- the original element in changed to:

```html
<link rel="preload" type="text/css" href="https://example.org/file4.css" onload="console.log('Hello');this.onload=null;this.media='screen';this.rel='stylesheet'" media="print" as="style">
```

What changed in the new element?

1. The double quotes around the `"Hello"` inside of the `console.log` function changed to single quotes.
2. The extra whitespace is gone. Less bytes.
3. It added `this.onload=null;this.media='screen';this.rel='stylesheet'`.
    - Notice that it grabs the original `media` attribute value, if one exists, and inserts into the `onboard`.

There are many more examples in the [test fixture here](https://github.com/wp-media/wp-rocket/blob/fix/2658-update-async-css-method/tests/Fixtures/inc/Engine/CriticalPath/AsyncCSS/modifyHtml.php)

## Notes

`DOMDocument` automatically fixes malformed HTML. What do I mean by that?

- If there an element is missing its required closing tag, it will insert it. There are examples in the test fixture.
     - Note: The browser does the same thing as it tries to fix malformed HTML.
- Single quotes around the attribute value is changed to double quotes. For example: `rel='stylesheet'` is changed to `rel="stylesheet".
- Whitespace is automatically removed. For example: `rel="     stylesheet   "` is changed to `rel="stylesheet"`. `rel="stylesheet"     href=".."` is changed to ``rel="stylesheet" href=".."`.

inc/Engine/CriticalPath/AsyncCSS.php
```diff
@@ -0,0 +1,376 @@
...
+	 * @var array
+	 */
+	private $css_urls_to_exclude;
+
+	/**
+	 * The "onload" attribute defaults.
+	 *
+	 * @var array
+	 */
+	private $onload_defaults = [
```
crystinutzaa, line 52: This does not "feel" right. Shouldn't it be something like `attribute_defaults` to make it feel way global than `$onload_defaults` ? 🤔 Because it contains also `this.rel` which is not related to `onload` attributes or changes required.

crystinutzaa:
I cannot review it because I'm the owner of this PR. I added a single comment on it. In rest it seems great from here 👍 
Great Job @hellofromtonya 🏅

inc/Engine/CriticalPath/AsyncCSS.php
hellofromtonya: This is only used for the `onload` attribute values. It's used to create this:

```html
onload="this.onload=null;this.media='all';this.rel='stylesheet'"
```

Notice that the `rel` in there too, which will change from `rel="preload"` to `rel="stylesheet"` when "onload" event fires.

hellofromtonya:
@wp-media/qa This PR is ready for QA. It's a complete rewrite of the `async_css` method, as we have moved away from working with the raw HTML as a string (via regex) to using the Document Object Model (DOM). [See my comment here for more details](https://github.com/wp-media/wp-rocket/pull/2786#issuecomment-653250718).

The [test data fixture](https://github.com/wp-media/wp-rocket/blob/fix/2658-update-async-css-method/tests/Fixtures/inc/Engine/CriticalPath/CriticalCSSSubscriber/asyncCss.php#L56-L693) includes all of the scenarios with the raw HTML and expected HTML after it runs. The `key` for each test scenario tells you the behavior for this test data. 

For example, this test scenario returns the HTML without touching the stylesheet `<link>` when the `href` is missing or empty.

```php
		// Check for empty href.
		'shouldNotProcessWhenHrefIsEmpty' => [
			'html'     => <<<HTML
<!doctype html>
<html lang="en-US">
<head>
	<meta charset="UTF-8">
	<link rel="stylesheet" type="text/css" href="" media="all">
	<link rel="stylesheet" type="text/css" href=" " media="all">
	<link rel="stylesheet" type="text/css" href='' media="all">
	<link rel='stylesheet' type="text/css" media="all">
</head>
<body>
	<div>
		<h1>Testing</h1>
		<p>Testing when CSS links do not have a href or href is empty.</p>
	</div>
</body>
</html>
HTML
			,
			'expected' => <<<HTML
<!DOCTYPE html>
<html lang="en-US">
<head>
	<meta charset="UTF-8">
	<link rel="stylesheet" type="text/css" href="" media="all">
	<link rel="stylesheet" type="text/css" href="" media="all">
	<link rel="stylesheet" type="text/css" href="" media="all">
	<link rel="stylesheet" type="text/css" media="all">
</head>
<body>
	<div>
		<h1>Testing</h1>
		<p>Testing when CSS links do not have a href or href is empty.</p>
	</div>
	<script>const wprRemoveCPCSS = () => { \$elem = document.getElementById( "rocket-critical-css" ); if ( \$elem ) { \$elem.remove(); } }; if ( window.addEventListener ) { window.addEventListener( "load", wprRemoveCPCSS ); } else if ( window.attachEvent ) { window.attachEvent( "onload", wprRemoveCPCSS ); }</script>
</body>
</html>
HTML
			,
		],
```

Notice that the scenario includes the original `'html'` and the returned `'expected'` HTML.

hellofromtonya:
@Tabrisrp Can you review this one on Monday please?

Davidacu (changes requested):
testing environment:
http://mega.wp-rocket.me/qa

Test agaisnt `fix/2658-update-async-css-method` branch
- CPCSS is added in the head - **Passed**
- a `<noscript><link ....></noscript>` with original stylesheet is added just above the `</body>` - **Passed**
- The stylesheets attributes are updated  to
	`as="style"` - **Passed**
	`rel="preload"` - **Passed**
	`media="print"` - **Passed**
	`onload="this.onload=null;this.media='all';this.rel='stylesheet'"` - **Passed**
	
**Issues:**
The page layout is broken, there seems to be something wrong with loading the CSS.
when inspecting the HTML, I can see that the `<link>` still have attribute `media="print"`
according to this new technique that we are using to load the CSS asynchronously,
shouldn't we update the media attribute to "all" in `onload`?

The page layout breaks right after we remove the CPCSS from the header
when I manually set the media attribute of the stylesheets to "all" in the browser, 
the layout starts getting back to normal.
http://sc.com.ly/show/Llkj6UPQnfoZFE9M0tuayw3GRcKxY2B8

hellofromtonya:
@Davidacu Doh on my part. That was happening because I kept the old behavior of changing the `rel` attribute from `rel="stylesheet"` to `rel="preload"`. This PR should remove that behavior in favor of the new approach of changing `media` attribute from `media="all"` to `media="print"`.

Fixed in commit https://github.com/wp-media/wp-rocket/pull/2786/commits/71c79c3f97a3ad1f000e92b883b3ccbd69988fa6

Go ahead and re-do the QA. Thanks.

GeekPress:
To double check, we also put a preload link tag in addition the one which contains the onload attribute, isn’t it?

hellofromtonya:
>To double check, we also put a preload link tag in addition the one which contains the onload attribute, isn’t it?

We are applying this async change to all stylesheets that are not currently `rel="preload"`. Adding an additional preload `<link rel="preload"...>` means non-critical stylesheets would also receive the highest-priority for fetching the resource from the server.

@GeekPress is this what we want? Aren't we seeking to defer non-critical styles to fetch later? Or do we want to prioritize all stylesheets to the highest priority?

GeekPress:
> is this what we want?

@hellofromtonya Please check my benchmarks here: https://github.com/wp-media/wp-rocket/issues/2658#issuecomment-630191512

The preload tag has a very important impact on the loading time. 

> Aren't we seeking to defer non-critical styles to fetch later? Or do we want to prioritize all stylesheets to the highest priority?

When we don't have the concatenated option turned on, this question is totally legitimate. I'm agree with you that non critical CSS should be fetch later. 
However, since day 1 on CPCSS option, we are preloading all CSS files. If we change this behavior, our customer won't understand why with 3.6.2 their CSS will be loaded slower.

hellofromtonya:
@GeekPress Thanks for the explanation. I appreciate it. And yes, this PR has the separate `<link rel="preload" ...>` element.

Davidacu (changes requested):
**Testing environment:**
http://mega.wp-rocket.me/qa


Tested with 
- No Minification or Combined CSS - Passed
- Minified CSS - Passed
- Minified and Combined CSS Files - Passed

For all variants, we generate 2 links for each stylesheet, one with preload and the one with the new technique that adds media attribute.

**Issues:**
The CSS seems to be loaded fine now, but the resulting layout doesn't seem to match the original layout before applying Optimize CSS delivery.
1. The `` in the header is not displayed
2. The sidebar in the right-handside which includes the search bar, recent posts etc... 
   is loaded as regular content and not as a sidebar on the right-handside.
  
I believe this is caused because for some reason we are removing the classes applied to the `<body>` tag.

Before Enabling Optimize CSS delivery:
https://i.imgur.com/eZ1uMag.png
After Enabling Optimize CSS delivery:
https://i.imgur.com/yyhWFCJ.png

Video: http://sc.com.ly/show/Xu7tj3LSzV29qHdeJQnRDmlUNOwYh1oP

remyperona (approved):
Approving the code of this. What I'm worried about is DOMDocument itself, as we had issues with it in the past, breaking pages and markup. This really needs to be well tested

hellofromtonya:
@Davidacu Does this QA report apply to this PR? https://github.com/wp-media/wp-rocket/pull/2786#pullrequestreview-443190516

Davidacu:
>@Davidacu Does this QA report apply to this PR? #2786 (review)

Yes, it is for PR https://github.com/wp-media/wp-rocket/pull/2786 @hellofromtonya 
Please let me know if there is anything I can help with.
I have just added this video to the last report http://sc.com.ly/show/Xu7tj3LSzV29qHdeJQnRDmlUNOwYh1oP

hellofromtonya:
Thank you @Davidacu. The extra video helps. 

Creating a test site on our Siteground engineering account for more extensive testing. I'm setting it up to match the QA testing environment. Let's see what's causing the malformed `<head>` HTML which is causing the problem David notes.

Tabrisrp:
Since we're working on this specific part and making big changes, I think we should include the fix for #2059 at the same time

hellofromtonya:
The problem that David identified is due to the template tag we insert into the HTML as a placeholder for the IE conditional comments, i.e. `{{WP_ROCKET_CONDITIONAL}}`. This is not valid HTML, but rather a placeholder for future insertion.

One solution is to wrap the template tag in a HTML comment converting it to `<!--{{WP_ROCKET_CONDITIONAL}}-->`. 

@Tabrisrp Do you foresee any problems with me do this in the `class-ie-conditionals-subscriber.php` file?

Tabrisrp:
@hellofromtonya I believe that should be working too with this formatting. One thing to not is that there is a `hide_comments()` method in the `AbstractOptimization` class, that is removing the comments blocks like this from the HTML before parsing.

But after that the optimization is applied on the original provided HTML, so there should be no impact.

hellofromtonya:
@Tabrisrp Please do a code review of the following additions:

- Refactoring of `IEConditionalSubscriber`
    - Moves into the new architecture
    - Changes the template tag into a comment, i.e. `<!--{{WP_ROCKET_CONDITIONAL}}-->` instead of `{{WP_ROCKET_CONDITIONAL}}`. Why? A comment is valid HTML whereas the template tag without being a comment is not. This is what caused the error that @Davidacu found.
   - Adds tests
- Removes the `id` attribute from each stylesheet link in the `<noscript>` node

hellofromtonya:
>I believe that should be working too with this formatting. One thing to not is that there is a hide_comments() method in the AbstractOptimization class, that is removing the comments blocks like this from the HTML before parsing.
>
>But after that the optimization is applied on the original provided HTML, so there should be no impact.

@Tabrisrp I added test cases to validate that we are getting the expected HTML from the initial HTML sent into `rocket_buffer` and processes through all of the callbacks in the Optimization and Critical Path classes.

remyperona (approved):

hellofromtonya:
@Davidacu This PR is ready for another QA session. What's changed?

- The issue you found has been fixed. It was due to how we were handling IE conditionals. 
How so? We were adding a temporary placeholder template tag `{{WP_ROCKET_CONDITIONAL}}` which is invalid HTML. The DOM was autofixing it. This PR changes that tag to be a HTML comment `<!--{{WP_ROCKET_CONDITIONAL}}-->`. You won't see this in your QA tests as it happens behind-the-scenes as part of the `'rocket_buffer'` filtering and optimization process.

- This PR now fixes issue #2059 too.
How? It removes the `id` attribute from the original stylesheet `<link>` elements that are copied into the `<noscript>` at the bottom of the web page. There should be no `id'` attributes in the added `<noscript>` `<link>` elements.

This PR is 🔥  priority 🔥  as it's blocking the 3.6.2 release.

Davidacu (approved):
- Previous issues caused by IE conditional statements in HTML is no longer reproduced. - **Passed**
- I can confirm that id attribute from stylesheets is not added in `<noscript>` - **Passed**

Made last round of regression tests:
- CPCSS is generated for each post type
- CPCSS is removed in onload
- `media=print` attribute is added for stylesheets with non-empty `href` and updated to `media=all`
- Stylesheets are pre-loaded
- CPCSS fallback works as expected in Firefox, Chrome, IE11, Safari, and Opera.
- Exclude page from Optimize Delivery works as expected.
- Malformed HTML is handled as expected
- Validated switching themes regenerates proper CPCSS
- All variants of Minified CSS/JS, Combined CSS/JS

Issue #2682, opened 2020-05-26
Get rid of Minify HTML option


GeekPress:
## **What are we proposing to do?**

With this update we are hoping to remove the « Minify HTML » option from WP Rocket UI (File Optimization tab).

## Why are we doing this?

GT Metrix has deprecated this recommandation as the effect on overall performance is very small :

<https://gtmetrix.com/blog/recommendation-updates-http-2-changes-and-more/?fbclid=IwAR1ZQ2FUmlbazFScrVIAUG08MV5-yVXVXU4J36vubnvZ6MeclF43QwAla44>

GT Metrix was the latest performance tool to recommend HTML minification. PageSpeed Insight and Pingdom Tools ignore this recommandation since a while. 

Here a GT Metrix test comparaison:

**With Minify HTML**

<https://gtmetrix.com/reports/wp-rocket.me/JUC11exR>

- Grade: 100 / 96

- Loading time: 0.8s

- Page Size: 956KB

**Without Minify HTML**

<https://gtmetrix.com/reports/wp-rocket.me/Xv7fZ4DG>

- Grade: 100 / 96

- Loading time: 0.8s

- Page Size: 957KB

The loading time is exactly the same. Why? The page size only increased by 0.01KB...

## How will we do this feature?

From version 3.7, all users will not have this option in WP Rocket UI and HTML minification won't be applied anymore.

Both existing and new users will have the same experience.

## Development

inc/Plugin.php

- Remove the `minify_html_subscriber` entry from the subscribers

inc/admin/upgrader.php

- Remove the entry for `minify_html` in `rocket_first_install()`

- Remove the checks for `minify_html` in `rocket_new_upgrade()`

inc/admin/ui/meta-boxes.php

- Remove `minify_html` from the `$fields` in `rocket_display_cache_options_meta_boxes()` and `rocket_save_metabox_options()`

inc/admin/ui/notices.php

- Remove the checks for `minify_html` in `rocket_plugins_to_deactivate()`

inc/3rd-party/hosting/wp-serveur.php

- Deprecate the `rocket_deactivate_inline_js_on_wp_serveur()` function

inc/3rd-party/plugins/appbanner.php

- Deprecate the `rocket_deactivate_js_minifier_with_appbanner()` function

- Delete the file

- Remove the require in `inc/3rd-party.php`

inc/3rd-party/plugins/autoptimize.php

- Deprecate the `rocket_maybe_deactivate_minify_html()` function

[...]

PR #2806, opened 2020-06-23
Remove minify HTML option: front (1/3)

crystinutzaa:
Sub-task for epic #2682 

### inc/Plugin.php ✔️ 
- Remove the minify_html_subscriber entry from the subscribers

### inc/classes/subscriber/Optimization/class-minify-html-subscriber.php ✔️ 
- Deprecate the class

### inc/Engine/Optimization/ServiceProvider.php ✔️ 
- Remove registering minify HTML subscriber in $provides and register()

### inc/vendors/classes/class-minify-html.php ✔️ 
- Deprecate the class

### DIFF
--- a/inc/Engine/Optimization/ServiceProvider.php
+++ b/inc/Engine/Optimization/ServiceProvider.php
@@ -28,7 +28,6 @@ class ServiceProvider extends AbstractServiceProvider {
 		'buffer_subscriber',
 		'cache_dynamic_resource',
 		'ie_conditionals_subscriber',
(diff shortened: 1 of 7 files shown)


crystinutzaa:
@hellofromtonya & @Tabrisrp I don't think we need QA for this one. What do you think? Can this be merged and tested when the full Epic is done?

hellofromtonya:
@crystinutzaa I agree with you. Let's merge it into the feature epic and then do QA on the whole feature.

PR #2808, opened 2020-06-23
Automatic licensing file check

ercchy:
A hook to check out the `license-data.php` if it is missing in the PR


ercchy:
@hellofromtonya I have made some tests with this, that is why it has so many commits.

So the way this works is when I create a PR to the `develop` or `master` a Github action [runs](https://github.com/wp-media/wp-rocket/actions?query=workflow%3A%22Rebuild+Licence+data+file%22+branch%3Alicencing_file_check). 
If it works correctly, the issue of missing file will not happen again, but developers will not know anything has changed apart from seeing the check here.

p.s:. I am not sure why Travis is not passing, but I will take a look at this tomorrow.

hellofromtonya:
>p.s:. I am not sure why Travis is not passing, but I will take a look at this tomorrow.

Travis is failing due to a change on RocketCDN. [See the Slack thread here](https://wp-media.slack.com/archives/C015Q3UDNRF/p1592937722001200).

hellofromtonya:
@Tabrisrp What do you think?

Tabrisrp:
Looks good to me! I like the use of Github actions for this 👍

Issue #2734, opened 2020-06-05
Remove minify HTML option: 3rd party (3/3)


Tabrisrp:
Sub-task for epic #2682

### inc/3rd-party/hosting/wp-serveur.php
- Deprecate the `rocket_deactivate_inline_js_on_wp_serveur()` function

### inc/3rd-party/plugins/appbanner.php
- Deprecate the `rocket_deactivate_js_minifier_with_appbanner()` function
- Delete the file
- Remove the require in `inc/3rd-party.php`

### inc/3rd-party/plugins/autoptimize.php
- Deprecate the `rocket_maybe_deactivate_minify_html()` function
- Update `rocket_activate_autoptimize()` to remove the check for `minify_html`
- Deprecate `rocket_maybe_disable_minify_html()`

### inc/3rd-party/plugins/slider/revslider.php
- Deprecate the `rocket_deactivate_js_minifier_with_revslider()` function
- Delete the file
- Remove the require in `inc/3rd-party.php`

Issue #2733, opened 2020-06-05
Remove minify HTML: backend (2/3)


Tabrisrp:
Sub-task for epic #2682 

### inc/admin/upgrader.php
- Remove the entry for `minify_html` in `rocket_first_install()`
- Remove the checks for `minify_html` in `rocket_new_upgrade()`

### inc/admin/ui/meta-boxes.php
- Remove `minify_html` from the `$fields` in `rocket_display_cache_options_meta_boxes()` and `rocket_save_metabox_options()`

### inc/admin/ui/notices.php
- Remove the checks for `minify_html` in `rocket_plugins_to_deactivate()`

### inc/Engine/Beacon/Beacon.php
- Remove `minify_html` from `session_data()`

### inc/Engine/Admin/Deactivation/DeactivationIntent.php
- Remove `minify_html` from `activate_safe_mode()`

### inc/Engine/Admin/Settings/Page.php
- Remove references to minify HTML option

### inc/Engine/Admin/Settings/Settings.php
- Remove references to `minify_html` in `sanitize_callback()`

Issue #2732, opened 2020-06-05
Remove minify HTML option: front (1/3)


Tabrisrp:
Sub-task for epic #2682 

### inc/Plugin.php
- Remove the minify_html_subscriber entry from the subscribers

### inc/classes/subscriber/Optimization/class-minify-html-subscriber.php
- Deprecate the class

### inc/Engine/Optimization/ServiceProvider.php
- Remove registering minify HTML subscriber in $provides and register()

### inc/vendors/classes/class-minify-html.php
- Deprecate the class

PR #2813, opened 2020-06-23
Get rid of Minify HTML option

hellofromtonya:
This PR is for the epic #2682. All 3 sub-tasks have been completed and merged:

- Part 1: Front issue #2732 
- Part 2: Backend issue #2733 
- Part 3: Option - 3rd Party #2734 

No QA was done in the sub-tasks. Rather, we want to do full QA here for this epic.


hellofromtonya:
@arunbasillal @Davidacu This epic is ready for QA.

Davidacu:
**Environment**: local
**branch**: enhancement/remove-minify-html-option

All the cases below have **passed** and confirmed to work as expected

**WP Rocket Settings - File Optimization**
- Validate that Minify HTML option is removed from File optimization tab 
- Validate the actual page source code HTML is not minified. 

**Admin - Edit Post - WP Rocket Options**
- Validate that the option "Minify HTML"  under the right-hand sidebar is not displayed when editing a post

**Activate WP Super Minify plugin:**
- when minifying CSS and javascript is disabled on WP Rocket - notice does not appear - OK
- when minifying CSS or javascript is enabled on WP Rocket - notice appears. - OK

**3rd Party:**
- Validated that when HTML minification is enabled in Autoptimize there is no message printed under File Optimization - "Basic settings" in WP Rocket
- Validated that when CSS/JS minification is enabled in Autoptimize there is a message printer under File Optimization - Css Files / Js Files

**Smoke tests**:
1. Activate/Deactivate WP Rocket plugin - OK but some remarks, see more details on items that need attention.
2. Save changes in WP Rocket settings under file optimization without making any changes. - OK.
3. Enable/Disable Minify CSS/JS - OK but some remarks, see more details on items that need attention.
4. Update Post without making changes - OK.

**Items that need attention**:
1. Debug log prints PHP Notice about Minify_HTML being depecrated in version > 3.7, - doesn't seem ok ->
logs  :point_right:  http://snippi.com/raw/bw6fx66
video :point_right: https://screencast-o-matic.com/watch/cY1Z3T5nQN

2. Debug log prints `PHP warnings call_user_func() expects parameter 1 to be a valid callback, class 'JSMin' not found` - Doesn't seem ok, I confirm is not regression, it is present in current master branch
video 👉  https://screencast-o-matic.com/watch/cY1TnQ5nG1
full log 👉 http://snippi.com/raw/uma6bhe , Please advise if we can create a separate issue for this one.

Both of these issues are reproduced during plugin activation or when loading WP Rocket settings UI and trying to enable/disable Minify CSS/JS, whenever there is a conflict with a 3rd party plugin such as Autoptimize or WP Super Minify plugin.
Please have a look @hellofromtonya

hellofromtonya:
>Debug log prints PHP Notice about Minify_HTML being depecrated in version > 3.7

@Davidacu 

The notice is the correct and expected behavior. Why? The `Minify_HTML` class and file was deprecated in 3.7. Since these 3rd party plugins are using the class, here's the expected behavior:

- The class will load into memory and works as it did before.
- A notice will be triggered to alert that a deprecated class is being used.

That notice alerts people to upgrade. For us, we should let these 3rd party plugins know that this class has been deprecated and their customers/users are getting a PHP notice.

arunbasillal:
@hellofromtonya What about the `PHP warnings call_user_func() expects parameter 1 to be a valid callback, class 'JSMin' not found` warning? 

Should we do something about this? Or is it related to https://github.com/wp-media/wp-rocket/pull/2813#issuecomment-649714462 as well?

Davidacu:
> That notice alerts people to upgrade. For us, we should let these 3rd party plugins know that this class has been deprecated and their customers/users are getting a PHP notice.

I see, that makes sense! Thanks for clearing that out.

hellofromtonya:
>Debug log prints PHP warnings call_user_func() expects parameter 1 to be a valid callback, class 'JSMin' not found

@arunbasillal @Davidacu This warning is due to a logic problem in the 3rd party plugins, i.e. that is not attributed to Rocket. Let me explain here so we all have context.

The 3rd party plugins have their own minify logic that loads up dependencies _when_ our `Minify_HTML` class is not in memory (hasn't been loaded). The problem though is their plugins need the `JSMin` for other things. We don't load `JSMIn`. That has nothing to do with Rocket. Rather, it's a logic error in how and when they are loading dependencies they need in their plugin.

PR #2819, opened 2020-06-25
Update $html_options to enable inline JavaScript minification.

### LINKED ISSUE #2818: Inline JavaScript is not minified when HTML minification is on
Inline JavaScript is not minified when HTML minification is on.
I've tested this with the following versions:
* 3.6.1
* 3.5.5.1
* 3.4.4

This is because of the HTML options not including the method to `minify_inline_js` that handles minification of inline JavaScript:
https://github.com/wp-media/wp-rocket/blob/0455940481777146a4677dbf59ed928223351639/inc/classes/subscriber/Optimization/class-minify-html-subscriber.php#L69-L70

**To Reproduce**

1. Add some inline JavaScript: http://snippi.com/s/wjutn82
2. Make sure HTML minification is enabled.
3. Check the source code of the page. JavaScript will not be minified.

**Expected behavior**
Inline JavaScript should be minified.

**Additional context**

HTML minification will be removed in one of the next WP Rocket versions, but this is an easy fix if we decide to do it.

**Related ticket**

https://secure.helpscout.net/conversation/1204067401/175422?folderId=2135277

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

vmanthos:
Fixes #2818

### DIFF
--- a/inc/classes/subscriber/Optimization/class-minify-html-subscriber.php
+++ b/inc/classes/subscriber/Optimization/class-minify-html-subscriber.php
@@ -68,6 +68,7 @@ class Minify_HTML_Subscriber implements Subscriber_Interface {
 
 		$html_options = [
 			'cssMinifier' => [ $this, 'minify_inline_css' ],
+			'jsMinifier'  => [ $this, 'minify_inline_js' ],
 		];
 
 		/**


GeekPress:
Closing this PR for the reasons explained here https://github.com/wp-media/wp-rocket/issues/2818#issuecomment-649239359

PR #2823, opened 2020-06-25
WP Rocket options page missing icons

### LINKED ISSUE #2723: WP Rocket options page missing icons
- Made sure you’re on the latest version :heavy_check_mark:
- Used the search feature to ensure that the bug hasn’t been reported before :heavy_check_mark:

**Describe the bug**
On some servers, and if WooCommerce has the geolocation option enabled, fonts/icomoon.eot?-4yq2oj will be redirected to fonts/icomoon.eot?v=XXX, which will ultimately fail.

**To Reproduce**
Steps to reproduce the behavior (only on faulty servers):
1. Go to WooCommerce and activate Geolocation.
2. Go to WP Rocket options page.
3. See error on WP Rocket options pages and browse console.

**Expected behavior**
No redirects. Icons showing properly.

**Screenshots**
vgy.me
vgy.me
vgy.me

**Additional context**
It only happens on misconfigured servers AND WooCommerce has geolocation enabled. 
Removing the "-" from fonts/icomoon.eot?-4yq2oj fixes the issue.

Ticket - https://secure.helpscout.net/conversation/1183758870/170104/

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
Closes #2723 

### Scope a solution ✅ 
- Remove the query string from the font URL in `src/scss/base/_icons.scss`
- Regenerate the production CSS file

### Estimate the effort ✅ 
Effort `[XS]`


hellofromtonya (approved):

hellofromtonya:
@arunbasillal @Davidacu This PR is ready for QA.

arunbasillal (commented):
@Davidacu and I tried to recreate this, but couldn't replicate the original issue. 

We simulated the "misconfigured server" by doing a 301 redirect on the font file. However, the redirect was for fonts with and without the ending `?-4yq2oj`, so the update didn't fix it. 

But we could confirm that nothing was broken visually on a normal website 👍 

@crystinutzaa Were you able to recreate the original issue? Why did these fonts have the query strings in the first place? 

One way to test this would be to send the fix version to the customer. Is the `fix/2723-wp-rocket-missing-icons-on-options-page` branch stable enough to be sent to the customer?

crystinutzaa:
@arunbasillal I was unable to recreate this issue. Remy groomed it and stated that the solution will be to remove that query string.
Yes, a solution to test it is to send to the customer this branch for testing. Another solution is to send only the file: ` assets/css/wpr-admin.css` to be uploaded in their current version of WP Rocket and see if it fixes the issue.

arunbasillal:
@crystinutzaa Thank you for the feedback. Some interest research suggested that the query string is for cache busting as usual. Hopefully we won't have an issue with it. 

I have sent the fix branch to the customer. Will report back when the customer confirms.

arunbasillal:
The customer has not replied yet to us. I just followed up with the customer. 🤞

arunbasillal (approved):
I don't think the customer will get back to us. Since we tested to make sure it doesn't break anything for normal installs and since it's only a query string removal, lets merge this and move forward. 

What do you think?

crystinutzaa:
@hellofromtonya & @Tabrisrp what do you think? It's a safe merge ?

Tabrisrp:
IMO yes, we're not going to update the font anyway, so the version doesn't matter.

PR #2824, opened 2020-06-25
Disable the Varnish purge on Cloudways when Varnish has been disabled…

### LINKED ISSUE #2668: Disable the Varnish purge on Cloudways when Varnish has been disabled on the server side
**Describe the bug**
When a customer has its site hosted on Cloudways WP Rocket it automatically activates cache synchronization with Varnish and the toggle button for Varnish add-on doesn't show on dashboard. but when the customer disables Varnish on the server-side there is no way to disable the Varnish purge from the WP Rocket side, this causes the site's error log to be filled in with messages due to the failed request. 

**To Reproduce**
1. In a Cloudways test environment, disable Varnish on the server-side.
2. Check in the WP Rocket dashboard that the toggle button on the Varnish add-on is not present.
image

**Expected behavior**
When Varnish is turned off on the server-side, Varnish synchronization is automatically turned off on WP Rocket side and can be clearly seen on the dashboard

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

crystinutzaa:
Closes #2668 

### Scope a solution ✅ 
In `inc/ThirdParty/Hostings/Cloudways.php`, update the check in `get_subscribed_events()` to also check for `HTTP_X_VARNISH` & `HTTP_X_APPLICATION` values, following this logic:
- X-Varnish → True → X-Application → varnishpass → return false
- X-Varnish → True → X-Application → other than varnishpass → Go to function
- X-Varnish → False → return false

### Estimate the effort ✅ 
Effort `[XS]`

### DIFF
--- a/inc/ThirdParty/Hostings/Cloudways.php
+++ b/inc/ThirdParty/Hostings/Cloudways.php
@@ -18,7 +18,7 @@ class Cloudways implements Subscriber_Interface {
 	 * @return array
 	 */
 	public static function get_subscribed_events() {
-		if ( ! isset( $_SERVER['cw_allowed_ip'] ) ) {
+		if ( ! isset( $_SERVER['cw_allowed_ip'] ) || ! self::is_varnish_running() ) {
 			return [];
 		}
 
@@ -28,6 +28,26 @@ class Cloudways implements Subscriber_Interface {
 			'rocket_varnish_field_settings'           => 'varnish_addon_title',
 			'rocket_varnish_ip'                       => 'varnish_ip',
 		];
+	}
+
+	/**
+	 * Determine if the Varnish server is up and running.
+	 *
+	 * @since 3.6.1
+	 */
+	private static function is_varnish_running() {
+		if ( ! isset( $_SERVER['HTTP_X_VARNISH'] ) ) {
+			return false;
+		}
+
+		if ( ! isset( $_SERVER['HTTP_X_APPLICATION'] ) ) {
+			return false;
+		}
+
+		if ( 'varnishpass' === trim( strtolower( $_SERVER['HTTP_X_APPLICATION'] ) ) ) { // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.ValidatedSanitizedInput.MissingUnslash
+			return false;
+		}
+		return true;
 	}
 
 	/**
--- a/inc/ThirdParty/Hostings/Cloudways.php
+++ b/inc/ThirdParty/Hostings/Cloudways.php
@@ -44,10 +44,7 @@ class Cloudways implements Subscriber_Interface {
 			return false;
 		}
 
-		if ( 'varnishpass' === trim( strtolower( $_SERVER['HTTP_X_APPLICATION'] ) ) ) { // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.ValidatedSanitizedInput.MissingUnslash
-			return false;
-		}
-		return true;
+		return ( 'varnishpass' !== trim( strtolower( $_SERVER['HTTP_X_APPLICATION'] ) ) ); // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.ValidatedSanitizedInput.MissingUnslash
 	}
 
 	/**
--- a/inc/ThirdParty/Hostings/Cloudways.php
+++ b/inc/ThirdParty/Hostings/Cloudways.php
@@ -18,13 +18,13 @@ class Cloudways implements Subscriber_Interface {
 	 * @return array
 	 */
 	public static function get_subscribed_events() {
-		if ( ! isset( $_SERVER['cw_allowed_ip'] ) || ! self::is_varnish_running() ) {
+		if ( ! isset( $_SERVER['cw_allowed_ip'] ) ) {
 			return [];
 		}
 
 		return [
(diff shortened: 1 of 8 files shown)


inc/ThirdParty/Hostings/Cloudways.php
```diff
@@ -30,6 +30,26 @@ public static function get_subscribed_events() {
...
+		}
+
+		if ( ! isset( $_SERVER['HTTP_X_APPLICATION'] ) ) {
+			return false;
+		}
+
+		if ( 'varnishpass' === trim( strtolower( $_SERVER['HTTP_X_APPLICATION'] ) ) ) { // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.ValidatedSanitizedInput.MissingUnslash
+			return false;
+		}
+		return true;
```
hellofromtonya, line 50: No need to explicitly return `false` or `true` as the conditional expression returns a boolean.

```suggestion
		return ( 'varnishpass' !== trim( strtolower( $_SERVER['HTTP_X_APPLICATION'] ) ) )  // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.ValidatedSanitizedInput.MissingUnslash
```

hellofromtonya:
@arunbasillal @Davidacu This PR is ready for QA.

arunbasillal:
Thanks for the detailed report @Davidacu 

@crystinutzaa I think having the option there could create issues in the future. Some user might enable it and forget about it. 

We can detect the Varnish status reliably, right? If yes, let's please: 

- Hide the toggle
- When disabled, in place of: `Your site is hosted on Cloudways, we have enabled Varnish auto-purge for compatibility` display: `Varnish seems to be disabled on your Cloudways server. Varnish auto-purge will be automatically enabled once Varnish is enabled`. 

Regarding the message, I am [waiting on feedback](https://wp-media.slack.com/archives/CFSCRLVGB/p1593536636002500) from @webtrainingwheels to confirm.

arunbasillal:
@GeekPress For your attention. 

David noted in his [QA review](https://github.com/wp-media/wp-rocket/pull/2824#pullrequestreview-439640205): 

> Varnish addon toggle button can be manually switched to ON when varnish is disabled in cloudways - so in this case I'm not sure if this is ok, or if it makes sense notify the user that varnish is not really detected by wp-rocket and this may cause unintended behavior or error logs in the wp site?

Based on that, I suggested this: https://github.com/wp-media/wp-rocket/pull/2824#issuecomment-651924717
Please let me know if you have any feedback or ideas.

GeekPress:
@arunbasillal We are in into a normal situation, isn't it? If we don't have an auto-detection, then the add-on could be turned on.

If the add-on is turned on even if Varnish isn't activated, it won't cause issues. If Varnish is activated later on Cloudways, I suppose that WP Rocket will detect it and the toggle button will be hidden.

We have other integration like o2switch where Varnish can be enable / disable from the hosting configuration and it doesn't to generate confusions / tickets.

arunbasillal:
@GeekPress 

> If the add-on is turned on even if Varnish isn't activated, it won't cause issues

Isn't this the reason for the whole issue and PR? Mirna noted in her issue description - https://github.com/wp-media/wp-rocket/issues/2668#issue-621112912

> when the customer disables Varnish on the server-side there is no way to disable the Varnish purge from the WP Rocket side, this causes the site's error log to be filled in with messages due to the failed request.

So if the user enables the Varnish add-on again then the original issue will reappear I believe.

GeekPress:
> Isn't this the reason for the whole issue and PR? Mirna noted in her issue description - #2668 (comment)
> So if the user enables the Varnish add-on again then the original issue will reappear I believe.

The difference with the original issue is we didn't have a way to turn off the option. When it will be possible now.

arunbasillal:
@GeekPress What is the advantage of letting a user turn on the option (and cause potential issues), when we already automate it for them? We enable the add-on when Varnish is enabled and we already disable the option when Varnish is disabled. 

We can reduce one more decision for a user. Don't we have the `Decision, not option` philosophy? 

- Let's say a user installs WP Rocket on Cloudways and Varnish is disabled
- The user discovers the add-on
- The user now needs to check if their server uses Varnish or not and make a choice. 

In fact, if we can detect Varnish reliably on other hosts, we could even get rid of the whole add-on thus cleaning up WP Rocket UI even more.

GeekPress:
@arunbasillal Ok, let's go with your suggestion :)

hellofromtonya:
@arunbasillal and @GeekPress Is there more work to be done in this specific PR? Can we merge this PR and capture any new work for the option in a new issue?

arunbasillal:
@hellofromtonya These are the changes requested - https://github.com/wp-media/wp-rocket/pull/2824#issuecomment-651924717

If it doesn't take too much time, lets do it together because everything including the test environment is readily available. However if that would slow down the sprint, then sure lets plan for later 👍

hellofromtonya:
@arunbasillal Okay, moving this PR and issue back to `In Process`.

hellofromtonya:
@crystinutzaa @engahmeds3ed This PR needs changes. See @arunbasillal 's comment https://github.com/wp-media/wp-rocket/pull/2824#issuecomment-652337648

crystinutzaa:
@hellofromtonya & @engahmeds3ed can you please review the last changes?

hellofromtonya:
@wp-media/qa The requested changes have been applied. Please re-do the QA on this PR.

PR #2825, opened 2020-06-25
Fallback CSS insert load CSS code in head

crystinutzaa:
Add insert load CSS in head when fallback CSS is used

### DIFF
--- a/inc/Engine/CriticalPath/CriticalCSSSubscriber.php
+++ b/inc/Engine/CriticalPath/CriticalCSSSubscriber.php
@@ -510,7 +510,7 @@ class CriticalCSSSubscriber implements Subscriber_Interface {
 			return;
 		}
 
-		if ( ! $this->critical_css->get_current_page_critical_css() ) {
+		if ( empty( $this->critical_css->get_current_page_critical_css() ) && empty( $this->options->get( 'critical_css', '' ) ) ) {
 			return;
 		}
 


crystinutzaa:
@engahmeds3ed this method will be removed by this PR https://github.com/wp-media/wp-rocket/pull/2786/ 
I don't think we should spend time to revisit the order of the conditions in there, however you have a valid point of view related to the order of those bailout conditions.

crystinutzaa:
@arunbasillal & @Davidacu ready for QA

arunbasillal:
@engahmeds3ed Really interesting thought here, glad to see you think about this - https://github.com/wp-media/wp-rocket/pull/2825#pullrequestreview-438777225

Some feedback. 
- Constants `DONOTROCKETOPTIMIZE` and `DONOTASYNCCSS` very rarely used. In fact I have never encountered one all these years, so my theory is based on that. 
- Don't apply on wp-login.php/wp-register.php. These pages are rarely accessed. 

Shouldn't we target the most common ones first and the rare ones later? If this assumption is right, my list would be `1 > 2 > 6 > 8 > 3 > 7 > 4 > 5 > 9`. 

I love this process ❤️

crystinutzaa:
@Davidacu yes, this is a hotfix. The permanent fix will be in another PR which totally removes this function because a new method for async CSS is implemented.

hellofromtonya:
@crystinutzaa This one will be packaged up with 3.6.2. Can you change the base to `develop` please?

crystinutzaa:
@hellofromtonya base branch changed to develop.

PR #2831, opened 2020-06-29
Closes #1523 Wildcards do not work in Always Purge URL(s)

### LINKED ISSUE #1523: Wildcards do not work in Always Purge URL(s)
Here is a video recording of the issue: https://www.youtube.com/watch?v=h-uebLmqE38

To recreate: 

- Add `/category/(.*)` to Always Purge URL(s)
- Add any other non-wildcard-ed URL to the list (eg: `/sample-page/`)
- Clear the cache of a page or post
- Make sure preload is turned off
- The cache for `sample-page`is cleared but the `category` is not

Reproduced both on local and mega test site. 

Related ticket: https://secure.helpscout.net/conversation/772518070/95724?folderId=1724284
@vmanthos tested this with mixed results - https://wp-media.slack.com/archives/C08N8J6VC/p1549542729626300

crystinutzaa:
**Reproduce the issue** ✅ 
Identified the issue on my localhost.

**Identify the root cause** ✅ 
This was never working: RegEx never worked for Always purge URL(s) field though. The code is expecting standard paths.

**Scope a solution** ✅ 
I made some prototyping / discovery work and this code should work just fine:
```
function get_recursive_dir_files_by_regex( $dir, $regex ) {
	try {
		$iterator = new RecursiveIteratorIterator(
			new RecursiveDirectoryIterator( $dir )
		);
		return new RegexIterator( $iterator, $regex,  RecursiveRegexIterator::MATCH);
	} catch ( Exception $e ) {
		return [];
	}
}
$files = get_recursive_dir_files_by_regex( WP_CONTENT_DIR . '/cache/wp-rocket/wp.local/' , '/category\/(.*)\/sub-category\/(.*)/i');


```

**Effort** ✅ 
I would assume this will be an [S]. 

TO DO: @engahmeds3ed 
- [x] - Unit tests
- [x] - Integration tests

This will need both Unit and Integration tests for `rocket_get_recursive_dir_files_by_regex()` & `rocket_get_purge_urls()`

### DIFF
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -113,24 +113,20 @@ function rocket_get_purge_urls( $post_id, $post ) {
 		}
 
 		foreach ( $cache_purge_pages as $page ) {
-			if ( false !== strpos( $page, '(.*)' ) ) {
+			//check if it contains regex pattern
+			if ( strstr( $page, '*' ) ) {
 				$matches_files = _rocket_get_recursive_dir_files_by_regex( '#' . $page . '#i' );
 				foreach ( $matches_files as $file ) {
-					$pathname = $file->getPathname();
-					$filetype = $file->getType();
-					if ( ! empty( $filetype ) ) {
+					if ( ! empty( $file->getType() ) ) {
 						continue;
 					}
 
 					// Convert path to URL.
-					$home_url_slashed = untrailingslashit( $home_url ) . '/';
-					$page             = str_replace( [ rocket_get_constant( 'WP_ROCKET_CACHE_PATH' ) ], [ $home_url_slashed ], $pathname );
-					$purge_urls[] = $page;
+					$purge_urls[] = str_replace( rocket_get_constant( 'WP_ROCKET_CACHE_PATH' ), untrailingslashit( $home_url ) . '/', $file->getPathname() );
 				}
 				continue;
 			}
-			$page = trailingslashit( $home_url ) . $page;
-			$purge_urls[] = $page;
+			$purge_urls[] = trailingslashit( $home_url ) . $page;
 		}
 	}
 
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -137,11 +137,11 @@ function rocket_get_purge_urls( $post_id, $post ) {
 	$parents = get_post_ancestors( $post_id );
 	if ( (bool) $parents ) {
 		foreach ( $parents as $parent_id ) {
-			array_push( $purge_urls, get_permalink( $parent_id ) );
+			$purge_urls[] = get_permalink( $parent_id );
 		}
 	}
 
-	return $purge_urls;
+	return array_flip( array_flip( $purge_urls ) );
 }
 
 /**
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -112,13 +112,15 @@ function rocket_get_purge_urls( $post_id, $post ) {
 			restore_current_blog();
 		}
 
+		$cache_path = rocket_get_constant( 'WP_ROCKET_CACHE_PATH' ) . wp_parse_url( $home_url, PHP_URL_HOST );
+
 		foreach ( $cache_purge_pages as $page ) {
 			// Check if it contains regex pattern.
 			if ( strstr( $page, '*' ) ) {
 				$matches_files = _rocket_get_recursive_dir_files_by_regex( '#' . $page . '#i' );
 				foreach ( $matches_files as $file ) {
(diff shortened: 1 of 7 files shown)


tests/Integration/inc/common/rocketGetPurgeUrls.php
```diff
@@ -0,0 +1,163 @@
...
+
+	public function setUp() {
+		$this->set_permalink_structure( "/%postname%/" );
+		parent::setUp();
+
+		$this->create_posts( $this->config['urls']['posts'] );
+
+		$this->create_authors( $this->config['urls']['authors'] );
+
+		Functions\expect( 'get_rocket_option' )->withAnyArgs()->andReturnUsing( function( $option_name, $default=null ) {
```
crystinutzaa, line 34: @engahmeds3ed why is this needed in here? Can you please explain me? Probably I am missing something

engahmeds3ed: mocking this function that is used here:

https://github.com/wp-media/wp-rocket/blob/e5d7a6446d518f950bcbd86aa92fe03ac9276d72/inc/common/purge.php#L103

crystinutzaa: This is an integration test. We should try to bypass the mocking and set the system to run with our configuration.
For this one this should do the trick:
```
add_filter( 'pre_get_rocket_option_cache_purge_pages', return_our_value);

// RUN THE CODE	

remove_filter( 'pre_get_rocket_option_cache_purge_pages', return_our_value );
```

engahmeds3ed: Ok

crystinutzaa:
@Davidacu this is ready for QA

Tabrisrp:
There is a mistake in the code and in the tests of this PR which prevents the feature from working. The folder containing the cache files uses the domain name as its name, and it was not taken into account. I'll add this as a review.

@Davidacu Only this pattern is currently allowed as a regex in this field: `(.*)`

inc/common/purge.php
```diff
@@ -124,24 +113,31 @@ function rocket_get_purge_urls( $post_id, $post ) {
...
 
 		foreach ( $cache_purge_pages as $page ) {
-			$page = trailingslashit( $home_url ) . $page;
-			array_push( $purge_urls, $page );
+			// Check if it contains regex pattern.
+			if ( strstr( $page, '*' ) ) {
+				$matches_files = _rocket_get_recursive_dir_files_by_regex( '#' . $page . '#i' );
+				foreach ( $matches_files as $file ) {
+					// Convert path to URL.
+					$purge_urls[] = str_replace( rocket_get_constant( 'WP_ROCKET_CACHE_PATH' ), untrailingslashit( $home_url ) . '/', $file->getPath() );
```
Tabrisrp, line 121: This string replacement doesn't give the expected result:

Using `/category/(.*)`, here is the result:
```
Before: /app/public/wp-content/cache/wp-rocket/example.org/category/unpublished
After:    https://example.org/example.org/category/unpublished
```

Why?

Because `WP_ROCKET_CACHE_PATH` returns the filepath for the `cache/wp-rocket/` folder, but not the one for the current domain.

One solution is to do this before the first foreach:
`$cache_path = rocket_get_constant( 'WP_ROCKET_CACHE_PATH' ) . wp_parse_url( $home_url, PHP_URL_HOST );`

And then

`str_replace( $cache_path, untrailingslashit( $home_url ), $file->getPath() );`

tests/Fixtures/inc/common/rocketGetPurgeUrls.php
```diff
@@ -0,0 +1,186 @@
...
+			'cache' => [
+				'wp-rocket' => [
+					'test_purge' => [
+						'index.html' => ''
+					],
+					'folder_1' => [
+						'test_purge' => [
+							'index.html' => ''
+						]
+					]
```
Tabrisrp, line 17: This structure is missing one folder level, the one related to the current domain used, i.e. `example.org`

engahmeds3ed:
Checking that now

inc/common/purge.php
engahmeds3ed: Done

engahmeds3ed:
let me try it locally and back to you in a while.

Tabrisrp:
@Davidacu Ready for a new test, I pushed a change to make the code working when the home URL uses a path, like on the QA site

PR #2835, opened 2020-06-30
Fix #2242 - PDFEmbedder compatibility (Free and Paid versions)

### LINKED ISSUE #2242: Plugin Incompatibility with PDF Embedder (Free and Paid versions)
For the FREE version - https://wordpress.org/plugins/pdf-embedder/

When our **JS combine** and Remove query strings from static resources are enabled, the PDF viewer doesn’t work. It displays an error like this in the place where the PDF file should be embedded:

_Setting up fake worker failed: "Cannot load script at: https://www.example.com/wp-content/cache/busting/1/wp-content/plugins/PDFEmbedder-premium-secure/js/pdfjs/pdf.min-4.4.worker.js"._

Disabling **Remove query strings from static resources** and excluding the plugin JS files from the JS optimizations solves the issue:

`/wp-content/plugins/pdf-embedder/js/(.*).js`

This plugin also has a premium version:
**PDF Embedder Premium** - https://wp-pdf.com/premium/

In this case the same fix works, you just need to change the plugin folder:
`/wp-content/plugins/PDFEmbedder-premium-secure/js/(.*).js`

crystinutzaa:
Closes #2242 

### Scope a solution ✅ 
Create a compatibility file in `inc/ThirdParty/Plugins` for this plugin, using the code Alfonso already wrote as a base.
This was the code: http://snippi.com/s/pgsfher

We no longer need to exclude from `Remove Query Strings` since the option was removed, so it will be specifically to exclude from minify/combine JS.

### Estimate the effort ✅ 
Effort `[S]`

Needs clarification:
1. Free - `core_pdf_embedder`
2. Premium - ???? ( I used it as `pdfemb_premium_pdf_embedder` but I am not sure if this is the correct one)
3. Premium Secure - `pdfemb_commerical_pdf_embedder`

To Do:
- [x] - Unit tests - `No unit tests are required for the exclude_pdfembedder_scripts method because all of the code would be mocked. The integration tests provide complete coverage.`
- [x] - Integration tests


crystinutzaa:
Files from `wp-rocket\tests\Fixtures\inc\ThirdParty\Plugins\PDFEmbedder` are core files from both paid plugins + minimal files required for the integration test to run.

hellofromtonya (approved):

hellofromtonya:

crystinutzaa:
@arunbasillal & @Davidacu we have both paid plugins (Premium & Secure). Please let me know if you need a copy of the plugins.

arunbasillal:
@crystinutzaa Can you please upload these plugins to our shared resources folder? https://drive.google.com/drive/u/0/folders/1rY4YzgrSg0KJf878VoP-Dgs8XT_EMSVs

GeekPress:
@arunbasillal Download links are available on Bitwarden

Davidacu (changes requested):
testing environment:
http://mega.wp-rocket.me/qa


Reproduced in 3.6.1 error-> https://i.imgur.com/bTxK7WQ.png


Test against `fix/2242-pdf-embedder-compat` branch

It looks like all versions of the plugin still don't work when minify and combine js is enabled
this time we get a `ReferenceError: jQuery is not defined`
excluding `/wp-includes/js/jquery/jquery.js` fixes the issue.
https://i.imgur.com/ufZT1Gp.png, http://sc.com.ly/show/MJny5zFjiaCE314g0h8xLRVBrSoQTWfO

crystinutzaa:
This issue seems to be a bug described in here: https://github.com/wp-media/wp-rocket/issues/2846
This bug was introduced in this https://github.com/wp-media/wp-rocket/pull/2783

Davidacu (approved):
Re-tested in a new environment using local, I can confirm that all versions of the plugin work as intended when "Safe mode for JQuery" is enabled.

@crystinutzaa found the root cause of the issue reported earlier in the environment http://mega.wp-rocket.me/qa.
> This issue seems to be a bug described in here: #2846
> This bug was introduced in this #2783

so approving this PR, as #2846 is out of the scope.

PR #2857, opened 2020-07-08
Closes #2791 Automatically apply font-display: swap in CSS files

### LINKED ISSUE #2791: Automatically apply font-display: swap in CSS files

With this enhancement, WP Rocket will add “font-display: swap;” to font files mentioned in CSS files. 

<p></p>

We do the same for Google fonts when we optimize Google fonts by adding “&display=swap” to the Google font url. 

## Why are we doing this? 

When a font file is discovered in a CSS file, some browsers by default keep the text as invisible while the font is being downloaded. This invisible text is later displayed once the font download is complete. This is commonly called Flash of Invisible Text (FOIT). 

<p></p>

Adding the “font-display: swap;” property to the font will instruct the browser to use a system font while the font is being downloaded. Once the custom font download is complete, the system font is replaced with the custom font. 

<p></p>

This means users will be able to interact with the website faster without waiting for the font to download. 

## Benchmark Tests

Tested on: [<u>http://mega.wp-rocket.me/divi/</u>](http://mega.wp-rocket.me/divi/)

<p></p>

Before adding “font-display”: PageSpeed suggested this: 

<p></p>

Added “font-display: swap;” property to the font

<p></p>

After the edit, the PageSpeed suggestion was satisfied

<p></p>

WP Rocket will add the font-display property when a CSS file is minified or combined. This is safe to add and can be done without any user interaction. 

## Technical Specs

- Add the “font-display: swap;” to “@font-face” CSS rule. 

- If the font already contains a “font-display” property, we shouldn’t do anything to it. 

## Risks

Risk should be minimal and next to none. Adding font-display property is recommended practice and we shouldn’t be expecting any issues because of this enhancement. We already do this for Google fonts.

## Acceptance Criteria

- Sample @font-face rule: [<u>http://snippi.com/raw/8zod34q</u>](http://snippi.com/raw/8zod34q) (from Divi)

- When the CSS file is minified or combined, then add “font-display: swap;” to the rule. 

- If the @font-face rule already contains [<u>any font-display descriptor</u>](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-display#Syntax), then do nothing. 

- Check how the font works in the browser. Everything should be as it used to be once the page fully loads. 

<p></p>

**Edge cases:**

I couldn’t think of any edge cases. Can the Dev team think of anything? 

How should we document this? None of our [<u>current docs</u>](https://docs.wp-rocket.me/category/1231-file-optimization) are related to this. Maybe create a new doc if we get questions about this? For now, we can simply leave it as part of internal documentation.

crystinutzaa:
Closes Epic https://github.com/wp-media/wp-rocket/issues/2791


Davidacu (commented):
For this PR you can find the following:
1. [Acceptance criteria ](https://www.notion.so/wpmedia/Apply-font-display-descriptor-AC-60679da0f68b44548c4c18bf1c158343#615887f71f44447f81f71d65575653c7) for the linked EPIC 
2. The Test cases can be found [here](https://www.notion.so/wpmedia/14710a35d30c478ca9e6d8fda8eb2f0e?v=ecad62a50087434eb5faf61d90face9e)
3. [Test execution status](https://www.notion.so/wpmedia/Test-Execution-Summary-Apply-font-display-descriptor-3ca9aef714b34134948984b2c45b2439) with test evidence, 11/11 Test cases **Passed**.

**Some remarks:**
1. I need to confirm if PR #2785 should be part of the AC for PR #2857 since it's about support for combining 3rd party CSS, External CSS may have @font-face rules, and we may want to apply `font-display:swap;` there too,  but since that work is in another branch we cannot test it, @arunbasillal  could you please confirm if we should apply `font-display` to those external CSS files once WP Rocket combines them?
2. I noticed that we process existing `.min.css` files and cache them inside `wp-content/cache/min` regardless they have `@font-face` rule or not, I assume this is fine, correct ? @arunbasillal 
3. I found that it's possible to inject CSS for pages through `WP Admin -> Apperance -> Customize -> Additional CSS` , you may load some custom fonts from there with `@font-face` if desired, is this out of the scope of this PR ? currently, we don't handle such case. @arunbasillal 
4. Please feel free to take a look at the AC, and test cases, and let me know if you have any questions, feedback or if I'm missing something important.

Thank you!

crystinutzaa:
@arunbasillal can you please check the comments made by @Davidacu ?
2. Yes, all files even if they do not contain any font face will be processed and saved in /wp-content/cache/min. Should I bailout in case there is no font-face? Will this not confuse our customers why some CSS files are in the min folder and why some are not in there?

3. @Tabrisrp is this case handled in any way in WP Rocket right now?

iCaspar:
@crystinutzaa @Davidacu I think all the files should be checked/processed for `@font-face` and added to the cache (unless the user has specifically excluded them), since even previously minimized files may still need to have the `@font-face` property applied.

iCaspar:
On (3), above, I can see a case for doing it either way:
 - Pro: We should be able to cover _everything_.
 - Con: As a user, I might want a way to reliably add _my_ CSS (right or wrong) and know that nobody is going to mess with it.

Tabrisrp:
1. Most likely yes, we will want to add the property to external CSS combined by the plugin.
2. > Should I bailout in case there is no font-face? Will this not confuse our customers why some CSS files are in the min folder and why some are not in there?

Having CSS files in the min folder and some not is already the case currently, since we're ignoring `.min.css` files up to now. So I don't think this would cause any issue on the users side.

3. The customizer CSS is inserted inline, and inline CSS is left alone by WP Rocket for now, so this is expected behaviour. Also since it's manual input, they can add the property themselves, without risk of it being overridden by a theme or plugin update.

crystinutzaa:
@Tabrisrp for no 2. I need to bailout in minify() function with a `return false;` so this notice will appear for all .min.css files in our log:
https://github.com/wp-media/wp-rocket/blob/53f7e983986520c7244d5952cfc495cc0d2886f5/inc/Engine/Optimization/Minify/CSS/Minify.php#L73-L79

Will this be ok?

Tabrisrp:
That could be confusing. What if you apply the font display swap outside of the `minify()` function? That also makes more sense since it's not minify-related, it's better to separate those actions

crystinutzaa:
@Tabrisrp if I move it outside minify() function, then if I need to re-read and re-write the same file on disk, then this operation will be costly. That's why I applied it in the same operation. 
Any other ideas?

Tabrisrp:
The content is not written yet when it's returned by the `minify()` function, You could apply the swap right before the call to `write_file()` in `replace_url()`

crystinutzaa:
@Tabrisrp, @hellofromtonya, @engahmeds3ed  & @iCaspar  in here we have 2 solutions:
1. Refactor the code and change the logic in `optimize()` and `replace_url()` to accomodate the bailout for `.min.css` files which do not contain the font-face.
2. Do not bailout and move the `min.css` file in `cache/min` folder even if we do not alter the file at all. 

Each solution has it own pros and cons.
1. Pro: we will add complexity to the code even if we refactor and split it in multiple smaller functions. 
Cons: Only files which are altered by WP Rocket will be in `/cache/min` folder. So questions may arise why some CSS files were not moved ( however this was the old functionality )
2. Pro: works as is now. No code changes required.
Cons: There will be duplicated identifical files (in WP and also in `/cache/min` folder)

What do you think? Which is the most suitable?

arunbasillal:
Great feedback so far, nice job @Davidacu 

1. Any CSS file that we process can benefit from this. So for both internal and external 👍 
2. I prefer the bail out. If the file is not needed, then let's not create it. In the end, it's less files to create. 

3. For inline CSS, do we actually see this happening? It should be really rare and as Remy mentioned, customers can add this themselves. If we do see a lot of this, we can think about doing this as a new enhancement. 
^ This is assuming that we have to do a lot more work to make this happen right now. If it's not so much effort, let's do this as well and cover all cases? cc: @crystinutzaa, @Tabrisrp 

> Please feel free to take a look at the AC, and test cases, and let me know if you have any questions, feedback or if I'm missing something important.

It looks like you have covered plenty of cases. Couldn't think of anything else at the moment. Very detailed 🙌

crystinutzaa:
@Tabrisrp I refactored the code and bailout if min.css file does not contain any font-face. Can you please review the code before resubmitting to QA?

remyperona (approved):

crystinutzaa:
@Davidacu this is ready for QA again.
The new changes are related to min.css files. If the min.css files do not contain font-face it will not affect the file and will not be created in cache folder.

Davidacu (changes requested):
- I can confirm that now we don't generate a file if the `.min.css` file doesn't contain `@font-face` rule.
this check is now included in one of the test cases.
- I also confirmed that now we apply `font-display:swap;` to the combined external CSS.

12/12 [Test cases](https://docs.google.com/spreadsheets/d/1t0vAmmIRJ4RN_tjS7EPQuL5BNJju1UjUppn5LmWhdlk/edit#gid=180417299&range=A2) passed, everything is looking good, however, there is something important I may have missed, it seems that when the CSS files are combined, the original files are combined and not the minified versions, I noticed that the combined CSS contains comments and they belong to both internal and external CSS files, when comparing to the master version of the plugin, indeed these comments are stripped.

**To reproduce:**
1. Navigate to http://mega.wp-rocket.me/qa/
2. in the page source, look for the combined file and open

you'll notice there are comments in the combined file. https://youtu.be/w6WY48vCFqU
image

Tabrisrp:
I understand what is happening there: The file containing this code uses this filename `example.min.css`

In the combine 3rd party CSS PR, an update to the code was made to check for the filename, and if it contains `.min.css`, it's being combined but doesn't go through the minification process. We use this as a hint that the file is already minified, and preserve processing time & resources.

In this case, the file is using the suffix, but the content is not _actually_ minified.

As a test, I updated the filename to remove the `.min` part, and now it's being minified as expected.

I would consider this as an edge case, as the suffix should not be used if the content is not minified.

Davidacu:
>if it contains .min.css, it's being combined but doesn't go through the minification process.

Right! I do see that for external CSS we don't minify them regardless of the `.min.css` ending
they are combined as is, is it intended? @Tabrisrp

Tabrisrp:
Do you have an example for that? They should be minified if they don't use the `.min` filename.

To note, some comments will always be preserved, notably if the start with `/*!`, or contain `@license` for example. But the CSS itself should still be minified.

Davidacu:
Ok I see, yes it looks like it's actually removing spaces, but not the comment.
you can check it in http://mega.wp-rocket.me/qa/wp-content/cache/min/1/e07fbf015b1e40717feaa20324f4918f.css
the line that starts with `* Styles from sk8pedia`  are from an external URL https://peaceful-kepler-839115.netlify.app/sk8pedia_styles.css
but yes the comments do indeed start with `/*!` so that seems all right.

Tabrisrp:
Yes in this case it's expected. The content is minified, but the comment is preserved because it starts with `/*!`

Davidacu (approved):

PR #2860, opened 2020-07-08
Improve activation/deactivation processes

Tabrisrp:
- Add an `Activation` and a `Deactivation` class, each with a static method to serve as the 2nd argument callback for the `register_(de)activation_hook` functions, replacing `rocket_activation()` and `rocket_deactivation()`
- Update the capabilities manager to the new architecture
- Update the `WPCache` class to add new methods to handle the `activation` and `deactivation` events
- Update the `AdvancedCache` class to add new methods to handle the `activation` and `deactivation` events


inc/Engine/Deactivation/Deactivation.php
```diff
@@ -0,0 +1,98 @@
+<?php
+
+namespace WP_Rocket\Engine\Deactivation;
+
+use League\Container\Container;
+use WP_Rocket\Engine\Cache\AdvancedCache;
+
+class Deactivation {
```
iCaspar, line 8: Should implement `DeactivationInterface` here.

inc/Engine/Deactivation/DeactivationInterface.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace WP_Rocket\Engine\Deactivation;
+
+interface DeactivationInterface {
+	/**
+	 * Executes this method on plugin deactivation
+	 *
+	 * @return void
+	 */
+	public function deactivate();
```
iCaspar, line 11: Add `static`

inc/Engine/Activation/ActivationInterface.php
```diff
@@ -0,0 +1,12 @@
...
+
+namespace WP_Rocket\Engine\Activation;
+
+interface ActivationInterface {
+	/**
+	 * Executes this method on plugin activation
+	 *
+	 * @return void
+	 */
+	public function activate();
```
iCaspar, line 11: Add `static`

tests/Fixtures/inc/Engine/Cache/WPCache/maybePreventDeactivation.php
```diff
@@ -0,0 +1,11 @@
...
+
+return [
+	'vfs_dir' => 'public/',
+
+	'test_data' => [
+		[
+
+		]
+	]
+];
```
iCaspar, line 11: @Tabrisrp @hellofromtonya Is there a better way to do this? We don't really need any testdata for this one, but we do need to use the filesystem, so we're extending the FilesystemTestCase (which seems to require the fixture to establish the `vfs_dir`.

Tabrisrp: In your test file, you can group up the different testing methods into one, and use the data provided from here to change the way the test behave. For example:

```
testShouldBailOutWhenConfigFileFound => [
    'expected' => [],
],
testShouldBailOutWhenSetCacheConstFilterFalse => [
    'set_constant' => false,
    'expected'       => [],
],
testShouldAddCauseToCausesWhenNotPrevented => [
    'file_exist'   => false,
    'expected'  => [ 'config' ],
],
```

You can check the existence of the array keys to perform different actions in the test code and get the expected result

tests/Integration/inc/Engine/Cache/AdvancedCache/updateAdvancedCache.php
```diff
@@ -0,0 +1,49 @@
+<?php
+namespace WP_Rocket\Tests\Integration\inc\Engine\Cache\AdvancedCache;
+
+use WP_Rocket\Engine\Cache\AdvancedCache;
+use WP_Rocket\Tests\Integration\FilesystemTestCase;
+use Brain\Monkey\Filters;
```
Tabrisrp, line 6:
```suggestion
```

tests/Integration/inc/Engine/Cache/AdvancedCache/updateAdvancedCache.php
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * @covers \WP_Rocket\Engine\Cache\AdvancedCache::update_advanced_cache
+ * @uses   ::rocket_get_filesystem_perms
+ *
+ * @group  AdvancedCache
+ */
+class Test_UpdateAdvancedCache extends FilesystemTestCase {
+	protected $path_to_test_data = '/inc/Engine/Cache/AdvancedCache/updateAdvancedCache.php';
+
+	public function testShouldBailOutWhenShortCircuitFilterReturnsFalse() {
```
Tabrisrp, line 17: You can add this test to the test_data too, the same way as described in my other comment.

For the filter, you can use a config value from the data provider, and do the following:
```
if ( false === $config['generate'] ) {
    add_filter( 'rocket_generate_advanced_cache_file', [ $this, 'return_false' ] );
}
```

We have custom methods for tests to return common values like this. Also since we're adding the filter in the test, we want to remove it on the `tearDown()` method to make sure it doesn't interfere with other tests

```
public function tearDown() {
    parent::tearDown();
    remove_filter( 'rocket_generate_advanced_cache_file', [ $this, 'return_false' ] );
}
```

iCaspar:
Ready for QA!

Tabrisrp:
@wp-media/qa 

It contains internal changes, so there is no new functionalities. Though we want to check there is no regression, specifically:

### Activation process
- The `advanced-cache.php` file is correctly generated, except if the `rocket_generate_advanced_cache_file` filter is set to `false`
- The `WP_CACHE` constant is set to `true` in `wp-config.php`, except if the `rocket_set_wp_cache_constant` filter is set to `false`
- WP Rocket custom capabilities are correctly set for the `administrator` role (WP Rocket settings page is accessible)

### Deactivation process
- The `advanced-cache.php` file is correctly emptied, except if the `rocket_generate_advanced_cache_file` filter is set to `false`
- The `WP_CACHE` constant is set to `false` in `wp-config.php`, except if the `rocket_set_wp_cache_constant` filter is set to `false`
- The deactivation process is stopped if `wp-config.php` is not writable, except if the `rocket_set_wp_cache_constant` filter is set to `false`

Also:
- the `htaccess` file should be updated on activation/deactivation
- the `wp-content/cache` and `wp-content/wp-rocket-config` folders should be created on activation (if they don't exist yet)
- Config files in `wp-content/wp-rocket-config/` should be deleted on deactivation

Davidacu:
As there are not any new features that end-users will notice, we agreed on the AC for this PR is
- Validate that the activation/deactivation process works as before.
- There aren't any defects introduced.

For the testing results reported bellow, I have included the scope of PR #2874 , so far I have covered Cloudways, and WP Engine, the remaining 2 hosting (SpinupWP and Pressable) will be done once access is provided, we're waiting for those.

**WP Engine** - [Test results](https://docs.google.com/document/d/1QHlaHSqvgyB1MWf_GuUfo_EjThZNogIk9cPfSxz9V94/edit?usp=sharing)

All cases passed, except I notice this:
- The `WP_CACHE` constant is set to 1 during activation in `wp-config.php`
- When `rocket_set_wp_cache_constant` filter is set to false it doesn't cause any effect on `WP_CACHE` on activation, the constant is always set to 1 no matter what. (looks like it's the same behavior on previous versions of WP Rocket, 3.6.1 for instance)
Is this because WP Engine only caching plugin allowed is WP Rocket? @Tabrisrp 
therefore not updating this constant probably doesn't cause any issues.

**Cloudways** - [Test results](https://docs.google.com/document/d/1TYYA-RjE-DxYmxqcaGy47pWM9NaEVFpLcT_0tuav7aI/edit?usp=sharing)

All cases passed.

Tabrisrp:
@Davidacu Good catch about `WP_CACHE` on WPEngine. Previously we forced the value to `true` all the time because our optimizations were depending on it, but it's no longer the case. So I updated our code to not write in `wp-config.php`at all, since we don't care about it (WPEngine handles the cache part).

Davidacu:
Hi @jbma @arunbasillal @Tabrisrp 
Any chance you could please help with getting access to SpinupWP and Pressable?
@jbma did you hear anything from SpinupWP support?

jbma:
@Davidacu About SpinupWP that's because the server on DigitalOcean has been removed.
@arunbasillal Can you spawn the server or show David how to do it?

GeekPress:
@Davidacu SpinUpWP and Pressable issues has been fixed.

PR #2869, opened 2020-07-09
Exclude Phone Wagon script from combine JS

ScottTravisHartley:
Excludes phone wagon JS from being merged to avoid the following error. 
Uncaught Error: PhoneWagon: pw_dns.js <script> not found.

example script: https://beta.phonewagon.com/DynamicNumbers/7c4e4b04-faa1-4798-af67-c13350ea015c/88/9/pw_dns.js


inc/Engine/Optimization/Minify/JS/Combine.php
```diff
@@ -729,6 +729,7 @@ protected function get_excluded_external_file_path() {
 			'static.zdassets.com',
 			'feedbackcompany.com/widgets/feedback-company-widget.min.js',
 			'widget.gleamjs.io',
+			'beta.phonewagon.com',
```
Tabrisrp, line 732: is there a possibility that the domain changes at some point? The `beta` subdomain raises doubt, maybe excluding the top domain instead would be safer?

ScottTravisHartley:
That should be fine it's been in "beta" for quite a long time but the full domain I could see being more future proof.

PR #2874, opened 2020-07-10
Add a factory for hosting subscribers

iCaspar:
This builds on the new Subscriber architecture for ThirdParty hosting.

Currently, we instantiate and register Subscribers into the container for every host, even if we don't need any special host; and the most we would ever need is to subscribe for a single host. This is both a performance hit and an unnecessary use of memory.

What we do here is to add a `HostSubscriberFactory` that knows how to figure out if we need to subscribe a host, and if so which one, and allows us to instantiate the only the correct one at runtime.

We have also implemented a null object pattern into the hosting subscribers, to guarantee a consistent return object type when invoking the factory's `get_subscriber()` method.

To implement, we have also done the following refactorings:
- The 4 current hosting subscribers are refactored to remove checks that are now performed in a single place by the `HostSubscriberFactory`. This gives us a single point in the codebase to hold the "Which host are we on" concern.
- The `ThirdParty/ServiceProvider` is refactored to build the correct Subscriber, only when we need one, and register it into the container, and
- The `Plugin.php::init_common_subscribers()` is refactored to query the `HostSubscriberFactory` about which tags it should be aware of, passing only the relevant single hosting subscriber into the `$common_subscribers` array.


tests/Unit/inc/ThirdParty/Hostings/HostResolver/getHostService.php
```diff
@@ -0,0 +1,52 @@
...
+	 * @dataProvider configTestData
+	 */
+	public function testShouldReturnExpectedValue( $expected ) {
+		switch( $expected ) {
+			case 'cloudways':
+				$_SERVER['cw_allowed_ip'] = true;
+				break;
+			case 'pressable':
+				Functions\when( 'rocket_get_constant' )
+					->justReturn( true );
```
hellofromtonya, line 34: Need to qualify this with which  constant to get. Else, it'll override all mocks.

inc/Plugin.php
```diff
@@ -254,12 +255,15 @@ private function init_common_subscribers() {
...
 			'simple_custom_css',
-			'cloudways',
-			'wpengine',
-			'spinupwp',
 			'pdfembedder',
 		];
 
+		$host_type = HostResolver::get_host_service();
+
+		if ( $host_type ) {
```
hellofromtonya, line 263: The data type is a string. Let's be more strict with our data type checks. Use `empty()` or `'' !== $host_type` here to avoid type juggling of a boolean check.

inc/ThirdParty/Hostings/HostResolver.php
```diff
@@ -0,0 +1,66 @@
...
+	 *
+	 * @since 3.6.3
+	 *
+	 * @param bool $ignore_cached_hostname (optional) Don't use cached hostname when true.
+	 *
+	 * @return string Name of the hosting service or '' if no service is recognized.
+	 */
+	public static function get_host_service( $ignore_cached_hostname = false ) {
+
+		if ( ! $ignore_cached_hostname && self::$hostname ) {
```
hellofromtonya, line 30: The hostname is a string. Avoid boolean type juggling by checking it either via `empty()` or `'' !== self::$hostname` or even adding a `has` method.

inc/ThirdParty/Hostings/HostResolver.php
```diff
@@ -0,0 +1,66 @@
...
+			return self::$hostname;
```
engahmeds3ed, line 31: I think using cache with long lifetime to save this hostname value will help a lot here.

tests/Unit/inc/ThirdParty/Hostings/HostResolver/getHostService.php
Tabrisrp: Done in https://github.com/wp-media/wp-rocket/pull/2874/commits/ffc56472c788b4d609fe3ef753af420a16d3ac44

inc/Plugin.php
Tabrisrp: Done in https://github.com/wp-media/wp-rocket/pull/2874/commits/9a9a5970182e068a13e9703d1860ef151813e05a

inc/ThirdParty/Hostings/HostResolver.php
Tabrisrp: Done in https://github.com/wp-media/wp-rocket/pull/2874/commits/9a9a5970182e068a13e9703d1860ef151813e05a

inc/ThirdParty/Hostings/HostResolver.php
Tabrisrp: I don't think we need a longer cache than the static variable for the duration of the current request

Tabrisrp:
@wp-media/qa This is ready for QA

This is mainly an internal change, as in term of functionalities everything is still the same. I guess we would want to give this a test run it on Cloudways, SpinUpWP & WPEngine to make sure the compatibilities classes are correctly loaded as expected on there?

inc/ThirdParty/Hostings/HostResolver.php
engahmeds3ed: Ok, I thought that using cache here will be over all requests longer time and the cache is already related to the hosting itself so if customer changed the hosting company this cache won't be found.
it's only idea :D

iCaspar: So, yeah, I included this as an internal short-circuit -- mainly to make unit testing easier on a static value. It's probably poorly named, since it's not really a cached value. It's probably poorly named, since it's not really a cached value.
Why use a static value or store it at all?
Currently, this is only checking for 4 hosts, but as we move more hosts into the new architecture this will become more and more expensive as we have more and more possibilities to evaluate. While not overwhelmingly expensive, there are several places in our codebase that will call it, so it seems best not to expend all the resources every time. That said, a real cached value is probably overkill.

Tabrisrp:
QA for this will be done on PR #2860 QA, as it's the parent branch for this one

PR #2888, opened 2020-07-14
Silently run phpstan in CI

szepeviktor:
Everything is taken care of.

Now it is possible to see what is up in the codebase. 👀 
Everyone is encouraged to contribute!


szepeviktor:
kép

Tabrisrp:
Thank you for that, that will definitely be useful!

szepeviktor:
You're welcome.

Somehow it does not start + does not fail. Strange.

PR #2895, opened 2020-07-16
Revert PR #2786 Revert using DOMDocument for async CSS feature parsing

Tabrisrp:


inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -575,20 +633,78 @@ public function exclude_inline_js( array $excluded_inline ) {
...
 	 * @return string Updated HTML code
 	 */
-	public function async_css( $html ) {
-		$instance = AsyncCSS::from_html( $this->critical_css, $this->options, $html );
-		if ( ! $instance instanceof AsyncCSS ) {
-			return $html;
+	public function async_css( $buffer ) {
+		if ( rocket_get_constant( 'DONOTROCKETOPTIMIZE' ) ) {
+			return;
+		}
```
iCaspar, line 645: Should this be `return $buffer;`? Not sure.

Tabrisrp: yes it should you're right https://github.com/wp-media/wp-rocket/pull/2895/commits/9317abe8d97f163c02fffe768baf885e561b6d2e

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -488,6 +491,61 @@ public function warning_critical_css_dir_permissions() {
...
+		// Don't apply on search page.
+		if ( is_search() && ! $rocket_cache_search ) {
+			return;
+		}
+
+		// Don't apply on 404 page.
+		if ( is_404() ) {
+			return;
+		}
+
```
engahmeds3ed, line 533:
```suggestion
//group posts related conditions
		if (
			(
				is_search()
				&&
				! apply_filters( 'rocket_cache_search', false )
			)
			||
			is_404()
			||
			is_rocket_post_excluded_option( 'async_css' )
		){
			return;
		}

		if (
			empty( $this->critical_css->get_current_page_critical_css() )
			&&
			empty( $this->options->get( 'critical_css', '' ) )
		){
			return;
		}
```
what about grouping conditions related to posts?

Tabrisrp: `is_search()` is going to be the search results page, and `is_404()` can be anything, so they are not conditions tied to a post specifically.

iCaspar: You could almost extract _all_ these checks into their own private method.

engahmeds3ed: @Tabrisrp Ok, we can reorder the conditions to put `get_current_page_critical_css` at the end as this is using the filesystem so it will be almost slower than checking database for example.
Also like @iCaspar stated, I can see that these conditions or most of them are used on `insert_critical_css_buffer` buffer.
be calm with our code review we are studying the code on that :D

Tabrisrp: Good idea for the private method, done here: https://github.com/wp-media/wp-rocket/pull/2895/commits/5415787628da77aa59ee73a3c69e8a8635c2e15e

Tabrisrp:
@wp-media/qa 

For the QA, everything should be working as on 3.6.1 related to Optimize CSS delivery, while:
- Preserving the fixes for Firefox for which there was FOUC in some cases
- Preserving the removal of the `id` attribute in the styles added in the `<noscript>` tags at the end of the HTML, to prevent w3c validator warnings

Davidacu:
1. New Load async CSS method is indeed reverted, the old method is now being used.
2. I can confirm that the firefox fixes are included, didn't notice any FOUC or broken layouts.
3. There is no issue with duplicates ids, the id of the stylesheets only appear at the bottom in the `<noscript>`

Smoke/Regression tests.
**Optimize CSS delivery**
1. CPCSS is generated
2. Fallback CPCSS works (Chrome, FF, Safari, Opera)
3. CSS is prioritized with preload and loaded asynchronously.
4. Generate specific CPCSS for page from page editor.
5. Exclude page from OCD by disabling option from page-editor.

**File Optimization**
1. Minify CSS/HTML/JS
2. Combine CSS/JS
3. Load JS deferred

**Lazy loading**
1. Images are lazyloaded as expected.
2. No issues with encoded characters in `alt` or background image url.

Looking good.

Issue #2899, opened 2020-07-20
Delay JS execution time (LazyLoad Script)


GeekPress:
## What are we proposing to do? 

With this feature WP Rocket will delay the execution of JavaScript files. This will essentially be a LazyLoad for JavaScript.

**Note**: We once planned a similar feature, but only for third party scripts (#[<u>2581</u>](https://github.com/wp-media/wp-rocket/issues/2581)). With this feature we are extending this to both internal and external JS files. 

## Why are we doing this? 

Not all scripts are relevant for the initial page render and delaying them can improve the loading time. For example, a script useful for an “Exit Popup” is not relevant during initial page load and can be delayed. This can have huge performance gains as evident from the Benchmark tests. 

As a side effect, delaying the execution of JavaScript will help improve PageSpeed score and speed test results. 

<p></p>

**Note**: We decided not to go ahead with #2581 because we thought of a better way to do it. However, that approach will take some development time and we are hoping the current approach will serve as an intermediate solution for our customers. 

## Benchmark Tests

Jonathan has been using this feature on our own website for a while with impressive results. We also did detailed benchmarking for #[<u>2581</u>](https://github.com/wp-media/wp-rocket/issues/2581) and are available in the issue. 

## User expectations and experience. How will we do this feature?

We are going to keep it simple and lean for our first iteration.   

- We will do it with a simple checkbox to enable the option. 

- The whole feature will be opt-in. Meaning that only the scripts that the users have selected will be delayed. 

- Enabling the option will reveal a text area where users can enter the scripts that they wish to delay. 

- We will pre-populate this textarea with some scripts that we have identified as safe to delay. Example: Facebook Social Plugins.

- Users can add their own scripts (both internal and external) to the list. They can also specify a keyword from the inline JS to be delayed. 

- If any of the scripts we include by default causes issues, users can remove them from the list as well. 

- A link « Restore default settings » will restore the default list if the user clicks on it. This will provide users an easy way to restore the default list. 

<p></p>

**Updating the default list as we discover more scripts that can be safely delayed**

<p></p>

- For now, let’s not do an automatic update for existing users. We can update our documentation and maintain a public list. Users can copy from our public list when needed. After they set-up the page initially, it wouldn’t make sense to modify the list. 

- Once we have more feedback and also see how we can handle this feature in other ways, we can work on an update strategy. Let’s keep this lean for now. 

- The updated list can be included for new users.

[...]

PR #2913, opened 2020-07-22
Delay JS execution time: admin part (1/3) 

crystinutzaa:
Sub-task for epic #2899 

### Update `Admin\Settings\Page`
- Update the `assets_section()` to add 2 new fields:
    - A checkbox for the option itself
    - A textarea for the inclusion list

We also need to add the restore defaults button, which will trigger an AJAX request to replace the textarea content with the default values.

### New `Optimization\DelayJS\Admin\Settings`
- a property `$defaults` containing the default values
- a method `add_option()` to add the `delay_js` option to WP Rocket options array. Default value is `1`. Also the `delay_js_include` option. Default value is the `$defaults` array property value.
- a method `set_option_on_update()` to set the default value to `0` on plugin update to 3.7 for existing users
- a method `restore_defaults()`, to serve as the callback for the AJAX request to restore the default content
- Write tests

### New `Optimization\DelayJS\Admin\Subscriber`
- inject the above class as a dependency
- Add the callbacks methods for each event:
    - `rocket_first_install_options`
    - `wp_rocket_upgrade`
    - `wp_ajax_restore_default_delay_js`

### New `Optimization\DelayJS\ServiceProvider`
- Instantiate the 2 classes above, only in the admin context

### Update `Admin\Deactivation\DeactivationIntent`
- Add `delay_js` to the `$reset_options` array in `activate_safe_mode()`
- Update tests

### Update `src/js/global/ajax.js`
- Add the script to trigger the AJAX request when clicking on restore defaults

### Update `Admin\Beacon\Beacon`
- Add `delay_js` to the `$options_to_send` array in `session_data()`


inc/Engine/Admin/Settings/Page.php
```diff
@@ -826,6 +826,34 @@ private function assets_section() {
 					'default'           => 1,
 					'sanitize_callback' => 'sanitize_checkbox',
 				],
+				'delay_js'           => [
+					'container_class'   => [
+						'wpr-isParent',
+					],
+					'type'              => 'checkbox',
+					'label'             => __( 'Delay JavaScript execution', 'rocket' ),
+					'description'       => sprintf( __( 'Improves initial loading time by delaying the loading of JavaScript files.', 'rocket' ) ),
```
Tabrisrp, line 835:
```suggestion
					'description'       => __( 'Improves initial loading time by delaying the loading of JavaScript files.', 'rocket' ),
```

The `sprintf` is not needed in this case

inc/Engine/Admin/Settings/Page.php
```diff
@@ -826,6 +826,34 @@ private function assets_section() {
...
+					'section'           => 'js',
+					'page'              => 'file_optimization',
+					'default'           => 0,
+					'sanitize_callback' => 'sanitize_checkbox',
+				],
+				'delay_js_scripts'      => [
+					'type'              => 'textarea',
+					'label'             => __( 'Scripts to delay', 'rocket' ),
+					'description'       => sprintf( __( 'Specify keywords that can identify inline or JavaScript files to be delayed (one per line).', 'rocket' ) ),
```
Tabrisrp, line 844:
```suggestion
					'description'       => __( 'Specify keywords that can identify inline or JavaScript files to be delayed (one per line).', 'rocket' ),
```

Same here

tests/Fixtures/inc/Engine/Optimization/DelayJs/Admin/Subscriber/setOptionOnUpdate.php
```diff
@@ -1,15 +1,20 @@
...
-	'ShouldUpdateOptionWithVersionAbove3_7' => [
-		'old_version' => 3.9,
-		'valid_version' => false
-	],
-	'ShouldNotUpdateOptionWithVersionBelow3_7' => [
-		'old_version' => 3.5,
-		'valid_version' => true
+	'settings' => [
+		'delay_js' => 1,
 	],
```
Tabrisrp: @engahmeds3ed The `settings` key in  the fixtures can be used to set initial value for the WPR options array when needed, thanks to the `SettingsTrait`. For example in this test, we expect the `delay_js` option to already be set, and to change to zero if it corresponds to the version, so it's good to have the options set.

tests/Integration/inc/Engine/Optimization/DelayJs/Admin/Subscriber/restoreDefaults.php
```diff
@@ -1,33 +1,40 @@
...
+		CapTrait::hasAdminCapBeforeClass();
+		CapTrait::setAdminCap();
+	}
+
+	public function tearDown() {
+		parent::tearDown();
+
+		CapTrait::resetAdminCap();
+	}
+
```
Tabrisrp, line 30: @engahmeds3ed We can use the setup and teardown methods to handle the capacity setup, and make sure it's correctly reset between & after tests

tests/Integration/inc/Engine/Optimization/DelayJs/Admin/Subscriber/setOptionOnUpdate.php
```diff
@@ -1,45 +1,33 @@
...
 	 * @dataProvider configTestData
 	 */
-	public function testShouldDoExpected( $old_version, $valid_version ){
-
+	public function testShouldDoExpected( $old_version, $valid_version ) {
 		do_action( 'wp_rocket_upgrade', '', $old_version );
 
-		$options      = new \WP_Rocket\Admin\Options( 'wp_rocket_' );
-		$option_array = new Options_Data( $options->get( 'settings' ) );
+		$options = get_option( 'wp_rocket_settings' );
```
Tabrisrp, line 19: @engahmeds3ed We can use the `get_option()` function directly to get the option value:
- It makes the test cleaner
- It also integrates with WordPress

tests/Integration/inc/Engine/Optimization/DelayJs/Admin/Subscriber/setOptionOnUpdate.php
```diff
@@ -1,45 +1,33 @@
...
+			$this->assertSame(
+				0,
+				$options['delay_js']
+			);
 		} else {
-			$this->assertNull( $option_array->get('delay_js', null) );
+			$this->assertSame(
+				1,
+				$options['delay_js']
+			);
```
Tabrisrp: @engahmeds3ed I updated this after updating the fixture to use the `SettingsTrait`, to check for the value of the option key which should always be set

tests/Fixtures/inc/Engine/Optimization/DelayJs/Admin/Settings/restoreDefaults.php
```diff
@@ -1,21 +1,11 @@
...
-			]
-		],
-		'restored' => true
+	'testShouldNotRestoreDefaultsWhenNoCapabilities' => [
+		'capability' => false,
+		'restored'   => false,
+	],
+	'testShouldRestoreDefaultsWhenCapabilities'      => [
+		'capability' => true,
+		'restored'   => '',
```
Tabrisrp, line 9: @engahmeds3ed Simplified the fixture a bit, we know the capability we want to check, so we can mock only the returned value. We also can use the `restored` key to return the expected value

tests/Unit/inc/Engine/Optimization/DelayJs/Admin/Settings/restoreDefaults.php
```diff
@@ -1,45 +1,35 @@
...
-		if ( $capabilities ) {
-			foreach ( $capabilities as $capability ){
-				Functions\expect( 'current_user_can' )->with( $capability )->andReturn( true );
-			}
-		} else {
-			Functions\when( 'current_user_can' )->justReturn( false );
-		}
+		Functions\expect( 'current_user_can' )
+			->with( 'rocket_manage_options' )
+			->andReturn( $capability );
```
Tabrisrp, line 25: @engahmeds3ed Following the change in the fixtures, this simplifies the test too, where we can directly handle the expected return in one method

tests/Unit/inc/Engine/Optimization/DelayJs/Admin/Subscriber/addOptions.php
```diff
@@ -1,32 +0,0 @@
-<?php
```
Tabrisrp: @engahmeds3ed Removed this test as it's not necessary, the method in the subscriber is directly calling the dependency method with no processing

PR #2951, opened 2020-07-27
Preserve WP_CACHE & advanced-cache.php if on multisite with WPR active on other sites

remyperona:
When on a multisite configuration, all sites share the same `wp-config.php` & `advanced-cache.php` files.

When deactivating WP Rocket on a sub-site, the `WP_CACHE` value & the `advanced-cache.php` content should be preserved, except if this site was the last one with WP Rocket active. We check this on deactivation by looking at the number of config files found in `wp-content/wp-rocket-config`.

When adding the new deactivation process, this behaviour was not preserved (oversight on my part). This PR is fixing that, reverting it to the intended, while still using the new process.

As comparison, here is how the code looked on version 3.6.2.1: https://github.com/wp-media/wp-rocket/blob/b1fdd13fbe78c8fc54f26a586061fd79b9a1966b/inc/main.php#L154

### Acceptance criteria
On a _single site_, the `WP_CACHE` value & advanced-cache.php content should be updated respectively to `false` & an empty content when deactivating WP Rocket.

On a _multisite_, the following should happen:
- If WP Rocket is still active on another site of the network, the value & content should not be touched
- If the site where WP Rocket is deactivated is the last one with the plugin active, the same behaviour as for a single site is expected

### DIFF
--- a/inc/Engine/Cache/WPCache.php
+++ b/inc/Engine/Cache/WPCache.php
@@ -47,7 +47,7 @@ class WPCache implements ActivationInterface, DeactivationInterface {
 	 * @since 3.6.3
 	 *
 	 * @param int $sites_number Number of WP Rocket config files found.
-	 * @return bool
+	 * @return void
 	 */
 	public function update_wp_cache( $sites_number = 0 ) {
 		if ( ! rocket_valid_key() ) {
--- a/inc/Engine/Cache/WPCache.php
+++ b/inc/Engine/Cache/WPCache.php
@@ -34,14 +34,9 @@ class WPCache implements ActivationInterface, DeactivationInterface {
 	/**
 	 * Performs these actions during the plugin deactivation
 	 *
-	 * @param int $sites_number Number of WP Rocket config files found.
 	 * @return void
 	 */
-	public function deactivate( $sites_number = 0 ) {
-		if ( is_multisite() && 0 !== $sites_number ) {
-			return;
-		}
-
+	public function deactivate() {
 		add_action( 'rocket_deactivation', [ $this, 'update_wp_cache' ] );
 		add_filter( 'rocket_prevent_deactivation', [ $this, 'maybe_prevent_deactivation' ] );
 	}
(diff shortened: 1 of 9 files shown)


iCaspar (approved):

Tabrisrp:
@wp-media/qa Ready for QA, acceptance criteria in the opening comment

Davidacu (commented):
There seems to be something wrong with multisite installations, for what I have noticed, the behavior in `3.6.2.1` is the expected behavior described in this issue, and the behavior in `fix/multisite-deactivation` seems to be the exact problem described in this ticket.

**Testing Environment**: local

**branch**: `3.6.2.1`

Could not reproduce the issue with deactivation
- when deactivating the plugin on a subsite and other subsites have the plugin activated,
  nothing is written to wp-config.php nor advanced-cache.php
- when activating the plugin on a subsite, the files that seem to be modified are advanced-cache.php and .htaccess

**branch**: `fix/multisite-deactivation`

The issue with deactivation is reproduced.
- when deactivating the plugin on a subsite, and other subsites have the plugin activated,
  the WP_CACHE=false is written to wp-config.php.
  then if you go to plugins section of a subsite that has the plugin activated the value of WP_CACHE=true is written to wp-config.php
  https://screencast-o-matic.com/watch/cYiuckEgit
- when activating the plugin on a subsite, `wp-config.php` , `advanced-cache.php` and `.htaccess` are modified.

Tabrisrp:
@iCaspar requesting a new code review, because I realized I didn't modify the correct methods, and the code was not working.

inc/Engine/Cache/WPCache.php
```diff
@@ -46,14 +46,23 @@ public function deactivate() {
 	 *
 	 * @since 3.6.3
 	 *
-	 * @return bool|void False if rocket key is invalid.
+	 * @param int $sites_number Number of WP Rocket config files found.
+	 * @return bool
```
iCaspar, line 50: We are actually returning void here.

iCaspar (changes requested):

Tabrisrp:
@Davidacu This is ready for QA again, sorry about the first one

Davidacu (commented):
Find the test results attached.  
[testrail-report-multisite-activation-deactivation.pdf](https://github.com/wp-media/wp-rocket/files/4991954/testrail-report-multisite-activation-deactivation.pdf)
3/4 cases **Passed**
1/4 cases **Failed**
There is one failed case, which actually I want to confirm if it should be a failure or not.
The failed case is T14, I would expect that if activating the plugin on Site 1, when I activate the plugin on Site 2 it should not write to `.htacess` or `advanced-cache.php` but it does, if this is not really an issue then everything is looking good!

image

Tabrisrp:
@Davidacu That is expected, it's the same behaviour on the current production version. It's not ideal, but it's also out of scope of this PR, so it's ok

PR #2959, opened 2020-07-30
Closes #2810: Minify external CSS and JS files

### LINKED ISSUE #2810: Host external CSS and JS files locally when minification is active

WP Rocket already combines externally hosted JS files into the combined JS file. And with the recent enhancement #[<u>2689</u>](https://github.com/wp-media/wp-rocket/issues/2689), externally hosted CSS files are also combined into the combined CSS file. 

<p></p>

However, if only minification of CSS or JS is enabled, the external CSS and JS files are not hosted locally. 

<p></p>

With this enhancement, WP Rocket will download and cache the external CSS and JS files on the same server and serve them. 

## Why are we doing this? 

Hosting external CSS and JS files locally can have the following advantages: 

- Reduce the number of DNS resolutions needed. 

- Control the browser cache headers of these files when needed. 

- If the externally hosted files are not minified, we can minify them for our customers. 

- Serve these files from a CDN if the customer uses a CDN service. 

The change will be transparent to new and existing customers and no user input is required. They can exclude any file from this using the “Exclude CSS files” and “Exclude JavaScript files” text areas in the File Optimization tab of WP Rocket. 

## Technical Specs

- Current exclusions need to work as they do now. If a user excludes an external CSS or JS file from combination using the text area in the UI or the filter, exclude these files from minification as well. 

- If the external file is not minified, then minify the contents of the file. 

- Remove references to the external file in the source code.

- Only process files with .css and .js extensions for now. 

## Risks

Since we are not modifying these files or its order, but only changing its location, this feature should be safe and we are not expecting any issues because of this. 

## Acceptance Criteria

- When minification is active, externally hosted CSS and JS files are hosted locally. 

- The reference to the external file is replaced with the local one. 

- When combination is active, these files are combined into the combined CSS or JS file.

- If the file is excluded using the UI or filter, the file is not minified or combined. 

- When CDN is enabled, the files are rewritten to the CDN CNAME

<p></p>

**Edge Cases**

- What if the external CSS changes from time to time (i.e. dynamic CSS)? 

- ^ The user can fix this by excluding this CSS file from minification. This should be really rare. 

- Update this documentation to include minification - [<u>https://docs.wp-rocket.me/article/39-excluding-external-js-from-concatenation</u>](https://docs.wp-rocket.me/article/39-excluding-external-js-from-concatenation)

Once #[<u>2689</u>](https://github.com/wp-media/wp-rocket/issues/2689) is done and if we create a doc for that, update that as well.

remyperona:
closes #2810 

### Code changes
The `Minify\CSS` & `Minify\JS` classes now use the `AssetsLocalCache` class during the minification process, to save a local copy of external CSS/JS files. This local copy is then used to create a minified version, saved in the `min` folder.

Code optimizations have been made mostly in the `Minify\CSS` and `Minify\JS` classes. For the `JS` one, we also added an auto-exclusion for Google Analytics JS, as we already handle it through our Google Tracking add-on.

There is now a `ProcessorInterface`, that is implemented by the Minify/Combine classes, and allows correct typehinting in the `AbstractMinifySubscriber` when using the `set_processor_type()` method.

### Acceptance Criteria

- When minification is active, externally hosted CSS and JS files are hosted locally. 
- The reference to the external file is replaced with the local one. 
- When combination is active, these files are combined into the combined CSS or JS file.
- If the file is excluded using the UI or filter, the file is not minified or combined. 
- When CDN is enabled, the files are rewritten to the CDN CNAME


iCaspar (approved):

Davidacu (changes requested):
Here are some issues I have found so far:

**Minify CSS:**
When the external CSS file contains relative URLs , those requests are failing.
**To reproduce:**
1. Navigate to https://rocketlabsqa.dev/
2. Observe the network pane for requests to `fontawesome-webfont.woff?`

The request fails:
https://i.imgur.com/ny13E92.jpg
https://i.imgur.com/BioDkFM.png

**Minify JS:**
1. The minified files that wp rocket generates from the external JS contains whitespaces
https://i.imgur.com/DGZgP26.png
2. When the external JS is already minified, we don't seem to host the file locally
https://rocketlabsqa.dev/external-minified-css/
visit the page and scroll to the bottom looking for the jquery script, is loaded from the external source and not locally from cache:
https://i.imgur.com/HAXIrjN.png

Tabrisrp:
@Davidacu ready for another round of QA, both issues should be fixed

Davidacu (changes requested):
All issues reported have been fixed 🥳  , 97% Pass rate.
Find the test results here -> [testrail-report-14.pdf](https://github.com/wp-media/wp-rocket/files/5042896/testrail-report-14.pdf)


There is 1 failed test case, the URLs from external CSS in the combined file is still relative.
image

Davidacu (approved):
The last reported issue with the relative URL from external CSS in the combined file has been fixed.
All cases have passed QA 🎉

Davidacu:
For visibility, there is an issue that may potentially break some functionality on customer sites.
it is related to Minify External Javascript Files. 

The case is the following:
image

Notice the domain part is removed in the request that was originated from the minified file that WP Rocket generated.
@Tabrisrp Took a look at this, and the cause of the issue is that the initial script on the page is creating another script dynamically to be loaded, and the dynamic one uses the same domain as the initial, so it's breaking.

The script in question has been included in the exclusion list, but still, there may be similar scripts that does this and will break.

PR #2963, opened 2020-07-31
Enhancement/integration theme fixture

iCaspar:
This adds an integration testcase for use with WP themes and child themes, allowing us to create WP_Theme objecgts with basic 3rd party theme properties (stylesheet, name, and template).

Also revises Integration tests and fixtures for the `ThirdParty/Themes/Divi` Subscriber and fixes a found bug when checking for a Divi child theme.

Updates corresponding unit tests to account for fixture revisions.


tests/Integration/WPThemeTestcase.php
```diff
@@ -0,0 +1,81 @@
...
+namespace WP_Rocket\Tests\Integration;
+
+use WP_Theme;
+
+/**
+ * Integration testing framework for themes (and child themes).
+ *
+ * @since 3.8
+ */
+abstract class WPThemeTestcase extends \WP_Rocket\Tests\Integration\FilesystemTestCase {
```
Tabrisrp, line 12:
```suggestion
abstract class WPThemeTestcase extends FilesystemTestCase {
```

tests/Integration/WPThemeTestcase.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+namespace WP_Rocket\Tests\Integration;
+
+use WP_Theme;
```
Tabrisrp, line 5:
```suggestion
use WP_Theme;
use WP_Rocket\Tests\Integration\FilesystemTestCase;
```

iCaspar: Turns out we don't need the `use` here -- they are in the same namespace :)

PR #2971, opened 2020-08-06
Delay JS execution: front part (3/3)

### LINKED ISSUE #2905: Delay JS execution: front part (3/3)
Sub-task for epic #2899 

### New `Optimization\DelayJS\HTML`
- inject `Options_Data` as a dependnecy
- a method `delay_js()` to parse the HTML, find the scripts, and modify the HTML if it matches with the inclusion list set in the settings and the option is enabled
- write tests

- Use RegEx to find scripts in the HTML
- Use `data-rocketlazyloadscript` as the attribute to store the URL or the base64 encoded inline script

### New `Optimization\DelayJS\Subscriber`
- Inject the above as a dependency
- Add a callback method on `rocket_buffer` to call `delay_js()` from `DelayJS::HTML`

### Update `Optimization\Minify\JS\Combine`
- Update the exclusion check to exclude any script using the `data-rocketlazyloadscript` attribute from combine
- Update tests

engahmeds3ed:
Closes #2905 

### New `Optimization\DelayJS\HTML`
- inject `Options_Data` as a dependnecy
- a method `delay_js()` to parse the HTML, find the scripts, and modify the HTML if it matches with the inclusion list set in the settings and the option is enabled
- write tests

- Use RegEx to find scripts in the HTML
- Use `data-rocketlazyloadscript` as the attribute to store the URL or the base64 encoded inline script

### New `Optimization\DelayJS\Subscriber`
- Inject the above as a dependency
- Add a callback method on `rocket_buffer` to call `delay_js()` from `DelayJS::HTML`

### Update `Optimization\Minify\JS\Combine`
- Update the exclusion check to exclude any script using the `data-rocketlazyloadscript` attribute from combine
- Update tests


inc/Engine/Optimization/DelayJS/HTML.php
```diff
@@ -0,0 +1,172 @@
...
+			return false;
+		}
+
+		if ( rocket_get_constant( 'DONOTROCKETOPTIMIZE' ) ) {
+			return false;
+		}
+
+		if ( rocket_get_constant( 'DONOTDELAYJS' ) ) {
+			return false;
+		}
```
Tabrisrp, line 74: we don't need a custom constant check for this feature

inc/Engine/Optimization/DelayJS/HTML.php
```diff
@@ -0,0 +1,172 @@
...
+
+		if ( ! $this->options->get( 'delay_js' ) ) {
+			return false;
+		}
+
+		return true;
```
Tabrisrp, line 80: This can be simplified to
```suggestion
		return (bool) $this->options->get( 'delay_js', 0 ) );
```

inc/Engine/Optimization/DelayJS/HTML.php
```diff
@@ -0,0 +1,172 @@
...
+	/**
+	 * Parse the html and add/remove attributes from specific scripts.
+	 *
+	 * @param string $html Buffer html for the page.
+	 *
+	 * @return string
+	 */
+	private function parse( $html ) {
+		$regex_pattern = '<script\s*(?<attr>[^>]*)?>(?<content>.*)?<\/script>';
+		return preg_replace_callback( '/' . $regex_pattern . '/Uis', [ $this, 'replace_scripts' ], $html );
```
Tabrisrp, line 92: No need to assign the pattern to a variable if you're going to use it only once after
```suggestion
		return preg_replace_callback( '/<script\s*(?<attr>[^>]*)?>(?<content>.*)?<\/script>/Uis', [ $this, 'replace_scripts' ], $html );
```

inc/Engine/Optimization/DelayJS/HTML.php
```diff
@@ -0,0 +1,172 @@
...
+	 *
+	 * @return string
+	 */
+	public function delay_js( $html ) {
+
+		if ( ! $this->is_allowed() ) {
+			return $html;
+		}
+
+		$this->allowed_scripts = $this->prepare_allowed_scripts_regex();
```
Tabrisrp, line 51: We could bailout early here if there is no scripts in the list of allowed scripts, since nothing needs to happen. Small performance improvement

inc/Engine/Optimization/DelayJS/Subscriber.php
```diff
@@ -0,0 +1,96 @@
...
+			'',
+			true
+		);
+		wp_enqueue_script( 'rocket-delay-js' );
+
+		$script_filename = rocket_get_constant( 'SCRIPT_DEBUG' ) ? 'lazyload-scripts.js' : 'lazyload-scripts.min.js';
+
+		wp_add_inline_script(
+			'rocket-delay-js',
+			$this->filesystem->get_contents( rocket_get_constant( 'WP_ROCKET_PATH' ) . 'assets/js/' . $script_filename )
```
Tabrisrp, line 92:
```suggestion
			$this->filesystem->get_contents( rocket_get_constant( 'WP_ROCKET_PATH' ) . "assets/js/{$script_filename}" )
```

Small improvement in readability and performance

inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
```diff
@@ -101,7 +101,13 @@ protected function is_minify_excluded_file( array $tag ) {
 		}
 
 		// File should not be minified.
-		if ( false !== strpos( $tag[0], 'data-minify=' ) || false !== strpos( $tag[0], 'data-no-minify=' ) ) {
+		if (
+			false !== strpos( $tag[0], 'data-minify=' )
+			||
+			false !== strpos( $tag[0], 'data-no-minify=' )
+			||
+			false !== strpos( $tag[0], 'data-rocketlazyloadscript=' )
```
Tabrisrp, line 109: This check should be in the combine JS class specifically, because having the file minified is fine (it stays in place and can be delayed the same), but having it combined (and thus, removed from the HTML) is not

tests/Unit/inc/Engine/Optimization/DelayJS/Subscriber/delayJs.php
```diff
@@ -0,0 +1,74 @@
...
+
+		if ( $this->donotrocketoptimize || $bypass ) {
+			$this->options->shouldReceive( 'get' )
+			              ->with( 'delay_js', 0 )
+			              ->never();
+
+			$this->options->shouldReceive( 'get' )
+			              ->with( 'delay_js_scripts', [] )
+			              ->never();
+		}else{
```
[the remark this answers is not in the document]
iCaspar: Fixed in cd253ef along with other code spacing. :)

PR #2975, opened 2020-08-10
Subtask 2902 - Adds the Preload Pages script

### LINKED ISSUE #2902: Preload links: preload script (3/3)
Sub-task for epic #2898 

### New `assets/js/preload.js`
- Write the script to preload the links

hellofromtonya:
Closes #2902 

This PR adds the script that handles preloading pages. The script does the following:

- Adds a `<link rel="prefetch" href="url">` link for a page(s) to be preloaded - after being triggered.
- When triggered:
    - Mobile: when user touches the link before releasing it
    - Desktop: when hovering over a link (after a false trigger delay) or when clicking a link before releasing it
- Triggering:
     - Hover trigger has a false trigger delay to handle "Do not preload if the mouse just passes over the link. If the user scrolls over a menu, that will be a lot of preloaded links. "
     - Limits the number of preloads per second. This will also help with the previous situation where the user scrolls over a bunch of links. 
     - Disables the feature if the browser uses data saver mode. Instant page does this. 

## TODO
- [x] Analyze instant.pages script.
- [x] Analyze flying pages script.
- [x] Merge in the new gulp script for compiling a single script.
- [x] Add `<link rel="prefetch" href="url">` link for a page(s) to be prefetched
- [x] Non-mobile trigger: when hovering over the link after the false trigger delay
- [x] Mobile trigger: only when a user touches the link (before releasing it)
- [x] Non-mobile trigger: when clicking/touching the link (before releasing it)
- [x] Limit the number of preloads per second.
- [x] Disable the feature if the browser uses data saver mode.
- [x] Strip out the browser checker script. It's now in PR https://github.com/wp-media/wp-rocket/pull/2984

>Limit the number of preloads per second.

A max of 2 prefetches can happen per second. How? The onHoverDelayTime is set at 0.5 seconds (500 ms). What does this mean? A user would need to hover over a link for at least 0.5 seconds before the script adds the prefetch link. Too low of a time here can cause unnecessarily adding prefetch links that the user might not click through to go view. 

## Testing

We don't yet have a JS testing suite. Instead, let's use JSFiddle. [Here's this class in JSFiddle](https://jsfiddle.net/hellofromtonya/1Lybwa86/7/). To use:

1. In the right panel, right click on one of the links at the top and inspect in Dev Tools.
2. Then open `<head>` and watch above the closing tag.
3. Hover over one of the links. When you hover for at least 5 seconds, a `<link rel="prefetch" href="">` should be added.
4. Click on hold a link. It should add a link for it.

Things it should not do:
1. Add a prefetch link for the current URL.
2. Add duplicate prefetch links.


hellofromtonya:
This PR depends upon the `RocketBrowserCompatabilityChecker` in the `browser-checker.js` script. This script is encapsulated in PR https://github.com/wp-media/wp-rocket/pull/2984. Why? This feature and the delay js feature both need it. Keeping it DRY.

PR #2976, opened 2020-08-10
Use common external exclusions list for Minify JS

Tabrisrp:
Now that we are locally hosting & minifiying 3rd party JS following enhancement #2810, We noticed during QA that change could cause issues, like what is described in [this comment](https://github.com/wp-media/wp-rocket/pull/2959#issuecomment-670712275).

To prevent that, this PR is taking the existing list of excluded externals from combine JS, moving it to the parent class, and also using it in the minify JS class when checking for the external exclusions.


Tabrisrp:
@wp-media/qa Ready for QA, this change should prevent the issues found in PR #2959

PR #2977, opened 2020-08-10
Add Gulp task for compiling a single JS script

hellofromtonya:
Breaking this PR out of the delay JS feature branch as it's also needed for Prefetch Pages script.

This new Gulp task allows us to compile a single `assets/js/` script by:

- Running babel to transpile modern JS back into older ES5 version for wider browser compatibility
- Running uglify to minify the script
- Renaming the new compiled script to `.min.js`

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -68,7 +68,7 @@ function compile(watch) {
         });
     }
 
-    rebundle(); 
+    rebundle();
 }
 
 function watch() {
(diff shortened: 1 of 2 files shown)


hellofromtonya:
@Tabrisrp This PR splits out the gulpfile changes so that we can reuse them in both the delay JS and preload pages features.

PR #2984, opened 2020-08-11
Adds browser checker script

hellofromtonya:
Both the delay js and preload links scripts have common browser checker code. This PR pulls the common code out of those scripts to encapsulate and keep it DRY.

## TODO
- [x] Split out the code into separate class
- [x] Compile (transpile and minify) into a `.min.js` script
- [x] Enqueue as an inline script.


hellofromtonya:
@Tabrisrp Where should we enqueue this script? We'll want to use it _when_ either the delay js and/or preload links is turned on. Do you want to do it here in this PR? Or capture that work in each of the other feature PRs?

Tabrisrp:
@hellofromtonya We can update that in both branches when the other sub-tasks are merged, before QA, do you agree?

hellofromtonya:
@Tabrisrp Yes. I'd suggest merging this into the 3.7 branch and then update each of the feature branches.

PR #2969, opened 2020-08-05
Delay JS execution: delay script (2/3)

hellofromtonya:
Closes #2904 

Adds the JavaScript file to:

1. Find all `<script>` elements with the `data-rocketlazyloadscript` attribute (the value holds the `src`).
2. Moves the attribute value to the `src` attribute.
3. Removes the data attribute.

The above is triggered when the user interacts with the page, i.e. on a window event.

## TODO:

- [x] Trigger on user interactivity.
- [x] Remove the `data-rocketlazyloadscript` attribute after triggering (not needed anymore).
- [x] Remove the listener from the `window` once triggered.
- [x] Test with base64.
- [x] Add tests. 👈  we don't have JS testing bootstrapped into the repo...yet. Using JSFiddle for now.
- [x] Review the listener options - what if passive is not supported in an older browser? [Added](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#Safely_detecting_option_support)
- [x] Modify `gulpfile.js`: add a task for compiling single script.
To run:
```
gulp js:compile_single --script=lazyload-scripts.js
```
- [x] Optimized the new script: `assets/js/lazyload-scripts.min.js`.

## Tests:
Using JSFiddle as we don't yet have JUnit or another JS test suite in Rocket. [See it in action here](https://jsfiddle.net/hellofromtonya/79efrd2L/20/).

Browser Tests:
- [x] Chrome: Version 84.0.4147.89 (Official Build) (64-bit)
- [x] Safari: Version 13.1.2 (14609.3.5.1.5)
- [x] Firefox: Version 78.0.2 (64-bit)
- [x] Firefox Developer Edition: Version 80.0b4 (64-bit) 

## Notes:

- Use the optimized (transpiled and minified) `assets/js/lazyload-scripts.min.js` file.
- And then insert the minified code as in line `<script>`. Why? Faster. No external resource to request from the web server.

PR #2986, opened 2020-08-11
Updates the Delay JS Script to Use the Browser Checker class.

hellofromtonya:
Relates to PR #2969 and PR #2984.

Sub-task for epic #2899 

This PR updates the delay js script, i.e. `lazyload-scripts.js` script, to use the new `RocketBrowserCompatabilityChecker`. Why? Both the delay js and preload links scripts had redundant browser checking code. This code was extracted and encapsulated into the new browser checker script.

It also borrows the same design pattern for adding a named static constructor. Why? To encapsulate the knowledge of how to create the object.

Here is the [JSFiddle to view this change actually running](https://jsfiddle.net/hellofromtonya/79efrd2L/26/).


hellofromtonya:
@Tabrisrp This PR can't be QA'd yet. Why? We need to enqueue the browser checker and load this new script. I'm thinking it should be QA'd as part of the whole delay js feature. What do you think?

Tabrisrp:
Yes, for both delay JS and preload links, QA is being done when every sub-tasks are merged together

PR #2987, opened 2020-08-12
Prevent PHP notice when trying to serve the WC empty cart cache

Tabrisrp:
The `$wp` global is not created yet on `plugins_loaded`, which was triggering the following notices:

```
PHP Notice:  Trying to get property 'query_vars' of non-object
PHP Notice:  Trying to get property 'request' of non-object
```

Changing the hook to `after_setup_theme` solves the issue, at that point the global is populated with the necessary object.


Tabrisrp:
@wp-media/qa Ready for QA.

The notices showed up when trying to serve the WC empty cart cache. You will have to delete the local storage cache for the `wc_cart_fragments` to trigger it if it exists.

PR #2993, opened 2020-08-12
Closes #2899 New Feature: Delay JS execution

### LINKED ISSUE #2899: Delay JS execution time (LazyLoad Script)

Tabrisrp:
Closes epic #2899 

Adds a new option in WP Rocket settings page to delay JS execution, with a textarea to add the JS URLs or patterns to delay. A default list of values is used, which can be restored by clicking the button under the textarea.

On the frontend, Delayed JS are modified to use a `data-rocketlazyloadscript` attribute that will be containing either the original URL from the `src` attribute, or a base64 encoded version of the content for inline scripts.

An inline script is also inserted to handle the conversion from the `data-rocketlazyloadscript` back to the `src` attribute.

### Acceptance Criteria

- On activation of the setting, the text area will be pre-populated with the scripts that we have identified as safe to delay.
- New scripts can be added and existing scripts can be removed from the text area. 
- If a keyword matches with a JavaScript file, this file won't be minified / concatenated and deferred. If a keyword matches an inline JS, this script won't be inserted into the concatenated file if the « Combine JavaScript files » option is activated.
- These scripts will not be loaded when the page loads until a user interaction. 
- On user interaction, the script will be loaded. 
- Restore the default scripts list when a user clicks on « Restore default settings ». 

**Testing notes:**

- Cross browser testing is recommended for this to make sure we don’t break things on different browsers.


iCaspar:
@wp-media/qa Ready for QA, acceptance criteria in OP

Tabrisrp:
@hellofromtonya Can you have a look at point 2, 4 & 5 above?

@Davidacu For point 1, it could be browser caching of the WPR admin script, seeing how the page reloaded when you clicked on the button instead of executing the AJAX action. Does it work if you force a browser cache clear?

For point 3, it's doable, but it was not part of the spec for this release. Everyone can add/change the values they need in the field, and it's unlikely you'll have a ton of different languages versions on your site. @arunbasillal What do you think?

arunbasillal:
> For point 3, it's doable, but it was not part of the spec for this release. Everyone can add/change the values they need in the field, and it's unlikely you'll have a ton of different languages versions on your site. @arunbasillal What do you think?

I agree with this. Let's keep it simple for now. Even if they use multiple languages, they can add all that to the list manually. EN and GB covers plenty of use cases. Our aim is to have a starting point, not to be comprehensive.

arunbasillal:
On a related note, now that we know that the `connect.facebook.net/ru_RU/sdk.js` exist and it can be delayed, shall we add it to the safe list @Tabrisrp ?

Tabrisrp:
It wouldn't make sense to me to add the RU version before the FR one, knowing our userbase. I would vote to keep what we have now, and let our users decide what they need.

arunbasillal:
@Davidacu I tested leadpages - https://jmp.sh/YGoPWsq and it seems to work well with WP Rocket. 
As you can see, the script is not loaded when the page is loaded. As soon as I move my mouse, it starts downloading.

hellofromtonya:
>Added a simple script that prints "hello world" and I made sure to add it to the list ofDelay scripts, however the script never runs, the attribute `data-rocketlazyloadscript` is added but after user interaction there is no request made.

Bringing the slack conversation to here.

The original script:

```html
<script src='https://rocketlabsqa.dev/wp-content/rocket-test-data/scripts/print-helloworld.js' >       </script>
```

Notice the extra spaces within the open tag and between the opening and closing tags.

The script after this PR processes it:

```html
<script data-rocketlazyloadscript='https://rocketlabsqa.dev/wp-content/rocket-test-data/scripts/print-helloworld.js' >       </script>
```

After our lazyload-script runs with the above HTML, the external script is not fetched from the browser; thus, the `console.log` inside of it never runs because the script is never fetched.

If we remove the whitespace in the HTML, it works fine.

```html
<script data-rocketlazyloadscript='https://rocketlabsqa.dev/wp-content/rocket-test-data/scripts/print-helloworld.js'></script>
```

## What's going on? Why does extra whitespaces in the HTML cause this to occur?

Here's what I _think_ is going on:

In the browser, as it parses the HTML, this `<script>` is seen as inline script. There’s no external `src` specified and there are characters between the opening and closing tags. I suspect the browser's HTML parser interprets this element as an inline script, not external script. When it builds the DOM tree, that node is an inline script. 

When our script changes the data attribute to `src` attribute, it doesn't work because the node is an "inline" element, not external `script` element. So the browser doesn't know to go fetch the external script from the server.

## How can we fix this?

Option 1: In PHP (server side), strip the extra whitespaces to ensure this element is kept in the right format for the browser.

Option 2: In JavaScript (browser side): delete the `script` element and insert a new `script` element.

### Which option is better?

Option 1: relatively easy solution and serves the element to the browser in the right format.

Option 2: could cause a little bitty performance hit. Why?
Removing and inserting nodes causes the browser engine to reprocess. Depending upon the change, it can cause it to redo the rendering process. That blocks the main thread of the browser to do this work. It could have a performance hit if the user is trying to interact at the same time.

Davidacu:
@Tabrisrp regarding point 1, weird but I cannot reproduce it anymore 🤔 .
Point 2 - fixed.
Point 3 - won't fix.
Point 4. Pending investigation?
point 5. hmm strange I cannot reproduce it anymore, leadpage script is delayed as expected 👍

Tabrisrp:
@Davidacu we removed eucookielaw from the default list, so point 4 is fixed too

Davidacu:
I can confirm the eucookielaw has been removed from the default list 👍

PR #2994, opened 2020-08-13
Fixes #2962 Prevent whitescreen when Optimizing Google Fonts in some cases

### LINKED ISSUE #2962: Combine Google Fonts causes whitescreen in specific case
**Describe the bug**
When `Combine Google Fonts` feature is enabled with Asset CleanUp `Inline CSS Files` we might experience a white screen. That's because this regex is breaking and returning empty value:
https://github.com/wp-media/wp-rocket/blob/bdcfdca33b5040219bb3a9c5f8d6d3916290a8bb/inc/Engine/Optimization/GoogleFonts/Combine.php#L79

**To Reproduce**
Steps to reproduce the behavior on customer's website:
1. In WP Rocket's settings enable the `Combine Google Fonts` feature
2. In Asset CleanUp settings in the CSS section enable `Inline CSS Files` 
3. Open Woo product's category (in this specific case)

**Expected behavior**
`Combine Google Fonts` working fine with `Inline CSS Files` feature.

**Additional context**
The amount of added CSS is pretty huge, however, I wasn't able to reproduce it on our Mega even with more CSS added. I've tested it with regex101 and it also broke the tool due to the amount of code:
https://regex101.com/r/dtxzF7/1/

We'll need to improve our regex here, I've already talked with @Tabrisrp about that.

Related ticket:
https://secure.helpscout.net/conversation/1232413326/181791?folderId=2415573

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

Tabrisrp:
Fixes #2962 

The current regEx used was causing a catastrophic backtracking fatal error when the content inside the `<head>` was too big.

This PR changes the regEx to a simpler expression, similar to the one used for combine CSS or inserting the critical CSS.

### DIFF
--- a/inc/Engine/Optimization/GoogleFonts/Combine.php
+++ b/inc/Engine/Optimization/GoogleFonts/Combine.php
@@ -76,7 +76,7 @@ class Combine extends AbstractOptimization {
 			return $html;
 		}
 
-		$html = preg_replace( '@(<head[^>]*>.*<title[^>]*>.*</title>)(.*</head>)@isU', '$1' . $this->get_combine_tag() . '$2', $html, 1 );
+		$html = preg_replace( '@<\/title>@i', '$0' . $this->get_combine_tag(), $html, 1 );
 
 		foreach ( $fonts as $font ) {
 			$html = str_replace( $font[0], '', $html );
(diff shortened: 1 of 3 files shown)


Tabrisrp:
@wp-media/qa Ready for QA, you can find the HTML sample of the page with the issue here: https://regex101.com/r/dtxzF7/1/

PR #3000, opened 2020-08-18
Improves Preload Links Feature

hellofromtonya:
Changes from alpha testing:

- [x] Moves "on hover" delay and rate throttle to the configuration passed from Rocket, i.e. to make it easier to "tune".
- [x] Makes preload on-hover faster: changed from 500ms to 200ms.
- [x] Adds a rate throttle for "on hover" to limit the number of links preloaded per second.


remyperona (approved):

Davidacu (approved):
- I can confirm that now the links are prefetched faster 👍 
- The rate throttle can not be perceived with the default hover delay of 200ms. 👍

Issue #2713, opened 2020-06-04
Enable assets optimizations on uncached pages


GeekPress:
## What are we proposing to do? 
With this enhancement our users will be able to use WP Rocket’s features on uncached pages. 

Currently these features ARE available on uncached pages: 
- Rules applied via the htaccess file (e.g. GZIP rules, browser caching rules etc)
- Optimization of the WooCommerce get_refreshed_fragments request
- Conversion of dynamic CSS/JS files to static files
- Disable Emoji
- Disable WordPress embeds
- Heartbeat settings
- Preload for fonts

We could safely add the following features as well: 

- Optimize Google Fonts 
- Optimize CSS Delivery
- Remove jQuery Migrate
- Defer JS
- LazyLoad
- Prefetch DNS
- Preload for fonts (coming in 3.6 and already added by design)
- CDN
- Google Tracking Add-on
- Facebook Pixel Add-on

Users will be able to turn some of these off on a post by post basis. 

## Why are we doing this? 

Currently if a page is excluded from WP Rocket cache, most optimizations are not applied to the excluded page. Uncached pages can still benefit from WP Rocket with the other features we have.

## Benchmark Tests

**Test site:** [<u>**http://mega.wp-rocket.me/**</u>](http://mega.wp-rocket.me/)

The main purpose of the test is to see if any of our options are massively slowing down the load time. The homepage barely has assets or images that could greatly benefit from our optimizations. 

Control test: WP Rocket disabled

Fully loaded time average: 2.44s ([<u>#1</u>](https://gtmetrix.com/reports/mega.wp-rocket.me/yN0NCElI), [<u>#2</u>](https://gtmetrix.com/reports/mega.wp-rocket.me/1bMckdaM), [<u>#3</u>](https://gtmetrix.com/reports/mega.wp-rocket.me/1YhNgIxx), [<u>#4</u>](https://gtmetrix.com/reports/mega.wp-rocket.me/Rdcd6H5V), [<u>#5</u>](https://gtmetrix.com/reports/mega.wp-rocket.me/LJTjNDGl))

For the rest of the tests, WP Rocket page caching is disabled using our helper plugin. 

Test with WP Rocket enabled but all options are disabled. 

[...]

PR #3002, opened 2020-08-18
Fixes #3001 Delay JS alpha issues

### LINKED ISSUE #3001: Delay JS alpha issues
- [x] When I updated from a previous version, I didn’t get the pre-filled list of of JS files until I clicked “Restore Defaults” https://recordit.co/ujbBNTejit

- [x] Could we increase the height of the textarea for Delay JS to be able to view all inclusions without needed to scroll inside the textarea?

- [x] When delay JS is disabled and try enable/disable any addon it will enable delay JS automatically https://share.getcloudapp.com/o0u8q59o

engahmeds3ed:
Fixes #3001 

- [x] When I updated from a previous version, I didn’t get the pre-filled list of of JS files until I clicked “Restore Defaults” https://recordit.co/ujbBNTejit

- [x] Could we increase the height of the textarea for Delay JS to be able to view all inclusions without needed to scroll inside the textarea?

- [x] When delay JS is disabled and try enable/disable any addon it will enable delay JS automatically https://share.getcloudapp.com/o0u8q59o


remyperona (approved):

hellofromtonya (approved):

Davidacu:
Getting a fatal error when activating the plugin
image

Davidacu (commented):
**Working as expected**:
- The default list of safe scripts to delay does appear when updating from previous versions.
- The delay js option appears enabled by default for new users
- The delay js option appears disabled when updating from previous version.
- The height of the textarea for introducing the scripts to delay has now increased.
  before:https://i.imgur.com/bfw3QUj.png
  after: https://i.imgur.com/JqWbwJr.png
- Issue with auto-enabling delay js option is fixed.

**Issues:**
- The delay js option is still not applied in uncached pages.
I noticed the scripts are now added in the page source, however the feature doesn't seem to work.
when the page can be cached the feature does work -> https://screencast-o-matic.com/watch/cYjF0UGR7R

Tabrisrp:
@hellofromtonya @arunbasillal @GeekPress 

I looked back at it, and in fact in the current state we can't apply it on uncached pages, because the optimization process using `rocket_buffer` is not applied for now (see related issue #2713). This is because a page excluded from page doesn't trigger the hook, the process bails out early.

This means that while the script are inserted (since they use the `wp_enqueue_scripts` hook), the rewrite of the HTML to lazyload the scripts is not happening.

We _could_ switch to a different hook like `template_redirect`, but it would come at the cost of doing another output buffering, additional checks to make sure we're on a real front end page, etc. I'm not sure the trade-off is worth it, and it would increase the time to finish this feature.

It could be valuable to keep this for cached page only for now, and switch it with the other optimizations when we work on the epic linked above.

Thoughts?

GeekPress:
> It could be valuable to keep this for cached page only for now, and switch it with the other optimizations when we work on the epic linked above.

To not delay the 3.7 release, let's do like that. Having this feature on cached page is already a big improvement :)

PR #3004, opened 2020-08-18
Fixes #3003 Prevent fatal error following WooCommerce 4.4 update

### LINKED ISSUE #3003: [WIP] Possible conflict with Woocommerce 4.4.0
We're receiving tickets from customers having a fatal error after updating Woocommerce to 4.4.0.
Still investigating this. According to Remy, it looks like they are using the same container package as us, but not the same version, they are on the new one that requires PHP 7 and that includes breaking changes.

Errors are similar to this:
`Fatal error: Declaration of League\Container\ImmutableContainerAwareTrait::getContainer() must be compatible with League\Container\Argument\ArgumentResolverTrait::getContainer(): Psr\Container\ContainerInterface in /home/446954.cloudwaysapps.com/krxgzndejz/public_html/wp-content/plugins/wp-rocket/vendor/league/container/src/Inflector/Inflector.php on line 32`

Tabrisrp:
Namespace the container inside our own namespace to prevent conflict when WooCommerce 4.4 uses league/container version 3

Also prevent the PHP notice related to the REST API with WP 5.5

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,7 @@
 		"php": ">=5.6.0",
 		"wp-media/background-processing": "^1.3",
 		"composer/installers": "~1.0",
+		"container-interop/container-interop": "1.2.0",
 		"matthiasmullie/minify": "1.3.*",
 		"monolog/monolog": "^1.0",
 		"wp-media/rocket-lazyload-common": "^2"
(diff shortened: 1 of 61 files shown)


hellofromtonya:
@Davidacu Ready for QA. Please ping us in Slack when done.

PR #3007, opened 2020-08-19
Improves link prefetching for 3.7 Preload Links feature.

hellofromtonya:
This PR further improves the link prefetching for the 3.7 Preload Links feature.

Improvements:

- [x] Stops double prefetches when hover is processing at the same time a tap/click event is processing.
- [x] Adds the ability to disable "on hover" trigger by setting the `onHoverDelayTime` config param to `-1` in the `Engine\Preload\Links\Subscriber::add_preload_script()` method.
- [x] Ensure prefetch happens when a link is double clicked. 
- [x] Preload current page
- [x] Figure out why it works ~50% of time on smashingcoding, but ~90+ elsewhere
- [x] Get click/tap start accuracy ~90%.
- [x] Remove on hover trigger - not needed if clicking/tapping prefetches links
- [x] Compress the script more - currently too big.

Results:

- Click accuracy is 99+% - after moving the adder into a `Promise`.

Tested on these development sites:

http://smashingcoding.com/
https://tonyam4.sg-host.com/
https://wp-rocket.dev/


hellofromtonya:
We've noticed double entries of links in the Network tab. Here are some snapshots of what's happening and why:

### Example 1
To reproduce:

1. Hover over the link. Prefetch link is added and during idle the browser will request the document (low priority).
2. Server returns 304.
3. Click the same link.
4. Browser starts a prefetch request. Notice the status of "pending". 

Screen Shot 2020-08-20 at 10 56 36 AM

hellofromtonya:
Why did I remove the “on hover”?

Technical side:
- Less CPU cycles and calls to server.
Hovering over multiple items before clicking through means: (a) more CPU cycles, (b) more work for the browser to do and (c) more hits back to the server.
- Less chance of the browser being busy processing other prefetch requests when the user clicks on a link.
Prefetch is a low priority task. It happens in the background when the browser is idle. Imagine that the user hovered over several links and the browser is busy processing those prefetches. Now imagine the user clicks a link while the browser is still busy with the other ones. What happens? No prefetch. Whoops.
- Less code.
- Less complexity.
- Smaller script size for less bytes the browser has to load and less to store on the customer’s hosting.

Feature side:
- Hovering doesn’t mean the user will click through to the page.
- We achieve the impact we want by increasing the robustness of prefetching (ie serving documents from the browser’s prefetch cache) by clicking.

hellofromtonya:
Closed. Using PR #3017

PR #3011, opened 2020-08-20
Fixes multiple delay JS issues

remyperona:
- Add additional patterns and remove too specific ones in favour of more generics
- Add Delay JS option to Post metabox to be able to disable the option on specific posts
- Update the upgrade routine to get a fresh options array instead of re-using the one passed at runtime to prevent missing data


iCaspar (approved):

Tabrisrp:
@wp-media/qa Ready for QA, updating the list of delayed JS

Davidacu:
I confirm that all links included in the default list of scripts are delayed.

Please find the test results attached.
[testrail-report-18.pdf](https://github.com/wp-media/wp-rocket/files/5105955/testrail-report-18.pdf)

There is one test case failure, it is about the **Cookie Notice** plugin by dFactory.
The script is delayed, however, the popup to accept the cookies never appears.
which may not really in the scope of this PR, but could you please take a look @Tabrisrp .

you can test it in https://rocketlabsqa.dev/
The plugin is installed, notice that if you bypass rocket the popup appears.

Tabrisrp:
I can see the issue, but I don't know why it's happening. In term of code and behaviour, it's as expected:
- Scripts have the custom data on loading
- User interaction triggers the switch to the `src` attribute

@arunbasillal @hellofromtonya Note the above

Davidacu (approved):

PR #3021, opened 2020-08-25
Closes #2318 Add i18n.min.js to the $excluded_files array

### LINKED ISSUE #2318: WooCommerce Gutenberg blocks: "wp not defined" error with JS combination/deferring
When WooCommerce Gutenberg blocks are being used on a page, there is a `wp not defined` error when the following WP Rocket options are enabled:

* JavaScript combination
* JavaScript deferring

The solution is to exclude the following files from both processes:

```PHP
/wp-includes/js/dist/api-fetch.min.js
/wp-includes/js/dist/hooks.min.js
```

Depending on the blocks that are used more files and inline JavaScript have to be excluded.

For example, if the **All Reviews** block is used these two exceptions have to also be added:

```PHP
/wp-includes/js/dist/url.min.js
```
and this inline JavaScript:

```JavaScript
wcSettings
```

This can be reproduced in our MEGA site.

**Related ticket:** https://secure.helpscout.net/conversation/1079564082/145400?folderId=377611#thread-3068022924

vmanthos:
Ticket: https://secure.helpscout.net/conversation/1252699151/186413/#thread-3591428397
GitHub issue: #2318 

Related Slack conversation: https://wp-media.slack.com/archives/C43T1AYMQ/p1598275277020100

### DIFF
--- a/inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
+++ b/inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
@@ -48,6 +48,7 @@ abstract class AbstractJSOptimization extends AbstractOptimization {
 	 */
 	protected function get_excluded_files() {
 		$excluded_files = $this->options->get( 'exclude_js', [] );
+		$excluded_files[] = '/wp-includes/js/dist/i18n.min.js';
 		$jquery_urls    = $this->get_jquery_urls();
 
 		if ( ! empty( $jquery_urls ) ) {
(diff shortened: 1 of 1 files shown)


inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
```diff
@@ -48,6 +48,7 @@ public function __construct( Options_Data $options, AssetsLocalCache $local_cach
 	 */
 	protected function get_excluded_files() {
 		$excluded_files = $this->options->get( 'exclude_js', [] );
+		$excluded_files[] = '/wp-includes/js/dist/i18n.min.js';
 		$jquery_urls    = $this->get_jquery_urls();
```
Tabrisrp, line 52:
```suggestion
		$excluded_files    = $this->options->get( 'exclude_js', [] );
		$excluded_files[] = '/wp-includes/js/dist/i18n.min.js';
		$jquery_urls         = $this->get_jquery_urls();
```

PR #3026, opened 2020-08-26
Fixes 3025: Abstracts and transpiles CPCSS removal script

### LINKED ISSUE #3025: CPCSS Removal script causes browser console errors when browser does not fully support ES6.
**Describe the bug**
We are serving ES6 JavaScript code for the CPCSS removal script. The script is not transpiled (via Babel) to convert it from ES6 to ES5 for those browsers which do not fully support ES6. The specific error is for the arrow function.

**To Reproduce**
1. Use IE11 browser and open up the console tab
2. With CPCSS on, load a site with Rocket running on it
3. Notice the error in the console: `SCRIPT1002: Syntax error`.
4. Click on the `arrangement ...` link to see the source of the problem in the Debugger tab. It points to the `() => ` arrow function.

**Expected behavior**
The script should not throw errors in the browsers we support. The script should be transpiled to make it fully compatible.

**Screenshots**
Console tab (the error):
Screen Shot 2020-08-26 at 1 34 04 PM

Debugger tab (this is where the error is occurring):
Screen Shot 2020-08-26 at 1 34 29 PM

**Additional context**
- The script is intermixed in the PHP. It should be abstracted out to the `assets/js/` folder as a JS script and then loaded where needed. Mixing concerns.

- [See browser compatibility here](https://caniuse.com/#feat=arrow-functions)

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

hellofromtonya:
Closes #3025 

- [x] Abstracts the script code out of the PHP code (mixing of concerns) and moves it to `assets/js/cpcss-removal.js`.
- [x] Compiles (transpile and minify) using `gulp js:compile_single --script=cpcss-removal.js` to create a new `assets/js/cpcss-removal.min.js` file


iCaspar (approved):

hellofromtonya:
@Davidacu This PR is ready for QA.

remyperona (approved):

Davidacu (commented):
- I can confirm that the original error generated by the arrow function is gone. 👍 
- CPCSS is generated and properly attached and removed after page load. 👍 
- A different error is still displayed and the debugger points to the rocket-critical-css removal script.
Could you please take a look @hellofromtonya 
image
image

hellofromtonya:
@engahmeds3ed The change in the script needs to be applied to the test fixtures too.

@Davidacu We've fixed the error you found. Can you recheck please?

Davidacu (commented):
I can still see the error in the console on IE11 @engahmeds3ed @hellofromtonya 
1. Go to https://rocketlabsqa.dev/
2. Look the error in the console
image
When I click the error the debugger takes me here:
image

hellofromtonya:
@Davidacu @engahmeds3ed The reason it failed in IE11 is because [IE doesn't support `remove()` method](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove).

Screen Shot 2020-09-11 at 1 39 03 PM


There are a couple of ways to handle this:

- [Add this polyfill](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/remove#Polyfill) - Yuck. Not necessary as we'd have to insert this code for all browsers.
- Check if the method is supported with `elem. hasOwnProperty( 'remove' )`. If no, bail out.

hellofromtonya:
@wp-media/qa @Davidacu Fixed. Ready for QA.

piotrbak:
@hellofromtonya No errors in the IE 11 console, however, is it expected that the Critical CSS is not removed from the source there?

On other browsers works as expected.

Tabrisrp:
Yes that's expected, see https://github.com/wp-media/wp-rocket/pull/3026#issuecomment-691253641

PR #3048, opened 2020-08-31
Fixes 3047 - Guards deprecated Minify_HTML class to avoid fatal errors.

### LINKED ISSUE #3047: 3.7 - Clearfy - PHP fatal error from using deprecated Minify_HTML class
**Describe the bug**

This affects users that:
* Have [Clearfy](https://wordpress.org/plugins/clearfy/) installed.
* They are updating to WP Rocket `3.7`.

The error is this one:
```
PHP Fatal error:  Cannot declare class Minify_HTML, because the name is already in use in /home/domains/example.com/public_html/wp-content/plugins/wp-rocket/inc/deprecated/vendors/classes/class-minify-html.php on line 22
```

Clearfy is also declaring a `Minify_HTML` class in:
```
/wp-content/plugins/clearfy/components/html-minify/includes/classes/ext/php/class-minify-html.php
```

**To Reproduce**

1. Rollback to `3.6.4` if that's not the version already installed.
2. Install [Clearfy](https://wordpress.org/plugins/clearfy/).
3. Enable WordPress debugging.
4. Update WP Rocket to `3.7`.
5. Check the `debug.log` for the error.

**Expected behavior**

No fatal error should occur.

**Additional context**

**Related ticket:** https://secure.helpscout.net/conversation/1268667344/191001?folderId=2135277

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

hellofromtonya:
After discussions, we decided to wrap the deprecated classes with `class_exist` to avoid loading when the class already exists in memory.

Closes #3047

### DIFF
--- a/inc/deprecated/3.7.php
+++ b/inc/deprecated/3.7.php
@@ -5,7 +5,9 @@ defined( 'ABSPATH' ) || exit;
 /**
  * Require deprecated classes.
  */
-require_once __DIR__ . '/vendors/classes/class-minify-html.php';
+if ( ! class_exists( 'Minify_HTML' ) ) {
+	require_once __DIR__ . '/vendors/classes/class-minify-html.php';
+}
 require_once __DIR__ . '/subscriber/admin/Optimization/class-minify-html-subscriber.php';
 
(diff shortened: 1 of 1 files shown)


hellofromtonya:
@wp-media/qa @Davidacu This hotfix PR is ready for QA.

PR #3053, opened 2020-08-31
Excluding aos.js from DeferJS

alfonso100:
The aos.js library (https://github.com/michalsnik/aos) is used by  https://oxygenbuilder.com/
When DeferJS is enabled, there is and "AOS is undefined" error on AOS.init, excluding the JS file from deferring solves the issue.

Ticket: https://secure.helpscout.net/conversation/1267928807/190825?folderId=2683093


Tabrisrp:
You'll need to update the related tests too

Tabrisrp (changes requested):
Automated tests need to be updated to get Travis passing

Tabrisrp (approved):

PR #3057, opened 2020-09-01
Add index.html inside each directory created by wp rocket

engahmeds3ed:
creating directories without `index.html` file may lead to security hole.

- [x] Add index.html file into wp rocket config directory `/wp-content/wp-rocket-config`

- [x] Add index.html file into directories ( min, busting, critical-css )

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -263,14 +263,26 @@ function rocket_init_cache_dir() {
 		rocket_mkdir_p( WP_ROCKET_MINIFY_CACHE_PATH );
 	}
 
+	if ( ! rocket_direct_filesystem()->is_file( WP_ROCKET_MINIFY_CACHE_PATH . 'index.html' ) ) {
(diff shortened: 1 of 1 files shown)


Tabrisrp:
@wp-media/qa Ready for QA

piotrbak (approved):
Worked as expected, 
- .index.html files created
- .htacess inside /wp-content/cache/wp-rocket/ created
- Pages served from the cache
- Everything is removed when deleting WP Rocket

Davidacu:
- When we tested this same security update on Imagify, we needed to make sure that the `index.html` file is created not only at the root of the directories created by the plugin, but also under each sub-directory created, can you please confirm if we should validate the same in rocket?
- Another scenario is to make sure that if we manually delete the `index.html` file, is recreated again by the plugin when that directory is regenerated.
- Another thing to check is to validate that when updating from an older version the `index.html` file is added to the existing directories
- what about `/wp-content/cache/wp-rocket/` , shouldn't we add `index.html` files there too?

@Tabrisrp @piotrbak

PR #3059, opened 2020-09-01
Fixes #2846 Correctly exclude jQuery from combine when WP is in a subfolder

### LINKED ISSUE #2846: Exclude jQuery from combine is not working when WP is in a subfolder
Used latest version

**Describe the bug**
If we have a wordpress site installed in a subfolder: http://example.org/qa/ then exclud jQuery from combine and defer is not working because the regex fails to match the url:
https://github.com/wp-media/wp-rocket/blob/0036b86142cc4083a9192a892cacb4223ddb456c/inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php#L110
https://regex101.com/r/XHUL9t/1

**To Reproduce**
1. http://mega.wp-rocket.me/qa
2. Enable Combin JS, Minify JS, load js deffered & Safe Mode for jQuery (recommended)
3. Clean cache and see that jquery is combined on all pages

**Expected behavior**
JQuery to be excluded from combination.

A solution is to modify the regex to ignore the start of the string comparison like this:
https://regex101.com/r/cpqPkE/1
 
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

Tabrisrp:
Fixes #2846 

### Code changes
Instead of directly using the value from the `src` property for jQuery, we check if the `host` is set for it. If not, we call `site_url()` to make sure we get the absolute URL for the file, and then keep only the path.

This is reverting to how we did it previously.

### DIFF
--- a/inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
+++ b/inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
@@ -166,11 +166,11 @@ abstract class AbstractJSOptimization extends AbstractOptimization {
 		$exclude_jquery = [];
 		$jquery         = wp_scripts()->registered['jquery-core']->src;
 
-		if ( isset( $jquery  ) ) {
-			if ( '' === wp_parse_url( $jquery , PHP_URL_HOST ) ) {
-				$exclude_jquery[] = wp_parse_url( site_url( $jquery  ), PHP_URL_PATH );
+		if ( isset( $jquery ) ) {
+			if ( '' === wp_parse_url( $jquery, PHP_URL_HOST ) ) {
+				$exclude_jquery[] = wp_parse_url( site_url( $jquery ), PHP_URL_PATH );
 			} else {
-				$exclude_jquery[] = $jquery ;
+				$exclude_jquery[] = $jquery;
 			}
 		}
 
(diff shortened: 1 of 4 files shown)


iCaspar (approved):

iCaspar:
@wp-media/qa Ready for QA

Davidacu (changes requested):
- The jQuery library loaded from `/wp-includes/js/jquery/` is not combined as expected in https://rocketlabsqa.dev/ 👍 
- The jQuery library loaded from `/wp-includes/js/jquery/` is still combined when wp is installed in a  subdirectory 🐞 http://mega.wp-rocket.me/qa/
- The jQuery library hosted in external sources such as cloudfare and ajax.googleapis is combined in both type of wp installs 🐞  (these are supposed to be excluded from combine js right?)
Can be tested here:
https://rocketlabsqa.dev/combine-js/
http://mega.wp-rocket.me/qa/combine-js

Find the test results attached
[testrail-report-30.pdf](https://github.com/wp-media/wp-rocket/files/5211555/testrail-report-30.pdf)

Tabrisrp:
@iCaspar Can you re-review my latest changes?
@Davidacu Thank you for the reports, found the issues, ready for another QA

iCaspar (approved):

piotrbak (approved):
- The jQuery library loaded from `/wp-includes/js/jquery/` is not combined as expected in https://rocketlabsqa.dev/
- The jQuery library loaded from `/wp-includes/js/jquery/` is not combined as expected in subdirectory http://mega.wp-rocket.me/qa/
- The jQuery library hosted by external sources such as Cloudflare and ajax.googleapis is not combined in both cases. Tested with deregistering the original jQuery code and registering the new one:
https://snippi.com/s/in79atv with https:// protocol and relative one //.
- If the jQuery is not registered, just added into the source using the `/wp-includes/js/jquery/` path it is not being combined too.

PR #3062, opened 2020-09-02
Closes #2789 Automatic Dreampress compatibility

### LINKED ISSUE #2789: Improve DreamPress hosting compatibility
As part of our Hosting compatibility initiative, we can have a new look at our compatibility with DreamPress.

### Current state
- Helper plugin: https://github.com/wp-media/wp-rocket-helpers/tree/master/compatibility/wp-rocket-compat-dreampress
- Documentation: https://docs.wp-rocket.me/article/1016-dreampress--varnish-compatibility

This works, but discoverability for those is low, as users might:
- Not know there is an issue when using WP Rocket on DreamPress without the helper
- Not find the helper plugin and/or documentation
- There is nothing in the core plugin to indicate that something might not be working well

### Ideal state
- Be able to integrate the helper plugin directly in core, if we can get a way to auto-detect DreamPress
- Update the code for our new architecture

#### Related tickets
- https://secure.helpscout.net/conversation/1196851806/173436?folderId=637040

Tabrisrp:
This PR adds automatic compatibility with DreamPress managed hosting.

### Code changes
We add a new `DreamPress` class to the 3rd party hostings namespace.

This class extends the `NoCacheHost` and has methods to:
- Automatically enable our Varnish Addon + update the description
- Remove the expire rules in the `.htaccess` for HTML to prevent no cache issue with Varnish
- Disable the creation of cache files, `advanced-cache.php`, the `WP_CACHE` constant, and all the related notices

New nit & integration tests have been added for this class methods


iCaspar (approved):

Tabrisrp:
Blocked: waiting for test site

piotrbak:
Our compatibility doesn't work. It seems that the `x-hosted-by` header is not present in the `$_SERVER` array. You can see the whole array here:
https://snippi.com/s/mfw76zr

Tabrisrp:
Updated the code checking for the compatibility. The header is set by the proxy itself, so it won't be available in the PHP `$_SERVER` global variable.

Instead, I'm doing a request to the home URL to get the headers, and check if the header is there. If it is, I save a transient for a week, to prevent us from doing the extrat request on every page load.

Tested this change directly on our DreamPress test site, and the detection is working correctly.

@iCaspar if you can re-review the latest changes
@wp-media/qa This is ready for a new QA, the change is already applied on the test site

piotrbak:
@Tabrisrp Shouldn't we remove the transient when deactivating the plugin? I can see this kind of potential issue:

1. We add another hosting compatibility (let's say the hosting called **xyz**) below the 
https://github.com/wp-media/wp-rocket/blob/88a12ffe232a230a8e5fe9d1aaf7f77ce7bc354a/inc/ThirdParty/Hostings/HostResolver.php#L86
2. The customer is on DreamPress, but he's migrating to the **xyz**
3. After the migration, the compatibility will be still pointing to the DreamPress, deactivation/activation won't help, since the transient is still present.

Also, after migrating to other hosts, we'll be spamming their servers with bad requests when saving the posts (it can lead to problems with resources), etc.

piotrbak:
1. When removing the transient manually, using the following code I was able to trigger the fatal error:
https://snippi.com/s/b9pm6c8

Error itself:
`[22-Sep-2020 09:59:38 UTC] PHP Fatal error:  Uncaught WP_Rocket\Engine\Container\Exception\NotFoundException: Alias (dreampress) is not being managed by the container in /home/wp_74cgrq/wprocket.breakwpdh.com/wp-content/plugins/wp-rocket/inc/Engine/Container/Container.php:266
Stack trace:
#0 /home/wp_74cgrq/wprocket.breakwpdh.com/wp-content/plugins/wp-rocket/inc/Engine/Container/Container.php(93): WP_Rocket\Engine\Container\Container->getFromDelegate('dreampress', Array)
#1 /home/wp_74cgrq/wprocket.breakwpdh.com/wp-content/plugins/wp-rocket/inc/Plugin.php(112): WP_Rocket\Engine\Container\Container->get('dreampress')
#2 /home/wp_74cgrq/wprocket.breakwpdh.com/wp-content/plugins/wp-rocket/inc/main.php(48): WP_Rocket\Plugin->load()
#3 /home/wp_74cgrq/wprocket.breakwpdh.com/wp-includes/class-wp-hook.php(287): rocket_init('')
#4 /home/wp_74cgrq/wprocket.breakwpdh.com/wp-includes/class-wp-hook.php(311): WP_Hook->apply_filters(NULL, Array)
#5 /home/wp_74cgrq/wprocket.breakwpdh.com/wp-includes/plugin.php(478): WP_Hook->do_action(Array)
#6 /home/wp_74cgrq/wprocke in /home/wp_74cgrq/wprocket.breakwpdh.com/wp-content/plugins/wp-rocket/inc/Engine/Container/Container.php on line 266`

2. When updating the post, it seems that the whole cache is cleared. Requests seem to be correct though. In the video I have updated a `Single Page`, all the posts got cleared (their path is `/2020/09/22/(.*)`, not present in the request we're sending to the Varnish)
https://jmp.sh/MCaJoQM

Requests and args sent to the varnish during the update:
https://snippi.com/s/w7owa3n

That's not critical, but good to note.

What do you think about this critical error @Tabrisrp, customers that could use some DB cleaning plugins might experience it? I'm not 100% sure how to recreate it, unfortunately.

iCaspar (approved):

Tabrisrp:
@piotrbak Ready for a new QA. We removed the transient, and used the `DH_USER` server variable instead to perform the verification.

piotrbak:
@Tabrisrp The detection works as expected now. 

Unfortunately, **we're not clearing** the cache at all. There were no changes in the code related to the requests, so I assume I had their cache clearing solution enabled when doing the previous QA (that would explain why the whole cache was purged then).

I tried various different settings to see if they'll help (BAN method, settings 127.0.0.1 as Varnish IP, changing the scheme to HTTPS) but they didn't help. Sorry for missing that in the previous round of QA :(

piotrbak:
@Tabrisrp Didn't want to leave it like this. It looks like we just need to change the IP in our request in order to work:
- Filter Varnish IP to the `localhost` using this filter: `rocket_varnish_ip`

When sending requests like those (`$purgeme` + `$args`), the purge works:
https://snippi.com/s/3khrccl

Here's how their built-in solution is sending the requests:
https://snippi.com/s/bgngzsc

Thanks!

Tabrisrp:
@piotrbak It's ready to be tested again

piotrbak (commented):
@Tabrisrp The only problem we're having is that when we're clearing the homepage, everything is cleared. I think that's not very serious in terms of Varnish, what's your opinion?

- Should not write to wp-config.php or advanced-cache.php at any time 👍
- Should not show any notices due to writing permissions with wp-config.php and advanced-cache.php 👍
- Should not display Varnish add-on section 👍
- Varnish addon should indicate which is the hosting detected 👍
- Should not create cache files from WP Rocket 👍
- Should clear the whole varnish server cache when clearing WP Rocket cache 👍
- Should clear specific URL's varnish cache when using 'Purge This URL' button 🛑
- Should clear varnish server cache when updating a post 🛑

The last two fails because of what I wrote.

piotrbak (approved):
The fails are also happening for the default DreamPress cache clearing solution.

PR #3079, opened 2020-09-09
Fix #3078 prevents naming conflicts on our scripts

### LINKED ISSUE #3078: JS Console errors when Preload Links and Combine JS are both ON
We have 2 customers who have reported that when Preload Links and Combine JS features are both ON, they get console errors. These errors cause JS to fail and our scripts to stop working.

This happens when a customer script has a naming collision with our script. 

Why? Our gulp uglify mangle process removed our scripts' unique namespaces and replaced them with generic single character names. If another script has that same variable name, this happens.

**To Reproduce**

1. [Install this script](https://github.com/danielstjules/blankshield/blob/master/blankshield.min.js). You can enqueue in a theme or helper plugin. Make sure it enqueues _before_ the Preload Links script.
2. Turn on the Preload Links feature.
3. Turn on the Combine Js feature.
4. Open the web page in the browser (incognito)
5. Inspect anything on the page.
6. Open the Console tab.

Notice the error in the console. Turn off Combine JS feature. Refresh. Notice no error in the console.

**Expected behavior**
WPR scripts should retain their scope (either by namespace or IIFE) and not have naming conflicts with other scripts.

**Root Cause**
The problem is due to scope. When we optimize the minified script for production, we use a gulp task. That task runs Babel (to convert from modern JS to an older version that more browsers support), minify, and mangle (compresses names and space to free up more bytes). The mangle part saves us about 1000 bytes in the script size. Yes, we want that. But it's also renaming our unique namespace. Darn it. 

props to @engahmeds3ed 

**Solution**
We need the mangle task as it compresses the Preload Links script by ~1kb. We need it.

We will add an IIFE around the Preload Script and LazyLoad Scripts scripts. The IIFE provides a functional scope, which thereby encapsulates each script and prevents naming conflicts.

- [X] Reproduce the problem
- [X] Identify the root cause
- [X] Scope a solution
- [X] Estimate the effort

engahmeds3ed:
Fixes #3078 

- [x] encapsulate preload-links.js script with self called function

- [x] encapsulate lazyload-scripts.js script with self called function


assets/js/lazyload-scripts.min.js
```diff
@@ -1 +1 @@
...
...("Cannot call a class as a function")}!function(){function r(e,t){n(this,r),this.attrName="data-rocketlazyloadscript",this.browser=t,this.options=this.browser.options,this.triggerEvents=e,this.userEventListener=this.triggerListener.bind(this)}(e(r,[{key:"init",value:function(){this._addEventListener(this)}},{key:"reset",value:function(){this._removeEventListener(this)}},{key:"_addEventListener",value:function(t){this.triggerEvents.forEach(function(e){return window.addEventListener(e,t.userEventListener,t.options)})}},{key:"_removeEventListener",value:function(t){this.triggerEvents.forEach(function(e){return window.removeEventListener(e,t.userEventListener,t.options)})}},{key:"_loadScriptSrc",value:function(){var r=this;document.querySelectorAll("script["+this.attrName+"]").forEach(function(e){var t=e.getAttribute(r.attrName);e.setAttribute("src",t),e.removeAttribute(r.attrName)}),this.reset()}},{key:"triggerListener",value:function(){this._loadScriptSrc(),this._removeEventListener(this)}}],[{key:"run",value:function(){RocketBrowserCompatibilityChecker&&new r(["keydown","mouseover","touchmove","touchstart"],new RocketBrowserCompatibilityChecker({passive:!0})).init()}}]),r).run()}();
```
hellofromtonya, line 1: The mangler has added an `var e=function(){ ... }();` to the IIFE. That `e` variable is in global scope. That means it could conflict with another script when combining JS and another script has a var or function named `e`.

The mangler and IIFE then isn't going to work. 

Let me take a look at the configurations for the mangler task. Maybe we can exclude the global var naming to preserve the namespace.

assets/js/preload-links.min.js
```diff
@@ -1 +1 @@
...
...f,n.onload=e,n.onerror=t,document.head.appendChild(n)}).catch(function(){})}},{key:"_prepareUrl",value:function(e){if(null===e||"object"!==(void 0===e?"undefined":r(e))||!1 in e||-1===["http:","https:"].indexOf(e.protocol))return null;var t=e.href.substring(0,this.config.siteUrl.length),n=this._getPathname(e.href,t),i={original:e.href,protocol:e.protocol,origin:t,pathname:n,href:t+n};return this._isLinkOk(i)?i:null}},{key:"_getPathname",value:function(e,t){var n=t?e.substring(this.config.siteUrl.length):e;return n.startsWith("/")||(n="/"+n),this._shouldAddTrailingSlash(n)?n+"/":n}},{key:"_shouldAddTrailingSlash",value:function(e){return this.config.usesTrailingSlash&&!e.endsWith("/")&&!this.regex.fileExt.test(e)}},{key:"_isLinkOk",value:function(e){return null!==e&&"object"===(void 0===e?"undefined":r(e))&&(!this.prefetched.has(e.href)&&e.origin===this.config.siteUrl&&-1===e.href.indexOf("?")&&-1===e.href.indexOf("#")&&!this.regex.excludeUris.test(e.href)&&!this.regex.images.test(e.href))}}],[{key:"run",value:function(){"undefined"!=typeof RocketPreloadLinksConfig&&new n(new RocketBrowserCompatibilityChecker({capture:!0,passive:!0}),RocketPreloadLinksConfig).init()}}]),n).run()}();
```
hellofromtonya, line 1: Same thing here, except this time, mangler added a global var of `r` for the IIFE.

hellofromtonya (changes requested):

hellofromtonya (approved):
🥇 Well done @engahmeds3ed

engahmeds3ed:
> @engahmeds3ed This PR though should be based off of `develop` as it will part of the next minor release instead of a hot fix. Can you fix that please?

Done

hellofromtonya:

Davidacu (approved):
- I created a template from the customer site having this issue, I can confirm that the error in the console not longer appears, and preload-links and delay js scripts work as expected.
- Ran smoke/regression tests for Minify JS, Combine JS, Preload links and Delay JS. 
100% Pass rate, find the test execution results attached.
[testrail-report-29.pdf](https://github.com/wp-media/wp-rocket/files/5210638/testrail-report-29.pdf)

Davidacu (approved):

PR #3095, opened 2020-09-14
Auto exclude preload delay js scripts

GeekPress:
This PR will automatically exclude our Preload and Delay JS inline scripts. It will ensure us that these features will always work. By being included into the combined JS files, our features can stop to work if a JS error is triggered.

Slack conversation: https://wp-media.slack.com/archives/GUT7FLHF1/p1599831075030100

### DIFF
--- a/inc/Engine/Optimization/Minify/JS/Combine.php
+++ b/inc/Engine/Optimization/Minify/JS/Combine.php
@@ -408,7 +408,6 @@ class Combine extends AbstractJSOptimization implements ProcessorInterface {
 	 */
 	protected function get_excluded_inline_content() {
 		$defaults = [
-			'RocketBrowserCompatibilityChecker',
 			'document.write',
 			'google_ad',
 			'edToolbar',
@@ -761,6 +760,7 @@ class Combine extends AbstractJSOptimization implements ProcessorInterface {
 			'wp_post_blocks_vars.listed_posts=[',
 			'captcha-hash',
 			'mapdata={',
+			'RocketBrowserCompatibilityChecker'
 		];
 
 		/**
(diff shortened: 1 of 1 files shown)


inc/Engine/Optimization/Minify/JS/Combine.php
```diff
@@ -408,6 +408,9 @@ protected function add_to_minify( $content ) {
 	 */
 	protected function get_excluded_inline_content() {
 		$defaults = [
+			'RocketBrowserCompatibilityChecker',
+			'RocketLazyLoadScripts',
+			'RocketPreloadLinks',
```
hellofromtonya, line 413: The `RocketLazyLoadScripts` and `RocketPreloadLinks` variable names are not present in the minified scripts as we use a mangler in the gulp task to further reduce the bytes in each of the scripts.

However, we do retain the `RocketBrowserCompatibilityChecker` name. All 3 of these scripts have that variable name in them. Therefore, we can simplify this to just `RocketBrowserCompatibilityChecker`.

hellofromtonya (changes requested):

hellofromtonya (approved):

hellofromtonya:

How to test?

1. Turn on the Delay JS and Preload Links features.
2. Turn on the Combine JS feature.
3. Check that these 3 inline scripts are not combined but are actually in the HTML as separate `<script>` elements:
   
```
<script id="rocket-browser-checker-js-after">...the script code....</script>
<script id="rocket-delay-js-js-after">...the script code....</script>
<script id="rocket-preload-links-js-extra">...the script code....</script>
<script id="rocket-preload-links-js-after">...the script code....</script>
```

piotrbak (approved):
It worked as expected. Scripts mentioned by @hellofromtonya are not combined. The following script wasn't combined in the previous version too:
`<script id='rocket-preload-links-js-extra'>`

It's good to add that this specific one is added above the combined JS file.

PR #3098, opened 2020-09-14
Add pixel-caffiene/frontend.js to auto-delay list

### LINKED ISSUE #3088: Delay JavaScript - Pixel Caffeine is breaking

**Describe the bug**

When the Pixel Caffeine plugin is used along with the **Delay JavaScript scripts** feature, there is a JavaScript console error:
```
 aepc_extend_args is not defined
```

This is because they are using `fbq( 'init' )` and we automatically delay `fbq(`.

**To Reproduce**
1. Install/Set up the [Pixel Caffeine](https://wordpress.org/plugins/pixel-caffeine/) plugin.
2. Enable **Delay JavaScript scripts**
3. Check the JavaScript console for the error.

**Expected behavior**

There shouldn't be any errors in the JavaScript console.

**Screenshots**
screenshot

**Additional context**

Delaying the following script fixes the error:
```
/wp-content/plugins/pixel-caffeine/build/frontend.js
```

**Related Ticket:** https://secure.helpscout.net/conversation/1279175035/193966/

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

iCaspar:
Closes #3088.

- [X] Add caffeine frontend.js to default delayed scripts.
- [X] Update test fixtures.


remyperona (changes requested):
In addition, you need to add a new method to hook on the upgrade routine, to add this new script to the saved user list in the settings

iCaspar:
Modified the current set_options_on_update() to check for the new script and add if it's not already there. Ready for another look @Tabrisrp .

remyperona (approved):

iCaspar:
@wp-media/qa This one is ready for your magic!

remyperona (approved):

Tabrisrp:
@wp-media/qa Ready for QA

- The pixel-caffeine script should be delayed too
- The script path should be added automatically to the default list
- The script path should be added to the saved value in the option if the `fbq` pattern is also present, but not if it's not

iCaspar:
@Tabrisrp Just a last quick check on the integration tests I added there. QA can continue during review.

tests/Integration/inc/Engine/Optimization/DelayJS/Admin/Subscriber/optionUpdate372.php
```diff
@@ -0,0 +1,41 @@
...
+		parent::tearDown();
+
+		$this->restoreWpFilter( 'wp_rocket_upgrade' );
+	}
+
+	/**
+	 * @dataProvider configTestData
+	 */
+	public function testShouldDoExpected( $config, $expected ) {
+		global $wp_filter;
```
Tabrisrp, line 29:
```suggestion
```

tests/Integration/inc/Engine/Optimization/DelayJS/Admin/Subscriber/optionUpdate372.php
```diff
@@ -0,0 +1,41 @@
...
+namespace WP_Rocket\Tests\Integration\inc\Engine\Optimization\DelayJS\Admin\Subscriber;
+
+use WP_Rocket\Tests\Integration\TestCase;
+
+/**
+ * @covers \WP_Rocket\Engine\Optimization\DelayJS\Admin\Subscriber::option_update_3_7_2
+ *
+ * @group  DelayJS
+ */
+class Test_optionUpdate372 extends TestCase{
```
Tabrisrp, line 12:
```suggestion
class Test_optionUpdate372 extends TestCase {
```

tests/Unit/inc/Engine/Optimization/DelayJS/Admin/Settings/optionUpdate372.php
```diff
@@ -0,0 +1,43 @@
...
+use WP_Rocket\Admin\Options_Data;
+use WP_Rocket\Engine\Optimization\DelayJS\Admin\Settings;
+use WP_Rocket\Tests\Unit\TestCase;
+
+/**
+ * @covers \WP_Rocket\Engine\Optimization\DelayJS\Admin\Settings::option_update_3_7_2
+ *
+ * @group  DelayJS
+ */
+class Test_OptionUpdate372 extends TestCase{
```
Tabrisrp, line 15:
```suggestion
class Test_OptionUpdate372 extends TestCase {
```

remyperona (approved):
just minor formatting

piotrbak (approved):
Worked as expected, tested scenarios:
- Should add the new string to the default list during the clean installation
- Should add the new string to the default list after the update when the list was default one
- Should not add the new string to the default list after the update when the string was already added there
- Should not add pixel caffeine script if fbq( script is not present on the list
- Should validate Delay JS is compatible with pixel caffeine

antoscarface:
Hi, I'm the Pixel Caffeine developer :) Just a heads up: first of all thank you for adding this into WP Rocket. Secondly, I've already fixed my self this into the latest version of Pixel Caffeine (before to notice this PR, I'm seeing it only now!) by splitting the main `<script>`, which include global variables definition and the Facebook Pixel snippet initialization, loading them separately, so WP Rocket will delay the snippet init, but not the global variable's definition, in order to prevent the `aepc_something not found` error.

Anyway, that's great you added directly the frontend.js file, it makes sense as well.
Thank you :)

PR #3105, opened 2020-09-16
Add excluded inline JS for urls_wpwidgetpolylang

nicomollet:
Polylang adds an inline JS like:
`
//<![CDATA[
var urls_wpwidgetpolylang = {"fr":"myfrpage","en":"myenpage"};
document.getElementById( "lang_choice_wp-widget-polylang" ).onchange = function() {
location.href = urls_wpwidgetpolylang[this.value];
}
`

It creates JS combine multiple files.
Can `urls_wpwidgetpolylang` be added to the ignored patterns default setting?

Thank you
FYI the doc https://docs.wp-rocket.me/article/1138-large-cache-folder-size-caused-by-dynamic-inline-js is really well done and helped a lot.


Tabrisrp (approved):

Tabrisrp:
Thank you for the PR, it makes sense to exclude this 👍🏼

PR #3116, opened 2020-09-17
Fixes #3100 Exclude RocketPreloadLinksConfig from combine JS

### LINKED ISSUE #3100: Exclude our scripts from combine JS and add them after the combined file
Exclude our 3 new scripts (browser checker, lazyload-scripts, and preload-links) from the Combine JS feature, but make sure it appears after the combined script in the DOM.

Tabrisrp:
This prevents it from being combined if the theme is not supporting the `html5` argument in its config.

### DIFF
--- a/inc/Engine/Optimization/Minify/JS/Combine.php
+++ b/inc/Engine/Optimization/Minify/JS/Combine.php
@@ -636,6 +636,7 @@ class Combine extends AbstractJSOptimization implements ProcessorInterface {
 			'archives-dropdown',
 			'loftloaderCache',
 			'SmartSliderSimple',
+			'RocketPreloadLinksConfig',
 		];
 
 		$excluded_inline = array_merge( $defaults, $this->options->get( 'exclude_inline_js', [] ) );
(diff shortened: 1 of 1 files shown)


engahmeds3ed (approved):

Tabrisrp:
@wp-media/qa ready for QA

Davidacu (changes requested):
- The rocket inline scripts are removed from combined JS file, however, it seems they are not added after the combined JS file in the DOM as requested in #3100 
- There is a console error coming from `rocket-preload-links-js-after` script
image
- The issue above breaks the preload link feature.

Tabrisrp:
@Davidacu Thanks, I changed the way the scripts are excluded

All scripts related to delay JS and preload links should stay where they are on the page, and not combined.

Ready for another QA

piotrbak:
@Tabrisrp In the issue description @engahmeds3ed mentioned that the scripts should be added after the combined file:
`but make sure it appears after the combined script in the DOM`

Has something changed since then?

Tabrisrp:
I think there was some kind of miscommunication about the subject at some point, but I was away when this was discussed.

Inline scripts to move after the combined JS files are scripts that depends on something that is part of the combined file (jQuery for example).

But in the instance of the preload links and delay JS, they are independant of anything combined. Everything is excluded with this PR, so there is no need to move them, they can remain where they are in the code.

piotrbak (approved):
Reproduced the issue, fix works as expected. Tested also known common Delayed JS scripts and they work too.

PR #3122, opened 2020-09-18
Fixes PHP Fatal error in 3.7.1 because of missing autoload for Cloudflare module

Tabrisrp:
The Cloudflare module PSR-4 is not in the main composer.json file, which is preventing the classes from being added to the autoloader, and causing a fatal error when the addon is activated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -78,7 +78,8 @@
 			"inc/classes/class-wp-rocket-requirements-check.php"
 		],
 		"psr-4": {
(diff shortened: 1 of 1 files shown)


iCaspar (approved):

Davidacu (approved):
I was able to reproduce the issue by enabling the Cloudflare addon in `master`
I can confirm the error is not longer reproducible in `hotfix/missing-autoload`

PR #3148, opened 2020-09-25
Fixes #3080 Update Trustpilot widget pattern for delay JS

### LINKED ISSUE #3080: Delay JS - TrustPilot additional scripts
Trustpilot is in our default list of "safe to delay" scripts. 

A customer is using the Trustpilot plugin: https://wordpress.org/plugins/trustpilot-reviews/

Turns out, this plugin loads its files everywhere, and even without authenticating a TrustPilot account.🙃 So, to reproduce: 

- Activate the plugin on a site (no need to connect to a TrustPilot account)
- Activate Delay JS with default list
- Visit any page. You'll see numerous JS errors in the console:https://jmp.sh/0n8Wf12
- Using [Vasilis' list of additional scripts to delay from his comment](https://github.com/wp-media/wp-rocket/issues/3080#issuecomment-692049990), will resolve the errors.

Plugin is installed on Smashing Coding.

Ticket: https://secure.helpscout.net/conversation/1277705875/193437?folderId=377611

I could not test this on Mega because the Trustpilot plugin requires an https connection.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

remyperona:
Fixes #3080 

Replaces `widget.bootstrap.com` by `widget.bootstrap.com/bootstrap` to be more accurate in the matching of the script during the delay JS process.

### Acceptance criteria
- The value should be updated in the default list
- The value should be updated when updating the plugin if the original value is present
- The value should not be added if the original value is not present
- There should no longer be a JS error on the frontend when using the trustpilot plugin


iCaspar (approved):
Not sure why this is buggered in Travis as composer wasn't changed in this set.

Tabrisrp:
Yeah I don't know why wpackagist seems to not be working suddenly.

@piotrbak ready for QA

piotrbak (approved):
Worked as expected.

PR #3164, opened 2020-09-28
Exclude jQuery from code.jquery.js from being deferred

vmanthos:
Exclude jQuery added from:
https://code.jquery.com/jquery/

**RegEx:** https://regex101.com/r/QMXIUi/1
**Ticket:** https://secure.helpscout.net/conversation/1291100148/197185


Tabrisrp (approved):

Tabrisrp:
@wp-media/qa Ready for QA

Acceptance criteria:
- When using jQuery from `code.jquery.com`, it should be correctly excluded from defer when jQuery safe mode is enabled
- It should also be excluded from combine JS

piotrbak (approved):
- All scripts are excluded from the Defer JS when the Safe Mode is enabled.
- All scripts are excluded from the Combination/Minification when the Safe Mode is enabled.

Issue #2664, opened 2020-05-18
Improve compatibility with Pagely


JohnTorvik:
WP Rocket is compatible with Pagely, but we received a refund request where the customer pointed:

- Pagely's caching system did not play nicely with WP Rocket. I then installed the WP Rocket disable caching add-on as well and still was having issues with Pagely

- WP Rocket does several thousand cache purge requests, impacting the server performance.

Ticket link: https://secure.helpscout.net/conversation/1166399003/165639/

GeekPress:
Another related ticket (and refund):
https://secure.helpscout.net/conversation/1114214010/152669/

GeekPress:
@arunbasillal Could you do some investigation about that? 

I don't know if we have a Pagely test account. Let me know if not.

ScottTravisHartley:
#1759 is relevant to this as well

GeekPress:
@ScottTravisHartley Thanks a lot for the reminder :)

@wp-media/wprocketplugin Do you have any idea?

hellofromtonya:
@Tabrisrp Can you look into this and provide feedback? Thank you.

webtrainingwheels:
FYI we have a testing account with Pagely now, [more info here](https://wp-media.slack.com/archives/G8L4EE8N9/p1595963468028200).
They are also doing some tests from their side.

klong-pagely:
Hello,

There is a hook for pagely_cache_purge_after to run a rocket_clean_domain. Then on rocket_clean_domain there is a hook for after_rocket_clean_domain to run a pagely cache purge.

The natural result of this is an infinite loop until the process is killed.

I created a simple pull request with a solution which sets a global variable that stops the purge after a single cycle.  See the pull request above for details.

Please take a look.

Thanks,
Keelan

DahmaniAdame:
Related ticket - https://secure.helpscout.net/conversation/1303211209/200535/

jorgemartine00:
Related ticket - https://secure.helpscout.net/conversation/1314607610/204012

Issue #1759, opened 2019-06-05
Cache Purge Timing Out On Pagely Hosting


ScottTravisHartley:
An interesting issue was occurring on a website that was running on Pagely that would cause the purge action to result in a time out error. This happened on a site where all plugins were disabled, and a default theme was being used. 

Pagely Support was contacted and a response was given regarding the issue. 

> This looks like an issue we've seen before with WPRocket where the plugin does an insane amount of PURGE requests to the server. Here's an excerpt of the logs, notice how all these requests are within the same second:

> This specific issue seems to be related to the rocket_purge_time_event wp-cron item. That possibly gets fired off when a post is saved.

> Specifically this function of the plugin code: https://github.com/wp-media/wp-rocket/blob/master/inc/3rd-party/hosting/pagely.php

> which seems it's firing off on a /wp-admin/ save post item, a cache purge, as well as being a wp-cron event.

> To verify, I'm suspecting that trying to save a post will time out/503 as well. We have a workaround plugin to stop the WP-rocket cache purges as we've seen something like this before. 

To resolve this, Pagely provided me with a custom plugin as follows:

```
add_filter('pagely_cache_purge_path', 'pagely_stop_homepage_wildcard_purges');

// To disable purging, "return null" from the filter. (Useful during batch imports, etc.)
function pagely_stop_homepage_wildcard_purges($url) {
    if ($url == "/(.*)") {
        // Only stop homepage from purging
        return null;
    } else {
        return $url;
    }
}
```

After adding it as a plugin and uploading it to the site the issue seems to have been resolved. I think there needs to be testing on Pagely to make sure the third-party compatibility file is working as expected.

PR #3166, opened 2020-09-28
Stop the infinite loop when purging cache on Pagely

klong-pagely:
This relates to the following two issues:

https://github.com/wp-media/wp-rocket/issues/2664
https://github.com/wp-media/wp-rocket/issues/1759

There is a hook for pagely_cache_purge_after to run a rocket_clean_domain.  Then on rocket_clean_domain there is a hook for after_rocket_clean_domain to run a pagely cache purge.  This causes a loop where it just keeps purging and purging until a timeout happens.  

This pull simply adds a global variable which is set when either hook is called that stops the purge from continuing beyond a single cycle.

### DIFF
--- a/inc/3rd-party/hosting/pagely.php
+++ b/inc/3rd-party/hosting/pagely.php
@@ -10,6 +10,13 @@ defined( 'ABSPATH' ) || exit;
  * @return void
  */
 function rocket_clear_cache_after_pagely() {
+        //  End the cycle of purges after a single purge has happened
+        if (!empty($GLOBALS['pagely_rocket_stop_cache_purge'])){
+            $GLOBALS['pagely_rocket_stop_cache_purge']=false;
+            return;
+        }
+        $GLOBALS['pagely_rocket_stop_cache_purge']=true;
+
 	// Clear all caching files.
 	rocket_clean_domain();
 }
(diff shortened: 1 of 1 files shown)


Tabrisrp (changes requested):
Thank you for the PR!

Though, we don't want to go the way of using global variables in our code, as much as possible. Globals are hard to test, and need to be avoided whenever we can.

Ideally this part of the code would be refactored to a hosting subscriber in ìnc/ThirdParty/Hostings`, and instead of using a global, we can use a class property to check if the purge already ran in one method or the other.

armanly:
Thank you for the update @Tabrisrp 

> Ideally this part of the code would be refactored to a hosting subscriber in ìnc/ThirdParty/Hostings`, and instead of using a global, we can use a class property to check if the purge already ran in one method or the other.

Just to clarify, are you asking Pagely to make those changes? We're happy to follow your feedback, just wanted clarification if that was something you're asking us to do or if you're able to take things from here.

We would really like to get this fixed soonish because it is causing issues for customers of WP Rocket using Pagely, as it causes purge calls to get stuck and it is generating support inquiries. If it will take substantially longer to refactor this the way you described, I would vote to remove what it does currently because it does not work correctly.

Thank you again and hope we can get this solved soon.

Tabrisrp:
Hey @armanly,

The refactoring would take significantly longer, as we would also be writing automated tests for it, as part of our CI process.

Our team is also very busy right now working on a whole new feature, which doesn't give us time to work on this issue properly.

If that's ok with you, we could start with removing the problematic code, and code back to it when we have more availability, to do the refactor and improvements discussed.

Could you do the code changes to remove the problematic parts?

klong-pagely:
Hello @Tabrisrp,

Updated pull request to simply comment out the code leading to a loop with a comment leading people back here to see your comment in case someone wants to take a stab at refactoring according to the wp-rocket guidelines.  This removes the infinite loop issue.

Downsides to this method:

-  If Pagely cache is cleared it won't clear wp-rocket cache.

Pro's

-  Left in place the Pagely cache clear after wp-rocket cache is cleared so a wp-rocket cache clear should work correctly.

Tabrisrp (approved):

PR #3185, opened 2020-10-05
User class to manage user data access

Tabrisrp:
This classes provides getter methods to get specific data from the user API, providing guard clauses in case some data is not defined.

### Methods
- `get_license_type()`: Gets the user license type, based on a numerical value
- `get_license_expiration()`: Gets the user license expiration timestamp
- `is_license_expired()`: Checks if the user license is expired
- `get_creation_date()`: Gets the user license creation date
- `is_auto_renew()`: Checks if auto-renew is enabled for this user license
- `get_upgrade_plus_url()`: Gets the URL to upgrade to plus
- `get_upgrade_infinite_url()`: Gets the URL to upgrade to infinite

### DIFF
--- a/inc/Engine/License/API/User.php
+++ b/inc/Engine/License/API/User.php
@@ -16,7 +16,7 @@ class User {
 	 * @param object $user The user object.
 	 */
 	public function __construct( $user ) {
-		$this->user = $user;
+		$this->user = is_object( $user ) ? $user : new \stdClass();
 	}
 
 	/**
@@ -25,11 +25,7 @@ class User {
 	 * @return int
 	 */
 	public function get_license_type() {
-		if (
-			! is_object( $this->user )
-			||
-			! isset( $this->user->licence_account )
-		) {
+		if ( ! isset( $this->user->licence_account ) ) {
 			return 0;
 		}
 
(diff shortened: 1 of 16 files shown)


inc/Engine/License/API/User.php
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	/**
+	 * Gets the user license expiration timestamp
+	 *
+	 * @return int
+	 */
+	public function get_license_expiration() {
+		if (
+			! is_object( $this->user )
```
hellofromtonya, line 46: As this check is repeating, how about if we create a private method to encapsulate the check? Then there's only 1 place where we need to change the code in the future (if needed).

hellofromtonya: Or another idea: 

- Check that it's an object in the constructor. 
- If no, instantiate to `stdClass`. 
- Then remove all of the `is_object` checks.

inc/Engine/License/API/User.php
```diff
@@ -0,0 +1,132 @@
...
+	 */
+	private $user;
+
+	/**
+	 * Instantiate the class
+	 *
+	 * @param object $user The user object.
+	 */
+	public function __construct( $user ) {
+		$this->user = $user;
```
hellofromtonya, line 19:
```suggestion
		$this->user = is_object( $user ) ? $user : new stdClass;
```

This step eliminates all of the `is_object` checks in the class.

hellofromtonya (approved):

PR #3195, opened 2020-10-08
Closes #3067 Upgrade from the plugin

### LINKED ISSUE #3067: Be able to upgrade from the plugin

With this enhancement, WP Rocket users with a Single or Plus license will be able to upgrade their license by clicking a link on the WP Rocket dashboard. They will be redirected to our checkout page. No payment will proceed from the customer website directly.

<p></p>

More info:

<https://docs.google.com/document/d/1g68IUjNW4dOiCe9QR0jWEkZpF6JMOK6hJhG40iL4_ds/edit#>

<p></p>

remyperona:
Closes #3067 

This change enables our users to see the upgrade options and be redirected directly to the upgrade checkout. A new section is added in the "My Account" section for single and plus users, to call for action, and a popin displays the upgrade choices

#### My account section
The sentence `You can use WP Rocket on more websites by upgrading your license (you will only pay the price difference between your current and new licenses). Upgrade Now` should be visible in this section when the license is not infinite.

#### Upgrade Popin
- The popin should be displayed when clicking on "Upgrade now"
- The popin should display 2 upgrade choices if the current license is single, with the accurate name, price, number of websites and upgrade URL
- The popin should display 1 upgrade choice if the current license is plus, with the accurate name, price, number of websites and upgrade URL
- The popin HTML code should not be present when on an infinite license, or when the license is expired


Tabrisrp:
@wp-media/qa Ready for QA, acceptance criteria in the original comment

piotrbak:
@Tabrisrp This is still just a prototype, right? Things like the wrong expiration date format, etc. should be omitted and I should be testing only the display of the message for different types of users, correct?
image

Tabrisrp:
You can ignore the values For `License` and `Expiration date` for now, I'll be fixing those in another PR specifically for that part

piotrbak:
@Tabrisrp There are some issues:
1. This notice is present in the logs. I'm changing the license type, removing the transient, and refreshing the page:
https://snippi.com/s/1rlzbsl
2. The upgrade links contain space: https://wp-rocket.me/checkout/upgrade/piotr@wp-media.me**%20**/371d3a01/infinite/
3. When someone has 2 single licenses, the `Upgrade Now` button is present but the window doesn't contain the Upgrade options: 
image How we should handle that?

Tabrisrp:
@piotrbak 
1. Is expected for now, I'll be fixing it later too
2. I'm taking the values directly from the API, is the link not working correctly? If yes, it's change that needs to be made on the API side
3. I just pushed a change to fix this one

piotrbak:
@Tabrisrp Yes, the link is taking me to the 404, no matter if with or without space:
image

piotrbak:
@Tabrisrp Could you also confirm, that the prices when upgrading from `2xSingle` should be exactly the same as when upgrading from `1x`?

Tabrisrp:
@piotrbak Yes, the price is the same going from 1 website or 2 websites, both are considered single level.

For the upgrade URL, I'm checking with Jo to see what's the issue. But consider that part as not critical, as once we fix it on the API side, it will be automatically fixed on the plugin side too.

piotrbak (approved):
@Tabrisrp So all the issues are not blockers, the last commit fixed the `2xSingle` license problem. The fails are related to:
1. PHP Notice https://snippi.com/s/1rlzbsl
2. Upgrade link not working

[testrail-report-41.pdf](https://github.com/wp-media/wp-rocket/files/5357091/testrail-report-41.pdf)

GeekPress:
@piotrbak The rewrite rule for this URL isn't set yet on our server. It explains the 404 page. I will do it Monday :)

PR #3196, opened 2020-10-09
Prevent duplicate addition of inline JS on multiple wp_enqueue_script calls

### LINKED ISSUE #3125: Calling wp_enqueue_scripts more than once can cause some of our inline scripts to be added twice

**Describe the bug**
When a 3rd party is re-calling `wp_enqueue_scripts` some of our scripts like lazyload-scripts.js will be enqueued twice, causing a JavaScript error like this:
` RocketLazyLoadScripts class is defined `

The same will happen if the Preload Links feature is enabled. 

For example the [Table of Contents Plus](https://wordpress.org/plugins/table-of-contents-plus/) plugin is calling `wp_enqueue_scripts` under certain circumstances [here](https://github.com/zedzedzed/table-of-contents-plus/blob/master/trunk/toc.php#L277)

**To Reproduce**
1. Enable **Table of Contents Plus**
2. Add a table of contents to the page
3. Enable DelayJS
4. See the console error

**Expected behavior**
Our scrips should be included only once.

**Additional context**
**Related ticket:** 
https://secure.helpscout.net/conversation/1278505927/193773?folderId=2683093

**Slack convo:**
https://wp-media.slack.com/archives/C43T1AYMQ/p1599768257124600

Per Tonya's feedback, we can check and guard our code from this happening.
There are 2 possible ways:
1 - `wp_script_is` [https://developer.wordpress.org/reference/functions/wp_script_is/]
2 - Set a property in the subscriber and check as a guard clause.

The 2nd approach is more performant. Both guard Rocket from 3rd parties doing weird stuff like firing the enqueue event more than once.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

vmanthos:
Fixes #3125 

Use a private property `is_enqueued` to check if the method has already been called once, in the delay JS and preload links subscribers.

### Acceptance criteria
- The inline JS scripts from delay JS and preload links are not inserted multiple times if `wp_enqueue_scripts` is called more than once

### DIFF
--- a/inc/Engine/Optimization/DelayJS/Subscriber.php
+++ b/inc/Engine/Optimization/DelayJS/Subscriber.php
@@ -23,6 +23,14 @@ class Subscriber implements Subscriber_Interface {
 	 * @var \WP_Filesystem_Direct
 	 */
 	private $filesystem;
+	
+	/**
+	 * Script enqueued status.
+	 *
+	 * @since 3.7
+	 *
+	 */
+	private $is_enqueued = false;
 
 	/**
 	 * Subscriber constructor.
@@ -70,6 +78,9 @@ class Subscriber implements Subscriber_Interface {
 	 * @return void
 	 */
 	public function add_delay_js_script() {
+		if ( $this->is_enqueued ) {
+			return;
+		}
 		if ( ! $this->html->is_allowed() ) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


Tabrisrp (approved):

Tabrisrp:
Just some coding standards issues to fix: https://travis-ci.com/github/wp-media/wp-rocket/jobs/397267577

Tabrisrp:
@wp-media/qa ready for QA

piotrbak (approved):
Worked as expected.
[testrail-report-40.pdf](https://github.com/wp-media/wp-rocket/files/5356628/testrail-report-40.pdf)

PR #3199, opened 2020-10-09
Fix #3140 : IE11 forEach issue on delayJS script

### LINKED ISSUE #3140: Delay JS - IE11: Object doesn't support property or method 'forEach'

**Describe the bug**

The ""rocket-delay-js-js-after" script we use contains a `forEach` method:
https://github.com/wp-media/wp-rocket/blob/7e8d54aaf5ad8ddffda404d7011931f1a43f3a53/assets/js/lazyload-scripts.js#L59

That's not supported on IE11 when using ES6, and results in the following error:
```JavaScript
Object doesn't support property or method 'forEach'
```
which breaks the feature.

I'm quoting @engahmeds3ed who was kind enough to clarify the "why" this doesn't work in IE11:

> forEach() is actually working on IE11, just be careful on how you call it.
> querySelectorAll() is a method which return a NodeList. And on Internet Explorer, foreach() only works on Array objects. (It works with NodeList with ES6, not supported by IE11).

**To Reproduce**

1. Enable **Delay JavaScript execution**.
2. Visit a site using IE11 on your PC or **browserstack.com** .
3. Open the JavaScript console.
4. See error.

**Expected behavior**

The feature should work on IE.

**Screenshots**

image

**Additional context**

**Related ticket:** https://secure.helpscout.net/conversation/1287835559/196361?folderId=2135277

**Potential solution:** https://rimdev.io/foreach-for-ie-11/

**IE11 worldwide market share:** [2.1% the previous year]( https://gs.statcounter.com/browser-version-market-share/desktop/worldwide/#monthly-201908-202008)

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

engahmeds3ed:
Fixes #3140 

- [x] Convert NodeList into an array with the slice.call() method so it can easily work with forEach

### Acceptance criteria
- No error in the console on IE11 when using the delay JS feature


Tabrisrp (approved):

engahmeds3ed:
I tested that change on Browsers:-

IE11
Chrome Version 85.0.4183.121 (Official Build) (64-bit)
Firefox 81.0 (64-bit)

assets/js/lazyload-scripts.js
```diff
@@ -54,7 +54,12 @@ class RocketLazyLoadScripts {
 	 * load the script.
 	 */
 	_loadScriptSrc() {
-		const scripts = document.querySelectorAll( `script[${ this.attrName }]` );
+		const scripts = Array.prototype.slice.call(
+			document.querySelectorAll(
+				`script[${ this.attrName }]`
+			)
```
hellofromtonya, line 60: This is safe to do here. Why? If there are no scripts with the `data-rocketlazyloadscript` attribute, `querySelectorAll` returns an empty `NodeList`. 

[Per the MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll#Return_value):

>Return Value
>A non-live NodeList containing one Element object for each element that matches at least one of the specified selectors or an empty NodeList in case of no matches.

However, it's not as efficient as it could be. Why? If there are no scripts, there's no need to do the slice or the `forEach`. Therefore, a tiny little micro-optimization would be:

```
const scripts = document.querySelectorAll( `script[${ this.attrName }]` );

if ( 0 === scripts ) {
     return;
}

Array.prototype.slice.call( scripts ).forEach( elem => {          
```

hellofromtonya (changes requested):
Tiny micro-optimization

assets/js/lazyload-scripts.js
engahmeds3ed: Good idea, I also called `this.reset();` before bailing out so removing the event listener as it's not needed anymore.

hellofromtonya (approved):

engahmeds3ed:
@wp-media/qa @piotrbak 
ready for QA, please test that on IE11, Chrome, Firefox, Safari, and Opera.

Thanks.

Tabrisrp:
@wp-media/qa ready for QA

piotrbak (approved):
Worked as expected, checked also on the Edge. Scripts delayed on each browser without errors in the console.

[testrail-report-42.pdf](https://github.com/wp-media/wp-rocket/files/5357252/testrail-report-42.pdf)

PR #3203, opened 2020-10-13
Fixes License data display on the dashboard

Tabrisrp:
This PR fixes an issue in the display of the license data on the dashboard, after the introduction of the code to manage the upgrade.

Some old code in the `Page` class was conflicting with the new one, and caused unexpected results.

In here, We use the new `UserClient` class in `Page` to get the data we need, and display it the same way as it was before.

Displayed result below:

license


inc/Engine/Admin/Settings/Page.php
```diff
@@ -75,25 +76,34 @@ class Page {
...
-		$this->title      = $args['title'];
-		$this->capability = $args['capability'];
-		$this->settings   = $settings;
-		$this->render     = $render;
-		$this->beacon     = $beacon;
-		$this->optimize   = $optimize;
+	public function __construct( $args, Settings $settings, Render_Interface $render, Beacon $beacon, Optimization $optimize, UserClient $user_client ) {
+		$this->slug        = $args['slug'];
+		$this->title       = $args['title'];
+		$this->capability  = $args['capability'];
```
hellofromtonya, line 101: We shouldn't assume that the given array of args contains these keys _when_ the class is available publicly for anyone to use. To help ensure we get what is expected:

1. Type hint to `array` for the `$args`
2. Merge the args with a default structure or use `isset` to valid or bail out if the structure isn't right

hellofromtonya (changes requested):

hellofromtonya (approved):

PR #3205, opened 2020-10-14
Fixes delay JS scripts textarea sanitization

remyperona:
The value passed to `rocket_sanitize_textarea_field()` was incorrect, and caused some of the default list to be stripped of the appropriate values.

The new value passed fixes the issue.


Tabrisrp:
Yes it's a non-issue. We can remove them the next time we update the default list.

piotrbak (approved):

PR #3217, opened 2020-10-16
Fix #3209 avada lazyload is detected as always active

### LINKED ISSUE #3209: Avada Lazyload is detected as always active

**Describe the bug**
Avada has added WordPress Lazyload option: https://jmp.sh/3SpVHLf

In our [compatibility file](https://github.com/wp-media/wp-rocket/blob/6c811c18c1739349a51753b07816b7260b1e2c07/inc/3rd-party/themes/avada.php#L22), we are checking Avada’s lazyload setting to be **empty**, and in that case, our Lazyload is available.

But now, the value is never empty, it’d be one of:
- none
- Wordpress
- avada

**To Reproduce**
1. Go to MEGA Avada
2. Deactivate Lazyload in Avada's settings, set it to "None": https://jmp.sh/3SpVHLf
3. Visit WP Rocket settings page
4. WPR Lazyload option is disabled: https://jmp.sh/7tNbWYk

**Expected behavior**
We should detect Avada's Lazyload **none** setting is deactivated, and enable WPR Lazyload option.

**Screenshots**
Avada's new option: https://jmp.sh/3SpVHLf

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

alfonso100:
Proposing a fix for [#3209](https://github.com/wp-media/wp-rocket/issues/3209)
(thank you @engahmeds3ed for the help!)

starting from Avada 7.2 version the `lazy_load` value is never empty, it’d be one of:
- none
- wordpress
- avada

Added verifications for the "none" value to deactivate and disable our LazyLoad option, while keeping the "empty" value verification for backward compatibility

### DIFF
--- a/inc/3rd-party/themes/avada.php
+++ b/inc/3rd-party/themes/avada.php
@@ -18,9 +18,12 @@ if ( 'Avada' === $current_theme->get( 'Name' ) ) {
 	 * @param string $value New Avada option value.
 	 * @return void
 	 */
-	function rocket_avada_maybe_deactivate_lazyload( $old_value, $value ) {
-		if ( empty( $old_value['lazy_load'] ) && ! empty( $value['lazy_load'] ) ) {
+	 function rocket_avada_maybe_deactivate_lazyload( $old_value, $value ) {
+		if ( (empty( $old_value['lazy_load'] ) && ! empty( $value['lazy_load'] ) ) 
+			|| ( 'none' === $old_value['lazy_load'] && 'none' !== $value['lazy_load'] ) ) {
+			
 			update_rocket_option( 'lazyload', 0 );
+			
 		}
 	}
 	add_action( 'update_option_fusion_options', 'rocket_avada_maybe_deactivate_lazyload', 10, 2 );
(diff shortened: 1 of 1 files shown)


Tabrisrp:
I'm not sure this is the best approach. When selecting `none` in Avada, it means "neither the WordPress lazyload, nor the Avada one", but it doesn't extend to WP Rocket itself.

How about deactivating WP Rocket lazyload only when the `Avada` value is selected?

This is the one value that _can_ conflict with our own lazyload, the other won't cause any issue.

inc/3rd-party/themes/avada.php
```diff
@@ -38,11 +38,11 @@ function rocket_avada_maybe_disable_lazyload() {
...
 	$current_theme = wp_get_theme();
 
-	if ( 'Avada' === $current_theme->get( 'Name' ) && ! empty( $avada_options['lazy_load'] ) ) {
-		return true;
+	if  ( ( 'Avada' === $current_theme->get( 'Name' ) ) && ( empty( $avada_options['lazy_load'] ) || 'avada' !== $avada_options['lazy_load']  ) ) {
+		return false;
 	}
 
-	return false;
+	return true;
```
Tabrisrp, line 45:
```
if  ( ( 'Avada' !== $current_theme->get( 'Name' ) ) {
    return false;
}

if ( empty( $avada_options['lazy_load'] ) ) {
        return false;
}

if (  ! empty( $avada_options['lazy_load'] && 'avada' !== $avada_options['lazy_load']  ) ) {
	return false;
}

return true;
```

inc/3rd-party/themes/avada.php
```diff
@@ -54,4 +54,4 @@ function rocket_avada_maybe_disable_lazyload() {
 function rocket_avada_clear_cache_fusion_patcher() {
 	rocket_clean_domain();
 }
-add_action( 'fusion_cache_reset_after', 'rocket_avada_clear_cache_fusion_patcher' );
+add_action( 'fusion_cache_reset_after', 'rocket_avada_clear_cache_fusion_patcher' );
```
Tabrisrp, line 57:
```suggestion
add_action( 'fusion_cache_reset_after', 'rocket_avada_clear_cache_fusion_patcher' );

```

Tabrisrp (changes requested):

Tabrisrp (approved):

alfonso100:
Tested this on MEGA/Avada, confirmed to be working as expected with Avada 7.2

PR #3222, opened 2020-10-19
Closes #3066 Auto-enable disable emoji

### LINKED ISSUE #3066: Auto-enable Disable Emoji

We are proposing to disable WordPress emoji automatically for all users. We will also remove the UI from WP Rocket. 

## Why are we doing this? 

Disable Emoji has been one of the “safe” options that is [<u>automatically enabled</u>](https://github.com/wp-media/wp-rocket/blob/f55ec0386e82a68394662b7b46b451ad124f18d7/tests/Fixtures/inc/admin/rocketFirstInstall.php#L10) on WP Rocket install. Based on mixpanel data, this option is currently enabled by 94% of our users. 

<p></p>

More importantly, the option generated just 12 tickets in the last 4 years. The impact of this feature is very low. 

<p></p>

It is our goal to clean up the WP Rocket user interface and make it even more user friendly. We want to take away as many questions and “decisions” from our users thus making WP Rocket faster and easier to configure. 

- We will remove the UI but keep the option as a “hidden” option in the database. (This is similar to what we did for [<u>SSL Cache</u>](https://github.com/wp-media/wp-rocket/issues/701))

- Existing users who disabled the option will continue to have it disabled on update. 

- New users will not have the option in the UI. 

- If new users run into an issue, we can provide them a helper plugin to disable the option. 

- Since very few percentage (6%) of our users have disabled the feature, and they probably disabled for good reason (the option is enabled by default), we will not be providing them with a “Tool” to enable the option. They can use a helper plugin to enable the option when needed.

## Technical Specs

Please refer the previous section. Additional details if any can be added by the dev team. 

## Risks

We are not expecting any issues since we preserve the existing state for current users. New users who face issues might reach out to support if they wish to use emoji’s. Even then, the feature is not likely to break any website and customers can refer to our documentation for the helper plugins. 

## Acceptance Criteria

- On a new install, WordPress emoji’s are disabled by default. There is no UI in WP Rocket.

- On update, if disable emoji was enabled before, it stays enabled. If disable emoji was disabled before it stays disabled. 

- The option is saved in the database as “emoji” and should be available when exporting WP Rocket settings. 

- We need to create a documentation about this “hidden” feature. 

- We need to create a helper plugin to enable and disable this feature. We will need one to disable the option and another to enable the option. (It can be combined into one; i.e. disables the feature on activation and enables it on deactivation or vice versa, but it might be counterintuitive to some customers. Will leave the final ddecision to Lucy and support team.)

Tabrisrp:
Closes "epic" #3066 

This PR removes the disable emoji option from the UI, while still keeping its value in the hidden fields. It also remove unnecessary code that had no impact on the front-end performance.

#### inc/Engine/Admin/Settings/Page.php
- Remove `emoji_section` and `emoji` indexes from the `media_section()` method
- Remove `emojis` from the `add_page_section()` `menu_description` in `media_section()` method
- Add `emoji` to the array in `hidden_fields()` method

#### inc/Engine/Media/Emojis/EmojisSubscriber.php
- deprecate `disable_emoji_tinymce()` method, and the associated filter. This methods has no impact on the front-end loading time
- in `disable_emoji()`, remove the following, which have no impact on the front-end pages:
    - `remove_action( 'admin_print_scripts', 'print_emoji_detection_script' );`
    - `remove_filter( 'the_content_feed', 'wp_staticize_emoji' );`
    - `remove_filter( 'comment_text_rss', 'wp_staticize_emoji' );`
    - `remove_filter( 'wp_mail', 'wp_staticize_emoji_for_email' );`

#### Tests
- Removed tests for deprecated method

### Acceptance criteria
- Remove the UI but keep the option as a “hidden” option in the database
- Existing users who disabled the option will continue to have it disabled on update
- New users will not have the option in the UI
- Optimization is the same as before on the front-end


hellofromtonya (approved):

Tabrisrp:
@wp-media/qa ready for QA

piotrbak:
@Tabrisrp You're merging it into `branch-3.8`, is this expected?

Tabrisrp:
yes it's a change for 3.8

piotrbak (approved):
- Should disable emoji by default
- Should be able to export and import it via settings
- Should not disable emoji on update if the option was disabled before
- Should disable emoji on update if the option was enabled before
[testrail-report-48.pdf](https://github.com/wp-media/wp-rocket/files/5408925/testrail-report-48.pdf)

PR #3224, opened 2020-10-20
Fix #3133 combine css/js increase cache/min size when versioning contains variable strings/timestamps

### LINKED ISSUE #3133: Combine CSS/JS increase cache/min size when versioning contains variable strings/timestamps

**Describe the bug**

When combining CSS/JavaScript files that use a timestamp as their version, e.g. 
```
https://www.example.com/wp-content/plugins/web-directory-pro-v2/assets/css/frontend.css?ver=1600668548
```
a different combined CSS/JavaScript file is created because of the way we currently handle the combined file's creation.

That can lead to an increased size of the `/cache/min/` folder.

We use a hash of the files that are to be combined:
https://github.com/wp-media/wp-rocket/blob/10a5ab2606759dd82d2a4c89732d6ef76ea4120d/inc/Engine/Optimization/Minify/CSS/Combine.php#L188

If the version of a file is different, even though the content is the same, the hash will be different, and a new combined file will be created because this conditional isn't evaluating to `true`:
https://github.com/wp-media/wp-rocket/blob/10a5ab2606759dd82d2a4c89732d6ef76ea4120d/inc/Engine/Optimization/Minify/CSS/Combine.php#L193

___
**UPDATE**

Upon further investigation, this is an issue related to the fact that in `3.7` **we are using the URLs of files including their version**:
```HTML
//www.example.com/wp-content/themes/visionclimax/style.css?ver=1601532890
```

In `3.6.4` we used to use their file path without the version:
```HTML
/home/www.example.com/public_html/wp-content/themes/twentytwenty/style.css
```

**To Reproduce**

1. Make sure you are using WP Rocket 3.7.
2. Add timestamp versioning to a CSS/JS file.
3. Enable CSS/JS combination.
4. Visit different pages or preload cache.
5. A different CSS/JS combined file will be created for each page.

**Expected behavior**

Avoid the increase of the `cache/min` size increase when timestamp versioning is used.

**Additional context**

Currently, handling this involves excluding the culprit files from CSS/JS combination.

**Screenshots** 

**Related tickets**
* https://secure.helpscout.net/conversation/1279890356/194227/
* https://secure.helpscout.net/conversation/1268568522/190972/

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

alfonso100:
Proposing a fix for [#3133](https://github.com/wp-media/wp-rocket/issues/3133). 
**Combine CSS/JS increase cache/min size when versioning contains variable strings/timestamps**

Applying Remy's proposed fix here: [#3133 (comment)](https://github.com/wp-media/wp-rocket/issues/3133#issuecomment-703837954)

The fix is applied in the minify CSS and JS files, and also in the Combine CSS and JS files

### Acceptance criteria
- An URL using a query string version number randomly generated (or with the current timestamp) should not create duplicate combine/minify files

### DIFF
--- a/inc/Engine/Optimization/Minify/CSS/Combine.php
+++ b/inc/Engine/Optimization/Minify/CSS/Combine.php
@@ -121,7 +121,7 @@ class Combine extends AbstractCSSOptimization implements ProcessorInterface {
 			$this->styles[ $style['url'] ] = [
 				'type' => 'internal',
 				'tag'  => $style[0],
-				'url'  => $style['url'],
+				'url'  => strtok( $style['url'], '?' ),
 			];
 		}
 
--- a/inc/Engine/Optimization/Minify/CSS/Combine.php
+++ b/inc/Engine/Optimization/Minify/CSS/Combine.php
@@ -98,7 +98,7 @@ class Combine extends AbstractCSSOptimization implements ProcessorInterface {
 				$this->styles[ $style['url'] ] = [
 					'type' => 'external',
 					'tag'  => $style[0],
-					'url'  => rocket_add_url_protocol( $style['url'] ),
+					'url'  => rocket_add_url_protocol( strtok( $style['url'], '?' ) ),
 				];
 
 				continue;
--- a/inc/Engine/Optimization/Minify/CSS/Minify.php
+++ b/inc/Engine/Optimization/Minify/CSS/Minify.php
@@ -44,7 +44,7 @@ class Minify extends AbstractCSSOptimization implements ProcessorInterface {
 				continue;
 			}
 
-			$minify_url = $this->replace_url( $style['url'] );
+			$minify_url = $this->replace_url( strtok( $style['url'], '?' ) );
 
 			if ( ! $minify_url ) {
 				Logger::error(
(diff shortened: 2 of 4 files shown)


Tabrisrp (approved):

Tabrisrp:
@wp-media/qa ready for QA

piotrbak:
@Tabrisrp @alfonso100 @vmanthos I wasn't able to reproduce this problem for JavaScript Combination. The problem is here for:
-JS Minification
-CSS Minification
-CSS Combination

Tests:
https://jmp.sh/yLyB3mW 

In this case, should we include this change in the JavaScript Combination?

piotrbak:
PR passed the QA, @Tabrisrp let me know what do you think about this JS Combination change. If it's going to be left like this, I'll approve it immediately.

Tabrisrp:
For combine JS, the filename is generated based on the content. So the change here doesn't have any impact, good nor bad, so it can be left as is for consistency with the rest.

piotrbak (approved):
- Should not create separate cached JS files when random query string is applied
- Should combine external JS files with .JS extension and inline scripts
- Should combine external CSS files with .CSS extension
- Should not create separate combined CSS files when random query string is applied
- Should not combine JS files when they have been excluded from Combine JS
- Should not combine CSS files when they have been excluded from Combine CSS
- Should not create separate cached CSS files when random query string is applied
- Should not create separate combined JS files when random query string is applied

[testrail-report-50.pdf](https://github.com/wp-media/wp-rocket/files/5411312/testrail-report-50.pdf)

PR #3225, opened 2020-10-20
Fixes #3119 Update adsbygoogle value in default delay JS list

### LINKED ISSUE #3119: DelayJS default adsbygoogle keyword breaking Adsense functionality on Newspaper Theme

**Describe the bug**
The **Newspaper** theme creates its responsive google Ads spots using Javascript and `document.write`, like this: 
[https://jmp.sh/YH2vvdQ
](https://jmp.sh/YH2vvdQ)

When our **DelayJS** feature is enabled, we rewrite this code because it contains the **adsbygoogle** keyword and encode it as base64: https://jmp.sh/zKLJaU5 

So, the ad spots doesn’t exist in the HTML, triggering an error like this for each ad spot:
https://jumpshare.com/v/Mj2lb7BBWwSuucmb2TO7 

**To Reproduce**
1. On Newspaper theme, set some ad sizes
2. The theme will inject the ads like this: https://jmp.sh/YH2vvdQ
3. Enable DelayJS
4. The ads won't be created because the containers won't be generated: https://jmp.sh/zKLJaU5 

**Expected behavior**
We should exclude this inline Javascript from the DelayJS feature

**Additional context**
If I replace the DelayJS keyword to use **googlesyndication** instead of **adsbygoogle**  everything works as expected, the JS is delayed, and the ad containers are intact: https://jumpshare.com/v/APJLAmvb50QTsRbNyYQW

The specific console error is: 
`Object { message: "adsbygoogle.push() error: All ins elements in the DOM with class=adsbygoogle already have ads in them.", stack: "O@https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js:69:574\nZl@https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js:228:210\nTl@[.....]\nm.ba@https://pagead2.googlesyndication.com/pagead/js/adsbygoogle.js:64:294\225\n", name: "TagError", pbr: true }
adsbygoogle.js:216:110`

Related Issue: 
https://github.com/wp-media/wp-rocket/issues/3085

HelpScout Ticket:
https://secure.helpscout.net/conversation/1282673567/194973/

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

Tabrisrp:
Fixes #3119 

Change `adsbygoogle` to `adsbygoogle.js` in the default delay JS list, and update the old value to the new one during plugin update if the value was in the field.

### Acceptance criteria
- The value is updated on plugin update if the original value is in the list
- The new value is added when using the `restore defaults`
- The delay works on the frontend with the new value


iCaspar (approved):

Tabrisrp:
@wp-media/qa ready for QA

piotrbak (approved):
-Should validate that updated AdSense exclusion doesn't cause conflicts
-Should not add the string when the old one wasn't present on the list
-Should update the string when the old one was present before
-Should update the string in the list during the clean installation
[testrail-report-49.pdf](https://github.com/wp-media/wp-rocket/files/5411283/testrail-report-49.pdf)

Issue #3118, opened 2020-09-17
3.7: External CSS Minify exclusions are not working as intended


webtrainingwheels:
The specs for excluding an external CSS file from minify are: 

>3rd Party: Use either the full URL path or only the domain name, to exclude external CSS. 

However there are some cases where this is not working as expected. 

Samples: 
- https://use.fontawesome.com/releases/v5.14.0/css/all.css
Had to be excluded like this: /releases/v5.14.0/css/all.css

- https://cloud.typography.com/7231672/6270372/css/fonts.css
Had to be excluded like this: /7231672/6270372/css/fonts.css

Seems only to be an issue with Minify, not Combine

Tickets: 
- https://secure.helpscout.net/conversation/1281498132/194658?folderId=3572910
- https://secure.helpscout.net/conversation/1278389385/193723?folderId=2675957

**To Reproduce**
1. Include one of the above URLS in the HTML of a site
2. Enable Minify CSS
3. Try to exclude the file with either the domain or full URL

**Expected behavior**
The files should be excluded using either the domain or full URL

**Backlog Grooming (for WP Media dev team use only)**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

vmanthos:
**Related ticket:** https://secure.helpscout.net/conversation/1310458122/202765?folderId=2135277

I further investigated this.

This is because of the  `preg_match()` we run here:
https://github.com/wp-media/wp-rocket/blob/6c811c18c1739349a51753b07816b7260b1e2c07/inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php#L141

If the excluded file is e.g. `cdn.jsdelivr.net/npm/bootstrap@4.5.3/dist/css/bootstrap.min.css`, the `$file_path` is `/npm/bootstrap@4.5.3/dist/css/bootstrap.min.css`, and `preg_match()` returns `0`.

Hence the file is _not_ excluded and is cached locally.

vmanthos:
**Related ticket:** https://secure.helpscout.net/conversation/1308669776/202288/

iCaspar:
Can Reproduce [X]

Identified cause: per @vmanthos, the regex fails to identify domain names. [X]

Scope Solution:
- [ ] Add testcases for domain names and full URLs to fixtures for `inc/Engine/Optimization/Minify/CSS/Minify`
[...]

PR #3236, opened 2020-10-22
Fix #3121: Local cache - minified 3rd-party resources having the integrity attribute are blocked

### LINKED ISSUE #3121: Local cache - minified 3rd-party resources having the integrity attribute are blocked

**Describe the bug**

Since 3.7 WP Rocket saves locally 3rd-party JavaScript/CSS files.

If the original URL contains the `integrity` attribute, and CSS/JavaScript minification is enabled, this leads to that resource being blocked, since the file's hash doesn't match with what's saved in the `integrity`.

**To Reproduce**

1. Load Bootstrap's[ CSS/JavaScript files](https://getbootstrap.com/docs/4.5/getting-started/introduction/).
2. Enable CSS/JavaScript minification, NOT combination.
3. Check the JavaScript console for errors.

**Expected behavior**

The files should load fine when saved locally, despite the `integrate` attribute.

Approved ideas for solution:
- https://github.com/wp-media/wp-rocket/issues/3121#issuecomment-697616840
- https://github.com/wp-media/wp-rocket/issues/3121#issuecomment-697972206

**Screenshots**

**Additional context**

A few thoughts on this:

If there is an `integrity` attribute in the URL we can:
1. remove it when we add that file from the local cache or
2. [calculate the hash](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) of the local file and replace the value of the attribute.

To serve the original purpose of the `integrity`, **1** doesn't look right, and **2** would most likely require a check of the original file's integrity.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

engahmeds3ed:
Fixes #3121 

In the `AssetsLocalCache` class, add 2 new methods:
- `has_integrity()`: checks if the current processed CSS/JS tag has the `integrity` attribute. Returns boolean
- `is_integrity_valid()`: checks if the content integrity corresponds to the expected value. Use the `integrity` attribute value to determine which algorithm was used to created the hash (sha256/384/512) Use the same algorithm on the downloaded content to verify that the value we generated is the same as the one in the attribute. Returns boolean, false if the check failed, true if it passed.

If the integrity check fails, bail-out and don't process the file. If it's successful, remove the integrity attribute from the tag.

For the above, we will have to modify all the minify/combine CSS/JS classes.


inc/Engine/Optimization/AssetsLocalCache.php
```diff
@@ -109,4 +113,55 @@ protected function write_file( $content, $file ) {
...
+	 * @param string $asset Link HTML to be tested.
+	 *
+	 * @return array|false Matched array with integrityhashmethod, integrityhash keys.
+	 */
+	private function has_integrity( $asset ) {
+		if ( ! preg_match( '#\s*integrity\s*=[\'"](?<integrityhashmethod>.*)-(?<integrityhash>.*)[\'"]#Ui', $asset, $integrity_matches ) ) {
+			return false;
+		}
+
+		if ( ! isset( $integrity_matches['integrityhashmethod'] ) || ! isset( $integrity_matches['integrityhash'] ) ) {
```
Tabrisrp, line 131:
```suggestion
		if ( ! isset( $integrity_matches['integrityhashmethod'], $integrity_matches['integrityhash'] ) ) {
```

You can group multiple checks in one `isset` call, it makes it easier to read

inc/Engine/Optimization/AssetsLocalCache.php
```diff
@@ -109,4 +113,55 @@ protected function write_file( $content, $file ) {
...
+		}
+
+		$content      = $this->get_content( $asset_matched['url'], false );
+		$content_hash = base64_encode( hash( $integrity_matches['integrityhashmethod'], $content, true ) );// phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode
+
+		if ( $integrity_matches['integrityhash'] !== $content_hash ) {
+			return false;
+		}
+
+		return str_replace( $integrity_matches[0] . '', '', $asset_matched[0] );
```
Tabrisrp, line 164: Why is there a need for a concatenation with an empty string `$integrity_matches[0] . ''`?

inc/Engine/Optimization/AssetsLocalCache.php
```diff
@@ -45,16 +45,18 @@ public function __construct( $cache_path, $filesystem ) {
 	 * @since 3.1
 	 *
 	 * @param string $url URL to get the content from.
+	 * @param bool   $save Save the resource locally or just get contents.
+	 *
 	 * @return string
 	 */
-	public function get_content( $url ) {
+	public function get_content( $url, $save = true ) {
```
Tabrisrp, line 52: Instead of using a new parameter, how about splitting the method in 2:
- Keep this method as is
- Create a new method `get_raw_content( $url )`, which would always fetch the remote URL instead of checking if we have a local cache first
- Then use this new method inside `get_content()` for when the content needs to be fetched

That way we avoid a boolean parameter, which is usually discouraged (breaks single responsability), and split the complexity into 2 methods instead of one.

inc/Engine/Optimization/AssetsLocalCache.php
```diff
@@ -109,4 +113,55 @@ protected function write_file( $content, $file ) {
...
+		if ( false === $integrity_matches ) {
+			return $asset_matched[0];
+		}
+
+		// validate the hash algorithm.
+		if ( ! in_array( $integrity_matches['integrityhashmethod'], hash_algos(), true ) ) {
+			return false;
+		}
+
+		$content      = $this->get_content( $asset_matched['url'], false );
```
Tabrisrp, line 157: Use the new `get_raw_content()` method proposed above here

remyperona (changes requested):

remyperona (approved):

remyperona (approved):

Tabrisrp:
@wp-media/qa ready for QA

piotrbak:
@Tabrisrp @engahmeds3ed When trying to exclude the external assets the host part is stripped here (`$file_path`):
https://github.com/wp-media/wp-rocket/blob/e49355167e75f1d3cf51ee65c429b75f3e28682e/inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php#L141

It means, that in order to exclude:
`maxcdn.bootstrapcdn.com/bootstrap/4.0.0/css/bootstrap.min.css`
we need to add:
`/bootstrap/4.0.0/css/bootstrap.min.css`

Second example:
`mega.wp-rocket.me/wp-content/uploads/integrity/integrity-test-file.css`
the `$filepath`:
`/wp-content/uploads/integrity/integrity-test-file.css`

What do you think about this?

Tabrisrp:
@piotrbak Related to this issue #3118?

piotrbak (approved):
Worked as expected, the only problem was related to the already existing issue:
https://github.com/wp-media/wp-rocket/issues/3118
[testrail-report-52.pdf](https://github.com/wp-media/wp-rocket/files/5440560/testrail-report-52.pdf)

PR #3240, opened 2020-10-23
Closes #3168 Add the server protocol to our analytics

### LINKED ISSUE #3168: Add HTTP protocol version to Rocket Analytics
To help us to make some important decisions related to combine options, we will need to know the HTTP protocol version used by ours users.

Having the version like 1.1, 2 and 3 will be enough.

Tabrisrp:
Closes #3168 

### Code changes
Send the server protocol to our analytics. We parse the `SERVER_PROTOCOL` value to only keep the version number based on the `http\/(?.*)$` pattern.

### Acceptance criteria
The server protocol is correctly sent to Mixpanel, using the `http_protocol` index, and the version number as value


inc/admin/admin.php
```diff
@@ -397,10 +397,11 @@ function rocket_analytics_data() {
...
 
-	$theme              = wp_get_theme();
-	$data               = array_diff_key( get_option( WP_ROCKET_SLUG ), $untracked_wp_rocket_options );
-	$locale             = explode( '_', get_locale() );
-	$data['web_server'] = 'Unknown';
+	$theme                 = wp_get_theme();
+	$data                  = array_diff_key( get_option( WP_ROCKET_SLUG ), $untracked_wp_rocket_options );
+	$locale                = explode( '_', get_locale() );
+	$data['web_server']    = 'Unknown';
+	$data['http_protocol'] = '';
```
engahmeds3ed, line 404: I think no need to use this initialization as always the function `wp_get_server_protocol` will return valid value.

engahmeds3ed (approved):
Only one optimization

Tabrisrp:
@wp-media/qa ready for QA

piotrbak:
@Tabrisrp Could you give me access to the Mixpanel?

Tabrisrp:
You should have it in Bitwarden

piotrbak:
@Tabrisrp I'm wondering how accurate data we're going to receive. 

We'll get the HTTP/1 protocol for all sites using Cloudflare:
image
Probably because all connections between Cloudflare and the origin server are done via HTTP/1:
https://support.cloudflare.com/hc/en-us/articles/200168076-Are-the-HTTP-2-or-SPDY-protocols-supported-between-Cloudflare-and-the-origin-server-#6ncFUWOVRaVtPzYN1euBIC

We'll get the same results for the servers when there's Nginx as a reverse proxy for Apache working.

I also tested it also on the HTTP3:
image

It's not even possible to get an HTTP/3 protocol from this function, so it'll become a default HTTP/1.0:
https://developer.wordpress.org/reference/functions/wp_get_server_protocol/

When it comes to the sending data itself. I can see that the protocol data is present in the sent array, but I'm not able to find it in the Mixpanel. It's not present in the list:
image
Do you know if I need to set something up there?

Tabrisrp:
@GeekPress From QA, accuracy of the data for this might not be good, do we still want to collect it?

GeekPress:
@Tabrisrp Indeed if the accuracy is bad, we don't have the interest to collect it. Let's forget that for now?

ping @arunbasillal :)

arunbasillal:
Really amazing QA there @piotrbak 

I agree with @GeekPress Bad accuracy can lead to bad data and bad judgements. Better not to have it. A lot of users use Cloudflare, so it's not negligible.

PR #3248, opened 2020-10-27
Closes #2893 Create a REST API support endpoint to get support data

### LINKED ISSUE #2893: Automatically fetch customer's site/WP Rocket information for support
When customers submit tickets from the plugin interface, Beacon provides all this info which greatly assists with support. 

We need a way to do the same when the customer submits a ticket from the support form on our website. We don't want them to have to check boxes or copy/paste info which is annoying and could be inaccurate. 

We have to automatically retrieve: 

- List of active plugins
- List of WP Rocket settings
- WP Rocket version
- WordPress version
- Site URL
- Theme
- Hosting company (per @DaniJacome) 

This is related to the support form for the new website.

remyperona:
Closes #2893 

### What does it do?
With this PR, a new REST API POST endpoint `wp-rocket/v1/support` is added. This endpoint provides some useful support information, that will be used to populate Helpscout when opening a new ticket via the support form on the new website.

The endpoint will return the data only:
- If a valid key and email corresponding to the website license holder are provided along the request
- If the request comes from our website

### Code changes
- New classes in `Engine\Support` to manage data collection and REST registration
- Update `Beacon` class to use the same data collection class as a dependency

### Acceptance criteria
- Errors are returned if the above conditions are not met when doing the endpoint request
- The support data is returned when the above conditions are met when doing the endpoint request


hellofromtonya (approved):

piotrbak:
@Tabrisrp Using Postman to perform the check. Our website is still not sending that request to the customers, right?

The postman tests look good. My only concern is that we can send this request from, for example:
`wp-rocket.me.com` website and it'll go through the `strpos`.

Do we want to be that detailed?

Tabrisrp:
Yes the website is not yet sending the requests, it will be on the new one.

I'd say it's fine, the referer check is an additional measure, but he can't be relied on too much either (the header itself). The important part is the key/email validation.

piotrbak (approved):
@Tabrisrp Worked as expected, the only thing is that the referer header can not be trusted very much.
[testrail-report-54.pdf](https://github.com/wp-media/wp-rocket/files/5459590/testrail-report-54.pdf)

@GeekPress It will definitely need testing on the website side when implemented (one site that belongs to two accounts, and so on)

PR #3250, opened 2020-10-27
Closes #3247 Update minimum PHP version to 7.0

### LINKED ISSUE #3247: Bump minimum support versions to PHP 7.0
We would like to bump the minimum required PHP version to 7.0.
We have today only 1.9% of our users using PHP 5.6.

Tabrisrp:
Closes #3247 

Update minimum supported PHP version to 7.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,10 +25,8 @@ matrix:
     env: WP_VERSION=latest
   - php: 7.2
     env: WP_VERSION=latest
-  - php: 5.6
-    env: WP_VERSION=latest
   - name: Legacy
-    php: 5.6
+    php: 7.0
     env: WP_VERSION=5.2.7
   - name: Coding Standards
     php: 7.3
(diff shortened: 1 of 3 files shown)


piotrbak:
@Tabrisrp after the update, when the customer was on the PHP5.6 we are not letting access the WP Rocket settings but our optimizations are still applied to the frontend, is that correct?

Would it be possible that after the update some of our code will not be backward compatible and cause a fatal error on the frontend?

Tabrisrp:
It's correct in a way since the advanced-cache.php is still there.

Right now there is no code that is not compatible with PHP 5.6, so there is no risk of creating an error.

piotrbak:
The change is minimum and it works as expected. My concern is that someone in the future will update from WP Rocket 3.7.5, to let's say 3.9 and we will have some code that's not PHP5.6 compatible then.

What do you think?

Tabrisrp:
At the time the compatibility check is performed, we don't have anything else loaded, so we can't remove the advanced-cache.php file for example.

PR #3254, opened 2020-10-29
Fix CSS/JS minification when the theme's name contains a space

### LINKED ISSUE #3178: CSS/JS combination breaks when the theme's folder contains a space
- Made sure you’re on the latest version: This was tested against `3.7.2` ✅ 

**Describe the bug**

When the theme's folder name contains a space, e.g. `twenty twenty`, the theme's files aren't combined, while their URLs are removed from the source code.

This leads to broken styling/functionality.

When the issue occurs, the `$filepath` here is empty:
https://github.com/wp-media/wp-rocket/blob/443f2e1b902c592bd517cbaa641571bb2ddce1a5/inc/Engine/Optimization/Minify/CSS/Combine.php#L283

because `rocket_url_to_path()` returns `false` here:
https://github.com/wp-media/wp-rocket/blob/443f2e1b902c592bd517cbaa641571bb2ddce1a5/inc/functions/formatting.php#L512

The following file path is an example that returns `false` when passed to the `is_readable()` method.
```
/home/domains/example.com/public_html/wp-content/themes/twenty%20twenty/style.css
```

@alfonso100 noted the following:
_In `3.6.4`, these CSS files coming from a theme folder containing spaces, are automatically excluded from being combined: [http://recordit.co/AFuaw5L88l](http://recordit.co/AFuaw5L88l)_

_Javascript files are not combined in any version (`3.7.2` and `3.6.4`) when the folder contains spaces._

_In `3.7.2` after updating the plugin, or activating any option in WP Rocket the error is present: the CSS fles coming from the theme are stripped from the combine…. But, if you clear the cache, the CSS file is excluded as we used to do in the previous version: [http://recordit.co/zjC6U3l8hB](http://recordit.co/zjC6U3l8hB)
And if you enable another option, the files are removed from the combine… but if you clear the cache, the CSS files are excluded and present again_ 

**To Reproduce**

1. Rename a theme folder so it contains a space.
2. Enable CSS minification/combination
3. Visit the front-end and search for a theme-specific class.
4. It shouldn't be there.

**Expected behavior**

CSS/JavaScript combination should work even when there are spaces in the theme's folders.

**Screenshots**

**Screencast of the issue:** [https://youtu.be/8ubrkkR-K_k](https://youtu.be/8ubrkkR-K_k)

**Additional context**

**Conversation in Slack:** https://wp-media.slack.com/archives/C43T1AYMQ/p1601535360090600?thread_ts=1601407796.062500&cid=C43T1AYMQ

**Related ticket:** https://secure.helpscout.net/conversation/1294708853/198137?folderId=2135277

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

vmanthos:
Fixes #3178

### DIFF
--- a/inc/Engine/Optimization/Minify/CSS/Minify.php
+++ b/inc/Engine/Optimization/Minify/CSS/Minify.php
@@ -60,7 +60,7 @@ class Minify extends AbstractCSSOptimization implements ProcessorInterface {
 
 			$style['final'] = $integrity_validated;
 
-			$minify_url = $this->replace_url( strtok( $style['url'], '?' ) );
+			$minify_url = $this->replace_url( rawurldecode( strtok( $style['url'], '?' ) ) );
 
 			if ( ! $minify_url ) {
 				Logger::error(
--- a/inc/Engine/Optimization/Minify/CSS/Minify.php
+++ b/inc/Engine/Optimization/Minify/CSS/Minify.php
@@ -60,7 +60,7 @@ class Minify extends AbstractCSSOptimization implements ProcessorInterface {
 
 			$style['final'] = $integrity_validated;
 
-			$minify_url = $this->replace_url( rawurldecode( strtok( $style['url'], '?' ) ) );
+			$minify_url = $this->replace_url( strtok( $style['url'], '?' ) );
 
 			if ( ! $minify_url ) {
 				Logger::error(
@@ -134,7 +134,7 @@ class Minify extends AbstractCSSOptimization implements ProcessorInterface {
 
 		$unique_id     = md5( $url . $this->minify_key );
 		$filename      = preg_replace( '/\.(css)$/', '-' . $unique_id . '.css', ltrim( rocket_realpath( $parsed_url['path'] ), '/' ) );
-		$minified_file = $this->minify_base_path . $filename;
+		$minified_file = rawurldecode( $this->minify_base_path . $filename );
 		$minify_url    = $this->get_minify_url( $filename, $url );
 
 		if ( rocket_direct_filesystem()->exists( $minified_file ) ) {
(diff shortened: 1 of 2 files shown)


remyperona (approved):

Tabrisrp:
@wp-media/qa ready for QA

piotrbak:
The resources are loaded correctly during the minification and combination. Tested cases:
[testrail-report-55.pdf](https://github.com/wp-media/wp-rocket/files/5460368/testrail-report-55.pdf)

@Tabrisrp @vmanthos On the other hand, spaces are not allowed in the `href` attributes, is that something we should care about, especially when this seems to be some kind of edge case?
image

vmanthos:
> spaces are not allowed in the href attributes, is that something we should care about

Our customers are doing this, i.e. using themes with spaces in their name, despite the specs.

If we do not fix this:
* WP Rocket will not work as expected.
* We are going to have tickets for this.

For me, it's a one-way street.

Tabrisrp:
it's likely we will have complain about it if we go that way.

@vmanthos 

Alternate solution: in CSS\Minify and JS\Minify, remove the changes made.

Instead use `rawurldecode()` on the `$minified_file` variable in the `replace_url()` method. That way, the minified files path will be correctly using a space instead of the `%20` characters, and the URL can remain untouched and valid.

I tested this quickly, and it seems to work.

piotrbak:
@vmanthos Yes, I know this is a weird setup... The thing is that when WordPress is handling it without WP Rocket, the spaces are replaced by `%20`, so the output is correct HTML:
image

remyperona (approved):

Tabrisrp:
@piotrbak can you re-test this with the new changes

piotrbak (approved):
@Tabrisrp Worked as expected now.

[testrail-report-58.pdf](https://github.com/wp-media/wp-rocket/files/5466396/testrail-report-58.pdf)

PR #3260, opened 2020-11-03
Fixes #3033 Don't display RocketCDN campaign banner when campaign is not active

### LINKED ISSUE #3033: RocketCDN campaign name display only if is_discount_active and end_date are valid
Currently RocketCDN will display the campaign name if the campaign is filled and do not take into consideration the `end_date` and `is_discount_active`.

So we end up in the situation when the campaign is expired or not active, however in WP Rocket the Campaign Name is still displayed only because this field is filled. 

The issue is in here because it checks only the campaign_name. Instead it should check also that the campaign is active and end_date is still valid.
https://github.com/wp-media/wp-rocket/blob/7e8d54aaf5ad8ddffda404d7011931f1a43f3a53/views/settings/rocketcdn/cta-big.php#L20

remyperona:
Fixes #3033 

### Acceptance criteria
- The campaign banner should not be displayed when either `is_discount_active` is `false`, or the end date is in the past, even if the campaign name is not empty


piotrbak:
@Tabrisrp The `is_discount_active` is now empty in the response from API, not false, I assume that the promotion is over. This is the response I received:
`   [is_discount_active] =>
    [discounted_price_monthly] => 6.99
    [discounted_price_yearly] => 95.88
    [discount_campaign_name] => 
    [end_date] => 2020-11-03
    [monthly_price] => 7.99
    [annual_price] => 79.99`

piotrbak:
@Tabrisrp If the following conditions are met, we're displaying the discounted price but not the banner:
-`is_discount_active` = true
-`discount_campaign_name` = empty 
-`end_date` in the future
image

Is it something we should take care of in the code or it won't ever happen? 🤔

Tabrisrp:
If a campaign is active, there should always be a campaign name associated with it. If it's missing, it's an input error on our side when updating the pricing endpoint. So it should never happen.

piotrbak (approved):
Worked as expected:
[testrail-report-60.pdf](https://github.com/wp-media/wp-rocket/files/5488727/testrail-report-60.pdf)

PR #3262, opened 2020-11-03
Fixes #3257 Update license promotion CSS

### LINKED ISSUE #3257: Changes to Promo notice
**Notice bubble:** 
Change the color of the bubble when a promotion is active. Now it’s a red bubble with an exclamation point.To make it less alarming I’d like to change the color to green.

**Countdown timer:**
According to the design, It should be made of white squares but right now they all have different width.

Capture d’écran 2020-11-02 à 12 02 15
Capture d’écran 2020-11-02 à 12 02 49

Tabrisrp:
Fixes #3257 

Update the CSS for the promotion bubble background color and the countdown items width.

### Acceptance criteria
- Promotion bubble background color is now green
- Countdown items are the same width (though items are flexible when changing screen size)


piotrbak (approved):
Works as expected. Changing the screen might cause the squares to be flexible though. Seems that we've no other possibility there and it's mentioned in the acceptance criteria:
image

PR #3263, opened 2020-11-03
Closes #3243 Disable auto-updates for WP Rocket

### LINKED ISSUE #3243: After updating, WordPress Automatic updates link is not available on last version

**Describe the bug**
if you have the last version of any plugin (WPR on his case), the link for Enable auto-updates on plugins page will disappear but if there is an update for the plugin, this link will appear.
So we need always to be able to enable/disable it regardless there is an update or not.

**To Reproduce**
1. Install WP Rocket the last version.
2. Open Plugins admin page to see list of available plugins.
3. Check `Automatic Updates` column for WP Rocket row.

**Expected behavior**
We need to be able to see the enable/disable link with last installed version.

**Screenshots**
With last version 3.7.4
image

With the version before 3.7.3
image

**Additional context**
We have two options here:-
1. We can force this auto update to be always disabled by using this filter auto_update_plugin as documented [here](https://wordpress.org/support/article/configuring-automatic-background-updates/#plugin-theme-updates-via-filter).

and this may be [here at this class](https://github.com/wp-media/wp-rocket/blob/5434080ccd89b36f3312846e4288a50c9d6b7c3e/inc/classes/subscriber/Plugin/class-updater-subscriber.php#L117)

```
public function force_disable_autoupdate( $update, $item ) {
	if ( 'wp-rocket/wp-rocket.php' === $item->plugin ){
		return false;
	}
	return $update;
}
```
and the output will be like:
image

2. Make this feature enabled/disabled by the user:
we need to adjust two things here:-
**a.** change [this code](https://github.com/wp-media/wp-rocket/blob/5434080ccd89b36f3312846e4288a50c9d6b7c3e/inc/classes/subscriber/Plugin/class-updater-subscriber.php#L247-L249) to be:

```
// If a newer version is available, add the update.
if ( version_compare( $this->plugin_version, $remote_data->new_version, '<' ) ) {
	$transient_value->response[ $remote_data->plugin ] = $remote_data;
}else{
	$transient_value->no_update[ $remote_data->plugin ] = $remote_data;
}
```

**b.** adjust [this method code](https://github.com/wp-media/wp-rocket/blob/5434080ccd89b36f3312846e4288a50c9d6b7c3e/inc/classes/subscriber/Plugin/class-updater-subscriber.php#L312) to handle the API response when there is no update and return the object again not error (I sent @Tabrisrp to ask him about the following API response if it's expected or not and he told me that it's valid one)
```
{"success":false,"data":{"reason":"NO_UPDATE","consumer_key":"f760c170","consumer_email":"ahmed@wp-media.me"}}
```
with response code 202 but the code handles not 200 response to be error.

@wp-media/productrocket I think you need to take decision on which approach is needed.
@Tabrisrp to be validated from technical side, if u have any moment.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [ ] Estimate the effort

Tabrisrp:
Closes #3243 

Disable auto-updates for WP Rocket, as discussed in the related issue.

### Acceptance criteria
- `Auto-updates disabled` should be displayed in the automatic updates column, when the row is for WP Rocket.

### DIFF
--- a/inc/classes/subscriber/Plugin/class-updater-subscriber.php
+++ b/inc/classes/subscriber/Plugin/class-updater-subscriber.php
@@ -264,7 +264,7 @@ class Updater_Subscriber implements Subscriber_Interface {
 	 * @since 3.7.5
 	 *
 	 * @param bool|null $update Whether to update. The value of null is internally used to detect whether nothing has hooked into this filter.
-	 * @param object $item The update offer.
+	 * @param object    $item The update offer.
 	 * @return bool|null
 	 */
 	public function disable_auto_updates( $update, $item ) {
--- a/inc/classes/subscriber/Plugin/class-updater-subscriber.php
+++ b/inc/classes/subscriber/Plugin/class-updater-subscriber.php
@@ -2,23 +2,21 @@
 namespace WP_Rocket\Subscriber\Plugin;
 
 use WP_Rocket\Event_Management\Subscriber_Interface;
+use WP_Rocket\Traits\Updater_Api_Tools;
 
 /**
  * Manages the plugin updates.
  *
  * @since  3.3.6
- * @author Grégory Viguier
  */
 class Updater_Subscriber implements Subscriber_Interface {
-	use \WP_Rocket\Traits\Updater_Api_Tools;
+	use Updater_Api_Tools;
 
 	/**
 	 * Full path to the plugin.
 	 *
 	 * @var    string
 	 * @since  3.3.6
-	 * @access private
-	 * @author Grégory Viguier
 	 */
 	private $plugin_file;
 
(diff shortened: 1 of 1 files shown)


engahmeds3ed (approved):
why travis failed on preloading!

engahmeds3ed:
travis failed because `https://smashingcoding.com` is not available now.

Tabrisrp:
@wp-media/qa Ready for QA

piotrbak (approved):
The text is displayed correctly.

[testrail-report-62.pdf](https://github.com/wp-media/wp-rocket/files/5489405/testrail-report-62.pdf)

Issue #1803, opened 2019-06-20
Conflict between Divi video module & Replace YouTube iframe with preview image option


WordPresseur:
When you enable Divi video module sub-option called **Overlay** and you set an **Overlay image** and that the **Replace YouTube iframe with preview image** option is enable, you can’t play the video anymore.

Ticket: https://secure.helpscout.net/conversation/882236070/112834?folderId=1213662

WordPresseur:
**Describe the bug**
When you use the Divi video module and you set an Overlay image, if the Replace YouTube iframe with preview image option is enable, you can’t play the video anymore.

**To Reproduce**
Steps to reproduce the behavior:
1. Create a page with a **Divi video module** & add an **Overlay** image to it like here: https://share.getcloudapp.com/v1urQg4j
2. It works fine as you can see here https://share.getcloudapp.com/qGuox7B0
3. Activate the **Replace YouTube iframe with preview image** sub-option in Media tab.
4. Test the page, it doesn’t work anymore as you can see here https://share.getcloudapp.com/bLum5KLr

**Expected behavior**
I was expecting the video to load.

Tabrisrp:
**Reproduce the issue** ✅ 
Reproduced on http://mega.wp-rocket.me/divi/iframe-ll-test/

**Find the root cause** ✅ 
This is happening because our replacement is removing the `iframe` element from the page, and the Divi JavaScript code is looking for it to launch the video.

**Scope a solution** ✅ 
Unfortunately, right now we don't have a way to make those compatible. The solution:

We can auto-disable the option when we detect the current theme is Divi to prevent the issue, it will make the iframe lazyloaded normally instead of using the replacement.

**Estimate the effort** ✅ 
Effort `[S]`
- Move our Divi compatibility file to the new architecture
- Add a callback method to disable the replacement option when Divi is the current theme
- Disable the checkbox on the settings page, with a message explaining why
- Write tests

WordPresseur:
another case: https://secure.helpscout.net/conversation/1192067794/172127/

kierantaylorio:
Related ticket: https://secure.helpscout.net/conversation/1230856807/181473/

SimonThSorcerer:
[...]

PR #3294, opened 2020-11-11
Fixes #3293 Apply our divi compatibility for parent and child theme

### LINKED ISSUE #3293: Divi child theme issue with divi video module + overlay image when wp-rocket "Replace youtube iframe" option is enabled
Divi video is not working when we are using a child theme as parent theme the "Divi" and we have the wp-rocket -> media -> "Replace Youtube iframe with preview image" enabled and the divi video module settings has enabled the overlay image

**Steps to reproduce the issue:**
1. Create a child theme as parent the Divi theme

i.e.
```
/*
 Theme Name: Test child theme
 Description: This is a test child theme for Divi
 Template: Divi
 Text Domain: test-divi-theme
*/
```
2. Switch to this new child theme
3. Go to admin dashboard -> wp-rocket -> Media, enable the "Enable for iframes and videos" and also the checbox below "Replace Youtube iframe with preview image"
4. Create a new page, add a video module, add an overlay image for this video module
5. Preview this new page

**Expected behavior**
The video to work correctly 

**Actual behavior**
The video is not working, when we click the play button it just takes us to the top of the page and we get a js error in the console log.
Below is the js error:
```
 Uncaught TypeError: Cannot read property 'play' of undefined
    at window.et_pb_play_overlayed_video (57142510dd94504b2fb78c81aaedafca.js:216)
    at HTMLDivElement.<anonymous> (57142510dd94504b2fb78c81aaedafca.js:216)
    at HTMLDivElement.dispatch (jquery.js?ver=1.12.4-wp:3)
    at HTMLDivElement.r.handle (jquery.js?ver=1.12.4-wp:3)
```

**Additional comments**
I noticed that you fixed this issue on this PR #2875 but actually for us is not fixed because the actual template name(folder name) is "Divi" and not 'divi' as is the actual fix in your code.

File of the fix, Divi.php:121
`return ( 'Divi' === $theme->get( 'Name' ) || 'divi' === $theme->get_template() );`
This can be fixed with two ways, either by changing the 'divi' to 'Divi' or converting the $theme->get_template() to lower case
but i am not sure there is a reason to convert the template_name to lower case since the actual folder name of the template is "Divi" so i believe the best would be to just change the 

`'divi' === $theme->get_template()`

to 

`'Divi' === $theme->get_template() `

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

mitsios:
Based on my comments from the issue #3293

We are using a child theme that has as parent theme "Divi" but we are facing issues with the divi video module when we have enabled the overlay image

The old fix from the PR #1803 for this issue was checking for the template_name 'divi' all in lower case but actually the Divi folder name is 'Divi' so the fix was not working for the child themes.

I renamed it from 

`return ( 'Divi' === $theme->get( 'Name' ) || 'divi' === $theme->get_template() );`
to

`return ( 'Divi' === $theme->get( 'Name' ) || 'Divi' === $theme->get_template() );`


inc/ThirdParty/Themes/Divi.php
```diff
@@ -118,6 +118,6 @@ public function add_divi_to_description( $disable_youtube_lazyload ) {
 	private static function is_divi( $theme = null ) {
 		$theme = $theme instanceof WP_Theme ? $theme : wp_get_theme();
 
-		return ( 'Divi' === $theme->get( 'Name' ) || 'divi' === $theme->get_template() );
+		return ( 'Divi' === $theme->get( 'Name' ) || 'Divi' === $theme->get_template() );
```
Tabrisrp, line 121:
```suggestion
		return ( 'divi' === strtolower( $theme->get( 'Name' ) ) || 'divi' === strtolower( $theme->get_template() ) );
```

With this change, we force the returned values to be lowercase, eliminating the naming issue of the parent/child theme. This improves consistency on our side of the code.

remyperona (changes requested):

PR #3300, opened 2020-11-13
3118 Fix external css minify exclusions

### LINKED ISSUE #3118: 3.7: External CSS Minify exclusions are not working as intended

iCaspar:
Per #3118 -
[X] - Add testcases for external css with URL in domains
[X] - Modify regex to make tests pass


inc/Engine/Optimization/Minify/CSS/Minify.php
```diff
@@ -32,6 +33,19 @@ public function optimize( $html ) {
 		}
 
 		foreach ( $styles as $style ) {
+			foreach ( $this->get_excluded_externals() as $excluded ) {
```
Tabrisrp, line 36: 2 suggestions: 
- Initialize the excluded outside of the loops, that way the method is only called once
- Instead of a second loop, how about imploding the array into a string pattern, and using `preg_match()` to do the matching in the URL?

inc/Engine/Optimization/Minify/CSS/Minify.php
```diff
@@ -323,9 +347,30 @@ protected function minify( $file_path, $minified_file, $file_content ) {
...
+		/**
+		 * Filters CSS external URLs to exclude from the minify process
+		 *
+		 * @since 3.7.6
+		 *
+		 * @param array $pattern Patterns to match.
+		 */
+		$excluded_externals = (array) apply_filters( 'rocket_minify_css_excluded_external', [] );
+
+		return array_merge( $excluded_externals, $this->options->get( 'exclude_css', [] ) );
```
engahmeds3ed, line 374: I don't like using array_merge as it's slow, can u check the following and test it.
```suggestion
		return [ ...$excluded_externals, ...$this->options->get( 'exclude_css', [] ) ];
```
This code is using [Spread Operator](https://wiki.php.net/rfc/spread_operator_for_array) which I think it's included from PHP 5.6 (If I remember well)

and also do we need to make this array unique?
if Yes, I think using `array_flip` twice will do the trick faster.

@iCaspar What do u think?

remyperona (changes requested):

engahmeds3ed (changes requested):

inc/Engine/Optimization/Minify/CSS/Minify.php
iCaspar: Found a way to eliminate this method altogether :)

inc/Engine/Optimization/Minify/CSS/Minify.php
iCaspar: Moved this check into `AbstractCSSOptimization::is_minify_excluded_file()`.

remyperona (approved):

Tabrisrp:
@wp-media/qa ready for QA

piotrbak (approved):
Worked as expected:
[Uploading testrail-report-71.pdf…]()

PR #3308, opened 2020-11-16
Exclude stylesheets with media queries from combine CSS

### LINKED ISSUE #3206: CSS files with media queries are not excluded during CSS combination

**Describe the bug**

Currently, when combining CSS files we check for media queries here:
https://github.com/wp-media/wp-rocket/blob/443f2e1b902c592bd517cbaa641571bb2ddce1a5/inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php#L124

We do not exclude `link` elements containing media queries like:
* `max-width`
* `min-width`

For example, the following file will be combined:
```HTML
<link rel='stylesheet' media='screen and (max-width: 900px)' href='css/medium.css' />
```

When that happens, we do not wrap the file's content around the media query, which would be the ideal solution. 
Instead, we add it directly to the combined file.

The rules contained there may be applied to elements regardless of what the original media query dictated.

**To Reproduce**
1. Add a CSS file using a media query, e.g. `max-width`. In that, change, e.g. the color of `h1`.
2. Enable CSS combination.
3. Visit the site. The rule from step 1 will be applied regardless of the width of the viewport.

**Expected behavior**

Files containing such media queries should be either automatically excluded or combined while wrapped in the original media query.

**Additional context**

**Related ticket:** https://secure.helpscout.net/conversation/1305858288/201378?folderId=2135277

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

engahmeds3ed:
Fixes #3206 

We need to handle the following cases

**Excluded from combine CSS feature**

```
<link rel='stylesheet' media='screen AND (max-width: 900px)' href='http://www.example.org/assets/css/test.css' />
<link rel='stylesheet' media='screen AND (min-width: 500px)' href='http://www.example.org/assets/css/test.css' />
<link rel='stylesheet' media='screen not (min-width: 500px)' href='http://www.example.org/assets/css/test.css' />

<link rel='stylesheet' media='all AND (max-width: 900px)' href='http://www.example.org/assets/css/test.css' />
<link rel='stylesheet' media='all AND (min-width: 500px)' href='http://www.example.org/assets/css/test.css' />
<link rel='stylesheet' media='all not (min-width: 500px)' href='http://www.example.org/assets/css/test.css' />
```

**Included into combine CSS feature**

```
<link rel='stylesheet' media='screen, (max-width: 900px)' href='http://www.example.org/assets/css/test.css' />
<link rel='stylesheet' media='all, (max-width: 900px)' href='http://www.example.org/assets/css/test.css' />

<link rel='stylesheet' media='(max-width: 900px), screen' href='http://www.example.org/assets/css/test.css' />
<link rel='stylesheet' media='(max-width: 900px), all' href='http://www.example.org/assets/css/test.css' />

<link rel='stylesheet' media='screen' href='http://www.example.org/assets/css/test.css' />
<link rel='stylesheet' media='all' href='http://www.example.org/assets/css/test.css' />
<link rel='stylesheet' href='http://www.example.org/assets/css/test.css' />
```

**Note:** Comma(,) means OR [Documentation here](https://www.w3schools.com/tags/att_link_media.asp)

### DIFF
--- a/inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php
+++ b/inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php
@@ -121,7 +121,7 @@ abstract class AbstractCSSOptimization extends AbstractOptimization {
 			return true;
 		}
 
-		if ( false !== strpos( $tag[0], 'media=' ) && ! preg_match( '/media=["\'](?:\s*|[^"\']*?\b(?:\s*?,\s*?)?(all|screen)(?:\s*?,\s*?)?)["\']/i', $tag[0] ) ) {
+		if ( false !== strpos( $tag[0], 'media=' ) && ! preg_match( '/media=["\'](?:\s*|[^"\']*?\b(?:\s*?,\s*?)?(all|screen)(?:\s*?,\s*?[^"\']*)?)["\']/i', $tag[0] ) ) {
 			return true;
 		}
 
--- a/inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php
+++ b/inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php
@@ -121,10 +121,6 @@ abstract class AbstractCSSOptimization extends AbstractOptimization {
 			return true;
 		}
 
-		if ( false !== strpos( $tag[0], 'media=' ) && ! preg_match( '/media=["\'](?:\s*|[^"\']*?\b(?:\s*?,\s*?)?(all|screen)(?:\s*?,\s*?)?)["\']/i', $tag[0] ) ) {
-			return true;
-		}
-
 		$file_path = wp_parse_url( $tag['url'], PHP_URL_PATH );
 
 		// File extension is not css.
--- a/inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php
+++ b/inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php
@@ -121,7 +121,7 @@ abstract class AbstractCSSOptimization extends AbstractOptimization {
 			return true;
 		}
 
-		if ( false !== strpos( $tag[0], 'media=' ) && ! preg_match( '/media=["\'](?:\s*|[^"\']*?\b(all|screen)\b[^"\']*?)["\']/i', $tag[0] ) ) {
+		if ( false !== strpos( $tag[0], 'media=' ) && ! preg_match( '/media=["\'](?:\s*|[^"\']*?\b(?:\s*?,\s*?)?(all|screen)(?:\s*?,\s*?)?)["\']/i', $tag[0] ) ) {
 			return true;
 		}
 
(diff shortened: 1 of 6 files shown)


remyperona (changes requested):
Can you add unit/integration tests cases to validate the change?

remyperona (approved):

Tabrisrp:
@wp-media/qa ready for QA

piotrbak:
@engahmeds3ed @Tabrisrp Is it expected that the files will be also excluded from the minification? Not sure if the last update caused this problem, but the files are not minified/hosted locally:
https://rocketlabsqa.dev/combine-css-media-queries-excluded/

inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php
```diff
@@ -121,14 +121,6 @@ protected function is_minify_excluded_file( array $tag ) {
 			return true;
 		}
 
-		if ( false !== strpos( $tag[0], 'media=' ) && ! preg_match( '/media=["\'](?:\s*|[^"\']*?\b(all|screen)\b[^"\']*?)["\']/i', $tag[0] ) ) {
```
Tabrisrp, line 124: keep this line to avoid processing stylesheets for `print` or `speech`

remyperona (changes requested):

inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php
engahmeds3ed: Ok and what about this
```
if ( false !== strpos( $tag[0], 'only screen and' ) ) {
			return true;
		}
```

remyperona (approved):

inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php
```diff
@@ -121,11 +121,7 @@ protected function is_minify_excluded_file( array $tag ) {
...
-			return true;
-		}
-
-		if ( false !== strpos( $tag[0], 'only screen and' ) ) {
+		if ( false !== strpos( $tag[0], 'media=' ) && ! preg_match( '/media=["\'](?:\s*|[^"\']*?\b(?:\s*?,\s*?)?(all|screen)(?:\s*?,\s*?)?)["\']/i', $tag[0] ) ) {
```
Tabrisrp, line 124:
```suggestion
		if ( false !== strpos( $tag[0], 'media=' ) && ! preg_match( '/media=["\'](?:\s*|[^"\']*?\b(all|screen)\b[^"\']*?)["\']/i', $tag[0] ) ) {
```

remyperona (changes requested):

inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php
engahmeds3ed: So we removed any check on media attribute for minify only and kept this check only for combining (local and external files)

piotrbak (approved):
Worked as expected:
[testrail-report-72.pdf](https://github.com/wp-media/wp-rocket/files/5569711/testrail-report-72.pdf)

PR #3309, opened 2020-11-16
Prevent combination of excluded files from defer JS

remyperona:
With this changes, files excluded from defer JS are also excluded from combine JS, to make sure the exclusion is preserved.

Previously, only jQuery files were excluded from combine JS. Now, all files in defer JS exclusion will be.

### DIFF
--- a/inc/Engine/Optimization/Minify/JS/Combine.php
+++ b/inc/Engine/Optimization/Minify/JS/Combine.php
@@ -178,18 +178,7 @@ class Combine extends AbstractJSOptimization implements ProcessorInterface {
 							}
 						}
 
-						if (
-							! empty( $this->excluded_defer_js )
-							&&
-							preg_match( '#(' . $this->excluded_defer_js . ')#i', $matches['url'] )
-						) {
-							Logger::debug(
-								'Script is excluded from defer JS.',
-								[
-									'js combine process',
-									'tag' => $matches[0],
-								]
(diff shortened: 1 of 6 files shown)


inc/Engine/Optimization/Minify/JS/Combine.php
```diff
@@ -202,6 +212,21 @@ function( $script ) {
 						return;
 					}
 
+					if (
```
engahmeds3ed, line 215: This part is being used twice, one inside the condition checking if url is external and one if not external.
so this code will be executed twice when url is external, right?

engahmeds3ed (changes requested):

inc/Engine/Optimization/Minify/JS/Combine.php
Tabrisrp: It happens only once for each type: once for external, once for internal.

engahmeds3ed: If so, I think making it inside a new method would be better

engahmeds3ed: sorry I just saw the return keyword,fine.
it's minor change, I will approve PR and you can take the decision of adding a new local method for these two checks ot not

engahmeds3ed (approved):

PR #3325, opened 2020-11-19
Fix Missing Width And Height Attributes On Images - Frontend [2/2]

### LINKED ISSUE #3307: Add Missing Width And Height Attributes On Images - Frontend [2/2]
Sub-task for epic https://github.com/wp-media/wp-rocket/issues/3301

## Frontend

- Move `https://github.com/wp-media/wp-rocket/blob/master/inc/front/images.php` into new structure `/inc/Engine/Media/Images/ImagesSubscriber.php` 
- In case the option `image_dimensions` is enabled it will apply this change.
- The filter 'rocket_specify_image_dimensions' still works and enables the feature even if the UI is disabled. (To use the filter add add_filter( 'rocket_specify_image_dimensions', '__return_true' ); to functions.php)
- Modify the ServiceProvider.php to provide the ImagesSubscriber
- Deprecate `rocket_specify_image_dimensions()`
- Revisit the compatibility with  LayerSlider - https://github.com/wp-media/wp-rocket/blob/3b8b2d7debc83641045f9af82baa08b14e41daed/inc/3rd-party/plugins/slider/layerslider.php#L10-L13
- When the option is enabled, any images with missing “width” and “height” attributes have the appropriate width and height added. 
- Write Unit & Integration tests

engahmeds3ed:
Fixes #3307


crystinutzaa (changes requested):
This seems fine, however I think it still needs to handle the integration with `LayerSlider`
- Revisit the compatibility with  LayerSlider - https://github.com/wp-media/wp-rocket/blob/3b8b2d7debc83641045f9af82baa08b14e41daed/inc/3rd-party/plugins/slider/layerslider.php#L10-L13

engahmeds3ed:
Thanks @crystinutzaa 
@wp-media/php  I adjusted the code and changed the layerslider code a bit please check.
Also I added integration test case that uses the fixture related to the main image dimensions feature to assure that all cases not changed (input images = output images) without applying the feature of image dimensions even if it was active.
I don't know if we need it or not but I added it so feel free to advise about that.

inc/Engine/Media/Images/Frontend.php
```diff
@@ -0,0 +1,306 @@
...
+
+		// Get all images without width or height attribute.
+		preg_match_all( '/<img(?:[^>](?!(height|width)=[\'\"](?:\S+)[\'\"]))*+>/i', $html, $images_match );
+
+		if ( empty( $images_match ) ) {
+			return $html;
+		}
+
+		$replaces = [];
+		$images   = apply_filters( 'rocket_specify_dimension_images', $images_match[0] );
```
Tabrisrp, line 72: Missing docblock for the filter

inc/Engine/Media/Images/Frontend.php
```diff
@@ -0,0 +1,306 @@
...
+		return str_replace( content_url(), rocket_get_constant( 'WP_CONTENT_DIR' ), $url );
+	}
+
+	/**
+	 * Check if we can specify dimensions for external images.
+	 *
+	 * @return bool Valid to be parsed or not.
+	 */
+	private function can_specify_dimensions_external_images() {
+		return ini_get( 'allow_url_fopen' ) && apply_filters( 'rocket_specify_image_dimensions_for_distant', true );
```
Tabrisrp, line 186: Missing docblock for filter

inc/Engine/Media/Images/Frontend.php
```diff
@@ -0,0 +1,306 @@
...
+		 *
+		 * @since 2.2
+		 *
+		 * @param bool Do the job or not.
+		 */
+		if ( ! apply_filters( 'rocket_specify_image_dimensions', true ) ) {
+			return false;
+		}
+
+		return true;
```
Tabrisrp, line 251:
```suggestion
		return apply_filters( 'rocket_specify_image_dimensions', true );
```

inc/Engine/Media/Images/Frontend.php
```diff
@@ -0,0 +1,306 @@
...
+		if ( $this->is_external_file( $image_url ) ) {
+			if ( ! $this->can_specify_dimensions_external_images() ) {
+				return false;
+			}
+
+			if ( ! $this->image_exists( $image_url, true ) ) {
+				return false;
+			}
+
+			return getimagesize( $image_url );
```
Tabrisrp, line 294: why not return directly the dimensions index instead of returning the whole array from `getimagesize()`?

inc/Engine/Media/Images/Frontend.php
```diff
@@ -0,0 +1,306 @@
...
+	 * @return false|string false if we can't specify for this image otherwise get img src attribute.
+	 */
+	private function can_specify_dimensions_one_image( $image ) {
+		// Don't touch lazy-load file (no conflict with Photon (Jetpack)).
+		if (
+			false !== strpos( $image, 'data-lazy-original' )
+			||
+			false !== strpos( $image, 'data-no-image-dimensions' )
+			||
+			! preg_match( '/\s+src\s*=\s*[\'"]([^\'"]+)/i', $image, $src_match )
```
Tabrisrp, line 268:
```suggestion
			! preg_match( '/\s+src\s*=\s*[\'"](?<url>[^\'"]+)/i', $image, $src_match )
```

inc/Engine/Media/Images/Frontend.php
```diff
@@ -0,0 +1,306 @@
...
+		) {
+			return false;
+		}
+
+		return $src_match[1];
```
Tabrisrp, line 273:
```suggestion
		return $src_match['url'];
```

Named capturing groups help to make the code more readable

inc/ThirdParty/Plugins/Slider/LayerSlider.php
```diff
@@ -0,0 +1,34 @@
+<?php
```
Tabrisrp, line 1: Did you test if we still need this compatibiliy file? Seeing how old the original code is, a lot can have happened since

Tabrisrp (changes requested):

inc/ThirdParty/Plugins/Slider/LayerSlider.php
engahmeds3ed: @Tabrisrp Yes I tried that locally and I can see that layerslider alreay uses img tag with dimensions already set as in the following screenshot.

image

So our code won't affect it at all, even our code affected it and added dimensions, this won't make a conflict with layerslider.

@wp-media/php what do u think?

Tabrisrp: Ok, so we don't need the compatibility file, right?

inc/Engine/Media/Images/Frontend.php
```diff
@@ -0,0 +1,328 @@
...
+		return $src_match['url'];
+
+	}
+
+	/**
+	 * Get Image sizes.
+	 *
+	 * @param string $image_url Image url to get sizes for.
+	 *
+	 * @return array|false Get image sizes otherwise false.
```
Tabrisrp, line 292:
```suggestion
	 * @return string|false Get image sizes otherwise false.
```

With the change, it's a string instead of an array, right?

Tabrisrp (changes requested):

inc/Engine/Media/Images/Frontend.php
engahmeds3ed: Yes exactly, Done

inc/ThirdParty/Plugins/Slider/LayerSlider.php
engahmeds3ed: @Tabrisrp Yes I think so.

Tabrisrp: great, so let's just remove it along the old compatibility file 👍🏼

engahmeds3ed: Ok will so that now.

Tabrisrp (approved):

crystinutzaa (approved):

Tabrisrp (approved):

Issue #3264, opened 2020-11-04
Issue when Google fonts using different APIs are combined


DahmaniAdame:
**Before submitting an issue please check that you’ve completed the following steps:**
- [x] Made sure you’re on the latest version
- [x] Used the search feature to ensure that the bug hasn’t been reported before

**Describe the bug**
Google fonts API v2 has introduced a new syntax and the main detail causing the issue on our end is that they now use semicolons to separate style instead of comas.

V1 - https://fonts.googleapis.com/css?family=Poppins:wght@100,200,300,400,500,600,700,800,900
V2 - https://fonts.googleapis.com/css2?family=Poppins:100;200;300;400;500;600;700;800;900

What happens is that if both APIs are used, our combined Google fonts is delivering the V2 fonts in a way that is not compatible with V1 (which is the one we use when delivering combined fonts).

https://fonts.googleapis.com/css?family=Lato:wght@100;300;400;700;900|Poppins:wght@100;200;300;400;500;600;700;800;900|Merriweather:400,700,900,400italic,700italic,900italic|Montserrat:400,700|Inconsolata:400&subset=latin,latin-ext&display=swap

As a result, in the combined file above, we have a lot of Lato and Poppins styles not loading.

**To Reproduce**
1. Add Google fonts using V1 and V2 APIs
2. Use the Google fonts optimization feature
3. See the error on the combined file

**Expected behavior**
We need a way to normalize fonts to use one API version before combining them.
Or we need to set a combined file per API version.

**Screenshots**
N/A

**Additional context**
APIs documentation: 
- V1 - https://developers.google.com/fonts/docs/getting_started
- V2 - https://developers.google.com/fonts/docs/css2

**Backlog Grooming (for WP Media dev team use only)**
- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

DahmaniAdame:
Related tickets: 

- https://secure.helpscout.net/conversation/1298260935/199079 
- https://secure.helpscout.net/conversation/1324384282/207231

DahmaniAdame:
Related tickets: 

- https://secure.helpscout.net/conversation/1298260935/199079 
- https://secure.helpscout.net/conversation/1324384282/207231

crystinutzaa:
### Reproduce the problem ✅ 
Reproduced on local installation

### Identify the root cause ✅ 
[...]

PR #3329, opened 2020-11-19
Google Fonts API Enhancement, Part 1 of 4 - New Class

iCaspar:
Epic #3264 - Part 1 of 4
Creates a new CombineV2 class to handle combining API v2 webfont links.


Tabrisrp (commented):
What about GFv2 URLs that would be using the `text` parameter?

While it might be less used, it can still occur. I'm thinking combining them would cause issues, as it would apply the parameter to all fonts in the combined URL.

Maybe if we detect the parameter, we should skip the URL?

PR #3330, opened 2020-11-19
Fixes #3323 High CPU usage when combining CSS

### LINKED ISSUE #3323: 3.7 - High CPU usage when CSS combination is used

**Describe the bug**

In `3.7` when CSS combination is used and there is a lot of CSS content, the CPU usage can substantially increase.
This is because contrary to what we did in `3.6.4`, where optimization was applied separately to every CSS file, in `3.7` it's applied to the combined content.

On a customer's site, where the combined CSS content was ~2.2MB, the following part takes 20-25s to complete on WP StageCoach, where no traffic exists:
https://github.com/wp-media/wp-rocket/blob/92b9036c6a7492d8790e8558ef60c8ce8b721470/inc/Dependencies/Minify/CSS.php#L311

This can cause timeouts and bring down a busy site.

**To Reproduce**

1. Use the latest WP Rocket.
2. Enable CSS combination.
3. Enqueue a few large CSS files
4. Measure the time it takes for this part to complete:
https://github.com/wp-media/wp-rocket/blob/92b9036c6a7492d8790e8558ef60c8ce8b721470/inc/Dependencies/Minify/CSS.php#L311

**Expected behavior**

The CPU usage should be affected as little as possible by our file optimization features.

**Additional context**

**Related ticket:** https://secure.helpscout.net/conversation/1327441678/208364?folderId=2135277
**Slack conversation:** https://wp-media.slack.com/archives/C43T1AYMQ/p1605710940034300

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
Fixes #3323 

Changes the way the content is passed to the minify library: instead of passing the whole content at once, add it in smaller data chunk, and let the library handle combining.

This should prevent the high CPU usage reported in the issue, as the library's regex will process smaller data chunks instead of one big.

### DIFF
--- a/tests/Fixtures/inc/Engine/License/API/UserClient/getUserData.php
+++ b/tests/Fixtures/inc/Engine/License/API/UserClient/getUserData.php
@@ -37,15 +37,5 @@ return [
 		],
 		'expected' => $data,
 	],
-	'testShouldReturnDataWhenSuccess'  => [
-		'config'   => [
-			'transient' => false,
-			'response'  => [
-				'code' => 200,
-				'body' => $json,
-			],
-		],
-		'expected' => $data,
-	],
 ];
 
--- a/inc/Engine/Optimization/Minify/CSS/Combine.php
+++ b/inc/Engine/Optimization/Minify/CSS/Combine.php
@@ -276,7 +276,7 @@ class Combine extends AbstractCSSOptimization implements ProcessorInterface {
 	 * @return string
 	 */
 	private function get_content( $combined_file ) {
-		$content = '';
+		$minifier = new MinifyCSS();
 
 		foreach ( $this->styles as $key => $style ) {
 			if ( 'internal' === $style['type'] ) {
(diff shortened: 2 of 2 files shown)


tests/Fixtures/inc/Engine/License/API/UserClient/getUserData.php
```diff
@@ -37,15 +37,5 @@
 		],
 		'expected' => $data,
 	],
-	'testShouldReturnDataWhenSuccess'  => [
```
crystinutzaa, line 40: May I ask you why this fixture is deleted? It does not seem related to the code changes you applied.

crystinutzaa (approved):

tests/Fixtures/inc/Engine/License/API/UserClient/getUserData.php
Tabrisrp: the test is too fragile, because the data on the API end changes to often. I want to do it another way, but for now I'm removing it to stop it from failing our test suites every 2 days.

Tabrisrp:
@wp-media/qa ready for QA

I guess the easiest way would be to test on client site reporting the issue, ir copying their css to do so

piotrbak:
@Tabrisrp I checked this on the staging site of the customer, unfortunately, I don't have access to the CPU utilization statistics.

The difference in the time spent on the `execute` function is significant though:
https://github.com/wp-media/wp-rocket/blob/92b9036c6a7492d8790e8558ef60c8ce8b721470/inc/Dependencies/Minify/CSS.php#L296

Before fix:
19~ seconds

After fix:
1.5~ seconds

Should I try to create an environment where we could measure the CPU usage or that's enough and I'll just check if there's no regression in the functionality?

tests/Fixtures/inc/Engine/License/API/UserClient/getUserData.php
crystinutzaa: Ok. Good to know. On my last PR there are a lot of failures due to "fragile" test data related to API too. 
Like expiration data on licence. Probably this one is  related too.

Tabrisrp:
As discussed during the daily standup, we can consider the time difference enough to validate the change.

You can go ahead to check for regression 👍🏼

piotrbak (approved):
Worked as expected:
[testrail-report-74.pdf](https://github.com/wp-media/wp-rocket/files/5575469/testrail-report-74.pdf)

PR #3331, opened 2020-11-19
Closes #1906 - RocketCDN iframe - resend token & cname from unsubscri…

### LINKED ISSUE #1906: Lazy loading adds data-lazy-data-lazy-srcset attribute
**Describe the bug**

When lazy loading is enabled, something like `<picture></picture>` gets converted to `<picture></picture>` and this causes lazy loading to not work correctly.

**To Reproduce**

1. Have a picture element with img containing srcset on a cached page
2. View the page source and you'll see the img contains the data-lazy-data-lazy-srcset attribute

**Expected behavior**

Attribute should be *data-lazy-srcet* instead of *data-lazy-data-lazy-srcset*

**Additional context**

It looks like vendor/wp-media/rocket-lazyload-common/src/Image.php executes both  `lazyloadPictures()` and `lazyloadResponsiveAttributes()` in this case, causing *data-lazy* to be added twice.

Perhaps the `str_replace` in `lazyloadResponsiveAttributes()` could simply expect a space char or quote in front of the srcset and sizes attributes to prevent this issue?

Instead of:

`$html = str_replace('srcset=', 'data-lazy-srcset=', $html);` and `$html = str_replace('sizes=', 'data-lazy-sizes=', $html);`

Maybe it could be changed to this?

`$html = preg_replace('/([\s"\'])srcset=/i', '\1data-lazy-srcset=', $html);` and `$html = preg_replace('/([\s"\'])sizes=/i', '\1data-lazy-sizes=', $html);`

crystinutzaa:
Closes https://github.com/wp-media/wp-rocket.me/issues/1906

### Scope a solution ✅ 

**### 1. Modify rocketcdn.js and create a new function `validateTokenAndCNAME()`.**

In `window.onmessage` this function will be called. 

The function will receive 2 params `rocketcdn_validate_token` and `rocketcdn_validate_cname` and will make an AJAX call with the action `validate_rocketcdn_token_cname`.

**### 2.  \inc\Engine\CDN\RocketCDN\DataManagerSubscriber.php**

Create a new subscribed event: 'wp_ajax_validate_rocketcdn_token_cname'     => 'validate_token_cname',

In this function it will validate if the token and cname match.
In case the token and cname both are missing will update both values.
In case token is missing it will update token.
In case token is present will do nothing (ignoring what is in CNAME param).

### Estimate the effort ✅ 
Effort `[M]`


Tabrisrp (approved):

Tabrisrp:
@wp-media/qa ready for QA

piotrbak:
@Tabrisrp Going to test this on Monday, I'm preparing test scenarios now, don't want to miss anything and it's Friday afternoon

piotrbak:
@Tabrisrp It seems that everything went well on the smashingcode in terms of the scenarios. I do have some questions about the UX though.

1. When the site is already on the StackPath and the customer tries to buy our CDN, he's getting into the popup loop until the site is removed either from StackPath or RocketCDN:
https://jmp.sh/Jvc9zKf
Shouldn't we give the possibility to close this popup and make him not appear for some time at least?

2. In the process of recreation of the token/CNAME we're logging inside the iframe and later we have that kind of choice:
image
Wouldn't it be better to add `Refresh token` or something similar somewhere during this process?

In terms of the feature itself, it works as expected:
[testrail-report-75.pdf](https://github.com/wp-media/wp-rocket/files/5585293/testrail-report-75.pdf)

Tabrisrp:
@piotrbak Can you create a separate issue for your notes? That way it can be discussed with the product team.

PR #3347, opened 2020-11-23
Closes #3345 Remove defer JS safe mode: front (2/2)

### LINKED ISSUE #3345: Remove “Safe Mode for jQuery” option - Frontend (2/2)
Sub-task for epic #3303

### **Frontend modifications**

**1. Move `/inc/front/deferred-js.php` into the new structure `/inc/Engine/Optimization/DeferJs`** 
IT'S ALREADY MOVED 😃 
https://github.com/wp-media/wp-rocket/blob/3b8b2d7debc83641045f9af82baa08b14e41daed/inc/front/deferred-js.php#L14
Create the files and functionality for:
 - Create the file:  `/inc/Engine/Optimization/DeferJs/Subscriber.php`
 - Modify the file and share the Defer JS Subscriber : `/inc/Engine/Optimization/ServiceProvider.php`
 - Modify the functionality inside `rocket_defer_js()` function and split it in 2. Handle first the scripts with src and the second part will handle the inline scripts.
- External scripts will maintain the current functionality.
- Inline scripts will need much more functionality:
    a. Check if there is jQuery code inside ( `$(` , `$.(` , `jQuery.`, `jQuery` - and other variations 😄 ) - if it does not contain will be skipped.
    b. Check if inline JS contains the pattern “DOMContentLoaded” or “document.write”, these scripts will be skipped.
    c. Check if jQuery is excluded from JavaScript defer, the inline scripts are not delayed.  (https://github.com/wp-media/wp-rocket/issues/3233) - if it contains it will be skipped
     d. If all condition pass, then the current code needs to be wrapped inside: 
```
  window.addEventListener('DOMContentLoaded', function() {
        // Add inline JS 
  });
```

- Deprecate the function: `rocket_defer_js()` 
- Write tests (unit & integration )

- Make sure that the new code pass the acceptance criteria:
```
Acceptance Criteria
The UI for “Safe Mode for jQuery” is removed after the update. 

jQuery is added to the combined JS file. 

Inline JS is delayed and there are no console errors. 

If inline JS contains keywords to be delayed using our “Delay JavaScript execution” feature, these scripts are to be excluded from the delay. 

If inline JS contains the pattern “DOMContentLoaded” or “document.write”, these scripts will be excluded from the process.

If jQuery is excluded from JavaScript defer, the inline scripts are not delayed. 
```

remyperona:
Closes #3345 

### Code changes
Add a new method `defer_inline_js()`. This method:
- Checks if jquery is excluded from defer. If yes, bailout
- Finds all inline scripts on the page
- For each inline script:
    - Check if empty. If yes, skip and continue
    - Check if it contains `DomContentLoaded` or `document.write`.  If yes, skip and continue
    - Check if it contains any jQuery pattern. If not, skip and continue
- Wrap inline jQuery code in `window.addEventListener('DOMContentLoaded', function() {});` and replace it in the HTML

Also add unit/integration tests for the new method.


tests/Unit/inc/Engine/Optimization/DeferJS/DeferJS/deferInlineJs.php
```diff
@@ -0,0 +1,47 @@
...
+ * @group  DeferJS
+ */
+class Test_DeferInlineJs extends TestCase {
+	/**
+	 * @dataProvider configTestData
+	 */
+	public function testShouldReturnExpected( $config, $html, $expected ) {
+		$this->donotrocketoptimize = $config['donotrocketoptimize'];
+
+        $options  = Mockery::mock( Options_Data::class );
```
crystinutzaa, line 24: There are some minor alignment issues in here

crystinutzaa (approved):

inc/Engine/Optimization/DeferJS/DeferJS.php
```diff
@@ -75,6 +75,83 @@ public function defer_js( string $html ) : string {
...
+		$jquery_patterns = apply_filters( 'rocket_defer_jquery_patterns', 'jQuery|\$\.\(|\$\(' );
+
+		/**
+		 * Filters the patterns used to find inline JS that should not be deferred
+		 *
+		 * @since 3.8
+		 *
+		 * @param string $inline_exclusions A RegEx pattern to find inline JS that should not be deferred.
+		 */
+		$inline_exclusions = apply_filters( 'rocket_defer_inline_exclusions', 'DOMContentLoaded|document\.write' );
```
engahmeds3ed, line 121: I don't know if this is valid case or not but the following two cases will fail:-
```
	jQuery('document').ready(function(){
		//document.write('Hi');
	});
</script>
	document.writename = function(){
	    alert('asa');
	};
	jQuery('document').ready(function(){
	    document.writename();
	});
</script>
```
Both of these cases will not be enclosed inside `DOMContentLoaded` event and I think they should.

inc/Engine/Optimization/DeferJS/DeferJS.php
```diff
@@ -75,6 +75,83 @@ public function defer_js( string $html ) : string {
...
+			if ( preg_match( "/({$inline_exclusions})/msi", $inline_js['content'] ) ) {
+				continue;
+			}
+
+			if ( ! preg_match( "/({$jquery_patterns})/msi", $inline_js['content'] ) ) {
+				continue;
+			}
+
+			$tag  = str_replace( $inline_js['content'], $this->inline_js_wrapper( $inline_js['content'] ), $inline_js[0] );
+			$html = str_replace( $inline_js[0], $tag, $html );
```
engahmeds3ed, line 137: I was talking with @hellofromtonya last night about this:-
which approach would be better (performance wise)

Approach 1
```
LOOP START
str_replace(string1, string2, subject);
LOOP END
```
Approach 2
```
LOOP START
array[string1] = string2;
LOOP END
str_replace(array_keys(array), array, subject);
```

I think Approach 2 would be faster, but I added this as a hint if you have input ( really it's for me to know for my knowledge ).

inc/Engine/Optimization/DeferJS/DeferJS.php
```diff
@@ -75,6 +75,83 @@ public function defer_js( string $html ) : string {
...
+	/**
+	 * Adds the JS wrapper for inline jQuery code
+	 *
+	 * @since 3.8
+	 *
+	 * @param string $content Inline JS content.
+	 * @return string
+	 */
+	private function inline_js_wrapper( string $content ) : string {
+		return "window.addEventListener('DOMContentLoaded', function() {" . $content . '});';
```
engahmeds3ed, line 152: image

do u think that we need here to support old versions of IE? like the following schema:-
```
if (window.addEventListener) {                    // For all major browsers, except IE 8 and earlier
  window.addEventListener(...........
} else if (window.attachEvent) {                  // For IE 8 and earlier versions
  window.attachEvent(................
}
```

engahmeds3ed (changes requested):

inc/Engine/Optimization/DeferJS/DeferJS.php
Tabrisrp: No, at best we would for IE 11, but not under.

inc/Engine/Optimization/DeferJS/DeferJS.php
Tabrisrp: Do you see a way of making the cases pass while complying with the technical specs?

engahmeds3ed: for the commented code the solution will be expensive but for `writename` I'd adjust the regex to use write and writeln only.
after thinking about that, if it made a problem support can guide client how to use the filter.

inc/Engine/Optimization/DeferJS/DeferJS.php
engahmeds3ed: Ok Great

engahmeds3ed (approved):

inc/Engine/Optimization/DeferJS/DeferJS.php
hellofromtonya: Our conversation was hypothetical. Now that I see the implementation in its context, this line of code bothers me

```
$html = str_replace( $inline_js[0], $tag, $html );
```

Why? It means the entire HTML string is walked to search and then replace matches on each iteration of the `foreach`. 

Let me think about this and do some benchmarking to determine which is more performant.

hellofromtonya: Benchmarked 2 approaches

- Approach 1: code above [link](https://3v4l.org/jXYuV)
- Approach 2: do the HTML search once [link](https://3v4l.org/aVprY)
```php
$replace = [];
$search  = [];
        
foreach ( $matches as $inline_js ) {
	if ( empty( $inline_js['content'] ) ) {
		continue;
	}

	if ( preg_match( "/({$inline_exclusions})/msi", $inline_js['content'] ) ) {
		continue;
	}

	if ( ! preg_match( "/({$jquery_patterns})/msi", $inline_js['content'] ) ) {
		continue;
	}

	$search[]   = $inline_js[0];
	$replace[]  = str_replace( $inline_js['content'], $this->inline_js_wrapper( $inline_js['content'] ), $inline_js[0] );
		}
		
	$html = str_replace( $search, $replace, $html );
````

Notes:

- Used `microtime` as a cheap benchmark
- 10,000 cycles for an average time for each PHP version
- [Here's a link to the spreadsheet](https://docs.google.com/spreadsheets/d/1Njo2A_OS4-ekAUP9pFlLf0m4IqOBd-jeRCdFKq3i-1k/edit?usp=sharing)

Results:

|                                        | Approach 1 | Approach 2 |
|----------------------- | ----------- | ------------ |
| Average microseconds |           8.62  |             9.69  |
|Total number of cycles  |     620,000 |       640,000 |
| code                               | [link](https://3v4l.org/jXYuV) | [link](https://3v4l.org/aVprY) |


Conclusion:
Go with the code in this PR. Why?

1. Teeny tiny bit faster 
2. Uses less memory
3. Less code
4. Easier to understand
5. Easier to maintain

hellofromtonya (approved):

PR #3354, opened 2020-11-24
Fix respect site's trailing slash settings

### LINKED ISSUE #1392: Respect site's trailing slash settings
Some customers believe having a trailing slash, or not, is important. We know that it's not a real SEO issue since Google uses the canonical and does not index content with a slash in addition to the version without a slash. 

However, would it be possible to detect a site's permalink settings and add the necessary htaccess rules automatically, to prevent complaints?

It appears to be an approach used by Comet Cache: 
https://github.com/websharks/comet-cache/issues/554#issuecomment-229243833

Caveat: multisite is a bit different and I think requires a different set of rules, so we would have to test carefully for that.

hs tag: `trailing slash`

engahmeds3ed:
fixes #1392 

### Acceptance criteria
- On Apache, the pages with/without trailing slash are correctly redirected to the corresponding permalink structure
- The `htaccess` file is updated when the permalink structure is changed from the WP permalink settings
- On NGINX, the redirection happens only when the page is not yet cached


remyperona (approved):

inc/functions/htaccess.php
```diff
@@ -694,3 +698,37 @@ function rocket_check_htaccess_rules() {
...
+ * Return rules for trailing slash redirection based on permalink schema.
+ *
+ * @return string htaccess Rules.
+ */
+function rocket_htaccess_trailingslash() {
+	global $wp_rewrite;
+
+	$rules  = '<IfModule mod_rewrite.c>' . PHP_EOL;
+	$rules .= 'RewriteEngine On' . PHP_EOL;
+	$rules .= 'RewriteCond %{REQUEST_URI} !wp-admin' . PHP_EOL;
```
Tabrisrp, line 712:
```suggestion
	$rules .= 'RewriteCond %{REQUEST_URI} !(?:^|/)wp\-admin(?:/|$)' . PHP_EOL;
```

remyperona (changes requested):
~~In the `AdminSubscriber`, you can add a new method hooked on `wp_rocket_upgrade`, to update the `htaccess` file when updating to version 3.8~~

remyperona (approved):

engahmeds3ed:
@wp-media/qa You need to check this change carefully please.

I'm sure you know what to do exactly but I'm highlighting some scenarios that need to be check here:-

1. Site is using apache with htaccess and no pages on cache and update permalink structure to use trailing slash and try open any inner page of website with no trailing slash you should be redirected to trailing slash url and try again with the same steps with setting permalink structure to no trailing slash.

2. Try again with cached pages the same at point 1

3. Try with nginx ( it should redirect not cached pages only as we can't control nginx rules )

4. Try opening admin page and make sure that there is no redirection there

5. If you can install any plugin that changes admin login url to be sure that admin page doesn't go into redirection loop.

6. When upgrading, `.htaccess` should be flushed.

7. Vasilis saw some issues before when he enabled `Separate cache files for mobile devices` option so please make sure of that.

piotrbak:
@Tabrisrp @engahmeds3ed I can see a couple of issues here and will move forward with the further investigation with external plugins tomorrow. Here's what I've found so far:

1. Multisite on subdirectory:
- When permalinks settings are saved on site A, it changes the trailing settings on the site B
- Sites are using `domain.com/site1/wp-admin/` paths. Since `/wp-admin/` is not at the [beginning](https://github.com/wp-media/wp-rocket/blob/7bbc33a45f12503e105c084e779ad4448ab2c39b/inc/functions/htaccess.php#L712), we're rewriting admin URLs to the trailing slashes (also the resources):
image
2. We're checking if the Force Trailing Slash plugin [exists](https://github.com/wp-media/wp-rocket/blob/7bbc33a45f12503e105c084e779ad4448ab2c39b/inc/functions/htaccess.php#L142), but we're not doing this for the one that [removes the trailing slash](https://github.com/wp-media/wp-rocket-helpers/blob/78142b653c120020103c85d382dc2a346dbcba17/htaccess/wp-rocket-htaccess-remove-trailing-slash/wp-rocket-htaccess-remove-trailing-slash.php#L27). It can end up with a redirect loop.
3. The redirect loop can be a potential issue for people that are using the cache, their WordPress is set to use `/` but they have a custom .htaccess that points to the non `/`, and vice versa. But seems to be an edge case.
4. I still need to check this with some plugins that are:
-Hiding `wp-admin`
-Changing the images URLs, I think ShortPixel was doing it
- etc.

This seems to be a challenging update.

Tabrisrp:
I agree that it's quite risky to go ahead with it like this.

@arunbasillal Maybe this can be delayed and refined more? As far as I know, this was not originally planned for this major version.

It could be good for us to take a bit more time on this one, see if we need to exclude multisite, do more compatibility testing, etc. Especially since we're in the last week of the 3.8 sprint.

The change itself has zero impact on performance, it's more of a QoL change for some customers, and possibly reduce tickets. But right now, there is a risk to have more tickets than solving them.

arunbasillal:
@Tabrisrp I agree with you. If you think this needs more time, definitely yes. (cc @webtrainingwheels)
But can we plan this for another upcoming sprint so that we do not loose the momentum?

Tabrisrp:
@arunbasillal Yes, we continue working on it, but it's no longer tied on the 3.8 release.

Your output could be useful for the following too: should this be applied automatically in the `.htaccess`, or should we put it behind a filter to enable it?

My current thinking is that since:
- it doesn't affect performance of the website
- it solves something that is not a real issue as long as you have canonical link on the page
- it comes with risks of conflicts or errors

It might be better to make it opt-in. That would avoid unnecessary tickets, while still allowing the customers who want to preserve this behaviour to do so.

maxplateau:
Hello, I think not respecting trailing slash is a huge issue and i have a solution :
You just need to add a test at the end of the function can_init_process() in class-tests.php file.

// Don't process if not trailing slash
if( substr($this->get_request_uri_base(),-1) != "/" ) {
	return false;
}

My code just check if the uri is ending with a slash and if it's not then the cache is not processed.

Please add this code in your next update. Thanks !

PR #3365, opened 2020-11-25
Add Missing Width And Height Attributes On Images

### LINKED ISSUE #3301: Add Missing Width And Height Attributes On Images

With this enhancement, WP Rocket will add any missing “width” and “height” values to images used on a website. 

## Why are we doing this? 

When a browser is parsing a page it loads the HTML first and waits for the images to be downloaded. If width and height values are available, browsers can use that information to reserve the space needed for the images on the page. 

<p></p>

When space for an image is not reserved, the browser will move content around to accommodate the image once the image is downloaded. This layout shift leads to a poor reading experience for the user. Having this feature will help improve the user experience for our users. 

<p></p>

In the new version of PageSpeed, this layout shift is counted towards the [<u>Cumulative Layout Shift</u>](https://web.dev/cls/) metric. This feature will also help with improving the CLS score. 

## Benchmark Tests

Test page: [<u>http://mega.wp-rocket.me/cls-test</u>](http://mega.wp-rocket.me/cls-test)

<p></p>

Test method: Twelve images were added first without the “width” and “height” attribute. CLS was recorded using PageSpeed thrice. Then the “width” and the “height” attribute were added and rested with PageSpeed. Three tests were recorded in each case and the average was calculated. 

<p></p>

**Results**

- Without “width” and “height” attributes. Average CLS on desktop = 1.875

- With “width” and “height” attributes. Average CLS on desktop = 1.015

Before ^

<p></p>

After ^

We will add a UI for this feature. UI Location: Media tab. After “LazyLoad”. The option will be disabled by default for existing users and new users alike. 

<p></p>

Heading: **Image dimensions**

<p></p>

Description: **Add missing width and height attributes to images. Helps prevent layout shifts and improve the reading experience for your visitors. More info**

<p></p>

Option name: **Add missing image dimensions**

<p></p>

<p></p>

**Recommendations for a future iteration**

- We can enable this option by default for new users if we notice that the option is generally safe to use. 

- Option can be added to the WP Rocket metabox on every post if users need it to disable on a per post basis. 

## Technical Specs

<p></p>

This feature is already part of WP Rocket since version 1.1 as a “secret option”. Details can be found here: [<u>https://github.com/wp-media/wp-rocket/blob/master/inc/front/images.php</u>](https://github.com/wp-media/wp-rocket/blob/master/inc/front/images.php)

<p></p>

This was changed in WP Rocket 2.2.2 with the following note in the changelog: 

<p></p>

*Fix: Width and Height attributes will no longer be added automatically to prevent conflicts. Simply add this filter: add_filter( 'rocket_specify_image_dimensions', '__return_true' );.*

## Risks

Since there are countless themes and plugins that insert images, we are most likely going to be incompatible with some of them and cause layout issues. Since it’s a visual change, this would be easy to identify. 

<p></p>

This is the main reason why we have the UI for the option so that it can be easily disabled.

## Acceptance Criteria

- UI is as per specification

- When the option is enabled, any images with missing “width” and “height” attributes have the appropriate width and height added. 

- The filter 'rocket_specify_image_dimensions' still works and enables the feature even if the UI is disabled. (To use the filter add add_filter( 'rocket_specify_image_dimensions', '__return_true' ); to functions.php)

<p></p>

- A new documentation is to be created to explain the feature for our users.

engahmeds3ed:
Fixes #3301 epic

I will add here some logs to make it easy for support team to know what happened to the images

### Acceptance Criteria

- UI is as per specification
- When the option is enabled, any images with missing “width” and “height” attributes have the appropriate width and height added. 
- The filter 'rocket_specify_image_dimensions' still works and enables the feature even if the UI is disabled. (To use the filter add add_filter( 'rocket_specify_image_dimensions', '__return_true' ); to functions.php)


remyperona (approved):

Tabrisrp:
@wp-media/qa ready for QA

piotrbak:
@Tabrisrp @engahmeds3ed It looks like the filter is not working as expected, it might be because we're returning false [here](https://github.com/wp-media/wp-rocket/blob/4e53e2df43cb98db98341b7adaa96f104b83a3b3/inc/Engine/Media/ImagesDimensions/ImagesDimensions.php#L285).

There are two scenarios to consider when we're breaking the layout. In both of them, the `height` of the `img` is not set in CSS:
1. `max-width` or `width` is set for the image. Then the width will be set by the CSS and we'll force **real** height of the image. The images won't scale nicely. I believe that the same will happen when the width will not be specified.

To reproduce, check this [URL ](https://rocketlabsqa.dev/images-width-height/) in the mobile view and look for the `blog-post-10.jpg` image.

2. When we're using the `<picture>` element tag, we'll add the dimensions only to the  element, not to the <source>. If the <source> element width/height atts are not specified, the browser will use the dimensions of the ``, which were not calculated for the <source> image.

To reproduce, check the same URL in the desktop view and look for the `geeks-221.png`

Tabrisrp:
@piotrbak On your first test example, why is the `height:auto;` value commented-out?

The `width` and `height` attribute in the HTML are indicator of ratio for the browser, but in the end the CSS properties always take priority over the attributes.

If someone uses `max-width: 100%` to make the image responsive, the `height:auto;` property is required to preserve the aspect-ratio.

On the second example, there is an issue with the dimensions calculation, the image is 667x184, but the width and height attributes added are 258 and 261.

Also, it's not possible to set `width` and `height` attributes on `sources` elements, they are not valid attributes for it. Currently for `picture` and `source` those attributes are only set on the `img` fallback. So this part of the behaviour is correct.

piotrbak:
@Tabrisrp Yes, the first one might be an edge, I mentioned it because without width and height attributes added, the ratio is preserved.

When it comes to the second, the calculation is good I think, it's related to this [](https://media.geeksforgeeks.org/wp-content/uploads/20190808102629/geeks15.png) the problem is that when there are no CSS rules, the <source> images inherit those values. So it ends up that [667x184](https://media.geeksforgeeks.org/wp-content/uploads/20190825000042/geeks-221.png) `source` image is using the [258x261](https://media.geeksforgeeks.org/wp-content/uploads/20190808102629/geeks15.png) values.

Tabrisrp:
Second point is interesting, it seems that the dimensions set on the fallback img element is forcing the aspect-ratio for all images contained in the picture element, even if each one have a different ratio originally. I found this [stackoverflow question](https://stackoverflow.com/questions/64898864/how-set-width-and-height-attributes-for-img-element-within-picture-element-t) which seems to be related, but without an actual solution.

I'm wondering if we should skip img element within a picture one to prevent the issue?

piotrbak:
@Tabrisrp I think that would be the safest way. We could have a filter to bypass this exclusion and add dimensions to the `<picture>` elements too. Then, for interested customers, there would be a possibility to take advantage of this setting for `picture` elements if they know what they're doing.

arunbasillal:
> I'm wondering if we should skip img element within a picture one to prevent the issue?

@Tabrisrp I agree with this too. 

> We could have a filter to bypass this exclusion and add dimensions to the <picture> elements too. Then, for interested customers, there would be a possibility to take advantage of this setting for picture elements if they know what they're doing.

@piotrbak Do we need a filter for this? Customers who know enough to do all this can fix the issue directly by specifying dimensions. True, if it's part of a theme or plugin, it will be harder and won't be able to do it. But it's still a tiny factor. Having a filter to include and another filter to exclude ([if we end up doing it that is](https://github.com/wp-media/wp-rocket/issues/3301#issuecomment-729146661)) can be confusing I think.

Tabrisrp:
I think having the filter is a good idea, because it's likely we will get the question/need at some point, especially with plugins offering WebP optimization, where the picture element is used in place of the image element.

engahmeds3ed:
@Tabrisrp Now I added a new filter called `rocket_specify_dimension_images_inside_pictures` and by default will be true so it will skip all img tags inside picture tags from being processed to get dimensions.
Ready for review.

remyperona (approved):

Tabrisrp:
@wp-media/qa ready for another round of tests

piotrbak:
@Tabrisrp @engahmeds3ed I'm still having problems with filters:

1. I thought the filter `rocket_specify_image_dimensions` will overwrite the value in the UI. Is that the approach we're trying to achieve?

If so, here's what's happening:
- Filter is set to `true`, UI is set to `false`, the option is enabled 👍 
- Filter is set to `false`, UI is set to `true`, the option is enabled 👎 ?
Or if any of the above is set to true, then we'll just enable it?

2. When it comes to the `rocket_specify_dimension_skip_pictures`, it's not skipping the `<pictures>` at all in my case. Check the markup here, with no `add_filter` present:
https://rocketlabsqa.dev/images-width-height/

piotrbak:
@engahmeds3ed It worked as expected on the rocketlabs. Can you push that change?

piotrbak (approved):
Worked as expected:
[testrail-report-82.pdf](https://github.com/wp-media/wp-rocket/files/5624174/testrail-report-82.pdf)

Issue #3233, opened 2020-10-22
Excluding JS files from Defer JS


GeekPress:
## What are we proposing to do? 

WP Rocket will include a UI to exclude JavaScript files from being deferred. 

## Why are we doing this? 

Excluding JavaScript files from defer currently needs a helper plugin and is not the most user-friendly experience. This update will make it easier for all of our users to configure WP Rocket faster. 

<p></p>

This is a feature that has been requested by our users many times as well. [<u>#2290</u>](https://github.com/wp-media/wp-rocket/issues/2290)

## User expectations and experience. How will we do this feature?

This feature will be implemented after we remove the “Safe Mode for jQuery” option. The UI proposal is as follows:

![](https://lh3.googleusercontent.com/twAbxUk3WyqU_rguZLb43eWghbBCU5fkX6jIZE0nXwfW74VijOk29oHJumL_ItSmTkwOsfraVMe1RK7RlrliGh4Wfw1ZanrUyn5iA7DuzTyhb7LMNXRk51ta5V8PJdD9eztE6DUQ)

<p></p>

Text area title: **Excluded JavaScript Files**

Text area description: **Specify URLs of JavaScript files to be excluded from defer (one per line).**

Placeholder: **/wp-content/themes/some-theme/(.*).js**

## Technical Specs

- Current filter rocket_exclude_defer_js continues to work. This will make sure our current helper plugin works and existing customers won’t be impacted. 

- Exclusions can be made with keywords from the URL, just the domain part, wildcards, or the entire URL. Examples of how it is right now are mentioned in [<u>our doc</u>](https://docs.wp-rocket.me/article/976-exclude-files-from-defer-js). 

- Files excluded from defer can also be excluded from JavaScript combination if JS combination is enabled. They can still be minified. 

## Risks

Since this is only a UI enhancement, we shouldn’t see risks of any kind. The existing helper plugin and the filter will continue to work as before. 

<p></p>

If anything, this should make things easier for users and support to resolve issues related to defer JS.

## Acceptance Criteria

- The UI is as per specification.

- JavaScript files are excluded from defer based on full file name, keyword, or wildcards. All cases mentioned in the [<u>current doc</u>](https://docs.wp-rocket.me/article/976-exclude-files-from-defer-js) works.

- When external domains are mentioned, they are deferred too. 

- Files excluded from defer are not combined by JavaScript combination. However, they are still minified. 

[...]

PR #3366, opened 2020-11-25
Closes #3303 Remove defer JS safe mode option

### LINKED ISSUE #3303: Remove “Safe Mode for jQuery” option

Currently, “Load JavaScript deferred” option in the File Optimization tab of WP Rocket includes a sub-option, “Safe Mode for jQuery (recommended)”. With this update, we propose to remove this sub-option. 

## Why are we doing this? 

This update will be part of making WP Rocket simpler and easier to configure and use. 

<p></p>

As per Mixpanel data, almost all users (98%) with “Load JavaScript deferred” option have enabled the “Safe Mode for jQuery” sub-option. The UI isn’t adding value to many users and can be confusing for novice users. Removing it will make WP Rocket easier to configure. 

<p></p>

The reason why we need this option is because jQuery references in in-line JavaScript will need the jQuery file to be not deferred. With this enhancement, we will be delaying inline JS and this feature will no longer be needed and the jQuery file can be deferred. 

<p></p>

Since inline JS is delayed with this enhancement, more JavaScript execution will be delayed and this should mean that the website is interactive faster (i.e. time to interactive is lower). This will reflect in better user experience and better ranks on PageSpeed insights. 

## Benchmark Tests

Test website: [<u>http://mega.wp-rocket.me/</u>](http://mega.wp-rocket.me/)

<p></p>

Test method: Tests were done first with “Safe Mode for jQuery” enabled and then with it disabled. The website was repeated on PageSpeed thrice with each setting. Combine JS was enabled for all these tests. 

<p></p>

Average PageSpeed score. With Safe Mode for jQuery enabled = 68, disabled = 78

<p></p>

This test doesn’t account for delayed inline JS. That should only improve the scores. 

This update will be completely transparent to users. The UI will simply disappear on update and everything will be handled automatically in the background. 

During the QA process, some changes were made to how the change would affect existing users and new users. Details are available here: https://github.com/wp-media/wp-rocket/pull/3366#issuecomment-735850052

## Technical Specs

By wrapping all jQuery inline scripts into a DOMContentLoaded event, we will be able to get rid of Safe Mode for jQuery options.

Some code example here with jQuery deferred: [<u>https://playcode.io/679488/</u>](https://playcode.io/679488/)

To know which inline scripts contain a jQuery inline script, we will detect them by checking these patterns:

- jQuery

- jQuery.

- $.(

- $(  

**Notes:** 

- There are two methods used in the above example. One where the inline JS is converted to base64 code and then deferred. Second where inline JS is wrapped into “window.addEventListener('DOMContentLoaded', function()”. 

- The second approach essentially delays inline JS and can give better performance. This is the one we hope to use. 

- For comparison, “LightSpeed Cache” has the option to use the second approach. Check Page Optimization > JS Settings tab > Load inline JS

<p></p>

**Other specifications:** 

- If inline JS contains keywords to be delayed using our “Delay JavaScript execution” feature, these scripts can be excluded. 

- If inline JS contains the pattern “DOMContentLoaded” or “document.write”, these scripts will be skipped.

- If jQuery is excluded from JavaScript defer, then we can skip the entire process. This will be similar to how “Safe Mode for jQuery” worked. This can act as a backup plan if our update doesn’t suit some websites.

- [<u>With the UI that will be added for file exclusions</u>](https://github.com/wp-media/wp-rocket/issues/3233), excluding the jQuery file from JS defer will be easy for our users. If they see a “jQuery not defined” error in spite of our mechanisms, they can simply exclude the jQuery file (based on our documentation) and all can go back to normal. 

## Risks

About 60% of our users will be impacted by this update, so we need to roll this out with a major version release to take advantage of our staggered release. 

## Acceptance Criteria

- The UI for “Safe Mode for jQuery” is removed after the update. 

- jQuery is added to the combined JS file. 

- Inline JS is delayed and there are no console errors. 

- If inline JS contains keywords to be delayed using our “Delay JavaScript execution” feature, these scripts are to be excluded from the delay. 

- If inline JS contains the pattern “DOMContentLoaded” or “document.write”, these scripts will be excluded from the process.

- If jQuery is excluded from JavaScript defer, the inline scripts are not delayed. 

The following documentation needs to be updated:

<p></p>

- [<u>https://docs.wp-rocket.me/article/1265-load-javascript-deferred</u>](https://docs.wp-rocket.me/article/1265-load-javascript-deferred)

- [<u>https://docs.wp-rocket.me/article/1291-find-the-best-settings-for-your-site</u>](https://docs.wp-rocket.me/article/1291-find-the-best-settings-for-your-site)

<p></p>

remyperona:
Closes #3303 

### Acceptance Criteria

- The UI for “Safe Mode for jQuery” is removed after the update. 
- jQuery is not added to the combined JS file, but is deferred (except if excluded from defer from the UI)
- Inline JS is delayed and there are no console errors. 
- If inline JS contains keywords to be delayed using our “Delay JavaScript execution” feature, these scripts are to be excluded from the delay. 
- If inline JS contains the pattern “DOMContentLoaded” or “document.write”, these scripts will be excluded from the process.
- If jQuery is excluded from JavaScript defer, the inline scripts are not delayed.
- When updating from a previous version, jQuery is excluded in the new field if safe mode was enabled


Tabrisrp:
@wp-media/qa Ready for QA

piotrbak:
@arunbasillal I can see some potential (and confirmed on test environment) issues:

1. `Load JS Deferred` is enabled, `Combine JS` is enabled, one of the files that are excluded from `Combine JS` is dependent on the jQuery (I've checked some of the files we're excluding [automatically](https://github.com/wp-media/wp-rocket/blob/92b9036c6a7492d8790e8558ef60c8ce8b721470/inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php#L192), and already [found one](https://embed.lpcontent.net/leadboxes/current/embed.js), moreover, customers might have specific exclusions like that on their installs). The above scenario is causing jQuery dependent script to be loaded before the jQuery placed in the combined file.

2. `Load JS Deferred` is enabled, `async` attribute is being used on the jQuery dependent file below the jQuery file. `async` attribute is not respecting the order of the loaded files. It's being loaded before the jQuery library is parsed.

3. `Combine JS` and `Load JS Deferred` are enabled, the script that returns an error is placed **before** the jQuery file. We'll combine that script and also jQuery one. Because of the error, the jQuery library won't be processed in the combined file. When jQuery wasn't present in the combined file, then all jQuery dependent scripts that are outside the combined file worked just fine. After this update, all jQuery dependent scripts outside the combined file will fail with `jQuery not defined`.

piotrbak:
@Tabrisrp @arunbasillal The happy path scenarios went just correct. Except for the above, there are 2 things more so far:

1. What's the reason we're excluding `document.write`? I used the following code and it caused `Uncaught ReferenceError: $ is not defined` error because of the feature (this is some random code from the internet):
https://snippi.com/s/w95cb2s

2. When excluding jQuery from the Defer JS when the combination is disabled, we need to use the `minified` URL. It's a bit misleading, especially when hosting external jQuery locally. What do you think?

[testrail-report-79.pdf](https://github.com/wp-media/wp-rocket/files/5604672/testrail-report-79.pdf)

arunbasillal:
Point 1 and 2 are going to be of serious concerns I think since it's going to be quite common. 

- We could detect if a JS file is using jQuery either by scanning it or checking for dependencies in `wp_enqueue_script`?
- Then delay those files just as we would do for inline JS. 

If this solution is feasible, wouldn't that solve the issue? 

For Point 3, what if we always combine jQuery to the top of the combined JS file? (credit to Piotr for the idea). 

cc: @GeekPress

arunbasillal:
> What's the reason we're excluding document.write? 

Because `document.write` is always executed after dom is loaded. Which should be after our combined JS is loaded? 

> When excluding jQuery from the Defer JS when the combination is disabled, we need to use the minified URL. It's a bit misleading, especially when hosting external jQuery locally. What do you think?

This is indeed confusing. Since we advise users and our team to check the uncached page for the files and links. Can this be updated? 🤔

Tabrisrp:
This is honestly starting to get messy. Having to download the content or check dependencies would make for substantial code changes. And how would you delay the files when they are external? We can't wrap them like we do for inline JS.

Same thing for moving jQuery at the top of the combined file, messing up with the loading order is risky + requires big changes in the code.

arunbasillal:
> And how would you delay the files when they are external? We can't wrap them like we do for inline JS.

For the files, I was suggesting delay just as we do for the `Delay JS Execution` feature. But I can understand things getting messy. 

But without something like that, this feature is going to generate tickets, especially since it is going to be applied to everyone who updates to this version. 

Some broad approaches (thinking out loud). 

> 1. Load JS Deferred is enabled, Combine JS is enabled, one of the files that are excluded from Combine JS is dependent on the jQuery 

- Right now, the excluded JS file will be loaded before the combined JS file.
- What if we move the JS file to after the combined JS file and still defer it? Wouldn't that solve the issue?

> 2. Load JS Deferred is enabled, async attribute is being used 

- For this scenario, what if we delay every file that use `async`. Files with async should be fine if they are delayed, right? 
- Or switch `async` to `defer`

What do you think? cc: @Tabrisrp @GeekPress

Tabrisrp:
> What if we move the JS file to after the combined JS file and still defer it? Wouldn't that solve the issue?

There is risks with that approach. What if there is inline JS depending on this file is on the page? Now the scripts are out of order, and we're going to get errors too. Modifying scripts order is a pretty bad idea.

> For this scenario, what if we delay every file that use async. Files with async should be fine if they are delayed, right?

Async and delay are not the same thing. the async attribute tells the browser to load & execute the file asynchronously, but it doesn't wait for any browser event. The delay does.

> Or switch async to defer

Same thing, they're not the same. `defer` waits for the complete parsing of the document, `async` doesn't.

I don't think you can not expect a ticket increase when you straight-up remove an option that is there specifically to avoid this kind of situation as much as possible, and is used by 98% of our users.

The whole feature, from the feature doc, is also based on the idea that only "inline JS" was impacted by the safe mode, which is not the case, as found by the QA process.

We're also encountering conflicts between the different conditions of the technical specs, like with the `document.write` case which also includes jQuery code.

Maybe the approach for this is not fleshed-out enough, and needs further considerations before continuing?

GeekPress:
One thing to consider which is important: the goal of this enhancement is to be able to defer jQuery and stop being render-blocking.

It means that jQuery can still be automatically excluded from the `Combine JS` if `Defer JS` is enabled. 

Why?
Because even if jQuery is loaded in the header with the defer attribute, it won't be considered as render-blocking anymore.

If keeping jQuery on the top will fix our issues and is safer, then let's do that.

GeekPress:
Another thing we can do if we aren't comfortable is to automatically add jQuery in the `Exclude JS files from Defer JS` for users who activated previously the Safe Mode. It will require a user action (remove the file from the exclusion) to get the benefits of the enhancement.

GeekPress:
About `document.write`, it was mainly used by the 3rd party to insert their script. It's something that Google doesn't recommend to use: https://web.dev/no-document-write/

I will be surprised to see issue related to document.write and jQuery.

arunbasillal:
> Another thing we can do if we aren't comfortable is to automatically add jQuery in the Exclude JS files from Defer JS for users who activated previously the Safe Mode. It will require a user action (remove the file from the exclusion) to get the benefits of the enhancement.

Since we have discovered flaws in our original plan, my vote is to proceed with this. This means we can have a "risk free" upgrade. We just have to document this so that our users will know (cc: @webtrainingwheels).

Can we agree on this @Tabrisrp @piotrbak?

GeekPress:
@arunbasillal What are your thoughts about that comment?
https://github.com/wp-media/wp-rocket/pull/3366#issuecomment-734994323

arunbasillal:
@GeekPress 

> One thing to consider which is important: the goal of this enhancement is to be able to defer jQuery and stop being render-blocking.
> It means that jQuery can still be automatically excluded from the Combine JS if Defer JS is enabled.
> Because even if jQuery is loaded in the header with the defer attribute, it won't be considered as render-blocking anymore.

Sorry I missed this reply. I think this solution would still affect cases where JS file with jQuery users `async`. Pro is that, the adoption for this feature will be quite high. 

Con is that we can expect tickets. These tickets can be solved with a saved reply to exclude jQuery from Defer JS though.

GeekPress:
@arunbasillal I’m not sure to understand why it should be one or another solution

What about doing both? :smile:

We know that not combining jQuery will still fix the render-blocking recommendations, and it will be safer for compatibility

The idea about auto-excluding for existing customers is to provide an « extra » compatibility for the previous users.

On the other side, we have no cons to automatically exclude jQuery from Combine JS once it will be deferred. We know it will help us to reduce conflicts too.

arunbasillal:
Had a chat with Jonathan to speed up conversation. Here is the result. Please let me know your thoughts @Tabrisrp 

For existing users with Safe Mode for jQuery enabled: 
- We exclude jQuery from combine JS
- We exclude jQuery from defer JS
- jQuery exclusion file is part of the new `Defer JS exclusion textarea` #3233. This will be the new equivalent of having `Safe Mode for jQuery` enabled. If a user wants to defer jQuery, they can simply remove this entry from the textarea. 

When new users enable `Defer JS`:
- jQuery is excluded from combine JS.
- jQuery is deferred.
- They can fix the edge cases by excluding jQuery from defer JS via the UI manually.

Tabrisrp:
Looks good to me, I'll work on adapting the code today to reflect that change.

inc/Engine/Optimization/DeferJS/AdminSubscriber.php
```diff
@@ -44,4 +45,22 @@ public static function get_subscribed_events() : array {
 	public function add_defer_js_option( array $options ) : array {
 		return $this->defer_js->add_option( $options );
 	}
+
+	/**
+	 * Adds jQuery to defer JS exclusion field if safe mode was enabled before 3.8
+	 *
+	 * @since 3.7
```
engahmeds3ed, line 52: I think this should be 3.8, right?

```suggestion
	 * @since 3.8
```

engahmeds3ed (approved):
small documentation point

Tabrisrp:
@wp-media/qa ready for another QA

piotrbak:
@Tabrisrp Just a question regarding the new updates + acceptance criteria. In acceptance criteria, there's a point `If jQuery is excluded from JavaScript defer, the inline scripts are not delayed.` and for the customers that are updating and had `Safe Mode` enabled, `We exclude jQuery from deferring JS`. 

Does it mean for all customers that are updating and had `Safe Mode` enabled we're preventing inline JS delay?

Tabrisrp:
That should be the case yes, since jQuery is excluded from defer, we don't need to defer the inline JS

piotrbak:
@Tabrisrp There's a problem with sanitization on the `Excluded JavaScript Files`, the one related to Defer JS. After the update, only `(` is left from the whole regex `(?:.*)/jquery-?[0-9\.]*(?:\.min|\.slim|\.slim\.min)?\.js`. Looks like `?` are the problem.

Other tests worked as expected

piotrbak (approved):
Worked as expected now:
[testrail-report-83.pdf](https://github.com/wp-media/wp-rocket/files/5629524/testrail-report-83.pdf)

PR #3377, opened 2020-12-01
Improve support for WP CLI Rocket extension

engahmeds3ed:
Adjust the code to be called correctly from WP CLI.
For example: WP CLI doesn't use `current_user_can` WP function to check the permissions but we need to add OR condition when using it to add a check on `WP_CLI` defined constant.


inc/Engine/Cache/WPCache.php
```diff
@@ -300,7 +300,7 @@ public function notice_wp_config_permissions() {
 	 * @return bool
 	 */
 	private function is_user_allowed() {
-		return current_user_can( 'rocket_manage_options' ) && rocket_valid_key();
+		return ( defined( 'WP_CLI' ) || current_user_can( 'rocket_manage_options' ) ) && rocket_valid_key();
```
Tabrisrp, line 303:
```suggestion
		return ( rocket_get_constant( 'WP_CLI' ) || current_user_can( 'rocket_manage_options' ) ) && rocket_valid_key();
```

remyperona (changes requested):

remyperona (approved):

PR #3385, opened 2020-12-03
Fixes #3384 Update delay JS sanitization

### LINKED ISSUE #3384: Sanitization Delay JS field problem

**Describe the bug**
In 3.7.6 we made a change in the sanitization for JS/CSS exclusion fields:
https://github.com/wp-media/wp-rocket/pull/3295

Delay JS is using the same sanitization:
https://github.com/wp-media/wp-rocket/blob/82b99b9b15e9b1144e2e24f06993d05c825fb8a4/inc/functions/formatting.php#L190

It ends up that only `.js` ending files are accepted

**To Reproduce**
1. Update to 3.7.6
2. Enable Delay JS
3. Add a new pattern without the .js ending and save settings
4. Confirm that it's not present after saving settings

**Expected behavior**
List saved correctly

**Additional context**
Change `rocket_validate_js` to the `sanitize_text_field` might be a solution.
https://github.com/wp-media/wp-rocket/blob/82b99b9b15e9b1144e2e24f06993d05c825fb8a4/inc/functions/formatting.php#L190

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
Fixes #3384 

Use `sanitize_text_field` to sanitize the delay JS input, as the textarea can contain either URLs or patterns. The patterns won't work with `rocket_validate_js` and are getting stripped in 3.7.6.


crystinutzaa (approved):

piotrbak (approved):
I've already tested this change and the sanitization is good after the patch. There's still comment from Arun here:
https://github.com/wp-media/wp-rocket/issues/3384#issuecomment-738150470

iCaspar:
@crystinutzaa @piotrbak I've added a check-and-reset-to-defaults method for updating from 3.7.6 (and the accompanying tests). Needs a code review on the new stuff, and then good to QA for that case.
Acceptance criteria: should reset delayJS setting _only_ when coming _from_ 3.7.6 or otherwise leave everything as is.

crystinutzaa (approved):

iCaspar:
@Tabrisrp Ready for review.
@wp-media/qa Change to acceptance criteria: After update from 3.7.6 the new DelayJS settings list should be the default list _with any additions the user has made appended to the end_ (see the testcase [here](https://github.com/wp-media/wp-rocket/blob/04e43de46df378112d486724da2c8f198badbf32/tests/Fixtures/inc/Engine/Optimization/DelayJS/Admin/Subscriber/optionUpdate3761.php#L20) as an example.

inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -217,7 +217,7 @@ public function option_update_3_7_6_1( $old_version ) {
 
 		$options = get_option( 'wp_rocket_settings', [] );
 
-		$options['delay_js_scripts'] = $this->defaults;
+		$options['delay_js_scripts'] = array_values( array_unique( array_merge( $this->defaults, $options['delay_js_scripts'] ) ) );
```
hellofromtonya, line 220: Do we know with certainty that `$options['delay_js_scripts']` exists and it's an array?

iCaspar: Since we're updating from a specific version where the array has been defined previously, it should be there. Of if it's not it  something else has gone (badly) wrong.

hellofromtonya (approved):

PR #3390, opened 2020-12-07
Mod Pagespeed admin notification

### LINKED ISSUE #3369: Add a notification when Mod Pagespeed is enabled and tell customers that it is likely to conflict with WP Rocket
Add a notification when Mod Pagespeed is enabled and tell customers that it is likely to conflict with WP Rocket, and that they should turn it off on their server to avoid any issues.

We can check if Mod Pagespeed is enabled, maybe using the response header they send, and show notification on WP Rocket dashboard. 

N/A

**Additional context**
Mod Pagespeed adds the following headers: 
* X-Mod-Pagespeed for Apache
* X-Page-Speed for Nginx

engahmeds3ed:
fixes #3369


engahmeds3ed:
I found that `apache_get_modules()` is very unreliable, because most hosting companies will not run Apache as a **DSO**, it is run via **CGI** in most cases. Making `apache_get_modules()` difficult to rely on.

So We used it only just in case return true if found our pagespeed module there but won't return false if not found there and then go to the next reliable check to get headers of the homepage and search on `X-Mod-Pagespeed` OR `X-Page-Speed`.

Then I will use transient to cache this condition.

inc/ThirdParty/Plugins/ModPagespeed.php
```diff
@@ -0,0 +1,106 @@
...
+	 * Check if mod_pagespeed is enabled on this server.
+	 *
+	 * @since  3.8.2
+	 *
+	 * @return bool
+	 */
+	private static function has_pagespeed(): bool {
+		global $is_apache;
+
+		if ( $is_apache && function_exists( 'apache_get_modules' ) ) {
```
iCaspar, line 58: For performance and to avoid checking on impossible headers, maybe start with:
```if ( ! $is_apache ) {
       return false;
}
```
Then you can eliminate the `$is_apache &&` in the next check as well.

iCaspar (approved):

engahmeds3ed:
@wp-media/qa ready for QA

inc/ThirdParty/Plugins/ModPagespeed.php
```diff
@@ -0,0 +1,106 @@
...
+
+		if ( empty( $headers ) ) {
+			return false;
+		}
+
+		if ( ! empty( $headers['X-Mod-Pagespeed'] ) || ! empty( $headers['X-Page-Speed'] ) ) {
+			return true;
+		}
+
+		return false;
```
Tabrisrp, line 74:
```suggestion
		return ( ! empty( $headers['X-Mod-Pagespeed'] ) || ! empty( $headers['X-Page-Speed'] );
```

inc/ThirdParty/Plugins/ModPagespeed.php
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		$notice_name = 'rocket_error_mod_pagespeed';
+
+		if ( in_array( $notice_name, (array) get_user_meta( get_current_user_id(), 'rocket_boxes', true ), true ) ) {
+			return;
+		}
+
+		// translators: %1$s is WP Rocket plugin name, %2$s is opening <a> tag, %3$s is closing </a> tag.
+		$error_message = '<p>' . sprintf( __( '<strong>%1$s</strong>: Mod PageSpeed is not compatible with this plugin and may cause unexpected results. %2$sMore Info%3$s', 'rocket' ), rocket_get_constant( 'WP_ROCKET_PLUGIN_NAME' ), '<a target="_blank" href="https://docs.wp-rocket.me/article/1376-mod-pagespeed">', '</a>' ) . '</p>';
```
Tabrisrp, line 94:
```suggestion
		$error_message = sprintf( __( '<strong>%1$s</strong>: Mod PageSpeed is not compatible with this plugin and may cause unexpected results. %2$sMore Info%3$s', 'rocket' ), rocket_get_constant( 'WP_ROCKET_PLUGIN_NAME' ), '<a target="_blank" href="https://docs.wp-rocket.me/article/1376-mod-pagespeed">', '</a>' );
```

remyperona (changes requested):

piotrbak:
@engahmeds3ed @Tabrisrp I should wait for the QA with this one until the changes are made, correct?

engahmeds3ed:
@piotrbak working on them now

remyperona (approved):

Tabrisrp:
@wp-media/qa ready for QA

piotrbak (changes requested):
@engahmeds3ed There's a problem with this PR. When a transient is not present, the site is timing out.

Details:
- Timing out function is `get_headers`
- The site is timing out, but when I refresh the page, WP Rocket is enabled and the transient is set correctly (the PageSpeed is recognized)
image (9)
image (10)

Another thing is related to this if statement:
https://github.com/wp-media/wp-rocket/blob/031db9efe46297e9aae355cfb2d49f98fae2c018/inc/ThirdParty/Plugins/ModPagespeed.php#L58

When the PHP is run as an Apache module and PageSpeed is disabled, we'll get to the `get_headers` anyway. We should bail out here I believe:
https://github.com/wp-media/wp-rocket/blob/031db9efe46297e9aae355cfb2d49f98fae2c018/inc/ThirdParty/Plugins/ModPagespeed.php#L62

engahmeds3ed:
@piotrbak 
Thanks for your feedback.
I will try to replicate the issue mentioned above but please send me privately the admin account for this testing site.

for the second thing, please check my comment here: https://github.com/wp-media/wp-rocket/pull/3390#issuecomment-750334876
the main idea is that this function is very unreliable, so what I did here is  to check if this module is found or not, so if found I will return true, else I will depend on get_headers to double check as I tried it on multiple hosts and found different results.

Tabrisrp:
A couple of things to look into:
- Replace `get_headers()` by `wp_remote_get()` + `wp_remote_retrieve_headers()`
- Perform the `has_pagespeed()` check in the method itself, instead of in the `get_subscribed_events()` method, as we don't need to do it everywhere, only on the settings page

inc/ThirdParty/Plugins/ModPagespeed.php
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	/**
+	 * Check if mod_pagespeed is enabled on this server.
+	 *
+	 * @since  3.8.2
+	 *
+	 * @return bool
+	 */
+	private static function has_pagespeed(): bool {
```
Tabrisrp, line 51: the method doesn't need to be static anymore

inc/ThirdParty/Plugins/ModPagespeed.php
```diff
@@ -0,0 +1,118 @@
...
+			home_url(),
+			[
+				'timeout'   => 0.01,
+				'sslverify' => apply_filters( 'https_local_ssl_verify', false ), // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
+			]
+		);
+
+		if ( is_wp_error( $home_request ) ) {
+			return false;
+		}
```
Tabrisrp, line 70: That's not needed, this check is part of `wp_remote_retrieve_headers()`

inc/ThirdParty/Plugins/ModPagespeed.php
```diff
@@ -0,0 +1,118 @@
...
+		if ( $is_apache && function_exists( 'apache_get_modules' ) ) {
+			if ( in_array( 'mod_pagespeed', apache_get_modules(), true ) ) {
+				return true;
+			}
+		}
+
+		$home_request = wp_remote_get(
+			home_url(),
+			[
+				'timeout'   => 0.01,
```
Tabrisrp, line 63: that's a very low timeout, what if we don't receive the response in that timing?

remyperona (changes requested):

inc/ThirdParty/Plugins/ModPagespeed.php
engahmeds3ed: Ok I will leave it as default (5 seconds)

remyperona (approved):

piotrbak:
@Tabrisrp @engahmeds3ed I have problems with my FlyWheel local environment, it's always timing out. It would be good to check it on another instance of this environment.

I tested it on 4 other environments and it worked as expected. In each case, the request took less than 1 second. What would you tell about decreasing the timeout value for this request?

Tabrisrp:
@piotrbak @engahmeds3ed We can set it to 3 seconds instead of 5.

piotrbak:
@Tabrisrp @engahmeds3ed Sounds good. Not sure what to do with this FlyWheel local environment... Everywhere else I tested it, everything went just fine:
[testrail-report-98.pdf](https://github.com/wp-media/wp-rocket/files/5772518/testrail-report-98.pdf)

engahmeds3ed:
@Tabrisrp Ok I will set it to 3 seconds with test cases in a while

@piotrbak I will try tomorrow to install it locally to check, I'm sure it's only configuration.

engahmeds3ed:
@wp-media/php @piotrbak 

Now the code is using wordpress function [apache_mod_loaded](https://developer.wordpress.org/reference/functions/apache_mod_loaded/) and adjusted the tests to pass.

crystinutzaa (approved):

engahmeds3ed:
@piotrbak ready for QA

piotrbak (approved):
Worked as expected. In the future release, we should consider adding the `X-LS-Pagespeed` header:
[testrail-report-99.pdf](https://github.com/wp-media/wp-rocket/files/5777359/testrail-report-99.pdf)

PR #3402, opened 2020-12-10
Update the pattern used for jQuery exclusion from defer

Tabrisrp:
The previous pattern could be matching unexpected files.

This new pattern has been tested against the jQuery URLs we were previously excluding, and also some of the unexpected URLs that were matching. See https://regex101.com/r/nQMi70/3

### DIFF
--- a/inc/Engine/Admin/Beacon/Beacon.php
+++ b/inc/Engine/Admin/Beacon/Beacon.php
@@ -635,7 +635,7 @@ class Beacon extends Abstract_Render implements Subscriber_Interface {
 			],
 			'exclude_lazyload'           => [
 				'en' => [
-					'id' => '5418c792e4b0e7b8127bed99',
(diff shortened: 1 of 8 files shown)


hellofromtonya:

hellofromtonya (approved):

iCaspar (approved):

piotrbak (approved):
Worked as expected:
- Before, unexpected files were excluded:
https://regex101.com/r/nQMi70/2
- After the change, all jQuery files are still excluded but without the accidental scripts:
https://regex101.com/r/nQMi70/3
- The fix was already tested on the customer's site and it helped:
https://secure.helpscout.net/conversation/1363634414/222156?folderId=4130695
- Sanitization is fine
- Update process worked just as before

hellofromtonya (approved):

PR #3425, opened 2020-12-17
Fix/set proper non grandfathered create date

### LINKED ISSUE #3388: Set a customer as non-grandfathered is date_created = 0
**Describe the bug**
For an unknown reason, some of our customers have the date_created set to 0 in Redis. Each time we noticed the issue, it was from non-grandfathered customers.

The problem is they will get the wrong renewal offer from the plugin.

**Expected behavior**
If we detect that date_created is equal to 0, we need to consider the customer as a non-grandfather customer.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

iCaspar:
Fixes #3388 

- Checks for both unset create_date and `0` create_date, returning current `time()` in either case instead of `0`.
- Updates unit test and fixtures for `Engine/Licence/API/User::get_creation_date()`

### DIFF
--- a/inc/Engine/License/API/User.php
+++ b/inc/Engine/License/API/User.php
@@ -61,10 +61,12 @@ class User {
 	 */
 	public function get_creation_date() {
 		if ( ! isset( $this->user->date_created ) ) {
-			return 0;
+			return time();
 		}
 
-		return (int) $this->user->date_created;
+		return (int) $this->user->date_created > 0
+			? (int) $this->user->date_created
+			: time();
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


remyperona (approved):

iCaspar:
@wp-media/qa This one is ready to go. Acceptance criteria are in the OP: #3388

piotrbak (approved):
Worked as expected:
[testrail-report-90.pdf](https://github.com/wp-media/wp-rocket/files/5716941/testrail-report-90.pdf)

PR #3427, opened 2020-12-18
Closes #3408 - Avada - Google Maps not being displayed (defer_js)

### LINKED ISSUE #3408: Avada - Google Maps not being displayed
- Made sure you’re on the latest version ✅ I used WP Rocket 3.8 patched with #3402 

**Describe the bug**

Google Maps aren't displayed when they are added using the Avada Builder.

The issue is resolved when excluding the following file from deferring:
```HTML
https://maps.googleapis.com/maps/api/js
```

**To Reproduce**

1. Use Avada's builder to add a Map to a page.
2. Enable JavaScript deferring.
3. The map won't be loaded.

**Expected behavior**

Google Maps should be loading. 

**Additional context**

**Related tickets:**
* https://secure.helpscout.net/conversation/1364618300/222465?folderId=2135277
* https://secure.helpscout.net/conversation/1365314430/222679?folderId=2135277

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

crystinutzaa:
**Closes https://github.com/wp-media/wp-rocket/issues/3408**

> - Refactored all code and moved into new structure.
> - Deprecated old methods
> - Write Integration tests for full code coverage ( Unit tests does not seem to be so relevant )**

### Reproduce the issue ✅ 
Reproduced on our Avada test site

### Identify the root cause ✅ 
The inline JS added by Avada when there is a Google Maps added with the Avada Builder is not compatible with our defer inline JS in 3.8. The inline JS is using jQuery, and the variables defined in it need to be in the global scope, which is no longer the case when wrapping them with our event listener.

### Scope a solution ✅ 
We need to always exclude both jQuery and Google Maps from deferring when using Avada to prevent issues. This can be done with a new function in `3rd-party/themes/avada.php`, hooked on our `rocket_exclude_defer_js` filter.

### Estimate the effort ✅
Effort `[S]`


inc/ThirdParty/Themes/Avada.php
```diff
@@ -0,0 +1,124 @@
...
+	 * Return an array of events that this subscriber wants to listen to.
+	 *
+	 * @since  3.3.1
+	 *
+	 * @return array
+	 */
+	public static function get_subscribed_events() {
+		$current_theme = wp_get_theme();
+
+		if ( 'Avada' !== $current_theme->get( 'Name' ) ) {
```
engahmeds3ed, line 28:
```suggestion
		if ( 'Avada' === $current_theme->get( 'Name' ) || 'Avada' === $current_theme->get_template() ) {
```

I think we need to do that to work with child themes

inc/ThirdParty/Themes/Avada.php
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	/**
+	 * Excludes Avada Google Maps JS files from defer JS
+	 *
+	 * @param array $exclude_defer_js Array of JS filepaths to be excluded.
+	 * @return array
+	 */
+	public function exclude_defer_js( $exclude_defer_js ) {
+		$exclude_defer_js[] = '/jquery-?[0-9.]*(.min|.slim|.slim.min)?.js';
```
engahmeds3ed, line 86: this issue is related to avada and google maps, why do we need to force exclude jquery file?

inc/ThirdParty/Themes/Avada.php
```diff
@@ -0,0 +1,124 @@
...
+
+		return $exclude_defer_js;
+	}
+
+	/**
+	 * Disable WP Rocket lazyload field if Avada lazyload is enabled
+	 *
+	 * @since 3.3.4
+	 * @param  array $disable_images_lazyload Array with plugins which disable lazyload functionality.
+	 * @return bool
```
engahmeds3ed, line 97:
```suggestion
	 * @return array
```

engahmeds3ed (changes requested):

inc/ThirdParty/Themes/Avada.php
crystinutzaa: Based on the proposed solution by Remy I followed the instructions from here:

> We need to always exclude both jQuery and Google Maps from deferring when using Avada to prevent issues. 

Also there is an internal Doc for this one including both:
https://docs.wp-rocket.me/article/1372-3-8-avada-and-google-maps

inc/ThirdParty/Themes/Avada.php
crystinutzaa: Hem. @Tabrisrp I think this is a valid point. What do you think? Should this case be handled too?

Tabrisrp: yes that's true, and to be sure you should use `strtolower` around the returned values from the methods, + use `avada` with no caps, to cover all cases

inc/ThirdParty/Themes/Avada.php
Tabrisrp: Yes both needs to be excluded, because the avada inline code for maps uses jQuery

inc/ThirdParty/Themes/Avada.php
```diff
@@ -0,0 +1,124 @@
...
+
+	/**
+	 * Excludes Avada Google Maps JS files from defer JS
+	 *
+	 * @param array $exclude_defer_js Array of JS filepaths to be excluded.
+	 * @return array
+	 */
+	public function exclude_defer_js( $exclude_defer_js ) {
+		$exclude_defer_js[] = '/jquery-?[0-9.]*(.min|.slim|.slim.min)?.js';
+		$exclude_defer_js[] = rocket_clean_exclude_file( 'maps.googleapis.com' );
```
Tabrisrp, line 87:
```suggestion
		$exclude_defer_js[] = 'maps.googleapis.com';
```

No need for the function call in that case

inc/ThirdParty/Themes/Avada.php
```diff
@@ -0,0 +1,124 @@
...
+	/**
+	 * Clears WP Rocket's cache after Avada's Fusion Patcher flushes their caches
+	 *
+	 * @since 3.3.5
+	 *
+	 * @return void
+	 */
+	public function clear_cache_fusion_patcher() {
+		rocket_clean_domain();
+	}
```
Tabrisrp, line 123: You have 2 methods doing the same thing in the class, that's not necessary, you can just re-use the first one

remyperona (changes requested):

iCaspar (approved):

iCaspar:
@wp-media/qa Ready for QA.
Acceptance criteria:
1. Google Maps work correctly with the Avada builder (see OP in #3427)
2. No regressions.

piotrbak (approved):
Worked as expected:
[testrail-report-91.pdf](https://github.com/wp-media/wp-rocket/files/5725443/testrail-report-91.pdf)

PR #3436, opened 2020-12-22
Remove last of GLOB_BRACE

### LINKED ISSUE #146: GLOB_BRACE is not available on all systems
Most notably it's not supported on Solaris and Alpine (a quite popular lightweight linux distribution used mostly for running docker containers), so it's throwing a PHP notice

Related ticket: https://secure.helpscout.net/conversation/246673063/21677/?folderId=637040
Potential fix: https://github.com/moderntribe/the-events-calendar/commit/29a24e519eac3fd672e8542a606d81774b2baf94

@GeekPress what do you think about this?

iCaspar:
Closes #146 

This PR is to remove the last 3 instances of GLOB_BRACE from `rocket_clean_home()`
Also adds unit/integration tests for the function.


inc/functions/files.php
```diff
@@ -613,10 +613,13 @@ function rocket_clean_home( $lang = '' ) {
 	do_action( 'before_rocket_clean_home', $root, $lang ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
 
 	// Delete homepage.
-	$files = glob( $root . '/{index,index-*}.{html,html_gzip}', GLOB_BRACE | GLOB_NOSORT );
```
crystinutzaa, line 616: Will this delete also other files and folders? I did not tested it, but I am concerned it might delete other files except index files.

iCaspar: The new regex should not delete any other files. @piotrbak has given it some initial tests (and found in the process this separate issue #3444) and confirmed that it should be ok, but has not had full QA yet.
Some discussion came up last week re: glob use in general. See my comment [here](https://github.com/wp-media/wp-rocket/issues/146#issuecomment-753994031) 
@crystinutzaa tagging you for Code Review.

iCaspar:
Looks like we have some integration test breakage around the PDFEmbedder ☹️ 
But I've added a test here and cleaned up the regex a bit.
Ready for review again.
We should definitely merge this and get rid of the issue; the other non-critical glob ideas are "nice to have" possibilities for the future.

remyperona (approved):

Mai-Saad (commented):
Working as expected, failing tests either need test maintenance or it's an issue on the trunk
[testrail-report-307.pdf](https://github.com/wp-media/wp-rocket/files/7557877/testrail-report-307.pdf)
Notes:
- Assuming that update theme only clears the wp-rocket cache 
- Switching theme clear wp-rocket and /min/1 cache

PR #3445, opened 2020-12-24
Force all processed CSS and Google Fonts to use font-display:swap

### LINKED ISSUE #3375: Force font-display: swap for CSS minify/combine
Currently we only add `font-display:swap` if no font-display descriptor has been specified. 

This means that when a plugin/theme has set a less optimal font-display, the customer will still see the PSI recommendation to: "Ensure text remains visible during webfont load"

Could we have a filter to _replace_ another font-display with `swap` so that the customer could still resolve this recommendation?

Examples:
- https://secure.helpscout.net/conversation/1352414191/218208?folderId=273766
- https://secure.helpscout.net/conversation/1335646807/211171/
- https://secure.helpscout.net/conversation/1350068654/217032/
- https://secure.helpscout.net/conversation/1332559179/210165/

iCaspar:
Closes #3375 

- [x] Revises `AbstractCSSOptimization::apply_font_display_swap()` to replace non-swap attributes with `swap`
- [x] Revises `AbstractGFOptimization` methods to replace display attrib with `swap`
- [x] Revises Unit and integration tests


inc/Engine/Optimization/CSSTrait.php
```diff
@@ -207,11 +207,14 @@ private function apply_font_display_swap( $css_file_content ) {
 			'/(?:@font-face)\s*{(?<value>[^}]+)}/',
 			function ( $matches ) {
 				if ( false !== strpos( $matches['value'], 'font-display' ) ) {
-					return $matches[0];
+					preg_match( '/font-display:\s*(?<attrib>\w*);?/', $matches['value'], $attribute );
```
engahmeds3ed, line 210: Why searching twice, one with strpos and another one with regex?
what about using only one check
```suggestion
					if ( preg_match( '/font-display:\s*(?<attrib>\w*);?/', $matches['value'], $attribute ) ) {
```
also please use regex `i` modifier (case insensitive).

inc/Engine/Optimization/CSSTrait.php
```diff
@@ -207,11 +207,14 @@ private function apply_font_display_swap( $css_file_content ) {
...
+					return 'swap' === $attribute['attrib']
```
engahmeds3ed, line 211: here we can do two changes:
1. Use `strtolower` function so you can check non-case-sensitively.
2. Use descriptive name instead of `attrib` :D, you can use `font_swap_value` for example.

tests/Fixtures/inc/Engine/Optimization/CSSTraitApplyFontDisplaySwap.php
```diff
@@ -46,7 +46,7 @@
 CSS
 		],
 
-		'shouldNotChangeFontDisplayAttributeWhenAlreadySetInRule' => [
+		'shouldChangeFontDisplayAttributeWhenAlreadySetInRule' => [
```
engahmeds3ed, line 49: I think we need to add more test cases on combineV2, minify CSS, combine CSS.
To handle forcing font-display to be swap.

engahmeds3ed (changes requested):

tests/Fixtures/inc/Engine/Optimization/CSSTraitApplyFontDisplaySwap.php
iCaspar: @engahmeds3ed I've added a testcase for CombineV2 below. I don't think we can add for Minify/Combine CSS, as those testcases don't evaluate the contents of the results, rather that the expected files are created and combined -- both of these use the CSSTrait to perform the swap, and that is under test here in the `CSSTraitApplyFontDisplaySwap` fixture - so the code is covered.

inc/Engine/Optimization/CSSTrait.php
iCaspar: So, I've combined the strpos and regex into a single regex eval with the `i` flag. It passes the tests, but please check the new regex with care (tricky!)

tests/Fixtures/inc/Engine/Optimization/CSSTraitApplyFontDisplaySwap.php
engahmeds3ed: Ok great.

inc/Engine/Optimization/CSSTrait.php
engahmeds3ed: regex pattern is fine

engahmeds3ed (approved):

iCaspar:
@wp-media/qa Ready for QA.
Acceptance criteria: All `@font-face` rules should now have the `font-display:swap` property added, regardless of previous `font-display` value.

piotrbak:
@iCaspar @Tabrisrp I tested the following:
- If changes are made in minified CSS files, when there is another `font-display` value, when there's no font-display or if the swap is already there
-Google fonts v1 and v2 with various values

The above worked just correctly. I'd like to ask you about the following two cases:
1. Should we handle this way of adding Google Fonts?
`
<style>
@import url('https://fonts.googleapis.com/css2?family=Open+Sans:wght@300&display=block');
</style>
`
2. The fonts added in CSS which are present in the Critical CSS will not have those attributes applied, since we're fetching the not cached version of the page when creating the Critical CSS.

iCaspar:
@piotrbak 
On the `@import`, my vote is no, we don't support that. Why? Because Google's own docs recommend using `<link>`, and using `@import` will not create any speed benefit by forcing to swap -- `@import` will block parallel downloads.

On critical CSS, I'm pretty sure that would require a change in our CriticalCSS API processing rather than in the plugin.

piotrbak (approved):
Worked as expected then:
[testrail-report-95.pdf](https://github.com/wp-media/wp-rocket/files/5757310/testrail-report-95.pdf)

My only comment is that the fonts in Critical CSS won't force the `display:swap`

PR #3457, opened 2020-12-31
Change the never cache cookie textarea description

### LINKED ISSUE #3334: Update Never Cache Cookies textarea description to explain partial matches are allowed
Our **Never Cache Cookies** exclusion box accepts a partial match,  so adding `wp_woocommerce_session_` will also exclude any cookie including that string on its name, like in the case of tokenized cookies, ie: `wp_woocommerce_session_[TOKEN]`

This is not clearly documented and it might lead users to understand the cookie match has to be literal.  

- Update the field description to let users know that a partial match is allowed: "Partial matches are allowed"
- Create a DOC about **Never Cache Cookies**, explaining its usage, documenting the possibility of using partial matches. and linked from UI like in other exclusion boxes with the "More Info" links: https://jmp.sh/ZcUiRfo

**Additional context**
On [this doc](https://docs.wp-rocket.me/article/875-using-wp-rocket-with-yith-woocommerce-wishlist) we already recommend the usage of a partial match to exclude `yith_wcwl_session_` type cookies. 

On this issue: https://github.com/wp-media/wp-rocket/issues/1358 we are requesting the usage of **wildcards**, and it could lead users to think that any cookie exclusion has to be literal.

Example ticket: 
https://secure.helpscout.net/conversation/1341907384/213435/

**Suggested update to the help text** (https://github.com/wp-media/wp-rocket/issues/3334#issuecomment-731976401): Specify full or partial IDs of cookies that, when set in the visitor's browser, should prevent a page from getting cached (one per line). More info

engahmeds3ed:
Fixes https://github.com/wp-media/wp-rocket/issues/3334

### DIFF
--- a/inc/Engine/Admin/Beacon/Beacon.php
+++ b/inc/Engine/Admin/Beacon/Beacon.php
@@ -469,9 +469,9 @@ class Beacon extends Abstract_Render implements Subscriber_Interface {
 			],
 			'exclude_cookie'             => [
 				'en' => [
-					'id' => '5fe5462df24ccf588e3fe804',
-					'url' => 'https://docs.wp-rocket.me/article/1382-never-cache-cookies'
-				]
+					'id'  => '5fe5462df24ccf588e3fe804',
+					'url' => 'https://docs.wp-rocket.me/article/1382-never-cache-cookies',
+				],
 			],
 			'always_purge'               => [
 				'en' => [
--- a/inc/Engine/Admin/Beacon/Beacon.php
+++ b/inc/Engine/Admin/Beacon/Beacon.php
@@ -470,7 +470,7 @@ class Beacon extends Abstract_Render implements Subscriber_Interface {
 			'exclude_cookie'             => [
 				'en' => [
 					'id'  => '5fe5462df24ccf588e3fe804',
-					'url' => 'https://docs.wp-rocket.me/article/1382-never-cache-cookies',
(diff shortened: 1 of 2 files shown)


iCaspar (approved):

engahmeds3ed:
@wp-media/qa ready for QA

piotrbak (approved):
Worked as expected. There's a JS error from Beacon in the console and it's not loading the Help article. That's because it's not published yet:
https://docs.wp-rocket.me/article/1382-never-cache-cookies

[testrail-report-96.pdf](https://github.com/wp-media/wp-rocket/files/5764910/testrail-report-96.pdf)

inc/Engine/Admin/Beacon/Beacon.php
```diff
@@ -467,6 +467,12 @@ public function get_suggest( $doc_id ) {
 					'url' => 'https://fr.docs.wp-rocket.me/article/196-exclure-pages-cache/?utm_source=wp_plugin&utm_medium=wp_rocket',
 				],
 			],
+			'exclude_cookie'             => [
+				'en' => [
+					'id'  => '5fe5462df24ccf588e3fe804',
+					'url' => 'https://docs.wp-rocket.me/article/1382-never-cache-cookies',
```
Tabrisrp, line 473:
```suggestion
					'url' => 'https://docs.wp-rocket.me/article/1382-never-cache-cookies/?utm_source=wp_plugin&utm_medium=wp_rocket',
```

Issue #2973, opened 2020-08-07
Minified JS files are generated and deleted right away the first time "Minifiy Javascript files" is enabled


Davidacu:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version 👍 
- Used the search feature to ensure that the bug hasn’t been reported before 👍 

**Describe the bug**
The first time "Minify Javascript files" is enabled, if you navigate to the home page, and open the page source, you'll notice the links point to the `wp-content/cache/min/` directory, however when you click on the URL it returns 404, the file does not actually exist.

**Preconditions:**
1. Start with "Minify Javascript files" disabled.
2. A site with JS files

**To Reproduce**
1. Go to WP Rocket Settings -> File Optimization
2. Enable "Minify Javascript Files" and click "Save Changes"
3. Navigate to the site home page
4. Open the page source
5. Scroll to where the script is located
6. Click on the URL pointing to the cached file.

**Expected behavior**
It should open the minified JS file.

**Actual behavior**
Page is not found.
https://screencast-o-matic.com/watch/cYjioCGIVZ

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jazir555:
Should be considered a major bug, anything related to the homepage is critical

Mai-Saad:
Scenarios to be checked once this issue is fixed:
Scenario1 => Reproduce steps:
1- nothing enabled in file optimization + CDN enabled
2- clear cache
3- enable minify js
4- visit the home page -> check min folder in cache "js files added in 1 folder"
5- change CDN and save -> here the js files in min/1 are deleted
6- hard refresh the home page
7- Check cache/min/1 -> js files not added

Scenario 2 => steps:
1- minify js enabled + CDN not enabled
2- visit the home page and check min/1 folder "js files are there"
3- enable CDN and check min/1 folder "js files removed"
4- visit the home page and check min/1 folder " js files still not added"

remyperona:
### Reproduce the issue ✅ 
Reproduced on dev install

### Identify the root cause ✅ 
The issue is coming from the code in this function: https://github.com/wp-media/wp-rocket/blob/5db4357e030936938828aa2edb890e038f486a0f/inc/admin/options.php#L19

The order of events is causing it: 
- first we preload the homepage
- Then we remove the minify JS files if the option changed
[...]

PR #3463, opened 2021-01-05
Closes #3423 Add preconnect tag for CDN URLs

### LINKED ISSUE #3423: Add preconnect tag for CDN CNAME
« Establish third-party connections early » is one of the optimizations recommended by Google to improve LCP.

_Server requests to third-party origins can also impact LCP, especially if they're needed to display critical content on the page. Use rel="preconnect" to inform the browser that your page intends to establish a connection as soon as possible._

We are now using only a DNS-Prefetch on CDN CNAME which helps to resolve DNS lookups faster. The preconnect value is doing more than DNS-Prefetch: TCP, TLS, etc...

As CDN are using everywhere on a website, there are a lot of chances to load critical content, this enhancement is welcoming.

Add `<link rel="preconnect" href="{CNAME_URL}">` in addition to DNS-Prefetch.

Preconnect isn't yet compatible with Firefox for HTTPS origin (https://caniuse.com/?search=dns-prefetch). We can still use DNS-Prefetch as a fallback

iCaspar:
Closes #3423 

- [x] Add `add_precnnect_cdn()` method to `wp_resource_hints` filter.
- [X] Add integration tests for new method.

### DIFF
--- a/inc/Engine/CDN/Subscriber.php
+++ b/inc/Engine/CDN/Subscriber.php
@@ -271,20 +271,20 @@ class Subscriber implements Subscriber_Interface {
 			return $urls;
 		}
 
-		$cdn_urls = $this->cdn->get_cdn_urls([ 'all', 'images', 'css_and_js', 'css', 'js' ]);
+		$cdn_urls = $this->cdn->get_cdn_urls( [ 'all', 'images', 'css_and_js', 'css', 'js' ] );
 
 		if ( empty( $cdn_urls ) ) {
 			return $urls;
 		}
 
 		foreach ( $cdn_urls as $cdn_url ) {
-			//Note: As of 22 Feb, 2021 we cannot add more than one instance of a domain url
-			//on the wp_resource_hint() hook -- wp_resource_hint() will
-			//only actually print the first one.
-			//Ideally, we want both because CSS resources will use the crossorigin version,
-			//But JS resources will not.
-			//Jonathan has submitted a ticket to change this behavior:
-			//@see https://core.trac.wordpress.org/ticket/52465
+			// Note: As of 22 Feb, 2021 we cannot add more than one instance of a domain url
+			// on the wp_resource_hint() hook -- wp_resource_hint() will
+			// only actually print the first one.
+			// Ideally, we want both because CSS resources will use the crossorigin version,
+			// But JS resources will not.
+			// Jonathan has submitted a ticket to change this behavior:
+			// @see https://core.trac.wordpress.org/ticket/52465
 			// Until then, we order these to prefer/print the non-crossorigin version.
 			$urls[] = [ 'href' => $cdn_url ];
 			$urls[] = [
--- a/inc/Engine/CDN/Subscriber.php
+++ b/inc/Engine/CDN/Subscriber.php
@@ -270,6 +270,7 @@ class Subscriber implements Subscriber_Interface {
 		) {
 			return $urls;
 		}
+
 		$cdn_urls = $this->get_cdn_hosts();
 
 		if ( empty( $cdn_urls ) ) {
@@ -277,14 +278,19 @@ class Subscriber implements Subscriber_Interface {
 		}
 
 		foreach ( $cdn_urls as $cdn_url ) {
-			//Todo: We cannot add more than one instance of a domain url
+			//Note: As of 22 Feb, 2021 we cannot add more than one instance of a domain url
 			//on the wp_resource_hint() hook -- wp_resource_hint() will
 			//only actually print the first one.
-			$urls[] = '//123456.rocketcdn.me';
+			//Ideally, we want both because CSS resources will use the crossorigin version,
+			//But JS resources will not.
+			//Jonathan has submitted a ticket to change this behavior:
+			//@see https://core.trac.wordpress.org/ticket/52465
+			// Until then, we order these to prefer/print the non-crossorigin version.
 			$urls[] = [
-				'href' => '//123456.rocketcdn.me',
-				'crossorigin',
+				'href'        => $cdn_url,
+				'crossorigin' => 'anonymous',
 			];
+			$urls[] = [ 'href' => $cdn_url ];
 		}
 
 		return $urls;
--- a/inc/Engine/CDN/Subscriber.php
+++ b/inc/Engine/CDN/Subscriber.php
@@ -281,6 +281,10 @@ class Subscriber implements Subscriber_Interface {
 			$url_parts = get_rocket_parse_url( $url );
 
 			if ( empty( $url_parts['scheme'] ) ) {
+				if ( preg_match( '/^(?![\/])(?=[^\.]+\/).+/i', $url ) ) {
+					continue;
+				}
+
 				$url       = '//' . $url;
 				$url_parts = get_rocket_parse_url( $url );
 			}
--- a/inc/Engine/CDN/Subscriber.php
+++ b/inc/Engine/CDN/Subscriber.php
@@ -277,7 +277,18 @@ class Subscriber implements Subscriber_Interface {
 			return $urls;
 		}
 
-		foreach ( $cdn_urls as $cdn_url ) {
+		foreach ( $cdn_urls as $url ) {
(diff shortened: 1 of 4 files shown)


iCaspar:
[WIP] @see note on #3423 -- need to re-think how to add more than 1 instance of the preconnect tag for a domain -- `wp_resource_hint()` allows only 1.

iCaspar:
@Tabrisrp @GeekPress I've gone with basically what Jonathan put [here](https://github.com/wp-media/wp-rocket/issues/3423#issuecomment-774985965) as option 2, namely the function we will run on the `wp_resource_hints` filter will return the CDN domains with and without the `crossorigin="anonymous"`, such that the non-crossorigin will be printed as things stand right now, and both will be printed if ever core adopts Jo's patch.

GeekPress:
@iCaspar Thanks for the info. For the default preconnect, as our CDN is used on images, JS and CSS, I think the one without the crossorigin is better. The preconnect with crossorigin needs to be added for fonts.

remyperona (approved):

inc/Engine/CDN/Subscriber.php
```diff
@@ -241,6 +248,54 @@ public function maybe_replace_url( $url, array $zones = [ 'all' ] ) {
...
+			// only actually print the first one.
+			// Ideally, we want both because CSS resources will use the crossorigin version,
+			// But JS resources will not.
+			// Jonathan has submitted a ticket to change this behavior:
+			// @see https://core.trac.wordpress.org/ticket/52465
+			// Until then, we order these to prefer/print the non-crossorigin version.
+			$urls[] = [ 'href' => $cdn_url ];
+			$urls[] = [
+				'href'        => $cdn_url,
+				'crossorigin' => 'anonymous',
```
Mai-Saad, line 307: @iCaspar Is there a typo here?  shall it be cross_origin?

Mai-Saad: shall the URLs in return here change based on CDN reserved for what? currently, it`s always the same independent on reserve CDN for which option i.e CSS, js, ..etc

Mai-Saad:
@iCaspar while minify CSS enabled, is this fine?
1- When CDN is domain.com i.e example.org then the href for preconnect will be over HTTP not HTTPS i.e 
`<link href='http://example.org' crossorigin='anonymous' rel='preconnect' />`  while the css file itself is HTTPS
2- When CDN is https://example.org , there is no crossorigin for preconnect  
`<link href='https://example.org' rel='preconnect' />`

inc/Engine/CDN/Subscriber.php
iCaspar: On the typo, no it is correct as it is. The intent is to print a link with `crossorigin="anonymous"`, so WordPress will be looking (eventually) for that exact key.

On the reason that as of now the `crossorigin` version of the link will not be printed is because of the way WordPress currently handles the printing of those preconnect links. We can only print 1 version. ☹️ @GeekPress has submitted a patch to WP Core to allow both, and if/when (🤞) that happens, we should start seeing both. But for now, printing just the first, without crossorigin, is the expected behavior. See [this discussion](https://github.com/wp-media/wp-rocket/issues/3423#issuecomment-755428707) for a little more background.

iCaspar:
@Tabrisrp On @Mai-Saad question 1 above, can you confirm whether that is correct, or should we always force a CDN url link to use https? As is, it uses whatever the customer has entered into the list of CDNs in the dashboard.

On 2, that is unfortunately the expected, though not ideal, behavior for now. (see above)

remyperona:
@iCaspar What do you think of this approach:
- Use the URL with the protocol if it's defined in the CDN value (`http(s)://cdn.example.org`)
- Use a protocol-relative URL if the protocol is not defined in the value (`//cdn.example.org`)

iCaspar:
@Tabrisrp Implemented your suggestion above. I think the new testcases should cover things, but maybe there are others.

remyperona (approved):

iCaspar:
@Mai-Saad @wp-media/qa - Revised now to use either the specified schema or simply `//` if no schema is provided by the user.

Mai-Saad:
@iCaspar @Tabrisrp 
1- Shall we have any kind of validation on CDN that includes at least a single dot and not empty? ( currently, if CDN is test/path then we will use //test as href)
2- Using CDN name as /test/path will result in applying this to all files i.e file href will be:  https:///test/path/wp-content/cache/busting/1/gtm-5baa953518f36f664ef5d618f4d839a2.js also no dns-prefetch nor preconnect .. I think it shouldn't be applied to files, should it?
3- Currently if the CDN name is `http://test.com` then 
dns-prefetch is `<link rel='dns-prefetch' href='//test.me' />`
while preconnect is : `<link href='http://test.me' rel='preconnect' />` and same for files... so is this fine or shall dns-prefetch be as same as preconnect here?
4- Currently, changing the CDN name will delete minified/combined files and these files won't come back till clear cache or deactivate the plugin, in this case, a new minified file shall be auto-created, right? (same behavior on the trunk, I guess we can open a separate issue for this)

iCaspar:
On 1 and 2: The preconnect is for a domain, not a file, as we are only concerned about connecting to the server, not accessing anything specific. So, the expected behavior should be to ignore any CDN entry that a is single file or directory, and it should remove any path after the domain host part of the URL.
On 3: The `dns-prefetch` links are generated from ~WordPress, not by us,~ a different method related to CDN, and will always print the `//test.me` version. I think it will be fine to use the customer's scheme, http(s), if they have specified it for any reason, as prefetch and preconnect are different things. In prefetch only the DNS is resolved, but in preconnect, an actual connection with the server is established, so having the specific schema to make connect could be helpful if the customer provides it.
On 4: Since that is already the behavior, we should handle it separately as another issue, yes. If you would go ahead and open that, would be awesome! Thanks.

remyperona:
On 4) When changing the CDN cname value, everything is deleted: cache files & minify/combine files, so there is nothing to do. The new cache & minify files will be generated on the next visit (or preload).

Mai-Saad:
@Tabrisrp on 4:
for minify css -> cached files in min are autogenerated currently without a new visit after changing CDN name. While for minify js -> cached files in min aren't autogenerated even with a hard refresh of the web page.
Scenario1 => Reproduce steps:
1- nothing enabled in file optimization + CDN enabled
2- clear cache
3- enable minify js
4- visit the home page -> check min folder in cache "js files added in 1 folder"
5- change CDN and save -> here the js files in min/1 are deleted
6- hard refresh the home page
7- Check cache/min/1 -> js files not added

Scenario 2 => steps:
1- minify js enabled + CDN not enabled
2- visit the home page and check min/1 folder "js files are there"
3- enable CDN and check min/1 folder "js files removed"
4- visit the home page and check min/1 folder " js files still not added"

remyperona:
Probably this bug: #2973

Mai-Saad:
> Probably this bug: #2973

@Tabrisrp I think it`s related but not sure if they are the same thing as even if minify js was enabled and working fine before enable/edit CDN, js files won't be regenerated till clear cache and visit the URL again

iCaspar:
In further conversation with @Mai-Saad, it turns out we were not eliminating a CDN entry like `test/tests` (though we were for `/test/tests`). @Tabrisrp, I've added a regex to skip those here. We have the same issue in the `add_dns_prefetch_cdn()` and probably `get_cdn_hosts()`. The latest change solves for the preconnect, since that's what's new here. Not catching it in prefetch and possibly other places can be a separate issue, not to hold this up getting into 3.8.6.

remyperona (approved):

Mai-Saad (commented):
CDN PR all handled now except for
1- ignore test , .com
2- add preconnect for //test.com
Note: we don't add crossorigin anymore also it was there with CDN name `example.org`

iCaspar:
@wp-media/qa Per conversation with @piotrbak and @Tabrisrp we're going to go with this as is and start a new issue to validate CDN values when the field gets saved in order to prevent invalid domains/hosts from being in the picture on this and in several other places.

PR #3483, opened 2021-01-13
Add support for WooCommerce Multilingual with multicurrency and geolocation

### LINKED ISSUE #3326: Cache is emptied when changing currency using WooCommerce Multilingual currency switcher
**Describe the bug**

**To Reproduce**
1. Install WooCommerce Multilingual currency switcher
2. Change currency when on live site
3. Navigate to cache folder from WP Rocket via FTP
4. See cache completely purged

**Expected behavior**
This was created as the expected behaviour for compatibility but needs to be investigated if feature can change from purging full cache to partial purge.

**Additional context**
Currently "clean_post_cache" function is tied to "wcml_switch_currency" hook thus it's clearing cache completely when any user changes currency at the moment they visit the site. Causing server load to be increased and cache not to be available most of the time the site is visited. This was done for compatibility with the currency switcher but needs to be revisited.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

### LINKED ISSUE #3418: WooCommerce Multilingual currency switcher compatibility
WooCommerce Multilingual currency switcher is not compatible with caching. Since the new, 4.11 version they are giving a possibility to store the currency in cookies.

We should use a `dynamic cookie`, disable `rewrite rules` and be compatible with this currency switcher. This is the code that will enable the usage of cookies:
https://snippi.com/s/108po0y

They also have a way to use geolocation, we should then make sure that `mandatory cookie` is used. I've just asked their team to provide us a way to determine whether the `geolocation` or `site's language` is used.

**Additional context**
- We're doing a similar thing for Aelia
- https://github.com/wp-media/wp-rocket/issues/3326 will be closed
- I was testing the 4.11 beta, the cookie way doesn't seem to work correctly there 👎

strategio:
https://github.com/wp-media/wp-rocket/issues/3418
https://onthegosystems.myjetbrains.com/youtrack/issue/comp-3534

This PR goes with some changes in WooCommerce Multilingual (https://git.onthegosystems.com/glue-plugins/wpml/woocommerce-multilingual/-/merge_requests/1461) but there is no hard dependency between the 2 projects and we don't have any error if the code exists in one but not in the other.

I based this code on the existing one in https://github.com/wp-media/wp-rocket/blob/master/inc/3rd-party/plugins/ecommerce/aelia-currencyswitcher.php.

### DIFF
--- a/inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php
+++ b/inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php
@@ -5,4 +5,111 @@ defined( 'ABSPATH' ) || exit;
 // Compatibility with the currency switcher in WooCommerce Multilingual plugin.
 if ( defined( 'WCML_VERSION' ) ) :
 	add_action( 'wcml_switch_currency', 'rocket_clean_domain' );
+
+	/**
+	 * Use Cookie instead of WCSession
+	 *
+	 * @return string
+	 */
+	function rocket_wcml_use_cookie_storage() {
+		return 'cookie';
+	}
+	add_filter( 'wcml_user_store_strategy', 'rocket_wcml_use_cookie_storage', 10, 2 );
+
+	/**
+	 * Add dynamic cookies for WCML.
+	 *
+	 * @param array $cookies Cookies.
+	 *
+	 * @return array
+	 */
+	function rocket_wcml_add_dynamic_cookies( $cookies ) {
+		$cookies[] = 'wcml_client_currency';
+		$cookies[] = 'wcml_client_currency_language';
+		$cookies[] = 'wcml_client_country';
+
+		return $cookies;
+	}
+	add_filter( 'rocket_cache_dynamic_cookies', 'rocket_wcml_add_dynamic_cookies' );
+
+	/**
+	 * Add mandatory cookies for WCML.
+	 *
+	 * @param array $cookies Cookies.
+	 *
+	 * @return array
+	 */
+	function rocket_wcml_add_mandatory_cookies( $cookies ) {
+		// phpcs:disable WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
+		if ( apply_filters( 'wcml_geolocation_is_used', false ) ) {
+			// phpcs:enable WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
+			$cookies[] = 'wcml_client_country';
+		}
+
+		return $cookies;
+	}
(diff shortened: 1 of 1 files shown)


strategio:
Hello @Tabrisrp, any update about this PR? Something's missing?

strategio:
Hi @Tabrisrp, @piotrbak!

My changes have been tested successfully in https://github.com/wp-media/wp-rocket/issues/3418.

Is there anything else missing to have this merged?

Thanks,
Pierre

piotrbak:
@Tabrisrp When do you think we could merge this PR to the core?

remyperona:
@piotrbak We can add it for 3.9, is it validated on your side?

piotrbak:
@Tabrisrp Yes. From my point of view, it works, but I think we should get rid of this line:
https://github.com/wp-media/wp-rocket/blob/436996870bd073db3119115c1c690c04ea492e97/inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php#L7
I think it was there to make sure that the correct currency is displayed before this compatibility.

We could also improve the compatibility by hooking some cache clearing actions (I'm thinking about adding or removing the currencies, also enabling/disabling geolocation). On the other hand, it's not necessary.

inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php
```diff
@@ -5,4 +5,111 @@
 // Compatibility with the currency switcher in WooCommerce Multilingual plugin.
 if ( defined( 'WCML_VERSION' ) ) :
 	add_action( 'wcml_switch_currency', 'rocket_clean_domain' );
```
remyperona: This action should be removed, it's no longer necessary now with the new cookie-compatibility code added

remyperona (changes requested):

inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php
strategio: Done @Tabrisrp :)

remyperona (approved):

strategio:
I rebased my branch on `develop` (tests were failing for some reason, not sure it's related...).

remyperona:
Thank you, I believe it's failing because the PR is done from a fork that doesn't contain our secrets used in the actions. So it's expected, no worries.

I'm going to merge the PR, and it will be in our next release!

strategio:
Thanks a lot @Tabrisrp :)

On our side, we have complementary changes that will release with WCML 4.12.0 (no ETA yet).

Issue #3453, opened 2020-12-29
Missing Image Dimensions are not being applied on Multisite


NataliaDrause:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version - yes
- Used the search feature to ensure that the bug hasn’t been reported before -yes

**Describe the bug**
Missing Image Dimensions are not being applied for images on Multisite.

**To Reproduce**

1. Created a post with the image markup:
```
<figure>
<img src="http://my.site.com/multisite/green/wp-content/uploads/sites/4/2020/12/photo-1.jpg" alt="my alt">
<img src="http://my.site.com/multisite/wp-content/uploads/2020/12/photo-2.jpg" alt="another alt">
</figure>
```
2. Enabled **Add Missing Image Dimensions** option. Check the site - the dimensions are not applied.
3. Added filter for including external images to functions.php:
`add_filter( 'rocket_specify_image_dimensions_for_distant', '__return_true' );`
4. Check the site - the dimensions are not applied.

**Expected behavior**
Image dimensions have to be applied.

**Additional context**
WP Rocket debug log:
_Specify Image Dimensions failed because internal image is not found._
Screenshot: https://jmp.sh/KZZb1xD

Related ticket: https://secure.helpscout.net/conversation/1375704680/225688/

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

iCaspar:
**Can Reproduce:**
On Multi-site, there are 4 possibilities to check:
Subdomain type multisite --
1. Main domain site (`mysite.com`)
2. Subdomain site (`site1.mysite.com`)
Subdirectory type multisite --
3. Main domain site (`mysite.com/blog/...`)
4. Subdomain site (`mysite.com/site1/...`)

Given WP Rocket Network Activated, and the `rocket_specifiy_image_dimensions_for_distant` filter `true` -- I found the following behaviors.

For the Subdomain types:
1. On `http://rocketmultisubdomain.local/2021/01/11/image-dimensions-test-page/` we have
```
<figure>
<img src="http://site1.rocketmultisubdomain.local/wp-content/uploads/sites/2/2021/01/hippo_birdie_card.jpg" alt="hippo birdie">
[...]

PR #3493, opened 2021-01-14
Apply missing image dimensions for images with a relative path

### LINKED ISSUE #3438: Apply missing image dimensions for images with a relative path

**Describe the bug**
Images added with a relative path don't have image dimensions applied.

**To Reproduce**
1. Activate _Add Missing Image Dimensions_ option in the Media tab
2. Add image with a markup like this: ``
3. The dimensions are not applied:
image
Full-screen image: https://jmp.sh/RNI1CNa

**Expected behavior**
Image dimensions have to be applied for images with a relative path.

**Additional context**
Debug log: https://jmp.sh/XfT839x
Specify Image Dimensions failed because the internal image is not found.

Slack discussion: https://wp-media.slack.com/archives/C43T1AYMQ/p1608732263037200

- [x] Reproduce the problem
- [x] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
Closes https://github.com/wp-media/wp-rocket/issues/3438


inc/Engine/Media/ImageDimensions/ImageDimensions.php
```diff
@@ -227,7 +226,15 @@ private function get_local_path( $url ) {
...
 
-		return str_replace( content_url(), rocket_get_constant( 'WP_CONTENT_DIR' ), $url );
+		if ( false === strpos( $url, content_url() ) ) {
+			$url = site_url( '/' ) . ltrim( $url, '/' );
+		}
+
+		return str_replace(
+			'/',
+			rocket_get_constant( 'DIRECTORY_SEPARATOR' ),
+			str_replace( site_url(), rocket_get_constant( 'ABSPATH' ), $url )
```
remyperona: I'm thinking that this might not be working on some WP setups, notably when WP is installed in a sub-directory like so:
- home URL is http://example.org
- site URL is http://example.org/wp

Can you test it?

engahmeds3ed: I adjusted the tests to accept changing `site_url` and `home_url` and tests passed.

You can check this fixture
https://github.com/wp-media/wp-rocket/blob/4035cff4b1ef03f07055022202a162f923e06ef8/tests/Fixtures/inc/Engine/Media/ImageDimensions/ImageDimensions/specifyImageDimensions.php#L213-L224

remyperona (approved):

remyperona:
@wp-media/qa ready for QA

If you can validate the change on different setups:
- home URL = site URL
- home URL = site URL on a website in a subdirectory installation
- home URL ≠ site URL, with the index.php at the root and the other inside a subdirectory (See https://wordpress.org/support/article/giving-wordpress-its-own-directory/)

piotrbak:
@engahmeds3ed Please find my comments here:
- Works on a regular WordPress site 👍 
- Works with custom `wp-content` 👍 
- Doesn't work with the subdirectory WordPress 👎 
- Doesn't work with the environment where `home_url =! site_url` 👎 

The last two seem to be related to the multisite issue:
https://github.com/wp-media/wp-rocket/issues/3453

remyperona (approved):

piotrbak:
@engahmeds3ed Two things found during the QA:
- When the site is in the subdirectory `example.com/qa/` and we're loading the images from the parent site `example.com/path/to/image.jpg`, then it's not possible to specify image dimensions even when filters are set. Possible that it's related to the https://github.com/wp-media/wp-rocket/issues/3453
- When the query string is added to the image, the process fails completely `example.com/path/to/image.jpg?hey`

engahmeds3ed:
@piotrbak adjusted the code now to solve those two issues but FYI, now we are dealing with image with query string as external image so you need to enable `allow_url_fopen` from php.ini and also use this filter to return true `rocket_specify_image_dimensions_for_distant` to make it works.

`I knot that you already know that, but for future reference.`

Ready for QA again.

piotrbak:
@engahmeds3ed @Tabrisrp Last one, probably an edge case... Relative URL + query string.

We're failing to add the missing attributes there.

engahmeds3ed:
Adjusted this edge case also, @piotrbak plz check now.

piotrbak (approved):
Working as expected:
[testrail-report-117.pdf](https://github.com/wp-media/wp-rocket/files/5914223/testrail-report-117.pdf)

PR #3517, opened 2021-01-22
Fixes #1495: Fix check that allows the cache_root to be built correctly.

### LINKED ISSUE #1495: Logic to get cache root in get_rocket_htaccess_mod_rewrite() seems incorrect.
The haystack and the needle seems to be revered here - https://github.com/wp-media/wp-rocket/blob/c92efa3a52ef285eb6230a04ac4c5e6d4e2f8e26/inc/functions/htaccess.php#L187

syntax for [strpos](http://php.net/manual/en/function.strpos.php): `strpos ( string $haystack , mixed $needle [, int $offset = 0 ] ) : int`

`WP_ROCKET_CACHE_PATH` ([when evaluated](https://github.com/wp-media/wp-rocket/blob/c92efa3a52ef285eb6230a04ac4c5e6d4e2f8e26/wp-rocket.php#L59-L62)) = `ABSPATH/wp-content/cache/wp-rocket/`

So in the current order, the `needle` is always bigger than the `haystack`, so it will ALWAYS be `false`. 

I tried it by swapping the values in our mega site. Here is the diff (original on the left, new on the right) - https://www.diffchecker.com/VhDkRQIX

Related ticket - https://secure.helpscout.net/conversation/763704797/94315/

ozmodiar:
Fix for #1495.

This issue was actually quite breaking for us. Because of the fact that the needle and the haystack are not in the correct order, the DOCUMENT_ROOT was not replaced with an empty string causing the path in the RewriteCond to be incorrectly structured. Without the change proposed we ended up with an htaccess like this (snippet):

```
RewriteCond "%{DOCUMENT_ROOT}/var/web/vd17118/releases/20210104103636/wp-content/cache/wp-rocket/%{HTTP_HOST}%{REQUEST_URI}/index%{ENV:WPR_SSL}%{ENV:WPR_WEBP}.html%{ENV:WPR_ENC}" -f
RewriteRule .* "/var/web/vd17118/releases/20210104103636/wp-content/cache/wp-rocket/%{HTTP_HOST}%{REQUEST_URI}/index%{ENV:WPR_SSL}%{ENV:WPR_WEBP}.html%{ENV:WPR_ENC}" [L]
```

This means the cache files are never found and thus never hit.

In my case, this were the values:
`ABSPATH=/var/web/vd17118/releases/20210104103636/`
`WP_ROCKET_CACHE_PATH=/var/web/vd17118/releases/20210104103636/wp-content/cache/wp-rocket/`

htaccess snippet after the fix:
```
RewriteCond "%{DOCUMENT_ROOT}/wp-content/cache/wp-rocket/%{HTTP_HOST}%{REQUEST_URI}/index%{ENV:WPR_SSL}%{ENV:WPR_WEBP}.html%{ENV:WPR_ENC}" -f
RewriteRule .* "wp-content/cache/wp-rocket/%{HTTP_HOST}%{REQUEST_URI}/index%{ENV:WPR_SSL}%{ENV:WPR_WEBP}.html%{ENV:WPR_ENC}" [L]
```

I'm not a Wordpress developer so let me know if anything else is needed on my part to get this merged.

Thanks!


remyperona (approved):

remyperona:
@wp-media/qa ready for QA

piotrbak:
@Tabrisrp Should I only test the regression on regular Apache as this is fixed for the customer? Or should I try to prepare a custom environment for that? What do you think?

remyperona:
@piotrbak You can do the regression test, which should be enough.

piotrbak:
@Tabrisrp I tested this on 3 different environments. 
1. With symlinks where I could reproduce the issue, it helped
2. On two different Apaches where it worked too

Making this change will cause a small change in the .htaccess though:
`RewriteRule .* "/wp-content/cache/wp-rocket/%{HTTP_HOST}%{REQUEST_URI}/index%{ENV:WPR_SSL}%{ENV:WPR_WEBP}.html%{ENV:WPR_ENC}" [L]`

will become:

`RewriteRule .* "wp-content/cache/wp-rocket/%{HTTP_HOST}%{REQUEST_URI}/index%{ENV:WPR_SSL}%{ENV:WPR_WEBP}.html%{ENV:WPR_ENC}" [L]`

Note the lack of the `/` in the second example. In my cases, everything worked as expected. I'm thinking if there's any potential risk because of this. What do you think?

remyperona:
I'm wondering if it could lead to issues in unusual structure where the `htaccess` and the `WP_CONTENT` directory are not in the same repository. To be safe, I'm going to suggest a change on the PR to preserve the leading slash.

inc/functions/htaccess.php
```diff
@@ -197,7 +197,7 @@ function get_rocket_htaccess_mod_rewrite() { // phpcs:ignore WordPress.NamingCon
 	$site_root = isset( $site_root ) ? trailingslashit( $site_root ) : '';
 
 	// Get cache root.
-	if ( strpos( ABSPATH, WP_ROCKET_CACHE_PATH ) === false && isset( $_SERVER['DOCUMENT_ROOT'] ) ) {
+	if ( strpos( WP_ROCKET_CACHE_PATH, ABSPATH ) === false && isset( $_SERVER['DOCUMENT_ROOT'] ) ) {
 		$cache_root = str_replace( sanitize_text_field( wp_unslash( $_SERVER['DOCUMENT_ROOT'] ) ), '', WP_ROCKET_CACHE_PATH );
```
remyperona:
```suggestion
		$cache_root = '/' . ltrim( str_replace( sanitize_text_field( wp_unslash( $_SERVER['DOCUMENT_ROOT'] ) ), '', WP_ROCKET_CACHE_PATH ), '/' );
```

remyperona (changes requested):

remyperona (changes requested):
preserve leading slash

inc/functions/htaccess.php
```diff
@@ -197,8 +197,8 @@ function get_rocket_htaccess_mod_rewrite() { // phpcs:ignore WordPress.NamingCon
 	$site_root = isset( $site_root ) ? trailingslashit( $site_root ) : '';
 
 	// Get cache root.
-	if ( strpos( ABSPATH, WP_ROCKET_CACHE_PATH ) === false && isset( $_SERVER['DOCUMENT_ROOT'] ) ) {
-		$cache_root = str_replace( sanitize_text_field( wp_unslash( $_SERVER['DOCUMENT_ROOT'] ) ), '', WP_ROCKET_CACHE_PATH );
+	if ( strpos( WP_ROCKET_CACHE_PATH, ABSPATH ) === false && isset( $_SERVER['DOCUMENT_ROOT'] ) ) {
+		$cache_root = '/' . ltrim( '/', str_replace( sanitize_text_field( wp_unslash( $_SERVER['DOCUMENT_ROOT'] ) ), '', WP_ROCKET_CACHE_PATH ) );
```
remyperona:
```suggestion
		$cache_root = '/' . ltrim( str_replace( sanitize_text_field( wp_unslash( $_SERVER['DOCUMENT_ROOT'] ) ), '', WP_ROCKET_CACHE_PATH ), '/' );
```

remyperona:
@piotrbak I updated the PR, the leading slash should be preserved now

piotrbak:
@Tabrisrp On all environments I'm using we're setting this rule in this line:
https://github.com/wp-media/wp-rocket/blob/60284bffbcd0810eba2a263376b5314c2983019e/inc/functions/htaccess.php#L203

It means that the `/` is still missing

inc/functions/htaccess.php
```diff
@@ -197,8 +197,8 @@ function get_rocket_htaccess_mod_rewrite() { // phpcs:ignore WordPress.NamingCon
 	$site_root = isset( $site_root ) ? trailingslashit( $site_root ) : '';
 
 	// Get cache root.
-	if ( strpos( ABSPATH, WP_ROCKET_CACHE_PATH ) === false && isset( $_SERVER['DOCUMENT_ROOT'] ) ) {
-		$cache_root = str_replace( sanitize_text_field( wp_unslash( $_SERVER['DOCUMENT_ROOT'] ) ), '', WP_ROCKET_CACHE_PATH );
+	if ( strpos( WP_ROCKET_CACHE_PATH, ABSPATH ) === false && isset( $_SERVER['DOCUMENT_ROOT'] ) ) {
+		$cache_root = '/' . ltrim( str_replace( sanitize_text_field( wp_unslash( $_SERVER['DOCUMENT_ROOT'] ) ), '', WP_ROCKET_CACHE_PATH ), '/' );
 	} else {
 		$cache_root = $site_root . str_replace( ABSPATH, '', WP_ROCKET_CACHE_PATH );
```
remyperona:
```suggestion
		$cache_root = '/' . ltrim( $site_root . str_replace( ABSPATH, '', WP_ROCKET_CACHE_PATH ), '/' );
```

remyperona (approved):

piotrbak (approved):
Worked as expected now:
[testrail-report-114.pdf](https://github.com/wp-media/wp-rocket/files/5876254/testrail-report-114.pdf)

remyperona:
@Ozmodiar Thank you for your contribution! The fix will be included in the next release of the plugin 👍🏼

PR #3532, opened 2021-01-28
Fix/3440 pricing client api transient

### LINKED ISSUE #3440: Guard stats API against too many failures that results in flooding server with requests
We noticed that the pricing API when returns status code other than `200`, it will try to get the pricing details with each time the administrator opens any admin page.

**Technically:**
It's bailing out here:
https://github.com/wp-media/wp-rocket/blob/55e722d07a857fea52eed9f186ed93231d23d24e/inc/Engine/License/API/PricingClient.php#L48

that is called here
https://github.com/wp-media/wp-rocket/blob/55e722d07a857fea52eed9f186ed93231d23d24e/inc/Engine/License/API/PricingClient.php#L24

so it will bailout also here:
https://github.com/wp-media/wp-rocket/blob/55e722d07a857fea52eed9f186ed93231d23d24e/inc/Engine/License/API/PricingClient.php#L27

without setting the transient `wp_rocket_pricing`.
and this mainly called here:
https://github.com/wp-media/wp-rocket/blob/5dd3fa10d54a6619aeefa1589e270e715ce6925c/inc/Engine/License/ServiceProvider.php#L47

that is called only on admin as in here:
https://github.com/wp-media/wp-rocket/blob/aef2714666ac9fc20eac8a1759feffe3a30885da/inc/Plugin.php#L163

As @hellofromtonya mentioned we need to find a way to guard hitting the API after failure and I would suggest that each time we find a failure we can increase a transient with time interval so we need to wait this time till sending the request again for example:

hit the API first time `=>` status != 200 `=>` set transient time with timenow + 5 minutes
hit the API again after this 5 minutes `=>` status != 200 `=>` increase transient value with 10 minutes
over and over, and only remove this transient when success.
This will help not to hit the API once it's live from all of our sites.

**Additional Context**
I think also user data has the same behavior that needs to be adjusted also:

https://github.com/wp-media/wp-rocket/blob/55e722d07a857fea52eed9f186ed93231d23d24e/inc/Engine/License/API/UserClient.php#L75-L77

iCaspar:
New PR from #3512, already code-review approved.
Ready for @wp-media/qa 

QA:
* See #3440 for Original issue and criteria.
* See #3512 for commit history on the issue.

### DIFF
--- a/tests/Fixtures/inc/Engine/License/API/PricingClient/getPricingData.php
+++ b/tests/Fixtures/inc/Engine/License/API/PricingClient/getPricingData.php
@@ -8,6 +8,8 @@ return [
 	'testShouldReturnDataWhenCached' => [
 		'config'   => [
 			'pricing-transient' => true,
+			'timeout-active'    => false,
+			'timeout-duration'  => false,
 			'response'          => false,
 		],
 		'expected' => [
@@ -19,6 +21,7 @@ return [
 		'config'   => [
 			'pricing-transient' => false,
 			'timeout-active'    => true,
+			'timeout-duration'  => false,
 			'response'          => false,
 		],
 		'expected' => [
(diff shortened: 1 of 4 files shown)


piotrbak (approved):
After changing the expiration of the transient everything works as expected:
[testrail-report-116.pdf](https://github.com/wp-media/wp-rocket/files/5889783/testrail-report-116.pdf)

Mai-Saad:
@iCaspar with failure whenever we reach wp_rocket_pricing_timeout = 1 day , then it will remain always 1 day, right?.. is this what we want or do we need to reset back to 300s and so on? i.e

tests/Fixtures/inc/Engine/License/API/PricingClient/getPricingData.php
```diff
@@ -5,47 +5,110 @@
...
+			'timeout-duration' => 600
+		],
+	],
+
+	'testShouldNotSetTimeoutDurationLongerThanADay' => [
+		'config'   => [
+			'pricing-transient' => false,
+			'timeout-active'    => false,
+			'timeout-duration'  => rocket_get_constant( 'DAY_IN_SECONDS' )
+								   - rocket_get_constant( 'HOuR_IN_SECONDS' ),
```
Mai-Saad: @iCaspar For consistency, shouldn't 'HOuR_IN_SECONDS' be 'HOUR_IN_SECONDS'

iCaspar: Yes, it should be all caps here. 😳 
Also, yes, it is intended that once the timeout reaches 1 day it remains at 1 day until the API call is successful and the transient is deleted.

remyperona (approved):

PR #3534, opened 2021-01-29
Fix the NextGEN gallery breaking issue

### LINKED ISSUE #3456: NextGEN Gallery UI breaks with WP Rocket

**Describe the bug**

When WP Rocket is used alongside [NextGen Gallery](https://wordpress.org/plugins/nextgen-gallery/) their UI breaks.
It should be like this:

Instead, it's like this:

The error is related to our compatibility code. Since `3.3.1` we are disabling NGG's resource manager:
https://github.com/wp-media/wp-rocket/blob/3f4de3116cbd9c97f1ac5bff0002f31587353d1d/inc/classes/subscriber/third-party/plugins/class-ngg-subscriber.php#L29

Here is the part of our changelog related to this change:
> Third Party Compatibility: Disable NGG resource manager to prevent conflict with JS minification/combine

Unhooking the `deactivate_resource_manager` method from `run_ngg_resource_manager` resolves the issue.

**To Reproduce**

1. Install [NextGen Gallery](https://wordpress.org/plugins/nextgen-gallery/)
2. Create a new post, and add a NextGen Gallery:

3. Click the **UPLOAD IMAGES** tab.
4. See the error,

**Expected behavior**

NextGen Gallery should work with WP Rocket.

**Additional context**

We should check whether the specific compatibility isn't necessary anymore.

**Related ticket:** https://secure.helpscout.net/conversation/1373096881/224963?folderId=2135277

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

vmanthos:
Fixes #3456

I've tested this both on my test site, and on the customer's website.

### DIFF
--- a/inc/classes/subscriber/third-party/plugins/class-ngg-subscriber.php
+++ b/inc/classes/subscriber/third-party/plugins/class-ngg-subscriber.php
@@ -39,6 +39,10 @@ class NGG_Subscriber implements Subscriber_Interface {
 	 * @return bool
 	 */
 	public function deactivate_resource_manager() {
+		if( is_admin() ) {
+			return true;
+		}
+		
 		return false;
 	}
 }
--- a/inc/classes/subscriber/third-party/plugins/class-ngg-subscriber.php
+++ b/inc/classes/subscriber/third-party/plugins/class-ngg-subscriber.php
@@ -39,10 +39,6 @@ class NGG_Subscriber implements Subscriber_Interface {
 	 * @return bool
 	 */
 	public function deactivate_resource_manager() {
-		if( is_admin() ) {
-			return true;
-		}
-		
-		return false;
+		return is_admin();
 	}
 }


inc/classes/subscriber/third-party/plugins/class-ngg-subscriber.php
```diff
@@ -39,6 +39,10 @@ public static function get_subscribed_events() {
 	 * @return bool
 	 */
 	public function deactivate_resource_manager() {
+		if( is_admin() ) {
+			return true;
+		}
+		
 		return false;
```
remyperona:
```suggestion
		return is_admin();
```

This block of code can be simplified to the suggestion above. `is_admin()` already returns a boolean value, so we can use that directly for the return, and turn 4 lines of code into one 🎉

remyperona (changes requested):

vmanthos:
That's a very nice touch, @Tabrisrp!

Thank you!

remyperona (approved):

remyperona (approved):

PR #3551, opened 2021-02-03
Add filters to set file name externally and force page loading

TzipporahWitty:
This pull request adds two filters:
1, a filter to change the file name at the point where it's generated for minified JS and CSS files.  This will let other plugins implement logic for determining the minified file name before the page is loaded. 
2, a filter to allow changing the URL inserted into the HTML of the page.  This will let other plugins modify the URL if needed, such as to add a cache busting query parameter.

### DIFF
--- a/inc/Engine/Optimization/Minify/CSS/Combine.php
+++ b/inc/Engine/Optimization/Minify/CSS/Combine.php
@@ -199,6 +199,14 @@ class Combine extends AbstractCSSOptimization implements ProcessorInterface {
 
 		$file_hash      = implode( ',', array_column( $this->styles, 'url' ) );
 		$this->filename = md5( $file_hash . $this->minify_key ) . '.css';
+		
+		/**
+		 * Allow overwriting file name to be saved.
+		 *
+		 * @since 3.8.9
+		 *
+		 * @param string $filename The file name already generated.
+		 */
 		$this->filename      = apply_filters( 'rocket_minified_css_file_name', $this->filename );
 
 		$combined_file = $this->minify_base_path . $this->filename;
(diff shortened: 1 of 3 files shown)


remyperona:
Hi 👋🏼 

First, thank you for contributing directly to the plugin like this, we appreciate it!

**About the proposed changes**
For the URL of the CSS/JS files, we already have filters available. See the following:
https://github.com/wp-media/wp-rocket/blob/ed5935ef5f7481a10f1933d0a3a119abe1cdea05/inc/Engine/Optimization/Minify/CSS/AbstractCSSOptimization.php#L102

https://github.com/wp-media/wp-rocket/blob/ed5935ef5f7481a10f1933d0a3a119abe1cdea05/inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php#L150

For the filenames, I think this is fine, but I have 2 feedbacks:
- We need the appropriate PHP Docblock added for each filter
- You also might need it not only in the `Combine` classes, but in the `Minify` ones too, as there is a specific filename generated too there

TzipporahWitty:
Hi! I'm sorry this took me so long to get back to. 

The reason that I didn't want to use the existing filters is that the get_minify_url() function is used in multiple places, and adding a cache busting parameter would only be relevant in this particular place.  Would it be better to name the filter something like 'rocket_css_stylesheet_source' to indicate that it is for use specifically when the need is to modify the URL used in the <link> tag?

I've added the PHP Docblocks for the filters now.

For our use case, we don't need to change the names generated in the Minify class, but if that would be a better contribution I can certainly add in the filters there as well.  In that case, would it be better to create two sets of filters, one for combining and one for minifying?

PR #3553, opened 2021-02-03
Exclude application/ld+json scripts from DeferJS

### LINKED ISSUE #3470: Exclude application/ld+json inline scripts from DeferJS

**Describe the bug**
We need to exclude `application/ld+json` from being deferred as this inline JS has json contents and on some edge cases it contains the keyword `jQuery` so it's enclosed between our
```
window.addEventListener('DOMContentLoaded', function(){
HERE
});
```
and this isn't valid.

**To Reproduce**
1. Create page template which has the following script
```
<script type="application/ld+json">
{"text": "Technology &amp; Services. xhtml. SEO campaign. Javascript &ndash; Jquery &amp; Mootols Library&rsquo;s."}
</script>
```
2. Enable DeferJS
3. Check the optimized page source.

**Expected behavior**
This json shouldn't be enclosed between `DOMContentLoaded` event.

**Additional context**
@GeekPress Commented on that issue with the following message:
> it seems to be an edge case. But to be sure it won’t happen again, we should exclude application/ld+json inline scripts. It’s safe as it won’t contains any executed code.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

vmanthos:
Fixes #3470

### DIFF
--- a/inc/Engine/Optimization/DeferJS/DeferJS.php
+++ b/inc/Engine/Optimization/DeferJS/DeferJS.php
@@ -147,8 +147,8 @@ class DeferJS {
 			if ( empty( $inline_js['content'] ) ) {
 				continue;
 			}
-			
-			if ( preg_match( "/(application\/ld\+json)/i", $inline_js[0] ) ) {
+
+			if ( preg_match( '/(application\/ld\+json)/i', $inline_js[0] ) ) {
 				continue;
 			}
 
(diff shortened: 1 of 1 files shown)


remyperona (approved):

remyperona:
@wp-media/qa Ready for QA

remyperona:
@vmanthos There is a couple of code styling issues that should be fixed before merging. Doesn't block QA though.

You can have a look at them here: https://github.com/wp-media/wp-rocket/pull/3553/checks?check_run_id=1822792674

They can be fixed automatically by running the following command in the root directory: `vendor/bin/phpcbf inc/Engine/Optimization/DeferJS/DeferJS.php`

piotrbak (approved):
Worked as expected:
[testrail-report-122.pdf](https://github.com/wp-media/wp-rocket/files/5946548/testrail-report-122.pdf)

PR #3562, opened 2021-02-05
Fixes #3036 Exclude Uncode scripts from deferJS

### LINKED ISSUE #3036: Exclude Uncode file from deferred JS in latest theme version
**Describe the bug**
Based on this previous [issue](https://github.com/wp-media/wp-rocket/issues/1470), these files should be excluded from Defer JS feature:
```
/wp-content/themes/uncode/library/js/min/init.min.js
/wp-content/themes/uncode/library/js/init.js 
```
to avoid some JS errors like:
image

And it was implemented on inc/3rd-party/themes/uncode.php, but it has been applied if the version is minor than 2.1 and Uncode latest version is 2.2.8.4.

**Expected behavior**
The init.js should be excluded automatically from Defer JS feature.

vmanthos:
Fixes #3036

### DIFF
--- a/inc/3rd-party/themes/uncode.php
+++ b/inc/3rd-party/themes/uncode.php
@@ -44,21 +44,20 @@ if ( 'uncode' === strtolower( $current_theme->get( 'Name' ) ) || 'uncode' === st
 	}
 	add_filter( 'rocket_excluded_inline_js_content', 'rocket_exclude_inline_js_uncode' );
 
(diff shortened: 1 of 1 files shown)


remyperona (approved):

piotrbak (approved):
Worked as expected:
[testrail-report-123.pdf](https://github.com/wp-media/wp-rocket/files/5946828/testrail-report-
123.pdf)

Issue #2456, opened 2020-03-19
License data not available when cloning to staging server


alfonso100:
**Describe the bug**
After enabling WP Rocket on a production site, and generating a staging website WP Rocket can't be activated because there is no license data available

**To Reproduce**
1. Install WP Rocket on production site
2. Clone to a staging site
3. Verify if the license is active or can be activated

**Expected behavior**
When cloning from production to staging, the WP Rocket license information should be kept.

**Screenshots**
![Captura de pantalla 2020-03-19 a las 11 32 45](https://user-images.githubusercontent.com/229406/77078720-bbec5680-69d5-11ea-8ce4-6c01e8958f33.png)

**Additional context**
Slack convo: https://wp-media.slack.com/archives/C08N8J6VC/p1584556692193700
I have created a staging site using WP Staging. Cloned from production https://rocket.alfonsocatron.com/
to staging
https://rocket.alfonsocatron.com/staging/
at the staging environment, I can’t activate WP Rocket because the license data is empty.
Reuploading the **license-data.php** file allows activating WP Rocket

These seem relevant: 
- https://github.com/wp-media/wp-rocket/issues/2456#issuecomment-604410958
- https://github.com/wp-media/wp-rocket/issues/2456#issuecomment-660229555

vmanthos:
Another ticket: https://secure.helpscout.net/conversation/1114379818/152705?folderId=2135277

GeekPress:
@alfonso100 During the migration, does the database has been migrated?

alfonso100:
@GeekPress yes, the DB is automatically migrated by the plugin. in this case, the staging database shares the same DB, just changing the prefix

GeekPress:
This is weird, if we have the license info in the DB, it shouldn't cause this issue.

@Tabrisrp Do you have an idea?

vmanthos:
I used the [WP Staging – DB & File Duplicator & Migration](https://wordpress.org/plugins/wp-staging/) plugin to create a staging site for a test I was running.

I didn't get the issue as Alfonso did.

The **license-data.php** wasn't there, but WP Rocket was properly activated on the staging site.

WP Rocket 3.5.1 was installed on the production site.

arunbasillal:
I had one of these issues where the prefix was different on staging and production. Exactly as this one - https://github.com/wp-media/wp-rocket/issues/2456#issuecomment-604410958

@vmanthos In your test, was the staging and the production using different prefixes?

[...]

PR #3563, opened 2021-02-05
Fixes #2804 Remove "disabled" attribute on Email address and API key fields

### LINKED ISSUE #2804: License validation sometimes fails when moving domains/creating staging sites
When creating a staging site or moving the site to a new domain, in many cases, a "Search & replace" is performed to replace the original domain, e.g. **domainA.com** with the new domain, e.g. **domainB.com**.

Since WP Rocket saves the WP Rocket's account email in the database, if the owner has an email like `user@domainA.com`, that will become `user@domainB.com`, and the license validation will fail.

The issue can be fixed by adding the license's info in the `wp-config.php`:
```PHP
define( 'WP_ROCKET_EMAIL', 'ACCOUNT_EMAIL');
define( 'WP_ROCKET_KEY', 'API_KEY');
```

**To Reproduce**
1. Create a site domainA.com.
2. Change its domain to domainB.com.
3. Perform a "Search & Replace" in the database to replace domainA.com with domainB.com.
4. Try to activate WP Rocket.

**Additional context**
WP Engine's staging creation mechanism seems to be doing the database replacement.

**Tickets:**
* https://secure.helpscout.net/conversation/1197140885/173483
* https://secure.helpscout.net/conversation/1194749805/172788/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

GeekPress:
Close #2804

### DIFF
--- a/inc/Engine/Admin/Settings/Page.php
+++ b/inc/Engine/Admin/Settings/Page.php
@@ -2064,6 +2064,28 @@ class Page {
 	 * @since 3.0
 	 */
 	private function hidden_fields() {
+		$hidden_fields = [
+			'secret_key',
+			'license',
+			'secret_cache_key',
+			'minify_css_key',
+			'minify_js_key',
+			'version',
+			'cloudflare_old_settings',
+			'sitemap_preload_url_crawl',
+			'cache_ssl',
+			'minify_google_fonts',
+			'emoji',
+		];
+
+		if ( ! empty( get_rocket_option( 'consumer_key', '' ) ) ) {
+			$hidden_fields[] = 'consumer_key';
+		}
+
+		if ( ! empty( get_rocket_option( 'consumer_email', '' ) ) ) {
+			$hidden_fields[] = 'consumer_email';
+		}
+
 		$this->settings->add_hidden_settings_fields(
 			/**
 			 * Filters the hidden settings fields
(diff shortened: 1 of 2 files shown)


remyperona (approved):

piotrbak:
@Tabrisrp I have some questions/concerns:
1. If the User/API is defined (and changed) in the wp-config.php, we won't be able to overwrite it anyway.
2. I'm not able to use this field correctly, it doesn't let me through when passing the correct data. `license-data.php` and `wp-config.php` definitions work just fine. Steps to reproduce:
- Validate your WP Rocket instance on the site
- Go to the Database Manager and modify `wp_rocket_settings` from the `wp_options`. Change your email to another one.
- Refresh WP Rocket dashboard, you should be redirected to the License Validation page
- Fill in the correct API data and Validate License
- I'm staying on the same screen with the following error:
`License validation failed. Our server could not resolve the request from your website.`
3. Do you think it's worth enabling the API field at all?

remyperona:
For 1), it's true but the issue also should not happen in that case, as those values in wp-config.php would not be modified during staging.

For 2), I pushed a code update to fix that

For 3), I'm thinking this change would also help us close #2456, what do you think?

piotrbak:
@Tabrisrp Here's what I did and it's not working as expected:
- Installed a site as `wp-media.me` using HOSTS file
- Used my license `piotr@wp-media.me`
- Changed the domain to `wp-medias.me`
- Local by Flywheel seems to be looking for `http://wp-media.me` when replacing, so I needed to run Search and Replace
- Did a search and replace
- WP Rocket licensing page is displayed with the `piotr@wp-medias.me`
- I'm changing it to the correct `piotr@wp-media.me` and validating the license
- Page reloads with the wrong `piotr@wp-medias.me`
- Tried a couple of times
Seems that the database has priority over the fields and the fields are not updating the DB.

Regarding 3. I think that customers will be still reaching our support to learn the API key. The difference is that they'll use fields instead of `wp-config.php`. What do you think?

remyperona:
Yeah, this issue is more complicated than expected to handle. I propose to put it back on the burner for now, and come back with a more robust approach.

remyperona:
@piotrbak Do we want to work on this in the future or close it?

piotrbak:
I think that the PR can be closed, we might look for a different approach at some point

PR #3571, opened 2021-02-08
Fixes #3437 Exclude image inside a picture element from lazyload

### LINKED ISSUE #3437: Images inside the <picture> tag are not excluded from lazy loading if the <picture> tag is excluded from lazy loading

**Describe the bug**
Images inside the <picture> tag are not excluded from lazy loading if the <picture> tag is excluded from lazy loading. 

**To Reproduce**
1. Add a `<picture class="picture_class">` with a class with a couple of `` inside.
2. Enable lazy loading for images.
3. Exclude "picture_class" from lazy loading.
4. `` are still lazy-loaded.

**Expected behavior**
When the picture tag is excluded, images should be excluded as well. 

**Screenshots**
Exclusions tested
https://i.vgy.me/6C1zyK.png

**Additional context**
Notes from Slack discussion: 

- Excluding picture via class is working correctly
- But the images inside picture are processed twice. So the LazyLoad rewrite happens on it regardless
- ^ This means the images in the picture needs to be excluded separately.
- The goal is to exclude all images mentioned within picture from LazyLoad if a picture is excluded.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

This PR adds the `data-skip-lazy` attribute to `img` inside a `picture`, when the picture is excluded from lazyload.

It also fixes a bug that would set the `data-lazy-srcset` on `source` when it should not.

Fixes https://github.com/wp-media/wp-rocket/issues/3437

## Type of change

## How Has This Been Tested?

- [x] Unit tests updated to validate the code update on the upstream repository

## Is there any change compared to the proposed solution during grooming?

Yes, the proposed solution was to ignore `picture` elements in the html when doing the lazyload for images, by changing the buffer value.

I took a different approach, to propose a self-contained solution, where we directly add an excluded attribute to the `img` inside a `picture`. That way, this `img` will be skipped by the lazyload for images.

It also avoids modifying the buffer, which can have unexpected side-effects.

## Checklist:

### DIFF
--- a/inc/Dependencies/RocketLazyload/Image.php
+++ b/inc/Dependencies/RocketLazyload/Image.php
@@ -253,6 +253,16 @@ class Image {
 
 		foreach ( $pictures as $picture ) {
 			if ( $this->isExcluded( $picture[0], $excluded ) ) {
+				if ( ! preg_match( '#<img(?<atts>\s.+)\s?/?>#iUs', $picture[0], $img ) ) {
+					continue;
+				}
+	
+				$img = $this->canLazyload( $img );
+	
+				if ( ! $img ) {
+					continue;
+				}
+
 				$nolazy_picture = str_replace( '<img', '<img data-skip-lazy=""', $picture[0] );
 				$html           = str_replace( $picture[0], $nolazy_picture, $html );
 
--- a/inc/Dependencies/RocketLazyload/Image.php
+++ b/inc/Dependencies/RocketLazyload/Image.php
@@ -253,25 +253,30 @@ class Image {
 
 		foreach ( $pictures as $picture ) {
 			if ( $this->isExcluded( $picture[0], $excluded ) ) {
+				$nolazy_picture = str_replace( '<img', '<img data-skip-lazy=""', $picture[0] );
+				$html           = str_replace( $picture[0], $nolazy_picture, $html );
+
 				continue;
 			}
 
-			$lazy_sources = 0;
-
 			if ( preg_match_all( '#<source(?<atts>\s.+)>#iUs', $picture['sources'], $sources, PREG_SET_ORDER ) ) {
-				$sources = array_unique( $sources, SORT_REGULAR );
+				$lazy_sources = 0;
+				$sources      = array_unique( $sources, SORT_REGULAR );
+				$lazy_picture = $picture[0];
 
 				foreach ( $sources as $source ) {
 					$lazyload_srcset = preg_replace( '/([\s"\'])srcset/i', '\1data-lazy-srcset', $source[0] );
-					$html            = str_replace( $source[0], $lazyload_srcset, $html );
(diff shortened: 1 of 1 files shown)


remyperona:
@wp-media/qa ready for QA

Mai-Saad (commented):
<strike> can you please check @Tabrisrp?
https://user-images.githubusercontent.com/76941962/107516558-41ac5e00-6bb5-11eb-85cf-722ea0db526f.mp4
</strike>
Update: Well seems it is not correct to add more than 1  inside the picture tag.
But even with single img in the picture tag, if we excluded by a filename, in case this file name exists in pic Elem but commented, it will exclude the pic Elem -> is it fine?

- code: https://snippi.com/s/ilx28ay 

- exclude 51qaqy1SAML.jpg will exclude src="http://mega.wp-rocket.me/wp-content/uploads/2020/10/plaskaziemia_xD-flat-earth-lol.png.webp"

**Notes**
1- At trunk: enable lazyload will apply on all images even commented one, so I think this is the root for what mentioned above

2- If img inside pic tag was already excluded from the code, we add another data-skip-lazy=" " attribute if we enabled LL -> I think we shouldn't add the attribute again here.
code sample before enabling LL:
http://snippi.com/s/0k6h8tj

3- If img was skipped in HTML using skip-lazy class, shall we add too data-skip-lazy attribute? (yes is what is happening now) while for img not in pic Elem we don't add skip-lazy if it was already in html

@Tabrisrp what do you think?

piotrbak:
@Tabrisrp @Mai-Saad Adding the same attribute twice to the HTML tag is a problem, we shouldn't do this:
https://html.spec.whatwg.org/multipage/syntax.html#attributes-2
`There must never be two or more attributes on the same start tag whose names are an ASCII case-insensitive match for each other.`

remyperona:
@wp-media/qa Thanks for finding this, I pushed a change to prevent adding the exclusion attribute again if the image already contains an excluded pattern

piotrbak (approved):
Worked as expected, covered scenarios:
1. Image is excluded using `img` src or `img` class - excluded without the attribute
2. Image contains the excluded keywords (`skip-lazy`) - excluded without the attribute
3. Picture class is excluded - excluded with the attribute
4. Source image is excluded - excluded with the attribute
5. Two exclusions are done, `img` and `picture` - excluded without the attribute
6. Two exclusions are done, `img` and `source` - excluded without the attribute
7. Helper plugin on picture class - excluded with the attribute

piotrbak:
@Tabrisrp As a comment:
If there's a commented out image in the `<picture>` and the exclusion pattern matches only it, not the visible images- we'll exclude the whole `<picture>` anyways.
https://snippi.com/s/7ssq72g

I think this is an edge case, what do you think?

remyperona:
I agree, this should not happen often. And we plan to move to native lazyload for images anyway, so it's not worth too much trouble.

PR #3575, opened 2021-02-09
Fix the NextGEN gallery regression

### LINKED ISSUE #3568: 3.8.4 breaks media library and many other plugins
On our websites:
PHP 7.3 and 7.4
WP 5.5.3

After the upgrade to WP Rocket 3.8.4 we cannot browse the library anymore. No content is shown. We cannot upload new pictures too in wp articles too. Even the plugin WP Optimize is broken.

We had another website still on V3.8.3 and we used those WP-Rocket files to downgrade to V 3.8.3 and now everything is fine again.

vmanthos:
Fixes #3568.

Solution based on the following comment: https://github.com/wp-media/wp-rocket/issues/3568#issuecomment-775919146

### DIFF
--- a/inc/classes/subscriber/third-party/plugins/class-ngg-subscriber.php
+++ b/inc/classes/subscriber/third-party/plugins/class-ngg-subscriber.php
@@ -36,6 +36,8 @@ class NGG_Subscriber implements Subscriber_Interface {
 	 * @since 3.3.1
 	 * @author Remy Perona
 	 *
+	 * @param bool $valid_request Indicates if the current request is valid for the NGG resource manager.
+	 *
 	 * @return bool
 	 */
 	public function deactivate_resource_manager( $valid_request ) {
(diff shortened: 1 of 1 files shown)


inc/classes/subscriber/third-party/plugins/class-ngg-subscriber.php
```diff
@@ -38,7 +38,11 @@ public static function get_subscribed_events() {
 	 *
```
remyperona, line 38:
```suggestion
	 *
	 * @param bool $valid_request Indicates if the current request is valid for the NGG resource manager.
	 *
```

remyperona (changes requested):

remyperona (approved):

piotrbak:
@Tabrisrp This one will be pretty hard to check. I think we'll be able to check only 2 scenarios:
1. NextGen + one of the affected plugins
2. NextGen itself

Do you think this is enough?

remyperona:
@piotrbak Could we validate the change on customers website affected by the issue?

piotrbak:
@Tabrisrp I'm albo to reproduce it and also will try on customer's site

piotrbak (approved):
@Tabrisrp Tested on customer's site, reproduced on my own, and tested too:
[testrail-report-128.pdf](https://github.com/wp-media/wp-rocket/files/5959025/testrail-report-128.pdf)

PR #3579, opened 2021-02-09
Fixes #3238 Remove empty values in $purge_urls array

### LINKED ISSUE #3238: PHP Warning: array_flip(): Can only flip STRING and INTEGER in inc/common/purge.php on line 144
- Made sure you’re on the latest version ✔️ 
- Used the search feature to ensure that the bug hasn’t been reported before ✔️ 

**Describe the bug**
A customer reported they are receiving this warning when they try to clear the cache:

```
Warning: array_flip(): Can only flip STRING and INTEGER values! in /wp-content/plugins/wp-rocket/inc/common/purge.php on line 144
Warning: Cannot modify header information - headers already sent by (output started at /wp-content/plugins/wp-rocket/inc/common/purge.php:144) in /wp-admin/post.php on line 231
```
It is happening here: 
https://github.com/wp-media/wp-rocket/blob/master/inc/common/purge.php#L144

**Expected behavior**
Per Tonya's comment, this is something we should guard against in Rocket. By using a guard clause or type cast to a string.

**Additional context**
Related ticket: https://secure.helpscout.net/conversation/1317241097/204840?folderId=2683093
Remy reviewed the ticket . Here are his notes:

> "not related to the update, they must have a strange value in there somewhere when we collect the URLs to purge. Something returning false maybe instead of the URL"

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

NataliaDrause:
## Description

Use `array_filter` to remove invalid values from the `$purge_urls` array.

All the values that didn't return`false` (such as random strings, integers, `true` boolean) and were passed into the $purge_urls array caused PHP notices regardless of the method `array_flip()` or `array_unique()` is used, but didn't cause the Warning from the GH issue.

Example: `PHP Notice: Trying to access array offset on value of type null in .../wp-content/plugins/wp-rocket/inc/functions/files.php on line 540`

Fixes #3238

## Type of change

## How Has This Been Tested?

Tested with a variety of values passed into the $purge_urls array. Tests and notes added in the ticket: https://secure.helpscout.net/conversation/1413525677/236164/

# Checklist:

### DIFF
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -140,7 +140,10 @@ function rocket_get_purge_urls( $post_id, $post ) {
 			$purge_urls[] = get_permalink( $parent_id );
 		}
 	}
-
+	
+	// Remove entries with empty values in array
+	$purge_urls = array_filter( $purge_urls );
+	
 	return array_flip( array_flip( $purge_urls ) );
 }
 
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -142,7 +142,7 @@ function rocket_get_purge_urls( $post_id, $post ) {
 	}
 
 	// Remove entries with empty values in array.
-	$purge_urls = array_filter( $purge_urls );
+	$purge_urls = array_filter( $purge_urls, 'is_string' );
 
 	return array_flip( array_flip( $purge_urls ) );
 }
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -140,10 +140,10 @@ function rocket_get_purge_urls( $post_id, $post ) {
 			$purge_urls[] = get_permalink( $parent_id );
 		}
 	}
-	
-	// Remove entries with empty values in array
+
+	// Remove entries with empty values in array.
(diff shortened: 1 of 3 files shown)


inc/common/purge.php
```diff
@@ -141,6 +141,9 @@ function rocket_get_purge_urls( $post_id, $post ) {
 		}
 	}
 
+	// Remove entries with empty values in array.
+	$purge_urls = array_filter( $purge_urls );
```
```suggestion
	$purge_urls = array_filter( $purge_urls );
```
```suggestion
	$purge_urls = array_filter( $purge_urls, 'is_string' );
```

This change will ensure we only preserve string values in the array

remyperona (changes requested):

remyperona (approved):

Mai-Saad (approved):
The PR itself working as expected, other issues exist in the clear cache itself when combined with AMP /Cookie/ user cache

PR #3582, opened 2021-02-10
Guard rocket_defer_inline_exclusions filter return when used by 3rd parties

### LINKED ISSUE #3576: Guard our code when the rocket_defer_inline_exclusions filter used by 3rd parties
- Made sure you’re on the latest version -yes

**Describe the bug**
3rd party plugins use our filter `rocket_defer_inline_exclusions` and pass a string instead of an array. This leads to a PHP warning in the debug.log:
`PHP Warning: Invalid argument supplied for foreach() in /home4/nataliad/testsite.nataliadrause.com/wp-content/plugins/wp-rocket/inc/Engine/Optimization/DeferJS/DeferJS.php on line 140`

Such errors were caused by compatibility codes in **ShortPixel AI** and **Smart Slider 3** plugins so far.

**To Reproduce**
1. Install and activate ShortPixel AI plugin https://wordpress.org/plugins/shortpixel-adaptive-images/
2. Enable _Load JavaScript Deferred_ option (with jQuery not being excluded from deferring)
3. Clear and preload cache.
4. The following error will be logged in debug.log:
```
09-Feb-2021 11:22:23 UTC] PHP Notice: Array to string conversion in .../wp-content/plugins/shortpixel-adaptive-images/includes/controllers/short-pixel-ai.class.php on line 1186
[09-Feb-2021 11:22:23 UTC] PHP Warning: Invalid argument supplied for foreach() in .../wp-content/plugins/wp-rocket/inc/Engine/Optimization/DeferJS/DeferJS.php on line 140
```

**Expected behavior**
There should be no errors.

**Additional context**
Code in ShortPixel AI:
```
public function wp_rocket_no_defer_spai_settings($regex) {
return $regex . '|spai_settings';
}
```

Comment by @engahmeds3ed:
> I think we need to guard it and make it backward compatible so we will detect if it's string we will change it to array by exploding using `|` needle

Slack convo: https://wp-media.slack.com/archives/C43T1AYMQ/p1612871496110000
Ticket: https://secure.helpscout.net/conversation/1412614288/235953/

Related issue with Smart Slider 3:
Slack: https://wp-media.slack.com/archives/C60QX4FC1/p1612456408103600
Ticket: https://secure.helpscout.net/conversation/1414913336/236541/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
## Description

I did two things here on this PR related to the return of `rocket_defer_inline_exclusions` filter:-
1. I casted its return to be always array.
2. For backward compatibility, I checked if the return of filter is string so I converted it to array by slicing it using separator `|`

Fixes #3576 

## Type of change

Yes It's like groomed solution.

## How Has This Been Tested?

- [x] Add the following code into active theme `functions.php` and then clear the cache and check `error_log` or `debug.log`
```
add_filter( 'rocket_defer_inline_exclusions', function( $exclusions ){
    return 'any_parameter';
} );
```
OR
```
add_filter( 'rocket_defer_inline_exclusions', function( $exclusions ){
    return [ 'any_parameter', 'second_parameter' ];
} );
```
OR
```
add_filter( 'rocket_defer_inline_exclusions', function( $exclusions ){
    return new StdClass();
} );
```

- [x] By using ShortPixel AI plugin as in the issue description.

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DeferJS/DeferJS.php
+++ b/inc/Engine/Optimization/DeferJS/DeferJS.php
@@ -294,11 +294,13 @@ class DeferJS {
 
 		$inline_exclusions = '';
 
-		if ( ! is_array( $additional_inline_exclusions_list ) ) {
-			$additional_inline_exclusions_list = explode( '|', (string) $additional_inline_exclusions_list );
+		// Check if filter return is string so convert it to array for backward compatibility.
+		if ( is_string( $additional_inline_exclusions_list ) ) {
+			$additional_inline_exclusions_list = explode( '|', $additional_inline_exclusions_list );
 		}
 
-		$inline_exclusions_list = array_merge( $inline_exclusions_list, $additional_inline_exclusions_list );
+		// Cast filter return to array.
+		$inline_exclusions_list = array_merge( $inline_exclusions_list, (array) $additional_inline_exclusions_list );
 
 		foreach ( $inline_exclusions_list as $inline_exclusions_item ) {
 			$inline_exclusions .= preg_quote( (string) $inline_exclusions_item, '#' ) . '|';
--- a/inc/Engine/Optimization/DeferJS/DeferJS.php
+++ b/inc/Engine/Optimization/DeferJS/DeferJS.php
@@ -142,7 +142,7 @@ class DeferJS {
 			}
 
 			foreach ( (array) $inline_exclusions_list as $inline_exclusions_item ) {
-				$inline_exclusions .= preg_quote( $inline_exclusions_item, '#' ) . '|';
+				$inline_exclusions .= preg_quote( (string) $inline_exclusions_item, '#' ) . '|';
 			}
 			$inline_exclusions = rtrim( $inline_exclusions, '|' );
 		}
--- a/inc/Engine/Optimization/DeferJS/DeferJS.php
+++ b/inc/Engine/Optimization/DeferJS/DeferJS.php
@@ -137,7 +137,11 @@ class DeferJS {
 
 		$inline_exclusions = '';
 		if ( ! empty( $inline_exclusions_list ) ) {
-			foreach ( $inline_exclusions_list as $inline_exclusions_item ) {
+			if ( is_string( $inline_exclusions_list ) ) {
+				$inline_exclusions_list = explode( '|', $inline_exclusions_list );
+			}
+
+			foreach ( (array) $inline_exclusions_list as $inline_exclusions_item ) {
 				$inline_exclusions .= preg_quote( $inline_exclusions_item, '#' ) . '|';
 			}
 			$inline_exclusions = rtrim( $inline_exclusions, '|' );
--- a/inc/Engine/Optimization/DeferJS/DeferJS.php
+++ b/inc/Engine/Optimization/DeferJS/DeferJS.php
@@ -294,13 +294,13 @@ class DeferJS {
 
 		$inline_exclusions = '';
 
-		if ( is_string( $additional_inline_exclusions_list ) ) {
-			$additional_inline_exclusions_list = explode( '|', $additional_inline_exclusions_list );
+		if ( ! is_array( $additional_inline_exclusions_list ) ) {
+			$additional_inline_exclusions_list = explode( '|', (string) $additional_inline_exclusions_list );
 		}
 
-		$inline_exclusions_list = array_merge( $inline_exclusions_list, (array) $additional_inline_exclusions_list );
+		$inline_exclusions_list = array_merge( $inline_exclusions_list, $additional_inline_exclusions_list );
 
-		foreach ( (array) $inline_exclusions_list as $inline_exclusions_item ) {
+		foreach ( $inline_exclusions_list as $inline_exclusions_item ) {
 			$inline_exclusions .= preg_quote( (string) $inline_exclusions_item, '#' ) . '|';
 		}
 
--- a/inc/Engine/Optimization/DeferJS/DeferJS.php
+++ b/inc/Engine/Optimization/DeferJS/DeferJS.php
@@ -126,26 +126,7 @@ class DeferJS {
 		 */
 		$jquery_patterns = apply_filters( 'rocket_defer_jquery_patterns', 'jQuery|\$\.\(|\$\(' );
 
-		/**
-		 * Filters the patterns used to find inline JS that should not be deferred
-		 *
-		 * @since 3.8
-		 *
-		 * @param array $inline_exclusions_list Array of inline JS that should not be deferred.
-		 */
-		$inline_exclusions_list = apply_filters( 'rocket_defer_inline_exclusions', $this->inline_exclusions );
-
-		$inline_exclusions = '';
-		if ( ! empty( $inline_exclusions_list ) ) {
-			if ( is_string( $inline_exclusions_list ) ) {
-				$inline_exclusions_list = explode( '|', $inline_exclusions_list );
-			}
-
-			foreach ( (array) $inline_exclusions_list as $inline_exclusions_item ) {
-				$inline_exclusions .= preg_quote( (string) $inline_exclusions_item, '#' ) . '|';
-			}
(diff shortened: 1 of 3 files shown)


remyperona:
@engahmeds3ed Is the PR ready for review? It's still in draft for now.

engahmeds3ed:
> @engahmeds3ed Is the PR ready for review? It's still in draft for now.

Yes it's ready for review.
I didn't notice that it's draft, changed now.

crystinutzaa (approved):

engahmeds3ed:
@wp-media/qa ready for QA

piotrbak:
@engahmeds3ed Before we were getting those errors with Smart Slider:
https://snippi.com/s/eznltbk

After this update, the notice `Array to string conversion in` in their plugin is still present. Is it expected?

Their code:
https://snippi.com/s/0keypyy

This is the 56th line:
`$regexp .= '|';`

engahmeds3ed:
@piotrbak 
Yes we are guarding our filter return not to let any plugin return any type of data from this filter so here we expect array and loop on it, so if any code returned string from this filter, our code will give the following warning
```
PHP Warning:  Invalid argument supplied for foreach() in /wp-content/plugins/wp-rocket/inc/Engine/Optimization/DeferJS/DeferJS.php on line 141
```

but the first warning on your snippet is from smartslider code so they need to adjust it.

piotrbak:
@engahmeds3ed When Smart Slider or ShortPixel are enabled, this is how the `$inline_exclusions` looks like:
`[2021-02-19 19:20:41] $inline_exclusions: Array|N2R|spai_settings`
https://github.com/wp-media/wp-rocket/blob/b8f237caa52d2dfe4de305ea2c0d25c17e41696d/inc/Engine/Optimization/DeferJS/DeferJS.php#L149

Do we expect a multidimensional array there?

engahmeds3ed:
@piotrbak 

No, they are trying to append string data to array so it appears like what you sent.
we expect array and guarded it from being changed.

piotrbak:
Our exclusions are not working when their plugins are enabled:
image

Is that something we can take care of on our end @engahmeds3ed ?

Before this update, the `$inline_exclusions` was empty, causing nothing to be enclosed in the DomContentLoaded.

Mai-Saad:
@engahmeds3ed Please check this case as it results in a fatal error (not sure if it`s a case to happen in reality):
Add the following code into active theme functions.php and then clear the cache and check error_log or debug.log
```
add_filter( 'rocket_defer_inline_exclusions', function( $exclusions ){
    return [ 1, 2];
} );
```

engahmeds3ed:
@piotrbak Yes we can handle this case, I added a check to fill the `$inline_exclusions` with the default list pattern when the filter return is empty or not valid.

@Mai-Saad I adjusted the code now to handle array of any type.

engahmeds3ed:
Now I changed the way we deal with this filter so this filter return will be added to our main list (no changing it anymore) and make the filter default value to be null so it can accept any time and then do the type check to convert this additional exclusions from string to be array (if it's string) or cast it to array (if it's not an array).

@wp-media/php Please validate before sending again to @wp-media/qa

inc/Engine/Optimization/DeferJS/DeferJS.php
```diff
@@ -289,4 +274,36 @@ public function exclude_jquery_upgrade() {
...
+
+		$inline_exclusions = '';
+
+		if ( is_string( $additional_inline_exclusions_list ) ) {
+			$additional_inline_exclusions_list = explode( '|', $additional_inline_exclusions_list );
+		}
+
+		$inline_exclusions_list = array_merge( $inline_exclusions_list, (array) $additional_inline_exclusions_list );
+
+		foreach ( (array) $inline_exclusions_list as $inline_exclusions_item ) {
```
crystinutzaa: I don't think type casted in there is needed.
`array_merge` function always return an array

inc/Engine/Optimization/DeferJS/DeferJS.php
```diff
@@ -289,4 +274,36 @@ public function exclude_jquery_upgrade() {
...
+		$inline_exclusions_list = $this->inline_exclusions;
+
+		/**
+		 * Filters the patterns used to find inline JS that should not be deferred
+		 *
+		 * @since 3.8
+		 *
+		 * @param array $inline_exclusions_list Array of inline JS that should not be deferred.
+		 */
+		$additional_inline_exclusions_list = apply_filters( 'rocket_defer_inline_exclusions', null );
```
crystinutzaa, line 293: Why do you use `null` on apply_filters? 
`$additional_inline_exclusions_list` should be an array, right?
So why not using the default empty array?

inc/Engine/Optimization/DeferJS/DeferJS.php
```diff
@@ -289,4 +274,36 @@ public function exclude_jquery_upgrade() {
...
+
+		$inline_exclusions = '';
+
+		if ( is_string( $additional_inline_exclusions_list ) ) {
```
crystinutzaa, line 298: Wouldn't be better to guard agaist `! is_array()`? 
And always `$additional_inline_exclusions_list` at this point to be converted to array?
Will reduce the numerous type castings from below.

tests/Fixtures/inc/Engine/Optimization/DeferJS/Subscriber/deferInlineJs.php
```diff
@@ -143,4 +143,126 @@ function newContent() {
 		'html'     => $html,
 		'expected' => $expected,
 	],
+
+	'testShouldExcludeUsingStringFilter' => [
+		'config' => [
+			'rocket_defer_inline_exclusions_filter' => 'first_string|third_string',
```
crystinutzaa, line 149: What about adding a Bool (true / false), Int, Float / double, Null exclusion?
Should we add these types to fixtures?

inc/Engine/Optimization/DeferJS/DeferJS.php
engahmeds3ed: as discussed with you on slack, we need to be compatible with both string and array types for backward compatibility.

inc/Engine/Optimization/DeferJS/DeferJS.php
engahmeds3ed: this will be good with all types except for objects so when casting object to string and this object didn't implement the magic method `__toString` it will give an error but if u tried to cast this object to array it will work.

crystinutzaa (approved):

engahmeds3ed:
@wp-media/qa ready for another round of QA

piotrbak:
@engahmeds3ed Tested scenarios and results:

```
add_filter( 'rocket_defer_inline_exclusions', function( $exclusions ){
    $exclusions[] = "test1";
    $exclusions[] = "test2";
    return $exclusions;
} );
```
1. Works when plugins are disabled
2. When plugins are enabled, it adds `Array` to the `$inline_exclusions` and we're not excluding the pattern 🐛 

```
add_filter( 'rocket_defer_inline_exclusions', function( $exclusions ){
    return "test3";
} );
```
1. Works when plugins are disabled
2. Works when plugins are enabled

```
add_filter( 'rocket_defer_inline_exclusions', function( $exclusions ){
    $exclusions .= "test4|test5";
    return $exclusions;
} );
```
1. Works when plugins are disabled
2. Works when plugins are enabled

piotrbak:
@engahmeds3ed @Tabrisrp As we talked, it's ready to go. However, the most reliable way to exclude more than one attribute from DeferJS, that will also work with the current compatibility of Smart Slider3 + ShortPixel Adaptive Images is this one:
```
add_filter( 'rocket_defer_inline_exclusions', function( $exclusions ){
    $exclusions .= "test4|test5";
    return $exclusions;
} );
```

piotrbak (approved):
Worked as expected:
[testrail-report-138.pdf](https://github.com/wp-media/wp-rocket/files/6045593/testrail-report-138.pdf)

PR #3587, opened 2021-02-12
Fixes #3083 Update delay JS RegEx to ignore space inside script tags

### LINKED ISSUE #3083: Delay JS - RegEx error when there are spaces in the </script> tag

**Describe the bug**

When there are spaces in the `</script>` tag, like `</script >` the RegEx in `preg_replace_callback()`:
```PHP
return preg_replace_callback( '/<script\s*(?<attr>[^>]*)?>(?<content>.*)?<\/script>/Uis', [ $this, 'replace_scripts' ], $html );
```
 will select the content until the next script tag, that doesn't have spaces.

If in that content there are scripts from our default delay JavaScript list, the whole content will be replaced, and the page can break.

You can see the RegEx issue here: [https://regex101.com/r/hZiXud/1/](https://regex101.com/r/hZiXud/1/)

**To Reproduce**

1. Add the following code on a page: http://snippi.com/raw/2uaob82
2. Enable **Delay JS**
3. Check the source code of the cached page for one of the classes in the snippet, e.g. `#thrive`
4. The code will not be there

**Expected behavior**

Properly pick up the closing script tag regardless of whether it includes spaces or not.

**Additional context**

**Related ticket:** https://secure.helpscout.net/conversation/1277671782/193430?folderId=2135277

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

vmanthos:
## Description

I've updated the RegEx to ignore spaces either in the opening or closing `script` tag.

Fixes #3083.

## Type of change

Yes, it is. I've added extra checks for spaces in the opening `<script>` tag, as well as the closing tag, to cover all cases.

## How Has This Been Tested?

A test run against the breaking sample:
https://regex101.com/r/hZiXud/1/

Which is working fine with the new RegEx: https://regex101.com/r/5RWVBd/1

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DelayJS/HTML.php
+++ b/inc/Engine/Optimization/DelayJS/HTML.php
@@ -88,7 +88,7 @@ class HTML {
 	 * @return string
 	 */
 	private function parse( $html ) {
-		$replaced_html = preg_replace_callback( '/<\s*script\s*(?<attr>[^>]*)?>(?<content>.*)?<\s*\/\s*script\s*>/Uims', [ $this, 'replace_scripts' ], $html );
+		$replaced_html = preg_replace_callback( '/<\s*script\s*(?<attr>.*)?>(?<content>.*)?<\s*\/\s*script\s*>/Uims', [ $this, 'replace_scripts' ], $html );
 
 		if ( empty( $replaced_html ) ) {
 			return $html;
--- a/inc/Engine/Optimization/DelayJS/HTML.php
+++ b/inc/Engine/Optimization/DelayJS/HTML.php
@@ -88,7 +88,7 @@ class HTML {
 	 * @return string
 	 */
 	private function parse( $html ) {
-		$replaced_html = preg_replace_callback( '/<\s*script\s*(?<attr>.*)?>(?<content>.*)?<\s*\/\s*script\s*>/Uims', [ $this, 'replace_scripts' ], $html );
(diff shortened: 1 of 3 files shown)


remyperona (approved):

Mai-Saad (commented):
@vmanthos @Tabrisrp I have a couple of questions/notes:
1- Shall we support nested script? (In case application/xhtml+xml) -> currently we don't
2- What about the case of having > inside the string attribute? shouldn't we ignore it?
3- Spaces before script tag (opening "here script considered as content" or closure "js error exist") will result in a problem without cache `?nocache`  {This is same behavior with defer js and combine js but not delay js } Currently we are ignoring those cases with delay js which makes sense but on the other hand, it changes the behavior of browser if no cache.... is this fine?

remyperona:
1) Nested scripts should not be supported, they are not valid
2) Probably an edge case, but we might be able to update the RegEx to `<\s*script\s*(?<attr>.*)?>(?<content>.*)?<\s*\/\s*script\s*>` to avoid that
3) Yes it's fine

remyperona:
@Mai-Saad Pattern is updated, case 2 should pass now

Mai-Saad:
@Tabrisrp #2 still consider > inside attribute string as the end of the attribute itself
Screenshot from 2021-02-24 08-27-02

remyperona:
@Mai-Saad Sorry about that, I did another update that should work now

Mai-Saad (approved):
Since ignoring space before script tag is fine then this PR works as expected.
Note: we won't handle ">" inside attribute

PR #3595, opened 2021-02-16
Closes #3564 Update minified filename structure to use query string instead of version in filename

### LINKED ISSUE #3564: 404 issues on CSS & JS resources
Currently, WP-Rocket uses a mechanism that deletes the old Minified files and replaces them with new ones to accomplish cache-busting. As someone who makes some kind of changes to the websites almost every week, the files get regenerated every few days.

While this cannot be avoided, the problem here is that the old resources return a 404 error whenever anyone requests that resource. As you may already know, not every request made to a website will ask for its HTML. There are cached requests such as "Search Engine caches", "Static page caches", etc. When such a request is made, they will not look for the new updated HTML version of the page. It will result in a completely broken web page due to missing resources.

I did a little search and found that there are mainly two types of cache-busting mechanisms in place. One is rewriting the file names as WP-Rocket does. The second one is using a query string like "example.css?v=1111". In the second method, even when the query string is changed (even while the older version is removed), there is an actual file returned instead of a 404.

So, coming to the matter. I understand that there should be a way for cache-busting, so we cannot use a single file name. But isn't it possible to implement a "**Redirect**" from the old deleted file requests to the new one with some form of patterns? Or use the version number method instead of the file renaming?

- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
## Description

As mentioned in the issue itself we will add a query string (?ver=) with the last modified time of the CSS or JS files. It means we won't rely anymore on the minify key and the filename of the minified file won't contain it too.

Fixes #3564 

## Type of change

No

## How Has This Been Tested?

- [x] Create page template and add some css/js files into it then enable minify css and minify js ( uncheck combine css/js ) and you will find the minified file structre changed to be the same filename with suffix `?ver=`

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/Minify/CSS/Minify.php
+++ b/inc/Engine/Optimization/Minify/CSS/Minify.php
@@ -137,7 +137,7 @@ class Minify extends AbstractCSSOptimization implements ProcessorInterface {
 
 		$filename      = ltrim( rocket_realpath( $parsed_url['path'] ), '/' );
 		$minified_file = rawurldecode( $this->minify_base_path . $filename );
-		$minify_url    = $this->get_minify_url( $filename, $url ) . "?ver=" . rocket_direct_filesystem()->mtime( $minified_file );
+		$minify_url    = $this->get_minify_url( $filename, $url ) . '?ver=' . rocket_direct_filesystem()->mtime( $minified_file );
 
 		if ( rocket_direct_filesystem()->exists( $minified_file ) ) {
 			Logger::debug(
--- a/inc/Engine/Optimization/Minify/CSS/Minify.php
+++ b/inc/Engine/Optimization/Minify/CSS/Minify.php
@@ -135,10 +135,9 @@ class Minify extends AbstractCSSOptimization implements ProcessorInterface {
 			$url = rocket_add_url_protocol( $url );
 		}
 
-		$unique_id     = md5( $url . $this->minify_key );
-		$filename      = preg_replace( '/\.(css)$/', '-' . $unique_id . '.css', ltrim( rocket_realpath( $parsed_url['path'] ), '/' ) );
+		$filename      = ltrim( rocket_realpath( $parsed_url['path'] ), '/' );
 		$minified_file = rawurldecode( $this->minify_base_path . $filename );
-		$minify_url    = $this->get_minify_url( $filename, $url );
+		$minify_url    = $this->get_minify_url( $filename, $url ) . "?ver=" . rocket_direct_filesystem()->mtime( $minified_file );
 
 		if ( rocket_direct_filesystem()->exists( $minified_file ) ) {
 			Logger::debug(
--- a/inc/Engine/Optimization/Minify/CSS/Minify.php
+++ b/inc/Engine/Optimization/Minify/CSS/Minify.php
@@ -138,7 +138,7 @@ class Minify extends AbstractCSSOptimization implements ProcessorInterface {
 		$filename            = ltrim( rocket_realpath( $parsed_url['path'] ), '/' );
 		$minified_file       = rawurldecode( $this->minify_base_path . $filename );
 		$minified_file_mtime = rocket_direct_filesystem()->mtime( $minified_file );
-		$minify_url          = add_query_arg( 'ver', $minified_file_mtime ? $minified_file_mtime : '', $this->get_minify_url( $filename, $url ) );
+		$minify_url          = add_query_arg( 'ver', $minified_file_mtime ? $minified_file_mtime : md5( $url . $this->minify_key ), $this->get_minify_url( $filename, $url ) );
 
 		if ( rocket_direct_filesystem()->exists( $minified_file ) ) {
 			Logger::debug(
--- a/inc/Engine/Optimization/Minify/CSS/Minify.php
+++ b/inc/Engine/Optimization/Minify/CSS/Minify.php
@@ -135,10 +135,8 @@ class Minify extends AbstractCSSOptimization implements ProcessorInterface {
 			$url = rocket_add_url_protocol( $url );
 		}
 
-		$filename            = ltrim( rocket_realpath( $parsed_url['path'] ), '/' );
-		$minified_file       = rawurldecode( $this->minify_base_path . $filename );
-		$minified_file_mtime = rocket_direct_filesystem()->mtime( $minified_file );
-		$minify_url          = add_query_arg( 'ver', $minified_file_mtime ? $minified_file_mtime : md5( $url . $this->minify_key ), $this->get_minify_url( $filename, $url ) );
+		$filename      = ltrim( rocket_realpath( $parsed_url['path'] ), '/' );
+		$minified_file = rawurldecode( $this->minify_base_path . $filename );
 
 		if ( rocket_direct_filesystem()->exists( $minified_file ) ) {
 			Logger::debug(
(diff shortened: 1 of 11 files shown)


inc/Engine/Optimization/Minify/CSS/Minify.php
```diff
@@ -135,10 +135,9 @@ private function replace_url( $url ) {
 			$url = rocket_add_url_protocol( $url );
 		}
 
-		$unique_id     = md5( $url . $this->minify_key );
-		$filename      = preg_replace( '/\.(css)$/', '-' . $unique_id . '.css', ltrim( rocket_realpath( $parsed_url['path'] ), '/' ) );
+		$filename      = ltrim( rocket_realpath( $parsed_url['path'] ), '/' );
 		$minified_file = rawurldecode( $this->minify_base_path . $filename );
-		$minify_url    = $this->get_minify_url( $filename, $url );
+		$minify_url    = $this->get_minify_url( $filename, $url ) . '?ver=' . rocket_direct_filesystem()->mtime( $minified_file );
```
remyperona:
```suggestion
		$minify_url    = add_query_arg( 'ver', rocket_direct_filesystem()->mtime( $minified_file ), $this->get_minify_url( $filename, $url ) );
```

We can use the built-in WP function to add the query arg to the URL, what do you think?

inc/Engine/Optimization/Minify/JS/Minify.php
```diff
@@ -183,12 +183,10 @@ protected function replace_url( $url ) {
...
 		// This filter is documented in /inc/classes/optimization/class-abstract-optimization.php.
-		$url       = apply_filters( 'rocket_asset_url', $url, $this->get_zones() );
-		$unique_id = md5( $url . $this->minify_key );
-		$filename  = preg_replace( '/\.js$/', '-' . $unique_id . '.js', ltrim( rocket_realpath( wp_parse_url( $url, PHP_URL_PATH ) ), '/' ) );
-
+		$url           = apply_filters( 'rocket_asset_url', $url, $this->get_zones() );
+		$filename      = ltrim( rocket_realpath( wp_parse_url( $url, PHP_URL_PATH ) ), '/' );
 		$minified_file = rawurldecode( $this->minify_base_path . $filename );
-		$minified_url  = $this->get_minify_url( $filename, $url );
+		$minified_url  = $this->get_minify_url( $filename, $url ) . '?ver=' . rocket_direct_filesystem()->mtime( $minified_file );
```
remyperona:
```suggestion
		$minify_url    = add_query_arg( 'ver', rocket_direct_filesystem()->mtime( $minified_file ), $this->get_minify_url( $filename, $url ) );
```

inc/Engine/Optimization/Minify/JS/Minify.php
```diff
@@ -183,12 +183,10 @@ protected function replace_url( $url ) {
 		}
 
 		// This filter is documented in /inc/classes/optimization/class-abstract-optimization.php.
-		$url       = apply_filters( 'rocket_asset_url', $url, $this->get_zones() );
-		$unique_id = md5( $url . $this->minify_key );
-		$filename  = preg_replace( '/\.js$/', '-' . $unique_id . '.js', ltrim( rocket_realpath( wp_parse_url( $url, PHP_URL_PATH ) ), '/' ) );
-
+		$url           = apply_filters( 'rocket_asset_url', $url, $this->get_zones() );
+		$filename      = ltrim( rocket_realpath( wp_parse_url( $url, PHP_URL_PATH ) ), '/' );
```
remyperona: Could you add the same checks we do on the URL in the CSS/Minify to this one? the code between line 126-135. It could help avoid some errors.

remyperona (changes requested):

inc/Engine/Optimization/Minify/CSS/Minify.php
engahmeds3ed: awesome idea.
Done

inc/Engine/Optimization/Minify/CSS/Minify.php
```diff
@@ -135,10 +135,10 @@ private function replace_url( $url ) {
...
 		}
 
-		$unique_id     = md5( $url . $this->minify_key );
-		$filename      = preg_replace( '/\.(css)$/', '-' . $unique_id . '.css', ltrim( rocket_realpath( $parsed_url['path'] ), '/' ) );
-		$minified_file = rawurldecode( $this->minify_base_path . $filename );
-		$minify_url    = $this->get_minify_url( $filename, $url );
+		$filename            = ltrim( rocket_realpath( $parsed_url['path'] ), '/' );
+		$minified_file       = rawurldecode( $this->minify_base_path . $filename );
+		$minified_file_mtime = rocket_direct_filesystem()->mtime( $minified_file );
+		$minify_url          = add_query_arg( 'ver', $minified_file_mtime ? $minified_file_mtime : '', $this->get_minify_url( $filename, $url ) );
```
remyperona: How about using the `$this->minify_key` value in case `mtime()` returned false? That way we always have a query string value, instead of an empty one.

remyperona (changes requested):

remyperona (approved):

engahmeds3ed:
@wp-media/qa Ready for QA.

piotrbak:
@Tabrisrp @GeekPress I have some questions/concerns about this update.
1. We're changing how we're recognizing outdated minified files. Before we were recreating the md5 key during the cache clear, so all files that were using an old filename were not used anymore, no matter if they were removed from the `/min/` directory or not. Now, we'll only check for the existence in the `/min/` directory without the md5 key. It means that if for any reason the minified files are not removed, we'll serve outdated ones. On my FlyWheel Local instance, I'm not able to clear the minified files when clearing the cache. I wonder how is it going to look like on IIS servers. Here's the scenario before and after the update:
**Before:**
- Visit the homepage
- Edit the `style.css`
- Clear the cache, `style.css` is not removed (for any reason)
- Changes visible, because the md5 was regenerated so we're looking for a different file 
**After:**
- Visit the homepage
- Edit the `style.css`
- Clear the cache, `style.css` is not removed (for any reason)
- No changes on the homepage because we're looking at the `style.css` file existence

2. We should make sure that the `rocket_clean_minify` is used at exactly the same places where we are regenerating the keys.

3. We should also check what are the default settings of the CDN providers:
- If they're caching all query strings by default
- If they're ignoring query strings
- If they're not caching query strings at all

remyperona (approved):

GeekPress:
@piotrbak @Tabrisrp In this case, we should try to understand why the minified file isn't removed and try to fix the problem. By the way, it's already not normal to not being able to delete cached files. Finding the reason could lead to fixing other issues.

piotrbak:
@engahmeds3ed @Tabrisrp 

1. We're applying the query string to the saved dynamic content:
https://docs.wp-rocket.me/article/1305-save-dynamic-css-js-as-static-files 
However, the previous cache busting is applied there too. We can not rely on the query string with those files. Not sure if we want to preserve 2 cache busting methods there or prevent query string from being applied:
`dynamic_styles-60341d034bc7e971078664.css?ver=1614028039`

On trunk, it looks like we have a double cache busting added:
`/wp-content/cache/min/1/wp-content/cache/busting/1/wp-content/rocket-test-data/styles/dynamic_styles-6035197ac0352018327556-5c0a1e8197e0cec4740706a755a74210.css`
Without WP Rocket it looks like this:
`/wp-content/rocket-test-data/styles/dynamic_styles.php?ver=5.6.2`

2. Added this not regression bug:
https://github.com/wp-media/wp-rocket/issues/3612

Rest seems to be just fine.

engahmeds3ed:
@webtrainingwheels I think this one needs documentation check

PR #3596, opened 2021-02-16
Fixes #2883 UI improvement for Never Cache URL(s) placeholder

### LINKED ISSUE #2883: UI improvement for Never Cache URL(s) placeholder
Our current placeholder contains **/members/(.*)**

image
https://jmp.sh/gCI4slS

It appears to be confusing for someone who uses BuddyPress for example, as they assume that /members/ pages are excluded by default, and the field is prefilled for them.

Adding something generic like **/example/(.*)** for a placeholder may solve this.

Example ticket: https://secure.helpscout.net/conversation/1221765567/179601/

Presskopp:
## Description

change wording of Never Cache URL(s) placeholder

fixes #2883

No


remyperona:
Duplicate of #3631

PR #3603, opened 2021-02-17
Closes #3539 Combine @import rules into the minified CSS files

### LINKED ISSUE #3539: @import rules aren't combined into the minified CSS files
**Describe the bug**
When the minify CSS option is turned on, the @import rules aren't combined into the current CSS file.

This is an important problem for performance:
- it creates useless HTTP requests
- The files needs to be preloaded to not decrease the PageSpeed Insight Grade

I did some tests on my side to know if the default behavior of the library was to combined @import rules.

The response is: YES

Please note that only internal @import rules are combined by the library. It means an @import rules from Google Fonts won't be merged.

I did a simple test on smashingcoding website. I minified this file which contains 2 @import rules:
http://smashingcoding.com/wp-content/plugins/forget-about-shortcode-buttons/public/css/button-styles.css

I used the last version of the library by using this code:

```
<?php
header('Content-type: text/css');
header('Cache-Control: max-age=31536000');

require_once 'minify/Minify.php';
require_once 'minify/CSS.php';
require_once 'path-converter/ConverterInterface.php';
require_once 'path-converter/Converter.php';

use MatthiasMullie\Minify;

$minifier = new Minify\CSS();
$minifier->add('wp-content/plugins/forget-about-shortcode-buttons/public/css/button-styles.css');

echo $minifier->minify();
```

The result is available here:
http://smashingcoding.com/minify.php

As you can notice the @import rules has been combined into the minified CSS file.

It means we have a problem into WP Rocket to be able to combine the @import rules.

/!\ The $maxImportSize variable doesn't have any relation with this problem. Increasing its value won't have any effect.

**To Reproduce**
1. Go to smashingcoding.com
2. Activate the Minify CSS option
3. Check the source code of the CSS file which starts by http://smashingcoding.com/wp-content/cache/min/1/wp-content/plugins/forget-about-shortcode-buttons/public/css/button-styles-
4. See error. The @import rules are still present

**Expected behavior**
The internal @import rules should be combined into the current CSS file.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
## Description

We added a layer to combine imported CSS files into the main minified/combined recursively.

Fixes #3539 

## Type of change

I added only a small part to handle rewrite paths and combine imports resursively.

## How Has This Been Tested?

- [x] Added new fixtures to be passed to unit/integration tests.
- [x] Create page template with CSS file that has multiple imports and one of those imports has another import with some image paths and check if those are imported Also add another import for external file and it shouldn't be imported.

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/CSSTrait.php
+++ b/inc/Engine/Optimization/CSSTrait.php
@@ -387,53 +387,53 @@ trait CSSTrait {
 	 * @param string $url URL of the file.
 	 * @return bool True if external, false otherwise.
 	 */
-	 function is_external_path( $url ) {
-		 $file = get_rocket_parse_url( $url );
+	protected function is_external_path( $url ) {
+		$file = get_rocket_parse_url( $url );
 
-		 if ( empty( $file['path'] ) ) {
-			 return true;
-		 }
+		if ( empty( $file['path'] ) ) {
+			return true;
+		}
 
-		 $parsed_site_url = wp_parse_url( site_url() );
+		$parsed_site_url = wp_parse_url( site_url() );
 
-		 if ( empty( $parsed_site_url['host'] ) ) {
-			 return true;
-		 }
+		if ( empty( $parsed_site_url['host'] ) ) {
+			return true;
+		}
 
-		 // This filter is documented in inc/Engine/Admin/Settings/Settings.php.
-		 $hosts   = (array) apply_filters( 'rocket_cdn_hosts', [], [ 'all' ] );
-		 $hosts[] = $parsed_site_url['host'];
-		 $langs   = get_rocket_i18n_uri();
+		// This filter is documented in inc/Engine/Admin/Settings/Settings.php.
+		$hosts   = (array) apply_filters( 'rocket_cdn_hosts', [], [ 'all' ] );
+		$hosts[] = $parsed_site_url['host'];
+		$langs   = get_rocket_i18n_uri();
 
-		 // Get host for all langs.
-		 foreach ( $langs as $lang ) {
-			 $url_host = wp_parse_url( $lang, PHP_URL_HOST );
+		// Get host for all langs.
+		foreach ( $langs as $lang ) {
+			$url_host = wp_parse_url( $lang, PHP_URL_HOST );
 
-			 if ( ! isset( $url_host ) ) {
-				 continue;
-			 }
+			if ( ! isset( $url_host ) ) {
+				continue;
+			}
 
-			 $hosts[] = $url_host;
-		 }
+			$hosts[] = $url_host;
+		}
 
-		 $hosts = array_unique( $hosts );
+		$hosts = array_unique( $hosts );
 
-		 if ( empty( $hosts ) ) {
-			 return true;
-		 }
+		if ( empty( $hosts ) ) {
+			return true;
+		}
 
-		 // URL has domain and domain is part of the internal domains.
-		 if ( ! empty( $file['host'] ) ) {
-			 foreach ( $hosts as $host ) {
-				 if ( false !== strpos( $url, $host ) ) {
-					 return false;
-				 }
-			 }
+		// URL has domain and domain is part of the internal domains.
+		if ( ! empty( $file['host'] ) ) {
+			foreach ( $hosts as $host ) {
+				if ( false !== strpos( $url, $host ) ) {
+					return false;
+				}
+			}
 
-			 return true;
-		 }
+			return true;
+		}
 
-		 return false;
+		return false;
 	}
 
 	/**
--- a/inc/Engine/Optimization/CSSTrait.php
+++ b/inc/Engine/Optimization/CSSTrait.php
@@ -371,18 +371,6 @@ trait CSSTrait {
 	 */
 	private function get_internal_file_contents( $file, $base_path ) {
 		if ( $this->is_external_path( $file ) && wp_http_validate_url( $file ) ) {
-			/**
-			 * Filter Import external urls.
-			 *
-			 * @since 3.8.6
-			 *
-			 * @param bool Allow status.
-			 * @param string $file File path/URL.
-			 */
-			if ( (bool) apply_filters( 'rocket_allow_import_externals', false, $file ) ) {
-				return [ $file, rocket_direct_filesystem()->get_contents( $file ) ];
-			}
-
 			return [ $file, false ];
 		}
 
--- a/inc/Engine/Optimization/CSSTrait.php
+++ b/inc/Engine/Optimization/CSSTrait.php
@@ -198,6 +198,8 @@ trait CSSTrait {
 	/**
 	 * Replace local imports with their contents recursively.
 	 *
+	 * @since 3.8.6
+	 *
 	 * @param string $content CSS Content.
 	 * @param string $target Target CSS file path.
 	 *
@@ -347,6 +349,8 @@ trait CSSTrait {
 	/**
 	 * Get internal file full path and contents.
 	 *
+	 * @since 3.8.6
+	 *
 	 * @param string $file Internal file path (maybe external url or relative path).
 	 * @param string $base_path Base path as reference for relative paths.
 	 *
@@ -357,7 +361,7 @@ trait CSSTrait {
 			/**
 			 * Filter Import external urls.
 			 *
-			 * @since 3.6.8
+			 * @since 3.8.6
 			 *
 			 * @param bool Allow status.
 			 * @param string $file File path/URL.
@@ -369,6 +373,7 @@ trait CSSTrait {
 			return [ $file, false ];
 		}
 
+		// Check if this file is readable or it's relative path so we add base_path at it's start.
 		if ( ! rocket_direct_filesystem()->is_readable( $this->get_local_path( $file ) ) ) {
 			$ds   = rocket_get_constant( 'DIRECTORY_SEPARATOR' );
 			$file = $base_path . $ds . str_replace( '/', $ds, $file );
@@ -382,7 +387,7 @@ trait CSSTrait {
 	/**
 	 * Determines if the file is external.
 	 *
-	 * @since 3.8
+	 * @since 3.8.6
 	 *
 	 * @param string $url URL of the file.
 	 * @return bool True if external, false otherwise.
@@ -439,6 +444,8 @@ trait CSSTrait {
 	/**
 	 * Get local absolute path for image.
 	 *
+	 * @since 3.8.6
+	 *
 	 * @param string $url Image url.
 	 *
 	 * @return string Image absolute local path.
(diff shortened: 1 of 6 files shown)


engahmeds3ed:
I created new filter here to allow importing external css files and it's disabled be default, you can find it here

https://github.com/wp-media/wp-rocket/blob/43840d58d34a0c8a4f6ff74bcbfc0b34f1f06235/inc/Engine/Optimization/CSSTrait.php#L361-L369

inc/Engine/Optimization/CSSTrait.php
```diff
@@ -35,6 +35,10 @@ public function rewrite_paths( $source, $target, $content ) {
 		 */
 		$target = apply_filters( 'rocket_css_asset_target_path', $target );
 
+		$content = $this->move( $this->get_converter( $source, $target ), $content, $source );
+
+		$content = $this->combine_imports( $content, $target );
```
remyperona, line 41: shouldn't the `combine_imports()` be before the `move()` call?

inc/Engine/Optimization/CSSTrait.php
```diff
@@ -191,6 +195,128 @@ protected function move( ConverterInterface $converter, $content, $source ) {
...
+		/ix',
+		];
+
+		// find all relative imports in css.
+		$matches = [];
+		foreach ( $import_regexes as $import_regexe ) {
+			if ( preg_match_all( $import_regexe, $content, $regex_matches, PREG_SET_ORDER ) ) {
+				$matches = array_merge( $matches, $regex_matches );
+			}
+		}
```
remyperona, line 287:
```suggestion
			if ( ! preg_match_all( $import_regexe, $content, $matches, PREG_SET_ORDER ) ) {
				return $content;
			}
```

Wouldn't this be enough to achieve the same result? We already get an array of matches from `preg_match_all()`, I'm not sure why it would be needed to merge with another empty array.

remyperona (changes requested):

piotrbak:
@GeekPress @Tabrisrp Just a quick confirmation, we're all aligned that this kind of situation might break the layout, right?
`@import internal.css`
`@import external.css`

After the minification, `internal.css` rules will be below the `external.css` ones.

inc/Engine/Optimization/CSSTrait.php
engahmeds3ed: Simply No, because I need to rewrite urls/paths as you some paths are relative to source file so I need to move them to the new url structure to be relative to our new target then be able to get their contents then you can see here

https://github.com/wp-media/wp-rocket/blob/43840d58d34a0c8a4f6ff74bcbfc0b34f1f06235/inc/Engine/Optimization/CSSTrait.php#L309

I'm calling the main method recursively to after combine_imports works for first time.
So don't worry about applying `move` after combining imports contents.

inc/Engine/Optimization/CSSTrait.php
engahmeds3ed: We have two regex patterns here to match the following schema:-
1. `@import url(xxx)`
2. `@import 'xxx'`

So I'm looping on the array of regexes `$import_regexes` and getting matches then add them to this new array.
What do you think?

GeekPress:
@piotrbak Yes we are aligned on that

remyperona (approved):

inc/Engine/Optimization/CSSTrait.php
```diff
@@ -191,6 +195,132 @@ protected function move( ConverterInterface $converter, $content, $source ) {
...
+		if ( empty( $matches ) ) {
+			return $content;
+		}
+
+		$search  = [];
+		$replace = [];
+
+		// loop the matches.
+		foreach ( $matches as $match ) {
+			if ( apply_filters( 'rocket_skip_import_replacement', false, $match['path'], $match ) ) {
```
remyperona, line 307: the new filter requires a docblock

engahmeds3ed: Done

engahmeds3ed:
@wp-media/qa Ready for QA

piotrbak:
@engahmeds3ed @Tabrisrp 
For now, the import of the rules worked only for the absolute, internal path (4th point):
` @import url('https://rocketlabsqa.ovh/wp-content/rocket-test-data/styles/import-css-fourth.css');`

**Tested scenarios:**
1. `@import url('import-css-third.css');`
Regular relative
2. `@import url('import-css-second.min.css');`
Relative with .min.css
3. `@import url('https://rocketlabsqa.ovh/wp-content/rocket-test-data/styles/import-css-fourth.css');`
Absolute internal
4. `@import url("import-css-fifth.css");`
Inside `import-css-fifth.css` there's another import inside, used `"` instead of `'`
5. `@import url('../../../wp-includes/css/dashicons.min.css');`
Relative path with ../../
6. `@import url("https://www.w3schools.com/w3css/4/w3.css?ver=5.5.3");`
External path, used with the filter
7. `@import url('import-css-eight.css?ver=1.2.3');`
Regular relative with query string
8. `@import url("https://rocketlabsqa.dev/something.css");`
Stylesheet loading `text/html` instead of `text/css`
9. `@import "import-css-ninth.css";`
Relative without `url`
10. `@import url(import-css-tenth.css);`
Relative without `"` nor `'`
11. `@import "import-css-eleventh.css" print;`
Relative without `url` with `print`
12. `@import "import-css-twelfth.css" screen and (max-width: 10768px);`
Relative without `url` with `media`

Do you already know that we shouldn't cover any of the above scenarios? They're going to be tested on various of envs.

engahmeds3ed:
@piotrbak checking that now.

I tried before sending it to you most of these schemes, let me try them again and will send you results.

engahmeds3ed:
@piotrbak now all cases are working, please validate.
except only one case related to query strings so we need to take the decision here is to remove completely query strings when trying to replace imports or this will affect the dynamic styles?

piotrbak:
@Tabrisrp @engahmeds3ed 
1. Should we import and download the `print` and `media` files too? They're now downloaded and enclosed in the correct queries:
`@import "import-css-eleventh.css" print;`
`@import "import-css-twelfth.css" screen and (max-width: 10768px);`

2. Should we be able to exclude the `@import` files from being downloaded/minified? Using UI/filter?

3. Query strings will be handled as internal files, as the external filter is going to be removed (that's just a confirmation)

remyperona:
@piotrbak 
1) Yes it's fine, it's still an improvement as it will reduce HTTP requests, and minification will be applied on everything
2) I'm thinking we can start with a filter, and we'll see if we need to also enable it in the UI if we have requests for that (@engahmeds3ed)

engahmeds3ed:
@Tabrisrp @piotrbak 

Ok I will remove now the external filter and try to check the exclusion filter and verify if it's working or not and why.

engahmeds3ed:
@piotrbak I removed external filter and also removed query strings from the css files to get contents properly and it works fine for me (please validate)

for the exclusion:
```
add_filter( 'rocket_skip_import_replacement', function( $skipped, $match_path ){
   if ( strstr( $match_path, 'file2.css' ) ) {
       return true;
    }
   return $skipped;
}, 10, 2 );
```

This code is working with me and I think I know why it wasn't working with u

Because this filter applies after moving relative paths to the new structure like
`file2.css`
is converted to
`../../../../../css/file2.css`
so we need to search with this path OR use regular expression to detect it.

so tomorrow morning I will adjust this code to accept array of files so support team would be able to use it.

engahmeds3ed:
@wp-media/qa ready for another round of testing.

piotrbak:
@engahmeds3ed Looks good on the single site, tomorrow I'm going to test on different environments + exclusions.

If I add .js file with `@import`, we're going to add the content of it to the css file. Shouldn't we make sure that the extension is `.css`?

piotrbak:
@engahmeds3ed @Tabrisrp I tested on all our environments. The problem we're having now:
- If we add `@charset "UTF-8";` at the very top, the external rules (ones that are not fetched) will be added before the charset, which is not correct.
image

**It seems that the problem is also present on the trunk though:**
image


To consider (that's an edge case, of course):
- Not importing files different than `.css`. The behaviour I'm experiencing on the not cached page:
1. This error in the JS console:
`Refused to apply style from '/wp-content/rocket-test-data/scripts/print-helloworld.js' because its MIME type ('application/javascript') is not a supported stylesheet MIME type, and strict MIME checking is enabled.`
2. In the network tab it's canceled with 0 bytes downloaded:
image

engahmeds3ed:
@piotrbak I think it's based on the content of the file, so in my case I created JS file with css contents inside it and I got this into netwrok tab

image

`file1.css` which is called into the page
image

`file2.js` which is imported into the previous file
image

remyperona:
If it's based on the mime type, we could do a check on that before importing?

Or if it's too complicated, we can just exclude any import that doesn't have the `.css` extension, as this is really what is expected in that context, and anything different is not normal.

GeekPress:
@piotrbak Have you ever see someone importing a .js into a CSS file? For which reason, a plugin/theme will do that?

I'm asking the question to know if we aren't trying to fox something which will never happen and/or a very rare case.

piotrbak:
@GeekPress That's rather an edge case, I'm trying to find as many possible faults as possible before customers will do.

I've brought this up for discussion since it's possible that we'll have different behavior with/without WP Rocket:
1. Without, the import will fail and we'll not load the JS (based on this [screenshot](https://user-images.githubusercontent.com/26767037/109178936-4a8a5b80-7781-11eb-8f31-ecea1bfa56d8.png))
2. With, we'll add the JS to our CSS and we'll load it

GeekPress:
> That's rather an edge case, I'm trying to find as many possible faults as possible before customers will do.

@piotrbak I understand. But we shouldn't complexify our code to fix a mistake a user can do. 

In this specific case, I can't see why a JS file could be added by a plugin / theme / something else.

GeekPress:
From the definition of the @import: 

> The @import rule allows you to import a style sheet into another style sheet.

It clearly indicates that an @import is to import .css into another. How can a JavaScript file be used to import CSS rules?

piotrbak:
@GeekPress You're totally right about this, and this hypothetical situation would be an obvious mistake of the developer or customer. My only concern is that without WP Rocket, the browser will prevent from downloading this file, with WP Rocket we'll download it.

The other scenarios passed and we'll be ready to merge.

engahmeds3ed:
@Tabrisrp @piotrbak 

firstly I used this WordPress function `wp_check_filetype_and_ext` that uses multiple ways to detect the file type and one of them is fileinfo module and when doing some tests I found the following bugs that may be in some servers (I found it also in my machine) which is mime-type not returns `text/css` and returns other unknown types.

https://forums.aws.amazon.com/thread.jspa?messageID=904363

so I used the WordPress function `wp_check_filetype` that checks the extension only.

remyperona:
We can move forward with that, it's enough to validate.

engahmeds3ed:
@piotrbak ready for QA

piotrbak:
@Tabrisrp We're good there. Ready to merge:
[testrail-report-139.pdf](https://github.com/wp-media/wp-rocket/files/6052378/testrail-report-139.pdf)

PR #3611, opened 2021-02-22
Closes #3113 Stop removing empty lines in the .htaccess -- replacement PR

### LINKED ISSUE #3113: Stop removing the empty lines on .htaccess
Stop removing the empty lines on .htaccess since it doesn't cause any issue or make it slower to WP Rocket proper work. 

The .htaccess files sometimes have many "Sections" - which are always separated by lines starting with # - and having the line breaks removed just adds more mind-work to find what we're looking for. So I can't see any real improvement on having those empty lines removed, just like wp-config.php.

**Additional context**
From productboard:
https://wp-media.productboard.com/insights/shared-inbox/notes/9673577

iCaspar:
Replaces #3599 (Accidentally closed). 😳 
See OPR for previous work from the branch.

### DIFF
--- a/inc/functions/htaccess.php
+++ b/inc/functions/htaccess.php
@@ -18,7 +18,6 @@ function flush_rocket_htaccess( $remove_rules = false ) { // phpcs:ignore WordPr
 	 * Filters disabling of WP Rocket htaccess rules
 	 *
 	 * @since 3.2.5
-	 * @author Remy Perona
 	 *
 	 * @param bool $disable True to disable, false otherwise.
 	 */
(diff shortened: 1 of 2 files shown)


Mai-Saad:
@iCaspar @Tabrisrp If we added an empty line inside the WPR section in htaccess then change WPR settings, WPR settings inside htaccess will be overwritten so is it normal to remove the added empty line within the WPR section or it shouldn't be removed as same as other sections than WPR? (currently, this empty line manually added to the WPR section in htaccess, is removed after update WPR settings)

remyperona:
@Mai-Saad That's normal. The WP Rocket block is completely removed and re-added when the `.htaccess` is updated, so it will re-initialize to the default spacing for that part.

Mai-Saad (approved):
Work as expected.
[testrail-report-136.pdf](https://github.com/wp-media/wp-rocket/files/6029339/testrail-report-136.pdf)

PR #3613, opened 2021-02-22
Fixes #3528 Add AMP query string to cached query strings when activating the plugin

### LINKED ISSUE #3528: AMP query string is not added during the AMP plugin activation

**Describe the bug**
When activating the official AMP plugin, we're not adding the `amp` query string automatically. We need to either disable WP Rocket or save settings in the AMP plugin.

**To Reproduce**
1. Enable WP Rocket
2. Enable AMP plugin
3. Check the WP Rocket config file
4. AMP is not there

**Expected behavior**
We should add the query string during the AMP activation

**Additional context**
The config file is regenerated, but the query string is not added.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Move the function existance check to avoid the race condition.

Fixes #3528

## Type of change

No

## How Has This Been Tested?

- [x] Tested by activating/deactivating the AMP plugin, and checking the value in the config file

# Checklist:

### DIFF
--- a/inc/ThirdParty/Plugins/Optimization/AMP.php
+++ b/inc/ThirdParty/Plugins/Optimization/AMP.php
@@ -46,12 +46,13 @@ class AMP implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() {
 		$events = [
-			'activate_amp/amp.php'   => 'generate_config_file',
-			'deactivate_amp/amp.php' => 'generate_config_file',
-			'wp'                     => 'disable_options_on_amp',
+			'activate_amp/amp.php'       => 'generate_config_file',
+			'deactivate_amp/amp.php'     => 'generate_config_file',
+			'wp'                         => 'disable_options_on_amp',
+			'rocket_cache_query_strings' => 'is_amp_compatible_callback',
 		];
+
 		if ( function_exists( 'is_amp_endpoint' ) ) {
-			$events['rocket_cache_query_strings'] = 'is_amp_compatible_callback';
 			$events['update_option_amp-options']  = 'generate_config_file';
 		}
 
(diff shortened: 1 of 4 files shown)


iCaspar (approved):

Mai-Saad:
@Tabrisrp @iCaspar Please find my notes/questions below:
1- Activating WPR after activating AMP results in the following warning/notice 
`[08-Mar-2021 14:23:01 UTC] PHP Notice:  Undefined property: DOMElement::$attributes in /var/www/rocketlabsqa.ovh/htdocs/wp-content/plugins/amp/includes/sanitizers/class-amp-style-sanitizer.php on line 2997
[08-Mar-2021 14:23:01 UTC] PHP Warning:  Invalid argument supplied for foreach() in /var/www/rocketlabsqa.ovh/htdocs/wp-content/plugins/amp/includes/sanitizers/class-amp-style-sanitizer.php on line 2997
[08-Mar-2021 14:23:01 UTC] PHP Warning:  Couldn't fetch DOMElement. Node no longer exists in /var/www/rocketlabsqa.ovh/htdocs/wp-content/plugins/amp/includes/sanitizers/class-amp-style-sanitizer.php on line 3001
[08-Mar-2021 14:23:01 UTC] PHP Notice:  Undefined property: DOMElement::$nodeName in /var/www/rocketlabsqa.ovh/htdocs/wp-content/plugins/amp/includes/sanitizers/class-amp-style-sanitizer.php on line 3001` , Is there anything needed from our side?
2- Activating AM and set mode to standard and save, then activate WPR. => no amp query string added to WPR config file in this case, shouldn't it be there?

remyperona:
@Mai-Saad 
1) No, this is not coming from/caused by us
2) No, the standard mode uses only one version for the pages, so there is no specific AMP version, and no need for the query string

Mai-Saad (approved):
Works as expected
[testrail-report-146.pdf](https://github.com/wp-media/wp-rocket/files/6108517/testrail-report-146.pdf)

piotrbak:
@Tabrisrp It seems that we're not removing the query string upon the AMP deactivation... However, if we recreate the config file after AMP deactivation (for example by adding something in the UI that will force the config reset), the query string will be removed.

I don't think anyone will complain about this though. What do you think?

remyperona:
@piotrbak I agree that it's not something blocking. We do have the callback on `deactivate_amp/amp.php`, but the function `is_amp_endpoint()` is still seen as existing in `is_amp_compatible_callback()` somehow.

PR #3615, opened 2021-02-23
Fixes #2041 CDN: prevent incorrect URL on srcset with duplicated relative URLS

### LINKED ISSUE #2041: CDN - Incorrect path rewriting for relative paths in a srcset attribute

remyperona:
## Description

This change prevents relative URLs in `srcset` from being incorrectly rewritten with the CDN URL multiple times, when the same URL is used multiple times in the `srcset` value.

Fixes #2041

## Type of change

Also slightly updated the RegEx after QA, to match `srcset` where the dimensions attribute is missing.

## How Has This Been Tested?

- [x] Tested with the following sample HTML, before and after the change

``
`Absolute&rel-scrset-images`

# Checklist:


iCaspar (approved):

Mai-Saad:
@Tabrisrp @iCaspar Please find notes/questions below:
1- Currently we don't add CDN to the relative path if srcset contains both relative and absolute paths
i.e `Absolute&rel-scrset-images`
Don't we need to handle this? (this is already on the trunk, not related to current PR, not sure if it`s actual case)

2- Shall we see latest updates of preconnect here? as only `<link href='https://fonts.gstatic.com' crossorigin rel='preconnect' />` displayed and no preconnect for CDN

remyperona:
@Mai-Saad 
1) Thank you for this test, I updated the RegEx slightly to make this case work
2) this branch was not up to date, I merge it with our base branch, it should be good now

Mai-Saad (approved):
[testrail-report-145.pdf](https://github.com/wp-media/wp-rocket/files/6107070/testrail-report-145.pdf)
Works as expected

Issue #3478, opened 2021-01-12
Install with composer v2 requirements fail


Alt0car:
I just trying to install last release of wp-rocket 3.8.2 using composer 2.0.8 and php 7.4
```
  Problem 1
    - woocommerce/woocommerce[3.9.0, ..., 3.9.3] require composer/installers 1.7.0 -> satisfiable by composer/installers[v1.7.0].
    - composer/installers v1.7.0 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
    - Root composer.json requires woocommerce/woocommerce ^3.9 -> satisfiable by woocommerce/woocommerce[3.9.0, 3.9.1, 3.9.2, 3.9.3].
```

This problem was `woocommerce/woocommerce 3.9` in dev requirements.

Updated with `woocommerce/woocommerce 4.8.0` solved this issue

remyperona:
Yes this is something we need to update, but it also requires some updates in our test suite to work correctly with newer versions of WooCommerce.

crystinutzaa:
**Reproduce the problem** ✅ 
Issue reproduced on local testing

**Identify the root cause** ✅ 
We need to upgrade to WooCommerce v3.9.2 (the latest version).
However, upgrading will bring a lot of new errors due to a lot of code modification inside WooCommerce.
Errors will popup at Integration test (Unit tests are running just fine)

**Scope a solution** ✅ 
The solution is to upgrade to WooCommerce v4.9.2
This needs to modify the composer.json file to use this: `"woocommerce/woocommerce": "^4.9.2",`

Next, in Integration/bootstrap.php file we need to manually include some WooCommerce files due to the errors which popup when running WithWoo integration tests.
https://github.com/wp-media/wp-rocket/blob/c88e099454ba5b1e2dba3e6cd46936f0525856ed/tests/Integration/bootstrap.php#L55-L60
We need to add a couple of includes and defines:
`define( 'WC_ABSPATH', WP_ROCKET_PLUGIN_ROOT . '/vendor/woocommerce/woocommerce/' );`
```
// Include the main WooCommerce class.
if ( ! class_exists( 'WooCommerce', false ) ) {
	include_once WP_ROCKET_PLUGIN_ROOT . '/vendor/woocommerce/woocommerce/includes/class-woocommerce.php';
[...]

Issue #3230, opened 2020-10-21
Add PHP 8 to our CI matrix


Tabrisrp:
PHP 8 is coming soon, and we need to add it to our testing matrix to ensure everything under test is working correctly with this version.

However, this comes with the issue that the current phpunit version we use are not compatible with it. So we have to come up with a strategy for our testing suite to be able to perform the tests on PHP 8.

Tabrisrp:
### Scope a solution ✅ 
In `composer.json`, we need to load phpunit 9 if the current PHP version is 8.

This can be done by changing the following value in `require-dev`:
- `"php": "^7",` to `"php": "^7  || ^8"`,
- `"phpunit/phpunit": "^7",` to `"phpunit/phpunit": "^7 || ^9",`

We also need to update `/workflows/test_wprocket.yml` to add PHP 8 to the matrix of our github actions

### Estimate the effort ✅
Effort `[S]`

PR #3624, opened 2021-02-26
Closes #3537 Update our dependency container version

### LINKED ISSUE #3537: Update our dependency container to latest stable version
Now that the plugin supports a minimum PHP version >= 7, we can update our dependency container (`league/container`) to the current stable v3 version.

**Benefits**
- The version currently in the plugin is no longer supported since February 2019
-  It also does support PHP 8
- PSR-11 interoperability
- It introduces a "tag" system for each definition registered, which can be very helpful for us at the initialization point, where we can tag all subscribers together, and do a one line call to get them `$container->get('subscribers');`. This would remove the need of the arrays of aliases we have in the `Plugin` class right now. This should improve the performance of the plugin initialization.

**Notable breaking changes**
- Some methods changed name between v2 and v3, notably `withArgument()` became `addArgument()` (remember the WC update falldown?)

remyperona:
## Description

This PR updates our dependency container `league/container` from version 2 to version 3.

- It imports the new version to our codebase using Mozart
- It updates the method name usage from `withArgument()` to `addArgument()` to follow with the breaking change between the 2 versions

Fixes #3537

## Type of change

Yes, because the new tag system currently doesn't work in service providers (See https://github.com/thephpleague/container/issues/223).

So all the subscribers are tagged, but the tags themselves are not used currently.

## How Has This Been Tested?

- Tested manually
- Automated tests in place to catch regression in the instantiation process

# Checklist:


iCaspar (approved):
Ran this with `composer install` after deleting `/vendor` and `composer.lock` and `composer clearcache` -- all worked fine with Composer 1.x
On Composer 2, we are still unable to install due to:
```
 Problem 1
    - woocommerce/woocommerce[3.9.0, ..., 3.9.3] require composer/installers 1.7.0 -> satisfiable by composer/installers[v1.7.0].
    - composer/installers v1.7.0 requires composer-plugin-api ^1.0 -> found composer-plugin-api[2.0.0] but it does not match the constraint.
    - Root composer.json requires woocommerce/woocommerce ^3.9 -> satisfiable by woocommerce/woocommerce[3.9.0, 3.9.1, 3.9.2, 3.9.3].
```
We have a separate issue for that, #3478, which should probably be included in the 3.9 milestone.
Noting also here PR #3230 currently in draft, that updates the CI matrix -- that would be good to include in 3.9, so long as we're introducing breaking changes.

remyperona:
Yes I'm working on #3230, it requires more work than just updating the CI matrix, because WP tests don't run with PHPUnit 9 by default, which is necessary for PHP 8 compatibility. I'm currently testing a polyfill package to make it work 🤞🏼

Mai-Saad (approved):
Tested on single env with PHP 7.4.16 and working as expected.
[testrail-report-156.pdf](https://github.com/wp-media/wp-rocket/files/6223293/testrail-report-156.pdf)

PR #3628, opened 2021-03-01
Fix #3621 move @charset rule should be added before the @import in the minified/combined file

### LINKED ISSUE #3621: @charset rule should be added before the @import in the minified/combined file

**Describe the bug**
The `@charset` rule should be very first in the minified/combined CSS file. When we have `@import` rules, we're moving them to the beginning:
https://developer.mozilla.org/en-US/docs/Web/CSS/@charset

**To Reproduce**
1. Add the `@charset` rule to the top of the CSS file where there's already an `@import` present
2. Enable CSS Minification
3. Check the source of the minified file 
4. Confirm that the `@charset` is moved under the `@import` (after 3.8.6 after the content of the `@import` file)

**Expected behavior**
We should make sure that the `@charset` is at the very top.

**Additional context**
I'm thinking about the case when we combine or import files that already have the `@charset` defined. We'll add it somewhere in the middle of the file.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
## Description

Move `@charset` CSS rule to the top of the combined/minified file.

Fixes #3621 

## Type of change

Not groomed as it's `XS` effort

## How Has This Been Tested?

- [ ] Create CSS file and import more than one file and in the second file add `@charset "utf-8";` then enable minify/combine CSS at WP Rocket settings page and check the minified/combined file you will find this charset at the top of the page

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/CSSTrait.php
+++ b/inc/Engine/Optimization/CSSTrait.php
@@ -40,6 +40,8 @@ trait CSSTrait {
 
 		$content = $this->combine_imports( $content, $target );
 
+		$content = $this->move_charset( $content );
+
 		/**
 		 * Filters the content of a CSS file
 		 *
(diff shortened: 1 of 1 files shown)


crystinutzaa (commented):
Is this tested somewhere? I think it can be easily be covered by Optimization -> CSS -> Combine or Minify fixtures.

PR #3632, opened 2021-03-01
Fixes #3394 PHP notice in Update_Subscriber::disable_auto_updates()

### LINKED ISSUE #3394: PHP notice in Update_Subscriber::disable_auto_updates()
**Describe the bug**

`Notice: Undefined property: stdClass::$plugin in /wp-content/plugins/wp-rocket/inc/classes/subscriber/Plugin/class-updater-subscriber.php on line 271`

**To Reproduce**
The issue likely happen with a plugin that is not passing the expected format for the update offer parameter, using an object missing the `$plugin` property.

We only had one report of this for now.

Nonetheless, we should guard clause the method to prevent the notice, by checking if the property is set before trying to use it.

**Expected behavior**
No PHP notice.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

NataliaDrause:
## Description

Adds guard close isset() to prevent the PHP notice, by checking if the property is set before using it.
Fixes #3394 

## Type of change

No

## How Has This Been Tested?

The issue was reproduced by adding
`unset($item->plugin);`
under this line: https://github.com/wp-media/wp-rocket/blob/f403aa56b4a08e527762a1637cfb8d60695c1e35/inc/classes/subscriber/Plugin/class-updater-subscriber.php#L270

This caused error in debug.log:
_PHP Notice:  Undefined property: stdClass::$plugin in /home4/nataliad/testsite.nataliadrause.com/wp-content/plugins/wp-rocket/inc/classes/subscriber/Plugin/class-updater-subscriber.php on line 272_

After the submitted code modification there are no errors in debug.log

# Checklist:

### DIFF
--- a/inc/classes/subscriber/Plugin/class-updater-subscriber.php
+++ b/inc/classes/subscriber/Plugin/class-updater-subscriber.php
@@ -268,7 +268,7 @@ class Updater_Subscriber implements Subscriber_Interface {
 	 * @return bool|null
 	 */
 	public function disable_auto_updates( $update, $item ) {
-		if ( 'wp-rocket/wp-rocket.php' === $item->plugin ) {
+		if ( isset( $item->plugin ) && ( 'wp-rocket/wp-rocket.php' === $item->plugin ) ) {
 			return false;
 		}
 


remyperona (approved):

piotrbak:
@Tabrisrp Can we add any code to simulate this issue?

edit. **Sorry, I can see it in the description.**

piotrbak (approved):
Tested scenario mentioned in the description. Auto-updates related to WP Rocket are still disabled (`Enable auto-updates` not present), related to other plugins are visible. No log entries:
[testrail-report-149.pdf](https://github.com/wp-media/wp-rocket/files/6110764/testrail-report-149.pdf)

@Tabrisrp Is that enough?

remyperona:
Yes that's good enough 👍🏼 

It's an edge case, and we added an extra guard clause to avoid the PHP notice.

PR #3648, opened 2021-03-08
Closes #2812 Add new filter to rewrite srcset to the CDN

### LINKED ISSUE #2812: Add apply_filter to CDN rewrite_srcset
Our CDN rewrite looks for specific patterns of `srcset`
If developers are using a custom attribute name, we won't detect it for rewriting. We could make the $pattern filterable to address this. 

Related ticket: https://secure.helpscout.net/conversation/1200037694/174367?folderId=377611
>Can you guys please add an apply_filter() to the $pattern? That is the best solution as it makes WPRocket a lot more flexible for all situations. As your code currently is assuming data-srcset/data-lazy-srcset, but due to plugin + theme incompatibilities, themes have to have their own custom attributes for things like this.

vmanthos:
## Description

A new filter `rocket_cdn_srcset_attributes` has been added to allow the addition of custom `srcset` attributes, in order for the URLs to be rewritten to the CDN.

Fixes #2812

## Type of change

No.

## How Has This Been Tested?

- [x] Testedlocally by adding a custom `srcset` attribute, and using the new filter, to check if that's rewritten to the CDN.

# Checklist:

- [x] I have commented on my code, particularly in hard-to-understand areas

### DIFF
--- a/inc/Engine/CDN/CDN.php
+++ b/inc/Engine/CDN/CDN.php
@@ -60,7 +60,7 @@ class CDN {
 	 * @return string
 	 */
 	public function rewrite_srcset( $html ) {
-		$pattern = '#\s+(?:data-lazy-|data-)?srcset\s*=\s*["\']\s*(?<sources>[^"\',\s]+\.[^"\',\s]+(?:\s+\d+[wx])?(?:\s*,\s*[^"\',\s]+\.[^"\',\s]+\s+\d+[wx])*)\s*["\']#i';
+		$pattern = '#\s+(?:' . $this->get_srcset_attributes() . ')?\s*=\s*["\']\s*(?<sources>[^"\',\s]+\.[^"\',\s]+(?:\s+\d+[wx])?(?:\s*,\s*[^"\',\s]+\.[^"\',\s]+\s+\d+[wx])*)\s*["\']#i';
 
 		if ( ! preg_match_all( $pattern, $html, $srcsets, PREG_SET_ORDER ) ) {
 			return $html;
@@ -367,5 +367,30 @@ class CDN {
 		);
 
 		return implode( '|', $files );
+	}
+	
+	/**
+	 * Get srcset attributes to rewrite to the CDN.
+	 *
+	 * @since 3.8.7
+	 *
+	 * @return string A pipe-separated list of srcset attributes.
+	 */
+	private function get_srcset_attributes() {
+		/**
+			* Filter the srcset attributes.
+			*
+			* @since 3.8.7
+			*
+			* @param array $srcset_attributes List of srcset attributes.
+		*/
+		$srcset_attributes = (array) apply_filters( 
+			'rocket_cdn_srcset_attributes',
+			[
+				'data-lazy-srcset',
+				'data-srcset',
+			]
+		);
+		return implode( '|', $srcset_attributes );
 	}
 }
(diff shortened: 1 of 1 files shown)


inc/Engine/CDN/CDN.php
```diff
@@ -60,7 +60,7 @@ function( $matches ) {
 	 * @return string
 	 */
 	public function rewrite_srcset( $html ) {
-		$pattern = '#\s+(?:data-lazy-|data-)?srcset\s*=\s*["\']\s*(?<sources>[^"\',\s]+\.[^"\',\s]+(?:\s+\d+[wx])?(?:\s*,\s*[^"\',\s]+\.[^"\',\s]+\s+\d+[wx])*)\s*["\']#i';
+		$pattern = '#\s+(?:' . $this->get_srcset_attributes() . ')?\s*=\s*["\']\s*(?<sources>[^"\',\s]+\.[^"\',\s]+(?:\s+\d+[wx])?(?:\s*,\s*[^"\',\s]+\.[^"\',\s]+\s+\d+[wx])*)\s*["\']#i';
```
remyperona: The value returned by `get_srcset_attributes()` is the following: `data-lazy-srcset|data-srcset`

Do you see what's missing in the new pattern?

The initial pattern `(?:data-lazy-|data-)?srcset` means that:
- `data-lazy-` or `data-` are optional string values that _can_ (indicated by the `?`) be before the `srcset` string
- `srcset` is required

So this pattern matches `data-lazy-srcset`, `data-srcset` and `srcset`.

Also, the new pattern looks like so: `(?:data-lazy-srcset|data-srcset)?`
It's making the whole capturing group optional, instead of only the "suffixes" values. This could have unintended consequences.

remyperona (changes requested):

inc/Engine/CDN/CDN.php
vmanthos: Thank you for the explanation @Tabrisrp.

I've made changes to the RegEx, and the `get_srcset_attributes()` method based on your comments.

remyperona (approved):

Mai-Saad:
@Tabrisrp @vmanthos Please find my notes/question below:
1- Currently CDN will be written to srcset if we have in the source (srcset or data-srcset or data-lazy-srcset), What about any other cases? like data-alt-srcset, data-base-srcset (existing here https://demos.codetipi.com/zeen/). What if any theme or plugin uses another naming convention i.e data-themeName-srcset or themeName-srcset?
2- Do we need to merge srcset relative URL (3615) fix here?

remyperona:
@Mai-Saad 
1) This is the goal of this PR, to allow other patterns to be added to the detection list with the filter, and be rewritten.
2) they can be merged separately

vmanthos:
@Mai-Saad About question 1, the purpose of this PR is to offer the ability to 3rd-party plugins to use our new filter to add their custom `srcset` attributes.

All the examples you've mentioned can be handled by using code like the following:
```PHP
add_filter( 'rocket_cdn_srcset_attributes', function ( $attributes ) {
   $attributes[] = 'data-alt-';
   $attributes[] = 'data-base-';
   return $attributes;
});

Mai-Saad (approved):
Works as expected.
- existing problems of duplicate CNAME or missed incase of absolute+relative URL will be fixed with 3615

Issue #3635, opened 2021-03-02
Remove Unused CSS


GeekPress:
## What are we proposing to do? 
To build a new feature within WP Rocket that will remove all the unused CSS from each page of the website. 

We will do this by building a CSS tree-shaker from the ground up, and host it on our server. WP Rocket will then interact with this service via its API. This is similar to what we do for our CPCSS generator. 

We will do this for each page of the website and embed the CSS right into the HTML of the page. 

([full doc](https://docs.google.com/document/d/15hFzxmi5lBMJ7qw0EaO-td7p0IhkDSuNXO8fPoMwlxw/edit?usp=sharing))

webtrainingwheels:
@Tabrisrp The doc link for this is going to be: https://docs.wp-rocket.me/article/1529-remove-unused-css
ID: 6076083ff8c0ef2d98df1f97

PR #3666, opened 2021-03-11
Collecting resources and call warmup

engahmeds3ed:
sub-task of #3635


inc/Engine/Optimization/RUCSS/ServiceProvider.php
```diff
@@ -36,10 +41,34 @@ public function register() {
...
+		$this->getContainer()->add( 'rucss_resources_query', 'WP_Rocket\Engine\Optimization\RUCSS\Database\Queries\ResourcesQuery' );
 		$this->getContainer()->add( 'rucss_database', 'WP_Rocket\Engine\Optimization\RUCSS\Admin\Database' )
 			->withArgument( $this->getContainer()->get( 'rucss_resources_table' ) );
 		$this->getContainer()->share( 'rucss_admin_subscriber', 'WP_Rocket\Engine\Optimization\RUCSS\Admin\Subscriber' )
 			->withArgument( $this->getContainer()->get( 'rucss_settings' ) )
 			->withArgument( $this->getContainer()->get( 'rucss_database' ) );
+
+		$this->getContainer()->add( 'local_cache', '\WP_Rocket\Engine\Optimization\AssetsLocalCache' )
+			->withArgument( rocket_get_constant( 'WP_ROCKET_MINIFY_CACHE_PATH' ) )
+			->withArgument( rocket_direct_filesystem() );
```
remyperona, line 52: This is already loaded elsewhere, why the need to re-add it here?

inc/Engine/Optimization/RUCSS/Warmup/APIClient.php
```diff
@@ -0,0 +1,56 @@
...
+			[
+				'body' => [
+					'resources' => [
+						$atts,
+					],
+				],
+			]
+		);
+
+		Logger::debug( 'asa_api_client', (array) $response );
```
remyperona: what does the `asa_` prefix correspond to?

inc/Engine/Optimization/RUCSS/Warmup/APIClient.php
```diff
@@ -0,0 +1,56 @@
...
+
+		if ( is_wp_error( $response ) ) {
+			return false;
+		}
+
+		if ( 200 !== wp_remote_retrieve_response_code( $response ) ) {
+			return false;
+		}
+
+		return true;
```
remyperona: This could be simplified to `return 200 === wp_remote_retrieve_response_code( $response );`

- `wp_remote_retrieve_response_code()` checks for `is_wp_error()` internally [doc](https://developer.wordpress.org/reference/functions/wp_remote_retrieve_response_code/)
- the return of this comparison is a boolean, so we can use it directly

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
```diff
@@ -0,0 +1,169 @@
...
+			];
+		}
+	}
+
+	/**
+	 * Get url file contents.
+	 *
+	 * @param string $url File url.
+	 *
+	 * @return false|string
```
remyperona: We can return an empty string instead of false, and that way we have only a string return instead of a mixed one

inc/Engine/Optimization/RUCSS/Warmup/WarmupProcess.php
```diff
@@ -0,0 +1,105 @@
...
+	 * @param int $resource_id Resource DB id.
+	 *
+	 * @return bool
+	 */
+	protected function task( $resource_id ) {
+		if ( ! is_int( $resource_id ) ) {
+			return false;
+		}
+
+		$resource_row = $this->resources_query->get_item( $resource_id );
```
remyperona: We already have the information we need from the previous background process, why do we do a new request in the database for them?

Can't we re-use directly the initial data, by passing everything to the `$send_to_warmup` array?

remyperona (changes requested):

inc/Engine/Optimization/RUCSS/ServiceProvider.php
engahmeds3ed: it's called using `new` not via the container here:

https://github.com/wp-media/wp-rocket/blob/f92eb9edbb0dd8710061d1643da9b5f6745d56b4/inc/Engine/Optimization/Minify/JS/Subscriber.php#L47

https://github.com/wp-media/wp-rocket/blob/f52d197c1da902b1c8b065340f7bff95853566a5/inc/Engine/Optimization/Minify/CSS/Subscriber.php#L45

this is why I created it here, do u think we need to refactor this code to use the container

inc/Engine/Optimization/RUCSS/Warmup/APIClient.php
engahmeds3ed: sorry it was just for debugging, removed.

inc/Engine/Optimization/RUCSS/Warmup/WarmupProcess.php
engahmeds3ed: really we don't have the DB row in all cases as this method 

https://github.com/wp-media/wp-rocket/blob/d7fdf63f9e5249dfe0ae6305b4a7b2b854129368/inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php#L65

returns the row ID only so in this case we would need to make another DB query to get this row.

also this update method

https://github.com/wp-media/wp-rocket/blob/d7fdf63f9e5249dfe0ae6305b4a7b2b854129368/inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php#L97

returns only ID, so if U used `$db_row` it wouldn't be updated.

for those two reasons I took the decision to pass only the IDs and there I can query to get them

inc/Engine/Optimization/RUCSS/ServiceProvider.php
remyperona: Not for now, it's out of scope of what we're doing here.

inc/Engine/Optimization/RUCSS/Warmup/APIClient.php
```diff
@@ -0,0 +1,46 @@
+<?php
+declare(strict_types=1);
+
+namespace WP_Rocket\Engine\Optimization\RUCSS\Warmup;
+
+use WP_Rocket\Logger\Logger;
```
remyperona: No need to import it if we don't use it

inc/Engine/Optimization/RUCSS/Warmup/WarmupProcess.php
remyperona: This is the flow I currently see:
- In `ResourceFetcher`, the `resources` property array contains the URL, content and type
- They are passed to the `ResourceFetcherProcess` to be saved in the DB, and return the ID
- In `WarmupProcess`, we get the same 3 values from the DB, by fetching them from the DB by the ID previously created

Do you see the issue here? At the start we have the data we need for the end, but instead of passing that data all the way, we add an additional step of reading again from the DB to get the same data.

inc/Engine/Optimization/RUCSS/ServiceProvider.php
iCaspar: I vote for reusing everywhere from container.

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
```diff
@@ -0,0 +1,169 @@
...
+				continue;
+			}
+
+			$this->resources[] = [
+				'url'     => rocket_add_url_protocol( $script['url'] ),
+				'content' => $contents,
+				'type'    => 'js',
+			];
+		}
+	}
```
iCaspar: These 2 methods are essentially the same. We can set class constants for the regexes and refactor to a single `find_resources()` method to eliminate most of the duplication here.

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
```diff
@@ -0,0 +1,169 @@
...
+	 *
+	 * @param string $html Page HTML.
+	 */
+	public function handle( $html ) {
+		// Grab resources from page HTML.
+		$this->find_styles( $html );
+		$this->find_scripts( $html );
+
+		// Send resources to the background process to be saved into DB.
+		$this->process->push_to_queue( $this->resources );
```
remyperona: In this case, why was the choice of pushing the full array to the background process made, instead of looping over the array to push every item of the array individually?

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php
```diff
@@ -0,0 +1,114 @@
...
+			$this->resources_query->update_item(
+				$db_row->id,
+				[
+					'content'          => $resource['content'],
+					'hash'             => md5( $resource['content'] ),
+					'resend_to_warmup' => 1,
+				]
+			);
+
+			$send_to_warmup[] = $db_row->id;
```
remyperona: Related to my other comment: why only send the ID to the warmup here, when we already have the content, url and type available?

remyperona (changes requested):

inc/Engine/Optimization/RUCSS/Warmup/WarmupProcess.php
engahmeds3ed: Got u, so u need me to pass the data directly to the warmup process, good idea!
Will adjust that now.

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php
```diff
@@ -0,0 +1,114 @@
...
+	 * Do the task for each page resources.
+	 *
+	 * @param array $resources Array of collected resources on the page.
+	 *
+	 * @return false
+	 */
+	protected function task( $resources ) {
+		if ( ! is_array( $resources ) ) {
+			return false;
+		}
```
iCaspar: In php 7 we can specify array type in signature and get rid of the type checking.
Better yet, since the parent class seems to require an iterable for `task()` (@see the description in the docblock), it might be better to change the signature to `abstract protected function task( iterable $item ) there rather than a mixed/unspecified.

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
engahmeds3ed: Ok, I was between two choices:-
1. Send array of resources to be sent to saas warmup bulky =>will take some time and may timeout
2. Send item per item to warmup process so they can be handled in parallel

what do you think ?

remyperona: Since we're sending to the background process, I think it makes more sense to go with option 2, as it's the whole idea of the background process: handle multiple, small items one by one instead of a big one

remyperona: But it also means that you will have to handle passing data to the next BG process differently, as in the `task()` method you won't have everything available, only the current item.

inc/Engine/Optimization/RUCSS/Warmup/WarmupProcess.php
remyperona: It does make more sense to do it that way, don't you think? It avoids an extra request to the DB.

inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
```diff
@@ -0,0 +1,106 @@
...
+		$this->warmup_process->save()->dispatch();
+	}
+
+	/**
+	 * If it's allowed to warmup resources.
+	 *
+	 * @return bool
+	 */
+	private function is_allowed() {
+		if ( rocket_get_constant( 'DONOTROCKETOPTIMIZE' ) ) {
```
iCaspar, line 72: should we also be checking for the ?nowprocket query string indicator here?

iCaspar (changes requested):

inc/Engine/Optimization/RUCSS/Warmup/WarmupProcess.php
engahmeds3ed: yes but when will we do the following DB update?
https://github.com/wp-media/wp-rocket/blob/504c70adbea734d97e7f4c73c678448b103d54fa/inc/Engine/Optimization/RUCSS/Warmup/WarmupProcess.php#L93-L98

really I don't think that we need this column `resend_to_warmup` anymore as we already pass the resource itself not making a query to get all resources need to be sent to warmup.

Or will we make cleanup cron for those not warmedup resources for any reason? => I don't think so.

what do u think ?

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php
iCaspar: Nevermind this.... I'm wrong. The parent method is passing a single item in an iterable. 🤦

inc/Engine/Optimization/RUCSS/Warmup/WarmupProcess.php
iCaspar: I was wondering that myself earlier. I think that unless we need it for some other reason, (will we ever need to preserve that info in case of a retry?) we can probably remove it now.

inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
remyperona: We can add it yes

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
```diff
@@ -0,0 +1,175 @@
...
+		$this->process->save()->dispatch();
+	}
+
+	/**
+	 * Find link styles on the page HTML.
+	 *
+	 * @param string $html Page HTML.
+	 */
+	private function find_styles( $html ) {
+		$links = $this->find( '<link(?:[^>]+[\s"\'])?href\s*=\s*[\'"]\s*(?<url>[^\'"\s]+)\s*?[\'"](?:[^>]+)?\/?>', $html );
```
iCaspar: This regex works for `<link rel="stylesheet" href="assets/css/styles.min.css" media="all">`,  
but not for `<link href="assets/css/styles.min.css" media="all" rel="stylesheet">` (doesn't seem to catch links that do not have anything before the `href`). I think we need to account for any order of attributes here.

crystinutzaa: I used this regex to catch all links and it works in both cases:
`<link\s+([^>]+[\s"\'])?href\s*=\s*[\'"]\s*?(?<url>[^\'"]+\.css(?:\?[^\'"]*)?)\s*?[\'"]([^>]+)?\/?>`

https://regex101.com/r/cwe74M/1/

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
```diff
@@ -0,0 +1,175 @@
...
+
+		if ( empty( $links ) ) {
+			return;
+		}
+
+		foreach ( $links as $link ) {
+			if ( ! (bool) preg_match( '/rel=[\'"]stylesheet[\'"]/is', $link[0] ) ) {
```
crystinutzaa, line 103: Is there a reason to bailout in case there is no `rel=stylesheet` on the link?

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php
```diff
@@ -0,0 +1,124 @@
...
+
+		if ( empty( $db_row ) ) {
+			// Create this new row in DB.
+			$resource_id = $this->resources_query->add_item(
+				[
+					'url'           => $resource['url'],
+					'type'          => $resource['type'],
+					'content'       => $resource['content'],
+					'hash'          => md5( $resource['content'] ),
+					'last_accessed' => gmdate( 'Y-m-d\TH:i:s\Z' ),
```
crystinutzaa: In Phase 3 I am using:
https://github.com/wp-media/wp-rocket/blob/65ad8426541fbd739133323aab4e434be4c0332b/inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php#L109
For all datetime DB values. 
We shall uniform all the code to use a single datetime format.

https://developer.wordpress.org/reference/functions/current_time/

What do you think?

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php
```diff
@@ -0,0 +1,124 @@
...
+
+
+	/**
+	 * Do the task for each page resources.
+	 *
+	 * @param array $resource Resource array consists of url, content and type.
+	 *
+	 * @return false
+	 */
+	protected function task( $resource ) {
```
crystinutzaa, line 61: Could this function be simplied?
Like:
1. Resource -> add or update in DB.
2. Check if content is changed -> send / resend to Warmup

inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	/**
+	 * Return an array of events that this subscriber listens to.
+	 *
+	 * @return array
+	 */
+	public static function get_subscribed_events() : array {
+		return [
+			'rocket_buffer' => 'collect_resources',
```
crystinutzaa, line 43: We need to discuss the priority for `collect_resources`.
Collect Resources should happen before TreeShake.
On `rocket_buffer` we have a couple of Optimizations which do not need to be sent to TreeShake ( SAAS expects the HTML without minified CSS, JS ). 
In Phase 3 I've sent TreeShake on priority 1, however, I think collect_resources should be 1 & TreeShake should be 2.

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
iCaspar: In the first regex above, we've collected ALL the link elements (theoretically). Now we only want the ones that are stylesheets here (as opposed to, say, links for prefetches, etc.). That said, it may be more efficient to eliminate them in the first pass with the previous regex rather than re-loop and check them here.

crystinutzaa: Ah, you are right @iCaspar. But I don't think we have an easy way to identify it in the REGEX due to all potential order of attributes. I need to update the code in Phase 3 too :)

iCaspar: Indeed, the regex to do it in 1 go would be a big hairy monster 👹

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
iCaspar: @crystinutzaa, this regex may be a solution to the need for a regex below... the check for a url ending with the `.css` extension may eliminate all the other unwanted links. I'm trying to think now of any other circumstances where a link would point to a .css file that we don't want to include. Any ideas? Would `.css` be found elsewhere in the URL path, perhaps, that might point to something that we don't want to collect?

engahmeds3ed: I adjusted the old regex a bit to be like
```
<link\s+(?:[^>]+[\s"\'])?href\s*=\s*[\'"]\s*(?<url>[^\'"\s]+)\s*?[\'"](?:[^>]+)?\/?>
```
so it can get both link schemas.

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
engahmeds3ed: so this is resolved :D

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php
engahmeds3ed: We agreed on using modified and last_accessed only and use the function `current_time( 'mysql', true )` to use current GMT time.

inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
engahmeds3ed: @crystinutzaa 
Good question, really I don't like using 1 and 2 we can start from 10 so other plugins (or maybe we) can hook it before both of them.
I think we can make warmup at first thing (priority 10) and let treeshaker comes at the last thing of optimizations so it will give some time to warmup to finish his work, 

what do u think?

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php
engahmeds3ed: Adjusted the code and put it inside the query class

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
engahmeds3ed: @iCaspar they are not exactly the same as on css we are checking the rel also so when I tried to refactor it I found that the code will be hard to read as I will check the type two times, one for regex and the other for checking rel,

what do u think ?

inc/Engine/Optimization/RUCSS/Database/Row/ResourceRow.php
```diff
@@ -0,0 +1,27 @@
...
+	 */
+	public function __construct( $item ) {
+		parent::__construct( $item );
+
+		$this->id            = (int) $this->id;
+		$this->url           = (string) $this->url;
+		$this->type          = (string) $this->type;
+		$this->content       = (string) $this->content;
+		$this->hash          = (string) $this->hash;
+		$this->modified      = (string) $this->modified;
```
crystinutzaa: Why string for modified?

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
```diff
@@ -0,0 +1,196 @@
...
+	/**
+	 * Resource constructor.
+	 *
+	 * @param AssetsLocalCache       $local_cache Local cache instance.
+	 * @param ResourceFetcherProcess $process     Resource fetcher process instance.
+	 */
+	public function __construct( AssetsLocalCache $local_cache, ResourceFetcherProcess $process ) {
+		$this->local_cache = $local_cache;
+		$this->process     = $process;
+		parent::__construct();
```
crystinutzaa: Probably move parent::__construct(); before the param initialization and add a space after it.
```
parent::__construct();

$this->local_cache = $local_cache;
$this->process     = $process;
```

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php
crystinutzaa: Looks better and cleaner now 👍 🎖️

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
iCaspar: I see. Let me play with it. 😝
(But don't hold up the merge for this. These are private methods, so they won't affect the testing if they get refactored after.)

inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
iCaspar: `collect_resources` should happen very early. Maybe priority 5 to allow for other things to get in first, but earlier than 10, because we don't want other things with default 10 to fire before.
I think we can't do treeshake last, b/c of @crystinutzaa's observation that some optimizations will happen after we have the shaken CSS (minify, specifically, maybe others). So it should be late, but before any of those.

iCaspar: I'm favoring a bigger priority gap between collect and optimize here, because it allows a few more ms for the SaaS warmup to collect maybe a few more resources before the call to treeshake hits the SaaS. It _may_ provide a slightly better outcome on those first few calls, which would be better early result for the customer and fewer re-queued resources for the next time warmup runs. 🤔

crystinutzaa: Hem....TreeShake need to kick in before any HTML optimizations. So it needs to be run before applying CDN, Minify CSS, Minify JS , lazyload. All of these start from priority 10 (some are on 14, 16,17). 
So, I assume Collect Resources will go on 10 and TreeShake on 11. I don't think we have other option.

iCaspar: Collect on 4, shake on 9?

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
crystinutzaa: @iCaspar I was thinking about `<link rel="preload" href="path/to/file.css"/>` 
This still needs to be bailout based on rel attribute

inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
crystinutzaa: Yeah. I think collect on 4 and Shake on 11. However, imagine that nothing else is running in between if we will set it on 4 and 9. 😅 
On priority 10, we run `cache_busting_facebook_tracking` , `cache_busting_google_tracking`, 
On Priority 1, we run: `extract_ie_conditionals`

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
iCaspar: @engahmeds3ed's revised regex is working for me. Going to mark this resolved. Agreed that we do need the second regex below.

PR #3684, opened 2021-03-17
Fix 3219 - Add preconnect link to google fonts only when google fonts are found

### LINKED ISSUE #3219: Google Font Preconnect is added even when no requests are made
The Google font preconnect is added in the header even when there are no Google fonts that are requested. For example, websites that host fonts locally don't need this to be added in the header adding extra load times. How to remove it?

iCaspar:
## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #3219

## Type of change

It is mostly as groomed. I've also included setting the flag in the concrete optimizers `optimize()` methods for cases where google fonts _are_ present as well as when they are not, rather than assuming a default flag status in the Combine classes. (The `AbstractGFOptimization::has_google_fonts()` method will return true if for any reason the flag fails to be set, so that preconnect link will not be inadvertently removed in any edge cases.)

## How Has This Been Tested?

- [X] Integration testcase has been added in the V2V1 fixture.
- [X] Steps to check: With Combine Google Fonts setting active ---
      1. Activate a theme -- or set a page where -- google fonts are present on the page.
      2. Clear rocket cache and load page. Observe that preconnect link is added to inside the page `<head>` html.
      3. Activate a theme -- or set a page where -- google fonts are NOT present on the page.
      4. Clear rocket cache and load page. Observe that there is NOT a preconnect link added inside the page `<head>` html.

# Checklist:

- [X] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/inc/Engine/Optimization/GoogleFonts/AbstractGFOptimization.php
+++ b/inc/Engine/Optimization/GoogleFonts/AbstractGFOptimization.php
@@ -35,7 +35,7 @@ abstract class AbstractGFOptimization extends AbstractOptimization {
 	 *
 	 * @var bool
 	 */
-	protected $has_google_fonts;
+	protected $has_google_fonts = true;
 
 
 	/**
@@ -46,7 +46,7 @@ abstract class AbstractGFOptimization extends AbstractOptimization {
 	 * @return bool Will default to true when extending classes have not set via the optimize() method.
 	 */
 	public function has_google_fonts() {
-		return isset( $this->has_google_fonts ) ? $this->has_google_fonts : true;
+		return $this->has_google_fonts;
 	}
 
 	/**
--- a/inc/Engine/Optimization/GoogleFonts/AbstractGFOptimization.php
+++ b/inc/Engine/Optimization/GoogleFonts/AbstractGFOptimization.php
@@ -31,9 +31,25 @@ abstract class AbstractGFOptimization extends AbstractOptimization {
 	/**
 	 * Flag for whether google fonts have been detected (Default: true)
 	 *
+	 * @since 3.8.8
+	 *
 	 * @var bool
 	 */
 	protected $has_google_fonts = true;
+
+
+	/**
+	 * Check whether the optimizer has found google fonts on the page.
+	 *
+	 * Note: Will always return true if the optimize() method has yet been called.
+	 *
+	 * @since 3.8.8
+	 *
+	 * @return bool
+	 */
+	public function has_google_fonts() {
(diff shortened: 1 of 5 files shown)


inc/Engine/Optimization/GoogleFonts/AbstractGFOptimization.php
```diff
@@ -28,6 +28,27 @@ abstract class AbstractGFOptimization extends AbstractOptimization {
...
 	];
 
+	/**
+	 * Flag for whether google fonts have been detected (Default: true)
+	 *
+	 * @since 3.8.8
+	 *
+	 * @var bool
+	 */
+	protected $has_google_fonts;
```
remyperona:
```suggestion
	protected $has_google_fonts = true;
```

We can set the default value here, and it will also simplify the related method. Else we declare it like this, the default value is `null`, not a boolean (it would trigger errors in PHPStan for example)

inc/Engine/Optimization/GoogleFonts/AbstractGFOptimization.php
```diff
@@ -28,6 +28,27 @@ abstract class AbstractGFOptimization extends AbstractOptimization {
...
+
+	/**
+	 * Check whether the optimizer has found google fonts on the page.
+	 *
+	 * @since 3.8.8
+	 *
+	 * @return bool Will default to true when extending classes have not set via the optimize() method.
+	 */
+	public function has_google_fonts() {
+		return isset( $this->has_google_fonts ) ? $this->has_google_fonts : true;
```
remyperona:
```suggestion
		return $this->has_google_fonts;
```

remyperona (changes requested):

inc/Engine/Optimization/GoogleFonts/AbstractGFOptimization.php
iCaspar: Indeed, this was how I had it initially. I changed it during the writing of the tests, but it should be reverted.

remyperona (approved):

iCaspar:
@wp-media/qa Ready for QA. Acceptance criteria are basically in the "Steps to Check" outlined in the description above.

Mai-Saad (approved):
Works as expected.

Notes:
- Deactivate of disable google font helper plugin shall auto enable GF optimization
- If preconnect tag was in non cached version while GF not used in HTML then with WPR we will remove the preconnect tag

PR #3691, opened 2021-03-19
Refactor RUCSS APIClient

crystinutzaa:
sub-task of #3635 

Refactor APIClient and created abstract class + for each side: TreeShake & Warmup, extended the AbstractAPIClient and implemented each required methods.

### DIFF
--- a/inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
+++ b/inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
@@ -1,0 +1,48 @@
+<?php
+declare(strict_types=1);
+
+namespace WP_Rocket\Engine\Optimization\RUCSS;
+
+abstract class AbstractAPIClient {
+	/**
+	 * API URL.
+	 */
+	const API_URL = 'https://central-saas.wp-rocket.me/';
+
+	/**
+	 * SAAS main API path.
+	 */
+	const TREESHAKE_PATH = 'api';
+
+	/**
+	 * API Warmup path.
+	 */
+	const WARMUP_PATH = 'warmup';
+
+	/**
+	 * Handle Saas request error.
+	 *
+	 * @param  array|WP_Error $request WP Remote request.
+	 *
+	 * @return array
+	 */
+	protected function handle_request_error( $request ) {
+		if ( is_wp_error( $request ) ) {
+			return [
+				'code'    => 400,
+				'message' => $request->get_error_message(),
+			];
+		}
+
+		$code = wp_remote_retrieve_response_code( $request );
+
+		if ( 200 !== $code ) {
+			return [
+				'code'    => $code,
+				'message' => __( 'Remove Unused CSS is not available at the moment. Please retry later', 'rocket' ),
+			];
+		}
+
+		return false;
+	}
+}
--- a/inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
+++ b/inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
@@ -3,6 +3,8 @@ declare(strict_types=1);
 
 namespace WP_Rocket\Engine\Optimization\RUCSS;
 
+use WP_Error;
+
 abstract class AbstractAPIClient {
 	/**
 	 * API URL.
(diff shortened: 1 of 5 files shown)


inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
```diff
@@ -0,0 +1,48 @@
...
+
+	/**
+	 * SAAS main API path.
+	 */
+	const TREESHAKE_PATH = 'api';
+
+	/**
+	 * API Warmup path.
+	 */
+	const WARMUP_PATH = 'warmup';
```
engahmeds3ed: @crystinutzaa 

what do you think about defining those constants at their APIClients so at treeshaker APIClient we can define this `TREESHAKE_PATH` and on warmup APIClient we can define this `WARMUP_PATH`?

inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
```diff
@@ -0,0 +1,48 @@
...
+
+	/**
+	 * Handle Saas request error.
+	 *
+	 * @param  array|WP_Error $request WP Remote request.
+	 *
+	 * @return array
+	 */
+	protected function handle_request_error( $request ) {
```
engahmeds3ed: so this method will be called only on treeshaker, right?

I think we need to unify both APIClients to be able to use this method generically, what do u think?

engahmeds3ed (changes requested):

inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
crystinutzaa: yeah...each path can be defined in each sub-class.

inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
crystinutzaa: Yeah. In TreeShaker I return both the code and the error message why it could not tree shake the CSS.
Not sure what Warmup needs, but a message should be good for logging and debuging.

inc/Engine/Optimization/RUCSS/ServiceProvider.php
```diff
@@ -51,7 +51,7 @@ public function register() {
 			->withArgument( $this->getContainer()->get( 'rucss_usedcss_table' ) );
 
 		$this->getContainer()->add( 'rucss_used_css_query', 'WP_Rocket\Engine\Optimization\RUCSS\Database\Query\UsedCSS' );
-		$this->getContainer()->add( 'rucss_api_client', 'WP_Rocket\Engine\Optimization\RUCSS\APIClient' );
+		$this->getContainer()->add( 'rucss_frontend_api_client', 'WP_Rocket\Engine\Optimization\RUCSS\Frontend\APIClient' );
 		$this->getContainer()->add( 'rucss_used_css_controller', 'WP_Rocket\Engine\Optimization\RUCSS\Controller\UsedCSS' )
 			->withArgument( $this->getContainer()->get( 'rucss_used_css_query' ) );
```
iCaspar, line 57: Question @Tabrisrp , will these need to be changed at all to conform to the new container (I'm thinking of the new tagging, etc., that was needed in the other recent container update PRs)?

remyperona: yes they will

inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
```diff
@@ -0,0 +1,55 @@
...
+	 *
+	 * @return array|WP_Error $request WP Remote request.
+	 */
+	protected function handle_post( string $path, array $args ) {
+		$request = wp_remote_post(
+			self::API_URL . $path,
+			$args
+		);
+
+		return $request;
```
remyperona:
```suggestion
		return wp_remote_post(
			self::API_URL . $path,
			$args
		);
```

inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
```diff
@@ -3,53 +3,67 @@
...
+
 	/**
 	 * Handle remote POST.
 	 *
-	 * @param string $path API path.
-	 * @param array  $args Array with options sent to Saas API.
+	 * @param array $args Array with options sent to Saas API.
 	 *
-	 * @return array|WP_Error $request WP Remote request.
+	 * @return bool $request WP Remote request.
```
iCaspar:
```suggestion
	 * @return bool $request True for successful WP Remote request; false when error.
```

iCaspar (changes requested):
Just a quick suggestion there for the docblock to clarify what is being returned. Could also add the bool return type to the signature on the next line as well. :)

iCaspar (approved):

PR #3706, opened 2021-03-23
Closes #3667 Don't add preload fonts links when CPCSS is enabled & remove font-face from CPCSS

### LINKED ISSUE #3667: Don't preload fonts when CPCSS is injected on the current page & remove font-face rules from CPCSS
When CPCSS is present on the current page, a combo will lead to a better LCP:
- Stop preloading fonts configured into our « Preload Fonts » option.
- Remove font-face rules from CPCSS

In our benchmarks, it can improve the LCP from 10 to 20%.

The CPCSS API is ready to be able to remove font-face rules from the CPCSS:
https://github.com/wp-media/criticalpath-css-generator/blob/master/app/controllers/register.js#L22  

We will need to send a new parameter in the CPCSS API config: https://github.com/wp-media/wp-rocket/blob/c9b302823c14481ccf1a6f56919b7fc9b446dd73/inc/Engine/CriticalPath/ProcessorService.php#L84

`'nofontface' => true,`

If this change causes a problem on a website, this could be solved by using this filter and set the `nofontface` parameter to `false`:
https://github.com/wp-media/wp-rocket/blob/develop/inc/Engine/CriticalPath/APIClient.php#L38

remyperona:
## Description

- Prevent adding the preload fonts links when the CPCSS is available for a page
- Add new parameter to CPCSS API request to prevent font-face from being added to the generated CPCSS

Fixes #3667

## Type of change

There is a couple changes:
- Simplified the method for maybe disabling the preload fonts
- I didn't add a new filter for the new `nofontface` parameter, because we already have a filter for CPCSS API request `rocket_cpcss_job_request` that can be used to modify the parameters sent

## How Has This Been Tested?

- [x] New unit & integration tests

# Checklist:


inc/Engine/CriticalPath/CriticalCSSSubscriber.php
```diff
@@ -688,6 +689,25 @@ public function maybe_regenerate_cpcss() {
...
+	public function maybe_disable_preload_fonts() : bool {
+		if ( ! $this->should_async_css() ) {
+			return false;
+		}
+
+		return (
+			! empty( $this->critical_css->get_current_page_critical_css() )
+			||
+			! empty( $this->options->get( 'critical_css', '' ) )
+		);
```
engahmeds3ed:
```suggestion
		return (
			$this->should_async_css()
			&&
			(
				! empty( $this->options->get( 'critical_css', '' ) )
				||
				! empty( $this->critical_css->get_current_page_critical_css() )
			)
		);
```

what about this small change?
**Why?**
- Check the database option before checking the filesystem.
- **(Optional)** merge `should_async_css` into the check. (maybe you did such line to return earlier or to make it more readable, just commented here to learn from u)

tests/Integration/inc/Engine/CriticalPath/CriticalCSSSubscriber/maybeDisablePreloadFonts.php
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    public function async_css() {
+        return $this->async_css;
+    }
+
+    public function critical_css() {
+        return $this->critical_css;
+    }
+}
```
engahmeds3ed: Add new empty line at the end of this file

engahmeds3ed (approved):
Very Minor changes, won't affect the testing process, 

@wp-media/qa Ready for QA

inc/Engine/CriticalPath/CriticalCSSSubscriber.php
remyperona: Good idea to check for the database first, it's likely a bit faster.

In my opinion, keeping the first check on its own makes the code more readable, when you start to accumulate multiple conditions and operators, it becomes hard to see the path.

engahmeds3ed: Got it, thanks.

piotrbak (approved):
Looks good:
[testrail-report-155.pdf](https://github.com/wp-media/wp-rocket/files/6214333/testrail-report-155.pdf)

PR #3711, opened 2021-03-24
Fixes #3449 rocket_need_api_key() - PHP error during license validation

### LINKED ISSUE #3449: rocket_need_api_key() - PHP Fatal error during license validation

**Describe the bug**

A PHP fatal error is thrown when the `$errors` is not an array/countable type, e.g. it's a `false` here:
https://github.com/wp-media/wp-rocket/blob/3f4de3116cbd9c97f1ac5bff0002f31587353d1d/inc/functions/admin.php#L23

Here is the log:
```
PHP Fatal error:  Uncaught TypeError: count(): Argument #1 ($var) must be of type Countable|array, bool given in /home/example.com/wp-content/plugins/wp-rocket/inc/functions/admin.php:23
Stack trace:
#0 /home/example.com/wp-includes/class-wp-hook.php(287): rocket_need_api_key('')
#1 /home/example.com/wp-includes/class-wp-hook.php(311): WP_Hook->apply_filters(NULL, Array)
#2 /home/example.com/wp-includes/plugin.php(484): WP_Hook->do_action(Array)
#3 /home/example.com/wp-admin/admin-header.php(292): do_action('admin_notices')
#4 /home/example.com/wp-admin/admin.php(239): require_once('/home/gunacom/p...')
#5 /home/example.com/wp-admin/options-general.php(10): require_once('/home/gunacom/p...')
#6 {main}
  thrown in /home/example.com/wp-content/plugins/wp-rocket/inc/functions/admin.php on line 23
```

**To Reproduce**
1. Enable WordPress's debugging.
2. Go to `/inc/functions/admin.php` and hardcode `$errors` to `false` in line 11.
3. .Visit WP Rocket's dashboard.
4. Check the `debug.log`.

**Expected behavior**

No fatal error should take place even when the `get_transient()` returns a boolean.

**Additional context**

**Related ticket:** https://secure.helpscout.net/conversation/1367704657/223307/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
## Description

if the `consumer_key` exists but its length is less/more than 8 characters (and this key is the same on our server), we fail without showing any error message so the transient which has this error isn't exist and when trying to get it it gives boolean false value so when trying to use `count` method it fails with warnings on some environments and fatal errors on others.

Fixes #3449 

## Type of change

Please describe in this section any change to the solution, and why.

## How Has This Been Tested?

- [x] I found the case on one of our customer's sites and this change solved the issue for them
- [x] I changed this number `8` on the following line to be `7`
https://github.com/wp-media/wp-rocket/blob/6ecc8b2df076136422dc5b0a89597eef4c978f38/inc/functions/options.php#L462

# Checklist:

### DIFF
--- a/inc/functions/admin.php
+++ b/inc/functions/admin.php
@@ -10,10 +10,8 @@ function rocket_need_api_key() {
 	$message = '';
 	$errors  = (array) get_transient( 'rocket_check_key_errors' );
 
-	if ( false !== $errors ) {
-		foreach ( $errors as $error ) {
-			$message .= '<p>' . $error . '</p>';
-		}
+	foreach ( $errors as $error ) {
+		$message .= '<p>' . $error . '</p>';
 	}
 
 	?>
--- a/inc/functions/admin.php
+++ b/inc/functions/admin.php
@@ -8,7 +8,7 @@ defined( 'ABSPATH' ) || exit;
  */
 function rocket_need_api_key() {
 	$message = '';
-	$errors  = get_transient( 'rocket_check_key_errors' );
+	$errors  = (array) get_transient( 'rocket_check_key_errors' );
 
 	if ( false !== $errors ) {
 		foreach ( $errors as $error ) {
--- a/inc/functions/options.php
+++ b/inc/functions/options.php
@@ -462,15 +462,20 @@ function rocket_valid_key() {
 	$valid_details = 8 === strlen( get_rocket_option( 'consumer_key' ) ) && hash_equals( $rocket_secret_key, hash( 'crc32', get_rocket_option( 'consumer_email' ) ) );
 
 	if ( ! $valid_details ) {
-
 		set_transient(
 			'rocket_check_key_errors',
-			// Translators: %1$s = opening link tag, %2$s = closing link tag.
-			[ __( 'Not valid License details.', 'rocket' ) . ' <br>' . sprintf( __( 'To resolve, please %1$scontact support%2$s.', 'rocket' ), '<a href="https://wp-rocket.me/support/" rel="noopener noreferrer" target=_"blank">', '</a>' ) ]
+			[
+				__( 'The provided license data are not valid.', 'rocket' ) .
+				' <br>' .
+				// Translators: %1$s = opening link tag, %2$s = closing link tag.
+				sprintf( __( 'To resolve, please %1$scontact support%2$s.', 'rocket' ), '<a href="https://wp-rocket.me/support/" rel="noopener noreferrer" target=_"blank">', '</a>' ),
+			]
 		);
-	}else {
-		delete_transient( 'rocket_check_key_errors' );
+
+		return $valid_details;
 	}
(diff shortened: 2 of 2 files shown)


inc/functions/options.php
```diff
@@ -459,7 +459,20 @@ function rocket_valid_key() {
...
+		set_transient(
+			'rocket_check_key_errors',
+			// Translators: %1$s = opening link tag, %2$s = closing link tag.
+			[ __( 'Not valid License details.', 'rocket' ) . ' <br>' . sprintf( __( 'To resolve, please %1$scontact support%2$s.', 'rocket' ), '<a href="https://wp-rocket.me/support/" rel="noopener noreferrer" target=_"blank">', '</a>' ) ]
+		);
+	}else {
+		delete_transient( 'rocket_check_key_errors' );
+	}
+
+	return $valid_details;
```
remyperona:
```suggestion
	if ( ! $valid_details ) {
		set_transient(
			'rocket_check_key_errors',
			// Translators: %1$s = opening link tag, %2$s = closing link tag.
			[ __( 'The provided license data are not valid.', 'rocket' ) . ' <br>' . sprintf( __( 'To resolve, please %1$scontact support%2$s.', 'rocket' ), '<a href="https://wp-rocket.me/support/" rel="noopener noreferrer" target=_"blank">', '</a>' ) ]
		);

		return $valid_details;
	}

	delete_transient( 'rocket_check_key_errors' );
	return $valid_details;
```

A quick suggestion to return early, and also update the error message

remyperona (changes requested):

inc/functions/admin.php
```diff
@@ -8,7 +8,7 @@
  */
 function rocket_need_api_key() {
 	$message = '';
-	$errors  = get_transient( 'rocket_check_key_errors' );
+	$errors  = (array) get_transient( 'rocket_check_key_errors' );
 
 	if ( false !== $errors ) {
```
remyperona: With this change, this condition will always be met, because `$errors` is always an array. Maybe it can be changed to `! empty( $errors )`

engahmeds3ed: @Tabrisrp 
I think we don't need this check anymore as we are using foreach,
resolved by removing it completely.

remyperona (approved):

piotrbak (commented):
There's one inconsistency only. When the key is not correct, we're setting the `The provided license data are not valid.` error. After clicking on the `Validate License` the error gets overwritten by `License validation failed: This license key is not recognized.`.

I don't think this is an issue but wanted to bring it up here.

Everything else worked as expected:
[testrail-report-153.pdf](https://github.com/wp-media/wp-rocket/files/6199941/testrail-report-153.pdf)

remyperona:
I feel that's ok as it's communicating the same issue with a different wording.

PR #3713, opened 2021-03-24
Fixes #2776 Process buffer only if there is a closing </html> tag during cache

### LINKED ISSUE #2776: LiteSpeed broken layout incompatibility
**Describe the bug**
When using LiteSpeed (and seems that in connection with WooCommerce), double click on the WooCommerce page creates a broken layout cache.

I wasn't able to replicate this on pages other than WooCommerce ones.

**To Reproduce**
1. On WooCommerce site hosted on LiteSpeed
2. Clear all caches, including OPCache
3. Go to the taxonomy page
4. Double-click on the chosen product
5. See if the layout is broken

**Expected behavior**
WP Rocket caching the pages correctly.

**Screenshots**
Double click on the product page:
https://jmp.sh/ve9YdcY

Double click on taxonomy page:
https://jmp.sh/4n8e1nJ

**Additional context**
I was working on one ticket related to this issue, I was pretty sure that it's related to the HTTP/3 protocol and QUIC, but I have a second ticket with HTTP/2.

Nothing special in the WP Debug log. The cached HTML file just stops at some point.

In the first ticket, it was happening only for mobiles (still not for all), in the second one, I've reproduced it on the desktop.

I'm getting more information regarding the second environment, as it seems that it has a shared cache directory.

**Related tickets:**
1. We have access to the full cPanel and we can test there whatever needed:
https://secure.helpscout.net/conversation/1173326494/167375?folderId=2415573

2. New ticket (from screencast), still getting more information about the environment:
https://secure.helpscout.net/conversation/1165493465/165403?folderId=2415573

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

vmanthos:
## Description

There are [reports](https://github.com/wp-media/wp-rocket/issues/2776) where pages are cached partially.

These happen randomly, and the issue is resolved by clearing the cache.

This PR will prevent a page that doesn't have a `</html>` to be saved on the disk. A check is in place for request from the REST API, in case the REST API cache is enabled.

Fixes #2776

## Type of change

Grooming wasn't done on this issue

## How Has This Been Tested?

- [x] Locally.
- [x] On a customer's website.

# Checklist:

### DIFF
--- a/inc/classes/Buffer/class-tests.php
+++ b/inc/classes/Buffer/class-tests.php
@@ -804,7 +804,7 @@ class Tests {
 	 * @return bool
 	 */
 	public function is_html( $buffer ) {
-		return (bool) preg_match( '/<\/html>/i', $buffer );
+		return (bool) preg_match( '/<\s*\/\s*html\s*>/i', $buffer );
 	}
 
 	/** ----------------------------------------------------------------------------------------- */
--- a/inc/classes/Buffer/class-tests.php
+++ b/inc/classes/Buffer/class-tests.php
@@ -66,6 +66,7 @@ class Tests {
 		'donotcachepage'   => 1,
 		'wp_404'           => 1,
 		'search'           => 1,
+		'is_html'          => 1,
 	];
 
 	/**
@@ -348,6 +349,12 @@ class Tests {
 		if ( $this->has_test( 'search' ) && $this->is_search() ) {
 			// Don't process search results.
 			$this->set_error( 'Search page is excluded.' );
+			return false;
+		}
+
+		if ( $this->has_test( 'is_html' ) && ! $this->is_html( $buffer ) ) {
+			// Don't process if there isn't a closing </html>.
+			$this->set_error( 'No closing </html> was found.' );
 			return false;
 		}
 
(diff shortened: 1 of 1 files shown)


remyperona:
I can see one issue with that change: this would prevent caching the REST API requests, when caching for them is enabled.

That is because the REST API responses are JSON, so they won't have the closing html tag.

vmanthos:
@Tabrisrp Thanks for checking this.

We can check if a request is a REST API one.

Changing this line :
```PHP
if ( $this->has_test( 'is_html' ) && ! $this->is_html( $buffer ) ) {
```
into this:
```PHP
if ( $this->has_test( 'is_html' ) && ! $this->is_html( $buffer ) && ! defined( 'REST_REQUEST' ) ) {
```
 should do it.

What do you think?

remyperona (approved):

piotrbak:
@Tabrisrp Seems that we're fine here. I do have a question though, we're very strict about the `</html>` here, do you think that possible that the closing tag will contain a space, or it's a bit different? 🤔 

If not, we're ready to go:
[testrail-report-164.pdf](https://github.com/wp-media/wp-rocket/files/6306204/testrail-report-164.pdf)

remyperona:
I updated the pattern because we never know.

PR #3733, opened 2021-03-29
Add tests for warmup resources and refactor code

engahmeds3ed:
## Description

Add tests for warmup resources


inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -147,6 +147,14 @@ public function remove_used_css_from_html( string $html, array $unprocessed_css
...
 		$unprocessed_styles = $this->unprocessed_flat_array( 'inline', $unprocessed_css );
 
+		if ( ! $link_styles ) {
+			$link_styles = [];
+		}
+
+		if ( ! $inline_styles ) {
+			$inline_styles = [];
+		}
+
```
remyperona: @wp-media/php I'm thinking this kind of things indicates a flaw in the design of the method. the `find()` method could return an empty array instead of `false`. It has 2 advantages:
- A single return type
- No need for this type of conditional checks, since the variables will always be set to an iteratable value

What do you think?

engahmeds3ed: You are totally right, I will refactor this part to also use UrlTrait and make it return empty array if regex search failed.

crystinutzaa:
@engahmeds3ed tests are failing. I think that first we should fix the tests before approving or reviewing the code of this PR.

Just take a look at the errors in tests:
```
The data provider specified for WP_Rocket\Tests\Unit\inc\Engine\Optimization\RUCSS\Warmup\APIClient\Test_SendWarmupRequest::testShouldSendWarmupRequest is invalid.
```

engahmeds3ed:
@crystinutzaa 
It's not the case locally, all tests are passing locally.
Let me check the errors on actions.

inc/classes/dependencies/mobiledetect/mobiledetectlib/Mobile_Detect.php
```diff
@@ -22,1464 +22,1464 @@
  */
```
remyperona, line 22: Why is this file changed?

tests/Integration/TestCase.php
```diff
@@ -92,27 +92,4 @@ protected function loadTestDataConfig() {
 
 		$this->config = $this->getTestData( dirname( $filename ), basename( $filename, '.php' ) );
 	}
-
```
Tabrisrp, line 95: These methods are used in quite a number of tests, why move them outside of the base testCase to a trait?

inc/classes/dependencies/mobiledetect/mobiledetectlib/Mobile_Detect.php
engahmeds3ed: I think this is pushed by mistake after running composer install, but I remember that I reverted it back, let me check

remyperona (approved):

PR #3735, opened 2021-03-30
Preserve backslash in the used CSS

crystinutzaa:
## Description

Related to : https://github.com/wp-media/tree-shaker/issues/73

Tested this and reproduced on 2 different sites. The backslash is removed from the CSS `content` attribute in the generated used CSS.

An example:

Original CSS rule:
```
.icon--calendar::before {
    content: "\e02b";
}
```

Used CSS rule:
```
.icon--calendar::before {
    content: "e02b";
}
```

This type of content value is often used by icons fonts. 

## Fixes  & Solution

The issue is related to `stripslashes` from Query on all strings. This is causing the error. In order to bypass this I replaced the `content: \e` with `shaker-parser:\'dashe` when storing it in the DB and when reading I am converting it back.

## Type of change

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
+++ b/inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
@@ -50,7 +50,7 @@ class Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() : array {
 		return [
-			'rocket_buffer' => [ 'treeshake', 1 ],
+			'rocket_buffer' => [ 'treeshake', 11 ],
 		];
 	}
 
--- a/inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
+++ b/inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
@@ -50,7 +50,7 @@ class Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() : array {
 		return [
-			'rocket_buffer' => [ 'treeshake', 11 ],
+			'rocket_buffer' => [ 'treeshake', 12 ],
 		];
 	}
 
--- a/inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
+++ b/inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
@@ -40,7 +40,7 @@ class Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() : array {
 		return [
-			'rocket_buffer' => [ 'collect_resources', 11 ],
+			'rocket_buffer' => [ 'collect_resources', 1 ],
 		];
 	}
 
--- a/inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
+++ b/inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
@@ -40,7 +40,7 @@ class Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() : array {
 		return [
-			'rocket_buffer' => 'collect_resources',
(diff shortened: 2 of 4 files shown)


inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
```diff
@@ -50,7 +50,7 @@ public function __construct( Options_Data $options, UsedCSS $used_css, APIClient
 	 */
 	public static function get_subscribed_events() : array {
 		return [
-			'rocket_buffer' => [ 'treeshake', 1 ],
+			'rocket_buffer' => [ 'treeshake', 11 ],
```
iCaspar: This is more a question than a change request:
Should this be 11 (running just after all the others on the default priority) or 9 (running just before)?
Seems like we had a conversation about needing to run some other optimizations on the CSS after it had been shaken (font-display, etc.)?

iCaspar (approved):

inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
remyperona: Optimizations on `rocket_buffer` start at priority 14, so the 11 is ok there.

remyperona:
@crystinutzaa Can you clean up the PR description to keep only things that are relevant to it?

inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
crystinutzaa: Yes. You are totally right.
The issue is that Warmup is setup on priority 10. And treeshaker needs to run after Warmup.
Right now there is a bug if I let it on priority 1 as Warmup will never pickup the CSS resources because are already stripped by TreeShake ( treeshake runs faster than warmup ).

This change is required, however these will not be the final priorities.

I think it would be better that Warmup to go on 5 and TreeShake on 7. So we have opened positions before, in between and after Warmup and TreeShake. However we need to decide it together

crystinutzaa: Just saw @Tabrisrp last reply with optimizations starting at 14. 
What about setting to 12 so we have some space also in between just in case we need to run something in between these 2 operations?

crystinutzaa:
@Tabrisrp cleaned it.  😄

inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
remyperona: I'm thinking we can have warmup at `1`, and treeshake at `12`?

crystinutzaa: Yeah....that should work too. Should I make the change on this PR?

remyperona: I'd say yes, that way it's updated right away

inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
```diff
@@ -40,7 +40,7 @@ public function __construct( Options_Data $options, ResourceFetcher $resource_fe
 	 */
 	public static function get_subscribed_events() : array {
 		return [
-			'rocket_buffer' => 'collect_resources',
+			'rocket_buffer' => [ 'collect_resources', 11 ],
```
Tabrisrp, line 43: I thought we agreed on `1`, not `11`?

remyperona:
```suggestion
			'rocket_buffer' => [ 'collect_resources', 1 ],
```

remyperona (changes requested):

remyperona (approved):

PR #3744, opened 2021-03-31
Fixes #3743 Improve i18n of add-ons list separated with ' and '

### LINKED ISSUE #3743: [i18n] The add-ons list separator ' and ' is not translatable
**Describe the bug**
The add-ons list on the below message is not fully translatable because it uses a non-i18n [implode()](https://github.com/wp-media/wp-rocket/blob/develop/inc/Engine/Admin/Settings/Page.php#L1584):
imagem

WordPress can add a fully localized list separated with `','` and `' and '`.
`wp_sprintf()` with `$l` can be used instead of regular `sprintf()` with custom non-i18n implode with ' and '.
Check out the function referrence:
https://developer.wordpress.org/reference/functions/wp_sprintf_l/

**To Reproduce**
1. Go to Settings > Add-ons
2. Enable both CloudPress and Sucuri Add-ons
3. Go to Settings > CDN
4. See error in translated string:
imagem

**Expected behavior**
The `' and '` should be also translatable, and can be done by WordPress core itself in a very elegant way :-)

pedro-mendonca:
## Description
Fixes i18n of list of add-ons separated by non-i18n `' and '`, that is now translated by WP core itself.

WordPress itself can add a fully localized list separated with `','` and `' and '`.
Use `wp_sprintf()` with `$l` instead of regular `sprintf()` with custom non-i18n implode with ' and '.
Check out the function referrence:
https://developer.wordpress.org/reference/functions/wp_sprintf_l/

Fixes #3743

## Type of change

# Checklist:

### DIFF
--- a/inc/Engine/Admin/Settings/Page.php
+++ b/inc/Engine/Admin/Settings/Page.php
@@ -1572,16 +1572,16 @@ class Page {
 		}
 
 		if ( ! empty( $addons ) ) {
-			$maybe_display_cdn_helper = sprintf(
-				// translators: %1$s = opening em tag, %2$s = add-on name(s), %3$s = closing em tag.
+			$maybe_display_cdn_helper = wp_sprintf(
+				// translators: %1$s = opening em tag, %2$l = list of add-on name(s), %3$s = closing em tag.
 				_n(
-					'%1$s%2$s Add-on%3$s is currently enabled. Configuration of the CDN settings is not required for %2$s to work on your site.',
-					'%1$s%2$s Add-ons%3$s are currently enabled. Configuration of the CDN settings is not required for %2$s to work on your site.',
+					'%1$s%2$l Add-on%3$s is currently enabled. Configuration of the CDN settings is not required for %2$l to work on your site.',
+					'%1$s%2$l Add-ons%3$s are currently enabled. Configuration of the CDN settings is not required for %2$l to work on your site.',
 					count( $addons ),
 					'rocket'
 				),
 				'<em>',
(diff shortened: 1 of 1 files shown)


remyperona (approved):

remyperona:
Thank you @pedro-mendonca for the PR, I reviewed and validated it, we will add it to the next major release of the plugin 👍🏼 (because it requires translations updates)

pedro-mendonca:
Thank you @Tabrisrp 👍

Mai-Saad:
@pedro-mendonca @Tabrisrp Currently only "and" is translated while rest of text is not affected, shouldn't it be translated as well? (it was already translated on trunk)

Example for Français language:

Screenshot from 2021-04-05 13-37-28

pedro-mendonca:
@Mai-Saad the original string and the translation must be changed according this code change.
The variable must be changed from `%2$s` to `%2$l`  (`s` -> `l`) in both source code and .po/.mo files.
You are probably using the old `%2$s` variable on the translation string, that do not match this change.

remyperona:
@Mai-Saad That's expected, the translation is handled separately in Transifex. It will be updated before we release the version with this change 👍🏼

Mai-Saad (approved):
Since the translation is handled separately in Transifex. It will be updated before we release the version with this change then this change is working as expected.

Notes:
- for prerelease we need to ensure that translations are fixed (both with/without and case)
- Our tests will fail now, but shall pass with translations fix
- Checked with Francais, Deutsch, Italiano, Turkce

PR #3746, opened 2021-03-31
Closes #3739 Delay JS v2: Admin side

### LINKED ISSUE #3739: Delay JS v2: Admin part (1/2)
Sub-task of epic #3738 

### `Engine\Optimization\DelayJS\Admin\Settings`
- Remove the `$defaults` property
- Remove the `$options` property (not used at all)
- Remove `__construct()` method
- Remove the `get_button_data()` method
- Remove the `option_update_3_7_4()` method
- Remove the `option_update_3_7_2()` method
- Remove the `option_update_3_7_6_1()` method
- Remove the `restore_defaults()` method

#### `add_options()`
- Replace the `delay_js_scripts` array index by `delay_js_exclusions` and the value by an empty array

#### `set_option_on_update()`
- Update the version to compare to `3.9`
- If the existing value for `delay_js` is `1`, update the `delay_js_exclusions` field with:
    - Add the exclusion pattern for the content and wp-includes directories
    - Add the exclusion pattern for jQuery

#### `sanitize_options()`
- Migrate the sanitization of these options from `Engine\Admin\Settings\Settings::sanitize_callback()` to this new method

#### Automated tests
- Remove tests for removed methods
- Update tests for existing methods
- Add tests for `sanitize_options()`

### `Engine\Optimization\DelayJS\Admin\Subscriber`
- Remove `Abstract_Render` as a parent class
- Remove `display_restore_defaults_button()` method
- Remove `option_update_3_7_2()` method
- Remove `option_update_3_7_4()` method
- Remove `option_update_3_7_6_1()` method
- Remove `restore_defaults()`

#### `__construct()`
- Remove the template parameter
- Remove the call to the parent construct

#### `get_subscribed_events()`
- Remove hooks and callbacks unused
- Add new callback for `rocket_input_sanitize` filter

#### `sanitize_options()`
- Callback for `rocket_input_sanitize`, calling `Settings::sanitize_options()`

#### Automated tests
- Remove tests for removed methods
- Update tests for existing methods
- Add tests for `sanitize_options()`

### `Engine\Optimization\DelayJS\Admin\ServiceProvider`
- Update instantiation of `delay_js_settings` to remove the `options` argument
- Update instantiation of `delay_js_admin_subscriber` to remove the `template` argument

### Estimate the effort
Effort `[S]`

remyperona:
## Description

Update the admin side of the delay JS feature:
- Remove no longer relevant methods
- Update existing methods for the new exclusion field
- Add new method for sanitization of the options
- Add/Update/Remove tests

Closes #3739

## Type of change

No

## How Has This Been Tested?

- [x] Add/Update unit/integration tests

# Checklist:


inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -1,253 +1,95 @@
...
 	 *
-	 * @return bool|string
+	 * @return string
 	 */
-	public function restore_defaults() {
-		if ( ! current_user_can( 'rocket_manage_options' ) ) {
-			return false;
+	private function get_excluded_internal_paths() : string {
+		$wp_content = wp_parse_url( content_url(), PHP_URL_PATH );
+		$pattern    = 'wp-includes(.*)';
```
engahmeds3ed: What about using `includes_url` function OR at least `WPINC` constant?

inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -1,253 +1,95 @@
...
+
+		if ( ! $wp_content ) {
+			return $pattern;
 		}
 
-		return implode( "\n", $this->defaults );
+		return "{$pattern}|{$wp_content}(.*)";
```
engahmeds3ed: I think we need here to return array and merge them with jquery regex to make the option uniformed on the textarea without using `|`

what do you think ?

engahmeds3ed (changes requested):
I have two questions here:-
1. Do we need to remove styles/scripts related to `delay_js_scripts` and `restore button` here on this PR?
2. Do we really exclude all internal scripts from being delayed?

Thanks.

inc/Engine/Optimization/DelayJS/Admin/Settings.php
remyperona: I was going with 2 different entries in the textarea:
- One for the pattern for wp-content and wp-includes
- One for jquery pattern

inc/Engine/Optimization/DelayJS/Admin/Settings.php
remyperona: true it's better since there is a filter there

engahmeds3ed: Yes but be careful as this will return the full url for `wp-includes` directory and I don't know if you need also to exclude any external url that has `wp-includes` keyword on it or not?!

remyperona:
> Do we need to remove styles/scripts related to delay_js_scripts and restore button here on this PR?

Since the option doesn't exist anymore, they will be automatically removed from the array the next time the option is updated.

> Do we really exclude all internal scripts from being delayed?

For users updating to 3.9 with the delay JS option enabled yes. This is to avoid issues as the logic changed between the 2 versions. They will be able to remove the exclusions manually if they want to.

For new installations, there is no exclusions by default.

remyperona:
@engahmeds3ed If you can re-review

engahmeds3ed (approved):

PR #3752, opened 2021-04-01
Save stylesheet media attribute into DB and send it to warmup

engahmeds3ed:
## Description

- [x] Add new DB column for media to resources DB
- [x] Remove unique constraint from resources DB
- [x] Detect Stylesheet media attribute.
- [x] Save media into resources DB.
- [x] Send media to warmup.
- [x] Remove unique from url column at used_css table


remyperona (approved):

remyperona:
just missing a docblock for the new method

remyperona (approved):

PR #3755, opened 2021-04-02
Fire action when any resource content changed

engahmeds3ed:
Fire a new action `rocket_rucss_file_changed` when any resource on the page changed its content so we can hook this action to empty used_css table as we don't know which resource is attached to which page.

**Note:** The sequence of emptying used_css table would be as follows:-

1. Page is loaded.
2. Resources are being collected via Async process.
3. In another background process we are making the DB checks for the content and detect changed resources then we call saas warmup and also fire this new action.
4. On the same page loading we call saas treeshaker lately and used_css row is created for this page and added to the cache file.
5. In most cases if content changed, used_css row will be deleted right after the cache file being created and used_css row is created also, SO this change will affect the next round of cache.

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php
+++ b/inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php
@@ -63,7 +63,7 @@ class ResourceFetcherProcess extends WP_Rocket_WP_Background_Process {
 	 *
 	 * @param array $resource Resource array consists of url, content, type and media (for css only).
 	 *
-	 * @return false
+	 * @return bool False for success, remove item from queue and True for failure so requeue this item.
 	 */
 	protected function task( $resource ) {
 		if ( ! is_array( $resource ) ) {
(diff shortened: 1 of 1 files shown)


inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php
```diff
@@ -64,6 +71,7 @@ protected function task( $resource ) {
 		}
 
 		if ( $this->resources_query->create_or_update( $resource ) ) {
+			$this->content_changed = true;
 			return ! $this->send_warmup_request( $resource );
```
iCaspar, line 75: Will this ever return true? Presumably `send_warmup_request()` will (happy case) return true, but what if there is an API outage and it comes back false? Then `task()` returns `true`. Do we care? If it should really always return false, per the docblock, we should just `return false` here. Otherwise, maybe update the docblock to say what happens for true.

iCaspar (changes requested):
Just a quick suggestion to clarify the return of `task()`

iCaspar (approved):

PR #3756, opened 2021-04-02
RUCSS cron for auto-retries after 30minutes

crystinutzaa:
## Description

Cron for retry regenerating Used CSS from Saas in case there are any missed CSS files (from unprocesscss list).

In case a page returns a not empty `unprocessed_css` it will trigger a cron (`wp_schedule_single_event`) which will:
1. Grab all URLs which have not empty `unprocessed_css` from the DB
2. Will reset the `retries = 1`
3. Will clean the page cache.

In this way we force the page cache to be regenerated with RUCSS after 30 minutes we identify any missing CSS files from Saas cache.

## Type of change

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
+++ b/inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
@@ -31,11 +31,18 @@ class Subscriber implements Subscriber_Interface {
 	private $used_css;
 
 	/**
+	 * Name of the cron.
+	 *
+	 * @var string
+	 */
+	const CRON_NAME = 'rocket_rucss_retries';
+
+	/**
 	 * Instantiate the class
 	 *
-	 * @param Options_Data $options      Plugin options instance.
-	 * @param UsedCSS      $used_css Settings instance.
-	 * @param APIClient    $api      Database instance.
+	 * @param Options_Data $options  Plugin options instance.
+	 * @param UsedCSS      $used_css UsedCSS instance.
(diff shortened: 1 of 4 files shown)


inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
```diff
@@ -123,6 +135,35 @@ public function treeshake( string $html ) : string {
...
+	 * @return void
+	 */
+	public function schedule_rucss_retry() {
+		$scheduled = wp_next_scheduled( self::CRON_NAME );
+
+		if ( $scheduled ) {
+			return;
+		}
+
+		wp_schedule_single_event( time() + ( 60 * 30 ), self::CRON_NAME );
```
engahmeds3ed: I have two comments here:-
1. What do u think of using this constant to make it more readable
```suggestion
		wp_schedule_single_event( time() + ( 0.5 * HOUR_IN_SECONDS ), self::CRON_NAME );
```

2. I think we need to add filter here so customers/helper plugins can change this period.

What do u think ?

engahmeds3ed (approved):
Just minor change, if u want to enhance more.

inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
crystinutzaa: Yeah. I will apply the first change.

However, for the filter I think we shall wait and implement it only if needed. If I remember well I had this talk with @Tabrisrp and he said to do it like this for now.

PR #3766, opened 2021-04-06
Closes #3738 Delay JS v2

### LINKED ISSUE #3738: Delay JS execution time (LazyLoad Script) - V2
#### What are we proposing to do? 

To build a new way to delay the execution of JavaScript and getting a significant boost on website performance. WP Rocket will delay all JavaScript instead of an inclusion list with the current feature.

#### Why are we doing this? 

With the new Core Web Vitals metrics, the goal for performance tools is to provide the best user experience to deliver as fast as possible content above the fold. To do this, we need to load the images when needed (Lazyload), to avoid render-blocking files (defer JavaScript and defer non-critical CSS), preload critical assets, etc...

Most websites are loading a ton of JavaScript used for non-critical features. With this feature, WP Rocket will preload and delay all JavaScript by respecting the loading order.

As a side effect, it will improve PageSpeed score and loading time.

[Full feature doc](https://docs.google.com/document/d/1s1xbAkFz_o5gK5CtZAGCki-VefBdikIb16dhypv6xsE/edit?usp=sharing)

remyperona:
## Description

WP Rocket will delay all JavaScript instead of an inclusion list with the current feature.

A new lazyload script for scripts has been created to manage that.

Closes #3738

## Type of change

No

## How Has This Been Tested?

- [x] Automated tests added
- [x] Local manual testing of the feature

# Checklist:


piotrbak:
@Tabrisrp
1. I think we should exclude the jQuery files from the combination as we do with Defer JS
2. We should make sure that the combined script is also delayed, right?
3. Shouldn't we add any notice in the UI telling that the DeferJS is not being applied now?

remyperona:
> I think we should exclude the jQuery files from the combination as we do with Defer JS

@GeekPress What do you think, should we exclude jQuery from combine JS when delay JS is enabled?

> We should make sure that the combined script is also delayed, right?

Yes, I retested that right now, and I can see it being delayed and loaded after. Can you confirm?

> Shouldn't we add any notice in the UI telling that the DeferJS is not being applied now?

This was not part of the specs, @GeekPress what do you think?

GeekPress:
@piotrbak Can you please elaborate why we should exclude jQuery? As all inline scripts and local JS files will be excluded, what are you afraid of?

> This was not part of the specs, @GeekPress what do you think?

I don't think it will be necessary.

piotrbak:
@GeekPress I was thinking about the same reason why we did it for DeferJS. JQuery dependent inline script is excluded from the combination and it's present in the source. 
If we combine the jQuery file we'll cause an issue.

piotrbak:
@GeekPress @Tabrisrp I've just found my previous comment and scenarios. I believe that it applies to this case too:
https://github.com/wp-media/wp-rocket/pull/3366#issuecomment-734374159

GeekPress:
@piotrbak Do you want to do that for existing users with default exclusions on Delay JS? Or also for new users?

GeekPress:
Ok, I see the problem, and excluding jQuery is the only workaround. But I don't like this solution as jQuery will becoming render-blocking.

To be sure at 100% we won't break any websites because of jQuery and Delay JS, I will prefer to disable Combine JS too when Delay JS is on. We will have a UI problem to solve, it will still better than breaking websites.

piotrbak:
@GeekPress Yes, this is a problem. For existing users, I think we should:
1. Exclude jQuery from the combination
2. Exclude jQuery from the DelayJS (as we are excluding `wp-content` and `wp-includes` that may contain jQuery dependent scripts)

For the new users, I think we could only exclude it from the combination. If everything is delayed and jQuery is at the very top, we'll preserve the correct order of the scripts. Would it be render-blocking in this case? 🤔 
I might be missing something though, what do you think?

GeekPress:
In 1. does jQuery will be delayed?

GeekPress:
> Would it be render-blocking in this case?

If jQuery is delayed, it won't be render-blocking

piotrbak:
1. jQuery will not be delayed. We're not delaying themes/plugins/core files, so if there's any that's using jQuery it will be a problem.
2. So, for the new users it won't be a problem I think.

remyperona:
@GeekPress What's the final decision on this?

GeekPress:
> @GeekPress What's the final decision on this?

This one 👇

> To be sure at 100% we won't break any websites because of jQuery and Delay JS, I will prefer to disable Combine JS too when Delay JS is on. We will have a UI problem to solve, it will still better than breaking websites.

But I'm seeing a new problem by doing that. For existing users, if we force to disable Defer JS and we automatically exclude all internal JS from Delay JS, it means that all internal JS files will be render-blocking.

So I think the best solution to do here will be : 
- Don't turn off Defer JS is Delay JS is on
- Turn off Combine JS is Delay JS is on

@piotrbak What do you think?

piotrbak:
@GeekPress Am I right to say that all scripts that have `defer` attribute will be executed before the DomContentLoaded event?

I'm also thinking about deferring the inline jQuery scripts feature we have. Are we going to enclose them in the DomContentLoaded, or the Delay JS will be applied before and we will exclude them from this optimization using the [attribute](https://github.com/wp-media/wp-rocket/blob/10978bae5984c359894d3b7cc19800af2a8731de/inc/Engine/Optimization/DeferJS/DeferJS.php#L136)?

GeekPress:
> @GeekPress Am I right to say that all scripts that have defer attribute will be executed before the DomContentLoaded event?

@piotrbak You're right :)

> I'm also thinking about deferring the inline jQuery scripts feature we have. Are we going to enclose them in the DomContentLoaded, or the Delay JS will be applied before and we will exclude them from this optimization using the attribute?

I don't see a problem to load them on DomContentLoaded. The new Delay JS is compatible with DomContentLoaded :)

piotrbak:
@Tabrisrp It works as expected. We're also delaying both Critical CSS scripts, I wasn't able to produce that kind of behaviour, but delaying the script that is removing Critical CSS from the source could cause the delayed FOUC. Do you think it's possible?

If not, we're ready to go with this:
[testrail-report-163.pdf](https://github.com/wp-media/wp-rocket/files/6288123/testrail-report-163.pdf)

remyperona:
WIth the way the delay script works, I believe it should be fine like this. We will see if we get any report about it, and add the exclusion if needed 👍🏼

PR #3769, opened 2021-04-08
When CPCSS + RUCSS are enabled, load the Used CSS async in a file

engahmeds3ed:
## Description

When CPCSS + RUCSS are enabled, load the Used CSS async in a file

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -3,10 +3,13 @@ declare(strict_types=1);
 
 namespace WP_Rocket\Engine\Optimization\RUCSS\Controller;
 
+use WP_Rocket\Admin\Options_Data;
 use WP_Rocket\Engine\Cache\Purge;
 use WP_Rocket\Engine\Optimization\RegexTrait;
 use WP_Rocket\Engine\Optimization\RUCSS\Database\Row\UsedCSS as UsedCSS_Row;
 use WP_Rocket\Engine\Optimization\RUCSS\Database\Queries\UsedCSS as UsedCSS_Query;
+use WP_Rocket\Engine\Optimization\RUCSS\Frontend\APIClient;
+use WP_Rocket\Logger\Logger;
 
 class UsedCSS {
 	use RegexTrait;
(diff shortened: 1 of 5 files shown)


inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
```diff
@@ -79,6 +79,8 @@ public function treeshake( string $html ) : string {
 		$is_mobile = $this->is_mobile();
 		$used_css  = $this->used_css->get_used_css( $url, $is_mobile );
 
+		$this->used_css->set_cpcss_enabled( (bool) $this->options->get( 'async_css', 0 ) );
```
crystinutzaa: Why `set_cpcss_enabled`? 
Why not inject Options in `used_css` and use in there directly?

inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
```diff
@@ -128,7 +130,8 @@ public function treeshake( string $html ) : string {
 		}
 
 		$html = $this->used_css->remove_used_css_from_html( $html, $used_css->unprocessedcss );
-		$html = $this->used_css->add_used_css_to_html( $html, $used_css->css );
+
+		$html = $this->used_css->add_used_css_to_html( $html, $used_css->css, $used_css->url, (bool) $used_css->is_mobile );
```
crystinutzaa: Here are 4 params sent? Why not creating an array and pass a single array?

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -25,6 +25,13 @@ class UsedCSS {
 	 */
 	private $purge;
 
+	/**
+	 * Status of CPCSS option.
+	 *
+	 * @var bool
+	 */
+	private $cpcss_enabled = false;
```
crystinutzaa: What if we will inject either Options or directly Options->cpccss_enable in the contructor and remove the set method?

inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
engahmeds3ed: Yes I thought about that and really I took this decision to separate concerns so the subscriber is dealing with the DB/options and pass it to the usedCss class, but I'm Ok with both options.

inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
engahmeds3ed: I thought of sending the full DB row with all of the used_css details and use what is needed from it, what do u think?

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -137,6 +144,9 @@ public function insert_used_css( array $data ) {
 			return false;
 		}
 
+		// Save used_css into filesystem.
+		$this->save_used_css_in_filesystem( $data['url'], $data['css'], $data['is_mobile'] );
```
crystinutzaa: Why not move the `save_used_css_in_filesystem()` in `save_or_update_used_css()` ?

engahmeds3ed: I did that to only save into filesystem in case the used_css row is successfully inserted or successfully updated and to do that at `save_or_update_used_css` I will change it a bit to check the response from insert/update methods and then call it again twice.

inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
remyperona: I think it's a good idea to send the row object. I don't like using array as parameter too much, because you need to add additional isset validation when using that.

inc/Engine/Optimization/RUCSS/Frontend/Subscriber.php
remyperona: Could we imagine refactoring the Controller and Subscriber so that Options_Data is only injected in the Controller instead?

We can think about the Subscriber as the router for hooks to callbacks

remyperona (changes requested):

crystinutzaa (approved):

remyperona (approved):

PR #3777, opened 2021-04-09
Add tests for RUCSS/Admin/Subscriber

crystinutzaa:
## Description

Subtask of #3635

#### `Admin\Subscriber`
- [x] `cron_clean_rows()` - Cristina
- [x] `schedule_clean_not_commonly_used_rows()` - Cristina
- [x] `delete_used_css_on_update_or_delete()` - Cristina
- [x] `truncate_used_css()` - Cristina

## How Has This Been Tested?

Integration tests pass.

# Checklist:

### DIFF
--- a/tests/Integration/inc/Engine/Optimization/RUCSS/Admin/Subscriber/cronCleanRows.php
+++ b/tests/Integration/inc/Engine/Optimization/RUCSS/Admin/Subscriber/cronCleanRows.php
@@ -14,7 +14,7 @@ use WP_Rocket\Tests\Integration\TestCase;
 class Test_CronCleanRows extends TestCase{
 	use DBTrait;
 
-	private $posts;
+	private $input;
 
 	public static function setUpBeforeClass(): void {
 		self::installFresh();
--- a/tests/Integration/inc/Engine/Optimization/RUCSS/Admin/Subscriber/cronCleanRows.php
+++ b/tests/Integration/inc/Engine/Optimization/RUCSS/Admin/Subscriber/cronCleanRows.php
@@ -1,0 +1,93 @@
+<?php
+declare(strict_types=1);
+
+namespace WP_Rocket\Tests\Integration\inc\Engine\Optimization\RUCSS\Admin\Subscriber;
+
+use WP_Rocket\Tests\Integration\DBTrait;
+use WP_Rocket\Tests\Integration\TestCase;
+
+/**
+ * @covers \WP_Rocket\Engine\Optimization\RUCSS\Admin\Subscriber::cron_clean_rows
+ *
+ * @group  RUCSS
+ */
+class Test_CronCleanRows extends TestCase{
+	use DBTrait;
+
+	private $posts;
+
+	public static function setUpBeforeClass(): void {
+		self::installFresh();
+
+		parent::setUpBeforeClass();
+	}
+
+	public static function tearDownAfterClass() {
+		parent::tearDownAfterClass();
+
+		self::uninstallAll();
+	}
+
+	public function tearDown() : void {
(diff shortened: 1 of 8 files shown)


tests/Integration/inc/Engine/Optimization/RUCSS/Admin/Subscriber/cronCleanRows.php
```diff
@@ -0,0 +1,93 @@
...
+			$rucss_resources_query->add_item( $resource );
+		}
+
+		$result_resources = $rucss_resources_query->query();
+		$this->assertCount( count( $input['resources'] ), $result_resources );
+
+		do_action( 'rocket_rucss_clean_rows_time_event' );
+
+		$rucss_usedcss_query       = $container->get( 'rucss_used_css_query' );
+		$rucss_resources_query     = $container->get( 'rucss_resources_query' );
```
engahmeds3ed, line 76: Why defining them again here?
they are already defined at the start of this test method.

tests/Integration/inc/Engine/Optimization/RUCSS/Admin/Subscriber/deleteUsedCssOnUpdateOrDelete.php
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		$result = $rucss_usedcss_query->query();
+		$this->assertCount( count( $input['items'] ), $result );
+
+		foreach ( $this->posts as $post ) {
+			do_action( 'delete_post', $post->ID );
+		}
+
+		$rucss_usedcss_query = $container->get( 'rucss_used_css_query' );
```
engahmeds3ed, line 64: Also here why defining it twice?

engahmeds3ed (changes requested):
I think we need to add some more comments on the test code, I'm feeling like it's hard to understand the idea behind the test code, what do u think?

For now, just a small change request that is repeated 3 times.

tests/Integration/inc/Engine/Optimization/RUCSS/Admin/Subscriber/cronCleanRows.php
crystinutzaa: @engahmeds3ed we need to get both again from the container because BerlinDB has kinda caching system in place.
So when deleting, the cache value is not deleted so I need to grab a new instance of the Query to retrieve the current values.

crystinutzaa: Due to the previous reason. BerlinDB has caching system. So to re-read the values from the DB I need to grab a new instance of the Query.

tests/Integration/inc/Engine/Optimization/RUCSS/Admin/Subscriber/deleteUsedCssOnUpdateOrDelete.php
crystinutzaa: All these tests are used for deleting from DB, right?

So the idea behind each test is simple:
1. Insert into the DB
2. Count the number of inserted values into DB
3. Clean the DB
4. Count again the left values in the DB

Based on each test: delete which are older than 1 month, truncate table, delete on update, the expectations are setup.
eg. for truncate => always the select after truncate should match 0 items in the DB
eg. for older than 1 month => I insert 2 items, 1 is not old and 1 is older than 1 month. And the expectation is that : After insert I have 2 rows in the DB. After cleaning the old files I have only 1  row in the DB.

engahmeds3ed (approved):

PR #3793, opened 2021-04-15
Keep files on the safelist on the page without removing them

engahmeds3ed:
## Description

Keep files on the safelist on the page without removing them


remyperona (changes requested):
Need unit/integration tests

remyperona:
After discussing with the team, it was decided to go differently:
- Content of the file in the safelist will be included in the generated used CSS
- The file will still be removed from the page

Its's better for a couple reason:
- It makes sure the order of CSS is preserved
- It still reduces the number of HTTP requests on the page

So the changed will be made on the SaaS side instead of the plugin side.

PR #3797, opened 2021-04-16
Save full url into warmup DB

engahmeds3ed:
## Description

When we have relative CSS/JS files, we get their contents properly but saving them with the relative URL into the DB and we need to save the full absolute URL instead of that.


crystinutzaa:
@engahmeds3ed how is the URL sent to Saas?

Is this going to affect Saas side?

In TreeShaker, I am collecting the URLs from the HTML (not normalized). Then those are matched and asked in DataLayer. 

Let's make sure that is fully working.

engahmeds3ed:
This is for warmup only.
Do u remember the part of treeshaker related to getfullurl? So when getting data from datalayer we use the full url.
I tried this and it's working.
U can give it a try

crystinutzaa (approved):

PR #3805, opened 2021-04-19
Clear used_css directory contents

engahmeds3ed:
## Description

- [x] Clear the full used_css directory with truncate table cases.
- [x] Clear used_css one by one with their cases like 'wp_trash_post' and all hooks that uses the callback `delete_used_css_on_update_or_delete`
- [x] Unit tests
- [x] Integration tests

**Completely remove used_css directory contents:**
- Save options (with cpcss changed disabled)
- Click the clear used CSS button
- switch theme

**In the following cases we remove only one by one (I think those also are purging cache)**
- wp_trash_post
- delete_post
- clean_post_cache
- wp_update_comment_count

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
+++ b/inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
@@ -53,7 +53,10 @@ class Subscriber implements Subscriber_Interface {
 		return [
 			'rocket_first_install_options'       => 'add_options_first_time',
 			'rocket_input_sanitize'              => [ 'sanitize_options', 14, 2 ],
-			'update_option_' . $slug             => [ 'clean_used_css_and_cache', 10, 2 ],
+			'update_option_' . $slug             => [
+				[ 'clean_used_css_and_cache', 10, 2 ],
+				[ 'clean_used_css_directory_cpcss_disabled', 10, 2 ],
+			],
 			'wp_rocket_upgrade'                  => [
 				[ 'set_option_on_update', 14, 2 ],
 			],
@@ -136,6 +139,7 @@ class Subscriber implements Subscriber_Interface {
 		}
 
 		$this->database->truncate_used_css_table();
+		$this->used_css->delete_all_used_css_files();
 	}
 
 	/**
@@ -184,12 +188,12 @@ class Subscriber implements Subscriber_Interface {
 	 *
 	 * @since 3.9
 	 *
-	 * @param array $value     An array of previous values for the settings.
 	 * @param array $old_value An array of submitted values for the settings.
+	 * @param array $value     An array of previous values for the settings.
 	 *
 	 * @return void
 	 */
-	public function clean_used_css_and_cache( $value, $old_value ) {
+	public function clean_used_css_and_cache( $old_value, $value ) {
 		if ( ! current_user_can( 'rocket_manage_options' )
 			||
 			! $this->settings->is_enabled()
@@ -205,6 +209,34 @@ class Subscriber implements Subscriber_Interface {
 			$this->database->truncate_used_css_table();
 			// Clear all caching files.
 			rocket_clean_domain();
+			$this->used_css->delete_all_used_css_files();
+		}
+	}
+
+	/**
+	 * Clean used_css directory with CPCSS option disabled.
+	 *
+	 * @since 3.9
+	 *
+	 * @param array $old_value Options old value.
+	 * @param array $value Options new value.
+	 */
+	public function clean_used_css_directory_cpcss_disabled( array $old_value, array $value ) {
+		if ( ! current_user_can( 'rocket_manage_options' )
+			||
+			! $this->settings->is_enabled()
+		) {
+			return;
+		}
+
+		if (
+			isset( $value['async_css'], $old_value['async_css'] )
+			&&
+			$value['async_css'] !== $old_value['async_css']
+			&&
+			! (bool) $value['async_css']
+		) {
+			$this->used_css->delete_all_used_css_files();
 		}
 	}
 
@@ -237,6 +269,7 @@ class Subscriber implements Subscriber_Interface {
 
 		$this->database->truncate_used_css_table();
 		rocket_clean_domain();
+		$this->used_css->delete_all_used_css_files();
 
 		set_transient(
 			'rocket_clear_usedcss_response',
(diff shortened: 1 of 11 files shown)


tests/Integration/inc/Engine/Optimization/RUCSS/Admin/Subscriber/cleanUsedCssDirectoryCpcssDisabled.php
```diff
@@ -0,0 +1,78 @@
...
+		CapTrait::setAdminCap();
+
+		self::$admin_user_id = static::factory()->user->create( [ 'role' => 'administrator' ] );
+		self::$contributer_user_id = static::factory()->user->create( [ 'role' => 'contributor' ] );
+	}
+
+	public function tearDown() : void {
+
+		parent::tearDown();
+	}
```
iCaspar: I think this is not needed as it doesn't do anything in addition to the parent.

iCaspar: Or, possibly you intended to do
```
remove_filter( 'pre_get_rocket_option_remove_unused_css', [ $this, 'set_rucss_option' ] );
```
Which was added on L:54?

iCaspar (changes requested):

tests/Integration/inc/Engine/Optimization/RUCSS/Admin/Subscriber/cleanUsedCssDirectoryCpcssDisabled.php
engahmeds3ed: Awesome catch, Done

iCaspar (approved):

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
```diff
@@ -136,6 +139,7 @@ public function truncate_used_css() {
 		}
 
 		$this->database->truncate_used_css_table();
+		$this->used_css->delete_all_used_css_files();
```
crystinutzaa, line 151: I don't think we need this in here.

For theme change, indeed the full WP Rocket cache will be cleaned. However this truncate_used_css() will be triggered also on: `rocket_rucss_file_changed` hook which do not need to clean also the files.

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
```diff
@@ -205,6 +209,34 @@ public function clean_used_css_and_cache( $value, $old_value ) {
...
+		}
+
+		if (
+			isset( $value['async_css'], $old_value['async_css'] )
+			&&
+			$value['async_css'] !== $old_value['async_css']
+			&&
+			! (bool) $value['async_css']
+		) {
+			$this->used_css->delete_all_used_css_files();
```
crystinutzaa: I think in here we need to clean also Rocket Cache?

Or is handled in Critical CSS code when CPCSS is disabled?

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
engahmeds3ed: When WP Rocket cache is cleaned, `used-css` directory isn't removed.

But I think really we need to remove it from here but not sure.

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -220,6 +220,9 @@ public function delete_used_css( string $url ) : bool {
 			if ( empty( $used_css->id ) ) {
 				continue;
 			}
+			// Delete file from filesystem.
+			$this->delete_used_css_file( $used_css );
```
crystinutzaa, line 224: This should be triggered only when also the cache file is deleted. Otherwise can lead to a 404 page.
Are you 100% sure this happens on cache clean too?

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
engahmeds3ed: cache is removed with saving options, with cpcss and any other option

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
engahmeds3ed: this is called with the following hooks:-

https://github.com/wp-media/wp-rocket/blob/2910581723efa98c4f45127941c4ba23631dc9b6/inc/Engine/Optimization/RUCSS/Admin/Subscriber.php#L65-L68

and I think all of them are removing the cache also.

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
crystinutzaa: It should NOT be removed from here. Why? This function is hooked in 2 places: 
```
'switch_theme'                       => 'truncate_used_css',
'rocket_rucss_file_changed'          => 'truncate_used_css',
```

On switch_theme we clean the WP Rocket cache, right? Then we can safely delete also the used_css directory.
However, when is `rocket_rucss_file_changed` triggererd? From Warmup side, in case we identify a used css that was changed. During that operation we don't clean also WP Rocket cache, so we will end up in the 404 situation.

So, cleaning the used-css folder should happen only if the page cache file is deleted. Otherwise: 404 and broken page style.

engahmeds3ed: Theoretically, `rocket_rucss_file_changed` is called just after sending all resources to the warmup before calling treeshaker on the cache file generation and before putting the used_css file into the cached page.

But to be on the safe side, I will remove it only with `switch_theme`.

engahmeds3ed: Done

crystinutzaa: And `rocket_rucss_file_changed` could be called for a single page where we identify a new CSS file which do not exist on other pages. This means, that the cache will be created only for that specific page and the full cache will not be regenerating.

So cleaning the full used css file when we identify a single resource changed does not make sense.

What do you think?

engahmeds3ed: Exactly.

You can check it now, I adjusted it.

crystinutzaa (approved):

PR #3809, opened 2021-04-19
Fixes Delay JS escaping for ? character

remyperona:
## Description

We can't escape the `?` character specifically, because it's used in the patterns added during the upgrade.

Instead, let's do the escaping on a more specific pattern `?ver`.

Also added a test case for the upgrade exclusion patterns to make sure there is no regression.

## Type of change

## How Has This Been Tested?

- [x] Manual test
- [x] New test cases

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DelayJS/HTML.php
+++ b/inc/Engine/Optimization/DelayJS/HTML.php
@@ -70,8 +70,8 @@ class HTML {
 		$this->excluded = array_map(
 			function( $value ) {
 				return str_replace(
-					[ '+', '?', '#' ],
-					[ '\+', '\?', '\#' ],
+					[ '+', '?ver', '#' ],
+					[ '\+', '\?ver', '\#' ],
 					$value
 				);
 			},
--- a/tests/Fixtures/inc/Engine/Optimization/DelayJS/HTML/delayJs.php
+++ b/tests/Fixtures/inc/Engine/Optimization/DelayJS/HTML/delayJs.php
@@ -15,6 +15,21 @@ $html = '<html>
 </body>
 </html>';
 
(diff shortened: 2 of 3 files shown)


inc/Engine/Optimization/DelayJS/HTML.php
```diff
@@ -70,8 +70,8 @@ public function delay_js( $html ): string {
 		$this->excluded = array_map(
 			function( $value ) {
 				return str_replace(
-					[ '+', '?', '#' ],
-					[ '\+', '\?', '\#' ],
+					[ '+', '?ver', '#' ],
+					[ '\+', '\?ver', '\#' ],
```
engahmeds3ed, line 74: May I ask, why not using preg_quote method here?

remyperona: `preg_quote()` would be escaping characters we don't want to escape, as they are part of the patterns in the exclusion field.

engahmeds3ed: If you mean `(.*)` we can replace `\(\.\*\)` with `(.*)` after applying `preg_quote`.

My fears come from if the file is using `?v` not `?ver`!

if u r sure that those three cases are handling all cases, Ok we can go ahead.

remyperona: Those are the characters that can be part of an URL, and can cause a misinterpretation of the provided pattern when we check for the exclusion.

`?v` might happen, but it's a much rarer occurrence. Also, it can be manually removed from the pattern in the exclusion field if needed.

engahmeds3ed (approved):

PR #3811, opened 2021-04-19
Handle resources without url schema

engahmeds3ed:
## Description

Handle resources without URL schema, for example, this file

```
<link rel='stylesheet' id='woocommerce-general-css'  href='//new.rocketlabsqa.ovh/wp-content/plugins/woocommerce/assets/css/twenty-twenty.css?ver=5.2.2' media='all' />
```

needs to be saved into resources table with the URL
```
https//new.rocketlabsqa.ovh/wp-content/plugins/woocommerce/assets/css/twenty-twenty.css?ver=5.2.2
```
OR at least
```
http//new.rocketlabsqa.ovh/wp-content/plugins/woocommerce/assets/css/twenty-twenty.css?ver=5.2.2
```

based on the current site supports SSL or not.

### DIFF
--- a/inc/Engine/Optimization/UrlTrait.php
+++ b/inc/Engine/Optimization/UrlTrait.php
@@ -3,6 +3,8 @@ declare(strict_types=1);
 
 namespace WP_Rocket\Engine\Optimization;
 
+use WP_Rocket\Logger\Logger;
+
 trait UrlTrait {
 
 	/**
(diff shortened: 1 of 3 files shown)


inc/Engine/Optimization/UrlTrait.php
```diff
@@ -3,6 +3,8 @@
 
 namespace WP_Rocket\Engine\Optimization;
 
+use WP_Rocket\Logger\Logger;
```
iCaspar: I don't see where we are actually using Logger in here. Accedental IDE auto-insertion, maybe?

iCaspar (changes requested):

inc/Engine/Optimization/UrlTrait.php
engahmeds3ed: this was added during my investigation of the issue, good catch.

Done

iCaspar (approved):

PR #3819, opened 2021-04-20
Remove set_options_on_update() methods

remyperona:
## Description

This method is not really necessary, as the new options values will be set anyway on the next saving of the settings, and they don't differ from the default values.

It was also causing an issue when going to the settings page right after updating the plugin:

- Go to the settings
- Enable RUCSS and save settings
- Page is reloaded
- Refresh the page, RUCSS is still disabled
- Enable it once again and save settings
- Refresh the page
- It's enabled

This is happening because the `admin_init` is first triggered on the page load, triggering `wp_rocket_upgrade`. Meanwhile while saving the changes, the options get out of sync, and we end up with this situation.

## Type of change

# Checklist:


engahmeds3ed (approved):
Yes, the default value when getting the option is set to `0 (false)`.
I can't think if this may make any issue even in export settings the RUCSS option will not be there but also the code default is 0.

PR #3823, opened 2021-04-21
RUCSS bugfix: URLs not matching due to html special chars encoding

crystinutzaa:
## Description

URLs with `&` sometimes are coming `htmlspecialchars_encoded` and this will transform the `&` into `&#038;`.
This is triggering a lot of side effects:
1. The URL is coming from SAAS in the `unprocessedlist` with `&`, however in the HTML is identified as `&#038;`. This will cause the removal of the CSS file from being not processed due to not matching the full URL.
2. Sending the URL from WP Rocket to SAAS will be done with  `&#038;`. In Saas side, however is matched and decoded and the `&` will be used. So the effect is that even if the file is correctly warmed up then it will not be matched in SAAS cache due to the different URL.

Solution is to always send and use in WP Rocket the decoded value of the url.
Example:

- WP Rocket identified this URL:

`//example.org/wp-content/uploads/oxygen/css/home-26.css?cache=1616068597&#038;ver=5.7.1` => this is stored like this in resources DB table.

- Warmup SAAS saves the URL in cache with the following URL:

`//example.org/wp-content/uploads/oxygen/css/home-26.css?cache=1616068597&#038;ver=5.7.1` 

- In SAAS in TreeShaker the Dom picks the URL from the HTML page like this:

`//example.org/wp-content/uploads/oxygen/css/home-26.css?cache=1616068597&ver=5.7.1` => 

- TreeShake tries to read from cache file => identified it does not exist in cache ( because `&#038; != &` ) and returns to WP Rocket the `unprocessedlist`

```
 "unProcessedCss": [
            {
                "type": "link",
                "content": "//example.org/wp-content/uploads/oxygen/css/main-11.css?cache=1615444245&ver=5.7.1"
            },
```

- WP Rocket modified the page and tries to keep the URLs from `unProcessedCSS`, however it failed due to the same mismatch:

`&#038; != &`

## Type of change

## How Has This Been Tested?

On Adame website and also needs complete tests to make sure we cover these cases.

# Checklist:


iCaspar (approved):

licence-data.php
```diff
@@ -1,18 +0,0 @@
-<?php
```
remyperona: This file should not be deleted

remyperona (changes requested):

remyperona (approved):

PR #3827, opened 2021-04-21
Fixes #3821 Disable the combine JS option and field when delay JS is enabled

### LINKED ISSUE #3821: 3.9: Grey out and deactivate Combine JS option when Delay JS option is turned on.
Delay JS feature as the priority on the Combine JS option which is automatically deactivated on front when Delay JS is active. 

To make it more clear for users, we should grey out and deactivate Combine JS when Delay JS option is turned on.

remyperona:
## Description

Reflect on the settings page the disabling of the combine JS option when delay JS is enabled.

It is now:
- Disabled when updating to 3.9 and delay JS is already enabled
- Disabled whenever delay JS is enabled on the settings page and the changes saved

Fixes #3821 

## Type of change

No grooming done

## How Has This Been Tested?

- [x] Manual testing
- [x] Added new tests + updated existing tests

# Checklist:


iCaspar (approved):

Mai-Saad:
@Tabrisrp please find my questions/comments below:

1- When importing old settings in 3.9, while delay exclusion was having value, we clear the delay js exclusion list, shouldn't we keep it?

2- UI of combined gets unchecked and dimmed when enable delay js in 3.9 or update to 3.9 while delay js on but with import settings (while delay js and combine js are on), it`s dimmed checked .. shouldn't we keep it same UI in both cases. ( Recommend the unchecked dimmed one as when we disable delay js => combine will be enabled (unchecked even if it was checked before). what do you think?
Screenshot from 2021-04-22 10-51-29
Screenshot from 2021-04-22 10-50-52

remyperona:
@Mai-Saad These are both issues with the import system itself, it's good to know about them. Can you create a new issue for this?

PR #3838, opened 2021-04-23
RUCSS - Delete used files only after 1 month

crystinutzaa:
## Description

Remove delete used files only at the cron which cleans the unused CSS after 1 month.

Fixes # (issue)

# Checklist:


piotrbak (approved):
We're good to go here:
[testrail-report-170.pdf](https://github.com/wp-media/wp-rocket/files/6372969/testrail-report-170.pdf)

inc/Engine/Optimization/RUCSS/Admin/Database.php
```diff
@@ -81,4 +81,16 @@ public function delete_old_resources() {
...
+	 * Get old used css based on last accessed date.
+	 *
+	 * @return array
+	 */
+	public function get_old_used_css() : array {
+		$old_used_css = [];
+		if ( $this->rucss_usedcss_table->exists() ) {
+			$old_used_css = $this->rucss_usedcss_table->get_old_used_css();
+		}
+		return $old_used_css;
```
engahmeds3ed: Enhancement:

```suggestion
		if ( ! $this->rucss_usedcss_table->exists() ) {
			return [];
		}
		return $this->rucss_usedcss_table->get_old_used_css();
```

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
```diff
@@ -53,14 +54,8 @@ public static function get_subscribed_events() : array {
 		return [
 			'rocket_first_install_options'       => 'add_options_first_time',
 			'rocket_input_sanitize'              => [ 'sanitize_options', 14, 2 ],
-			'update_option_' . $slug             => [
-				[ 'clean_used_css_and_cache', 10, 2 ],
-				[ 'clean_used_css_directory_cpcss_disabled', 10, 2 ],
```
engahmeds3ed, line 58: I thought that we will keep that so we remove the directory completely when CPCSS is disabled!

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
```diff
@@ -84,6 +80,13 @@ public function cron_clean_rows() {
 			return;
 		}
 
+		$old_used_css_ids = $this->database->get_old_used_css();
+		foreach ( $old_used_css_ids as $old_used_css ) {
+			$used_css_item = new UsedCSS_Row( $old_used_css );
```
engahmeds3ed, line 85: We can use the dependency injection to instantiate this class. We can then use the `init` method to initialize the row instance there at `\WP_Rocket\Engine\Optimization\RUCSS\Admin\Database::get_old_used_css` so this method should return an array of rows to be deleted, This will make tests easier and also separate the concerns.

what do u think?

inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php
```diff
@@ -80,4 +80,25 @@ public function delete_old_used_css() {
...
+		// Get the database interface.
+		$db = $this->get_db();
+
+		// Bail if no database interface is available.
+		if ( empty( $db ) ) {
+			return false;
+		}
+
+		$prefixed_table_name = $this->apply_prefix( $this->table_name );
+		$query               = "SELECT * FROM `$prefixed_table_name` WHERE `last_accessed` <= date_sub(now(), interval 1 month)";
```
engahmeds3ed, line 98: what do u think if we selected only `ID` and at cron we can get the full row not with instantiated `UsedCSS_Row` but with  getting by the ID?

engahmeds3ed (changes requested):

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
crystinutzaa: If I remember well, we disabled the file cleaning in case there is external cache which can be out of sync with WP Rocket and might lead to 404 pages.
I think @Tabrisrp can clarify what should happen when CPCSS is disabled.

engahmeds3ed: Yes, got it.
Thanks for clarification @crystinutzaa

inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php
crystinutzaa: Not sure if this will be ok. Because there will be another DB SELECT to grab each UsedCSS_Row element. 
Currently, for this operation there will be only 1 DB select. With grabing a single id and then for each ids to fetch each element, will make too many DB calls.

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
engahmeds3ed: sorry, init method is private so we can't do anything here, the current implementation is the best available solution.

inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php
engahmeds3ed: Ok, we may enhance this query one day to use BerlinDB query and return rows directly.
After a talk with Cristina, it's hard to be done in such a way.

engahmeds3ed (approved):

PR #3847, opened 2021-04-26
Fix saving resource contents with backslash issue

engahmeds3ed:
## Description

Remove `stripslashes()` before saving string content, to avoid issues with content using `\` in it, which ends up being incorrectly stripped.

### DIFF
--- a/inc/Dependencies/Database/Query.php
+++ b/inc/Dependencies/Database/Query.php
@@ -2046,7 +2046,7 @@ class Query extends Base {
 
 			// Strip slashes from all strings
 			if ( is_string( $value ) ) {
-				$value = stripslashes( $value );
+				//$value = stripslashes( $value );// We removed this line at PR #3847 to solve if the content has backslash.
 			}
 
 			// Get the column
--- a/inc/Dependencies/Database/Query.php
+++ b/inc/Dependencies/Database/Query.php
@@ -2045,9 +2045,9 @@ class Query extends Base {
 		foreach ( $item as $key => $value ) {
 
 			// Strip slashes from all strings
-			if ( is_string( $value ) ) {
-				//$value = stripslashes( $value );// We removed this line at PR #3847 to solve if the content has backslash.
-			}
+			/*if ( is_string( $value ) ) {
+				$value = stripslashes( $value );// We removed this line at PR #3847 to solve if the content has backslash.
+			}*/
 
 			// Get the column
 			$column = $this->get_column_by( array( 'name' => $key ) );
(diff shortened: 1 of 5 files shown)


inc/Engine/Optimization/RUCSS/Database/Queries/ResourcesQuery.php
```diff
@@ -78,6 +78,8 @@ public function create_or_update( array $resource ) {
 		// check the database if those resources added before.
 		$db_row = $this->get_item_by( 'url', $resource['url'] );
 
+		$resource['content'] = addslashes( $resource['content'] );
```
crystinutzaa: This should happen in Resource and not in Used CSS?
Isn't this the error coming from Used CSS or I am missing something?

engahmeds3ed: No, the issue was at warmup, when trying to save the resource content into the DB.
But your comment gets into my mind that we need to apply the same at the used_css table also while saving it.

crystinutzaa: Please test it round up: Warmup => Saas Warmup => TreeShake => Used CSS to make sure that is 100% covered .
I am kinda 90% sure this will happen in Used CSS DB table too.

crystinutzaa (approved):

inc/Dependencies/Database/Query.php
```diff
@@ -2046,7 +2046,7 @@ private function validate_item( $item = array() ) {
 
 			// Strip slashes from all strings
 			if ( is_string( $value ) ) {
-				$value = stripslashes( $value );
+				//$value = stripslashes( $value );// We removed this line at PR #3847 to solve if the content has backslash.
 			}
```
remyperona: Should we comment the whole conditional? No need to check for the type if we're not doing anything with it.

engahmeds3ed: Done

remyperona (approved):

remyperona:
@wp-media/qa Ready to be tested. This PR removes the automatic stripping of backslashes in the content before saving it in the local database.

This was creating an issue with various CSS elements, mostly icons fonts, for things like:
- `content: "\f069";`
- `font-family: "Font Awesome\ 5 free";`

Now the content is stored as-is.

Mai-Saad (approved):
Working as expected (keeping the backslash)

Note: 
Having internal and external CSS for the same element will be added twice in used CSS (I think this is not related to this PR and we may not do anything regarding it)

PR #3855, opened 2021-04-27
Fixes #3852 Correctly show updated options values after plugin update

### LINKED ISSUE #3852: 3.9 alpha4: Default exclusion list isn't added when it should
- [x] Made sure you’re on the latest version (3.9-alpha4)

**Describe the bug**
After updating to 3.9-alpha4 on a website that already has Delay JS turned on, the default exclusion isn't here by default. We need to refresh the option page to have the default exclusion added.

**To Reproduce**
1. Be sure the current configuration has Delay JS on.
2. Update to 3.9-alpha4
3. See error, the default exclusion isn't added.
4. Refresh the page a,d the default list will be added

**Expected behavior**
The default exclusion list on Delay JS should be added right after the update.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

As described in the issue, we are in a race condition when loading the settings page as the first page after updating the plugin, which results in incorrect display of the options values sometimes.

Forcing a reload of the page once `wp_rocket_upgrade` is finished will prevent this.

Fixes #3852 

## Type of change

No

## How Has This Been Tested?

- [x] Tested manually by switching between versions of the plugin, and verifying the redirect was correctly triggered

# Checklist:

### DIFF
--- a/inc/admin/upgrader.php
+++ b/inc/admin/upgrader.php
@@ -40,6 +40,15 @@ function rocket_upgrader() {
 
 	$page = filter_input( INPUT_GET, 'page', FILTER_SANITIZE_STRING );
 
+	if (
+		did_action( 'wp_rocket_upgrade' )
+		&&
+		'wprocket' === $page
+	) {
+		wp_safe_redirect( esc_url_raw( admin_url( 'options-general.php?page=wprocket' ) ) );
+		exit;
+	}
+
 	if ( ! rocket_valid_key() && current_user_can( 'rocket_manage_options' ) && 'wprocket' === $page ) {
 		add_action( 'admin_notices', 'rocket_need_api_key' );
 	}
--- a/inc/admin/upgrader.php
+++ b/inc/admin/upgrader.php
@@ -41,9 +41,9 @@ function rocket_upgrader() {
 	$page = filter_input( INPUT_GET, 'page', FILTER_SANITIZE_STRING );
 
 	if (
-		did_action( 'wp_rocket_upgrade' )
-		&&
 		'wprocket' === $page
+		&&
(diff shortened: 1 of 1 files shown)


inc/admin/upgrader.php
```diff
@@ -40,6 +40,15 @@ function rocket_upgrader() {
 
 	$page = filter_input( INPUT_GET, 'page', FILTER_SANITIZE_STRING );
 
+	if (
+		did_action( 'wp_rocket_upgrade' )
+		&&
+		'wprocket' === $page
```
engahmeds3ed: very minor change request,
I think reversing the order of the conditions will be better as this method is called with `admin_init` so in most cases the `$page` will not equal `wprocket` so we don't need to do the second check `did_action`

```suggestion
		'wprocket' === $page
		&&
		did_action( 'wp_rocket_upgrade' )
```

also, I was thinking that here:
https://github.com/wp-media/wp-rocket/blob/7eb62835e5ffa2b3c79c7625fc7921fbe57739d3/inc/admin/upgrader.php#L21
we do this action so why not defining variable here instead of using `did_action`, I think this will work also.

This is very minor so I will approve it.

engahmeds3ed (approved):

Mai-Saad:
The issue is only reproducible when updating from the plugins page so in order to retest this fix we need to merge the branch to the alpha zip file .. @Tabrisrp @engahmeds3ed  what do you think?

Mai-Saad (approved):
(tested on Alpha5 zip) 
[testrail-report-173.pdf](https://github.com/wp-media/wp-rocket/files/6393913/testrail-report-173.pdf)
Working as expected

PR #3858, opened 2021-04-27
Fixes #3853 Exclude inline scripts added by wp_localize_script() from delay JS

### LINKED ISSUE #3853: Delay JS: automatically exclude script added with wp_localize_script
Inline scripts added with `wp_localize_script()` only contains a variable with a JS object value. There is no code execution, which makes it unnecessary to delay, and can potentially cause issues.

Automatically exclude from delay JS scripts with the `js-extra` pattern, which corresponds to inline script added from `wp_localize_script()`

This also mean we can remove the `js-extra` from the delay JS exclusion field.

remyperona:
## Description

Exclude inline scripts added by the `wp_localize_script()` function from delay JS. It is not required to delay them, as there is no code execution in it. And it can also avoid unexpected issues.

Fixes #3853 

## Type of change

No

## How Has This Been Tested?

- [x] Automated tests in place
- [x] Manual tests on pages with this type of inline script

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DelayJS/Admin/Settings.php
+++ b/inc/Engine/Optimization/DelayJS/Admin/Settings.php
@@ -52,7 +52,7 @@ class Settings {
 			$options['delay_js_exclusions']   = [
 				$this->get_excluded_internal_paths(),
 				'/jquery-?[0-9.]*(.min|.slim|.slim.min)?.js',
-				'js-(extra|before|after)',
+				'js-(before|after)',
 			];
 			$options['minify_concatenate_js'] = 0;
 		}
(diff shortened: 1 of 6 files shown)


iCaspar (approved):

engahmeds3ed:
Just for the history,

This PR will completely exclude all  scripts that have `js-extra` into it (the content OR attributes) instead of adding them to the exclusion list at the admin options.

piotrbak (approved):
Good to go here:
[testrail-report-172.pdf](https://github.com/wp-media/wp-rocket/files/6391582/testrail-report-172.pdf)

Issue #3836, opened 2021-04-23
Improve Remove Unused CSS onboarding


remyperona:
**Is your feature request related to a problem? Please describe.**
Our current solution generates partial used CSS on the first time the option is enabled, because of the ramp-up time to get the resources cached on the SaaS side during the cache generation process.

We currently provide no indicator around that, and this might generate unnecessary questions from our users.

**Describe the solution you'd like**
Our new solution to improve on that is to pre-cache the resources earlier, using a similar process than with CPCSS. We will fetch resources on the homepage, CPTs and taxonomies on option activation, and display a progress bar on the settings page to keep the user informed of the process.

Once the process is completed, we will start applying the used CSS on the cached page. This used CSS should be (close to) complete, providing good results on PSI.

**Describe alternatives you've considered**
We considered having the warmup run all the time, but this would be putting to much unnecessary pressure on our servers.

**Additional context**
Detailed doc: https://docs.google.com/document/d/1SMZqpO5UcrjOoB_BNo2bafIpVlgwq8aDyAKOjuEZsos/edit?usp=sharing

crystinutzaa:
@wp-media/php 

I made a plan / the flow on how I see it working, however I think that there are still a lot of missing things which needs brainstorming.

In **WP Rocket** side, we have 2 options for that loading bar.
We can display the progress accurate by displaying it in 2 steps:
1. **Step 1** - _Collect the resources from pages - homepage, CPT & taxonomies_
2. **Step 2** - _Warmup resources._ 
Basically, we will display first the progress of collecting the resources from their page. Below the loading bar we display a text:
`Collecting resources in progress...`
When this step finish and is at 100% , we reset the same loading bar and we make the progress for warmup resources. The text below will be changed to:
`Warming resources in progress...`

**How we will do this?** 
For collecting the resources we need a Background Process which will fetch each page. Will merge and store the found resources in a transient(???).
For warming up the resources, we need another Background Process. This Background process will grab the found resources and send to warmup.
We need an REST API endpoint which will share to our React component the status of each step. It can be the percentage or the X out of Y values and compute the percent in React.
Refreshing / Navigating from WP Rocket options page while this process runs, will always display the accurate progress of this step.


**Questions:**
1. What will happen if the user change their theme? This step need to take place again, right?
2. When disabling & enabling back the RUCSS option, should this step be reloaded? 
3. What will happen if the transient / location where we are storing the progress of this is cleaned?
[...]

PR #3861, opened 2021-04-28
RUCSS Onboarding step 1: Pages scanner

remyperona:
## Description

Add new `Scanner` and `ScannerProcess` classes. Those classes are used on RUCSS activation, to:
- Get URLs to scan and send them to the scanner process
- Scanner process gets the HTML of those URLs, and sends the found resources to the `ResourceFetcher`
- It also stores each processed URL in an option, to be used in the UI to keep track of the status

Sub-task for #3836 

## Type of change

## How Has This Been Tested?

- [x] Manual testing, checking that each step is correctly triggered, and that the resources table is correctly populated

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Warmup/Scanner.php
+++ b/inc/Engine/Optimization/RUCSS/Warmup/Scanner.php
@@ -52,8 +52,21 @@ class Scanner {
 	 * @return void
 	 */
 	public function start_scanner( $old_value, $value ) {
+		if ( ! isset( $value['remove_unused_css'] ) ) {
+			return;
+		}
+
 		if (
-			! isset( $value['remove_unused_css'] )
+			! isset( $old_value['remove_unused_css'] )
+			&&
+			1 === (int) $value['remove_unused_css']
+		) {
+			$this->dispatcher();
+			return;
+		}
+
+		if (
+			! isset( $old_value['remove_unused_css'] )
 			||
 			( $old_value['remove_unused_css'] === $value['remove_unused_css'] )
 			||
--- a/inc/Engine/Optimization/RUCSS/Warmup/Scanner.php
+++ b/inc/Engine/Optimization/RUCSS/Warmup/Scanner.php
@@ -2,6 +2,7 @@
 
 namespace WP_Rocket\Engine\Optimization\RUCSS\Warmup;
 
+use WP_Rocket\Admin\Options;
 use WP_Rocket\Engine\Optimization\ContentTrait;
 
 class Scanner {
@@ -15,6 +16,13 @@ class Scanner {
 	public $process;
 
 	/**
+	 * Options API instance.
+	 *
+	 * @var Options
+	 */
+	private $options_api;
+
+	/**
 	 * Items for which we get the resources to warmup.
 	 *
 	 * @var array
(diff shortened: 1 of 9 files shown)


inc/Engine/Optimization/RUCSS/Warmup/Scanner.php
```diff
@@ -0,0 +1,131 @@
...
+	 * @param array $value     New values for WP Rocket settings.
+	 *
+	 * @return void
+	 */
+	public function start_scanner( $old_value, $value ) {
+		if (
+			! isset( $old_value['remove_unused_css'], $value['remove_unused_css'] )
+			||
+			( $old_value['remove_unused_css'] === $value['remove_unused_css'] )
+			|| 1 !== (int) $value['remove_unused_css']
```
iCaspar: new line here to have condition on its own from `||`

inc/Engine/Optimization/RUCSS/Warmup/Scanner.php
```diff
@@ -0,0 +1,131 @@
...
+	 * @since 3.9
+	 *
+	 * @param array $old_value Previous values for WP Rocket settings.
+	 * @param array $value     New values for WP Rocket settings.
+	 *
+	 * @return void
+	 */
+	public function start_scanner( $old_value, $value ) {
+		if (
+			! isset( $old_value['remove_unused_css'], $value['remove_unused_css'] )
```
engahmeds3ed: I think we removed the code of setting the RUCSS option value with the upgrade so the first time we don't have `$old_value['remove_unused_css']` so this will bail out here, right?

inc/Engine/Optimization/RUCSS/Warmup/Scanner.php
```diff
@@ -0,0 +1,131 @@
...
+	 * @since 3.9
+	 *
+	 * @return void
+	 */
+	private function dispatcher() {
+		$this->set_items();
+
+		array_map( [ $this->process, 'push_to_queue' ], $this->items );
+
+		$this->options_api->set( 'scanner_start_time', current_time() );
```
engahmeds3ed: I like saving it with parameter `U` to get the timestamp so we can calculate easily how much time elapsed in seconds and comparing them with WP constants like `HOUR_IN_SECONDS`, what do u think ?

inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
```diff
@@ -40,7 +49,8 @@ public function __construct( Options_Data $options, ResourceFetcher $resource_fe
 	 */
 	public static function get_subscribed_events() : array {
 		return [
-			'rocket_buffer' => [ 'collect_resources', 1 ],
+			'rocket_buffer' => [ 'collect_resources', 11 ],
+			'update_option_' . rocket_get_constant( 'WP_ROCKET_SLUG' ) => [ 'start_scanner', 15, 2 ],
```
engahmeds3ed, line 53: formatting issue related to aligning this `=>`

iCaspar (changes requested):

engahmeds3ed (changes requested):

inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
remyperona: PHPCS doesn't agree, it shows an error if I change it

inc/Engine/Optimization/RUCSS/Warmup/Scanner.php
remyperona: changed it to use `time()` directly

engahmeds3ed (approved):

PR #3870, opened 2021-04-30
Use time instead of current_time

engahmeds3ed:
## Description

By mistake, we changed time to current_time without passing any argument.

this PR to solve that one

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Warmup/Status/Checker.php
+++ b/inc/Engine/Optimization/RUCSS/Warmup/Status/Checker.php
@@ -56,7 +56,7 @@ class Checker extends AbstractAPIClient {
 			return;
 		}
 
-		if ( time() > strtotime( '+1 hour', $start_time ) ) {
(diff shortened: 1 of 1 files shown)


remyperona (approved):
I kew this code sucked, you found the way of making it much better!

inc/Engine/Optimization/RUCSS/Warmup/Status/Checker.php
```diff
@@ -166,11 +166,11 @@ private function prepare_resources_array( array $items ): array {
 		];
 
 		foreach ( $items as $item ) {
-			if ( 'css' === $item['type'] ) {
-				$resources['css'][] = $item['url'];
-			} elseif ( 'js' === $item['type'] ) {
-				$resources['js'][] = $item['url'];
+			if ( ! in_array( $item->type, [ 'css', 'js' ] ) ){
```
remyperona:
```suggestion
			if ( ! in_array( $item->type, [ 'css', 'js' ], true ) ) {
```

PR #3864, opened 2021-04-29
Pre-warmup RESTAPI endpoint

engahmeds3ed:
## Description

Add a new RESTAPI Endpoint to get the status of resource URLs from the Database.

## Response types

1. Fail when don't have the authority to do that action
```
{
    "code": "rest_forbidden",
    "message": "Sorry, you are not allowed to do that.",
    "data": {
        "status": 401
    }
}
```

2. Fail when RUCSS option is disabled.
```
{
    "success": false,
    "message": "Remove unused CSS option is disabled.",
    "data": []
}
```

3. Fail when pre-warmup process didn't start yet.
```
{
    "success": false,
    "message": "Pre-Warmup process didn't start yet.",
    "data": []
}
```

4. Success with the scan process & some wramed-up and some unwarmed resources
```
{
    "success": true,
    "data": {
        "scan_status": {
            "total_pages": 6,
            "scanned": 6,
            "fetched": 6
        },
        "warmup_status": {
            "total": 3,
            "warmed_count": 2,
            "notwarmed_resources": [
                "http://wordpress.localhost/wp-includes/js/dist/vendor/lodash.js?ver=4.17.19"
            ]
        }
    }
}
```

5. Success without any warmed-up resource yet.
```
{
    "success": true,
    "data": {
        "scan_status": {
            "total_pages": 6,
            "scanned": 6,
            "fetched": 6
        },
        "warmup_status": {
            "total": 3,
            "warmed_count": 0,
            "notwarmed_resources": [
                "http://wordpress.localhost/wp-includes/js/dist/vendor/wp-polyfill-fetch.js?ver=3.0.0",
                "http://wordpress.localhost/wp-includes/js/dist/vendor/wp-polyfill.js?ver=7.4.4",
                "http://wordpress.localhost/wp-content/themes/twentytwenty/assets/js/index.js?ver=1.2"
            ]
        }
    }
}
```

6. Success with all resources warmed-up
```
{
    "success": true,
    "data": {
        "scan_status": {
            "total_pages": 6,
            "scanned": 6,
            "fetched": 6
        },
        "warmup_status": {
            "total": 3,
            "warmed_count": 3,
            "notwarmed_resources": []
        }
    }
}
```

PR #3872, opened 2021-05-02
Start prewarmup only after collecting resources

engahmeds3ed:
## Description

We need a way to start prewarmup process (RESTAPI and the finish admin notice) only after collecting all resources and add them into the DB.

As in some cases when enabling the `RUCSS` option, it shows a message directly that `prewarmup` finished and this is because the `resourcefetcher` is Async and maybe the `init` is called before filling the resources table with the collected resources which leads to this finish message.

Also, the progress bar will send a request to REST API and it returned total => 3 resources and on the second tick total => 10 because again the Async behavior of `resourcefetcher`

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -126,7 +126,7 @@ class UsedCSS {
 			return false;
 		}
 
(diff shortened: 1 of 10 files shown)


engahmeds3ed:
After merging this PR I will adjust the [RESTAPI PR](https://github.com/wp-media/wp-rocket/pull/3864) to return the status of fetching resources

remyperona (approved):

PR #3886, opened 2021-05-05
Change enqueue position of the delay JS script depending on exclusions field value

remyperona:
## Description

As discussed during stand-up, the position of the delay JS script should change depending on if we have exclusions or not.

This PR updates the code to change the enqueue position depending on the option value:
- In footer if the field is empty
- In header if it's not

## Type of change

## How Has This Been Tested?

- [x] Manual testing with and without values in the exclusions field

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DelayJS/ServiceProvider.php
+++ b/inc/Engine/Optimization/DelayJS/ServiceProvider.php
@@ -41,6 +41,7 @@ class ServiceProvider extends AbstractServiceProvider {
 		$this->getContainer()->share( 'delay_js_subscriber', 'WP_Rocket\Engine\Optimization\DelayJS\Subscriber' )
 			->addArgument( $this->getContainer()->get( 'delay_js_html' ) )
 			->addArgument( rocket_direct_filesystem() )
+			->addArgument( $this->getContainer()->get( 'options' ) )
 			->addTag( 'front_subscriber' );
 	}
 }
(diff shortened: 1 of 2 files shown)


crystinutzaa (approved):

GeekPress:
@Tabrisrp When the script is loaded in the header, it must be the first script. By keeping the default priority on `wp_enqueue_scripts `, I'm afraid it won't necessarily be the case.

remyperona:
@GeekPress I updated the priority to `1`. That should be ok in most cases, but we never know what other plugins or themes do with this.

remyperona:
@wp-media/qa Ready for QA

The delay JS script should now be included early in the `head` when there is a value in the delay JS exclusions field. If it's empty, it remains in the footer.

GeekPress:
> @GeekPress I updated the priority to `1`. That should be ok in most cases, but we never know what other plugins or themes do with this.

The most important is to have our inline script before jQuery when jQuery isn't delayed.

piotrbak (approved):
Works as expected **with the newest version of the script from alpha6 only**:
[testrail-report-175.pdf](https://github.com/wp-media/wp-rocket/files/6436048/testrail-report-175.pdf)

Note, that I haven't tested the compatibilities with 3rd-parties yet

remyperona:
Notion documentation updated to reflect this change.

PR #3888, opened 2021-05-06
Update BerlinDB

engahmeds3ed:
## Description

We created two issues at the BerlinDB repo and both are resolved now so we need to pull the last changes from there.

https://github.com/berlindb/core/issues/97
https://github.com/berlindb/core/issues/98


crystinutzaa (approved):

JJJ:
See: https://github.com/berlindb/core/commit/3f72aebb643d86e7c75c48aad4d6388eaa722e39

PR #3898, opened 2021-05-10
Minify before checking if contents are empty and Minify JS also

engahmeds3ed:
## Description

Here we fix an issue related to if CSS file or JS files have only comments into it, we need to also send `*` in case minified content is empty not the physical one.

We also need to Minify JS contents like what we do with CSS.


tests/Fixtures/inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher/Handle.php
```diff
@@ -107,7 +107,7 @@
 					],
 					[
 						'url'       => 'http://example.org/scripts/script2.js',
-						'content'   => 'var second = "content 2";',
+						'content'   => 'var second="content 2"',
```
crystinutzaa, line 123: Can you add a Fixture for this case?

engahmeds3ed: Done

crystinutzaa (approved):

PR #3899, opened 2021-05-10
Load DelayJS script right after the opening <head>

remyperona:
## Description

After discussions, it was determined that the best place to load the script is right at the start of the document, after the `<head>` opening tag.

This PR reflects that change, using `rocket_buffer` to insert the script in the HTML.

## Type of change

## How Has This Been Tested?

- [x] Manual test
- [x] Updated automated test

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DelayJS/Subscriber.php
+++ b/inc/Engine/Optimization/DelayJS/Subscriber.php
@@ -91,6 +91,8 @@ class Subscriber implements Subscriber_Interface {
 	 * @since 3.9 Hooked on rocket_buffer, display the script right after <head>
 	 * @since 3.7
 	 *
+	 * @param string $html HTML content.
+	 *
 	 * @return string
 	 */
 	public function add_delay_js_script( $html ): string {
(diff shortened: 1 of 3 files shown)


crystinutzaa (approved):

remyperona:
@wp-media/qa Ready for QA

piotrbak:
@Tabrisrp In this version, the script doesn't have any ID. It used to be:
`rocket-delay-js-js-after`

Is that fine?

remyperona:
it's ok we didn't use the ID for this script. And the script itself is inserted after we have transformed the HTML, so it can't be delayed.

piotrbak:
@Tabrisrp We're replacing only the first occurrence of the `<head>` here. I can see a potential problem when there's an HTML comment containing `<head>` above the real `<head>`.
image

Switching to replacing all occurrences of  `<head>` brings different cons. What do you think?

remyperona:
Replacing multiple occurrences will bring more issues than the first one which is an edge case

piotrbak (approved):
Good to go!
[testrail-report-179.pdf](https://github.com/wp-media/wp-rocket/files/6466246/testrail-report-179.pdf)

PR #3906, opened 2021-05-12
RUCSS refactor check_warmup_status

### LINKED ISSUE #3895: RUCSS refactor check_warmup_status
**Describe the bug**
`check_warmup_status()` is doing too much and mixins concerns and makes it a hard time to identify what happens.

Check warmup status is hooked on `init()` , however is bailing out on init if RUCSS is disabled or if the option is empty and fetch_finish_time is not set.

During the latest round of changed , the status of  the option `wp_rocket_prewarmup_stats` was changed. This means that this option will always be kept in the DB and used for metrics. Inside this option, there will be a bool indicator: `allow_optimization` which indicates if RUCSS is enabled or not.

### 1. Modify the  `WP_Rocket\Engine\Optimization\RUCSS\Warmup\Subscriber::check_warmup_status()` 
- Bailout in 2 cases: when RUCSS option is disabled
- when `allow_optimization` is true && `warmup` finished - for warmup finished implement a new method in `Checker.php` which determines this.

### 4. Connect these new methods where are needed 😄 
- On REST API, when the BTN is clicked and RUCSS is manually enabled by the customer, we need to set the warmup status (finish time & allow_optimization)
- When the list of the warmup items finished (no more items in the list) - need to setup the warmup status (finish time & allow_optimization)
- When 1 hour passed after we started the warmup process  - need to setup the warmup status (finish time & allow_optimization)

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

crystinutzaa:
## Description

check_warmup_status() is doing too much and mixins concerns and makes it a hard time to identify what happens.

Check warmup status is hooked on init() , however is bailing out on init if RUCSS is disabled or if the option is empty and fetch_finish_time is not set.

During the latest round of changed , the status of the option wp_rocket_prewarmup_stats was changed. This means that this option will always be kept in the DB and used for metrics. Inside this option, there will be a bool indicator: allow_optimization which indicates if RUCSS is enabled or not.

Fixes https://github.com/wp-media/wp-rocket/issues/3895


inc/Engine/Optimization/RUCSS/Warmup/Status/Checker.php
```diff
@@ -45,31 +45,108 @@ public function __construct( Options $options_api, Options_Data $options, Resour
...
+			$this->set_warmup_status_completed();
+		}
+	}
+
+	/**
+	 * Checks if warmup is completed.
+	 *
+	 * @return void
+	 */
+	public function check_warmup_completed() {
```
iCaspar:
```suggestion
	public function activate_optimization_on_warmup_completion() {
```

inc/Engine/Optimization/RUCSS/Warmup/Status/Checker.php
```diff
@@ -45,31 +45,108 @@ public function __construct( Options $options_api, Options_Data $options, Resour
...
 		}
 
 		if ( time() > strtotime( '+1 hour', (int) $prewarmup_stats['scan_start_time'] ) ) {
-			$this->set_warmup_status_finish_time();
+			$this->set_warmup_status_completed();
+		}
+	}
+
+	/**
+	 * Checks if warmup is completed.
```
iCaspar: This description is misleading. It is activating the optimization and setting the finish time on completion.

inc/Engine/Optimization/RUCSS/Warmup/Status/Checker.php
```diff
@@ -45,31 +45,108 @@ public function __construct( Options $options_api, Options_Data $options, Resour
...
 
+		return ! empty( $prewarmup_stats['scan_start_time'] );
+	}
+
+	/**
+	 * Checks warmup status for resources in the prewarmup process.
+	 *
+	 * @return void
+	 */
+	public function check_warmup_status() {
```
iCaspar:
```suggestion
	public function update_warmup_status_while_has_items() {
```

Again, this is doing something contingently. `check...` implies a bool return on a condition.

inc/Engine/Optimization/RUCSS/Warmup/Subscriber.php
```diff
@@ -130,6 +134,36 @@ public function check_warmup_status() {
...
 	}
 
+	/**
+	 * Checks the is prewarmup process is completed.
+	 *
+	 * @since 3.9
+	 *
+	 * @return void
+	 */
+	public function check_warmup_completed() {
```
iCaspar: Rename to conform to name change in checker.

iCaspar (changes requested):
Mostly requests/suggestions for better method names here.

iCaspar (approved):

PR #3917, opened 2021-05-17
Exclude lazyload styles from being removed from the page with RUCSS

engahmeds3ed:
## Description

We need to exclude lazyload styles from being removed from the page after RUCSS finishes its work on the page.

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -74,6 +74,16 @@ class UsedCSS {
 	private $base_url;
 
 	/**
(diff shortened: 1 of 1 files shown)


iCaspar (approved):

engahmeds3ed:
@wp-media/qa ready for QA

@piotrbak this is the change u saw yesterday, feel free to validate it.

piotrbak:
@engahmeds3ed I'm thinking, in the future, we might want to allow people to exclude the CSS using its attributes within the UI and using a filter, what do you think?

engahmeds3ed:
Yes it will be easier to add a filter here, we can do that in the next minor release for sure

piotrbak:
We're good to go here @engahmeds3ed

PR #3925, opened 2021-05-19
Stop preload while prewarmup is running

engahmeds3ed:
## Description

We found when preload is enabled and the RUCSS option is enabled, that it's a big headache on the client's server so we need to delay preload till prewarmup finishes its work.

## Type of change

## How Has This Been Tested?

- [ ] Manually enable RUCSS when preload is working => should stop preload and start prewarmup
- [ ] Manually enable RUCSS when preload is enabled but not working => should start prewarmup and when finished the preload will start.
- [ ] Manually enable RUCSS while preload is disabled => should start prewarmup only
- [ ] When preload is enabled and manually clicked the button for forcing preload from the top menu or settings dashboard while prewarmup is working => prewarmup will continue and nothing will happen with preload.

# Checklist:

### DIFF
--- a/inc/API/preload.php
+++ b/inc/API/preload.php
@@ -91,6 +91,12 @@ function do_admin_post_rocket_preload_cache() { // phpcs:ignore WordPress.Naming
 		die();
 	}
 
+	$prewarmup_stats = get_option( 'wp_rocket_prewarmup_stats' );
+	if ( empty( $prewarmup_stats['allow_optimization'] ) ) {
+		wp_safe_redirect( wp_get_referer() );
+		die();
+	}
+
 	$preload_process = new FullProcess();
 
 	if ( $preload_process->is_process_running() ) {
--- a/inc/API/preload.php
+++ b/inc/API/preload.php
@@ -92,7 +92,7 @@ function do_admin_post_rocket_preload_cache() { // phpcs:ignore WordPress.Naming
 	}
 
 	$prewarmup_stats = get_option( 'wp_rocket_prewarmup_stats' );
-	if ( empty( $prewarmup_stats['allow_optimization'] ) ) {
+	if ( get_rocket_option( 'remove_unused_css' ) && empty( $prewarmup_stats['allow_optimization'] ) ) {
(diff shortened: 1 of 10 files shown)


inc/API/preload.php
```diff
@@ -91,6 +91,12 @@ function do_admin_post_rocket_preload_cache() { // phpcs:ignore WordPress.Naming
 		die();
 	}
 
+	$prewarmup_stats = get_option( 'wp_rocket_prewarmup_stats' );
+	if ( empty( $prewarmup_stats['allow_optimization'] ) ) {
```
remyperona: We also need to check for the value of `remove_unused_css` here, else we would bail out all the time, even if the option is not enabled

engahmeds3ed: Done

remyperona:
Can you update the checklist items, notably the `How Has This Been Tested?` section, to detail how you have tested this change?

engahmeds3ed:
I will do that tomorrow morning after adjusting the method `maybe_cancel_preload` to also cancel preload when RUCC option is enabled so if preload was working that should stop it.

remyperona (approved):

piotrbak:
@engahmeds3ed I think we're good here. Please confirm that the below testing is enough:
1. Prewarmup working + Preload triggered from the top menu = Preload not activated
2. Prewarmup working +  Activate the preload (it was disabled before) = Preload not activated
3. Preload working + Activate RUCSS = Preload is stopped and Prewarmup starts
4. Preload enabled but not working + Activate RUCSS = Prewarmup starts

After each scenario, the Preload is triggered.
[testrail-report-182.pdf](https://github.com/wp-media/wp-rocket/files/6524510/testrail-report-182.pdf)

engahmeds3ed:
Thanks @piotrbak 
Yes, I think this is enough, just a small thing:
Preload also can be manually triggered from the dashboard button

image

so we need to make sure that point number 1 is valid with this button also.
(Not a big deal, both are sharing the same functionality)

Thanks.

piotrbak (approved):

PR #3928, opened 2021-05-19
Fixes #3478 Enable installation through composer v2

### LINKED ISSUE #3478: Install with composer v2 requirements fail

remyperona:
## Description

This PR updates the WooCommerce dev dependency to 4.9.2, and pulls it from wpackagist instead of the origin repository.

It removes the conflict that was preventing the usage of composer v2 to install/update the plugin.

Fixes #3478

## Type of change

Used `wpackagist-plugin` instead of `woocommerce/woocommerce` to work against the built version of the plugin.

## How Has This Been Tested?

- [x] Automated tests are running correctly after updating `integration/bootstrap.php`
- [x] Installation works with composer v2

# Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,6 @@
 		"phpunit/phpunit": "^7.5",
 		"roave/security-advisories": "dev-master",
 		"szepeviktor/phpstan-wordpress": "^0.7.0",
-		"woocommerce/woocommerce": "^3.9",
 		"wp-coding-standards/wpcs": "^2",
 		"wp-media/background-processing": "^1.3",
 		"wp-media/cloudflare": "^1.0",
@@ -67,6 +66,7 @@
 		"wpackagist-plugin/pdf-embedder": "^4.6",
 		"wpackagist-plugin/simple-custom-css": "^4.0.3",
 		"wpackagist-plugin/spinupwp": "^1.1",
+		"wpackagist-plugin/woocommerce": "^4.9.2",
 		"wpackagist-plugin/wp-smushit": "^3.0"
 	},
 	"autoload": {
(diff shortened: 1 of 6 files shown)


composer.json
```diff
@@ -67,6 +66,7 @@
 		"wpackagist-plugin/pdf-embedder": "^4.6",
 		"wpackagist-plugin/simple-custom-css": "^4.0.3",
 		"wpackagist-plugin/spinupwp": "^1.1",
+		"wpackagist-plugin/woocommerce": "^4.9.2",
```
engahmeds3ed, line 69: why not using the last version of woocommerce?

engahmeds3ed (changes requested):

composer.json
remyperona: When using a more recent version, there was one of the existing WooCommerce test failing (`WooCommerceSubscriber::maybe_revert_uid_for_nonce_actions()` => `testShouldNotValidateNonce()`). I don't know why, so to be able to fix the problem at hand (composer v2 usage), I went with a version that was still working with this test.

We will want to have another look at the above method anyway, since it's likely no longer necessary (see #3929)

engahmeds3ed: aha, got your point.
Thanks.

engahmeds3ed (approved):

PR #3972, opened 2021-05-31
Compatibility with Impreza theme

### LINKED ISSUE #3932: Impreza Theme - JS delay issue after upgrading to 3.9
- Used the search feature to ensure that the bug hasn’t been reported before (maybe related to #3931) ✅ 

**Describe the bug**
Every time I enable js delay within wp-rocket 3.9, I get the following error in the console (see screenshot 1). I get this on every website, where I use wp-rocket and the [Impreza theme ](https://themeforest.net/item/impreza-retina-responsive-wordpress-theme/6434280). The issue did not occur before 3.9 
On some websites, this issue does not cause any visual problems, but on other pages it does. 
I also tried to exclude the file from JS delay, that seems to cause the error (screenshot 2), but that does not change anything. 

**To Reproduce**
1. Enable `Delay Javascript execution` option
2. Use Impreza theme
3. Watch frontend console

**Expected behavior**
The option should not break theme functions, especially if the file is excluded from defer loading. 

**Screenshots**
image

image

**Additional context**

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

GeekPress:
## Description

Automatically exclude an inline JS to have our new Delay JS feature compatible with Impreza theme.

Fixes #3932

## Type of change

# Checklist:


remyperona (approved):

piotrbak:
@Tabrisrp @GeekPress I'm able to reproduce this issue only when `wp-content, wp-includes, js-after|js-before, jquery` are excluded from the feature. When we remove those exclusions and delay all scripts the issue doesn't happen.

This exclusion will be useful only for people that are updating from previous versions. It's not necessary for new installs.

On the other hand, those scripts are only defining the variables, so there's no performance difference. If we decide to exclude it, it's ready to go!

GeekPress:
@piotrbak Indeed, I forgot to mention in the issue that the bug happened when we have the default exclusion list.

We can exclude them safely even if we remove the default exclusions, we can go with the exclusion.

piotrbak (approved):

PR #3985, opened 2021-06-02
Fixes #4122 Compress HTML data before sending it to resource fetcher

### LINKED ISSUE #4122: Pre-warmup fetch resources are stopped in some cases

**Describe the bug**
I was checking an issue with @vmanthos and the issue was that the fetch process doesn't start at all till 1 hour elapsed and RUCSS is enabled.
After debugging on the customer's site, I found that the background process that we use on fetching resources from scanned pages failed saving the batches into the options table, exactly here on this line:

https://github.com/wp-media/wp-rocket/blob/3250a7ecdefeb1021bc780f2015566180195c9e7/inc/classes/dependencies/wp-media/background-processing/wp-background-process.php#L101

so all what I did is to base64_encode the resource content here: https://github.com/wp-media/wp-rocket/blob/a0882549879826b9fe5e18ef620dc705d497efba/inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php#L179

and then use base64_decode here
https://github.com/wp-media/wp-rocket/blob/b7b4424f9e85cb389b489c481c93a336dfcccb4[…]inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcherProcess.php

like this
```
$resource['content'] = base64_decode( $resource['content'] );
```

this is exactly what I did here on this PR (I was researching some time ago)
https://github.com/wp-media/wp-rocket/pull/3985/files

so maybe encode/decode make size lower (theoretically it should reduce 30% of its size)

**To Reproduce**

I can't find a pattern to reproduce the issue at any host.

**Expected behavior**
The fetch process should perform properly.

**Additional context**
- I found a similar issue with our test site reported by @Mai-Saad but the difference is that only one page was stopped fetching not all pages then solved it also with encode/decode.
- I think we will not need this encode/decode mechanism if we solved this issue: https://github.com/wp-media/wp-rocket/issues/4072 so the resource's content will not be added to the batch at all.
https://wp-media.slack.com/archives/G01AMD70N7P/p1625049019275000

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
## Description

This PR add a compress/decompression to the HTML data sent to the resource fetcher, to reduce errors related to data being to big to send through a POST request.

It uses gzip compression, in the same way we use it to compress cached/minified files.

## Type of change

- [x] Enhancement

Yes

## How Has This Been Tested?

- [x] Tested the pre-warmup locally, worked as expected

# Checklist:


mostafa-hisham (approved):

remyperona:
@wp-media/php I updated this PR to compress/decompress the HTML data sent to the resource fetcher.

As we are no longer storing any content inside the background processes queue, this should be the last thing that might create stuck issue with the pre-warmup, because of the limit on the POST size configuration.

Please review 🙏🏼

mostafa-hisham (approved):

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
```diff
@@ -279,4 +281,49 @@ public function get_zones() : array {
...
+	/**
+	 * Compress the HTML data before sending it to the resources fetcher
+	 *
+	 * @since 3.9.2
+	 *
+	 * @param string $data HTML data.
+	 *
+	 * @return string
+	 */
+	public function compress( string $data ): string {
```
mostafa-hisham, line 293: i think we should bail out if string data is empty

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
```diff
@@ -279,4 +281,49 @@ public function get_zones() : array {
...
+	 * Decompress the HTML data received by the resources fetcher
+	 *
+	 * @since 3.9.2
+	 *
+	 * @param string $data Compressed HTML data.
+	 *
+	 * @return string
+	 */
+	private function decompress( string $data ): string {
+		if ( ! function_exists( 'gzdecode' ) ) {
```
mostafa-hisham, line 321: i think we should bail out if string data is empty

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
remyperona: Done in https://github.com/wp-media/wp-rocket/pull/3985/commits/7d8f945523f7ac7bd0e84fef1e638f0901b354dd

inc/Engine/Optimization/RUCSS/Warmup/ResourceFetcher.php
remyperona: Done in https://github.com/wp-media/wp-rocket/pull/3985/commits/7d8f945523f7ac7bd0e84fef1e638f0901b354dd

mostafa-hisham (approved):

Mai-Saad (commented):
Done measures here https://docs.google.com/spreadsheets/d/1bIrRtsvQrYBytseOH8bPznq0G0excSCgVtvbrvtiyWg/edit#gid=730116432 sheet 3985_new .
notes:
- average time and CPU utilization > trunk
- on Mega/Avada => stuck at fetch twice when network is ?, tried to switch between NW slow 3G and offline to reproduce it but it worked fine

remyperona:
To re-iterate, this PR is focused on fixing the stuck issue on the scanning phase (getting the HTML of each page and passing it to the resource fetcher).

Anything after that (the fetching phase for the resources) is not modified by this PR.

Mai-Saad (commented):
@Tabrisrp I think we can open another issue for stuck at no network case .. Then we can merge this PR.

Mai-Saad (approved):
Tried to reproduce the stuck when the network is offline then back but it works normally on both mobile and laptop. (can't memic the exact same case of the network ?)

Issue #3945, opened 2021-05-24
Add incompatibility notice when WP Meteor is activated and our Delay JS too


GeekPress:
**Is your feature request related to a problem? Please describe.**
WP Meteor is a plugin that delays all JS scripts.

This plugin will be in conflict with our Delay JS when it's activated. 

**Describe the solution you'd like**
Add WP Meteor to our list of conflicted plugins when our Delay JS option is activated.

aguidrevitch:
Any examples of incompatibility issues? Or, how can I make WP Meteor compatible?

GeekPress:
WP Meteor is compatible WP Rocket, but not when our Delay JS is turned on. The notice is displayed only if our Delay JS option is turned. 

It doesn't make any sense to have two scripts that will delay all JS and by using different solutions by the way.

aguidrevitch:
> WP Meteor is compatible WP Rocket, but not when our Delay JS is turned on. The notice is displayed only if our Delay JS option is turned.
> 
> It doesn't make any sense to have two scripts that will delay all JS and by using different solutions by the way.

Absolutely no sense to have both. But giving 'Disable other plugin'  is, umm, let's say, a one way suggeston. I would have to add a notice about the other option - to disable WP Rocket's 'Delay Javascript Execution' and use WP Meteor instead.

GeekPress:
> But giving 'Disable other plugin' is, umm, let's say, a one way suggeston.

It's a way to avoid our users to have problem.

Why are doing that?
Because we already had tickets about users using both at the same. Our role is to « protect » our users from conflicts.

>  I would have to add a notice about the other option - to disable WP Rocket's 'Delay Javascript Execution' and use WP Meteor instead.

If you are adding a notice too, that's going to be very confused for the users... 

I don't really understand what is the problem on your side. We are doing this with all conflicted plugins and it has never been a problem.

aguidrevitch:
> I don't really understand what is the problem on your side. We are doing this with all conflicted plugins and it has never been a problem.

Because I'm the author of WP Meteor

GeekPress:
> Because I'm the author of WP Meteor

And what is the problem?

aguidrevitch:
> Because we already had tickets about users using both at the same. Our role is to « protect » our users from conflicts.
My role is to protect my users as well. My users come to me asking about the message and I have to explain each of them that only 'Delay Javascript Execution' is conflicting option, not the whole plugin.

[...]

PR #3986, opened 2021-06-02
3.9.0.1

remyperona:
- Enhancement: Prevent showing the Image Optimization tab on settings page when white label is enabled (#3120)
- Enhancement: Prevent showing the Beacon & ask support buttons when white label is enabled (#2876)
- Enhancement: Update our new delay JS script to fix some issues discovered (#3934)
- Bugfix: Enable plugin installation through composer when using composer v2 (#3478)
- Bugfix: Correctly preload fonts when Optimize CSS Delivery is enabled (#3927)
- Bugfix: Remove charset=utf-8 from application/json content type (#3645)
- 3rd party compatibility: Display a warning when WP Meteor is enabled while delay JS is enabled in WP Rocket (#3945)
-  3rd party compatibility: Add automatic exclusions from delay JS (#3932, #3941, #3950, #3959)


piotrbak:
@Tabrisrp We have a problem with DelayJS field when updating. Steps to reproduce:
1. WP Rocket on 3.9
2. Enable Delay JS, add some custom things there
3. Update to 3.9.0.1, the field gets overridden by default list

On the other hand, we need to make sure that the default list is working correctly for people that are updating from the `<3.9`

remyperona:
@piotrbak can you retry after the last commit?

PR #4049, opened 2021-06-15
delay js exclude scripts with invalid type

### LINKED ISSUE #3944: Delay JS - Automatically exclude scripts with invalid type

An invalid type will lead to a freeze during the delay JS process.

To fix that problem, we must not transform these scripts to be delayed.

Here a list of valid JS type:

```
'text/javascript',
'text/x-javascript',
'text/ecmascript',
'text/jscript',
'application/javascript',
'application/x-javascript',
'application/ecmascript',
'application/jscript'
'module'
```

Detect and exclude automatically scripts with an invalid type.

mostafa-hisham:
## Description

delay js- exclude scripts with invalid type

Fixes #3944

## Type of change

No
## How Has This Been Tested?

- [x] Local testing

# Checklist:


remyperona (changes requested):
We can remove the conditional for `ld+json` at lines 171-173 now, as the new one will skip those too

remyperona (approved):

piotrbak:
@Tabrisrp @mostafa-hisham How detailed we should be while testing this regex itself?

For example, it accepts `type="module2345"` not sure if that's an issue though.

remyperona:
@piotrbak I updated the regex a bit to avoid this case. But mainly if we validate that the valid script types are all correctly working, anything else can be treated as an edge case.

piotrbak (approved):
Looks good!
[testrail-report-206.pdf](https://github.com/wp-media/wp-rocket/files/6766220/testrail-report-206.pdf)

PR #4083, opened 2021-06-22
Add auto-exclusion for Smush Native Lazyload script

GeekPress:
## Description

Smush inserts two different JS files for its Lazyload option. We are removing the classic version, but we missed the JS file when the Native Lazyload option is turned on.

## Type of change

## How Has This Been Tested?

- [x] On customer website by adding the exclusions using the field
- [x] On smashingcoding by updating the plugin code and checking the exclusion worked properly

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DelayJS/HTML.php
+++ b/inc/Engine/Optimization/DelayJS/HTML.php
@@ -33,7 +33,7 @@ class HTML {
 		'window.\$us === undefined',
 		'js-extra',
 		'fusionNavIsCollapsed',
-		'/assets/js/smush-lazy-load.min.js',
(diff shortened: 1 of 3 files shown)


engahmeds3ed (approved):

Mai-Saad (commented):
- Working as expected with smush with/without native LL enabled, with/without default exclusion (1st image always displayed)
- Note: if LL was disabled in smush, Donn we need to display a clear cache msg of WPR as we will still exclude the script from delay js till cache is cleared? (not related to this PR)
- The same issue of not displaying 1st image is happening with shortpixel, will add another issue for it if it wasn't there. => already exists in doc so no need for issue to this

Mai-Saad (approved):

PR #4094, opened 2021-06-24
Fixes #3690 Correctly clear GoDaddy's Varnish cache

### LINKED ISSUE #3690: GoDaddy VIP - Varnish cache isn't cleared in sync with WP Rocket's

**Describe the bug**

On GoDaddy VIP, the Varnish's cache isn't cleared, when WP Rocket's cache is cleared.

This is because the following check is returning `false`:
https://github.com/wp-media/wp-rocket/blob/2be5659b88d0b6e279a239adbbb8e9ea6208698a/inc/3rd-party/hosting/godaddy.php#L104

The `vip()` method is defined in the `Helpers` trait in:
`/gd-system-plugin/includes/trait-helpers.php`

**To Reproduce**

1. Go to a GoDaddy VIP site.
2. Check the `age` header for the homepage.
3. Clear WP Rocket's cache.
4. Visit the homepage and check the `age` header again. It will have increased, instead of being `0`, or a value lower than step 2.

**Expected behavior**

Cache clearing should be done in sync.

**Additional context**

On the customer's site, I tried the following:
```PHP
if( ! ( defined( 'GD_VIP' ) ? (string) GD_VIP : false ) ){
	return;
}
```

This is what the `vip()` method is doing in the first place.

This cleared Varnish's cache when WP Rocket's cache was cleared.

**Related tickets**
* https://secure.helpscout.net/conversation/1453393326/247480/
* https://secure.helpscout.net/conversation/1433880108/241700/

**Slack Conversation**
https://wp-media.slack.com/archives/C43T1AYMQ/p1614957546110100

A copy of GoDaddy's mu-plugin can be found [here](https://wp-media.slack.com/archives/C43T1AYMQ/p1614958304111600?thread_ts=1614957546.110100&cid=C43T1AYMQ).

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

mostafa-hisham:
## Description

GoDaddy VIP, the Varnish's cache isn't cleared, when WP Rocket's cache is cleared.

Fixes #3690 

## Type of change

No

## How Has This Been Tested?

- [x] create new unit and integration tests for it
- [ ] we should test it on GoDaddy VIP site to check if the solution will work or not

# Checklist:


inc/3rd-party/hosting/godaddy.php
```diff
@@ -101,7 +101,7 @@ function rocket_clean_home_godaddy( $root, $lang ) {
  * @return void
  */
 function rocket_godaddy_request( $method, $url = null ) {
-	if ( ! method_exists( 'WPass\Plugin', 'vip' ) ) {
+	if ( ! method_exists( 'WPaas\Plugin', 'vip' ) ) {
```
engahmeds3ed, line 104: remove this file completely.

inc/ThirdParty/Hostings/Godaddy.php
```diff
@@ -0,0 +1,166 @@
...
+	/**
+	 * Call the Varnish server to purge the home with GoDaddy.
+	 *
+	 * @since 3.9.1
+	 *
+	 * @param string $root root URL.
+	 * @param string $lang language code.
+	 * @return void
+	 */
+	public function clean_home_godaddy( $root, $lang ) {
```
engahmeds3ed: remove this `$root` argument, not used at any place.

inc/ThirdParty/Hostings/Godaddy.php
```diff
@@ -0,0 +1,166 @@
...
+
+		$this->godaddy_request( 'PURGE', $home_url );
+		$this->godaddy_request( 'PURGE', $home_pagination_url );
+	}
+
+
+	/**
+	 * Perform the call to the Varnish server to purge
+	 *
+	 * @since 2.9.5
```
engahmeds3ed: add another since for our current version

tests/Integration/inc/ThirdParty/Hostings/Godaddy/cleanDomainGodaddy.php
```diff
@@ -0,0 +1,35 @@
...
+		parent::setUp();
+		add_filter( 'pre_http_request', [ $this, 'mock_response' ] );
+	}
+
+	protected function tearDown(): void {
+		parent::tearDown();
+		remove_filter( 'pre_http_request', [ $this, 'mock_response' ]);
+	}
+
+	public function testShouldDoBanRequest( ) {
```
engahmeds3ed: I think you need a data provider here to use fixture to verify some cases like variation of root, lang and home_url and change them to be subdomain home or WP in sub-directory

tests/Integration/inc/ThirdParty/Hostings/Godaddy/cleanHomeGodaddy.php
```diff
@@ -0,0 +1,37 @@
...
+
+	public function setUp(): void {
+		parent::setUp();
+		add_filter( 'pre_http_request', [ $this, 'mock_response' ] );
+	}
+
+	protected function tearDown(): void {
+		parent::tearDown();
+		remove_filter( 'pre_http_request', [ $this, 'mock_response' ]);
+	}
```
engahmeds3ed: I can see those lines are common between all GoDaddy tests, what do you think if we created a class and extended from it?

tests/Unit/inc/ThirdParty/Hostings/Godaddy/godaddyVarnishField.php
```diff
@@ -0,0 +1,31 @@
...
+		parent::setUp();
+		Functions\stubTranslationFunctions();
+	}
+
+	/**
+	 * @dataProvider configTestData
+	 */
+	public function testShouldDoExpected( $settings, $expected ) {
+
+		$godaddy = new Godaddy();
```
engahmeds3ed: I think for this instantiation you need to pass `vip_url` (here and in all Godaddy unit tests)

engahmeds3ed (changes requested):

inc/ThirdParty/Hostings/Godaddy.php
mostafa-hisham: done

inc/3rd-party/hosting/godaddy.php
mostafa-hisham: done

inc/ThirdParty/Hostings/Godaddy.php
mostafa-hisham: we cannot remove "root" because the action use two arguments in other areas of the code

```
wp-rocket/inc/functions/files.php line 168
do_action( 'before_rocket_clean_home', $root, $lang );
```

tests/Integration/inc/ThirdParty/Hostings/Godaddy/cleanDomainGodaddy.php
mostafa-hisham: the root is not used i this function and the URL is provided inside the function it self "home_url()"

inc/ThirdParty/Hostings/Godaddy.php
```diff
@@ -0,0 +1,191 @@
...
+	private $vip_url;
+
+	/**
+	 * Godaddy constructor.
+	 *
+	 * @param bool|string $vip_url Godaddy vip url.
+	 */
+	public function __construct( $vip_url = false ) {
+		$this->vip_url = method_exists( '\WPaas\Plugin', 'vip' ) ? \WPaas\Plugin::vip() : $vip_url;
+	}
```
remyperona: I'm not sure about the point of passing the `$vip_url` as a parameter of the constructor. When we instantiate the class, we don't pass anything to it, so the value will never be different than `false` or the value from the `vip()` method?

tests/Unit/inc/ThirdParty/Hostings/HostResolver/getHostService.php
```diff
@@ -34,10 +34,6 @@ public function testShouldReturnExpectedValue( $expected ) {
 			case 'spinupwp':
 				putenv( 'SPINUPWP_CACHE_PATH=/wp-content/spinupwp-cache/' );
 				break;
-			case 'wpengine':
-				require_once WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/ThirdParty/Hostings/WPEngine/wpe_param.php';
-				require_once WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/ThirdParty/Hostings/WPEngine/WpeCommon.php';
-				break;
```
remyperona, line 40: Why do we remove those?

tests/Unit/inc/ThirdParty/Hostings/HostSubscriberFactory/getSubscriber.php
```diff
@@ -47,10 +47,6 @@ public function testShouldReturnSubscriber( $host, $expected ) {
 			case 'spinupwp':
 				putenv( 'SPINUPWP_CACHE_PATH=/wp-content/spinupwp-cache/' );
 				break;
-			case 'wpengine':
-				require_once WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/ThirdParty/Hostings/WPEngine/wpe_param.php';
-				require_once WP_ROCKET_TESTS_FIXTURES_DIR . '/inc/ThirdParty/Hostings/WPEngine/WpeCommon.php';
-				break;
```
remyperona, line 53: Why do we remove those?

remyperona (changes requested):

inc/ThirdParty/Hostings/Godaddy.php
engahmeds3ed: We did that to make the code testable so you may notice that the test is passing the `vip_url` but in the main code we will keep it with the default value of false and it will call this VIP method to get the vip_url that will be used here

https://github.com/wp-media/wp-rocket/blob/935d7362a6f93523e42cbc25f9d504c605be2a4b/inc/ThirdParty/Hostings/Godaddy.php#L141

The old code was bailing out if it didn't find the `vip` method here:
https://github.com/wp-media/wp-rocket/blob/cf8b0b647456783d62d9f180b8e71643fe75d763/inc/3rd-party/hosting/godaddy.php#L104-L106

so to create unit tests for that we need to create a mock class (Stub) as we did with integration tests here:

https://github.com/wp-media/wp-rocket/blob/935d7362a6f93523e42cbc25f9d504c605be2a4b/tests/Fixtures/inc/ThirdParty/Hostings/Godaddy/Plugin.php

but if we did that, this will affect the other tests as we require this class so always GoDaddy will be detected as the current host, and no way to change it.

so we have two options here:-

1. Don't add unit tests for GoDaddy methods (just the integration tests) and remove this vip_url argument.
2. keep unit tests and keep this argument.

What do you think?

tests/Unit/inc/ThirdParty/Hostings/HostResolver/getHostService.php
engahmeds3ed: For unit tests, those files are useless, you will find that no test fails after removal.

tests/Unit/inc/ThirdParty/Hostings/HostSubscriberFactory/getSubscriber.php
engahmeds3ed: For unit tests, those files are useless, you will find that no test fails after removal.

tests/Unit/inc/ThirdParty/Hostings/Godaddy/removeHtmlExpireGoddady.php
```diff
@@ -0,0 +1,32 @@
...
+use Brain\Monkey\Functions;
+use WP_Rocket\Tests\Unit\TestCase;
+use WP_Rocket\ThirdParty\Hostings\Godaddy;
+
+class Test_removeHtmlExpireGoddady extends TestCase {
+
+	public function setUp() : void {
+		parent::setUp();
+		Functions\stubTranslationFunctions();
+	}
```
remyperona: Is this necessary for this test?

inc/ThirdParty/Hostings/Godaddy.php
remyperona: Ok I understand the need now 👍🏼 

What about having a setter method instead? `set_vip_url()`

inc/ThirdParty/Hostings/Godaddy.php
```diff
@@ -0,0 +1,191 @@
...
+
+class Godaddy implements Subscriber_Interface {
+	use ReturnTypesTrait;
+
+	/**
+	 * Godaddy vip url
+	 *
+	 * @var bool|string
+	 */
+	private $vip_url;
```
remyperona:
```suggestion
	 * Godaddy vip url
	 *
	 * @var string
	 */
	private $vip_url = '';
```

inc/ThirdParty/Hostings/Godaddy.php
```diff
@@ -0,0 +1,191 @@
...
+	 * @source WPaaS\Cache
+	 *
+	 * @param string $method can be BAN or PURGE.
+	 * @param string $url URL to purge.
+	 *
+	 * @return void
+	 */
+	private function purge_request( $method, $url = null ) {
+
+		if ( false === $this->vip_url ) {
```
remyperona:
```suggestion
		if ( empty( $this->vip_url ) ) {
```

inc/ThirdParty/Hostings/Godaddy.php
```diff
@@ -0,0 +1,191 @@
...
+	 *
+	 * @since 3.9.1
+	 * @source WPaaS\Cache
+	 *
+	 * @param string $method can be BAN or PURGE.
+	 * @param string $url URL to purge.
+	 *
+	 * @return void
+	 */
+	private function purge_request( $method, $url = null ) {
```
remyperona:
```suggestion
	private function purge_request( $method, $url = '' ) {
```

remyperona (changes requested):

remyperona (approved):

piotrbak (changes requested):
@mostafa-hisham The cache is not updated when updating the post (nor the post itself, nor the homepage). Going to investigate this a bit

piotrbak:
@mostafa-hisham 

In order to get it to work and be able to clean the cache when updating posts:
1. Changed the method to `BAN` [here](https://github.com/wp-media/wp-rocket/blob/011e2a8ea412aac2d36996ef9872c82dfb895981/inc/ThirdParty/Hostings/Godaddy.php#L107)
2. Removed the trailing slash from the `url` [here ](https://github.com/wp-media/wp-rocket/blob/011e2a8ea412aac2d36996ef9872c82dfb895981/inc/ThirdParty/Hostings/Godaddy.php#L152)

What's interesting, cleaning the homepage with the `clean_home` function works with both `BAN` and `PURGE`. Cleaning the specific posts only with `BAN`. 

Sending you website details on PM

@Tabrisrp Is there any specific reason why we are using `BAN` for clearing the whole website's cache and `PURGE` for different actions?

piotrbak:
@Tabrisrp @mostafa-hisham Cache clearing working now. However, the problem with Varnish is when we're clearing the homepage, we're basically clearing the whole cache then because it clears everything recursively:

```
[2021-07-28 10:18:45] URL: http://160.153.138.71
[2021-07-28 10:18:45] Method: BAN
[2021-07-28 10:18:45] Host: dxr.095.myftpupload.com
```
or

```
[2021-07-28 10:16:55] URL: http://160.153.138.71
[2021-07-28 10:16:55] Method: PURGE
[2021-07-28 10:16:55] Host: dxr.095.myftpupload.com
```

Are clearing the whole cache. That kind of request is sent while updating the post or using the `Purge this URL` button on frontend while being on homepage. This is a Varnish thing rather than our compatibility though. What do you think?

piotrbak (approved):
Looks good, there's a problem with excessive cache purging while clearing the homepage, but it's related to the Varnish, not this update specifically:
[testrail-report-228.pdf](https://github.com/wp-media/wp-rocket/files/6894050/testrail-report-228.pdf)

PR #4110, opened 2021-06-28
Closes #2772 Preload & load Google Fonts asynchronously

### LINKED ISSUE #2772: Load Google Fonts asynchronously
We are missing one enhancement to our Google Fonts optimization feature.

Like it's described on this post, we could load asynchronously the Google Fonts:
https://csswizardry.com/2020/05/the-fastest-google-fonts/

This method is also going to be implemented to load the CSS files when the « Optimize CSS delivery » option is activated:
https://github.com/wp-media/wp-rocket/issues/2658

mostafa-hisham:
## Description

Load Google Fonts asynchronously using preload
https://csswizardry.com/2020/05/the-fastest-google-fonts/

Fixes #2772 

## Type of change

No but this was missing

`<link rel="stylesheet" href="$CSS&display=swap"  media="print" onload="this.media='all'" />
`
## How Has This Been Tested?

- [x] test locally using ggole fonts
- [x] unit and integration tests

# Checklist:


crystinutzaa (approved):

remyperona (changes requested):
One thing that is missing is the removal of the conditional checking if `1 === $num_fonts`.

We want the process to continue even if there is only 1 font

mostafa-hisham:
@Tabrisrp
I removed the condition but we have a problem in this case for example
the the original <link> has attr ` id="dt-web-fonts-css"`

what we should do in this case `single font with attrs` when we combine fonts we remove all the attrs of the links ad combine them in one <link>?

https://github.com/wp-media/wp-rocket/blob/bcdffdf615d5132233a86c3d4eecdf7b0d2a813a/tests/Fixtures/inc/Engine/Optimization/GoogleFonts/CombineV2/optimize.php#L29-L52

remyperona:
@wp-media/php In the latest commits:
- I updated the `AbstractGFOptimization` class to remove the dependency on `AbstractOptimization`, as it was done only to use a small subsets of methods, so the dependency didn't really make sense
- In the same class, I added an abstract declaration for the `optimize()` method, and added a new `get_optimized_markup()` that is returning the HTML with everything needed to preload + async the fonts
- In `Combine` and `CombineV2`, I use the RegexTrait for the `hide_comments()` and `find()` methods
- I also updated the classes to use the new method, and changed `get_combined_tag()` to `get_combined_url()`
- Finally, I updated the fixtures and tests to clean them up and make sure they're up to our latest standards

remyperona (approved):

iCaspar (approved):
Nice!

remyperona:
@wp-media/qa Ready for QA

One important change relative to your potential existing test cases: we no longer bailout when there is only one google font on the page, to make sure we fully optimize it.

Also for the expectations, you should see:
- a preload link
- a stylesheet link with the media print value
- a noscript element with a stylesheet link inside

Mai-Saad (approved):
Working as expected.
[testrail-report-220.pdf](https://github.com/wp-media/wp-rocket/files/6829921/testrail-report-220.pdf)

PR #4132, opened 2021-07-06
Fixes #3282 Improve LiteSpeed compatibility

### LINKED ISSUE #3282: LiteSpeed compatibility
**Describe the bug**
In our LiteSpeed compatibility, we're looking for the [HTTP_X_LSCACHE](https://github.com/wp-media/wp-rocket/blob/e40b824350a165cf52f2285a681d5b6796ed5555/inc/classes/subscriber/third-party/Hostings/class-litespeed-subscriber.php#L22) server variable, while it seems that the [X_LSCACHE](https://docs.litespeedtech.com/lscache/lscwp/admin/#testing) is the correct one. I also tested the cache purging on one of the hostings and clearing the whole cache worked as expected but the partial purge nor single URL purge doesn't seem to work.

**To Reproduce**
1. Make sure that the LiteSpeed cache is enabled
2. Check the PHP INFO file
3. Confirm that the X-LSCACHE is there while HTTP_X_LSCACHE is not
4. Change the value in our code to X-LSCACHE
5. Visit the homepage in the incognito mode and confirm with the headers that it's cached 
6. Clear the whole cache and confirm with the headers that the cache was purged
7. Visit the single post page and confirm with the headers that it's cached 
8. Update this post in the dashboard and confirm with the headers that the cache of this post wasn't purged

**Expected behavior**
We should be able to detect LiteSpeed cache correctly and clear the whole cache and also single post's cache as expected.

**Additional context**
The above was tested on the Raiola Network hosting, I was able to clear the single post's cache by sending PURGE requests, just like we're doing with Varnish:
https://docs.litespeedtech.com/lscache/basics/#curl-command

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

mostafa-hisham:
## Description

I fixed the problem and changed the class to the new structure
the oe thing that is still missing is to make the old class deprecated in 3.9.php that I created in Godaddy VIP issue

Fixes #3282 

## Type of change

No
## How Has This Been Tested?

- [x]  uint and integration test
- [ ] We need to test this on LiteSpeed server to make sure that it will work 

# Checklist:


remyperona (approved):

piotrbak:
@mostafa-hisham The new approach doesn't seem to work. There are two things:
1. There's a lot of confusion in the ticket. The correct value in the array is `X-LSCACHE` not `X_LSCACHE`
2. The new approach with `wp_headers` doesn't seem to work at all. I replaced https://github.com/wp-media/wp-rocket/blob/662e44488b32ed7ba4ab97d293c42a69a29b2239/inc/ThirdParty/Hostings/LiteSpeed.php#L128 with https://github.com/wp-media/wp-rocket/blob/e40b824350a165cf52f2285a681d5b6796ed5555/inc/classes/subscriber/third-party/Hostings/class-litespeed-subscriber.php#L114-L118 and it worked correctly.

I'll send you access to the LiteSpeed server on Slack, if you need to test the new approach with headers

piotrbak (changes requested):
@mostafa-hisham Cleaning the posts' cache doesn't work. Could you recheck now on the website? Branch is uploaded there

piotrbak:
Tested and it seems to be working fine. Waiting for CR, as there's a new approach present

inc/ThirdParty/Hostings/LiteSpeed.php
```diff
@@ -0,0 +1,148 @@
...
+	 *
+	 * @since  3.9.2
+	 *
+	 * @return void
+	 */
+	public function litespeed_send_headers() {
+		if ( empty( $this->headers ) ) {
+			return;
+		}
+		foreach ( $this->headers as $x ) {
```
iCaspar:
```suggestion
		foreach ( $this->headers as $header ) {
```

inc/ThirdParty/Hostings/LiteSpeed.php
```diff
@@ -0,0 +1,148 @@
...
+			@header( "X-LiteSpeed-Purge: {$x}", false ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
```
iCaspar:
```suggestion
			@header( "X-LiteSpeed-Purge: {$header}", false ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
```

iCaspar (approved):

inc/ThirdParty/Hostings/LiteSpeed.php
mostafa-hisham: Fixed

PR #4139, opened 2021-07-07
Fixes #4072 Prevent high CPU load from resource fetcher

### LINKED ISSUE #4072: High CPU usage related to rocket_saas_warmup async process
**Describe the bug**
We’re getting some reports about high CPU usage with RUCSS enabled, notably on Kinsta/WP Engine, and both pointing to the asynchonous process of the resource fetcher.

As a reminder, this process is triggered asynchronously on rocket_buffer, so running on uncached pages during the caching process. the AJAX action is `rocket_saas_warmup`

From their logs, some of those calls are running for too long, up to 6 min.

Looking at the code, during this call, we are:
- parsing the HTML for CSS/JS files
- for each file, we are getting the URL, path, content, type (and media value for CSS)
- For the content itself, we are getting it from the local file, or from the remote file if it’s an external path, and then we are Minifying it
- We finally send those to the background process which sends to warmup
- 
I believe what would be causing the issue here is the content fetching + minifying step. Depending on the number of files found, it means a loop with a lot of processing in it.

I’m thinking this specific part should be delegated to the background processing task instead, as it can be running for longer in the background, and it has time and memory limit control.

We can investigate how to move `get_url_details()` from this class to the background processing class.

**Expected behavior**
No high CPU usage during this process

**Screenshots**

image003

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

engahmeds3ed:
## Description

We splitted the processing time between `ResourceFetcher` and `ResourceFetcherProcess` as follows:-

### ResourceFetcher Responsibility
1. Collect CSS/JS resources from the HTML.
2. Get URL paths for each resource.
3. Send URL and path to the `ResourceFetcherProcess` background process.

### ResourceFetcherProcess Responsibility
1. Receive CSS/JS resources (URL, path).
2. Get the content of the file based on its type (internal or external)
3. Minify the content.
4. Check the file on the local database.
5. Call the SaaS warmup.

So we moved the content-related methods to be on `ResourceFetcherProcess`

Fixes #4072

## Type of change

Yes exactly the same

## How Has This Been Tested?

- [x] Tested pre-warmup locally
- [x] Tested normal warmup locally
- [x] Adjust tests to pass.

# Checklist:


remyperona (approved):

engahmeds3ed:
@wp-media/qa 

To test this change we need to make sure of the following:-
- [ ] With fresh installation, try pre-warmup by enabling the RUCSS option for the first time then check the DB resources table and make sure that resources are added to this table with proper content.
- [ ] Fetching process should be faster than before.
- [ ] If WordOps (the infra that you use at our QA sites) has a CPU usage meter you can use it to test that after this change the CPU usage should be lower and also bandwidth.
- [ ] We should also test all types of resources' URLs (relative, absolute)

remyperona (approved):

Mai-Saad (commented):
Testing is done.
With the latest recheck of performance after external files change => Total Fetch time increased (compared to trunk) while CPU utilization decreased

Notes:
-  Recommend team check the time in red cases
-  Fetch time on trunk decreased a lot compared to 1st measurements
[testrail-report-217.pdf](https://github.com/wp-media/wp-rocket/files/6822857/testrail-report-217.pdf)
https://docs.google.com/spreadsheets/d/1bIrRtsvQrYBytseOH8bPznq0G0excSCgVtvbrvtiyWg/edit?usp=sharing

remyperona:
Fetch time increase can make sense since we're now doing more processing later instead of upfront, so each individual fetch job will take longer. The difference is small, which is acceptable.

The reduction in CPU load is more important in that case.

Mai-Saad (approved):

PR #4174, opened 2021-07-19
Fixes #4165 Exclude EWWW WebP rewrite inline script from delay JS

### LINKED ISSUE #4165: Delay JS: Exclude EWWW inline script for webp
EWWW image optimizer has an option for JS WebP Rewriting: https://jumpshare.com/v/9aPlSgSZ0qDLqA2ZPKrK
There is an inline script for this which must be excluded from Delay JS so that images will load correctly. When not excluded it can take multiple interactions, scrolling up and down the page for example, to display correctly. 

The script: https://snippi.com/s/d8h3vh3

In my case I used this exclusion: `ewwwWebPInit`

Ticket: https://secure.helpscout.net/conversation/1565708096/277916?folderId=377611

remyperona:
## Description

This script from the EWWW plugin is preventing correct display when delayed, as reported in the issue.

Fixes #4165

## Type of change

No

## How Has This Been Tested?

- [x] Added fixture test with the corresponding inline script

# Checklist:


iCaspar (approved):

piotrbak:
@Tabrisrp I can't reproduce the issue in the same way as @webtrainingwheels... In my case, when the EWWW LazyLoad feature is enabled, we need to exclude the following files in order to load the WebP correctly without the delay:
```
/wp-content/plugins/ewww-image-optimizer/includes/lazysizes-pre.js
/wp-content/plugins/ewww-image-optimizer/includes/lazysizes.js
/wp-content/plugins/ewww-image-optimizer/includes/lazysizes-post.js
```

When the LazyLoad is disabled, I still can not see the `ewwwWebPInit` in the source and the WebP rewrite is handled with this file:
```
/wp-content/plugins/ewww-image-optimizer/includes/load_webp.js
```

**EDIT** the `load_webp.js` script is the same as inline `ewwwWebPInit` one. In this case we should exclude everything I believe to cover all scenarios.

remyperona:
@piotrbak I can add the lazysizes patterns to the other PR, and update this one to also exclude the external script, what do you think?

piotrbak:
@Tabrisrp Sounds good! We'll add it to 3.9.2 too, right?

remyperona:
Yes both are planned for the same version

remyperona:
@piotrbak Updated, including the lazysizes changes from the other PR now merged in develop. So you should be able to test all cases directly in this PR.

piotrbak:
@Tabrisrp In the latest EWWW release (1 hour ago), they changed the filename from `load_webp.js` to `load-webp.js` 😅 😞  We should exclude both of them.

remyperona:
@piotrbak updated

piotrbak (approved):
Works perfectly now, thank you!

PR #4175, opened 2021-07-19
Fixes #4149 Exclude WPML browser redirect script from delay JS

### LINKED ISSUE #4149: Delay JS - WPML redirect according to browser language issue
- Made sure you’re on the latest version: YES (3.9.0.5)

**Describe the bug**
When using delay JS & having WPML automatically redirect to browser language (if it exists & isn't the default site language): 

The website loads all elements that don't require user interaction first. Once the user moves his/her mouse the page will load fully in the main language (dutch for my site). After it is fully loaded the website will detect it is actually a french speaking client. This creates a visual illusion of the site unloading elements again (e.g.: slider) because it goes to a different URL of the same page. After the user moves around again it will present the correct language and fully loaded page. This is a 4 step page load for them.

**To Reproduce**
1. Enable delay JS v2
2. Activate WPML automatically detect browser language (found in settings tab "languages")
3. Browse to the main URL using a browser language set to the non default language of the site (my main language is dutch and the translated is french)

**Expected behavior**
WPML should be able to redirect to the correct translated website version before full load.

**Additional context**
I found out this issue exists because I live in a country with multiple official languages and was playing around in a computer store with apple & windows computers setup in a different language than mine. I can imagine this is a jarring experience for those people. I always like to try my site in computer stores as it is difficult to understand how the site will perform in the real world. 

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

Edit: a fast reload may even be triggerd by the WPML setting if it is the same language. This is less visible because it is the exact same content

remyperona:
## Description

Delaying this script causes too many loads of the page. It needs to be executed early to work properly.

Fixes #4149

## Type of change

No

## How Has This Been Tested?

- [ ] Added test fixture to automated tests

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DelayJS/HTML.php
+++ b/inc/Engine/Optimization/DelayJS/HTML.php
@@ -52,6 +52,7 @@ class HTML {
 		'var billing_additional_field', // Woo Autocomplete Nish.
 		'var gtm4wp',
 		'var dataLayer_content',
+		'/dist/js/browser-redirect/app.js', // WPML browser redirect script.
 	];
 
 	/**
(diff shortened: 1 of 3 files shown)


iCaspar (approved):

piotrbak (approved):
Ready to go!
[testrail-report-224.pdf](https://github.com/wp-media/wp-rocket/files/6857717/testrail-report-224.pdf)

PR #4176, opened 2021-07-19
Fixes #4136 'gtag(' not working in Delay JS UI exclusions field

### LINKED ISSUE #4136: Delay JS Execution: PHP Warning when an unclosed parenthesis is in exclusion list
**Describe the bug**
When using exclusions like `gtag(` in Delay JavaScript Execution (version 3.9.0.5) it causes the following PHP warning and the **exclusion does not work**: 

`PHP Warning:  preg_match(): Compilation failed: missing closing parenthesis at offset 5 in /wp-content/plugins/wp-rocket/inc/Engine/Optimization/DelayJS/HTML.php on line 172`

**To Reproduce**
1. Add the following type of code to your site (it's commonly used Analytics code, both inline and external file together): 
```
<script async src="https://www.googletagmanager.com/gtag/js?id=UA-11111111111-1"></script>
<script>
window.dataLayer = window.dataLayer || [];
	  function gtag(){dataLayer.push(arguments);}
	  gtag('js', new Date());


	  	  gtag('config', 'UA-111111111111-1');
	  
	</script>
```
2. Ensure Delay JS is enabled, add this exclusion: `gtag(`

3. Observe on the site that the inline script is not excluded and find the warning in your error log.

**Expected behavior**
The script should be excluded without warning

**Additional context**
These types of exclusions are listed in our docs: https://docs.wp-rocket.me/article/1492-google-analytics-google-ads-tracking-issues
We can update those of course, but they should work anyway :)

This is a similar issue to this one: https://github.com/wp-media/wp-rocket/issues/3967

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

iCaspar:
## Description

Closes #4136 

## Type of change

Same.

## How Has This Been Tested?

Confirmed locally that with new pattern substitution warning does not occur.

**Note: This PR adds a new test and html/expected set in both the unit and integration test fixtures** in order to test specifically for user-added combinations of scripts added to the exclusion UI. (the others only test for the expected default exclusions and adding user data would invalidate the results of those. The new tests validate user-only UI entries, ignoring the defaults.

# Checklist:


remyperona (approved):

piotrbak:
@iCaspar It fixes this specific pattern but not any other, similar pattern (`example(`). Is there any way we can cover this in an universal way?

At some point we'll for sure have similar case reported.

remyperona:
Closing this PR in favour of #4195

PR #4187, opened 2021-07-21
Fix 4161 - Unnecessary database growth due to urls with same content

### LINKED ISSUE #4161: RUCSS - database grows large when versions of the CSS/JS files change frequently
- Made sure you’re on the latest version - 3.9.0.5

**Describe the bug**
When a page contains a CSS or JS file with a version that is changing dynamically very often (for example on each page load), the database may grow very large even for smaller sites.

**To Reproduce**
The issue is reported in the ticket https://secure.helpscout.net/conversation/1571200331/279553/
where a version on the file like this:` https://example.com/wp-content/uploads/essential-addons-elementor/cb70d11b8.min.css?ver=1626273953` was changing on every page load.

**Expected behavior**
Avoid regeneration of used CSS if the content of the file doesn't change.

**Screenshots**
Video shared in Slack: https://wp-media.slack.com/archives/C08N8J6VC/p1626265876027200?thread_ts=1626264564.026500&cid=C08N8J6VC

**Additional context**
Slack thread: https://wp-media.slack.com/archives/C08N8J6VC/p1626264564026500

@engahmeds3ed remarks:

> The main problem here is the version, as for each url we search for it on the DB and then compare the hash of the content on the DB with the hash of the content on the file system so in our case we have different url in each cache generation, it's something like the WPR issue related to the huge size of min directory inside cache when any resource has timestamp on the version.
> We may search with the hash not the url.

Ticket: https://secure.helpscout.net/conversation/1571200331/279553/

- [ ] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

iCaspar:
## Description

Fixes #4161 

## Type of change

- [X] Enhancement - Also adds exclude files pattern for `node_modules/*` to our phpcs

Per grooming, checking the URL first, and then the hash when url is not matched to make sure content is also different.
Also, because the 'prewarmup' item in the resource can now be unset when checking for change against hash we need to update how this check is made to avoid the php error in those conditions.

## How Has This Been Tested?

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Database/Queries/ResourcesQuery.php
+++ b/inc/Engine/Optimization/RUCSS/Database/Queries/ResourcesQuery.php
@@ -74,7 +74,7 @@ class ResourcesQuery extends Query {
 	 *
 	 * @param array $resource Resource array.
 	 *
-	 * @return bool
+	 * @return bool|int False if neither created or updated, or row-id of the new or modified resource.
 	 */
 	public function create_or_update( array $resource ) {
 		$hash = md5( $resource['content'] );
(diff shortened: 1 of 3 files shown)


remyperona (approved):

iCaspar:
Closing. See #4199 for reasons why this cannot be used.

Issue #3967, opened 2021-05-28
PHP Warning with some exclusions in Load JavaScript Deferred


webtrainingwheels:
A PHP warning is triggered with certain exclusions entered into Load JavaScript Deferred. 

`PHP Warning: preg_match(): Compilation failed: missing closing parenthesis at offset 943 in /wp-content/plugins/wp-rocket/inc/Engine/Optimization/DeferJS/DeferJS.php on line 78`

**To Reproduce**

1. Enter the following exclusions in Load JavaScript Deferred:
`gtag(`
`ga( '`
`ga('`
2. Check the error log for the above warning

**Additional context**
Related ticket: https://secure.helpscout.net/conversation/1522666492/266458?folderId=377611

Per Remy, the opening parenthesis are the issue since they are not escaped currently for this field.

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

webtrainingwheels:
https://secure.helpscout.net/conversation/1522119318/266286?folderId=377611
This customer reports that this issue prevents other exclusions in the field from working.

mostafa-hisham:
### Reproduce the issue ✅ 
Reproduced on local test site

### Identify the root cause ✅ 
As Remy said, the opening parenthesis are the issue since they are not escaped currently for this field.

### Scope a solution ✅ 
I think it is better to escape all the characters used in regex patterns 
we can use this method [preg_quote](https://www.php.net/manual/en/function.preg-quote.php)
which escape The special regular expression characters: . \ + * ? [ ^ ] $ ( ) { } = ! < > | : - #

We need to update get_excluded function
https://github.com/wp-media/wp-rocket/blob/4b22e022aebb2517f29b5e31fd47899829256e9e/inc/Engine/Optimization/DeferJS/DeferJS.php#L193

to escape the special regular expression characters for "exclude_defer_js" field
```
		$exclude_defer_js_option= $this->options->get( 'exclude_defer_js', [] );

		foreach ( $exclude_defer_js_option as $i => $exclude ) {
			$exclude_defer_js_option[ $i ] = preg_quote($exclude, '/');
		}

		$exclude_defer_js = array_unique( array_merge( $exclude_defer_js,$exclude_defer_js_option ) );
```
And also remove the for loop that escapes only "#" for all $exclude_defer_js array

[...]

PR #4195, opened 2021-07-22
Validate pattern fields input by escaping invalid regex strings instead of removing them

### LINKED ISSUE #4136: Delay JS Execution: PHP Warning when an unclosed parenthesis is in exclusion list

iCaspar:
## Description

This is a more generalized solution to #4136, and applies the same idea to all the pattern fields in the admin UI.
It addresses more generally the problem of recognizing when a use may enter either a literal string (like a URL) or a pattern (to identify, for example, a snippet of an inline script) for exclusion.

To now, we have been validating these fields by removing any entry that returns `false` when subjected to `preg_match()`. However, this removes certain snippets that a user might enter intending that they be interpreted literally rather than as a regex pattern. For example, `gtag(` commonly used to identify the inline google analytics script.

We cannot in these cases maintain an endless list of these exceptions or anticipate what users might enter with this intention, and if we strip them during validation, it is not obvious to the user why they are being stripped. However leaving them to be processed as regex generates errors and fails to be exclude the files/in-lines per the user's intention.

Instead of stripping them, we'll run them through `preg_match()` to validate any regular expression, and failing that validation, we can escape the entry (`preg_quote()` it) in order to force it to work as a string literal in any subsequent pattern evaluation.

Fixes #4136

## Type of change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] This change requires a documentation update

As described above, this differs from the original grooming of #4136 in that it takes a more comprehensive approach that will affect all pattern-related fields in the UI.

## How Has This Been Tested?

Verified locally that pattern fields now escape rather than remove line-entries that do not conform to valid regex patterns.

# Checklist:


iCaspar:
Tagging @GeekPress and @webtrainingwheels on this one. @Tabrisrp and I were discussing this on the dev channel earlier, as a more comprehensive way to address @piotrbak 's comment on #4176.
Any thoughts or concerns here? Also not sure what docs might be affected (or useful to generate) about this.

GeekPress:
@iCaspar Not sure to know which solution has been finally decided, do we have to use `gtag(` or `gtag\(`?

iCaspar:
@GeekPress @wp-media/qa 

The expected behavior now is that in any field that accepts a pattern (such as delayJS exclusion or RUCSS safelist), when users enter any content that could mess up a regex evaluation will be automatically preg_quoted when saving options. Thus, when a user will enter `gtag(` it will be saved as `gtag\(` and appear with the escaped backslashes in the field after saving.

This also changes the notice displayed when an entry is modified in this way. Previously, invalid entries were removed completely with the notice saying: `The following pattern is invalid and has been removed:`
Now we escape it for them and the notice says: `The following pattern has been escaped to prevent errors:`.

Finally, there was a (previously unreported) bug here that only the first invalid pattern in any given field was removed. We now check and escape all invalid patterns if there are more than one in the same field.

inc/admin/options.php
```diff
@@ -121,13 +121,14 @@ function rocket_pre_main_option( $newvalue, $oldvalue ) {
...
-		'cache_purge_pages' => __( 'Always Purge URL(s)', 'rocket' ),
-		'cdn_reject_files'  => __( 'Exclude files from CDN', 'rocket' ),
+		'exclude_css'         => __( 'Excluded CSS Files', 'rocket' ),
+		'exclude_inline_js'   => __( 'Excluded Inline JavaScript', 'rocket' ),
+		'exclude_js'          => __( 'Excluded JavaScript Files', 'rocket' ),
+		'delay_js_exclusions' => __( 'Delay JS Exclusions', 'rocket' ),
+		'cache_reject_uri'    => __( 'Never Cache URL(s)', 'rocket' ),
+		'cache_reject_ua'     => __( 'Never Cache User Agent(s)', 'rocket' ),
+		'cache_purge_pages'   => __( 'Always Purge URL(s)', 'rocket' ),
+		'cdn_reject_files'    => __( 'Exclude files from CDN', 'rocket' ),
```
remyperona, line 131: We should also add the remove unused CSS safelist to the list

remyperona (changes requested):

remyperona (approved):

Mai-Saad:
@iCaspar @Tabrisrp Please find below notes/questions during exploratory:
**Delay JS**
1- Using `gtag(){` in exclusion, didn't work i.e `\` not added and script still delayed, shouldn't it be handled?
2- Using `gtag (` in exclusion, `\` was added like `gtag \(` but script is still delayed, shouldn't it be excluded in this case?
sample script:
`<script async src="https://www.googletagmanager.com/gtag/js?id=UA-11111111111-1"></script>`
```
<script>
          window.dataLayer = window.dataLayer || [];
	  function gtag(){dataLayer.push(arguments);}
	  gtag('js', new Date());
         gtag('config', 'UA-111111111111-1');  
</script>
```

note: `gtag(` was handled correctly i.e turned into `gtag\(` and script was delayed

**RUCSS**
1- Can you please provide example here of possible handled safelist?
2- Tried `.second {` , `.grid-container >` but both didn't work
sample style:
```
div.second {
      opacity: 0.3;
      background-color: green;
    }
 .grid-container > div {
  background-color: rgba(255, 255, 255, 0.8);
  border: 1px solid black;
  text-align: center;
  font-size: 30px;
}
```

**General question:**
- Is it expected that this PR fixes the following issue https://github.com/wp-media/wp-rocket/issues/3967 ? (currently I can't see the fix for it)
- Currently can't see admin notice mentioned here, am I missing something?

> Now we escape it for them and the notice says: The following pattern has been escaped to prevent errors:

iCaspar:
@Mai-Saad 
On (1):
Your results there are "expected": `gtag(){` is a valid regular expression, so it will not be escaped. However as a regular expression it does not match the script sample, so it will also not exclude it from delay.

On (2):
It _may_ be a type-o, but `gtag (` (with the space) will not match (and will not be excluded even when escaped) because there is no space in the script sample. `gtag(` (no space) _should_ work.

RUCSS samples:
```
.se[c|gond      -->    \.se\[c\|gond
.grid-container\-\d     --> stays as is, but .grid-container-1 in a sample css will not be safelisted
.grid-container(      -->   \.grid-container\(
```
That said, except for `(.*)` wildcard regex, they will not be actually safelisted, because treeshaker recognizes _only_ the `(.*)` pattern. So for this field, that invalid regexes are escaped is all we expect: `.grid-container(` -> `grid-container\(` in the UI, but treeshaker will not safelist it. Also `.sec[|gond` should be escaped, but treeshaker will not safelist anything for it.

Issue #3967:
These are similar issues, but this will probably not fix, as there are still some changes needed in parsing deferred mentioned by @mostafa-hisham on that issue.

Message display:
I suspect it may be a different bug that is preventing display of message, as the actual display process is not changed in this PR, only the contents of the messsage. How to test: If you enter `gtag(` in the DelayJS field on `develop` or `trunk` and it is removed, does the old message about removal display? If not, we have a new issue for display failure.

Mai-Saad:
Thanks @iCaspar for clarifications!

Here is what's happening while using trunk:
- For Delay JS, When added `gtag(` for Delay JS exclusion, nothing was removed in exclusion, nothing was excluded from delay JS and no notice was displayed .. so we aren't expecting notice here? or is it bug on trunk?
- For Delay JS, If we have the following files:
```
<script src="/wp-content/rocket-test-data/scripts/test(filename.js"></script>
<script src="/wp-content/rocket-test-data/scripts/test(filename).js"></script>
<script src="/wp-content/rocket-test-data/scripts/testfilename.js"></script>
```
excluding `test(filename).js` will results in excluding `testfilename.js` , shouldn't it exclude `test(filename).js` ? {note: we do the same on branch} (not sure if this is valid URL format)
- For RUCSS:
When we safelist `div.(.*` , no used CSS is created  on trunk , while on branch, it turns into `div\.\(\(\.\*\)` and nothing is safelisted but we still have used CSS => is this expected? or shall safelist work? (Note: if we re-enabled RUCSS after that the safelist will be `div\\\.\\\(\\\(\\\.\\\(\.\*\)\\\)`, shouldn't it remain as is?

iCaspar:
@Tabrisrp See @Mai-Saad comment above re RUCSS escaping, then re-escaping.
I'm having second thoughts about using the pattern linting in the safelist UI, as the only pattern that treeshaker will accept (unlike the DelayJS) is the wildcard `(.*)` which is either correct and it is safelisted or not and it fails to safelist. The new safelist will account for escaping anything else in cases where it's needed, so I _think_ this field should assume everything is a string literal. Thoughts?

@Mai-Saad 
The `gtag(` in DelayJS field currently does as you noted there, which was the original issue (#4136) -- There was no visible result, but the PHP warning was posted in the logs. Currently on trunk, there is no regex checking, so you will not see a message there. (My previous comment for testing was wrong, sorry.)
To check for the removal/message display, you could try to enter `gtag(` into a field (such as Advanced Rules -> Always Purge URLs) that on trunk uses the regex pattern detection. On my install it removes the `gtag(` but I don't see any message... so I think the message failure should probably be a new issue. :) 

For the files you listed there, parenthesis are technically possible in urls... but since we are going by patterns, the correct one is selected for exclusion as you have described there. To exclude `test(filename).js` try `test\(filename\)\.js` -- on one hand, it's tricky, but it's what would be required if you wanted to exclude it; on the other hand parenthesis are possible, but not recommended, in filenames so a user that needed such a thing should probably expect some odd things to happen.

remyperona:
@iCaspar You're right we might want to treat this field differently. If it's too much, we could skip validation for this field for now, and re-visit again later.

iCaspar:
@Mai-Saad 
The double-escaping was from the checking for specific `(.*)` wildcard type-O. (It was re-adding a parenthesis when it shouldn't and throwing the escaping off.
I fixed that, so `div.(.*` will be automatically "fixed" to `div.(.*)` (note the final wildcard sequence is corrected) -- this results in a valid regex, so we should expect that it will be simply corrected without escaping at all. 🤞

iCaspar:
@Mai-Saad This should be resolved now.
The last change should also enable the regex escaping to be done on the affected fields on settings import as well.
🤞

Mai-Saad (approved):
Working as expected
[testrail-report-253.pdf](https://github.com/wp-media/wp-rocket/files/7135553/testrail-report-253.pdf)

piotrbak:
There is a problem with specific markups, it's hard to test all possible cases:
```
h(.*)apply(.*)(
h(.*)apply(.*))
\*
```

PR #4202, opened 2021-07-27
Closes #4018 Add delay JS compatibility with Adthrive

### LINKED ISSUE #4018: Automatically exclude AdThrive scripts from delay JS
**Describe the bug**
Our new delay JS is causing issues for users of AdThrive, preventing ads loading and CLS management before any user interaction.

Exclude the 2 inline scripts inserted in the HTML would avoid the issue, and improve the experience of our common users.

To do so, we can add `adthrive` to our exclusion patterns.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Add an automatic compatibility with Adthrive for delay JS, by excluding the `adthrive` pattern, adding it in the delay JS exclusion field when Adthrive Ads is enabled.

Closes #4018

## Type of change

The solution is different, as we want to add the pattern into the exclusion field.

It is added when:
- Delay JS is enabled in WP Rocket
- The pattern is not yet present in the exclusions field
- Adthrive Ads is activated
- When updating to 3.9.3

## How Has This Been Tested?

- [x] Automated tests
- [x] Manual tests with Adthrive Ads plugin

# Checklist:


mostafa-hisham (approved):

Mai-Saad:
@Tabrisrp while exploring the PR, Found that:
1- When disabling Adthrive, adthrive is not removed from the exclusion list => shouldn't we auto-remove it?
2- When downgrading to 3.9.1.1 or 3.9 while adthrive +delay js enabled => adthrive is still there in the exclusion list, shouldn't it be removed? **------> I think this one is expected as in general we preserve exclusion list values while upgrade/downgrade**
3- When the user decides to remove this exclusion and save => if we refresh the page, adthrive will be auto-added to exclusion, does this mean users can't remove it from exclusion? (I think we better have text here to clarify this, it's so confusing that it is removed with save and appear again when refresh page)
4- When import settings where delayjs is enabled + Adthrive is enabled , adthrive is not added in exclusion till we refresh the page...shouldn't it be displayed when open file optimization page after importing settings with no need to refresh?
5- If `adthrive` was inside any script i.e part of the string (or probably URL), this script will be excluded from delayjs. shouldn't it be ignored if it's inside a string? 
This will be excluded if adthrive + delay js are enabled `<script>
document.getElementById("page").innerHTML = "Helloooooo adthrive JavaScript!";
</script>`

remyperona:
@wp-media/php Ready for another review

I updated the code to add the pattern on plugin update to 3.9.2.
I also updated the code of WP Meteor to use `get_option()` and `update_option()`, because on update to 3.9.2, both Adthrive and WP Meteor are going to update the options, and if doing it with the instantiated class, the data won't be reflected.

@wp-media/qa Notice that this PR contains a change on the WP Meteor code, so a new check for the part that is doing changes on plugin update will be necessary to confirm it stills works as expected.

mostafa-hisham (approved):

ncolletti:
Could this PR also include an addition to the text $exclusion matching in the HTML class? This would ensure Adthrive's inline scripts are also not blocked, we have potentially more outside of just adthrive-ads.php that come from our plugin.

`'adthrive', // adthrive inline scripts`

I can also open a separate PR for this, let me know what the team would prefer, thanks!

remyperona:
@ncolletti the pattern we add in the exclusions field will match against URL and inline scripts, adding it to the array in the HTML class would end up with the same result, except it would not be removable by the user.

The implementation done in this PR is the one agreed on with the AdThrive team, if this needs to be re-considered, we need to discuss again.

ncolletti:
hey @Tabrisrp -- that is fine and will work properly on our end. We were curious if this effort would still work for other inline scripts. Thank you for confirming.

Mai-Saad (commented):
@Tabrisrp  Currently:

1. If we enable Adthrive while delay js on => adthrive added to exclusion ==> so even if the user remove the exclusion, it will be added automatically when reactivating the plugin
2. If we enable delay js while Adthrive is on => adthrive not added to exclusion ===> **We must include adthrive into the exclusion list in this case too as per discussion with @GeekPress** 
3. If we update while Adthrive + delay js on => adthrive added to exclusion

remyperona:
@wp-media/php Requesting a new review following the latest changes I made

iCaspar (approved):

remyperona:
@wp-media/qa Ready again for QA, now when activating delay JS and adthrive is active, we add the pattern too.

Mai-Saad (approved):
Works as expected.
[testrail-report-239.pdf](https://github.com/wp-media/wp-rocket/files/7046373/testrail-report-239.pdf)

PR #4211, opened 2021-07-30
Closes #4080 Emulate elementor pro animation with delay JS

### LINKED ISSUE #4080: Emulate Elementor Pro animation to improve user experience
Elementor Pro allows users to add animation to elements. By using our Delay JS and when we delay everything, the animation in the header isn't triggered until user interaction.

It leads to an experience like that:
https://share.getcloudapp.com/o0ueez1J

It's not really a good user experience ^^

We have developed a script that simulates the animation like Elementor is doing to have the content displayed without waiting for user interaction.

Only the animations above the fold are triggered by our script.

The inline script is available in the Delay JS repository into the `elementor` branch and in a separate file. 

**Please note that the delay JS script contains an update too!**

**Additional context**
The inline script related to Elementor animation should be added just before the `</body>` closure and added only if Elementor Pro is activated on the website.

mostafa-hisham:
## Description

ِAdd js script to fix Elementor pro animations if delay js is enabled

Fixes #4080 

## Type of change

No

## How Has This Been Tested?

- [x] create Unit and integration 

# Checklist:


assets/js/elementor-animation.js
```diff
@@ -0,0 +1,117 @@
+class RocketElementorAnimation {
```
remyperona: We should minify this file

inc/ThirdParty/Plugins/PageBuilder/ElementorPro.php
```diff
@@ -0,0 +1,76 @@
...
+		}
+
+		return [
+			'rocket_buffer' => [ 'add_fix_animation_script', 28 ],
+		];
+	}
+
+	/**
+	 * Add Fix Elementor Pro animations script.
+	 *
```
remyperona:
```suggestion
	 *
	 * @since 3.9.2
	 *
```

inc/ThirdParty/Plugins/PageBuilder/ElementorPro.php
```diff
@@ -0,0 +1,76 @@
...
+	 * @param html $html DelayJS HTML class.
```
remyperona:
```suggestion
	 * @param string $html HTML content.
```

remyperona (changes requested):

remyperona (approved):

Mai-Saad:
@Tabrisrp @mostafa-hisham While exploring the fix found that:
1- Current fix only handle motion effect on text (which already exists in normal Elementor) while still not handling animated headline feature (open with Elementor Pro) i.e animated headline will only appear with user interaction, Do we need it to appear without user interaction?
2- If Elementor Pro is deactivated, we aren't handling the Motion effect which was already handled in 1 => I think if we kept the script while Elementor is activated then we will fix this case
3- With the default exclusions, we still add the fixing script, Do we still need to add it?
4- Will the fixing script be added on all pages or only the page using Elementor animation? i.e this page https://new.rocketlabsqa.ovh/3279template/

Notes: 
- test page is  https://new.rocketlabsqa.ovh/elementor-pro-animation/
- Unhandled effects mean that it won't appear till user interaction occurs

remyperona:
@GeekPress see Mai's questions above

GeekPress:
1- Animated headline feature is out of the scope of this enhancement.

2- Let's add our script with Elementor is detected instead of Elementor Pro.

3- We can keep adding our Elementor script, it shouldn't create a conflict. The inline script is very small, so it won't have a bad impact on performance if we keep it.

4- The script can be added on all pages for the same reason as 3.

mostafa-hisham:
@Mai-Saad 
Now the script will be added if `Elementor` is active, not `ElementorPRo`

Mai-Saad (commented):
All test cases pass but we have a problem on the branch to display Desktop/ mobile at the bottom of the page while Delay js only is enabled (tested latest problem on both newlabs and rocketlabs)
Screenshot from 2021-08-06 15-25-44

mostafa-hisham:
@Mai-Saad  @GeekPress 
there is a problem with the script that is used to fix elementor animations

here the script appends this span to the end of the page
`<span id="elementor-device-mode"></span>`

and then after the js that was delayed by Delay js feature runs again when users interact with the page
it adds a new span with the same id and the class that should hide it
`<span id="elementor-device-mode" class="elementor-screen-only"></span>`

https://github.com/wp-media/delay-javascript-loading/blob/156b81eddf5a1d845b0f934ecea66dfcf24aad65/elementor-animation.js#L3-L9
```
    constructor() {
        // Injects a span, that will later provide the device type from some CSS media query rules
        this.deviceMode = document.createElement('span');
        this.deviceMode.id = 'elementor-device-mode';
        document.body.appendChild(this.deviceMode);
    }
```
I fixed the problem by removing this line but this needs to be checked by Gaël Métais

https://github.com/wp-media/delay-javascript-loading/blob/156b81eddf5a1d845b0f934ecea66dfcf24aad65/elementor-animation.js#L7
 `document.body.appendChild(this.deviceMode);`

remyperona (approved):

GeekPress:
@mostafa-hisham To fix the issue we also need to add the class `elementor-screen-only` to the span.

This code should fix our problem:

```
constructor() {
        // Injects a span, that will later provide the device type from some CSS media query rules
        this.deviceMode = document.createElement('span');
        this.deviceMode.id = 'elementor-device-mode';
        this.deviceMode.setAttribute('class', 'elementor-screen-only');
        document.body.appendChild(this.deviceMode);
}
```

mostafa-hisham:
@GeekPress 
I tried the code you send and it will fix the issue but also will append the span twice

image

GeekPress:
@mostafa-hisham Does it cause any issues to have it twice? If not, I think we can keep it.

You can try WP Meteor which has an option to do exactly the same, and check if the span is also added twice or not.

You can also have a look at WP Meteor code here:
- https://plugins.trac.wordpress.org/browser/wp-meteor/trunk/src/js/public/includes/elementor/device-mode.js
- https://plugins.trac.wordpress.org/browser/wp-meteor/trunk/src/js/public/includes/elementor/animations.js

mostafa-hisham:
@GeekPress 
I checked it with elementor it didn't cause any problems with animations
I updated the PR with the fix and I will send it to QA to check it

Mai-Saad (approved):
The latest problem is fixed now.
[testrail-report-235.pdf](https://github.com/wp-media/wp-rocket/files/6954134/testrail-report-235.pdf)

PR #4232, opened 2021-08-06
Add Consent management platforms to WPRocket delay js exclusions

patmmccann:
## Description
Publishers implementing  https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework often provide a consent management "stub" script that essentially lets other scripts know that a consent management platform is coming. 

In order to comply with the law, this script must execute before any other script which might need to be aware of consent signals. WP Rocket delaying it puts publishers at extreme legal risk.

https://github.com/InteractiveAdvertisingBureau/GDPR-Transparency-and-Consent-Framework/blob/d56817415ccdd7168dfafdd02c659d91451e74c9/TCFv2/IAB%20Tech%20Lab%20-%20CMP%20API%20v2.md#how-can-scripts-on-a-page-determine-if-there-is-a-cmp-present

"Is there a sample CMP “stub” API script?
You can find an iab-supported open-source implementation of the stub API here: https://github.com/InteractiveAdvertisingBureau/iabtcf-es/blob/master/modules/stub/

This code should be executed on the page before any other scripts that require the __tcfapi function – this usually means between the <head></head> tags of the HTML document. The sample script also includes the postMessage handler."


patmmccann:
@Tabrisrp @GeekPress I am new to this project; can you educate me a bit on if this PR meets requirements for inclusion and if it does, on what schedule that might make its way into a release?

inc/Engine/Optimization/DelayJS/HTML.php
```diff
@@ -54,6 +54,7 @@ class HTML {
 		'/ewww-image-optimizer/includes/load[_-]webp(\.min)?.js', // EWWW WebP rewrite external script.
 		'ewwwWebPInit', // EWWW WebP rewrite inline script.
 		'/dist/js/browser-redirect/app.js', // WPML browser redirect script.
+		'__tcfapi', // Consent managemant platforms implementing the legal framework developed by IAB Europe to comply with GDPR
```
remyperona:
```suggestion
		'__tcfapi', // Consent management platforms implementing the legal framework developed by IAB Europe to comply with GDPR.
```

remyperona (changes requested):

remyperona:
@patmmccann 👋🏼 

Thank you for the PR, only a minor fix to do on the code styling, else this looks good 👍🏼 

After that the PR will go trough our QA process, and if it passes, we will merge it be included in the next release of the plugin.

Are you able to provide a ready to use JS script containing the excluded pattern, so that our QA team can easily test the change?

patmmccann:
Here is ready to use script that implements the iab stub
https://gist.github.com/mavaddat/310e3a9c63a003dff3509417ee640a73

Here is another
https://github.com/InteractiveAdvertisingBureau/iabtcf-es/blob/master/modules/stub/src/stub.js

remyperona (approved):

remyperona:
Thank you, looks ready to go through QA with that info and the last change.

piotrbak:
@patmmccann Could you show us the example of implementation?

You added `__tcfapi` to the exclusion list, which will exclude the inline scripts that are calling this specific function. Is function definition also inline **in all cases**? I can see taht it's inline in this example:
https://gist.github.com/mavaddat/310e3a9c63a003dff3509417ee640a73

But this script will not be excluded:
https://github.com/InteractiveAdvertisingBureau/iabtcf-es/blob/master/modules/stub/src/stub.js

If it's not inline in all cases, we'll also need to exclude the file that contains the function definition, otherwise, it'll end up with `Function not defined error` and product being not usable at all.

patmmccann:
It isn't always inline, it is sometimes synchronous and sourced. For an example of it implemented, wsj.com is an excellent option. I am a little confused on your last comment about the function not defined issue. I assumed you have in-line pattern matching exclusions and external script pattern match exclusions and this PR would only affect the former.

piotrbak:
@patmmccann On wsj.com I can see that the script is inline one. It'd be excluded correctly.

What I meant is that if the inline and sourced script depend on each other, excluding only one from the feature will break things.

When it comes to the exclusions for external script, they can be done in the same array:
https://github.com/wp-media/wp-rocket/blob/f3d047fe745660805b96bcb9e0e927527a664a9d/inc/Engine/Optimization/DelayJS/HTML.php#L56

Please let me know if that makes sense and if you want to add external script too, please go forward. If not, I'll test if everything works and proceed with the current version.

Thank you!

patmmccann:
That does make sense; I think it makes sense to move forward as is, and handle sourced exclusions for IAB TCF compliant vendors such as Quantcast, OneTrust, or LiveRamp in seperate PRs.

patmmccann:
@piotrbak curious where this is landing?

patmmccann:
@Tabrisrp @piotrbak this has been tagged waiting for a while now, but I am not sure there are any open questions

PR #4301, opened 2021-08-31
Exclude Perfmatter Lazy Load from Delay JS

### LINKED ISSUE #4169: Exclude Perfmatter's Lazy Load from Delay JS
**Describe the bug**
When Perfmatters lazy load for images is activated in addition to Delay JavaScript Execution, images will only start showing after user interaction with the page.

This applies also to all lazy-loaded images above the fold.

**To Reproduce**

- Enable Lazy-load for images in Perfmatters
- Enable Delay JS in WP Rocket
- Check a page with images above the fold.
- No image will be disabled on page load

**Expected behavior**
Images should be disabled on page load.

**Additional context**
Excluding the following scripts from Delay JavaScript Execution fixes the issue:

```
/perfmatters/js/lazyload.min.js
lazyLoadInstance
```

Original inline script:

```
<script>document.addEventListener("DOMContentLoaded",function(){var lazyLoadInstance=new LazyLoad({elements_selector:"[loading=lazy],.perfmatters-lazy",thresholds:"200% 0px",callback_loaded:function(element){if(element.tagName==="IFRAME"){if(element.classList.contains("loaded")){if(typeof window.jQuery!="undefined"){if(jQuery.fn.fitVids){jQuery(element).parent().fitVids()}}}}}});});function perfmattersLazyLoadYouTube(e){var iframe=document.createElement("iframe");var params="ID?autoplay=1";params+=0===e.dataset.query.length?"":"&"+e.dataset.query;iframe.setAttribute("src",params.replace("ID",e.dataset.src));iframe.setAttribute("frameborder","0");iframe.setAttribute("allowfullscreen","1");iframe.setAttribute("allow","accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture");e.replaceChild(iframe,e.firstChild)}</script>
```

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

mostafa-hisham:
## Description

Exclude Perfmatter Lazy Load from Delay JS by adding
```
/perfmatters/js/lazyload.min.js
lazyLoadInstance
```
to `DelayJS\HTML::$excluded array`

Fixes #4169 

## Type of change

No

## How Has This Been Tested?

- [x] Unit and integration

# Checklist:


remyperona (approved):

Mai-Saad (approved):
The mentioned inline script is now excluded from delay js. 
Note: not tested using mentioned plugin.

PR #4302, opened 2021-08-31
return false if it is not wc_product

### LINKED ISSUE #4255: Delay JS (Error 500):  Uncaught Error: Call to a member function get_gallery_image_ids() on bool
- Made sure you’re on the latest version ✅ - 3.9.1.1

**Describe the bug**
We are using the function wc_get_product() and that function can return these possible values: WC_Product | null | false ([Reference](https://woocommerce.github.io/code-reference/namespaces/default.html#function_wc_get_product))

And here it seems that we are not handling all the possible return values: 

https://github.com/wp-media/wp-rocket/blob/a38f5b6aa1853d221b97ebbaa9ef9638cafd7e2c/inc/ThirdParty/Plugins/Ecommerce/WooCommerceSubscriber.php#L515

That causes when the function returns false, we get and error when trying this:

https://github.com/wp-media/wp-rocket/blob/a38f5b6aa1853d221b97ebbaa9ef9638cafd7e2c/inc/ThirdParty/Plugins/Ecommerce/WooCommerceSubscriber.php#L516

The full error is:

> PHP Fatal error:  Uncaught Error: Call to a member function get_gallery_image_ids() on bool in /srv/users/signetprelive/apps/signetprelive/public/wp-content/plugins/wp-rocket/inc/ThirdParty/Plugins/Ecommerce/WooCommerceSubscriber.php:516

**Expected behavior**
The site shouldn't throw an error 500.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

mostafa-hisham:
## Description

return false if it is not wc_product

Fixes #4255 

## Type of change

No

## How Has This Been Tested?

- [x] unit and integration tests

# Checklist:

### DIFF
--- a/inc/ThirdParty/Plugins/Ecommerce/WooCommerceSubscriber.php
+++ b/inc/ThirdParty/Plugins/Ecommerce/WooCommerceSubscriber.php
@@ -513,7 +513,7 @@ class WooCommerceSubscriber implements Event_Manager_Aware_Subscriber_Interface 
 	 */
 	private function product_has_gallery_images() {
 		$product = wc_get_product( get_the_ID() );
-		if ( ! $product || empty( $product ) ) {
+		if ( empty( $product ) ) {
 			return false;
 		}
 		return ! empty( $product->get_gallery_image_ids() );
--- a/inc/ThirdParty/Plugins/Ecommerce/WooCommerceSubscriber.php
+++ b/inc/ThirdParty/Plugins/Ecommerce/WooCommerceSubscriber.php
@@ -513,7 +513,7 @@ class WooCommerceSubscriber implements Event_Manager_Aware_Subscriber_Interface 
 	 */
 	private function product_has_gallery_images() {
 		$product = wc_get_product( get_the_ID() );
-		if( !$product || empty($product) ){
+		if ( ! $product || empty( $product ) ) {
 			return false;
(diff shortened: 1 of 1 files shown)


inc/ThirdParty/Plugins/Ecommerce/WooCommerceSubscriber.php
```diff
@@ -513,6 +513,9 @@ private function get_nonce_actions() {
 	 */
 	private function product_has_gallery_images() {
 		$product = wc_get_product( get_the_ID() );
+		if ( ! $product || empty( $product ) ) {
```
remyperona: empty will check for both null and false, so we don't need the first part of the condition

remyperona (changes requested):

remyperona (approved):

piotrbak (approved):
Good to go!
[testrail-report-249.pdf](https://github.com/wp-media/wp-rocket/files/7117334/testrail-report-249.pdf)

katielgc:
Hi all, could you let me know when this will be merged into the master for deployment on our site?

iCaspar:
@katielgc We expect, pending final QA, to merge and release this week. 🤞

PR #4312, opened 2021-09-01
add delay js script after meta charset if exists

### LINKED ISSUE #4121: Charset declaration is missing or occurs too late in the HTML (in some cases)

**Describe the bug**
On some occasions, Lighthouse & PageSpeed will show the following "Best Practice" notification: 

`Charset declaration is missing or occurs too late in the HTML`

**To Reproduce**
1. Enable Delay JavaScript Execution
2. Check a cached page source code
3. Run a test on Lighthouse with "Best Practices" enabled
4. See the error

From my testing, this will happen occasionally on some sites.

**Expected behavior**
Inject the IE fallback script after the meta tag defining the `charset` to prevent the recommendation from showing. 

**Screenshots**
image

**Additional context**
The notification is prompted because we add our IE fallback script (6300 bytes) right after the opening heading tag, which pushes the `<meta charset>` below the first 1024 bytes threshold. 

https://github.com/wp-media/wp-rocket/blob/4f86c478c74e962eb6361b59a0f0b575adab27a9/inc/Engine/Optimization/DelayJS/Subscriber.php#L98-L112

We could possibly run a check before defining the `$pattern`. If there is a meta tag defining the `charset`, it can be targeted with a pattern. 

`/<head[^>]*>/i` would be kept as a fallback in case there is no meta tag defining the `charset`.

Ticket - https://secure.helpscout.net/conversation/1551660251/273982/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

mostafa-hisham:
## Description
add delay js script after meta charset if exists

Fixes #4121 

## Type of change

No

## How Has This Been Tested?

# Checklist:


remyperona (approved):

piotrbak:
@mostafa-hisham I have some concerns here:
1. I think that we should also make sure that `<meta charset="UTF-8">` is in the <head> element. Right now, if HTML document doesn't have the charset defined in the head, but someone is writing a post and include it in the post content, we'll add our script in the post content:
image

2. That kind of markup is also acceptable:
```
<meta name="keywords" 
   charset="UTF-8"
   content="Hello!" />
```

3. That kind of markup too, but I think it's the same as above in terms of regex:
```
<meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
```

4. I remember that it was a product decision to put this script at the very top od the head. If charset is defined in, for example, bottom of the head (which doesn't produce errors in HTML validator), we'll add our script there. @GeekPress Is that acceptable from your point of view?

mostafa-hisham:
@piotrbak 
I updated the code to only find the correct meta charset tags
```
<meta charset="UTF-8">
<meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
```
in `<head></head>` only and add the script after it if found

can you please check if it works ?

piotrbak:
@mostafa-hisham When I add charset rule **above** the head, we're also adding our script there, outside the head.

I think this is an edge case and problem with HTML, not sure if we should handling this. Let's ask @Tabrisrp

mostafa-hisham:
@piotrbak @Tabrisrp 
I think Regex is not the best solution to handle this issue, there are a lot of cases that need to be handled 
- Meta tag is above <head>
- a lot of lines between Meta tag and closing </head> tag

I did some research and most people use 
- [get_meta_tags()](https://www.php.net/manual/en/function.get-meta-tags.php) 
but it takes The path to the HTML file, as a string. This can be a local file or an URL. not HTML String
-  DOMdocument() & xpath

I will do more research maybe i find a better regex for it

piotrbak (changes requested):
Unfortuantely, it doesn't work well.

1. When there's no `charset` definition in `head`, but it's present in the body, we'll take it and paste to the head.
2. When there's a `charset` definition in `head` and there's another one in the body, we'll move the one form head to the top and **erase** the second one

I believe, just like in the previous approach, that we should look for the meta charset only in the `<head></head>`

piotrbak (approved):
Seems that we're good here! 🎉 
[testrail-report-254.pdf](https://github.com/wp-media/wp-rocket/files/7137681/testrail-report-254.pdf)

PR #4333, opened 2021-09-07
Closes #4105 update monolog to version 2

### LINKED ISSUE #4105: Upgrade Monolog to version 2
I'm installing WpRocket via Composer and the monolog version conflicts with [another package](https://github.com/roots/acorn) which require v2 

Please update to monolog  ( should be pretty easy to update it and probably requires no changes [https://github.com/Seldaek/monolog/blob/main/UPGRADE.md](https://github.com/Seldaek/monolog/blob/main/UPGRADE.md)

remyperona:
## Description

Update Monolog to version 2. Also add a small update to the container package

Fixes #4105

## Type of change

Added the container update

# Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,7 @@
 	"require": {
 		"php": ">=7.0",
 		"composer/installers": "~1.0",
-		"monolog/monolog": "^1.0",
+		"monolog/monolog": "^2.0",
 		"psr/container": "1.0.0"
 	},
 	"require-dev": {
--- a/inc/Dependencies/League/Container/Container.php
+++ b/inc/Dependencies/League/Container/Container.php
@@ -191,7 +191,7 @@ class Container implements ContainerInterface
     /**
      * {@inheritdoc}
      */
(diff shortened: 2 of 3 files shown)


remyperona:
Possibly I went too fast on this, I see now that monolog 2 requires PHP >= 7.2. This could cause issues for us as we still support PHP 7.0 and 7.1.

Let's think how we could handle that?

aglipanci:
@Tabrisrp if the WP Rocket can work with both v1 & v2 what if we can allow them both? Based on the PHP version the right one will be installed. In the composer.json it will look something like this `"monolog/monolog": "^1.0 | ^2.0"`.

The current version what is restricted to version 1 will not allow the installation of the WP Rocket with Sage10 Acorn.

lhall-amphibee:
I'm also having a conflict with illuminate/log which is asking for Monolog 2 in the Themosis framework.

ogorzalka:
Confirmed in other projects which use composer (example: bedrock architecture type)

Issue #4303, opened 2021-08-31
Update Delay JS script to the latest version


GeekPress:
For the 3.9.3 release, we need to update the Delay JS script to the latest version available. This version fixes a bug to avoid the user clicks on a button and nothing happens.

Branch to be merged: `replay-events`

inewton24:
Nice, but now we have bug in real IOS 14 Safari (not emulator). replay-events not work, but if we will exclude main script in plugin, we will have double click event.

inewton24:
> Nice, but now we have bug in real IOS 14 Safari (not emulator). replay-events not work, but if we will exclude main script in plugin, we will have double click event.

i think it is because of this article - https://stackoverflow.com/questions/43335183/window-click-event-does-not-fire-on-ios-safari-javascript-only click event not work in IOS, but touchstart must be good

Issue #4243, opened 2021-08-13
RUCSS - PHP Notice:  Trying to get property 'id' of non-object in /RUCSS/Database/Queries/ResourcesQuery.php on line 144


vmanthos:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version ✅ `3.9.1.1`
- Used the search feature to ensure that the bug hasn’t been reported before ✅

**Describe the bug**

The following PHP notice appeared in the `debug.log`:
```
PHP Notice:  Trying to get property 'id' of non-object in /home/public_html/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/Database/Queries/ResourcesQuery.php on line 144
```

This is taking place here:
https://github.com/wp-media/wp-rocket/blob/d1fc8854869892b4ec5c061f88af98e49c6e450d/inc/Engine/Optimization/RUCSS/Database/Queries/ResourcesQuery.php#L144

On the specific site, none of the discovered pages are scanned, so no resources are fetched, i.e. `wpr_rucss_resources` is empty while `$used_css->unprocessedcss` isn't.
https://github.com/wp-media/wp-rocket/blob/d1fc8854869892b4ec5c061f88af98e49c6e450d/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L226

**To Reproduce**

Blocking the scanning process should be the way to go, maybe by short-circuiting it here:
https://github.com/wp-media/wp-rocket/blob/d1fc8854869892b4ec5c061f88af98e49c6e450d/inc/Engine/Optimization/RUCSS/Warmup/ScannerProcess.php#L85-L93

**Expected behavior**

No notices should appear, even when none of the pages are scanned.

**Additional context**

We could bail out if `$db_row` isn't an object :
```PHP
if( ! is_object(  $db_row ) ){
return;
}
```
**Ticket:** https://secure.helpscout.net/conversation/1600151165/287220?folderId=2135277

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
### Identify the root cause ✅ 
We are not checking if the `$db_row` variable is an object before trying to use its properties.

### Scope a solution ✅ 
As noted in the OP, we should check if `$db_row` is an object, and bail-out if it isn't

### Estimate the effort ✅
Effort `[XS]`

[...]

Issue #4241, opened 2021-08-11
Delay JS Execution prevents search engines from reading Mediavine's video schema


vmanthos:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version ✅ `3.9.1.1`
- Used the search feature to ensure that the bug hasn’t been reported before ✅

**Describe the bug**

Mediavine is adding a video schema like the following for their publishers who are running videos:
```Javascript
<script class="mediavine-video__ld-json" type="application/ld+json">{"@context":"https://schema.org","@id":"https://video.mediavine.com/videos/vhjhd087s4vdvjlorkae","@type":"VideoObject","contentUrl":"https://mediavine-res.cloudinary.com/video/upload/t_original/v1621439458/vhjhd087s4vdvjlorkae.mp4","description":"test","duration":"PT7S","keywords":"test","name":"Homemade Pesto Recipe","thumbnailUrl":["https://mediavine-res.cloudinary.com/image/upload/s--nRIdcqHi--/ar_16:9,c_fill,f_auto,fl_lossy,q_auto/v1626459236/oscshyf0thxmfqzv2wie.jpg","https://mediavine-res.cloudinary.com/image/upload/s--dtfsLUS5--/ar_4:3,c_fill,f_auto,fl_lossy,q_auto/v1626459236/oscshyf0thxmfqzv2wie.jpg","https://mediavine-res.cloudinary.com/image/upload/s--hxlskzXf--/ar_1:1,c_fill,f_auto,fl_lossy,q_auto/v1626459236/oscshyf0thxmfqzv2wie.jpg"],"uploadDate":"2021-05-19T00:00:00.000Z","hasPart":[{"@type":"Clip","name":"zero second","startOffset":0,"url":"https://support-sandbox.com/trellis-blocks-test/?mvs=0#mv-first-video"},{"@type":"Clip","name":"one second","startOffset":1,"url":"https://support-sandbox.com/trellis-blocks-test/?mvs=1#mv-first-video"},{"@type":"Clip","name":"two boi","startOffset":2,"url":"https://support-sandbox.com/trellis-blocks-test/?mvs=2#mv-first-video"},{"@type":"Clip","name":"three boi","startOffset":3,"url":"https://support-sandbox.com/trellis-blocks-test/?mvs=3#mv-first-video"}]}</script>
```

That's done by a script:
```HTML
[...]

PR #4348, opened 2021-09-10
3.9.4

remyperona:
- Enhancement: Update delay JS script to catch click events before the scripts are loaded (#4303)
- Enhancement: Add the delay JS script to the page after the meta charset if it's present (#4121)
- Enhancement: Remove the clear cache notice after using clear used CSS since it also clears the cache (#4212)
- Bugfix: Prevent a possible error on WooCommerce product pages with delay JS enabled (#4255)
- Bugfix: Prevent a possible PHP notice in some cases when using remove unused CSS (#4243) 
- 3rd party compatibility: Improve delay JS compatibility with a number of scripts (#4075, #4241, #4169)


Mai-Saad (approved):
Ready to go!
[testrail-report-256.pdf](https://github.com/wp-media/wp-rocket/files/7161874/testrail-report-256.pdf)

PR #4367, opened 2021-09-20
Update lazyload script version & fix bugs related to WP Retina 2x and the inline script

### LINKED ISSUE #2492: Remove WP Retina x2 Exclusion from the LazyLoad
Images from WP Retina 2x should be LazyLoaded.

1. Go to /wp-content/wp-rocket/vendor/wp-media/rocket-lazyload-common/src/Image.php
2. Comment out the lines 193 to 197.
3. It should look like this: https://i.imgur.com/8F7V9P3.png.
4. The images are LazyLoaded now.

Related original exclusion issue: https://github.com/wp-media/wp-rocket/issues/37

**Additional context**
Slack thread: https://wp-media.slack.com/archives/C08N8J6VC/p1585576241074400.
Ticket: https://secure.helpscout.net/conversation/1113564900/152540?folderId=2675957.

### LINKED ISSUE #3510: Update LazyLoad script to the latest version
Request made by the customer for updating LazyLoad script from version 16.1 to the latest version 17.3. There are some improvements made and the script is also 500 bytes lighter now.

### LINKED ISSUE #3604: Switch to Native LazyLoad for Images

With this update, WP Rocket will switch from using the JavaScript based LazyLoad for images and start supporting native browser LazyLoad by default. 

WP Rocket will still use it’s JavaScript based LazyLoad for background images and iframes since these are not supported by native LazyLoad yet on all browsers.

([full doc](https://docs.google.com/document/d/1z4pkyZdYbOfKQzcEwGrvxuV4ngGaooQyQcxNtOw4yIk/edit?usp=sharing))

### LINKED ISSUE #3976: Lazyload: Uncaught TypeError: Assignment to constant variable when a constant "i" is declared

**Describe the bug**
When a constant "i" is declared before this line (in the browser):

https://github.com/wp-media/wp-rocket/blob/aea8f96b0725f900deaa337af2f87c37277676ff/inc/Dependencies/RocketLazyload/Assets.php#L104

The error Uncaught TypeError: Assignment to constant variable appears in the browsers console

**To Reproduce**
1. Add this test code to the website's <head>:
`			<script>
				 const i = true;
			</script>`
2. Activate Lazyload
3. Clear the cache
4. Try  and Check the browsers console

**Expected behavior**
This error doesn't appear when the page loads.

**Screenshots**
image

**Additional context**
This is caused because the next line is not declaring the variable "i" locally:

` for (i = 0; i < mutation.addedNodes.length; i++) {`

This means that if the variable is already declared / defined in a scope above, that declared variable will be used, but if the variable is not declared in a scope above, then it will be immediately declared in that line.

A possible solution could be adding a "var" keyword before the "i" variable, like this:

` for (var i = 0; i < mutation.addedNodes.length; i++) {`

This will declare a new "i" variable in the local scope, so, the global "i" won't be affected.

I can do the PR if needed.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Fixes #2492
Fixes #3976
Closes #3510

## Type of change

- [x] Bug fix (non-breaking change which improves an existing functionality)

## How Has This Been Tested?

- [x] Manual testing
- [x] Added unit & integration tests

# Checklist:


mostafa-hisham (approved):

iCaspar (approved):

inc/Engine/Media/Lazyload/Subscriber.php
```diff
@@ -15,7 +15,7 @@
  * @since 3.3
  */
 class Subscriber implements Subscriber_Interface {
-	const SCRIPT_VERSION = '16.1';
+	const SCRIPT_VERSION = '17.4';
```
Mai-Saad: @Tabrisrp Currently latest version is 17.5 as per https://github.com/verlok/vanilla-lazyload/releases, Do we need to use this? (not sure if we can have an option where this can be auto updated always as per latest version whatever)

piotrbak (commented):
@Tabrisrp There's a difference in how we're handling the lazy load of iframes.

In my tests, the YT video is not lazy loaded, could it be because we added `[loading=lazy]` to the script (iframe contains both, `loading=lazy` and `data-lazy-src`)?

remyperona:
@piotrbak I pushed a change to avoid changes to the iframe behaviour, it's ready to be tested again.

inc/Engine/Media/Lazyload/Subscriber.php
remyperona: Updated the script version to 17.5. We can't auto-update it, we have to manually handle it.

Mai-Saad:
@Tabrisrp  While trying to use new test template and updating tests for LL, for this page https://new.rocketlabsqa.ovh/images_picture_template_003/, found the following:
1- LL isnot working when checked network tab, any idea why? (note: working fine on trunk)
2- We are adding` loading = "lazy"`  even if there was `skip-lazy`  or `data-skip-lazy=" "` or `class="skip-lazy"` , I think we shouldn't add `loading = " lazy "` in these cases. what do you think?

remyperona:
@Mai-Saad
1. With native lazyload the loading of the images is handled by the browser directly, the threshold for loading them is likely different than what we had before, so images might be loaded sooner than before.
2. This is part of the requirements, we removed those exclusions from the default list when using native lazyload

remyperona:
The preview image for Youtube videos should be correctly displayed now when using the corresponding option, and the image has the `loading="lazy"` attribute.

When native lazyload is disabled (with the filter), it falls back to the previous markup, using `data-lazy-src` instead.

piotrbak:
@Tabrisrp The tresholds in native LL are still very big, but I guess they know what they're doing:
https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/frame/settings.json5;l=951-979

The other thing that we most likely experienced is related to the loading of images + styling. Basically, if images are displayed one to each other and later put in the order by the styling (float:left) for example. They will be loaded immediately. This is what is happening in the template provided by Mai. The width of the page is 8000px and all images are loaded probably because styling is happening slowly and all images are within this big treshold initially.

It wasn't happening before, as JavaScript that was loading images was very late in the source, so images were styled already when we triggered the script.

remyperona:
@piotrbak At this point we're letting the browser handle everything related to lazyload for images. Is it creating any warning in PageSpeed related to the images? I believe this is the benchmark to use to validate the change (cc @GeekPress)

GeekPress:
I suppose this is expected behavior from the browser to be compatible with sliders or other stuff like that.

The vertical threshold is only 1250px.

By the way, our users will still be able to use the JS solution if they prefer by using our filter `rocket_use_native_lazyload` as mentioned on the epic card.

remyperona:
Updated this PR to remove the native lazyload changes (but preserving the other updates and new tests added).

iCaspar (approved):

piotrbak (approved):
Good to go
[testrail-report-290.pdf](https://github.com/wp-media/wp-rocket/files/7429017/testrail-report-290.pdf)

PR #4375, opened 2021-09-22
Fix some minor issues with new OptimizeCSS UI

iCaspar:
## Description
This fixes a couple minor issues with the new Optimize CSS noted by Product and Support team on the 1.10-alpha1:

- [X] Missing text for RUCSS description: "Recommended for best performance. Test Thoroughly!"
- [x] Unnecessary Warning/confirmation for RUCSS appears when switching UI back from CPCSS option, when RUCSS is already active.

## Type of change

## How Has This Been Tested?
Locally in Rocket settings page.

# Checklist:


mostafa-hisham (commented):
Nice Work @iCaspar
I just edited the code for `wprShowRadioWarning` so it can work with all radio buttons fields in case we add new ones in the future
Can you please have a look?

Mai-Saad:
@mostafa-hisham @iCaspar Can we please remove `for better performance` after `Removes unused CSS per page and helps to reduce page size and HTTP requests`
Note: we need to add supported translations as well

iCaspar:
Can't approve my own PR, but @mostafa-hisham changes look good, and text is updated per @Mai-Saad review.
Moving this to ready for @wp-media/qa 
Translations will be added as transifex is updated.

Mai-Saad (approved):
[debug.log](https://github.com/wp-media/wp-rocket/files/7222439/debug.log)

Note: The fatal error when switch to the trunk without deactivating the plugin appears here.. isn't the fix merged to this branch? (same case when switch develop to trunk)  => will be handled here #4381 fix class aliases in deprecated 3.10

PR #4377, opened 2021-09-23
Closes #4320 combine CSS divi 4.10 compatibility

### LINKED ISSUE #4320: Divi 4.10+ compatibility improvement

**Describe the bug**
When we have specific settings enabled, we'll change the order of the styles. Which might cause the layout problems. Problematic setup:
1. WP Rocket's **CSS Combination** enabled
2. WP Rocket's **RUCSS** disabled
3. Divi's **Load Dynamic Stylesheet Inline** enabled

This is happening because we're not combining internal styles `<style></style>`. This is how it works for a long time, however, Divi 4.10+, for any reason, is adding some important styles to CSS files and some others (that could be related to the same elements) to internal styles.

**To Reproduce**
1. Enable settings as described above
2. Check the source of the cached page
3. The combined CSS file is at the very top
4. Check the source of the not cached page
5. Divi CSS files are below the internal styles - **the order is changed!**

**Expected behavior**
We should preserve Divi's CSS order. We could:
1. Disable their feature when RUCSS is disabled + Combine CSS is enabled
2. Disable our Combine CSS when their option is enabled + RUCSS is disabled
3. Exclude their files from combination when RUCSS is disabled

**Screenshots**

**Additional context**

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

mostafa-hisham:
## Description

 exclude all Divi CSS files which are stored in /wp-content/et-cache/ From being Combined

Fixes #4320 

## Type of change

No
## How Has This Been Tested?

- [x] Tested locally

# Checklist:


iCaspar (approved):

Mai-Saad:
Well, I can't see a UI problem when the Combine CSS is enabled while Load Dynamic Stylesheet Inline is enabled at Divi, however, while checking that we keep Divi CSS order, here is the current state/questions:

1- When  Load Dynamic Stylesheet is enabled => All Divi related style sheets under the inline CSS are kept in the source (Is it expected that the sheet here is normally combined 
`<meta content="Divi v.4.10.7" name="generator"/><link rel='stylesheet' id='wp-block-library-css'  href='https://new.rocketlabsqa.ovh/wp-includes/css/dist/block-library/style.min.css?ver=5.8.1' type='text/css' media='all' /> ` also it was below an inline style  `<style id="et-divi-open-sans-inline-css">`?) => we agreed that this is fine

2- When Critical CSS is enabled while Load Dynamic Stylesheet is disabled => This file is combined  `<link rel='stylesheet' id='divi-style-css'  href='https://new.rocketlabsqa.ovh/wp-content/themes/Divi/style.min.css?ver=4.10.7' type='text/css' media='all' />` also it was below an inline style  `<style id="et-divi-open-sans-inline-css">` (order changed here, so shouldn't we keep this file uncombined?) => we agreed that this is fine (combine styles that was under fonts inline)

inc/ThirdParty/Themes/Divi.php
```diff
@@ -172,4 +174,24 @@ public function disable_divi_jquery_body() {
...
+	 *
+	 * @param array $exclude_css An array of CSS  to be excluded.
+	 *
+	 * @return array the updated array of paths
+	 */
+	public function exclude_divi_css_from_combine( $exclude_css ) {
+
+		if ( (bool) $this->options->get( 'remove_unused_css', 0 ) ) {
+			return $exclude_css;
+		}
```
remyperona: We don't need to check for the `remove_unused_css` value. If RUCSS is enabled, there won't be any minify/combine happening anyway.

What we need to check for is if `minify_concatenate_css` is enabled, and bail-out if it isn't.

inc/ThirdParty/Themes/Divi.php
```diff
@@ -172,4 +174,24 @@ public function disable_divi_jquery_body() {
...
+
+		$wp_content    = wp_parse_url( content_url( '/' ), PHP_URL_PATH );
```
remyperona: `wp_parse_url()` can return `null` when using the second parameter. We should check if `$wp_content` is not null before using it on the line after.

remyperona (changes requested):
Test & fixture should be updated based on the review's comments.

inc/ThirdParty/Themes/Divi.php
```diff
@@ -172,4 +174,27 @@ public function disable_divi_jquery_body() {
...
+	 *
+	 * @since 3.10
+	 *
+	 * @param array $exclude_css An array of CSS  to be excluded.
+	 *
+	 * @return array the updated array of paths
+	 */
+	public function exclude_divi_css_from_combine( $exclude_css ) {
+
+		if ( (bool) $this->options->get( 'minify_concatenate_css', 0 ) ) {
```
remyperona:
```suggestion
		if ( ! (bool) $this->options->get( 'minify_concatenate_css', 0 ) ) {
```

We should bail-out if the option is not enabled

remyperona (changes requested):
need to update the fixtures and test too

remyperona (approved):

Mai-Saad (approved):
Working as expected
[testrail-report-265.pdf](https://github.com/wp-media/wp-rocket/files/7287343/testrail-report-265.pdf)

Notes:
We agreed that the following is fine
1- While Critical CSS is enabled at Divi , we will combine Divi style file although it's under divi inline fonts
2- If both Load CSS and critical CSS are disabled at Divi, there will be no inline for Divi styles and we will still exclude Divi files from Combine CSS

PR #4381, opened 2021-09-24
fix class aliases in deprecated 3.10

mostafa-hisham:


remyperona (approved):

Mai-Saad:
log sample for the error we are fixing here
[debug (1).log](https://github.com/wp-media/wp-rocket/files/7225767/debug.1.log)
Scenario:
1- Install develop
2- Enable CPCSS
3- Switch to the trunk
4-  Run in cmd `composer dumpautoload`

remyperona:
@wp-media/qa ready for QA

Mai-Saad:
Still seeing Error in debug.log when switch from the current branch to trunk without deactivating WPR before switching branch.
[debug.log](https://github.com/wp-media/wp-rocket/files/7233058/debug.log)
Note: if we deactivated WPR plugin before switching to the trunk, nothing added to debug .log
**Update**:  switching while there is a process in the background i.e OCD or RUCSS not done yet is resulting in the error while if no process in the background then no error while switching

PR #4413, opened 2021-10-07
Fixes #4397 Guard against preg_replace() errors when moving meta charset to head

### LINKED ISSUE #4397: Delay JS - PREG_BACKTRACK_LIMIT_ERROR when moving meta to the head
- Made sure you’re on the latest version ✅ `3.9.10` but it should be the same with `>3.9.4`.

**Describe the bug**

On a customer's website, an `error 500` occurs for the homepage, when Delay JavaScript Execution is active.

This is because a `PREG_BACKTRACK_LIMIT_ERROR` error pop-ups here:
https://github.com/wp-media/wp-rocket/blob/0bdb9fced6dd474ce7f5e4b648a65b70d2500a0b/inc/Engine/Optimization/DelayJS/HTML.php#L220

This isn't happening on other pages.

The `backtrack.limit` is set to `1000000`.

**To Reproduce**

The raw HTML of the page is in the ticket's note. Using that in a template will likely trigger the error.

**Expected behavior**

We should fail gracefully when backtrack errors occur.

**Additional context**

**Ticket:** https://secure.helpscout.net/conversation/1645011344/297785/#thread-4801024582

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

This PR adds guards to `move_meta_charset_to_head()`, to always return a string in case one of the `preg_replace()` encounters an error and returns `null` instead of the expected string value.

Fixes #4397

## Type of change

No grooming done

## How Has This Been Tested?

- [x] Automated tests

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DelayJS/HTML.php
+++ b/inc/Engine/Optimization/DelayJS/HTML.php
@@ -216,16 +216,43 @@ class HTML {
 	 */
 	public function move_meta_charset_to_head( $html ): string {
 		$meta_pattern = "#<meta[^h]*(http-equiv[^=]*=[^\'\"]*[\'\" ]Content-Type[\'\"][ ]*[^>]*|)(charset[^=]*=[ ]*[\'\" ]*[^\'\"> ][^\'\">]+[^\'\"> ][\'\" ]*|charset[^=]*=*[^\'\"> ][^\'\">]+[^\'\"> ])([^>]*|)>(?=.*</head>)#Usmi";
-		if ( preg_match( $meta_pattern, $html, $matches ) ) {
-			$html = preg_replace( "$meta_pattern", '', $html );
-			if ( preg_match( '/<head\b/i', $html ) ) {
-				$html = preg_replace( '/(<head\b[^>]*?>)/i', "\${1}${matches[0]}", $html );
-			} elseif ( preg_match( '/<html\b/i', $html ) ) {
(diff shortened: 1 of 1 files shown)


mostafa-hisham (approved):

iCaspar (approved):

Mai-Saad (commented):
Using the page source mentioned here https://github.com/wp-media/wp-rocket/issues/4397#issue-1010966855, can see now that no 500 error in console while delay js is enabled with/without default exclusions (no errors now in debug.log as well). but can see a console error when minify/combine js is enabled (on both trunk and the PR), @Tabrisrp  do we need to handle anything for that? or is that expected from the test page?
`Uncaught SyntaxError: Unexpected identifier` for both wp-embed.min.js and jquery-migrate.min.js

remyperona:
@Mai-Saad This is probably coming from the test page, it doesn't seem to be related to the change in this PR.

piotrbak:
@Tabrisrp It works liek expected. I do have a question though:
1. I manipulated `pcre.backtrack_limit` and set it to low limit
2. Visited the page **without** RUCSS enabled
3. Received this error: https://snippi.com/s/5lck17t

Should anything be happening related to RUCSS when the feature is disabled?

piotrbak:
The PR itself is fine:
[testrail-report-275.pdf](https://github.com/wp-media/wp-rocket/files/7337587/testrail-report-275.pdf)

piotrbak (approved):

Issue #4077, opened 2021-06-20
Auto-exclusion for WooCommerce Product Gallery when Delay JS is on


GeekPress:
**Is your feature request related to a problem? Please describe.**
If we delay all JS files, the product image on a single product page won't be displayed until user interaction.

Here an example:
https://share.getcloudapp.com/9ZuddXp2

As you can see, it's not really a good user experience and even more on mobile.

**Describe the solution you'd like**
The product image on a single product page should be displayed without any user interaction when Delay JS is on.

We need to handle two cases:

1. when the product page doesn't have a gallery
2. when the product page has a gallery

--

1. In this case, we don't need to exclude any JS files from being delayed. We can still delay the WooCommerce JS files in this case if we had a small inline CSS in the header:

`<style>.woocommerce-product-gallery{ opacity: 1 !important; }</style>
`

2. In this case, when we are on a product page, we need to exclude these files from being delayed:

```
/jquery-?[0-9.]*(.min|.slim|.slim.min)?.js
/woocommerce/assets/js/zoom/jquery.zoom.min.js
/woocommerce/assets/js/photoswipe/
/woocommerce/assets/js/flexslider/jquery.flexslider.min.js
/woocommerce/assets/js/frontend/single-product.min.js
```

bwafels:
I have installed the latest update of wp rocket and I can confirm this is not a nice behaviour. On mobile screens the product image take up a large part of the screen. This means customers may get the impression the website is broken.

engahmeds3ed:
I did some tests and reached the following code that solves the issue
```
add_action( 'wp_enqueue_scripts', function() {
	if ( ! is_product() ) {
		return;
	}

	$product = wc_get_product( get_the_ID() );
	$gallery_images = $product->get_gallery_image_ids();
	if ( ! empty( $gallery_images ) ) {
		return;
	}

	$custom_css = ".woocommerce-product-gallery{ opacity: 1 !important; }";
	wp_add_inline_style( 'woocommerce-layout', $custom_css );
} );

add_filter( 'rocket_delay_js_exclusions', function( $exclusions ) {
	if ( ! is_product() ) {
		return $exclusions;
	}

	$product = wc_get_product( get_the_ID() );
	$gallery_images = $product->get_gallery_image_ids();
	if ( empty( $gallery_images ) ) {
		return $exclusions;
	}

	$exclusions[] = '/jquery-?[0-9.]*(.min|.slim|.slim.min)?.js';
[...]

PR #4423, opened 2021-10-13
Fixes #4224 Add exclusions from delay JS execution for Avada compatibility

### LINKED ISSUE #4224: 3.9.1.1 - Delay JS exclusions for WooCommerce cause issues on Avada
**Describe the bug**
On a site using Avada, updating from 3.9.0.5  → 3.9.1.1 breaks the layout of the WooCommerce product pages due to the changes here: https://github.com/wp-media/wp-rocket/issues/4077

When using 3.9.0.5 the customer had no exclusions for Delay JS and there were no issues (images displayed correctly on interaction) and no console errors. 

After updating to 3.9.1.1 the images are displayed vertically at full size instead of thumbnails: 
http://recordit.co/954umFKZ10
Correct display should be: https://jmp.sh/4moYhbK

This can be corrected by excluding the following files: 
/wp-content/themes/Avada/assets/min/js/general/avada-woo-product-images.js
/wp-content/themes/Avada/includes/lib/assets/min/js/library/jquery.flexslider.js

BUT then this leads to jquery errors on other pages where previously there were none.

**Expected behavior**
Updating the plugin should not break the layout or require further exclusions. 

**Additional context**

- Customer in this case is on a multisite. 
- Theme is Avada
- Issue can be seen on customer's staging site, found in my note here: 
https://secure.helpscout.net/conversation/1586750255/283985?folderId=377611#thread-4632327399

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Exclude 2 Avada scripts from delay JS execution to prevent display issue on the WooCommerce product gallery.

Fixes #4224 

## Type of change

No

## How Has This Been Tested?

- [x] Added integration test for the new method adding the exclusions

# Checklist:

### DIFF
--- a/inc/ThirdParty/Plugins/Ecommerce/WooCommerceSubscriber.php
+++ b/inc/ThirdParty/Plugins/Ecommerce/WooCommerceSubscriber.php
@@ -550,7 +550,7 @@ class WooCommerceSubscriber implements Event_Manager_Aware_Subscriber_Interface 
 	 *
 	 * @return array
 	 */
-	public function show_notempty_product_gallery_with_delayJS( array $exclusions = [] ) {
+	public function show_notempty_product_gallery_with_delayJS( $exclusions = [] ): array {
 		global $wp_version;
 
 		if ( ! $this->delayjs_html->is_allowed() ) {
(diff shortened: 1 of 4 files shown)


iCaspar (approved):

piotrbak:
@Tabrisrp It looks like the problem with display is there only when product has the gallery. However, those files are loaded always, no matter if image has gallery, thumbnail, etc.

Maybe we could exclude those files only if gallery is there, just like we did here?
https://github.com/wp-media/wp-rocket/issues/4077

remyperona:
I updated the code to add a new filter specifically for the files we exclude from delay JS on the WC product page when it has a gallery with images. And use this filter in Avada instead of the generic one.

piotrbak (approved):
Good to go!
[testrail-report-279.pdf](https://github.com/wp-media/wp-rocket/files/7373158/testrail-report-279.pdf)

PR #4428, opened 2021-10-14
Fixes #4331 Prevent removing style tags inserted with inline JS when RUCSS is enabled

### LINKED ISSUE #4331: CSS rules added with inline JS script $("head").append are removed and not added to the Used CSS
- Made sure you’re on the latest version - 3.9.3

**Describe the bug**
When _Remove Unused CSS_ option is enabled, the CSS rules that are added with inline JS script `$("head").append` are being removed from the source code and not added to the Used CSS. This leads to a broken layout.

**To Reproduce**

1. Add the script with CSS rules to the head as in example here:
```
<script src="https://code.jquery.com/jquery-3.6.0.min.js" integrity="sha256-/xUj+3OJU5yExlq6GSYGSHk7tPXikynS7ogEvDej/m4=" crossorigin="anonymous"></script>
	(function($) {$("head").append("<style>.my-style{color:red;}</style>")})(jQuery);
</script>
```

2. Add an HTML element on a page like here:
`<h3 class="my-style">Hello world</h3>`
3. Check the page and make sure the style is applied.
4. Enable _Remove Unused CSS_ and wait until the process is completed and the used-css is generated.
5. Check the page in the incognito browser window - the style is not applied. 
It is removed, but not added to the Used CSS.
image
Screenshot: https://jmp.sh/7KLsoIj

Safelisting `.my-style` doesn't resolve the issue.

**Expected behavior**
CSS rules added with the inline script must be processed and added to the Used CSS the same way as with inline styles.

**Additional context**
Related ticket with notes: https://secure.helpscout.net/conversation/1620666496/292153/#thread-4732067758

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

This PR prevents the removal of style tags inserted with inline JS when RUCSS is enabled. It also cleans up the code where needed, notably in tests after the changes in 3.10

Fixes #4331

## Type of change

No

## How Has This Been Tested?

- [x] Updated fixture to add an inline script inserting a style

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
+++ b/inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
@@ -83,7 +83,7 @@ abstract class AbstractAPIClient {
 	}
 
 	/**
-	 * Handle Saas request error.
+	 * Handle SaaS request error.
 	 *
 	 * @param array|WP_Error $response WP Remote request.
 	 *
--- a/inc/Engine/Optimization/RUCSS/Admin/Settings.php
+++ b/inc/Engine/Optimization/RUCSS/Admin/Settings.php
@@ -5,7 +5,6 @@ namespace WP_Rocket\Engine\Optimization\RUCSS\Admin;
 
 use WP_Rocket\Admin\Options_Data;
 use WP_Rocket\Engine\Admin\Settings\Settings as AdminSettings;
-use WP_Rocket\Engine\Admin\Settings\Render;
 
 class Settings {
 	/**
--- a/inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
+++ b/inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
@@ -4,10 +4,8 @@ declare(strict_types=1);
 namespace WP_Rocket\Engine\Optimization\RUCSS\Admin;
 
 use WP_Rocket\Admin\Options;
-use WP_Rocket\Admin\Options_Data;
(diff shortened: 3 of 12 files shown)


iCaspar (approved):

piotrbak:
@Tabrisrp In specific circumstances we can change the order of the styles. Will check if that's the case with reported tickets. For example:
```
<head>
<script src="https://code.jquery.com/jquery-3.6.0.min.js" integrity="sha256-/xUj+3OJU5yExlq6GSYGSHk7tPXikynS7ogEvDej/m4=" crossorigin="anonymous"></script>

<style>.my-style{color:black;}</style>

	(function($) {$("head").append("<style>.my-style{color:red;}</style>")})(jQuery);
</script>
<title>Title</title>
```

It'll move the style after `<title>` tag to Used CSS but the style added by script will be added higher in the source.

remyperona:
This seems very edge case though right?

GeekPress:
Indeed, this seems to be an edge case. I don't know for you, on my side I never saw an inline script inserted before the title tag.

I guess the reason is becaise inline CSS are injected by using wp_head hook or wp_add_inline_script function. Both solutions  will mandatory insert the inline CSS after the title tag.

piotrbak (approved):
Good to go!
[testrail-report-281.pdf](https://github.com/wp-media/wp-rocket/files/7374761/testrail-report-281.pdf)

PR #4432, opened 2021-10-15
Allow overriding the SaaS API URL

iCaspar:
## Description

Internal improvement to allow overriding the SaaS API URL from `wp-config.php` by setting the constant:
`WP_ROCKET_SAAS_API_URL`
(So we don't have to keep stashing this change during testing on local/develop envs.)

## Type of change

## Checklist:


GeekPress:
@iCaspar Why `WP_Media` rather than `WP_Rocket`?

iCaspar:
@GeekPress 😄 
Thinking of SaaS in terms of a company-wide item.
But you're right -- should be ROCKET.

tests/Unit/inc/Engine/Optimization/RUCSS/Warmup/APIClient/sendWarmupRequest.php
```diff
@@ -53,10 +53,14 @@ public function testShouldSendWarmupRequest( $atts, $success, $mockResponse ): v
 		];
 
 		Functions\when( 'wp_parse_args' )->returnArg( 1 );
+
+		$api_url = rocket_get_constant( 'WP_ROCKET_SAAS_API_URL', false )
+			? rocket_get_constant( 'WP_ROCKET_SAAS_API_URL', false )
```
remyperona: This should use the Stub method available in `StubTrait` instead of using the function directly

remyperona (changes requested):

tests/Unit/inc/Engine/Optimization/RUCSS/Warmup/APIClient/sendWarmupRequest.php
iCaspar: I think this works now.
Added a default return value for the new constant to the StubTrait.

remyperona (approved):

Mai-Saad (approved):
Working as expected

PR #4445, opened 2021-10-20
Fixes #4337 Only try to clear RUCSS scheduled event if it is scheduled

### LINKED ISSUE #4337: Add wp_next_scheduled() check for rocket_rucss_clean_rows_time_event
Feedback from the customer: https://secure.helpscout.net/conversation/1624163812/292950

The current implementation calls `wp_clear_scheduled_hook` each time the WP init hook is called.

The reason is this code:

```
public function schedule_clean_not_commonly_used_rows() {
  if ( ! $this->settings->is_enabled() ) {
   wp_clear_scheduled_hook( 'rocket_rucss_clean_rows_time_event' );

  return;
}
```

It would be better to check for a scheduled event before trying to unschedule something. Like this:

```
public function schedule_clean_not_commonly_used_rows() {
  if ( ! $this->settings->is_enabled() ) {
   if ( wp_next_scheduled( 'rocket_rucss_clean_rows_time_event' ) ) {

    wp_clear_scheduled_hook( 'rocket_rucss_clean_rows_time_event' );
  }

return;
}
```

Slack thread: https://wp-media.slack.com/archives/C029G1B5HD2/p1631088167085200

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Previously we were only checking if the RUCSS option is enabled before calling the unschedule function. So that call would happen every time a page is loaded and the RUCSS is disabled, doing unnecessary processing.

With that change, it will be called only if RUCSS is disabled AND there is an event scheduled.

Fixes #4337

## Type of change

A bit, as it checks for the scheduled event in addition to the setting being disabled in the same condition.

# Checklist:

### DIFF
--- a/inc/Engine/Deactivation/Deactivation.php
+++ b/inc/Engine/Deactivation/Deactivation.php
@@ -89,8 +89,7 @@ class Deactivation {
 		delete_site_transient( 'update_wprocket_response' );
 
 		// Unschedule WP Cron events.
-		wp_clear_scheduled_hook( 'rocket_facebook_tracking_cache_update' );
-		wp_clear_scheduled_hook( 'rocket_google_tracking_cache_update' );
+		wp_clear_scheduled_hook( 'rocket_rucss_clean_rows_time_event' );
 		wp_clear_scheduled_hook( 'rocket_cache_dir_size_check' );
 
 		/**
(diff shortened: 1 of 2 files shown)


engahmeds3ed (approved):
Just check the failed tests, It's not related to this PR, related to smush integration tests

Mai-Saad (approved):
Working as expected
[testrail-report-288.pdf](https://github.com/wp-media/wp-rocket/files/7424817/testrail-report-288.pdf)

Issue #3998, opened 2021-06-04
3.9 - WooCommerce Multilingual compatibility change causing errors 50x


vmanthos:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version ✅ `3.9.0.1`
- Used the search feature to ensure that the bug hasn’t been reported before ✅

**Describe the bug**

The compatibility change that was introduced in WP Rocket 3.9:
https://github.com/wp-media/wp-rocket/issues/3418

is causing issues on customers' servers, e.g. error 502, when trying to open the products page.

WCML is trying to set the `wcml_client_currency` cookie numerous times by sending the `Set-Cookie` header.
In the case I worked on that took place [584 times](https://snippi.com/s/ykksqa8)! 

That caused an` ERR_RESPONSE_HEADERS_TOO_BIG` error in a ticket I debugged, and `error 502` in cases the customers reported.

When I commented out the following line in WP Rocket's compatibility code, the headers I mentioned previously were not sent, and the issue wasn't there:
https://github.com/wp-media/wp-rocket/blob/59045e5ace9948b5ad844c6f4d5c46122089af7b/inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php#L15

**To Reproduce**

1. Use WP Rocket 3.9.
2. Install WooCommerce Multilingual.
3. Check the headers for the HTML file. They should have several references of `Set-Cookie`.

**Expected behavior**

No error 50x or similar should occur when WP Rocket is used with WCML.

**Additional context**

 @piotrbak [suggested](https://wp-media.slack.com/archives/C43T1AYMQ/p1622800555211700?thread_ts=1622540018.164300&cid=C43T1AYMQ) we open this issue to discuss if there is anything we should be doing on our side.

**Slack thread:** https://wp-media.slack.com/archives/C43T1AYMQ/p1622540018164300

**Tickets:**
https://secure.helpscout.net/conversation/1526046602/267321
https://secure.helpscout.net/conversation/1531446262/268694
https://secure.helpscout.net/conversation/1531659256/268726

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

strategio:
Thanks for pinging me @piotrbak.

It looks like we have something to improve on the way we deal with cookies.
[...]

PR #4462, opened 2021-10-27
Changes required for WCML integration

### LINKED ISSUE #4070: Revert Changes related to #3998 - Reenable WCML Currency Cookies.
## Describe the bug
We have temporarily commented out the `wcml_user_store_strategy` filter to prevent generating duplicate cookies on sites using WCML. This change needs to be reverted after an anticipated fix to the duplicate cookie issue is released by WCML.

The missing filter will causing a known side-effect (bug) that we cannot read the currrent locale/currency information on cached pages. We implemented this because we anticipate the temporary side-effect is not as severe as  the cookie duplication on large sites where it causes 50x errors due to excessive headers.

@see #3998

- [X] Reproduce the problem
- [X] Identify the root cause
- [X] Scope a solution
- [X] Estimate the effort

### Reproduced
Yes, locally.

### Root Cause
An intentional temporary disabling of the `wcml_user_store_strategy` filter that needs to be reverted at
https://github.com/wp-media/wp-rocket/blob/4b22e022aebb2517f29b5e31fd47899829256e9e/inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php#L15

### Effort Estimate
[XS]

strategio:
After some discussions with @piotrbak today, I ran some tests and adjusted the integration code:
1. The activation/deactivation hooks were missing the respective prefixes.
2. We can't keep the function definitions inside the `if` because it's required on WCML activation (and we also had to extract the activation logic).
3. Re-enabled the "cookie storage" (See #3998).

---

Test results:

# 1. WP Rocket is active, we're activating WCML
WCML dynamic cookies are added to config file (/wp-content/wp-rocket-config/yourdomain.php), they should be listed in "$rocket_cache_dynamic_cookies" array
If Geolocation was set in WCML settings before, we should also add a mandatory cookie to "$rocket_cache_mandatory_cookies" array

## Partially working
=> "If Geolocation was set in WCML settings before, we should also add a mandatory cookie to "$rocket_cache_mandatory_cookies" array". The mandatory cookie is missing because the plugin is activated on `plugins_loaded`, which fires `activate_woocommerce-multilingual/wpml-woocommerce.php` which fires `rocket_generate_config_file`. At this stage, the filter for `wcml_geolocation_is_used` is not loaded yet (it's loaded on `wpml_loaded`).

IMHO, WP Rocket should set a flag (an option) to regenerate the config on the next admin request (at a later priority, e.g. `admin_init`).

# 2. WCML is active, we're activating WP Rocket
The same things are happening as in the 1st point

## Confirmed

# 3. We're changing the Geolocation to Site Language

We're removing mandatory cookie from "$rocket_cache_mandatory_cookies" array

## Confirmed

# 4. We're deactivating WCML

We're removing mandatory cookie from "$rocket_cache_mandatory_cookies" array
We're removing dynamic cookies from "$rocket_cache_mandatory_cookies" array

## Confirmed

# 5. We're changing specific WCML options

Config file is recreated

## Confirmed when enabling the multi-currency or changing the mode (by location / by language).

---

So we have 1 small glitch left but I don't think it should be solved inside the integration.

### DIFF
--- a/inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php
+++ b/inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php
@@ -3,6 +3,8 @@
 defined( 'ABSPATH' ) || exit;
 
 /**
+ * Check if WCML is active and has minimum requirements.
+ *
  * @return bool
  */
 function rocket_wcml_has_requirements() {
--- a/inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php
+++ b/inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php
@@ -2,11 +2,6 @@
 
 defined( 'ABSPATH' ) || exit;
 
-// Compatibility with the currency switcher in WooCommerce Multilingual plugin.
-// @since 3.9.0.3
-// We are _temporarily_ disabling this filter, to be re-enabled after a future WCML release.
-// that contains a better fix for the duplicate generation of cookies on this filter.
-// @see #3998.
 if ( defined( 'WCML_VERSION' ) ) :
 	/**
 	 * Use Cookie instead of WCSession
@@ -16,7 +11,7 @@ if ( defined( 'WCML_VERSION' ) ) :
 	function rocket_wcml_use_cookie_storage() {
 		return 'cookie';
 	}
-	/* add_filter( 'wcml_user_store_strategy', 'rocket_wcml_use_cookie_storage', 10, 2 ); */
+	add_filter( 'wcml_user_store_strategy', 'rocket_wcml_use_cookie_storage', 10, 2 );
 
 	add_filter( 'rocket_cache_dynamic_cookies', 'rocket_wcml_add_dynamic_cookies' );
 	add_filter( 'rocket_cache_mandatory_cookies', 'rocket_wcml_add_mandatory_cookies' );
--- a/inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php
+++ b/inc/3rd-party/plugins/ecommerce/woocommerce-multilingual.php
@@ -99,7 +99,7 @@ if ( defined( 'WCML_VERSION' ) ) :
 		flush_rocket_htaccess();
 		rocket_generate_config_file();
 	}
-	add_action( 'woocommerce-multilingual/wpml-woocommerce.php', 'rocket_wcml_activate', 11 );
+	add_action( 'activate_woocommerce-multilingual/wpml-woocommerce.php', 'rocket_wcml_activate', 11 );
 
 	/**
 	 * Reset WP Rocket settings on WCML deactivation.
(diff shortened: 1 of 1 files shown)


remyperona:
@piotrbak What do you think? Should we move forward with this, and we can improve the one partially working point later?

piotrbak:
@Tabrisrp Yes, I think we can move forward with this. I'm thinking whether we should consider users that are using old versions of WCML / WPML or not and ask them for updates. This are two scenarios that 
1. pre 4.11 WCML -> we'll set up a mandatory cookie for them that will never exist in the browser
2. 4.11 and newer WCML + pre 4.4.11 WPML ->We'll have exactly the same problem https://github.com/wp-media/wp-rocket/issues/3998
3. For the updated versions everything should go as expected (still needs QA)

remyperona:
We can recommend to users with any of these issues to update to the latest version available for both WPML and WCML?

piotrbak:
Yes, assuming that they have not expired subscription for those products. If not, then our product will not be usable at all (mandatory cookie or broken WooCommerce product list).

remyperona:
Should a check be added in the compatibility code for minimum specific versions of each? Is it possible?

piotrbak:
For WCML we have a constant `WCML_VERSION` and for WPML we have `ICL_SITEPRESS_VERSION`. I think we could do this within our compatibility code 🤔

strategio:
@Tabrisrp, @piotrbak, I've improve the conditions to load the integration code, only with WPML 4.4.11 and WCML 4.12.0.

I'll let you check and run some more tests.

remyperona (approved):

engahmeds3ed:
We need to fix phpcs and tests before merging this PR.

@strategio can u plz try fixing them? as we don't have proper permission to do so at your PR directly?

OR we will need to attach another PR after merging this one to fix them.

strategio:
Hi @engahmeds3ed, it would be great if you could address the remaining issues.

I've have added you and @Tabrisrp and @piotrbak as contributors on our fork. I believe you should be able to alter my branch for this PR.

engahmeds3ed:
Many thanks @strategio will check that today.

iCaspar:
Failing tests here are due to ENV items not available when running the tests from non-WP Media repo. Pulling locally and running passes. So this can go to QA and will be good to merge.
Ping @wp-media/qa

piotrbak:
We'll need to bump WCML version check to 4.12.6.

Mai-Saad:
Most of the cases are working fine now but there is a case that is not handled:
1- Enable WCML/ WPML old versions, and Enable WPR new version
2- update both WCML/WPML (while client location used), clear WPR cache => no cookies added to our config file (edit currency or reactivate plugin will add the cookies)
Note: The same problem occurs if rollback WCML/WPML while WPR latest, cookies will remain there in the config file unless we reactivate the plugin.
@piotrbak Do you think we need to change anything here? or this can be handled later?

Mai-Saad:
As per discussion with @piotrbak, we will merge this one and enhance the missing scenario(s) later.

PR #4470, opened 2021-10-29
Closes #3526 display notice if the global exclusion pattern is used in never cache URLs field

### LINKED ISSUE #3526: Display a warning when the user adds "/" into the Never Cache URL(s) box
From time to time we get tickets where the customers claim cache is not working, performance is not improving, Lazyload doesn't work, etc,  and the cause is just a / or some form of global exclusion added in the **Never Cache URL(s) box**. 
 
When we detect a global exclusion is being used, display a notification message to the user on the WP Rocket settings page. 

For Example:
Captura de pantalla 2021-01-27 a la(s) 11 27 54

_full image  https://jmp.sh/i0WWnCB_

- Make the warning dismissable. 
- Instead of a notification, add a warning below the exclusion box, in red letters as we do in other cases for compatibility notices

**Additional context**
Example tickets: 
https://secure.helpscout.net/conversation/1365587907/222760?folderId=2683093
https://secure.helpscout.net/conversation/1305427220/201250#thread-3773680307
https://secure.helpscout.net/conversation/1365587907/222760?folderId=2683093#thread-4063587537

remyperona:
## Description

With this PR, a notice is displayed, and the value removed from the field, if the pattern `/(.*)` is used in the never cache URLs textarea.

Fixes #3526

before

after

## Type of change

No

## How Has This Been Tested?

- [x] Manual testing as shown in screenshots

# Checklist:

### DIFF
--- a/inc/Engine/Admin/Settings/Settings.php
+++ b/inc/Engine/Admin/Settings/Settings.php
@@ -697,6 +697,6 @@ class Settings {
 
 		add_settings_error( 'general', 'reject_uri_global_exclusion', __( 'Sorry! Adding /(.*) in Advanced Rules > Never Cache URL(s) was not saved because it disables caching and optimizations for every page on your site.', 'rocket' ) );
 
-		return array_diff_key( $field, array_keys( $field, '/(.*)', true ) );
+		return array_diff_key( $field, array_flip( array_keys( $field, '/(.*)', true ) ) );
 	}
 }
--- a/inc/Engine/Admin/Settings/Settings.php
+++ b/inc/Engine/Admin/Settings/Settings.php
@@ -263,6 +263,7 @@ class Settings {
 		// Option : Never cache the following pages.
 		if ( ! empty( $input['cache_reject_uri'] ) ) {
 			$input['cache_reject_uri'] = rocket_sanitize_textarea_field( 'cache_reject_uri', $input['cache_reject_uri'] );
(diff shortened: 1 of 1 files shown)


iCaspar (approved):

Mai-Saad:
@Tabrisrp Can you please check these cases:
- If we Never Cache 
```
/test
.*
```
`.*` will be removed but notice won't be displayed => Notice shall be displayed
- If we Never Cache 
```
/test
/.*
```
`/test` will be removed and notice will be displayed => `/test` shall remain in the field, `/*` shall be removed and notice shall be displayed
Screenshot from 2021-11-01 13-41-13


- Q: Currently notice is auto-dismissed with page refresh, is this expected?

remyperona:
Second case is fixed with the latest commit.

For the first case, this is already the case before `.*` is not a value preserved.

Yes it's expected for the notice to be removed on page refresh, it's not persistent. It displays after saving the settings.

Mai-Saad:
@Tabrisrp Thank you, now the second case is fixed, and yes anything in the field without ` / ` at the beginning will be removed. Almost all tests are passing now but having a question, currently, if we add `/(.*)/` in Never Cache URL => this is preserved in the field and will result in disabled caching for all pages other than the home page, is this ok?

remyperona:
The original ticket is only mentioning the `/(.*)` pattern

Mai-Saad (approved):
Working as expected then 
[testrail-report-300.pdf](https://github.com/wp-media/wp-rocket/files/7465858/testrail-report-300.pdf)

PR #4476, opened 2021-11-02
Fixes #3704 Prevent server error if charset value is empty in htaccess rules

### LINKED ISSUE #3704: AddDefaultCharset directive added to .htaccess missing $charset value causing internal server error. Function missing logic to check returned value.
**Describe the bug**
AddDefaultCharset from function "get_rocket_htaccess_charset()" returns without a $charset value via get_bloginfo(). This function contains no logic to check the emptiness of the return from get_bloginfo(). Would recommend checking $charset before creating and appending rules. 

On Cloudways (host) this caused an internal server error and took down 4 sites on a multisite setup. 

**To Reproduce**
Not been able to reproduce, manually fixed issue by adding 'UTF-8' after directive via ssh.

**Expected behavior**
To not crash the site.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

It seems in some cases the charset value can be empty, which will result in invalid htaccess rules and creating a server error.

To avoid the issue, we check if the value is empty, and if it is, we bail-out early.

Fixes #3704

## Type of change

No

# Checklist:

### DIFF
--- a/inc/functions/htaccess.php
+++ b/inc/functions/htaccess.php
@@ -506,6 +506,10 @@ function get_rocket_htaccess_charset() { // phpcs:ignore WordPress.NamingConvent
 	// Get charset of the blog.
 	$charset = preg_replace( '/[^a-zA-Z0-9_\-\.:]+/', '', get_bloginfo( 'charset', 'display' ) );
 
+	if ( empty( $charset ) ) {
+		return '';
+	}
+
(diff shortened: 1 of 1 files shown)


iCaspar (approved):

Mai-Saad (approved):
Working fine now.

[testrail-report-301.pdf](https://github.com/wp-media/wp-rocket/files/7486900/testrail-report-301.pdf)

PR #4484, opened 2021-11-04
Fixes #4382 Prevent unexpected exclusion from delay JS when type is present in the string

### LINKED ISSUE #4382: Impossible to delay script which contain term "type" on his url name. Condition not specific enough
- Version 3.9.4
- Used the search feature to ensure that the bug hasn’t been reported before. Yes

**Describe the bug**
The script url with a term "**type**" in his name cannot be delay by WP Rocket. My url script url is here /wp-content/plugins/my-custom-plugin-types/assets/js/name_script.min.js. So in the process to adjust attributes for scripts, and here add _type="rocketlazyloadscript"_, he detect the term "type" and return the script like he was before. He don't add _type="rocketlazyloadscript"_ for this script and cause trouble, because my file doesn't delay.

**To Reproduce**
1. Add a js with type on his url like "test-type.js" or "/wp-content/plugins/my-custom-plugin-types/assets/js/name_script.min.js"
2. Active delay and defer js in WP_rocket
3. Test anonyme page
4. See in the DOM <script src="/wp-content/plugins/my-custom-plugin-types/assets/js/name_script.min.js" defer></script> instead of <script type="rocketlazyloadscript" src="/wp-content/plugins/my-custom-plugin-types/assets/js/name_script.min.js" defer></script>

**Expected behavior**
I think they are a probleme with the condition line 191 on /wp-rocket/inc/Engine/Optimization/DelayJS/HTML.php : 

> strpos( $matches['attr'], 'type' ) !== false

The condition is not specific, this condition should be better like

> strpos( $matches['attr'], 'type=' ) !== false

or

> strpos( $matches['attr'], 'type="rocketlazyloadscript"' ) !== false

I try this modification and it's work. I think this is the best way. Because all plugin with type in is name can't delay his script, it's too bad.

remyperona:
## Description

Using just `type` can lead to unexpected matches in some cases, notably when this is used inside the URL for example.

What we want to target is the `type` attribute specifically, so searching for `type=` should avoid edge case exclusions.

Fixes #4382

## Type of change

## How Has This Been Tested?

- [x] Added new fixture to cover this case

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DelayJS/HTML.php
+++ b/inc/Engine/Optimization/DelayJS/HTML.php
@@ -188,7 +188,7 @@ class HTML {
 		if ( ! empty( $matches['attr'] ) ) {
 
 			if (
-				strpos( $matches['attr'], 'type' ) !== false
+				strpos( $matches['attr'], 'type=' ) !== false
 				&&
 				! preg_match( '/type\s*=\s*["\'](?:text|application)\/(?:(?:x\-)?javascript|ecmascript|jscript)["\']|type\s*=\s*["\'](?:module)[ "\']/i', $matches['attr'] )
 			) {
--- a/tests/Fixtures/inc/Engine/Optimization/DelayJS/HTML/delayJs.php
+++ b/tests/Fixtures/inc/Engine/Optimization/DelayJS/HTML/delayJs.php
@@ -2,6 +2,7 @@
 
 $html = '<html>
 <head>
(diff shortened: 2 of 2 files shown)


iCaspar (approved):

Mai-Saad (commented):
Working as expected now. 

Note: if the URL contains type= for any reason, it won't be delayed.

PR #4486, opened 2021-11-05
Fixes #4436 Improve accessibility of the lazyloaded Youtube videos

### LINKED ISSUE #4436: LazyLoad for Iframes: Preview image cannot receive focus and fails the a11y audit

**Describe the bug**
It appears the lazy load preview image can not receive focus, and even if hacked with a tabindex="0", pressing enter on the image doesn't load the video.

We don't have indeed the "highlighting" focus option and `tab` is even skipping the video. The original YT video has it, so it works if we bypass the optimization.

This fails a11y audits, it's an extremely high priority for us to get fixed. All American websites using your plugin can not legally use this feature.

**To Reproduce**
1. Add YouTube video with standard WP block
2. Enable the LL for YT with preview images in WPR
3. Visit the page and use `tab` to navigate between the elements.
4. The video will be skipped

**Expected behavior**
Our LL should preserve these accessibility attributes
 
**Additional context**
Ticket: https://secure.helpscout.net/conversation/1661178330/301214?folderId=1213662
Slack: https://wp-media.slack.com/archives/C08N8J6VC/p1634216849074000

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Import the changes from the common library to improve the accessibility of the lazyloaded Youtube videos.

Fixes #4436 

## Type of change

# Checklist:


iCaspar (approved):

Mai-Saad:
@Tabrisrp while exploring the fix, can see that now using tab + enter when we are on video => Video will work. but can't see any visual indicator that we are on video, Can we display a visual indicator?

https://jmp.sh/mYGSopp

remyperona:
There is an outline border displayed when the focus is set on the video, like this (this is on Firefox, but all browsers should be compatible with this CSS):

outline

Mai-Saad:
@Tabrisrp Seems the problem with Astra (independent on browser i.e on both FF and chrome), checked it with 2020, 2021, Avada, Flatsome, and focus is working there so not sure if we can handle it from our side for Astra or if other themes having the same problem. Will go for test execution and it will at least fix the problem for most of themes. what do you think?

remyperona:
We can't do much if the theme CSS is overridding ours, I think it's fine as is

Mai-Saad:
@Tabrisrp after test execution, almost smoke for LL is passing but can see these issues:

1- With 2021, while LL iframe + preview image are enabled (or if + LL image enabled) => preview image is black (working fine on 2020, 2015, 2022, Avada, Flatsome) => can we fix that our side?
Screenshot from 2021-11-18 22-51-02

2- With 2021, when only LL image is enabled, we can't access 1st video here https://new.rocketlabsqa.ovh/image-and-video/ (same with 2020, flatsome, 2016)

3- While LL iframe is only enabled, we have a problem accessing 1st video here https://new.rocketlabsqa.ovh/image-and-video/  and all videos here https://new.rocketlabsqa.ovh/10067-2/ => tested with 2020, 2021, Flatsome, Avada. Can you please have a look into this? (I think it's related to if the video is displayed with/without need for scroll as it's working fine here https://new.rocketlabsqa.ovh/one-video/)

remyperona:
@Mai-Saad Are those issues new with this PR, or do they already exist on trunk?

Mai-Saad:
@Tabrisrp 
1- not happening on the trunk => will we handle this case (knowing that if we hover the video, the preview image will be displayed)? or is it related to the theme overriding our CSS?
2- Same on the trunk => even happening with ?nowprocket (may be related to native LL), I think this is probably not our side
3- Same on the trunk => We can add a separate GH issue for this one.

Q: How does the same page https://new.rocketlabsqa.ovh/image-and-video/?nowprocket behave differently while image LL only enabled than iframe LL only enabled?

what do you think?

@engahmeds3ed Can you please add your findings here as well

remyperona:
Last commit fixes the issue with the 2021 theme overridding our CSS for the play button

Mai-Saad:
@Tabrisrp point 1 is fixed now, will put 2/3 in a separate ticket for further investigations as they are already on the trunk

Note:
- As per discussion, not being able to access LL video with tab is expected and if page loads then used tab, it will be accessed 
so for 3 not able to access 1st video here https://new.rocketlabsqa.ovh/image-and-video/ nor any video here, https://new.rocketlabsqa.ovh/10067-2/ is expected 
- The main issue now to be doc in other ticket is that we remove `loading = 'lazy'` from nowprocket when our LL image or LL iframe is enabled

Mai-Saad (approved):
Working as expected
[testrail-report-309.pdf](https://github.com/wp-media/wp-rocket/files/7583197/testrail-report-309.pdf)

Issue #4340, opened 2021-09-09
Delay JS add compatibility exclusion for PDF Embedder Plugin


markllego:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version
WP Rocket 3.9.3
- Used the search feature to ensure that the bug hasn’t been reported before
I have read the ff: https://docs.wp-rocket.me/article/1560-delay-javascript-execution-compatibility-exclusions

**Describe the bug**
When the WP-Rocket Delay JS Feature is enabled, the PDF Embedder plugin will not work as expected.

**To Reproduce**
1. Install and enable the PDF Embedder Plugin.
2. Make a post and attached a pdf file using the PDF Embedder shortcode.
3. Use Incognito mode and go to the post.
4. The embedded pdf file will fail to load, appear as a link or get stuck on loading.

**Expected behavior**
Add the following scripts to solve the issue:

/wp-content/plugins/pdf-embedder/js/all-pdfemb-basic-4.6.2.min.js,
/wp-content/plugins/pdf-embedder/js/pdfjs/pdf-4.6.2.min.js, and
/wp-content/plugins/pdf-embedder/js/pdfjs/pdf-4.6.2.worker.min.js
/wp-content/plugins/PDFEmbedder-premium/js/all-pdfemb-premium-4.4.3.min.js
/wp-content/plugins/PDFEmbedder-premium/js/pdfjs/pdf-4.4.3.min.js
/wp-content/plugins/PDFEmbedder-premium/js/pdfjs/pdf-4.4.3.worker.min.js
/wp-content/plugins/PDFEmbedder-premium-secure/js/all-pdfemb-premium-4.4.5.min.js,
/wp-content/plugins/PDFEmbedder-premium-secure/js/pdfjs/pdf-4.4.5.min.js, and
/wp-content/plugins/PDFEmbedder-premium-secure/js/pdfjs/pdf-4.4.5.worker.min.js
/wp-content/plugins/PDFEmbedder-premium-secure/js/all-pdfemb-premium-4.4.5.min.js,
/wp-content/plugins/PDFEmbedder-premium-secure/js/pdfjs/pdf-4.4.5.min.js, and
/wp-content/plugins/PDFEmbedder-premium-secure/js/pdfjs/pdf-4.4.5.worker.min.js

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
[...]

PR #4491, opened 2021-11-09
Add check for updated classname in PDFEmbedder compatibility

iCaspar:
This PR patches the recent PDFEmbedder update (v4.6.3) that changes the classname of the main plugin class.

We still check for the old classname to preserve backwards compatibility for Embedder users who have not yet updated.

The error was found as a result of failing integration test on the `PDFEmbedder` group; this PR restores test to passing.

### DIFF
--- a/inc/ThirdParty/Plugins/PDFEmbedder.php
+++ b/inc/ThirdParty/Plugins/PDFEmbedder.php
@@ -34,8 +34,13 @@ class PDFEmbedder implements Subscriber_Interface {
 	 * @return array
 	 */
 	public function exclude_pdfembedder_scripts( $excluded_js ) {
-		if ( class_exists( 'pdfemb_basic_pdf_embedder' ) ) {
+		if (
+			class_exists( 'PDF_Embedder_Basic' )
+			||
+			class_exists( 'pdfemb_basic_pdf_embedder' )
+		) {
 			// Exclude Free version.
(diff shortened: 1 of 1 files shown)


remyperona (approved):

Mai-Saad:
@iCaspar **PDF embedder (4.6.4)** is working fine with Minify JS, Defer JS, and Delay JS (no need for user interaction to display PDF, probably fixing this ticket https://github.com/wp-media/wp-rocket/issues/4340).

 Using Combine JS gives an error 
Screenshot from 2021-11-10 09-43-01

The error is fixed by the following exclusion from Combine JS

```
/pdf-embedder(.*)
jquery.min.js
jquery-migrate.min.js
```

**For  PDF embedder (4.6.2)** working fine with Minify JS and Delay JS

 But, when Combine JS is enabled, links are displayed instead of the PDF.
Screenshot from 2021-11-10 10-21-44

Excluding jquery and jquery-migrate will make PDF displayed

```
jquery.min.js
jquery-migrate.min.js
```

With Defer JS, the file is stuck in loading or displayed as a link.
Screenshot from 2021-11-10 10-20-29

The following exclusion fixes the problem

```
/pdf-embedder/js/(.*).js
jquery.min.js
jquery-migrate.min.js
```

Note: All issues mentioned above already exist on the trunk

Issue #4314, opened 2021-09-01
Items added to settings_errors transient are not displayed on settings update


Mai-Saad:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version => 3.9.2
- Used the search feature to ensure that the bug hasn’t been reported before =>yes

**Describe the bug**
When using an invalid regex, it is removed from UI without displaying a notice

**To Reproduce**
1. Go to Advanced rules
2. Add `gtag(`  in Always Purge URLs
3. Save
4. `gtag(` removed and no notice is displayed

**Expected behavior**
Notice is displayed

**Additional context**
As per change here https://github.com/wp-media/wp-rocket/pull/4195#issuecomment-902888918 , the notice shall be `The following pattern has been escaped to prevent errors:`

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

iCaspar:
This issue is blocking
#4472 
#4327 
#4013

iCaspar:
Grooming Notes:

Reproduced: ✔️ 

Root cause: ✔️ 
Adding settings_errors to the transient does not persist them across the page load because other calls to the settings API  `get_settings_errors()` deletes the transient before we have a chance in our settings to get the contents.

Solution scope: ✔️ 
Instead of setting the transient, we can use WP's `add_settings_error()` to store the errors/notices.

Estimated effort:
[XS]

PR #4501, opened 2021-11-12
Fix 4472 - Display notice when DelayJS is activated with Autoptimize Aggregate JS already in effect

### LINKED ISSUE #3707: Autoptimize - rocket-delay-js-js-after script and delayed scripts are being combined by Autoptimize
- Made sure you’re on the latest version ✔
- Used the search feature to ensure that the bug hasn’t been reported before ✔

**Describe the bug**
When Delay JavaScript Execution and JavaScript combination feature from Autoptimize are active, the Delay JavaScript Execution Script (rocket-delay-js-js-after) and all the delayed scripts (data-rocketlazyloadscript ones) are combined and disappear from the HTML, this prevents the delayed scripts for being executed.

**To Reproduce**
1. Activate Delay JavaScript Execution on WP Rocket
2. Add a script to be delayed to the "Scripts to delay" box if needed (Example GA, GTM)
3. Activate JavaScript combination on Autoptimize (Aggregate JS-files option) and "Also aggregate inline JS" if needed.

**Expected behavior**
These scripts shouldn't be combined by Autoptimize or WP Rocket should disable Delay JavaScript execution when Aggregate JS-files option is active on Autoptimize.

**Screenshots**
image

**Additional context**
Because the delayed scripts are "Empty" (there is nothing between the <script> tags and src is not present), they are just removed from the HTML and in the combined file they do not exist, since they are in the data-rocketlazyloadscript attribute.

Excluding **_data-rocketlazyloadscript_** and **_rocket-browser-checker-js-after_** from Autoptimize solves the problem, but by the moment this should be done manually from Autoptimize UI: 
image

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

### LINKED ISSUE #4472: Autoptimize aggregated JS file can't be delayed

**Describe the bug**
When Autoptimize is using their `JS, CSS & HTML` > `JavaScript Options` > `Aggregate JS-files`, WP Rocket `Delay JS` is not applied to the resulting combined file.

The combined file will be added as:

```
<script defer src="https://domain.ext/wp-content/cache/autoptimize/js/autoptimize_609b9fb609c9a893772168880e9ef32e.js">
```

Checking the buffer received before applying the Delay JS: 

[/inc/Engine/Optimization/DelayJS/Subscriber.php#L84-L86](https://github.com/wp-media/wp-rocket/blob/v3.10.2/inc/Engine/Optimization/DelayJS/Subscriber.php#L84-L86)

The combined files are not there on the HTML to process. 

https://gist.github.com/DahmaniAdame/51aa691bad4070275eb565d8dc1e3445

**To Reproduce**
1. Go to enable `Autoptimize` > `JS, CSS & HTML` > `JavaScript Options` > `Aggregate JS-files`
2. Enable WP Rocket `Delay JS` with no exclusions
3. See error

**Expected behavior**
We should either delay the resulting combined file or hint to the customer that this particular file won't be delayed.

**Screenshots**
N/A

**Additional context**

Temporary fix: disabling `Autoptimize` > `JS, CSS & HTML` > `JavaScript Options` > `Aggregate JS-files` will allow delaying all scripts.

Ticket - https://secure.helpscout.net/conversation/1681325316/304823/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

iCaspar:
## Description

Fixes #4472
Fixes #3707 

## Type of change

Yes
In the grooming I had thought we would be adding a helper notice into the DelayJS selection field. In fact, we are adding a notice to the notices area at the top of the page.
Also, to do this, we had been adding the notice via the '`settings_errors` transient,  but that was failing to display messages properly. Instead, this adds a method in the DelayJS Admin/Subscriber to check if Autoptimize Aggregate JS is active, and if so, uses `add_settings_errors()` for display.

## How Has This Been Tested?

- [X] New Integration tests to confirm that the notice is added under expected conditions.
- [x] Check that notice is displayed properly on local environment.

Screen Shot 2021-11-17 at 9 32 55 AM

# Checklist:

### DIFF
--- a/inc/ThirdParty/Plugins/Optimization/Autoptimize.php
+++ b/inc/ThirdParty/Plugins/Optimization/Autoptimize.php
@@ -36,7 +36,6 @@ class Autoptimize implements Subscriber_Interface {
 	 *
 	 * @return array
 	 * @since  3.10.4
-	 *
 	 */
 	public static function get_subscribed_events() {
 		if ( ! rocket_get_constant( 'AUTOPTIMIZE_PLUGIN_VERSION', false ) ) {
@@ -44,7 +43,7 @@ class Autoptimize implements Subscriber_Interface {
 		}
 
 		return [
-			'admin_notices' => [ 'warn_when_js_aggregation_and_delay_js_active' ]
+			'admin_notices' => [ 'warn_when_js_aggregation_and_delay_js_active' ],
 		];
 	}
 
@@ -70,12 +69,7 @@ class Autoptimize implements Subscriber_Interface {
 			return;
 		}
 
-		$message = '</strong>' .
-		           __(
-			           'We have detected that Autoptimize\'s JavaScript Aggregation feature is enabled. The Delay JavaScript Execution will not be applied to the file it creates. We suggest disabling it to take full advantage of Delay JavaScript Execution.',
-			           'rocket'
-		           ) .
-		           '<strong>';
+		$message = '</strong>' . __( 'We have detected that Autoptimize\'s JavaScript Aggregation feature is enabled. The Delay JavaScript Execution will not be applied to the file it creates. We suggest disabling it to take full advantage of Delay JavaScript Execution.', 'rocket' ) . '<strong>';
 
 		rocket_notice_html(
 			[
--- a/inc/ThirdParty/Plugins/Optimization/Autoptimize.php
+++ b/inc/ThirdParty/Plugins/Optimization/Autoptimize.php
@@ -38,10 +38,6 @@ class Autoptimize implements Subscriber_Interface {
 	 * @since  3.10.4
 	 */
 	public static function get_subscribed_events() {
-		if ( ! rocket_get_constant( 'AUTOPTIMIZE_PLUGIN_VERSION', false ) ) {
-			return [];
-		}
-
 		return [
 			'admin_notices' => [ 'warn_when_js_aggregation_and_delay_js_active' ],
 		];
@@ -53,23 +49,35 @@ class Autoptimize implements Subscriber_Interface {
 	 * @since 3.10.4
 	 */
 	public function warn_when_js_aggregation_and_delay_js_active() {
+		if ( ! rocket_get_constant( 'AUTOPTIMIZE_PLUGIN_VERSION', false ) ) {
+			return;
+		}
+
 		if ( ! current_user_can( 'rocket_manage_options' ) ) {
 			return;
 		}
 
-		$boxes = get_user_meta( get_current_user_id(), 'rocket_boxes', true );
+		$autoptimize_aggregate_js_setting = get_option( 'autoptimize_js_aggregate' );
+		$boxes                            = get_user_meta( get_current_user_id(), 'rocket_boxes', true );
+
+		if ( 'on' !== $autoptimize_aggregate_js_setting || false === (bool) $this->options->get( 'delay_js' ) ) {
+			if ( ! is_array( $boxes ) ) {
+				return;
+			}
+
+			$this->remove_warning_dismissal( $boxes, __FUNCTION__ );
+
+			return;
+		}
 
 		if ( in_array( __FUNCTION__, (array) $boxes, true ) ) {
 			return;
 		}
 
-		$autoptimize_aggregate_js_setting = get_option( 'autoptimize_js_aggregate' );
-
-		if ( 'on' !== $autoptimize_aggregate_js_setting && ! (bool) $this->options->get( 'delay_js' ) ) {
-			return;
-		}
-
-		$message = '</strong>' . __( 'We have detected that Autoptimize\'s JavaScript Aggregation feature is enabled. The Delay JavaScript Execution will not be applied to the file it creates. We suggest disabling it to take full advantage of Delay JavaScript Execution.', 'rocket' ) . '<strong>';
+		$message = '</strong>' . __(
+			'We have detected that Autoptimize\'s JavaScript Aggregation feature is enabled. The Delay JavaScript Execution will not be applied to the file it creates. We suggest disabling it to take full advantage of Delay JavaScript Execution.',
+				'rocket'
+			) . '<strong>';
 
 		rocket_notice_html(
 			[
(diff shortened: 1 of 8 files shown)


iCaspar:
This approach should work, as the integration tests prove that notice data is added to the `settings_errors` transient array. But the notice is not appearing, due to the same bug that is causing https://github.com/wp-media/wp-rocket/issues/4314, namely there is something preventing any notice being displayed from the `settings_errors` transient when settings are updated.

Marking this as blocked by #4314.

mostafa-hisham (approved):

Mai-Saad:
@iCaspar Please find notes below based on test analysis and exploratory test:

1- If Delay JS was enabled, then we enable auto-optimize with the aggregate JS features => no warning will be displayed (we will need to re-enable delay js to see the warning), can we display the warning without needing to re-enable Delay JS (maybe with cache clear, refresh settings)?

2- When Autooptimize aggregate js files and inline js is enabled along with WPR LL video and preview image, result in removing LL videos script from HTML => Do we need a separate GH issue for that? or is it something we can handle on this PR? (same already on trunk)

3- @piotrbak  Q: Why not handle the current conflict, the same way as we do with minify/combine CSS/JS (disable the option and display red msg)?
Screenshot from 2021-11-18 12-27-13

4- Just for UI confirmation => Current msg which is displayed only when enable delay js while aggregate options are enabled in auto-optimize is as attached, is this what we need?
Screenshot from 2021-11-18 12-28-03

piotrbak:
@Mai-Saad **When it comes to the 3:**
Minify and Combine CSS/JS are exactly the same features. It's not the case with DelayJS, as it's not causing errors, rather not working as good as it should, we decided to go with a educative warning approach.

@iCaspar Can we make the warning border yellow instead of blue?

piotrbak:
@Mai-Saad 

**For 2:**
Does it cause any issues with the video or lazyload?

**For 4:**
We're going to display the message when:
1. Delay JS is enabled
2. Autoptimize JS Aggregation is enabled

CSS files are disabled because the CSS optimisations are enabled in the Autoptimize, below we have an explanation

iCaspar:
@piotrbak @Mai-Saad 
On 1, please confirm that the User Story above is accurate. Per the current version of the story:
We will show the message when Autoptimize>JS Aggregation is already active and then DelayJS is activated (the behavior Mai observed). If we also want it to appear when the events happen in the reverse order (DelayJS is already active and then JSAggregation is activated) we should specify that behavior in the story. Otherwise, I think it is working as intended here.

On 2, interaction with Lazyload -- this PR is ONLY about messaging -- see the User Story. It will not include any fixes for conflicts with LazyLoad (or anything else).

On 3, see 2.

On 4, the message is (again) specified in the User Story.

piotrbak:
@iCaspar Will it be possible to cover both scenarios? We want to display the warning in all possible use cases. The same should be applied to other Autoptimize compatibilities. Sorry for the misunderstanding.

iCaspar:
@piotrbak No problem. Let's update the User Story and I can make that change this morning.

inc/ThirdParty/Plugins/Optimization/Autoptimize.php
```diff
@@ -0,0 +1,106 @@
...
+
+	/**
+	 * Add an admin warning notice when Delay JS and JS Aggregation are both activated.
+	 *
+	 * @since 3.10.4
+	 */
+	public function warn_when_js_aggregation_and_delay_js_active() {
+		if ( ! rocket_get_constant( 'AUTOPTIMIZE_PLUGIN_VERSION', false ) ) {
+			return;
+		}
```
engahmeds3ed, line 54: I think it would be better adding this check into `get_subscribed_events` method not to register the file events if the plugin is not installed, this will help if we added other events in the future.

remyperona (approved):

inc/ThirdParty/Plugins/Optimization/Autoptimize.php
```diff
@@ -0,0 +1,106 @@
...
+		];
+	}
+
+	/**
+	 * Add an admin warning notice when Delay JS and JS Aggregation are both activated.
+	 *
+	 * @since 3.10.4
+	 */
+	public function warn_when_js_aggregation_and_delay_js_active() {
+		if ( ! rocket_get_constant( 'AUTOPTIMIZE_PLUGIN_VERSION', false ) ) {
```
engahmeds3ed, line 52: `AUTOPTIMIZE_PLUGIN_VERSION` is introduced at autoptimize version 2.4.0 (3 years ago) and I think it wasn't there before, is this a problem or we are fine? @wp-media/productrocket

engahmeds3ed (changes requested):
Small minor comments

inc/ThirdParty/Plugins/Optimization/Autoptimize.php
iCaspar: Initially, I did have it where you are suggesting. However, when `get_subscribed_events()` returns and empty array, that was what seemed to cause the failure in the `removeAllCallbacksExcept()` from detatching the other things from `admin_notices`. I think we will have to leave it here for now, until we can fix the callbacks test method.

inc/ThirdParty/Plugins/Optimization/Autoptimize.php
piotrbak: I think we're fine, I don't believe anyone will be using 3 years old Autoptimize + newest WP Rocket

iCaspar: This is OK, per @piotrbak in slack.

engahmeds3ed (approved):

Mai-Saad (commented):
Working as expected.

As per discussion, we agreed on the following notes:
- Notice won't be displayed after dismissing with plugin(s) reactivation (with same settings), import same settings
- Notice will be normally displayed when delay js is enabled independent of exclusions (i.e with default exclusions, notice is displayed)
- When enabling autoptimize while LL video is enabled, LL video will be automatically unchecked (this is already on the trunk and added ticket for this)
[testrail-report-311.pdf](https://github.com/wp-media/wp-rocket/files/7589642/testrail-report-311.pdf)

PR #4502, opened 2021-11-12
Fix 4314 - Items in settings errors transient not displayed on settings update

### LINKED ISSUE #4314: Items added to settings_errors transient are not displayed on settings update

iCaspar:
## Description

Changes our settings error process to use `add_settings_error` instead of the settings_errors transient.

Fixes #4314

## Type of change

Same as grooming.

## How Has This Been Tested?

Checked that errors are being displayed on local development environment.

# Checklist:

### DIFF
--- a/inc/admin/admin.php
+++ b/inc/admin/admin.php
@@ -246,7 +246,7 @@ function rocket_rollback() {
 add_action( 'admin_post_rocket_rollback', 'rocket_rollback' );
 
 /**
- * After a rollback has been done, replace the "return to" link by a link pointing to WP Rocket's tools page.
+ * After a rollback has been done, replace the "return to" link by a link pointing to WP Rocket's tools page.
  * A link to the plugins page is kept in case the plugin is not reactivated correctly.
  *
  * @since  3.2.4
--- a/inc/admin/options.php
+++ b/inc/admin/options.php
@@ -155,13 +155,13 @@ function rocket_pre_main_option( $newvalue, $oldvalue ) {
 	}
 
 	if ( $errors ) {
-		$error_message  = _n( 'The following pattern is invalid and has been removed:', 'The following patterns are invalid and have been removed:', array_sum( array_map( 'count', $errors ) ), 'rocket' );
+		$error_message = _n( 'The following pattern is invalid and has been removed:', 'The following patterns are invalid and have been removed:', array_sum( array_map( 'count', $errors ) ), 'rocket' );
 
 		foreach ( $errors as $error ) {
 			$error_message .= '<ul><li>' . implode( '</li><li>', $error ) . '</li></ul>';
 		}
 
-		$errors         = [];
+		$errors = [];
 
 		$rocket_settings_errors[] = [
 			'setting' => 'general',
--- a/inc/admin/options.php
+++ b/inc/admin/options.php
@@ -120,13 +120,14 @@ function rocket_pre_main_option( $newvalue, $oldvalue ) {
 	$is_form_submit = WP_ROCKET_PLUGIN_SLUG === $is_form_submit;
 	$errors         = [];
 	$pattern_labels = [
-		'exclude_css'       => __( 'Excluded CSS Files', 'rocket' ),
-		'exclude_inline_js' => __( 'Excluded Inline JavaScript', 'rocket' ),
-		'exclude_js'        => __( 'Excluded JavaScript Files', 'rocket' ),
-		'cache_reject_uri'  => __( 'Never Cache URL(s)', 'rocket' ),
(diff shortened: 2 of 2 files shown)


mostafa-hisham (approved):

Mai-Saad:
@iCaspar please find my notes/Qs below while exploring the PR:

1- `PHP Warning:  preg_match(): Compilation failed: missing closing parenthesis` will exist in logs for  Defer, and Delay JS => just confirming that the current PR won't fix this problem, right?

2- Notice is not displayed for the following fields although `gtag(` was removed: Never Cache Url(s), Always Purge URL(s), Preload Cache, and Preload Fonts => will we handle their notice in the current PR?

3- Although  `gtag(` is removed and notice is displayed for combine js, we have this warning in debug.log `PHP Warning:  preg_match(): Compilation failed: missing closing parenthesis at offset 928 in /var/www/new.rocketlabsqa.ovh/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Optimization/Minify/JS/Combine.php on line 914` (same warning on the trunk while notice not displayed) => Will this be fixed with 1 related issue?

4- Q: which fields shall escape invalid regex by adding `\` and which fields shall remove invalid regex totally?

iCaspar:
On 1 and 3, indeed, this will not fix any warnings. This PR will ONLY address that the message is not being displayed when an invalid entry was removed.

On 2 and 4, the notice should be displayed for an invalid regex that is removed in the following fields:
```
		'exclude_css'       => __( 'Excluded CSS Files', 'rocket' ),
		'exclude_inline_js' => __( 'Excluded Inline JavaScript', 'rocket' ),
		'exclude_js'        => __( 'Excluded JavaScript Files', 'rocket' ),
		'cache_reject_uri'  => __( 'Never Cache URL(s)', 'rocket' ),
		'cache_reject_ua'   => __( 'Never Cache User Agent(s)', 'rocket' ),
		'cache_purge_pages' => __( 'Always Purge URL(s)', 'rocket' ),
		'cdn_reject_files'  => __( 'Exclude files from CDN', 'rocket' ),
```

Mai-Saad:
@iCaspar Notice is displayed for all mentioned fields except `'cache_reject_uri'  => __( 'Never Cache URL(s)', 'rocket' ),` and `'cache_purge_pages' => __( 'Always Purge URL(s)', 'rocket' ),` Can we handle those?

- Other Note: If we have multiple invalid regex, the notice is displayed only on the last item.

Example: When excluding the following from CSS :
```
h(.*)apply(.*)(
h(.*)apply(.*))
gtag(
(test
```
although all items were removed from the field, the notice is :
lastErrorOnly 

Shouldn't we display all removed items in the notice?

iCaspar:
@Mai-Saad @piotrbak 

> Shouldn't we display all removed items in the notice?

Yes!
I've added a bit to make that happen now. Should be like this:
Screen Shot 2021-11-22 at 10 59 52 AM

Also, looks like in the case of 
* Never Cache URLs
* Always Purge URLs, and
* Exclude files from CDN
Turns out we are also, _**in a different sanitization**_, removing anything from those fields that is not an actual file -- regardless of any regex.
It will probably be better to handle these 3 in a separate issue/pr.

Also, I've added the DelayJS to the pattern fields that will display here based on regex removal. So at least the "Warning" part of #4136 should be applied here (not the other things about notices in logs, etc.)

Mai-Saad (commented):
As per discussions and latest analysis, this PR works as expected to display notice for all mentioned fields (except Never Cache URL and Always purge URL => will be in another ticket)

Notes:
- Some notices have a typo (will be in another ticket)
- Some fields cause warning (will be in another ticket)
-  Exclusion for `test\(` is removed from js/CSS files (will be in another ticket)
- Analysis Sheet here https://docs.google.com/spreadsheets/d/1NQz8vhTpuTa9Jqaf8TcyfLmhinc0Lx59ulo_ZUR9r0c/edit?usp=sharing

PR #4525, opened 2021-11-23
Enhancement/4013 add notice for compat autoptimize aggregatecss rocket cpcss

### LINKED ISSUE #4013: Autoptimize compatibility: make CPCSS compatible when using their "aggregate inline CSS" option

**Describe the bug**
When Autoptimize `Aggregate CSS-files` > `Also aggregate inline CSS` is enabled, it will combine the CPCSS inline code.

WP Rocket won't make Autoptimize combined CSS file async. 

```
<link media="all" href="https://example.com/wp-content/cache/autoptimize/css/autoptimize_22e19a8b161c43025e4b28146a3a7e67.css" rel="stylesheet" />
<link media="print" href="https://oxygen.adame.io/wp-content/cache/autoptimize/css/autoptimize_f637c8737ec7525695a5301f7a69ccd9.css" rel="stylesheet" />
<link media="screen" href="https://example.com/wp-content/cache/autoptimize/css/autoptimize_1a1a4a4a3c1facc9c784acf7ad599d82.css" rel="stylesheet" /><title>Title</title>
<style id="rocket-critical-css"></style>
```

Probably because their buffer comes after WP Rocket's. 

**To Reproduce**
1. Enable CPCSS on WP Rocket
2. Enable `Aggregate CSS-files` > `Also aggregate inline CSS` on Autoptimize
3. Clear the cache
4. See error (missing CPCSS inline style and non-asynced CSS files)

**Expected behavior**
WP Rocket inline CPCSS should still be there even if the option is enabled. And all other CSS files should be loaded async.

**Screenshots**
image
image

**Additional context**
Excluding `rocket-critical-css` from `Autoptimize` > `CSS Options` > `Exclude CSS from Autoptimize` fixes the issue.

Autoptimize has the following filter to add CSS exclusions programmatically `autoptimize_css_exclude`.

It has also the following filter `autoptimize_html_after_minify` just before returning their processed buffer. It can be used to make their combined CSS file async.

Reference: 

https://github.com/futtta/autoptimize/blob/b9e3ff920ad9169559af1b2fb72fc1ab60033d58/classes/autoptimizeMain.php#L536

Ticket - https://secure.helpscout.net/conversation/1532468157/268917

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

iCaspar:
## Description

Adds a warning notice to advise users of conflict between Autoptimize Aggregate Inline CSS and CPCSS ("Optimize CSS Delivery")

Fixes #4013

## Type of change

- [X] This change requires a documentation update

Yes. This solution builds on the solution from https://github.com/wp-media/wp-rocket/pull/4501, using the Autoptimize compatibility class implemented there. It is no longer dependent on settings_transients.

## How Has This Been Tested?

Integration tests have been added for the new notice method.
Verified functionality locally.

# Checklist:

### DIFF
--- a/inc/Dependencies/RocketLazyload/Assets.php
+++ b/inc/Dependencies/RocketLazyload/Assets.php
@@ -246,7 +246,7 @@ class Assets {
 			}
 		}
 
-		return "<script>function lazyLoadThumb(e){var t='{$image}',a='<div class=\"play\"></div>';return t.replace(\"ID\",e)+a}function lazyLoadYoutubeIframe(){var e=document.createElement(\"iframe\"),t=\"ID?autoplay=1\";t+=0===this.dataset.query.length?'':'&'+this.dataset.query;e.setAttribute(\"src\",t.replace(\"ID\",this.dataset.src)),e.setAttribute(\"frameborder\",\"0\"),e.setAttribute(\"allowfullscreen\",\"1\"),e.setAttribute(\"allow\", \"accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture\"),this.parentNode.replaceChild(e,this)}document.addEventListener(\"DOMContentLoaded\",function(){var e,t,a=document.getElementsByClassName(\"rll-youtube-player\");for(t=0;t<a.length;t++)e=document.createElement(\"div\"),e.setAttribute(\"data-id\",a[t].dataset.id),e.setAttribute(\"data-query\", a[t].dataset.query),e.setAttribute(\"data-src\", a[t].dataset.src),e.innerHTML=lazyLoadThumb(a[t].dataset.id),e.onclick=lazyLoadYoutubeIframe,a[t].appendChild(e)});</script>";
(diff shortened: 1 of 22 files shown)


remyperona (approved):

Mai-Saad:
@iCaspar @piotrbak 
Now Notice is displayed even if `Aggregate CSS-files` is unchecked (at this point `aggregate inline CSS` is disabled with its previous state i.e if it was checked then it will be checked and dim) => in this case we normally apply CPCSS on existing files and I think no need for the notice in this case (probably same will be in similar issues). So what do you think?
Screenshot from 2021-11-23 19-59-11
Screenshot from 2021-11-23 19-58-52

iCaspar:
On the one hand, the notice isn't really necessary in this case of dimmed sub-options. On the other, since we're not actually changing the behavior of the features, it won't actually break anything.
@piotrbak Shall we try to fix this now for 3.10.4, or later for 3.10.5?

iCaspar:
@piotrbak @Mai-Saad 
The last change should take into account the "up the chain" settings for both this and the previous (DelayJS/AggregateJS) notices. The last one (#4528) is the main CSS setting, so shouldn't be affected by this.

Mai-Saad (approved):
Working as expected

Issue #3718, opened 2021-03-25
CDN srcset rewrite - Additional Variation Images Gallery for WooCommerce breaks


vmanthos:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version ✅
- Used the search feature to ensure that the bug hasn’t been reported before ✅

**Describe the bug**

The variable products gallery that's added by the [Additional Variation Images Gallery for WooCommerce](https://.wordpress.org/plugins/woo-variation-gallery/) plugin breaks when CDN rewriting is active.

They are using the following inline script: [https://snippi.com/s/0vxl9t9](https://snippi.com/s/0vxl9t9)

This includes `srcset="{{data.srcset}}`.

Our [RegEx](https://regex101.com/r/0jVKfX/1) picks that up, and rewrites it to this `srcset="https://adfa3ad.rocketcdn.me/{{data.srcset}}` breaking the functionality.

**To Reproduce**

1. Add the following script on a page: [https://snippi.com/s/0vxl9t9](https://snippi.com/s/0vxl9t9)
2. Enable CDN rewriting.
3. The `srcset="{{data.srcset}}` will be rewritten.

**Expected behavior**

We should have a way to exclude such scripts from CDN rewriting.

**Additional context**

**Related ticket:** https://secure.helpscout.net/conversation/1450824927/246580?folderId=2135277

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

DahmaniAdame:
Related - https://secure.helpscout.net/conversation/1536021331/269823/
Excluding `{{data.srcset}}` from the CDN worked.

juricazuanovic:
Related ticket: https://secure.helpscout.net/conversation/1630463365/294577/

camilamadronero-zz:
New ticket https://secure.helpscout.net/conversation/1835826416/335419/

camilamadronero-zz:
related https://secure.helpscout.net/conversation/2000992076/366738/

piotrbak:
It should be fixed here:
https://github.com/wp-media/wp-rocket/pull/4535

Issue #3138, opened 2020-09-23
CDN - Prevent Relative URLs to be replaced by CNAME inside <p> blocks


WordPresseur:
**Before submitting an issue please check that you’ve completed the following steps:**
- WP Rocket 3.7.1.1
- Similar issue for < code > blocks: https://github.com/wp-media/wp-rocket/issues/2322

**Describe the bug**
When there’s a mention to a file with a relative path inside a paragraph block, if CDN feature is enabled, the CNAME will take place.

![Capture d’écran 2020-09-23 à 15 40 40](https://user-images.githubusercontent.com/11939512/94022403-5a0e0e80-fdb5-11ea-916b-8e87c9abb6aa.png)

![Capture d’écran 2020-09-23 à 15 38 55](https://user-images.githubusercontent.com/11939512/94029583-3ea70180-fdbd-11ea-955a-e96e05010aff.png)

**To Reproduce**
1. create a paragraph inside a post and insert a relative path to a static file
2. Activate the CDN feature and enter a CNAME
3. Check your post

**Expected behavior**
The CNAME shouldn’t appear there. The CNAME doesn’t replace the domain name if it’s there.

Ticket: https://secure.helpscout.net/conversation/1284566998/195463?folderId=273766

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

piotrbak:
The issue will be resolved while using the `rocket_cdn_relative_paths` filter:
https://docs.wp-rocket.me/article/42-using-wp-rocket-with-a-cdn#troubleshoot

PR #4535, opened 2021-11-24
Fixes #2849 Ignore inline scripts content when applying the CDN rewrite

### LINKED ISSUE #2849: CDN RegEx picks up relative URLs when query strings have a specially-formatted value
When the source code includes:
1. relative URLs, e.g. `action='/fr/?_ga=2.213437008.512656707.1593765841-1928543133.1539673047'>`
2. containing a query string with a value formatted like this: `string_1.string2`

the [RegEx](https://github.com/wp-media/wp-rocket/blob/0455940481777146a4677dbf59ed928223351639/inc/Engine/CDN/CDN.php#L63) matches that.

Here is an example from a customer's website: https://regex101.com/r/VgMfih/1

When that's being used alongside WPML or Polylang, it can result in rewriting of the URL and, if it's used in, e.g. a form, its functionality can break.

**To Reproduce**

1. Use WPML or Polylang and set the language as a subfolder, e.g. `https://example.com/fr/`
2. Add the following code on the website: `action='/fr/?_ga=2.213437008.512656707.1593765841-1928543133.1539673047'>`
3. Enable the CDN.
4. Check the source code. The previous URL will be rewritten to:
`https://cdn.example.com/fr/?_ga=2.213437008.512656707.1593765841-1928543133.1539673047`

**Expected behavior**

That kind or URL shouldn't be rewritten.

**Additional context**

This is certainly an edge case. Excluding URLs like `/fr/(.*)` from CDN delivery, resolved the issue.

**Related ticket:** https://secure.helpscout.net/conversation/1212168255/177075/

@crystinutzaa said:
* The RegEx picks this up because it starts with a `/` and is inside quotes.
* The RegEx cannot be modified to exclude this scenario.
* A possible solution would be to exclude all forms.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

With this PR, by default we now remove inline scripts from the content to parse, to avoid unexpected rewritting of cases reported by the support.

This behaviour can be reverted by using the new filter `rocket_cdn_exclude_inline_scripts`.

We also introduce another filter `rocket_cdn_relative_paths`, which can prevent complete rewrite of relative paths.

Fixes #2849 

## Type of change

No

## How Has This Been Tested?

- [x] Added snippets to the automated tests to check for reported use case and verify they are correctly passing now

# Checklist:

### DIFF
--- a/inc/Engine/CDN/CDN.php
+++ b/inc/Engine/CDN/CDN.php
@@ -1,4 +1,6 @@
 <?php
+declare(strict_types=1);
+
 namespace WP_Rocket\Engine\CDN;
 
 use WP_Rocket\Admin\Options_Data;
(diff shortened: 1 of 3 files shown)


inc/Engine/CDN/CDN.php
```diff
@@ -41,14 +43,44 @@ public function __construct( Options_Data $options ) {
...
+		 * @since 3.10.5
+		 *
+		 * @param bool $enable True to exclude, false otherwise.
+		 */
+		if ( apply_filters( 'rocket_cdn_exclude_inline_scripts', true ) ) {
+			$buffer = $this->remove_inline_scripts( $html );
+		}
+
+		/**
+		 * Filters the CDN rewritting of relative paths
```
iCaspar:
```suggestion
		 * Filters the CDN rewriting of relative paths
```

iCaspar (approved):
Trivial typo in filter comment.

Mai-Saad (commented):
@Tabrisrp Almost all tests passed but have the following notes:
1- While the filter rocket_cdn_relative_pathsset to false, we still rewrite relative URL inside srcset => can we fix this?
2- We still didn't fix https://github.com/wp-media/wp-rocket/issues/3718 nor https://github.com/wp-media/wp-rocket/issues/2849 by default, but we need to set rocket_cdn_relative_paths to false to avoid rewriting relative URLs inside HTML or inside script containing HTML
3- We still didn't fix https://github.com/wp-media/wp-rocket/issues/3138 by default, set rocket_cdn_relative_paths to false will fix the problem for relative path but not for absolute path in code content

remyperona:
The filter can't be applied for srcset because the RegEx is in a different format, but relative paths are ok in srcset, it's not going to break things there

Mai-Saad (approved):
Working as expected then.
[testrail-report-323.pdf](https://github.com/wp-media/wp-rocket/files/7658690/testrail-report-323.pdf)

PR #4546, opened 2021-11-29
Fix 3967 PHP warning generated when unescaped regex character in deferJS field

### LINKED ISSUE #3967: PHP Warning with some exclusions in Load JavaScript Deferred

iCaspar:
## Description

This preg_replaces all regex special characters when retrieving defer JS options to allow expressions such as `gtag(` to be used without generating PHP warnings.

Fixes #3967 

## Type of change

No.
Escaping all the regex special characters has the side-effect of not allowing any legitimate regex to be used in the field, so entries like `/wp-content/plugins/(.*)/script.js` will not be included in exclusions where expected.

Instead, this does 3 things:
- Adds a testcase for properly evaluating regex entries to the integration tests.
- Adds a testcase for the expectation that if an invalid regex does sneak through, we guard it from triggering warning, and return early without trying to defer js patterns from the UI.
- Adds the `exclude_defer_js` to the set of pattern fields, which will remove invalid patterns from the field on save and trigger the notice that the pattern was invalid.

## How Has This Been Tested?

Confirmed that no warnings are generated in development environment.

# Checklist:

- [X] New integration testcases are added to cover when regex entries are included in the UI field.

Further Context: 
Per note on 3967 regarding warnings also being generated in combineJS fields, the warnings on combine were being generated by DeferJS field entries being passed through to combineJS when Combine was active. Escaping once in the Defer JS process resolves also the combine warnings.

### DIFF
--- a/inc/Engine/Optimization/DeferJS/DeferJS.php
+++ b/inc/Engine/Optimization/DeferJS/DeferJS.php
@@ -75,6 +75,10 @@ class DeferJS {
 
 		$exclude_defer_js = implode( '|', $this->get_excluded() );
 
+		if ( ! @preg_replace( '#(' . $exclude_defer_js . ')#i', '', 'dummy-string' ) ) {
+			return $html;
+		}
+
 		foreach ( $matches as $tag ) {
 			if ( preg_match( '#(' . $exclude_defer_js . ')#i', $tag['url'] ) ) {
 				continue;
--- a/inc/Engine/Optimization/DeferJS/DeferJS.php
+++ b/inc/Engine/Optimization/DeferJS/DeferJS.php
@@ -75,7 +75,7 @@ class DeferJS {
 
 		$exclude_defer_js = implode( '|', $this->get_excluded() );
 
-		if ( ! @preg_replace( '#(' . $exclude_defer_js . ')#i', '', 'dummy-string' ) ) {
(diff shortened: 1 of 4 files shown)


iCaspar:
@wp-media/qa 
This should take care of the PHP warnings. But please note that the intended behavior of this field is to exclude URLs and inline patterns that relate to jQuery specifically. So `gtag\(`, and other patterns like it which are not related to jQuery and usually inline in page html, will not be excluded from defer in any case.

remyperona (approved):

Mai-Saad (approved):
working as expected
[testrail-report-321.pdf](https://github.com/wp-media/wp-rocket/files/7640858/testrail-report-321.pdf)

PR #4551, opened 2021-12-01
Add WP Rocket like in line 126

Presskopp:
## Description
Let's say WP Rockets's function when it is a WP Rocket function.
Let's be consistent and let's be more clear.

## Type of change
Just a small text change. 

## Tests
No tests, because only text was added


piotrbak:
@Presskopp Thank you for the PR

@webtrainingwheels what do you think about this change?

webtrainingwheels:
@piotrbak Seems fine, it's clearer now.

remyperona:
Closing this PR as this has been fixed in a different one #4562

PR #4581, opened 2021-12-13
Fixes #3464 Remove empty values in terms urls array

### LINKED ISSUE #3464: Notice: Trying to access array offset on value of type null in WP_Rocket\Engine\Cache\Purge::purge_url
WP: 5.6
WP Rocket: 3.8.1
PHP: 7.4.3

**UPDATE:**
It was indeed a plugin causing the error by hooking into `term_link` and returning incorrectly. However, I still think the issue should be handled better by WP Rocket. 

--- 

Notice: Trying to access array offset on value of type null in WP_Rocket\Engine\Cache\Purge::purge_url at line 55

I've spent some time debugging this and I'm pretty sure it's an error caused by either some corrupted rewrite rules or a plugin hooking into one of the filters in `get_term_link()`. However it causes a massive amounts of errors/notices from WP Rocket when purging urls. 

You can see from the stack trace that the error happens here https://github.com/wp-media/wp-rocket/blob/master/inc/Engine/Cache/Purge.php#L55 
but the real error originates from https://github.com/wp-media/wp-rocket/blob/master/inc/Engine/Cache/Purge.php#L195 or https://github.com/wp-media/wp-rocket/blob/master/inc/functions/formatting.php#L364-L366 
depending on how you look at it. 

From my test I can see that the method [get_post_terms_urls](https://github.com/wp-media/wp-rocket/blob/master/inc/Engine/Cache/Purge.php#L178) returns the following term urls `["https:\\/\\/example.test\\/shopkatalog\\/?product_cat=dame","https:\\/\\/example.test\\/shopkatalog\\/?product_cat=skonhedsprodukter","https:\\/\\/example.test\\/shopkatalog\\/?product_cat=dame",null,null]`

Notice the two `null` values in the end - that's what's causing the notices. 

Solution 1
It could be fixed by expanding `get_post_terms_urls` to check for `null` and not only `is_wp_error()` [here](https://github.com/wp-media/wp-rocket/blob/master/inc/Engine/Cache/Purge.php#L195) and [here](https://github.com/wp-media/wp-rocket/blob/master/inc/Engine/Cache/Purge.php#L208)

Solution 2
It could also be fixed in the [get_rocket_parse_url](https://github.com/wp-media/wp-rocket/blob/master/inc/functions/formatting.php#L364-L366) where it checks for `is_string` and returns null if it's not. This maybe an error in itself as the method probably should always return the expected array because that's what is expected in rest of the methods where this is used (see Solution 3). 

Solution 3
Validate that [parse_url](https://github.com/wp-media/wp-rocket/blob/3f4de3116cbd9c97f1ac5bff0002f31587353d1d/inc/Engine/Cache/Purge.php#L52) actually returns what it should. 

I'll be happy to send a pull request for any or all of the above solutions but perhaps you could point me in the direction you prefer. 

Stack trace (https://sentry.io/share/issue/14cc63377db043fb8020c8b3b70e86c6/)

```
ErrorException: Notice: Trying to access array offset on value of type null
#27 -content/plugins/wp-rocket/inc/Engine/Cache/Purge.php(55): WP_Rocket\Engine\Cache\Purge::purge_url
#26 -content/plugins/wp-rocket/inc/Engine/Cache/Purge.php(164): WP_Rocket\Engine\Cache\Purge::purge_post_terms_urls
#25 -content/plugins/wp-rocket/inc/Engine/Cache/PurgeActionsSubscriber.php(106): WP_Rocket\Engine\Cache\PurgeActionsSubscriber::purge_post_terms_urls
#24 /wp-includes/class-wp-hook.php(289): WP_Hook::apply_filters
#23 /wp-includes/class-wp-hook.php(311): WP_Hook::do_action
#22 /wp-includes/plugin.php(484): do_action
#21 -content/plugins/wp-rocket/inc/common/purge.php(253): rocket_clean_post
#20 /wp-includes/class-wp-hook.php(289): WP_Hook::apply_filters
#19 /wp-includes/class-wp-hook.php(311): WP_Hook::do_action
#18 /wp-includes/plugin.php(484): do_action
#17 /wp-includes/post.php(6955): clean_post_cache
#16 -content/plugins/woocommerce/includes/data-stores/class-wc-product-data-store-cpt.php(257): WC_Product_Data_Store_CPT::update
#15 -content/plugins/woocommerce/includes/class-wc-data-store.php(179): WC_Data_Store::update
#14 -content/plugins/woocommerce/includes/abstracts/abstract-wc-product.php(1378): WC_Product::save
#13 -content/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-crud-controller.php(169): WC_REST_CRUD_Controller::save_object
#12 -content/plugins/woocommerce/includes/rest-api/Controllers/Version3/class-wc-rest-crud-controller.php(238): WC_REST_CRUD_Controller::update_item
#11 /wp-includes/rest-api/class-wp-rest-server.php(1160): WP_REST_Server::respond_to_request
#10 /wp-includes/rest-api/class-wp-rest-server.php(1007): WP_REST_Server::dispatch
#9 /wp-includes/rest-api/class-wp-rest-server.php(440): WP_REST_Server::serve_request
#8 /wp-includes/rest-api.php(354): rest_api_loaded
#7 /wp-includes/class-wp-hook.php(287): WP_Hook::apply_filters
#6 /wp-includes/class-wp-hook.php(311): WP_Hook::do_action
#5 /wp-includes/plugin.php(551): do_action_ref_array
#4 /wp-includes/class-wp.php(388): WP::parse_request
#3 /wp-includes/class-wp.php(745): WP::main
#2 /wp-includes/functions.php(1291): wp
#1 /wp-blog-header.php(16): require
#0 /var/www/borsen/releases/20210105-143001/public/index.php(3): null
```

NataliaDrause:
## Description

This PR will remove entries with empty or other non-valid values in terms archives `$urls` array in `Purge::get_post_terms_urls()` to prevent the following PHP Notice:

`PHP Notice:  Trying to access array offset on value of type null in /home4/nataliad/testsite.nataliadrause.com/wp-content/plugins/wp-rocket/inc/Engine/Cache/Purge.php on line 55`

Fixes #3464 

## Type of change

No.

## How Has This Been Tested?

On my test site with the following steps:

1. Hard-coded the $urls array with different values including empty and unexpected values:

Test 1:
`$urls = ["https:\\/\\/example.test\\/shop","https:\\/\\/example.test\\?product_cat=skonhedsprodukter", "https:\\/\\/example.test\\/shop","https:\\/\\/example.test\\/shopkatalog\\/?product_cat=dame",null,null, false, 1, 0, true];`

Test 2:
`$urls = [null];`

Test 3:
`$urls = ["https:\\/\\/example.com\\/shop"];`

2. Updated a post and product page for the code to run.
3. `debug.log` had no errors/notices logged.

# Checklist:

### DIFF
--- a/inc/Engine/Cache/Purge.php
+++ b/inc/Engine/Cache/Purge.php
@@ -47,6 +47,10 @@ class Purge {
 	 * @return void
 	 */
 	public function purge_url( $url, $pagination = false ) {
+		if ( ! is_string( $url ) ) {
+			return;
+		}
+
 		global $wp_rewrite;
 
 		$parsed_url = $this->parse_url( $url );
--- a/inc/Engine/Cache/Purge.php
+++ b/inc/Engine/Cache/Purge.php
@@ -211,6 +211,10 @@ class Purge {
 				}
 			}
 		}
+		
+		// Remove entries with empty values in array.
+		$urls = array_filter( $urls, 'is_string' );
+		
 		/**
 		 * Filter the list of taxonomies URLs
 		 *
--- a/inc/Engine/Cache/Purge.php
+++ b/inc/Engine/Cache/Purge.php
@@ -211,10 +211,10 @@ class Purge {
 				}
 			}
 		}
-		
+
 		// Remove entries with empty values in array.
 		$urls = array_filter( $urls, 'is_string' );
-		
+
 		/**
 		 * Filter the list of taxonomies URLs
 		 *


remyperona (approved):

remyperona:
There is a just a little code styling fix to do, whitespaces in empty lines around the new code 👍🏼

Mai-Saad (approved):
Reproduced only on local env. working as expected now.
[testrail-report-329.pdf](https://github.com/wp-media/wp-rocket/files/7733379/testrail-report-329.pdf)
@wp-media/php any idea why tests are failing here?

engahmeds3ed:
> Reproduced only on local env. working as expected now. [testrail-report-329.pdf](https://github.com/wp-media/wp-rocket/files/7733379/testrail-report-329.pdf) @wp-media/php any idea why tests are failing here?

Now tests are passing.

@Tabrisrp @NataliaDrause 
I have a performance suggestion here that we need to discuss about:
The solution proposed here to use `array_filter` function to filter urls array to return only valid string urls and that's working fine but why not adding bailout condition inside this method

https://github.com/wp-media/wp-rocket/blob/6f76ca8f392db34d76481f1a23185cd2f781dbee/inc/Engine/Cache/Purge.php#L49

like the following:-

```
if( ! is_string($url) ){
    return;
}
```

so here we will save some time not to loop on urls (array_filter) and get only strings and then loop again to call `purge_url`.

What do you think?

remyperona:
Well in my opinion, we should not replace one for another, but do both. It's important to keep the filtering in this method, but it can also be useful to add the check for string in the `purge_url()` one.

engahmeds3ed:
> Well in my opinion, we should not replace one for another, but do both. It's important to keep the filtering in this method, but it can also be useful to add the check for string in the `purge_url()` one.

Thanks for looking into that.
I can see that this method `get_post_terms_urls` is private so it's not used outside this class, but I'm fine to guard its return and also guard purge_url method.

I pushed this small fix.

mostafa-hisham (approved):

Mai-Saad (approved):
Working fine
[testrail-report-331.pdf](https://github.com/wp-media/wp-rocket/files/7757011/testrail-report-331.pdf)

PR #4608, opened 2021-12-23
Fixes #3009 Remove CPCSS style tag when media isn't "all" or "screen"

### LINKED ISSUE #3009: CPCSS style tag is not removed when link with attribute media=print is found.
- Made sure you’re on the latest version 👍 
- Used the search feature to ensure that the bug hasn’t been reported before 👍 

**Describe the bug**
After page is loaded, use the elements inspector in Chrome, and look at the `<head>` - The CPCSS style tag that WP Rocket generates is not removed if there are links containing attribute `media=print`

**Preconditions**
Reproducible in Chrome.

**To Reproduce**
1. Activate Theme  **Twenty Twenty**
2. Enable **Optimize CSS delivery**
3. Navigate to any page and observe the `<head>` with the dev tool inspector.

**Expected behavior**
The CPCSS style tag should be removed

**Actual behavior**
The CPCSS style tag is not removed

**Screenshots**
https://i.imgur.com/cwxp51n.png

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

vmanthos:
## Description

In the `async_css()` method a check has been added to prevent processing CSS files whose `media` isn't `all` or `screen`.

Files with no `media` will still be processed.

Fixes #3009 

## Type of change

No.

## How Has This Been Tested?

- [x] I checked files that `media` was:
- `all`
- `screen`
- values other than `all` or `screen`
- no `media`

# Checklist:


engahmeds3ed:
Thanks @vmanthos for the PR.
I was thinking about that issue and want to share some thoughts with you so we can think together to get the best solution here:-

The PR's approach to play with changing the `stylesheet` to `preload` at `rel` without touching the JS code that is responsible for removing CPCSS code.

I think adjusting this JS code is better by changing the following 

https://github.com/wp-media/wp-rocket/blob/228e3a92ea7ec244ad720308b2ab3487b326cd74/assets/js/cpcss-removal.js#L1-L10

to be 

```
const wprRemoveCPCSS = () => {
	let preload_stylesheets = document.querySelectorAll( 'link[data-rocket-async="style"][rel="preload"]' );

	if ( preload_stylesheets ) {
		for ( let stylesheet_index = 0;stylesheet_index < preload_stylesheets.length;stylesheet_index++ ){
			let media = preload_stylesheets[stylesheet_index].getAttribute('media') || 'all';
			if( window.matchMedia(media).matches ){
				setTimeout( wprRemoveCPCSS, 200 );
				return;
			}
		}
		return;
	}

	const elem = document.getElementById( 'rocket-critical-css' );
	if ( elem && 'remove' in elem ) {
		elem.remove();
	}
};
```

it's just a prototype but I believe it will fix the issue.
give it a try so we can talk about it.

remyperona:
@engahmeds3ed Can you explain the benefits of your approach?

engahmeds3ed:
> @engahmeds3ed Can you explain the benefits of your approach?

OK Boss, Think of it as follows:-
- We have a page with the following stylesheet
```
<link rel='stylesheet' href='test.css' media="(min-width: 700px)" />
```

At any normal page, this stylesheet will be applied in devices that have width more than or equals 700px.

When CPCSS is enabled based on the current PR's approach, we will not preload it and this will increase the page's stylesheets that will load beside the generated CPCSS styles so we will not take the benefit of CPCSS feature.

But in my proposed approach, this stylesheet will be preloaded and the CPCSS styles will be removed correctly from the page if this media condition is met or not.

@Tabrisrp  What do you think about that?

remyperona:
@engahmeds3ed I see, it makes sense and seems to be a more optimized approach. Looking at the code, shouldn't the `setTimeout( wprRemoveCPCSS, 200 );` be called only once, if there was at least one valid match, instead of calling it inside the `for` loop?

engahmeds3ed:
@Tabrisrp Thanks Boss.

Will check that tomorrow and test it to make sure that it's working or not, I just tried the idea on a normal site and it's working but need to test it inside WPR.

remyperona:
@engahmeds3ed have you been able to test and validate your proposed approach?

engahmeds3ed:
Will check that tonight and back to you, sorry for being late here.

engahmeds3ed:
Validated the following code and it's working like the attached video
```
const wprRemoveCPCSS = () => {
	let preload_stylesheets = document.querySelectorAll( 'link[data-rocket-async="style"][rel="preload"]' );
	console.log('preload_stylesheets', preload_stylesheets)
	if ( preload_stylesheets && preload_stylesheets.length > 0 ) {
		console.log('preload_stylesheets is found')
		for ( let stylesheet_index = 0;stylesheet_index < preload_stylesheets.length;stylesheet_index++ ){
			let media = preload_stylesheets[stylesheet_index].getAttribute('media') || 'all';
			console.log('media', media);
			if( window.matchMedia(media).matches ){
				console.log('matched, should wait 200ms and try remove again')
				setTimeout( wprRemoveCPCSS, 200 );
				return;
			}
		}
	}

	const elem = document.getElementById( 'rocket-critical-css' );
	if ( elem && 'remove' in elem ) {
		elem.remove();
	}
};

if ( window.addEventListener ) {
	window.addEventListener( 'load', wprRemoveCPCSS );
} else if ( window.attachEvent ) {
	window.attachEvent( 'onload', wprRemoveCPCSS );
}
```

https://user-images.githubusercontent.com/15707971/149470592-66846fee-4aa4-4baf-a78b-4385cf01ee47.mp4

BTW, we can remove those logs, I just added them to validate.

remyperona:
Let's go with that approach then

engahmeds3ed:
OK, will update the PR now

engahmeds3ed:
@Tabrisrp 
Do we need to remove this part?
https://github.com/wp-media/wp-rocket/blob/10656929a7514ce4cfaa907778e1a9fc47600f03/inc/Engine/CriticalPath/CriticalCSSSubscriber.php#L664-L666

remyperona:
@engahmeds3ed I think it's better to keep it because it's unnecessary to async loading of a stylesheet with the media print, it's not loaded anyway by the browser.

remyperona (approved):

remyperona:
@engahmeds3ed Can you also update the fixture for the tests?

Mai-Saad:
@engahmeds3ed Can still see the Load Async tag with this test template https://new.rocketlabsqa.ovh/cpcss-media2/ (not sure if the problem with the template itself). Can you please check it? :pray:

engahmeds3ed:
This is because this page has two CSS file returns with 404.
We have two options here but I think we need @wp-media/productrocket opinion here:-

1. Add `onerror` attribute along with `onload` to change the rel to stylesheet
2. Use `onerror` to remove the attribute `data-rocket-async`

In both cases the script will be able to remove CPCSS styles from the page, what do u think?

piotrbak:
@engahmeds3ed Since they're both valid, which one is more secure or requires less processing?

engahmeds3ed:
> @engahmeds3ed Since they're both valid, which one is more secure or requires less processing?

I think the second one is better as it will not change the rel to stylesheet for 404 file.

piotrbak:
@engahmeds3ed Let's go this way then

iCaspar:
@Mai-Saad Can you see if that little patch will clear up the error on the template you mentioned before.

Mai-Saad (commented):
Working as expected now with the following note (Thanks to @engahmeds3ed):
If the following markup is used `<link rel='stylesheet' media='print , (max-width: 900px)' href='/wp-content/rocket-test-data/styles/mediaPrint.css' />`, we are preloading this sheet, but when print, preloaded styles are not applied. If this can be a real case then we need to create a separate GH for this. @piotrbak what do you think?
https://jmp.sh/GX5TRTq
[testrail-report-354.pdf](https://github.com/wp-media/wp-rocket/files/8039889/testrail-report-354.pdf)

PR #4629, opened 2022-01-05
Close 4588 - Enable auto-compatibility XL WooCommerce Sales Triggers with DelayJS

### LINKED ISSUE #4588: Delay JavaScript Execution compatibility with XL WooCommerce Sales Triggers

- [X] Made sure you’re on the latest version: **3.10.5**
- [X] Used the search feature to ensure that the bug hasn’t been reported before

**Describe the bug**
When [XL WooCommerce Sales Triggers](https://xlplugins.com/woocommerce-sales-triggers/) is used with Delay JavaScript Execution, the notification box doesn’t work without triggering any errors in the console : https://jmp.sh/QfDKD3S

Excluding the following fixes the issue:
`/wp-content/plugins/xl-woocommerce-sales-triggers/assets/js/wcst_combined.min.js`

**To Reproduce**
1. Use XL WooCommerce Sales Triggers
2. add a notification in a product
3. activate Delay JS Execution

**Expected behavior**
XL WooCommerce Sales Triggers should work out of the box when Delay JavaScript Execution is enabled.

**Screenshots**
[Screenshot](https://jmp.sh/QfDKD3S)

**Additional context**
HS ticket: https://secure.helpscout.net/conversation/1721065729/313322?folderId=1213662

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

iCaspar:
## Description

This adds a new ThirdParty compatibility to correctly exclude the XL WooCommerce Sales Triggers combined js and the dependent jQuery on WooCommerce single product pages.

Fixes #4588

## Type of change

Yes. Initial grooming plan was to simply add the sales triggers script to the default exclusions list. However, the script is dependent on jQuery, which requires delaying that also, but ONLY on single WC product pages.

## How Has This Been Tested?
- [X] Integration test and fixtures are added for new expected behavior
- [X] Confirmed that scripts are excluded as expected on Product pages, but not on other (homepage) in local environment.

# Checklist:

### DIFF
--- a/inc/ThirdParty/Plugins/Ecommerce/XLWooCommerceSalesTriggersSubscriber.php
+++ b/inc/ThirdParty/Plugins/Ecommerce/XLWooCommerceSalesTriggersSubscriber.php
@@ -32,15 +32,19 @@ class XLWooCommerceSalesTriggersSubscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'rocket_delay_js_exclusions' => [ 'modify_delayjs_exclusions_on_product_pages', 10, 2 ]
+			'rocket_delay_js_exclusions' => [ 'modify_delayjs_exclusions_on_product_pages', 10, 2 ],
 		];
 	}
 
 	/**
 	 * Modify delay JS excluded scripts for compatibility.
+	 *
+	 * @param array $exclusions The initial Delay JS exclusion list.
+	 *
+	 * @return array The filtered Delay JS exclusion list.
 	 */
 	public function modify_delayjs_exclusions_on_product_pages( array $exclusions ) {
-		if ( empty( rocket_get_constant('WCST_VERSION' ) ) ) {
+		if ( empty( rocket_get_constant( 'WCST_VERSION' ) ) ) {
 			return $exclusions;
 		}
 
@@ -55,7 +59,7 @@ class XLWooCommerceSalesTriggersSubscriber implements Subscriber_Interface {
 					'xl-woocommerce-sales-triggers/assets/js/wcst_combined.min.js',
 					'/jquery-?[0-9.](.*)(.min|.slim|.slim.min)?.js',
 					'js-(before|after)',
-					'(?:/wp-content/|/wp-includes/)(.*)'
+					'(?:/wp-content/|/wp-includes/)(.*)',
 				]
 			)
 		);
--- a/inc/ThirdParty/Plugins/Ecommerce/XLWooCommerceSalesTriggersSubscriber.php
+++ b/inc/ThirdParty/Plugins/Ecommerce/XLWooCommerceSalesTriggersSubscriber.php
@@ -1,0 +1,63 @@
+<?php
+
+declare( strict_types=1 );
+
+namespace WP_Rocket\ThirdParty\Plugins\Ecommerce;
+
+use WP_Rocket\Engine\Optimization\DelayJS\HTML;
+use WP_Rocket\Event_Management\Subscriber_Interface;
+
+class XLWooCommerceSalesTriggersSubscriber implements Subscriber_Interface {
+
+	/**
+	 * Delay JS HTML class.
+	 *
+	 * @var HTML
+	 */
+	private $delayjs_html;
+
+	/**
+	 * WooCommerceSubscriber constructor.
+	 *
+	 * @param HTML $delayjs_html DelayJS HTML class.
+	 */
+	public function __construct( HTML $delayjs_html ) {
+		$this->delayjs_html = $delayjs_html;
+	}
+
+	/**
+	 * Get the subscribed events array.
+	 *
+	 * @return array|array[]
+	 */
+	public static function get_subscribed_events() {
+		return [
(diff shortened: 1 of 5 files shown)


inc/ThirdParty/Plugins/Ecommerce/XLWooCommerceSalesTriggersSubscriber.php
```diff
@@ -0,0 +1,67 @@
...
+		}
+
+		return array_unique(
+			array_merge(
+				$exclusions,
+				[
+					'xl-woocommerce-sales-triggers/assets/js/wcst_combined.min.js',
+					'/jquery-?[0-9.](.*)(.min|.slim|.slim.min)?.js',
+					'js-(before|after)',
+					'(?:/wp-content/|/wp-includes/)(.*)',
```
remyperona: are those also needed for the compatibility?

inc/ThirdParty/Plugins/Ecommerce/XLWooCommerceSalesTriggersSubscriber.php
```diff
@@ -0,0 +1,67 @@
...
+	 * @param array $exclusions The initial Delay JS exclusion list.
+	 *
+	 * @return array The filtered Delay JS exclusion list.
+	 */
+	public function modify_delayjs_exclusions_on_product_pages( array $exclusions ) {
+		if ( empty( rocket_get_constant( 'WCST_VERSION' ) ) ) {
+			return $exclusions;
+		}
+
+		if ( 'product' !== get_post_type() || ! is_single() ) {
```
remyperona: In the WC context, you can use `is_product()` directly instead

remyperona (changes requested):

inc/ThirdParty/Plugins/Ecommerce/XLWooCommerceSalesTriggersSubscriber.php
iCaspar: Nope!

inc/ThirdParty/Plugins/Ecommerce/XLWooCommerceSalesTriggersSubscriber.php
iCaspar: Good point.

inc/ThirdParty/Plugins/Ecommerce/XLWooCommerceSalesTriggersSubscriber.php
```diff
@@ -0,0 +1,65 @@
...
+	 * @param array $exclusions The initial Delay JS exclusion list.
+	 *
+	 * @return array The filtered Delay JS exclusion list.
+	 */
+	public function modify_delayjs_exclusions_on_product_pages( array $exclusions ) {
+		if ( empty( rocket_get_constant( 'WCST_VERSION' ) ) ) {
+			return $exclusions;
+		}
+
+		if ( ! is_product() || ! is_single() ) {
```
remyperona: won't the `is_single()` exclude jquery on any post type single page?

iCaspar: The logic here is, we're going to bail out whenever it's NOT a product OR whenever it's NOT single. So, I think, checking against not/OR in both sides will allow it to add the exclusions only when it IS a product AND IS single.

iCaspar: In other words, when it is not a product, it will meet the condition before it ever checks for single -- so it will have bailed out and not affect singles of other types.

remyperona: Checking back, `is_single()` only applies to the `post` post type, not all of them (`is_singular()` does).

Also, the underlying code for `is_product()` is `is_singular( array( 'product' ) )` so in the end the "single" check is already done at that point.

So I would suggest only keeping `is_product()` there, as it's going to check everything needed, what do you think?

iCaspar: Indeed, that will work. Thanks.

remyperona (approved):

Mai-Saad:
Couldn't reproduce the main problem on the test site, however, please find the notes below:
- Using the test template (not product page), can see a jQuery console error while Defer JS is enabled => on both trunk and branch (this error is not in nowprocket page)
- Q: do we need to exclude anything for` jquery-migrate`?
@Tabrisrp @piotrbak  what do you think? do we need to do further change(s) here?

Note for retest:
The script `/wp-content/plugins/xl-woocommerce-sales-triggers/assets/js/wcst_combined.min.js` is only delayed on the product page now (thanks to @alfonso100 )

piotrbak:
@Mai-Saad I think we need to reproduce the issue to move forward. For future exclusions, we need to rely on product pages and environemnts, not the templates.

PR #4632, opened 2022-01-05
Closes #4627 Update renewal banners

### LINKED ISSUE #4627: Update Texts for the Expiring Soon and Expired Banners
We need to update the text displayed in the Expiring Soon and Expired banners in the plugin. 

You'll find the new texts and translation [here.](https://docs.google.com/document/d/1gSwxpp0VikIDJkO4uRWERXZX-lepXmCJlGdTtKIuLKo/edit?usp=sharing)

remyperona:
## Description

Update the renewal banners to change the texts, and reflect the changes for the renewal policy.

Fixes #4627

## Type of change

No

## How Has This Been Tested?

- [x] Updated unit and integration tests

# Checklist:

### DIFF
--- a/inc/Engine/License/Renewal.php
+++ b/inc/Engine/License/Renewal.php
@@ -85,7 +85,7 @@ class Renewal extends Abstract_Render {
 			return;
 		}
 
-		echo $this->generate( 'renewal-expired-banner', $this->get_banner_data() ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		echo $this->generate( 'renewal-expired-banner', [ 'renewal_url' => $this->user->get_renewal_url() ] ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
 	}
 
 	/**
--- a/inc/Engine/License/views/renewal-expired-banner.php
+++ b/inc/Engine/License/views/renewal-expired-banner.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Renewal soon banner.
+ * Renewal expired banner.
  *
  * @since 3.7.5
  */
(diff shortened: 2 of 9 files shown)


iCaspar (approved):

Mai-Saad:
@Tabrisrp please find the notes below While exploring the PR:
1- Translations are still not handled (IT/ES not translated, while FR/DE title only translated) => Seems this will be ready after deployment to develop, right?
2- % is still 30% and still appearing for users registered at 2022(i.e on Jan 10, 2022), can we handle this on the branch?

remyperona:
@Mai-Saad
1. Translations are done later yes.
2. The values are going to be updated on the API side at a later time, and it will automatically be reflected on the plugin side.

Mai-Saad (commented):
[testrail-report-337.pdf](https://github.com/wp-media/wp-rocket/files/7845136/testrail-report-337.pdf)
Failed tests due to:
- known issue(s) on the trunk
- translation is not there yet (shall be updated with the deployment)
- new % not displayed in the banner (shall be updated with API updates)
- renew now link  while expired banner shows the discounted price instead of the original one (I think this shall be fixed with API too, right @Tabrisrp) 

Possible enhancements (can be in separate ticket(s)):
- use 0 instead of 00 to hrs, mins, secs .. @piotrbak what do you think?
Screenshot from 2022-01-11 09-43-01
- Do we need to auto dismiss the banner when all is 0? (note: if no auto dismiss then after API updates we need to check to renew now after time is 0 and see which price is displayed)
- Clicking refresh for user info better update banner with no need to manually refresh the page

remyperona:
@piotrbak Do we want to add those possible enhancements to the backlog?

piotrbak:
@Mai-Saad Those will need a separate issue for sure. When timer reaches 0, we should hide it, definitely. 

Let's discuss remaining ones on the Slack and we'll decide where to move forward.

PR #4637, opened 2022-01-06
Closes #4586 exclude from preload links URL patterns used by referral plugins

### LINKED ISSUE #4586: Preload Links inflating Thirsty Affiliates or Pretty Links referral click count
- Made sure you’re on the latest version - Yes
- Used the search feature to ensure that the bug hasn’t been reported before - Yes

**Describe the bug**
The Preload Links feature is inflating the Thirsty Affiliates or Pretty Links referral click count, when the links are setup as internal, because the visitors are preloading the referral links when they mouse over the specific links.

**To Reproduce**
1. Install Thirsty Affiliates on a website
2. Check stats for click count
3. Enable Preload Links
4. Check how stats for click count increased, as visitors hovered on links
5. Disable Preload Links
6. Check stats once more, click count should be decreased

This might be difficult to reproduce from scratch, but Nadia provided us with the click stats for the [WPBeginner](https://www.wpbeginner.com/) site where we can see how inflated the count was during the period Preload Links was enabled. See all at  https://wp-media.slack.com/archives/C08N8J6VC/p1639408185296700?thread_ts=1638814946.173200&cid=C08N8J6VC..

**Expected behavior**

Preload Links shouldn't inflate referral statistics. 

The following fix was suggested:

> It might be a good idea to auto-exclude `/refer/` and `/go/` paths from ThirstyAffiliates and Pretty Link plugins to ensure other users don't run into this issue.

**Additional context**
**Ticket:** https://secure.helpscout.net/conversation/1716780906/312607/
**Slack discussion:** https://wp-media.slack.com/archives/C08N8J6VC/p1638814946173200

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Add a new filter `rocket_preload_links_exclusions` to be used to add additional exclusions for the preload links.

Also add a default list of exclusions, and a compatibility file for ThirstyAffiliates plugin, updating the exclusion value based on the prefix value defined in the options.

Fixes #4586 

## Type of change

No

# Checklist:


piotrbak:
@Tabrisrp Can we make it a filter? Sorry for not mentioning this with AC

inc/Engine/Preload/Links/Subscriber.php
```diff
@@ -181,6 +181,9 @@ private function get_uris_to_exclude( $use_trailing_slash ) {
 			$uris .= '|' . str_replace( $site_url, '', $uri );
 		}
 
+		// exclude URL patterns used by referral plugins.
+		$uris .= '|/refer/|/go/';
```
engahmeds3ed: what do u think about adding a filter here so we can change with compatibility files whenever needed?
I believe that some plugins uses `recommend` and others open it to the user so he can change it, for this reason, using a filter will help a lot.

engahmeds3ed: Sorry I didn't see @piotrbak comment above :D

iCaspar (approved):

Mai-Saad:
@Tabrisrp while exploring the branch, noticed that if the filter was `Null, false, ' '`, the preload won't work at all. Can we handle this?

Mai-Saad:
@Tabrisrp while exploring the branch, noticed that if the filter was `Null, false, ' '`, the preload won't work at all. Can we handle this?

remyperona:
@Mai-Saad it should be fixed now

Mai-Saad:
@Tabrisrp Currently if we used the filter to exclude certain URL(s) i.e
```
add_filter( 'rocket_preload_links_exclusions', function ($links){
  $links = '/projects';
    return  $links;
} );
```
Then the auto exclusion for Thirsty Affiliate won't work.

While defined prefix is` go` + filter is added:
https://jmp.sh/0bzZXKL

While defined prefix is` go` + no filter is added:
https://jmp.sh/QKa20ov

Shouldn't auto exclusion work fine while using a filter to exclude other links?

remyperona:
@Mai-Saad I pushed a change to make sure the thirsty affiliates prefix is always applied when the plugin is enabled

Mai-Saad:
Currently:
1- If no filter => working correctly with/without TA
2- If have a filter using the correct format below => Working fine while no TA. With TA enabled, TA default prefix will only be applied (overwrite the filter) @piotrbak What do you think? Do we need to do any changes here or in a separate ticket?
```
add_filter( 'rocket_preload_links_exclusions', function ($links) {
  $links[] = '/projects';
    return  $links;
} );
```
3- If filter not used properly like below => With TA enabled will overwrite filter, With TA disabled the default exclusions won't work
`return  Null;`  or `return [Null, '/projects'] ;`or `return [];` or `return '/projects';`

piotrbak:
I think we need to take care of the 2nd point. We'd expect that filter works with and without TA (or any other future compatibility) enabled.

remyperona:
@piotrbak SO when TA is enabled, we need to remove the default values, but keep any value that could have been set using the filter before the TA callback is called?

What if someone uses the filter to set a value that is part of the default?

piotrbak:
@Tabrisrp The objective is to have the values from filter in all cases:
1. When TA is activated, we'll get their prefix and exclude it + we'll get any values from the filter and exclude them too
2. When TA (or any other future compatibility plugin) is deactivated, we'll get the default values + filtered ones

I believe that someone might want to exclude different links than the ones specified in the compatibility code.

remyperona:
Ok, PR is updated to handle that case. When TA is enabled, the default values are removed, but values added using the filter are preserved, in addition to the TA exclusion.

Mai-Saad:
Currently, if TA is enabled along with the filter, both will be excluded except if the filter is matching one of the defaults & != TA prefix 
i.e if the TA prefix is `go` while we excluded `/refer/` at filter then `/refer/` won't be excluded but if we excluded `/refer` in the filter, it will work. I think this is fine but needs doc. What do you think @piotrbak? (if fine too then we are ready to merge)
[testrail-report-353.pdf](https://github.com/wp-media/wp-rocket/files/8031102/testrail-report-353.pdf)

piotrbak:
@Mai-Saad Yes, that's fine

Mai-Saad:
Pinging @camilamadronero seems we may need a doc update here.

PR #4644, opened 2022-01-12
Exclusion for yanovis.Voucher.js from JS defer

DahmaniAdame:
https://secure.helpscout.net/conversation/1727751196/314385


remyperona (approved):

Mai-Saad (approved):
The error `yanovis.Voucher.js` is fixed now with Defer JS

PR #4645, opened 2022-01-12
Fix WP Search with Algolia seach feature

### LINKED ISSUE #4612: Delay JavaScript Execution compatibility with WP Search with Algolia
- [x] Made sure you’re on the latest version `3.10.5`

**Describe the bug**
When Delay JavaScript Execution is used with the [WP Search with Algolia](https://wordpress.org/plugins/wp-search-with-algolia/), the search featyre is not functional and the following console error shows up.

```
Uncaught ReferenceError: algoliasearch is not defined
    at HTMLDocument.<anonymous> ((index):890)
    at e (jquery.min.js?ver=3.6.0:2)
    at t (jquery.min.js?ver=3.6.0:2)
```

Excluding the following from Delay JavaScript Execution fixes the issue: 

```
/wp-search-with-algolia/js/algoliasearch/dist/algoliasearch-lite.umd.js
/wp-search-with-algolia/js/autocomplete-noconflict.js
/wp-search-with-algolia/js/autocomplete.js/dist/autocomplete.min.js
var algolia
```

**To Reproduce**
1. Enable WP Search with Algolia and link it to a working Algolia account
2. Enable Delay JavaScript Execution with no exclusions
3. See error

**Expected behavior**
WP Search with Algolia should work with no issues when using Delay JavaScript Execution. At least after user interaction.

**Screenshots**
image

**Additional context**
jQuery needs to be excluded as well as one of the scripts is initialized using jQuery: 

```
jQuery( function () {
		/* Initialize Algolia client */
		var client = algoliasearch( algolia.application_id, algolia.search_api_key );
```

Ticket - https://secure.helpscout.net/conversation/1736063475/315781/
Added to - https://docs.wp-rocket.me/article/1560-delay-javascript-execution-compatibility-exclusions#wp-search-with-algolia

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

DahmaniAdame:
https://secure.helpscout.net/conversation/1736063475/315781/

### DIFF
--- a/inc/Engine/Optimization/DelayJS/HTML.php
+++ b/inc/Engine/Optimization/DelayJS/HTML.php
@@ -66,6 +66,10 @@ class HTML {
 		'jetpack-boost-critical-css', // Jetpack Boost plugin critical CSS.
 		'wpformsRecaptchaCallback', // WPForms reCAPTCHA v2.
 		'booking-suedtirol-js', // bookingsuedtirol.com widgets.
+		'/wp-search-with-algolia/js/algoliasearch/dist/algoliasearch-lite.umd.js', // WP Search with Algolia search feature.
+		'/wp-search-with-algolia/js/autocomplete-noconflict.js', // WP Search with Algolia search feature.
+		'/wp-search-with-algolia/js/autocomplete.js/dist/autocomplete.min.js', // WP Search with Algolia search feature.
+		'var algolia', // WP Search with Algolia search feature.
 	];
 
 	/**


remyperona (approved):

Mai-Saad:
The issue was reproducible while `jQuery` is excluded in UI. However, on the branch (while excluding jQuery), the original error was fixed but another error was there in the console.
Screenshot from 2022-01-13 13-08-29

@DahmaniAdame can you please check if anything else needs to be excluded.

DahmaniAdame:
@Mai-Saad the issue is fixed adding the `underscore.min.js` and the `wp-util.min.js` libraries to the exclusions: 

```
/wp-includes/js/underscore.min.js
/wp-includes/js/wp-util.min.js
/wp-search-with-algolia/js/algoliasearch/dist/algoliasearch-lite.umd.js
/wp-search-with-algolia/js/autocomplete-noconflict.js
/wp-search-with-algolia/js/autocomplete.js/dist/autocomplete.min.js
var algolia
```

Can you check if it's working properly now on https://new.rocketlabsqa.ovh/youtube-regular-2/?

Mai-Saad:
@DahmaniAdame Please notice that (on branch):
1- Exclusions mentioned here https://github.com/wp-media/wp-rocket/pull/4645#issuecomment-1013653974 are working fine with delay js
2- We have a `jQuery ` error when combine js is enabled (same on the trunk) I think the same exclusions are needed for Combine JS (not sure if we will add this here or on different PR). what do you think?
Screenshot from 2022-01-17 10-38-31


Note:
For the Delay JS problem, it wasn't reproducible on trunk except if we excluded `jQuery` so do we need global exclusions here not site-based? (just confirming that)

piotrbak:
@Mai-Saad Do we have any problem when **nothing** is excluded from Delay JS? If the problem happens only when jQuery is excluded, I don't think we need a global exclusion for that.

Mai-Saad:
@piotrbak on the test site using trunk, the problem with Delay JS is there only when we exclude `jQuery` (if we auto exclude `jQuery` with any other plugin/theme then maybe that was causing the problem at the client site)
While the error with Combine JS is there without adding any exclusions.

piotrbak:
@Mai-Saad @DahmaniAdame I'm not able to reproduce the problem when no exclusions are there. Here's what I'm experiencing:

**Delay JS**
1. No exclusions are being made, no issue
2. `/jquery-?[0-9.](.*)(.min|.slim|.slim.min)?.js` exclusion, no issue
3. `jquery` exclusion, there's a problem mentioned by Adame (this exclusion doesn't seem to be correct though)

**Combine JS**
1. There's a problem, as Algolia inline script is automatically excluded because of dynamic inline exclusions. However, it's dependent on the jQuery, so it should be placed below the combined file. The solution is to move this [exclusion](https://github.com/wp-media/wp-rocket/blob/5ca409236a99429839cb124294ff2b0debcc8533/inc/Engine/Optimization/Minify/JS/Combine.php#L539) to this [array](https://github.com/wp-media/wp-rocket/blob/5ca409236a99429839cb124294ff2b0debcc8533/inc/Engine/Optimization/Minify/JS/Combine.php#L740). 

@DahmaniAdame Could you confirm the Delay JS part on the customer's website?

piotrbak:
Removing it from the 3.11.2 milestone. We still need an exploratory and confirmation here.

PR #4650, opened 2022-01-13
Enhancement - 3523 Add Opposite Image Dimension When Only One Is Given

### LINKED ISSUE #3523: Add Missing Image Dimensions is not applied if only single attribute (width or height) is missing

**Describe the bug**
When image is missing width or height, Add missing image dimensions is not working (it only works only when both Height and Width are missing)

**Precondition**
Having a page with image that is missing height or width

**To Reproduce**
1. Enable add missing image dimensions in media page
2. Clear cache
3. Visit the page mentioned in precondition
4. In source image dimensions are still missing

**Expected behavior**
Image dimensions are added as same as that when both width and height were missing

**Screenshots**
imgDimensions

**Additional context**
Add missing image dimensions is working fine whenever **both** height and width are missing

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

iCaspar:
## Description
This adds to the ImageDimension the additional capability for adding missing dimensions when only one is given.
The current (designed) behavior is to add both dimensions when neither is already present in the HTML.

Here, when a single dimension is given (width, but not height; or height, but not width) we calculate the ratio of the given dimension to the actual image size, then insert the opposite dimension scaled to the same ratio.

Fixes #3523

## Type of change

(See also extended commit messages for other details.)

This is essentially the same approach as the grooming. However a couple additional regexes needed to be modified to handle the "nested in <picture> tag cases, and for re-inserting the updated height_width HTML for cases where only a single dimension is supplied.

## How Has This Been Tested?
Additional unit and integration tests are included for each case (width, no height; height, no width)

# Checklist:

Not sure what documentation about the feature may need updated for this. (@camilamadronero ?)


camilamadronero-zz:
Thanks a lot for the ping, @iCaspar :)
I'll update [this doc](https://docs.wp-rocket.me/article/1366-add-missing-image-dimensions ) with the 3.10.7 version release.

remyperona (approved):

Mai-Saad:
@iCaspar @piotrbak Currently, we are adding missing dimension(s) but noticed the following:
1- If a certain image has "NaN" width or height, the missing dimension will be set to 0 and the following warning is added to debug.log 
` PHP Warning:  A non-numeric value encountered in /var/www/new.rocketlabsqa.ovh/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Media/ImageDimensions/ImageDimensions.php on line 146`
Not sure if this is what we need, was expecting that correct dimensions are added or images not processed and no warning in debug.log. what do you think?

2- If we used an image (that was h300 and w500) but we changed its height in the template to 500 => the calculated width is 833.3333333333 .. Do we need to approx. to a certain digit? (browser approx to 2 digits)

3- If we used this filter `add_filter( 'rocket_specify_dimension_skip_pictures', '__return_false' );` to add dimensions for img inside picture, sometimes added dimensions are different than browser. Do we need GH issue for this? (same on trunk)
Screenshot from 2022-01-14 12-07-34

4- How shall we handle this markup ``, currently we don't add height.. Do we need GH for this? (same on trunk)

iCaspar:
On 1 - We should not have a warning there. I think the expectation should be as you say... if NaN in one dimension, bail out.
On 2 - Indeed, for that html markup the expected values should always be integer, as these are pixel values.
On 3 - Let's handle this in a new issue. Shouldn't be too hard, but it's out of scope for this one.
On 4 - Going to defer to @piotrbak on how to handle that, but I think not adding anything is the right call here. What if the markup were, `width="100" width="250"` for example... which should we try to match the height to?

Mai-Saad (commented):
1,2 are working as expected now. for 3,4 we can add separate GH issues if needed.
[testrail-report-340.pdf](https://github.com/wp-media/wp-rocket/files/7881296/testrail-report-340.pdf)

PR #4655, opened 2022-01-13
Fixess #3570 Use WP tested version from our API

### LINKED ISSUE #3570: Get required WordPress version automatically from the API

**Describe the bug**
Now our plugin states that we support WordPress up to `5.5.1` but I'm pretty sure that we support latest one `5.6.1`

https://github.com/wp-media/wp-rocket/blob/5c698673a22509f5a4194ba8909a696a45f5bb68/wp-rocket.php#L25

**To Reproduce**
1. Install WP Rocket 3.8.3 or any version before latest one (3.8.4)
2. View details of upgrade
3. You will see a message that WP Rocket is tested up to `5.5.1` and not tested with latest WP version `5.6.1`

**Expected behavior**
We should show that we support latest WP release.

**Additional context**
Related ticket: https://secure.helpscout.net/conversation/1406799537/234401/

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

remyperona:
## Description

- Move the files to the new architecture
- Add a new method `add_wp_tested_version()` to update the value of the `tested` field, maintaining a fallback to the constant if needed

Fixes #3570 

## Type of change

No

## How Has This Been Tested?

- [x] Manual test by changing the WP Rocket version to an old one, and checking the compatibility value on the updates screen

# Checklist:

### DIFF
--- a/inc/Engine/Deactivation/Deactivation.php
+++ b/inc/Engine/Deactivation/Deactivation.php
@@ -87,6 +87,7 @@ class Deactivation {
 		delete_transient( 'rocket_check_licence_30' );
 		delete_transient( 'rocket_check_licence_1' );
 		delete_site_transient( 'update_wprocket_response' );
+		delete_site_transient( 'wp_rocket_update_data' );
 
 		// Unschedule WP Cron events.
 		wp_clear_scheduled_hook( 'rocket_rucss_clean_rows_time_event' );
--- a/inc/Engine/Plugin/InformationSubscriber.php
+++ b/inc/Engine/Plugin/InformationSubscriber.php
@@ -54,8 +54,8 @@ class InformationSubscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'plugins_api'        => [ 'exclude_rocket_from_wp_info', 10, 3 ],
-			'plugins_api_result' => [ 'add_rocket_info', 10, 3 ],
+			'plugins_api'              => [ 'exclude_rocket_from_wp_info', 10, 3 ],
+			'plugins_api_result'       => [ 'add_rocket_info', 10, 3 ],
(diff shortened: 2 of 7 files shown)


iCaspar (approved):

Mai-Saad (commented):
Currently, the API-tested version is reflected on the updates page.
[testrail-report-341.pdf](https://github.com/wp-media/wp-rocket/files/7888543/testrail-report-341.pdf)


Notes:
- (already on old versions of WPR) When we deactivate the plugin while update msg is displayed in plugins page, then we click view details => empty msg with the plugin not found will be displayed (if we deleted transients, the upgrade msg won't be displayed so maybe we need to automatically delete transients in this case) can be in another ticket. what do you think @Tabrisrp @piotrbak ?
 https://jumpshare.com/v/KHzVv8oltgZoliwsmm6s
- If we blocked external requests while having update msg then tried to view details of new update without clearing transients 1st, we will see this error => from discussions, this seems expected (it is the case as well for other plugins)
Screenshot from 2022-01-18 09-43-57
- Sometimes if we rollback multiple times, we can see 2 compatibility msg for the latest WP version displayed as follow => As per discussion, probably it's from core side
Screenshot from 2022-01-18 13-30-07

remyperona:
@Mai-Saad I added the deletion of the transient when deactivating the plugin

Mai-Saad:
Our transient is deleted now while deactivation but still, this didn't fix 1st point, we need to delete `update_plugins` as well and refresh the page so the update msg be dismissed. Maybe, something is needed at the core as well. We can open a separate ticket for this one, what do you think @piotrbak?

piotrbak:
That sounds like an edge case, we can have a separate issue for this to collect feedback though.

Mai-Saad (approved):

PR #4658, opened 2022-01-14
Closes #4654 Exclude StatsCounter from minify JS and delay JS execution

### LINKED ISSUE #4654: Compatibility with StatsCounter
- Made sure you’re on the latest version 🆗 
- Used the search feature to ensure that the bug hasn’t been reported before 🆗 

**Describe the bug**
Stats are not beign collected when Minify and Delay JS are enabled.

**To Reproduce**
1. Enable StatsCounter plugin and login to their dashboard
2. Enable Minify and Delay JS
3. Visit the page in incognito
4. Stats are not being collected

**Expected behavior**
Stats should be collected all the time

**Additional context**
We need to exclude from JS Minification:
1. `statcounter.com/counter/counter.js` so it's not rewritten to the domain name
and from the Delay JS:
1. `statcounter.com/counter/counter.js`
2. `var sc_project`

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Exclude StatsCounter scripts from minify JS and delay JS execution

Fixes #4654 

## Type of change

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DelayJS/HTML.php
+++ b/inc/Engine/Optimization/DelayJS/HTML.php
@@ -67,6 +67,8 @@ class HTML {
 		'wpformsRecaptchaCallback', // WPForms reCAPTCHA v2.
 		'booking-suedtirol-js', // bookingsuedtirol.com widgets.
 		'/gravityforms/js/conditional_logic.min.js', // Gravity forms conditions.
+		'statcounter.com/counter/counter.js', // StatsCounter.
+		'var sc_project', // Statscounter.
 	];
 
 	/**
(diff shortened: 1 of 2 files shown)


iCaspar (approved):

Mai-Saad (commented):
StatsCounter is excluded now from Minify / Combine JS and from Delay JS. 

Note: on the trunk, the counter was working while MinifyJS and Delay JS are enabled but we must have user interaction. excluding only from delay js will make it works without user interaction so don't think we need exclusion for Minify JS. @Tabrisrp @piotrbak What do you think?

piotrbak:
@Mai-Saad We need exclusion from Minify JS, because in one of StatCounter states:
1. Invisible
2. Visible

It adds external JS script. If we minify and rewrite it to our domain, the DelayJS exclusion won't work.

Mai-Saad (approved):

PR #4661, opened 2022-01-17
Fixes #2852 Update feeds cache exclusion pattern

### LINKED ISSUE #2852: Incorrect RegEx exclusion matches on URLs that contain the string "feed"
**WP Rocket version:** 3.6.1
Issue not found on the Github repository.
**Not happening on the previous major version**3.5.5.1

**Describe the bug**
WP Rocket config file applies this RegEx:

$rocket_cache_reject_uri = '/(.+/)?feed/?.+/?|/(?:.+/)?embed/|/(index\.php/)?wp\-json(/.*|$)';

but the URL is excluded from the cache if a customer adds some URL path containing the word "feed" like _domain.com/category/**feed**ing/_

**To Reproduce**
1. Create a category called "feeding"
2. Go to this category page (example: yourdomain.com/category/feeding/)
3. Go to page source and notice that isn't cached.

**Expected behavior**
Only URLs with the string /feed/ on the path should be excluded automatically.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Update the feeds cache exclusion pattern to prevent unexpected matching when the URL contains `feed` in its slug, but is not a WordPress feed.

Performs an update of the config file on plugin update to 3.10.8, to make sure its using the new pattern.

Fixes #2852 

## Type of change

No

## How Has This Been Tested?

- [x] Tested with RegEx101 with various URLs
- [x] Tested on user website by the support team

# Checklist:

@wp-media/tech-support This might require an update to the helper plugin to cache feed once this is released.

### DIFF
--- a/inc/admin/upgrader.php
+++ b/inc/admin/upgrader.php
@@ -366,10 +366,6 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 		rocket_clean_domain();
 	}
 
-	if ( version_compare( $actual_version, '3.6.1', '<' ) ) {
-		rocket_generate_config_file();
-	}
-
 	if ( version_compare( $actual_version, '3.7', '<' ) ) {
 		rocket_clean_minify( 'css' );
 		rocket_generate_advanced_cache_file();
@@ -401,6 +397,10 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 			rocket_rrmdir( $cache_path . 'used-css' );
 			update_option( rocket_get_constant( 'WP_ROCKET_SLUG' ), $options );
 		}
+	}
+
+	if ( version_compare( $actual_version, '3.10.8', '<' ) ) {
+		rocket_generate_config_file();
 	}
 }
 add_action( 'wp_rocket_upgrade', 'rocket_new_upgrade', 10, 2 );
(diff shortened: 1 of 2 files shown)


engahmeds3ed:
@Tabrisrp 
Plz check the following link: https://regex101.com/r/2ewwV6/1

Is it valid to detect `https://ahmed.com/posts/any-feed/` as feed url?

remyperona:
@engahmeds3ed There is the leading slash missing in your expression, see the updated one: https://regex101.com/r/2ewwV6/2

iCaspar (approved):

Mai-Saad (approved):
Working as expected with the new version and after update
[testrail-report-348.pdf](https://github.com/wp-media/wp-rocket/files/7993284/testrail-report-348.pdf)

Notes:
- The plugin mentioned [here](https://github.com/wp-media/wp-rocket/issues/2852#issuecomment-912574193) is not working on the trunk nor on the branch (home/feed won't be cached while the plugin is active)
- NeverCacheURL for `/feeding` is not working on the branch although it was working on the trunk. However, exclude`/feeding/` will work on the branch as on the trunk. (Thanks to @alfonso100 ) do we need to do anything for that @Tabrisrp ?

remyperona:
@alfonso100 The helper plugin will need to be updated to use the new pattern for the feed `/(?:.+/)?' . $wp_rewrite->feed_base . '(?:/(?:.+/?)?)?$`

For the second point, the exclusion should reflect the usage of a trailing slash (or not) in the site configuration. the `/feeding` exclusion was a side effect of the incorrect pattern in this case.

Mai-Saad (approved):
Note: we need to test the helper plugin after being updated

PR #4674, opened 2022-01-20
Closes #3727 Add missing dimensions on SVG files

### LINKED ISSUE #3727: Add missing image dimensions not working for SVG images
- Running on the latest version

**Describe the bug**
The "Add missing image dimensions" feature doesn't seem to be working with SVG images. I am talking about SVG images added using the `` tag. I read that the feature will not work on pure SVG. But is that the case for the .svg files as well? If so, isn't there any workaround for this?

**To Reproduce**
1. Visit any page with SVG images (.svg format) with "Add missing image dimensions" feature enabled.
2. Check the HTML output.

**Expected behavior**
Would be great if we add the support for the SVGs as well as more and more websites are adopting the SVG format over traditional image formats.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Add a new method to get the dimensions values for SVG files. Also extract to its own method the code to add a dimension when only one is already set.

Fixes #3727

## Type of change

No

## How Has This Been Tested?

- [x] Added new tests to validate

# Checklist:


iCaspar (approved):

iCaspar:
Ready for QA. Pinging also @camilamadronero to check for any possible docs updates for 3.10.8.

Mai-Saad:
@Tabrisrp Can't see missing dimensions added for SVG here https://new.rocketlabsqa.ovh/image-template-1/, do we need to use a certain filter? 
Screenshot from 2022-01-31 13-47-12

remyperona:
It's working fine on my local with this image, there is no filter to be used. I'll investigate what happens on the QA site

remyperona:
Updated the regex flag to make it ungreedy, and it's working on the test site now

Mai-Saad:
@Tabrisrp can see the dimension(s) added now on the test site but it's added too for excluded cases like
```
<img src="/wp-content/rocket-test-data/images/flower.svg" Width='auto'>
<img src="/wp-content/rocket-test-data/images/flower.svg" Width='NaN'>
<img src="/wp-content/rocket-test-data/images/flower.svg" data-actual-height>
```
will be 

```
<img width="148mm" height="210mm" src="/wp-content/rocket-test-data/images/flower.svg" >
  <img width="148mm" height="210mm" src="/wp-content/rocket-test-data/images/flower.svg" >
  <img width="148mm" height="210mm" src="/wp-content/rocket-test-data/images/flower.svg" data-actual-height>
```
Shouldn't auto exclusions mentioned [here](https://docs.wp-rocket.me/article/1366-add-missing-image-dimensions) apply on svg as well?

remyperona:
@Mai-Saad I pushed a fix for the first 2 cases, which was happening because the `Width` is capitalized, and we didn't have case insensitivity in the RegEx.

For the last case, I don't understand how it's noted as an exclusion in the doc, the only exclusions we have in the code are `data-lazy-original` and `data-no-image-dimensions`.

remyperona:
Discussed this with Piotr.

What should be excluded:
- `data-lazy-original` and `data-no-image-dimensions`
- `data-height-percentage` when using Divi

When there is attributes using `height` and `width`:
- Both should be set, if there is only one, it should not be excluded
- The markup should be `width|height="100"`, the attribute needs to have a value

so `data-actual-height` alone is not supposed to be excluded for example.

Mai-Saad:
@camilamadronero Seems we need documentation update [here](https://docs.wp-rocket.me/article/1366-add-missing-image-dimensions)
Note: The case of having `width="100" width="100"` is currently processed

Mai-Saad:
@Tabrisrp Confirmed that 1st and 2nd cases mentioned above are handled now (note: `data-height-percentage` is excluded independent on the theme)
But with the latest change:
1- There is `mm` added to SVG (in case of no height nor width is there), can you please check that. (note: the actual width /height for the SVG are Width='524' Height='744', do we expect to see these dimensions when both width/height are missing?)
Screenshot from 2022-02-03 13-07-02
2- can see these logs (this one was fixed on the trunk):
```
[03-Feb-2022 11:55:56 UTC] PHP Notice:  A non well formed numeric value encountered in /var/www/new.rocketlabsqa.ovh/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Media/ImageDimensions/ImageDimensions.php on line 306
[03-Feb-2022 11:55:56 UTC] PHP Notice:  A non well formed numeric value encountered in /var/www/new.rocketlabsqa.ovh/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Media/ImageDimensions/ImageDimensions.php on line 308
[03-Feb-2022 11:55:56 UTC] PHP Notice:  A non well formed numeric value encountered in /var/www/new.rocketlabsqa.ovh/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Media/ImageDimensions/ImageDimensions.php on line 291
[03-Feb-2022 11:55:56 UTC] PHP Notice:  A non well formed numeric value encountered in /var/www/new.rocketlabsqa.ovh/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Media/ImageDimensions/ImageDimensions.php on line 293
```

remyperona:
I updated the code to format SVG width/height value before trying to use it.

If the value has:
- no unit, we use the value directly
- a px unit, we remove the unit and use the numerical value
- another unit, we bail-out, and it will fallback to the viewBox values (if set)

This should also avoid the PHP notices from the previous comment.

Mai-Saad:
> I updated the code to format SVG width/height value before trying to use it.
> 
> If the value has:
> 
> * no unit, we use the value directly
> * a px unit, we remove the unit and use the numerical value
> * another unit, we bail-out, and it will fallback to the viewBox values (if set)
> 
> This should also avoid the PHP notices from the previous comment.

Confirmed that no PHP notice now and the latest rules for SVG are applied (notice that: when we fallback to viewBox while having a single dimension in source=> we will calculate the other dimension based on viewBox %).

But the only problem now is that when we use filter to add dimensions for external and/or query string:
it is working for external URL of SVG but not for SVG Query string (relative or absolute URL)
Used filter => `add_filter( 'rocket_specify_image_dimensions_for_distant', '__return_true' );`
Test page => https://new.rocketlabsqa.ovh/image-template-1/
@Tabrisrp What do you think? Will we handle this on the current PR?

remyperona:
@Mai-Saad Latest commit should fix this specific issue

Mai-Saad (approved):
Working fine now
[testrail-report-352.pdf](https://github.com/wp-media/wp-rocket/files/8025645/testrail-report-352.pdf)

PR #4675, opened 2022-01-20
Closes #3520 Exclude PWA service worker URL from cache

### LINKED ISSUE #3520: Exclude wp.serviceworker JS file created by PWA plugin from being cached as a page

**Describe the bug**
When PWA plugin is enabled, it is loading a JavaScript file with the path `example.com/wp.serviceworker`. The file is generating a cache file as for a non-existing page/post with the URL `example.com/wp.serviceworker` 

**To Reproduce**
1. Install PWA plugin https://wordpress.org/plugins/pwa/
2. Clear cache in WP Rocket
3. Open the site in the browser
4. Check the cache folder. The cache/example.com/wp.serviceworker cache file will be generated.

**Expected behavior**
The file has to be excluded from caching.

**Screenshots**

**Additional context**
Adding `/wp.serviceworker` to the Never Cache URS(s) resolves the issue.

Issue thread with the plugin developers: 
https://wordpress.org/support/topic/pwa-is-not-compatible-with-cloudflare-rocket-loader/

> `/wp.serviceworker` returns with these HTTP response headers:
> `cache-control: no-cache`
> `content-type: text/javascript; charset=utf-8`
> WP Rocket should include this in the list of URLs that are excluded, and/or avoid caching anything that returns with a non-HTML content type.

- [x] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Create a compatibility file for the PWA plugin, excluding the service worker URL from cache to prevent any issue.

Fixes #3520 

## Type of change

- [x] 3rd party compatibility

A difference on the conditional check to verify if the 3rd party plugin is enabled.

## How Has This Been Tested?

- [x] Added tests to validate the change

# Checklist:

### DIFF
--- a/inc/Plugin.php
+++ b/inc/Plugin.php
@@ -289,6 +289,7 @@ class Plugin {
 			'revolution_slider_subscriber',
 			'wordfence_subscriber',
 			'ezoic',
+			'pwa',
 		];
 
 		$host_type = HostResolver::get_host_service();
--- a/inc/ThirdParty/Plugins/PWA.php
+++ b/inc/ThirdParty/Plugins/PWA.php
@@ -1,0 +1,34 @@
+<?php
+declare(strict_types=1);
+
+namespace WP_Rocket\ThirdParty\Plugins;
+
+use WP_Rocket\Event_Management\Subscriber_Interface;
+
+class PWA implements Subscriber_Interface {
+	/**
+	 * Subscribed events.
+	 */
+	public static function get_subscribed_events() {
+		return [
+			'rocket_cache_reject_uri' => 'exclude_service_worker',
+		];
+	}
+
+	/**
+	 * Excludes the PWA service worker URL
+	 *
+	 * @param array $excluded Array of excluded URLs.
+	 *
+	 * @return array
+	 */
+	public function exclude_service_worker( $excluded ): array {
+		if ( ! function_exists( 'wp_get_service_worker_url' ) ) {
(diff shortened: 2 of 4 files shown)


iCaspar (approved):

Mai-Saad (commented):
Currently, access to any page won't auto create `wp.serviceworker` cache folder but if we access domain/wp.serviceworker i.e https://new.rocketlabsqa.ovh/wp.serviceworker/ => this will be cached, Was expecting that it won't be cached. What do you think @Tabrisrp?
Screenshot from 2022-02-01 11-34-25

Screenshot from 2022-02-01 11-31-53

remyperona:
The URL of the service worker shouldn't have a trailing slash, this is how it is registered in the script.

@piotrbak What do you think, should we also prevent the trailingslashed URL from being cached in case someone accesses it like this, even though it's unexpected?

piotrbak:
@Tabrisrp Yes, let's exclude it too, so we will be 100% sure it's resolved for all cases.

remyperona:
ok I updated the pattern, it's ready to be tested again 👍🏼

Mai-Saad (approved):
Working fine now even with URL access.
[testrail-report-344 (1).pdf](https://github.com/wp-media/wp-rocket/files/7985100/testrail-report-344.1.pdf)

PR #4676, opened 2022-01-21
Closes #1599 Reduce number of queries on the admin

### LINKED ISSUE #1599: Database queries in the WordPress dashboard
Right now WP Rocket is making approximately 18 queries on each page load in the WordPress admin area.

Queries are related to the transients and return up to 1 record. Basically, they're pretty fast queries.

However, in an environment where the wp_options table is enormous due to the plugins that use it to store a lot of data instead of creating their own tables, the situation might change. The wp_options table could also grow due to the plugins that don't delete their records correctly when deactivating.

We could replace those queries with PHP checks that should be faster even if the wp_options table is not spammed with records.

Execution times of those queries should be checked on environments with the flooded wp_options table. 

Slack thread:
https://wp-media.slack.com/archives/C43T1AYMQ/p1553874846017700

remyperona:
## Description

This PR reduces the number of queries that WP Rocket performs on  the admin pages, mostly related to transients usage in notices.

It either:
- adds an additional check for the current page or the option value before trying to get the transient
- removes excessive calls to delete transient

It's also the opportunity to move the Sucuri subscriber to the new architecture, and do some clean up.

Fixes #1599 

## Type of change

The requests for the `pricing_data` and `user_data` have not been changed, as it's not possible to do so.

## How Has This Been Tested?

- [x] Checked queries with Query Monitor on admin pages

# Checklist:


iCaspar (approved):

Mai-Saad:
@Tabrisrp Appreciate checking Qs below:

1- Do we need these transients on pages other than our settings pages `_transient_timeout_rocket_clear_usedcss_response` `_transient_rocket_clear_usedcss_response` while RUCSS is disabled ? => currently those transients exists for all pages.

2- When RUCSS is enabled, queries raise from 43 to 143 on the settings page, while from 8/10 to 106/108 on other pages. Do we need to do anything about that?

3-` rocket_critical_css_generation_triggered` and `rocket_preload_triggered` => Added to any **non WPR page** once related option is enabled. I think this is expected here, isn't it?

4- `sucuri_purge_result` is added to **all pages** including our settings once the option is enabled, is this what's needed? (Note: Cloudflare adds 4 queries on all pages including WPR settings once enabled)

5- Do we expect any of those to affect query count: Preloadlinks, fonts, dns prefetch , site map?

6- OP-cache purge transients are there on the settings page also the option is removed now, Do we need to remove it?

Note:
- Overall #queries compared to trunk decreased on WPR page from 60 to 45  and decreased from 23/ 24 to 10/12 on other pages while WPR options are disabled

remyperona:
1. No, I pushed a fix for that
2. I don't think so
3. Yes
4. Yes
5. No
6. I merged the develop branch into this branch to make it up to date. For the transient itself, it will be cleaned-up automatically once it's expired.

Mai-Saad (approved):
Working as expected
Note:
on both branch and trunk, if we hide Sucuri API key (without define key in config file) then do any changes in settings => Sucuri validation error msg will be displayed as the key will be deleted (probably need separate GH issue)

PR #4677, opened 2022-01-21
Fixes #3617 Clean busting directory when purging cache

### LINKED ISSUE #3617: Busting directory is not cleared during the cache clearing

**Describe the bug**
We're not removing the `/wp-content/cache/busting/1/wp-content/` content when clearing the cache from the top menu neither during deactivation.

**To Reproduce**
1. Visit the page when the dynamic CSS file is added (dynamic_css.php, etc.)
2. Confirm that the associated file is created in the busting directory
3. Clear the cache, deactivate WP Rocket
4. Confirm that the cache is not cleared

**Expected behavior**
We should be able to clear this cache at the correct moment, including the `.gz` files which are not listed in the `$extensions` array here:
https://github.com/wp-media/wp-rocket/blob/8c02750b89412eb7e395551c150ab06e4a4b8e2b/inc/functions/files.php#L401

**Screenshots**
image

**Additional context**
We should think when the busting directory should be cleared exactly.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Add a call to `rocket_clean_cache_busting()` when purging the whole cache, to make sure it's updated correctly.

Fixes #3617 

## Type of change

No

# Checklist:

### DIFF
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -437,6 +437,7 @@ function do_admin_post_rocket_purge_cache() { // phpcs:ignore WordPress.NamingCo
 				if ( '' === $lang ) {
 					// Remove all minify cache files.
 					rocket_clean_minify();
+					rocket_clean_cache_busting();
 
 					// Generate a new random key for minify cache file.
 					$options                   = get_option( WP_ROCKET_SLUG );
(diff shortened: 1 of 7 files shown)


iCaspar (approved):

Mai-Saad (commented):
Currently, clearing cache from the admin bar or from the quick actions in the dashboard tab will clear the dynamic style /js file from busting. but the .gz file won't be removed. However, with the next page visit, the .gz file will be overwritten. @Tabrisrp Do you think this is fine or do we need to do anything here?

remyperona:
@Mai-Saad I updated the code to also clear the .gz files when the function is called.

@wp-media/php Need another code review, as I refactored the code of the function, and also added unit/integration tests for it

mostafa-hisham (approved):

Mai-Saad (approved):
All files are deleted now from busting.

PR #4684, opened 2022-01-25
Closes #4682 Update delay JS script

### LINKED ISSUE #4682: Update DelayJS script
Issue for update the Delay JS script to the newest version that can be fetched [here](https://github.com/wp-media/delay-javascript-loading/pull/16).

remyperona:
## Description

Update the delay JS script with the latest changes made

Fixes #4682

## Type of change

# Checklist:


iCaspar (approved):

Mai-Saad:
Note:
While CDN is disabled:
- if crossorigin="test" => will change to anonymous
- if crossorigin is different for the same domain, we will have 2 preconnect tags one for each URL and crossorigin will be the same
- The preconnect tag won't be added to the external URL when minify js is enabled (as it is internally hosted now)
- Crossorigin for relative Internal script won't be kept when preload script

While CDN is enabled:
- CDN preconnect is added twice, one by CDN at the top of the head and the other by delay js and bottom of header (for the internal scripts hosted by CDN)=> Do we need to change this?

Screenshot from 2022-02-22 12-49-07

Mai-Saad (approved):
Working as expected assuming nothing is needed for the previous comment.
Note: preload scripts while the page is hidden seems to be working here (for most of scripts)
[testrail-report-359.pdf](https://github.com/wp-media/wp-rocket/files/8127372/testrail-report-359.pdf)

andyz02:
It looks like after this change you have to double click links on mobile safari browsers.

PR #4702, opened 2022-02-01
Enhancement #4439 Remove OPCache purge

### LINKED ISSUE #4439: Get rid of "Clear OPcache" action button
This improvement intends to simply the usage of WP Rocket by getting rid of an unnecessary action button from the Quick Actions : Clear OPcache

image

An average user (which is the target audience of your plugin) doesn't even know what is OPcache. I think there is no added-value to have different buttons for :
- clear OPcache
- clear all cache files

I've been using WP Rocket for years now and I've never found its interest. I think the "OPcache" clearing may just simply be merged into "all cache files" clearing.

**Additional context**
It's always a loss of time to hit both buttons even if I believe that the "clear all cache files" should be sufficient to also clear the OPcache). An average user like me would still hit both buttons to ensure that everything is cleared.

Also, if you look at competitors like FlyingPress, this doesn't exist neither as there is no need for it.

image

remyperona:
## Description

Removes the OPCache purge feature (UI, capability and actions), and deprecate the related functions.

@camilamadronero I guess this will require a documentation update.

Fixes #4439 

## Type of change

There was a bit more to remove than initially groomed.

# Checklist:

### DIFF
--- a/inc/Engine/Capabilities/Manager.php
+++ b/inc/Engine/Capabilities/Manager.php
@@ -17,7 +17,6 @@ class Manager implements ActivationInterface, DeactivationInterface {
 		'rocket_purge_posts',
 		'rocket_purge_terms',
 		'rocket_purge_users',
-		'rocket_purge_opcache',
 		'rocket_purge_cloudflare_cache',
 		'rocket_purge_sucuri_cache',
 		'rocket_preload_cache',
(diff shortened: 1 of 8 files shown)


mostafa-hisham (approved):

Mai-Saad (approved):
Working as expected.
Note: Test scope is UI with update/fresh install (deprecated code is not included)
[testrail-report-350.pdf](https://github.com/wp-media/wp-rocket/files/8018285/testrail-report-350.pdf)

PR #4705, opened 2022-02-02
Preload mobile homepage after changing options that require a cache purge

remyperona:
## Description

In this PR, we add the preloading of the mobile homepage after changing options that require a cache purge. To do so, it adds a new filter `rocket_options_changed`, and a new callback method in the preload subscriber.

This only applies if the option for a separate cache for mobile is enabled.

## Type of change

## How Has This Been Tested?

- [x] Manual testing by changing options and checking the generation of the cache files
- [x] Unit tests added for the new method 

# Checklist:

### DIFF
--- a/inc/Engine/Preload/PreloadSubscriber.php
+++ b/inc/Engine/Preload/PreloadSubscriber.php
@@ -1,4 +1,5 @@
 <?php
+declare(strict_types=1);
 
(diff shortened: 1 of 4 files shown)


iCaspar (approved):

piotrbak (approved):
@Tabrisrp We're good here

PR #4719, opened 2022-02-09
Refactor kinsta.php to not rely on CDN class

### LINKED ISSUE #4717: Update Kinsta Compatibility
Kinsta no longer use KeyCDN and they have updated their MU Plugin to remove the `\Kinsta\CDN_Enabler` class. 
That is causing the following notice to always show for users due to the logic in this file https://github.com/wp-media/wp-rocket/blob/develop/inc/3rd-party/hosting/kinsta.php :

> **Your installation seems to be missing core Kinsta files managing Cache clearing and CDN, which will prevent your Kinsta installation and WP Rocket from working correctly. Please get in touch with Kinsta support through your MyKinsta account to resolve this issue.**

This could also causes WP Rocket to no longer clear the Kinsta cache.

**Additional context**
HS Ticket from Kinsta support: https://secure.helpscout.net/conversation/1781767741/324931/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

alexstine:
## Description

Fixes #4717 

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,10 @@
 		"preferred-install": {
 			"wp-media/phpunit": "source"
 		},
-		"process-timeout": 0
+		"process-timeout": 0,
+		"allow-plugins": {
+			"composer/installers": true
+		}
 	},
 	"support": {
 		"issues": "https://github.com/wp-media/wp-rocket/issues",
--- a/inc/3rd-party/hosting/kinsta.php
+++ b/inc/3rd-party/hosting/kinsta.php
@@ -136,7 +136,7 @@ if ( isset( $kinsta_cache ) ) {
 					'status'      => 'error',
 					'dismissible' => '',
 					// translators: %1$s = opening link tag, %2$s = closing link tag.
-					'message'     => sprintf( __( 'Your installation seems to be missing core Kinsta files managing Cache clearing and CDN, which will prevent your Kinsta installation and WP Rocket from working correctly. Please get in touch with Kinsta support through your %1$sMyKinsta%2$s account to resolve this issue.', 'rocket' ), '<a href="https://my.kinsta.com/login/" target="_blank">', '</a>' ),
+					'message'     => sprintf( __( 'Your installation seems to be missing core Kinsta files managing Cache clearing, which will prevent your Kinsta installation and WP Rocket from working correctly. Please get in touch with Kinsta support through your %1$sMyKinsta%2$s account to resolve this issue.', 'rocket' ), '<a href="https://my.kinsta.com/login/" target="_blank">', '</a>' ),
 				]
 			);
 		}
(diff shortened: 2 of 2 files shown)


remyperona:
How does the Kinsta CDN URL works now? With this change, we are not automatically adding it in our config anymore, so it's not ideal.

alexstine:
@Tabrisrp Kinsta now uses Cloudflare and everything works via proxy. There is no need for CDN URL rewriting or anything like that anymore so should be no need to store the URL.

Thanks.

remyperona:
@alexstine Thank you, what about the caching part of the MU plugin, is this still being used? We are doing call to `kinsta-clear-cache` endpoint to clear the cache of specific URLs, is it still valid?

alexstine:
@Tabrisrp Yes, it's still valid. I left test site credentials in the linked Issue if you want to give it a test.

remyperona:
ok great, one thing I would suggest for this PR is to update the notice message to also remove the part about the CDN then, and that should be good to go.

alexstine:
@Tabrisrp Done.

remyperona (approved):

Mai-Saad:
@alexstine Can you please provide us with FTP access? Thank you

alexstine:
@Mai-Saad That really wouldn't be possible at this point. Do you need me to check error logs?

If it's really needed, happy to send via email.

Thanks.

Mai-Saad:
> @Mai-Saad That really wouldn't be possible at this point. Do you need me to check error logs?

@alexstine Thanks for the reply. Can you please remove the last test line in functions.php of the active theme?

alexstine:
@Mai-Saad Fixed.

Mai-Saad:
Can see the main problem fixed but with the following notes:
1- If we hard refresh the page multiple times i.e 3 before we clear cache from admin bar => we will need to hard refresh n times so Kinsta Cache be Miss ( same case if we clear cache from Kinsta itself, while if we didn't refresh the page multiple times then the clear cache will reflect with the 1st refresh of the page)
2- Always Purge URL is not working probably i.e if we added `/test/ `or `/test `in Always Purge URL, cache /test page then we edited any post, /test will still be Hit after refresh => /test shall be Miss (as per discussion with Piotr, this is not related to PR and will need separate GH issue, couldn't check trunk yet as the folder is R access) 
[testrail-report-357.pdf](https://github.com/wp-media/wp-rocket/files/8074530/testrail-report-357.pdf)

Mai-Saad:
@alexstine Can you please allow the delete/upload of the WPR plugin from the dashboard? (we will need this for regression test after merging the PR to ensure that everything else is working fine)

alexstine:
@Mai-Saad Fixed, plugin can now be deleted.

PR #4729, opened 2022-02-10
Add preload links for fonts found in the used CSS

remyperona:
## Description

In this PR, we use the used CSS content to check if there is font-face declaration in it, and add preload links for each font.

Considering the case of a font having multiple font type declarations, only the first one is preloaded.

## Type of change

## How Has This Been Tested?

- [x] Added a new case fixture to existing tests

# Checklist:


mostafa-hisham (approved):

piotrbak:
@Tabrisrp Can we exclude `EOT` format from being preloaded?

remyperona:
@piotrbak I pushed a change to skip `.eot` fonts

piotrbak:
@Tabrisrp There are 2 things more:
1. Can we make sure that if `EOT` is there, we'll preload the next font format (if exists)? Right now, if `EOT` is there, we'll bypass thsi font completely.
2. We'll also need a filter to disable this optimization

🙏

GeekPress:
> We'll also need a filter to disable this optimization

Must be a decision, not option here, isn't it?

For which reason should we allow to bypass our choice?

piotrbak:
@Tabrisrp If there are two the same `@font-face` definitions, we'll add 2 preload links related to the same font.

It doesn't seem to cause side effects, but maybe it's easily fixable?

remyperona:
It is, I updated the code to remove duplicate entries in the fonts array

piotrbak (approved):

PR #4732, opened 2022-02-15
Prevent PHP 8.1 deprecation notices

### LINKED ISSUE #4728: Deprecation notice when using PHP 8.1
- Made sure you’re on the latest version ☑️
- Used the search feature to ensure that the bug hasn’t been reported before ☑️

**Describe the bug**
A customer informed this deprecation notice is displayed when running PHP 8.1:

`Deprecated: Constant FILTER_SANITIZE_STRING is deprecated in /var/www/vhosts/f****g.com/httpdocs/wp-content/plugins/wp-rocket/inc/admin/upgrader.php on line 41`

**To Reproduce**
N/A

**Expected behavior**
There should  be no deprecation notice displayed when running PHP 8.1

**Screenshots**
N/A

**Additional context**
The customer took this from the debug file from wp-content.

Related ticket: [https://secure.helpscout.net/conversation/1783564514/325265](https://secure.helpscout.net/conversation/1783564514/325265)

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

### LINKED ISSUE #4842: PHP 8.1 Deprecation notice at yoast-seo.php: "Automatic conversion of false to array is deprecated"

**Describe the bug**

When using PHP 8.1, we'll get a Deprecation notice:

`Automatic conversion of false to array is deprecated wp-content/plugins/wp-rocket/inc/3rd-party/plugins/seo/yoast-seo.php:10`

it is happening here: 

https://github.com/wp-media/wp-rocket/blob/9d3313b2f974831d946b10bad7c475cd8c023831/inc/3rd-party/plugins/seo/yoast-seo.php#L10

**To Reproduce**
1. Update to PHP 8.1
2. Enable Yoast SEO and activate sitemap preload
3. See the deprecation notice

**Expected behavior**
We should not cause any deprecation notices

**Screenshots**
2022-03-22-at-14-51-10

**Additional context**
Ticket: https://secure.helpscout.net/conversation/1816071909/331622?folderId=2683093

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Replace the usage of `FILTER_SANITIZE_STRING` (deprecated in PHP 8.1) by using the WP function `sanitize_key()` instead.

Update the Yoast SEO compatibility file to prevent deprecation notice `Automatic conversion of false to array is deprecated`. Move the code to the new architecture.

Fixes #4728
Fixes #4842 

## Type of change

No

# Checklist:


engahmeds3ed (approved):

inc/ThirdParty/Plugins/SEO/Yoast.php
```diff
@@ -0,0 +1,171 @@
...
+			return $enabled;
+		}
+
+		if ( ! rocket_has_constant( 'WPSEO_VERSION' ) ) {
+			$enabled = false;
+
+			return $enabled;
+		}
+
+		$yoast_seo_xml = get_option( 'wpseo_xml', [] ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
```
engahmeds3ed, line 159: What do u think if we used here their options class instead? they handle there multisite in a good way:
https://github.com/Yoast/wordpress-seo/blob/trunk/inc/options/class-wpseo-options.php#L242-L263

engahmeds3ed (changes requested):

inc/ThirdParty/Plugins/SEO/Yoast.php
remyperona: A couple of reasons I prefer the current approach:
- less dependency on 3rd party code, if they ever change their code we won't have to change ours, since we use core functions
- using will re-introduce the deprecation notice on 8.1, because there is no guarantee it returns an array (even if the typehint says so), and we need this to be an array.

engahmeds3ed (approved):

Mai-Saad:
@Tabrisrp the main error is fixed but can see an error with an update using the following steps in query monitor (it will be removed by clearing cache)
1- PHP 8.1.4, Yoast SEO is enabled with sitemap
2- Install and activate wpr 3.10.9 or 3.11.0.3 and refresh the plugins page while the query monitor is open => some errors exist for WPR
3- Deactivate wpr plugin, switch to fix branch and `composer dumpautoload` then activate wpr and check query monitor

Screenshot from 2022-04-12 23-10-11
Do we need to do anything about this?

remyperona:
@Mai-Saad Can you click on the ➕ icon to get the details of the location?

Right now I don't know which WPR code it's coming from

Mai-Saad:
@Tabrisrp HYG
Screenshot from 2022-04-13 16-05-04

remyperona:
@Mai-Saad Thank you, the issue is coming from WP core itself, not from WP Rocket, so it's not something we can fix. (see https://github.com/WordPress/Requests/pull/500)

Mai-Saad (approved):
Working as expected

PR #4733, opened 2022-02-15
Fixes #4701 Prevent console error when using Elementor with combine JS and user cache

### LINKED ISSUE #4701: Console Errors for logged-in user when Combine js is enabled while Elementor plugin is active
- Made sure you’re on the latest version => 3.10.7

**Describe the bug**
Console error displayed for the logged-in user when combine js and user cache are enabled while Elementor is active

**To Reproduce**
- Elementor is enabled 
- User cache and Combine JS are enabled at WPR
1. open home page while logged-in
2. Console errors are there

**Expected behavior**
No console error

**Screenshots**
Screenshot from 2022-02-01 17-52-37

**Additional context**
- Deactivate Elementor or disable Combine JS or disable User Cache will fix the problem
- Tested on 3.10.7 and 3.8 => same results

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Add a new method to exclude the `hooks.min.js` file from combine JS when using:
- Elementor
- Combine JS
- User cache and the user is logged in

Fixes #4701 

## Type of change

No

## How Has This Been Tested?

- [x] Added unit/integration tests for the new method

# Checklist:

### DIFF
--- a/inc/ThirdParty/Plugins/PageBuilder/Elementor.php
+++ b/inc/ThirdParty/Plugins/PageBuilder/Elementor.php
@@ -165,7 +165,7 @@ class Elementor implements Subscriber_Interface {
 			return $excluded_files;
 		}
 
-		$excluded_files[] = '/wp-includes/js/dist/hooks.min.js';
+		$excluded_files[] = '/wp-includes/js/dist/hooks(.min)?.js';
 
 		return $excluded_files;
 	}
--- a/inc/ThirdParty/Plugins/PageBuilder/Elementor.php
+++ b/inc/ThirdParty/Plugins/PageBuilder/Elementor.php
@@ -1,16 +1,14 @@
 <?php
+declare(strict_types=1);
+
 namespace WP_Rocket\ThirdParty\Plugins\PageBuilder;
 
 use WP_Rocket\Admin\Options_Data;
-use WP_Rocket\Event_Management\Subscriber_Interface;
 use WP_Rocket\Engine\Optimization\DelayJS\HTML;
-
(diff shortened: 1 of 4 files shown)


inc/ThirdParty/Plugins/PageBuilder/Elementor.php
```diff
@@ -161,4 +142,31 @@ public function add_fix_animation_script( $html ) {
...
+
+		if ( ! $this->options->get( 'cache_logged_user', false ) ) {
+			return $excluded_files;
+		}
+
+		if ( ! is_user_logged_in() ) {
+			return $excluded_files;
+		}
+
+		$excluded_files[] = '/wp-includes/js/dist/hooks.min.js';
```
engahmeds3ed: Do we need to handle both `min` and source files? as if `SCRIPT_DEBUG` is true, the page will load the source file not the minified one.
Like what we do here: https://github.com/wp-media/wp-rocket/blob/f7667cd723f29484c222d49c274fda6a11abffd5/inc/Engine/Optimization/DeferJS/DeferJS.php#L231

engahmeds3ed (changes requested):

inc/ThirdParty/Plugins/PageBuilder/Elementor.php
remyperona: Indeed it's better to cover both, code is updated

engahmeds3ed (approved):

Mai-Saad (approved):
Working as expected

PR #4737, opened 2022-02-16
Fixes #4706 Exclude use.typekit.com from minify & Defer JS

### LINKED ISSUE #4706: Defer and minify JS compatibility with Typekit

**Describe the bug**
When using the Typekit fonts with Defer JS, fonts are not rendered properly on Safari mobile. But they work fine on Desktop on Safari, and on both Desktop and Mobile of Android devices. 

Sample of the script added by Typekit - https://use.typekit.com/sgs4pbn.js (the file name is a random string)

A separate cache for mobile devices was enabled to exclude that possibility. 

Only excluding `use.typekit.com` fixes the issue. 

The same should be excluded from JS Minify to avoid the exclusion to stop working when the script is hosted. 

**To Reproduce**
1. Use Typekit to add a font
2. Enable Defer JS
3. See error (font-family not applied to target selector)

**Expected behavior**
Typekit fonts should work out of the box.

**Screenshots**
N/A

**Additional context**
Ticket - https://secure.helpscout.net/conversation/1770086284/322313/
Related to - https://github.com/wp-media/wp-rocket/issues/1912 - the only difference is that JupiterX has its own Typekit integration through the API using a hosted file.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

DahmaniAdame:
https://secure.helpscout.net/conversation/1770086284/322313/

### DIFF
--- a/inc/Engine/Optimization/DeferJS/DeferJS.php
+++ b/inc/Engine/Optimization/DeferJS/DeferJS.php
@@ -233,6 +233,7 @@ class DeferJS {
 			'js-eu1.hsforms.net',
 			'yanovis.Voucher.js',
 			'/carousel-upsells-and-related-product-for-woocommerce/assets/js/glide.min.js',
+			'use.typekit.com',
 		];
 
 		$exclude_defer_js = array_unique( array_merge( $exclude_defer_js, $this->options->get( 'exclude_defer_js', [] ) ) );
--- a/inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
+++ b/inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
@@ -256,6 +256,7 @@ abstract class AbstractJSOptimization extends AbstractOptimization {
 			'js-eu1.hsforms.net',
 			'statcounter.com/counter/counter.js',
 			'snapppt.com',
+			'use.typekit.com',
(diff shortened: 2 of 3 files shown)


remyperona (approved):

Mai-Saad:
@Tabrisrp @DahmaniAdame In the test template we have 
```
<script type='text/javascript' src='//use.typekit.com/sgs4pbn.js'></script>
<script type='text/javascript'>try{Typekit.load();}catch(e){}</script>
```
1st script is excluded from Combine JS while the 2nd script is combined which may cause FOUC, As per discussion with Piotr, we think that this shall be excluded from Combine JS as well `try{Typekit.load` . What do you think?

remyperona:
I agree

DahmaniAdame:
@Mai-Saad sounds good 👍

Mai-Saad:
@Tabrisrp Seems working fine but the inline script now moves after the combined file, is this what we need or just exclude without changing location in the source (as same as if excluded from UI)?

remyperona:
@Mai-Saad That was a mistake on my part, I moved it to the correct exclusion so that it doesn't move.

Mai-Saad (approved):
Working as expected

PR #4778, opened 2022-03-02
Add RUCSS notices for processing & success

remyperona:
## Description

Add admin notices on WP Rocket settings page when enabling RUCSS or when clearing the used CSS.

## Type of change

## How Has This Been Tested?

- [x] Enabling RUCSS in the options
- [x] Clearing the used CSS from the action button

# Checklist:


piotrbak:
@Tabrisrp Could you merge new approach into this PR?

remyperona:
Done

piotrbak:
@Tabrisrp We need to handle better the update from previous version, right now:
1. RUCSS is enabled in previous version
2. We update the plugin
3. The success notice is displayed

I think we have two options here:
1. During the update clear the Used CSS + Cache, make a request to the homepage and display notice with counter.
2. Don't clear the Used CSS + cache and don't display the notice at all
What do you think?

The `seconds` and `batch size` values should be the ones that are coming from the actual batch + interval values, is this the case now? 🤔

remyperona:
> The seconds and batch size values should be the ones that are coming from the actual batch + interval values, is this the case now?

Yes

For the notice, currently I don't have anything to prevent its display except when the processing transient is set. We could clear the used CSS and cache, knowing that the new system is going to be better too, so it can be a good idea?

piotrbak:
@Tabrisrp Yes, I think so. Will we be able to preload the homepage too?

piotrbak:
@Tabrisrp The batch size + interval values are coming from those filters, right? 
https://github.com/wp-media/wp-rocket/blob/be15a9d516f543713449095042c895baf06afb51/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L486
https://github.com/wp-media/wp-rocket/blob/be15a9d516f543713449095042c895baf06afb51/inc/Engine/Optimization/RUCSS/Admin/Subscriber.php#L175

remyperona:
@piotrbak I added the clear of cache and used CSS + homepage preload

The filters are correct yes

piotrbak:
@Tabrisrp I think we're ready to go here.

PR #4790, opened 2022-03-08
Guard the backslashes before passing the css content to preg_replace

engahmeds3ed:
## Description

This PR solves the issue of having backslashes in the CSS content like the following CSS
```
#test:before {content: '\2605';}
```

when passing this content to this line: 
https://github.com/wp-media/wp-rocket/blob/6225cafed859756a4cb7817de7cb06bbed3b14b3/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L372

it adds the following CSS into the page:
```
#test:before {content: '05';}
```

and this will affect all icon fonts like fontawesome and others.

## Type of change

Not groomed, just talk about it with @Tabrisrp and @mostafa-hisham .

## How Has This Been Tested?

As described above, use mentioned CSS into the page and check the generated used CSS.

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -450,7 +450,8 @@ class UsedCSS {
 	 * @return string
 	 */
 	private function get_used_css_markup( UsedCSS_Row $used_css ): string {
-		$used_css_contents = $this->handle_charsets( $used_css->css, false );
+		$css               = str_replace( '\\', '\\\\', $used_css->css );// Guard the backslashes before passing the content to preg_replace.
+		$used_css_contents = $this->handle_charsets( $css, false );
 		return sprintf(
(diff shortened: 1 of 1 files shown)


mostafa-hisham (approved):

piotrbak:
@engahmeds3ed is it affecting only specific content values, it seems. What kind of side effects can we expect with this change? 🤔

engahmeds3ed:
> @engahmeds3ed is it affecting only specific content values, it seems. What kind of side effects can we expect with this change? thinking

I don't think that there is any place inside CSS that can include this backslash, other than content property, so I think it won't affect any other thing here.

piotrbak:
@engahmeds3ed We have for example this:
```
@font-face {
    font-family: Font Awesome\ 5 Free;
    font-style: normal;
    font-weight: 900;
    font-display: block;
```

But it seems that it works correctly with this update.

Mai-Saad (approved):
Working fine on related templates/examples

PR #4794, opened 2022-03-08
Preserve Gutenberg inline classes when using remove unused CSS

remyperona:
## Description

Gutenberg uses some dynamically named classes in inline styles, so we need to preserve those in the HTML when adding the used CSS.

Fixes https://github.com/wp-media/nodejs-treeshaker/issues/28

## Type of change

# Checklist:


Mai-Saad:
Using 2022 on  https://new.rocketlabsqa.ovh/, found the following: @Tabrisrp @piotrbak what do you think?
1- `class="wp-block-site-title"` element is removed from home page => probably needs to exclude `.wp-elements-` too (**note:** when tried this exclusion from UI safelist, it didn't work and not preserved in source.. but adding it in code exclusions works => shouldn't safelist work for that?)
2- Style for post links is different than nowprocket => probably need another kind of exclusion

https://user-images.githubusercontent.com/76941962/158469222-43328ebb-4ceb-4126-b454-c6d4489e4e41.mp4

piotrbak:
@Tabrisrp We also need to add the `.wp-elements-`, thanks @Mai-Saad 

When it comes to the styles, I'm still not sure what should be the safelist.

Mai-Saad:
1- fixed now
2- seems it's a pattern as the same problem on the following themes 2020, 2022, 2015, customizr (not sure which other themes could be affected). I think we can open a separate ticket for this. what do you think @piotrbak ?

engahmeds3ed (approved):

piotrbak:
@Mai-Saad Yes, we need a separate issue and check if that's a regression.

PR #4799, opened 2022-03-09
Fix clearing usedcss table in random cases

engahmeds3ed:
## Description

In some cases, usedcss rows are deleted just after creation in random basis, after investigation, I found that the problem is that we don't initialize `last_accessed` column whenb creating a new row there so once the CRON that is responsible for deleting not accessed rows from DB fires, it deletes all rows as this query returns all rows in this case:

https://github.com/wp-media/wp-rocket/blob/b1d16b7bd8ddfb0a998330fea09cd3fd687fb92f/inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php#L84

Fixes https://github.com/wp-media/nodejs-treeshaker/issues/26

## Type of change

Yes

## How Has This Been Tested?

1. Enable Preload cache.
2. Enable RUCSS.
3. Wait till used CSS is completed.
4. Refresh any admin page.
5. All rows in the used CSS table shouldn't be deleted.

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Database/Queries/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Database/Queries/UsedCSS.php
@@ -139,6 +139,7 @@ class UsedCSS extends Query {
 			'queue_name' => $queue_name,
 			'status'     => 'pending',
 			'retries'    => 0,
+			'last_accessed' => current_time( 'mysql', true ),
 		];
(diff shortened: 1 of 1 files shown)


remyperona (approved):

Mai-Saad (approved):
This is working fine now
clear cache not clearing used CSS from both admin bar and quick action

PR #4800, opened 2022-03-09
Preloads the homepage after cleaning the used CSS

remyperona:
## Description

Preloads the homepage (desktop & mobile version if enabled) after cleaning the used CSS in the database.

Fixes https://github.com/wp-media/nodejs-treeshaker/issues/27

## Type of change

# Checklist:


inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
```diff
@@ -356,6 +370,13 @@ public function truncate_used_css_handler() {
 
 		$this->set_notice_transient();
 
+		/**
+		 * Fires after the used CSS has been cleaned in the database
+		 *
+		 * @since 3.11
+		 */
+		do_action( 'rocket_after_clean_used_css' );
+
```
engahmeds3ed: I like the idea of making a new action to fire here and not adding the preload call here.
I can see that this action is defined twice so what do u think if we created a new private function here to truncate the table or remove completed rows based on the condition above and after this we can add the action like the following
```
private function delete_rows(){
    if ( 0 < $this->used_css->get_not_completed_count() ) {
        $this->used_css->remove_all_completed_rows();
        return
    }
}
```

@Tabrisrp what do u think?

engahmeds3ed (changes requested):

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
remyperona: Good idea I refactored the code to a private function 👍🏼

engahmeds3ed (approved):

Mai-Saad:
Checks (start after clear cache):
While preloading disabled/enabled, Separate mobile cache enabled/disabled
- Enable RUCSS => created used CSS for home page with success msg (if we enabled RUCSS after that, the completed entry in DB won't change except for last accessed which is expected) same if preload enabled
- Edit safelist => recreated home at used CSS but usually stuck at pending till action on DB (shall be handled in corn PR) 
- Import settings => if it's not changing RUCSS (i.e change safelist, enable it while disabled) then nothing is affected in used CSS, o.w will recreate the home page in used CSS as expected except for the case in the note below
- Clear used CSS => entry recreated to the home page usually with pending status and rarely with in progress or completed (probably status will be handled in corn PR)

**Note**: In the following scenario, only the desktop version of the home page will be added to the used CSS table. shouldn't we have entry too for mobile in this case? do we need another GH for this case? @Tabrisrp (if Separate mobile cache was enabled before import settings, then mobile entry will be normally added to used CSS of home page)
The scenario:
1- Separate mobile cache, preload are disabled and RUCSS is enabled with no safelist or disabled
2- Import settings where the separate mobile cache is enabled and safelist is added to RUCSS
3- Check used CSS table

remyperona:
@Mai-Saad Can you retry the last scenario with the latest code changes?

Mai-Saad:
Well now it's adding an entry for both mobile and desktop if the database was empty or has a pending home page but if it was completed then import (which changes the safelist) will clear used CSS and no auto-generation for the home page. I think it should regenerate used CSS in this case too. What do you think? @Tabrisrp.
Note: change safelist from UI directly in the completed state of the home page will delete used home page CSS too and not regenerate used CSS for home

remyperona:
I think there was an extra cache clearing that was interfering with the process, can you retry with the new changes?

Mai-Saad:
Working fine now. Thank you :+1:

PR #4811, opened 2022-03-11
Trigger cron when switching between RUCSS notices

remyperona:
## Description

Send a request to the cron when switching between RUCSS notices, to make sure the scheduled actions are running even if there is no activity on the website.

Fixes https://github.com/wp-media/nodejs-treeshaker/issues/29

## Type of change

## How Has This Been Tested?

- [x] Updated tests to validate the call

# Checklist:


inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
```diff
@@ -581,5 +581,13 @@ private function set_notice_transient() {
...
 			MINUTE_IN_SECONDS
 		);
+
+		wp_safe_remote_get(
+			site_url( '/wp-cron.php' ),
+			[
+				'blocking' => false,
+				'timeout'  => 0.01,
+			]
+		);
```
engahmeds3ed: What about using this function instead?
https://developer.wordpress.org/reference/functions/spawn_cron/

it's doing the same functionality with additional checks/validations to make sure that we will not affect the performance.

engahmeds3ed (changes requested):

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
remyperona: Indeed it's much better, I didn't know about that one 👍🏼

engahmeds3ed (approved):

Mai-Saad:
Checks:
- enable RUCSS while preload enabled/disabled=> both cases, the home page was still pending in used CSS table till user interaction, shouldn't it be completed with no need for user interaction on the dashboard? @Tabrisrp 
- clear used CSS while preload enabled /disabled => nothing added in db if preload disabled (will be handled in other PR), if preload was enabled, no entry added in the table for home unless if we preload cache (which is expected)
- edit safelist while preload enabled/disabled => while preload is enabled, entries are recreated but stuck in pending till DB interaction.. if preload is disabled, entry readded to home but stuck at pending (same as 1st point)

remyperona:
I changed the approach to send the cron request at the moment we switch notices, it's ready to be tested again

PR #4817, opened 2022-03-17
Fixed problems related to an outdated fixture file for RocketCDN CTA

CrochetFeve0251:
## Description

Since 16/03/2022 tests broke because of what seems like an actualization from the HTML code from the CTA for RocketCDN.
This PR update the fixture in question () to make it match the new HTML structure.

Fixes #(issue number)
None
## Type of change

There was no grooming for this issue as it is a problem directly in the tests.

## How Has This Been Tested?

- [x] Running automatic tests

# Checklist:


CrochetFeve0251:
@Tabrisrp The problem looks only appearing on my local environnement.
I tried to pass my PHP version to 7.1 and WP to 5.4.7 to exactly match the test environnement but it still giving same results as before.

I am continuing to invest the problem, tell me if you have any idea

CrochetFeve0251:
I installed wprocket inside a WP instead of using the one from bin/install-wp-tests.sh and it looks like it fixed the problem

remyperona:
I think the issue is because we updated the RocketCDN code on WP Rocket repository, but we didn't synchronize the changes in the https://github.com/wp-media/module-rocketcdn, which is where the code is pulled from when doing a new installation for CI for example.

I have a PR ready there, but it's containing additional changes, so it has not been merged yet.

PR #4821, opened 2022-03-17
Clear Used CSS rows with changing the CDN option

engahmeds3ed:
## Description

Check this issue: https://github.com/wp-media/nodejs-treeshaker/issues/35

What we do now is to apply CDN into our used CSS just before saving it into the DB so each time we need to change CDN we will need to remove used CSS row.

<strike>So I took another approach which will not require removing the used CSS by applying the filter `rocket_usedcss_content` when getting the Used CSS from the DB instead of when saving it to give the flexibility to enable/disable CDN option without the need to remove used CSS rows.</strike>

We need to clear Used CSS by removing completed/failed rows when we touch the CDN option (CDN/CNames/Zones).

## Type of change

Not groomed

## How Has This Been Tested?

- [x] Test enable/disable CDN manually when RUCSS is enabled => should clear Used CSS rows
- [x] Test change CDN CNames manually when RUCSS is enabled => should clear Used CSS rows
- [x] Test change CDN Zones manually when RUCSS is enabled => should clear Used CSS rows
- [x] Try the above tests again when RUCSS is disabled => should NOT clear Used CSS rows
- [x] Automated tests

# Checklist:


engahmeds3ed:
The problem here is that we apply the CDN into the page that we use to generate the used CSS for it so we still need to regenerate the used CSS if we made any change in the CDN option.

mostafa-hisham (approved):

remyperona:
@engahmeds3ed In your unit test and fixture, you don't need the filesystem, that can simplify the code there a bit more

engahmeds3ed:
> @engahmeds3ed In your unit test and fixture, you don't need the filesystem, that can simplify the code there a bit more

Good point, Done

Mai-Saad:
checks:
- Enable CDN, change mode, change CNAME, add CDN => regenerate used CSS for home ( for all with preload)
- All CDN options
- CDN exclusion

Mai-Saad:
@engahmeds3ed please find notes below based on the exploratory test:
1- If CDN was applied to CSS or CSS&JS => some image URL inside used CSS will be re-written to CNAME. do we need separate GH for this?
2- CDN exclusions aren't reflected on used CSS => can we handle this in current PR?
3- Clear used CSS while CDN enabled, preload disabled, and previous status-completed, will regenerate only mobile version of used CSS for home. can you please recheck if we need to do anything here?

engahmeds3ed:
> 1- If CDN was applied to CSS or CSS&JS => some image URL inside used CSS will be re-written to CNAME. do we need separate GH for this?

Yes plz

> 2- CDN exclusions aren't reflected on used CSS => can we handle this in current PR?

You can add it to the same issue with first point above.

> 3- Clear used CSS while CDN enabled, preload disabled, and previous status-completed, will regenerate only mobile version of used CSS for home. can you please recheck if we need to do anything here?

this is related to the `blocking` thing here: https://github.com/wp-media/wp-rocket/blob/c93c47e1ae9d3fe3d76b45778048559ffd2b13b8/inc/Engine/Preload/PreloadSubscriber.php#L431-L453

@Mai-Saad when u come back, I will check that last point with u but we can't do anything here in this PR.

Mai-Saad (approved):
Since nothing is needed on PR.
1- probably won't be a problem later (CDN will apply for all)
2- https://github.com/wp-media/nodejs-treeshaker/issues/46
3- Shall be considered with new preload

PR #4826, opened 2022-03-17
Add flatsome compatibility for RUCSS inline CSS preserve

remyperona:
## Description

Add a compatibility file for Flatsome, to add specific patterns to the inline CSS content to preserve when using remove unused CSS

Fixes https://github.com/wp-media/nodejs-treeshaker/issues/39

## Type of change

- [x] Third-Party compatibility

# Checklist:

### DIFF
--- a/inc/ThirdParty/Themes/Flatsome.php
+++ b/inc/ThirdParty/Themes/Flatsome.php
@@ -51,6 +51,7 @@ class Flatsome implements Subscriber_Interface {
 	private static function is_flatsome( $theme = null ) {
 		$theme = $theme instanceof \WP_Theme ? $theme : wp_get_theme();
 
-		return 'flatsome' === ( strtolower( $theme->get( 'Name' ) ) || strtolower( $theme->get_template() ) );
+		return 'flatsome' ===  strtolower( $theme->get( 'Name' ) )
+                || 'flatsome' ===  strtolower( $theme->get_template() );
 	}
 }
--- a/inc/ThirdParty/Themes/Flatsome.php
+++ b/inc/ThirdParty/Themes/Flatsome.php
@@ -1,0 +1,56 @@
+<?php
+declare(strict_types=1);
+
+namespace WP_Rocket\ThirdParty\Themes;
+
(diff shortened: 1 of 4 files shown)


mostafa-hisham (approved):

Mai-Saad:
@Tabrisrp currently there is a missing section, styles are different, and the grey background for contact is not there. can you please check?

inc/ThirdParty/Themes/Flatsome.php
```diff
@@ -0,0 +1,56 @@
...
+	 * Checks if the current theme (or parent) is Flatsome
+	 *
+	 * @since 3.11
+	 *
+	 * @param WP_Theme $theme Instance of the theme.
+	 */
+	private static function is_flatsome( $theme = null ) {
+		$theme = $theme instanceof \WP_Theme ? $theme : wp_get_theme();
+
+		return 'flatsome' === ( strtolower( $theme->get( 'Name' ) ) || strtolower( $theme->get_template() ) );
```
mostafa-hisham: the problem was in condition
```suggestion
		return 'flatsome' === ( strtolower( $theme->get( 'Name' ) ) ?: strtolower( $theme->get_template() ) );
```
OR
```suggestion
		return 'flatsome' ===  strtolower( $theme->get( 'Name' ) )
                || 'flatsome' ===  strtolower( $theme->get_template() );
```

mostafa-hisham (changes requested):

mostafa-hisham (approved):

Mai-Saad:
The missing section is added now, the styles are the same and grey background is added for contact.

The only difference now compared to ?nowprocket is the spacing above the contact section. Can we fix this point?

engahmeds3ed:
this spacing is happening from the following `gap-timestamp` id as follows:-

?nowprocket version
image

The cached version
image

So we need to add `#gap-` into the list of inline preserve list.

Mai-Saad (approved):
Working as expected on both Flatsome and ChildFlatsome

PR #4810, opened 2022-03-11
Fixes #4765 Prevent query string caching from overridding page exclusion

alfonso100:
## Description

Per grooming: Instead of using `get_clean_request_uri()` as the subject in which to match, change it to use `get_request_uri_base()`, which doesn't contain the query string.

Fixes #4765

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## Is the solution different from the one proposed during the grooming?

No, exactly as groomed 
https://github.com/wp-media/wp-rocket/issues/4765#issuecomment-1060886331

## How Has This Been Tested?

On my lab site.

# Checklist:

Please delete the options that are not relevant.

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings

PR #4834, opened 2022-03-21
Changed the function call to find the uri in the method can_process_uri

### LINKED ISSUE #4765: Query string caching overrides excluded pages from cache
- [x] Made sure you’re on the latest version `3.10.8`

**Describe the bug**
When a query string is set to be cached, it will be cached on pages excluded from caching as well. 

**To Reproduce**
1. Enable a specific query string to be cache
2. Exclude a specific page from caching
3. See that the page is cached when using the query string

**Expected behavior**
Page exclusions should be applied to query strings cache as well. 

**Screenshots**
N/A 

**Additional context**
N/A

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

CrochetFeve0251:
## Description

Fix a problem with caching a page when a page is excluded but it also has a query parameter that is forcing page to be cached.
Now the behavior is the expected one: the page should be cached.

For this the function for getting the URI has been changed in method can_process_uri from the Tests class.

Fixes #4765

## Type of change

No the grooming was right

## How Has This Been Tested?

- [ ] Tested on local environment

# Checklist:

### DIFF
--- a/inc/classes/Buffer/class-tests.php
+++ b/inc/classes/Buffer/class-tests.php
@@ -577,7 +577,7 @@ class Tests {
 			return self::memoize( __FUNCTION__, [], true );
 		}
 
-		$can = ! preg_match( '#^(' . $uri_pattern . ')$#i', $this->get_clean_request_uri() );
+		$can = ! preg_match( '#^(' . $uri_pattern . ')$#i', $this->get_request_uri_base() );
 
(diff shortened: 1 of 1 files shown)


remyperona:
We already have PR #4810 for this issue

PR #4857, opened 2022-03-25
Allow people to use v2 of composer/installers package

Mahjouba91:
## Description

Otherwise you cannot install WP Rocket if you're currently using v2 like this:

`"composer/installers": "^2.0",`

Fixes #(issue number)

## Type of change

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
 	],
 	"require": {
 		"php": ">=7.0",
-		"composer/installers": "~1.0",
+		"composer/installers": ">=1.0",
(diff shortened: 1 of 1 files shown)


Mahjouba91:
Or actually you can do something like this maybe:

```
"composer/installers": "^1.0 || ^2.0",
```

remyperona:
@Mahjouba91 👋🏼 , thank you for the PR, I created a new one with an additional change to make sure there will be no PHP conflict, so I'm going to close this one.

Duplicate of #4863

PR #4886, opened 2022-04-01
Fixes #4876 Update RUCSS safelist on update to 3.11.0.2

### LINKED ISSUE #4876: Add prefix to safelisted items during the update

**Describe the bug**
In the previous approach of RUCSS safelist worked a bit differently. Support Team was using `.class` to safelist the following CSS: `.parent .class{}`. Now we need either full match or `(.*).class`.

To preserve the current behaviour, we need to add `(.*)` before each entry in the safelist that doesn't have it (the line starts with `.`, `#`, `[` or a char and doesn't contain `.css`).

**Expected behavior**
We need to make sure that after the update the same CSS is safelisted.

**Additional context**
It's a hard decision to make, as we know that the new approach fixes most of the issues and adding the prefix might cause too much CSS to be safelisted. However, the most important thing is to not have visual regression after the update.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Update the safelist items to be compatible with the new used CSS generator.

Fixes #4876 

## Type of change

## How Has This Been Tested?

- [x] Automated tests

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Admin/Settings.php
+++ b/inc/Engine/Optimization/RUCSS/Admin/Settings.php
@@ -433,11 +433,11 @@ class Settings {
 				continue;
 			}
 
-			if ( str_starts_with( $value, '(') ) {
+			if ( str_starts_with( $value, '(' ) ) {
 				continue;
 			}
 
-			$options['remove_unused_css_safelist'][$key] = '(.*)' . $value;
+			$options['remove_unused_css_safelist'][ $key ] = '(.*)' . $value;
 		}
 
 		update_option( 'wp_rocket_settings', $options );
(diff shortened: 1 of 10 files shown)


engahmeds3ed (approved):

Mai-Saad:
Working with the following notes:
1- `(.*)` will be auto-added to the start of **any** expression in safelist except if it starts with `(.*) `or its .css file
2- upgrade from 3.11.0.1 won't change the existing safelist
3- adding any new safelist shall be an exact expression or `(.*)` element to work correctly
4- this problem already on trunk, safelist `test(.*` will be `test((.*)`

Mai-Saad (approved):

PR #4888, opened 2022-04-04
Fixes #4032 Disabling heartbeat causing dependency error in query monitor

### LINKED ISSUE #4032: Disabling Heartbeat causes Missing Dependencies error in Query Monitor

**Describe the bug**
When Heartbeat settings are set to "disable", query monitor reports the following error:
`Missing Dependencies	wp-auth-check	example.com	wp-includes/js/wp-auth-check.min.js	heartbeat (missing), wp-i18n`

If settings are set to "reduce activity", the error disappears.

**To Reproduce**
1. Instal and activate [Query Monitor](https://wordpress.org/plugins/query-monitor/) plugin.
2. Set Heartbeat settings in WP Rocket to "disable" position.
3. Check the Query Monitor for the error. Screenshot: 

image

**Expected behavior**
No errors

**Additional context**
Query monitor developers comment:

> The script should be overridden rather than deregistered.
> WP Engine used to do this too, there's an old thread here: https://wordpress.org/support/topic/suppress-presistent-errors-ex-broken-dependencies-wp-auth-check-on-wp-engine/

Related ticket: https://secure.helpscout.net/conversation/1536822724/269979?folderId=3864740

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description

Create empty heartbeat script and enqueue script for fix on dependency error on query monitor.

Fixes #4032 

## Type of change

This solution is in no way different from the proposed one.

## How Has This Been Tested?

- Enable Heartbeat and select 'disable' on all options.
- Open Query Monitor and click on the scripts tab and see the dependency is not thrown.

# Checklist:

### DIFF
--- a/inc/Engine/Heartbeat/HeartbeatSubscriber.php
+++ b/inc/Engine/Heartbeat/HeartbeatSubscriber.php
@@ -61,6 +61,12 @@ class HeartbeatSubscriber implements Subscriber_Interface {
 		}
 
 		wp_deregister_script( 'heartbeat' );
+
+		/**
+		 * Enqueue an empty heartbeat script to prevent query monitor error
+		 * Added to the footer
+		 */
+		wp_enqueue_script( 'heartbeat', WP_ROCKET_ASSETS_JS_URL . 'assets/js/heartbeat.js', null, WP_ROCKET_VERSION, true );
 	}
 
 	/**


remyperona (approved):

Mai-Saad (approved):
Working as expected now.

PR #4893, opened 2022-04-05
Fixes #4868 Apply CDN to used CSS on output instead of on saving

### LINKED ISSUE #4868: CDN exclusions are not reflected in Used CSS

**Describe the bug**
When using CDN and RUCSS, CDN exclusions are not taken into the consideration.

**To Reproduce**
1. Enable CDN for all files
2. Enable RUCSS
3. Exclude URL from CDN (URL existing in used CSS) i.e /test.svg
4. visit the page and check used CSS

**Expected behavior**
Excluded URL not rewritten to the CNAME

**Additional**
Moving it from private repo:
https://github.com/wp-media/nodejs-treeshaker/issues/46

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description

Solves a problem where CDN exclusions are not taken in consideration when using RUCSS.
To solve this problem the filter `rocket_usedcss_content` has been moved from `check_job_status` to `get_used_css_markup`

Fixes #4868 

## Type of change

No

## How Has This Been Tested?

- [x] Automated tests


remyperona (approved):

remyperona (approved):

remyperona:
One thing to fix in the test for PHP 8
```
TypeError: error_reporting(): Argument #1 ($error_level) must be of type ?int, string given

/tests/Unit/inc/Engine/Optimization/RUCSS/Controller/UsedCSS/treeshake.php:31
```

CrochetFeve0251:
@Tabrisrp this is the PR where we got the problem with mockery. That function call is here to prevent the error to raise but we agreed with @engahmeds3ed that wasn't clean.

remyperona:
If this is the solution for now it's fine, but you are passing a string to `error_reporting` and it expects `int` value, so removing the quotes should be enough to prevent this error from showing on PHP 8

Mai-Saad (approved):
Working as expected

remyperona:
I'd suggest to skip the test for now, as even with the updated test suite it's still not passing. The issue is coming from the dependency, so we can't really change anything at that time.

PR #4904, opened 2022-04-07
Preserve Divi's "divi-style-inline-inline-css" attribute after treeshaking to prevent console error

### LINKED ISSUE #4598: RUCSS should preserve some <style> in some cases
- [x] Made sure you’re on the latest version `3.10.5`

**Describe the bug**
Some themes and plugin will try to manipulate existing `<style>`. 

When RUCSS is used, the target `<style>` element will be deleted. Leaving a console error in some cases.

Example: 

```
<html>
<body>
<!--StartFragment-->

<style id='divi-style-parent-inline-inline-css' type='text/css'>
/*!
Theme Name: Divi
Theme URI: http://www.elegantthemes.com/gallery/divi/
Version: 4.14.3
Description: Smart. Flexible. Beautiful. Divi is the most powerful theme in our collection.
Author: Elegant Themes
Author URI: http://www.elegantthemes.com
License: GNU General Public License v2
License URI: http://www.gnu.org/licenses/gpl-2.0.html
*/
</style>
```
```
<script type="application/javascript">
(function() {
    var file     = ["https:\/\/www.domain.com\/ka_seite\/et-cache\/222441\/et-divi-dynamic-tb-222629-222441-late.css"];
    var handle   = document.getElementById('divi-style-parent-inline-inline-css');
    var location = handle.parentNode;

    if (0===document.querySelectorAll('link[href="' + file + '"]').length) {
        var link  = document.createElement('link');
        link.rel  = 'stylesheet';
        link.id   = 'et-dynamic-late-css';
        link.href = file;

        location.insertBefore(link, handle.nextSibling);
    }
})();
</script>
```

After processing RUCSS, the `<style>` is removed.  And the style inside `.insertBefore()` is not processed. 

In this case, it's resulting in a console error: 

```
(index):1183 Uncaught TypeError: Cannot read properties of null (reading 'parentNode')
    at (index):1183
    at (index):1193
(anonymous) @ (index):1183
(anonymous) @ (index):1193
```

Safelisting selectors from inside `.insertBefore()` doesn't fix the issue.

**To Reproduce**
1. Add the above and target an element on the page using `.insertBefore()`
2. . If the script doesn't have a fallback in cache the targeted element doesn't exist, it will result in a console error.

**Expected behavior**
Ideally, we should check if `.insertBefore()` and preserve a placeholder with the selectors after Used CSS is injected. 

A broader approach would be to simply preserve placeholders for all inline `<style>` after getting processed by RUCSS. It should also work for https://github.com/wp-media/wp-rocket/issues/4554

**Screenshots**
N/A

**Additional context**
Related to https://github.com/wp-media/wp-rocket/issues/4554 but since the `<style>` wasn't empty, I added it as a separate issue.

Ticket - https://secure.helpscout.net/conversation/1727751196/314385

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

vmanthos:
## Description

Added `divi-style-inline-inline-css` to the `$inline_atts_exclusions` attributes array.

Fixes #4598 

## Type of change

It hasn't been groomed.

## How Has This Been Tested?

- [x] Using a helper plugin (`rocket_rucss_inline_atts_exclusions` filter) on the customer's website.

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -65,6 +65,7 @@ class UsedCSS {
 		'woodmart-inline-css-inline-css',
 		'woodmart_shortcodes-custom-css',
 		'rs-plugin-settings-inline-css', // For revolution slider, it saves settings for each slider.
+		'divi-style-inline-inline-css',
 	];
 
 	/**
--- a/inc/ThirdParty/Plugins/Optimization/AMP.php
+++ b/inc/ThirdParty/Plugins/Optimization/AMP.php
@@ -53,7 +53,8 @@ class AMP implements Subscriber_Interface {
 		];
 
 		if ( function_exists( 'is_amp_endpoint' ) ) {
-			$events['update_option_amp-options'] = 'generate_config_file';
+			$events['update_option_amp-options']  = 'generate_config_file';
+			$events['rocket_delay_js_exclusions'] = 'exclude_script_from_delay_js';
 		}
 
 		return $events;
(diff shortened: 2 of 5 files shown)


remyperona (approved):

Mai-Saad (approved):
The inline style for Divi is now preserved (can see problems in new test templates but already on the trunk)

PR #4908, opened 2022-04-08
Added WordPress redirection to buffer subscriber

### LINKED ISSUE #1392: Respect site's trailing slash settings

CrochetFeve0251:
### Note: 
An PR already exist for this https://github.com/wp-media/wp-rocket/issues/1392.
However due to the fact we saw that after the code was written and the fact this PR is 2 years old and implements some stuffs that needs to be deleted we preferred create a new one.

@Tabrisrp can you confirm this option is fine?

## Description

Fix a problem with redirection between slash ended url and non slash.
To fix this a new method has been attached to the `template_redirect` action  that calls the WordPress `redirect_canonical`.

Fixes #1392 

## Type of change

Only the first part of grooming has been implemented as requested by @piotrbak 

## How Has This Been Tested?

- [x] Automated Tests
- [x] Manually on local env

# Checklist:

### DIFF
--- a/tests/Unit/inc/classes/subscriber/Optimization/BufferSubscriber/redirectCanonical.php
+++ b/tests/Unit/inc/classes/subscriber/Optimization/BufferSubscriber/redirectCanonical.php
@@ -18,9 +18,9 @@ class Test_RedirectCanonical extends TestCase
 
 	public function setUp(): void
 	{
+		parent::setUp();
 		$this->optimizer = \Mockery::mock(Optimization::class);
 		$this->subscriber = new Buffer_Subscriber($this->optimizer);
-		parent::setUp();
 	}
 
 	public function testShouldCallRedirect() {
--- a/tests/Unit/inc/classes/subscriber/Optimization/BufferSubscriber/redirectCanonical.php
+++ b/tests/Unit/inc/classes/subscriber/Optimization/BufferSubscriber/redirectCanonical.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace WP_Rocket\Tests\Unit\inc\classes\subscriber\Optimization\BufferSubscriber;
+
+use WP_Rocket\Buffer\Optimization;
+use WP_Rocket\Subscriber\Optimization\Buffer_Subscriber;
+use WP_Rocket\Tests\Unit\TestCase;
+use Brain\Monkey\Functions;
+
+/**
+ * @covers \WP_Rocket\Subscriber\Optimization\Buffer_Subscriber::redirect_canonical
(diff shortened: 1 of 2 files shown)


tests/Unit/inc/classes/subscriber/Optimization/BufferSubscriber/redirectCanonical.php
```diff
@@ -0,0 +1,31 @@
...
+
+	public function setUp(): void
+	{
+		$this->optimizer = \Mockery::mock(Optimization::class);
+		$this->subscriber = new Buffer_Subscriber($this->optimizer);
+		parent::setUp();
+	}
+
+	public function testShouldCallRedirect() {
+		Functions\expect('redirect_canonical')->with();
```
remyperona:
```suggestion
		Functions\expect('redirect_canonical')->once();
```

tests/Unit/inc/classes/subscriber/Optimization/BufferSubscriber/redirectCanonical.php
```diff
@@ -0,0 +1,31 @@
...
+class Test_RedirectCanonical extends TestCase
+{
+	protected $subscriber;
+	protected $optimizer;
+
+	public function setUp(): void
+	{
+		$this->optimizer = \Mockery::mock(Optimization::class);
+		$this->subscriber = new Buffer_Subscriber($this->optimizer);
+		parent::setUp();
```
remyperona:
```suggestion
		parent::setUp();
		$this->optimizer = \Mockery::mock(Optimization::class);
		$this->subscriber = new Buffer_Subscriber($this->optimizer);
```

remyperona (changes requested):

tests/Unit/inc/classes/subscriber/Optimization/BufferSubscriber/redirectCanonical.php
CrochetFeve0251: @Tabrisrp I don't get this one with no parameters is not more precise that once?

PR #4911, opened 2022-04-11
Closes #2842 delete minified files on theme switch

### LINKED ISSUE #2842: Delete files in cache/min when switching themes
- Made sure you’re on the latest version 👍 
- Used the search feature to ensure that the bug hasn’t been reported before 👍 

**Describe the bug**
When a new theme is activated, we do not clear out the old combined and minified files. Since we know that these files are going to change, it's best if we can clear them. 

Video: https://youtu.be/NRPUp6Fr_SU

**To Reproduce**
- Enable CSS and/or JS combination and make sure files are generated in `cache/min`
- Change the theme. Existing files are not deleted. New files are generated.
- Change the theme again. The above will repeat. 

**Expected behavior**
When switching a theme, clear everything in `cache/min` folder. 

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

jeawhanlee:
## Description

Added a new AdminSubscriber class to to listen for `switch_theme `event and clean cache/min with `rocket_clean_minify()`

Fixes #2842 

## Type of change

Solution is in no way different from the proposed one.

## How Has This Been Tested?

- Enable Minify & Merge Options for both CSS & JS
- Activate a new theme
- Load new site in incognito with Ctrl+f5(windows) for force refresh
- check the wp-content/cache/min folder to see minified files generated
- Activate another theme
- check to see that previous minified files have been removed.

# Checklist:


jeawhanlee:
@Tabrisrp Do I need to write tests for this new class?

remyperona:
@jeawhanlee Yes, we want to have automated tests as much as possible

inc/Engine/Optimization/Minify/AdminSubscriber.php
```diff
@@ -0,0 +1,31 @@
...
+			'switch_theme' => 'clean_minify_all',
+		];
+	}
+
+	/**
+	 * Delete all minified cache file
+	 */
+	public function clean_minify_all() {
+		// Delete all minify cache files.
+		rocket_clean_minify();
```
remyperona, line 53: We should check here if at least minify CSS or minify JS is enabled, to avoid doing an unnecessary clean call if the options are not used. You can use the `Options_Data` class as a dependency to do that

remyperona (changes requested):

inc/Engine/Optimization/Minify/AdminSubscriber.php
```diff
@@ -23,8 +40,15 @@ public static function get_subscribed_events() {
 
 	/**
 	 * Delete all minified cache file
+	 *
+	 * @return void
 	 */
 	public function clean_minify_all() {
+		// Bail out if minify_js or minify_css is not enabled.
+		if ( ! (bool) $this->options->get( 'minify_js', 0 ) & ! (bool) $this->options->get( 'minify_css', 0 ) ) {
```
remyperona:
```suggestion
		if ( ! (bool) $this->options->get( 'minify_js', 0 ) && ! (bool) $this->options->get( 'minify_css', 0 ) ) {
```

tests/Integration/inc/Engine/Optimization/Minify/AdminSubscriber/cleanMinifyAll.php
```diff
@@ -0,0 +1,30 @@
...
+	/**
+	 * @dataProvider providerTestData
+	 */
+	public function testCleanMinifyAll( $option, $expected ) {
+		$this->dumpResults = isset( $expected['dump_results'] ) ? $expected['dump_results'] : false;
+		$this->generateEntriesShouldExistAfter( $expected['cleaned'] );
+
+		$this->checkEntriesDeleted( $expected['cleaned'] );
+		$this->checkShouldNotDeleteEntries();
+	}
```
remyperona: we're missing the call to the `switch_theme` action to validate the expectations

remyperona (changes requested):

piotrbak:
@jeawhanlee this one is ready for review, right?

jeawhanlee:
@piotrbak Yes it is.

remyperona (approved):

vmanthos:
@piotrbak @jeawhanlee 

I tested this PR on:
* Single site
* Multisite

The files in `cache/min/` in both cases are removed. However, not all folders are.  For example, if there is a file:
`/wp-content/cache/min/1/wp-content/themes/twentytwenty/assets/js/something.js`

The file and the `/js/` folder will be removed but none of the parent folders will:
`/wp-content/cache/min/1/wp-content/themes/twentytwenty/assets/`

I'm guessing this is related to how `rocket_clean_minify()` works.

Is this something we need to handle as well?

I checked for an issue related to this but I didn't find one. Let me know if you want me to GitHub this.

piotrbak:
@vmanthos I think we're fine with directories, leaving them empty won't cause any issues. Do you see any bad sides of this?

vmanthos:
> Do you see any bad sides of this?

No, I don't other than increasing the number of inodes. But this shouldn't be a problem.

vmanthos:
> Do you see any bad sides of this?

No, I don't other than increasing the number of inodes. But this shouldn't be a problem.

vmanthos (approved):
Working as expected. 👍

PR #4928, opened 2022-04-14
Closes #4881 Added constant in Avada subscriber to deactivate combined CSS and JS from Avada theme

### LINKED ISSUE #4881: Disable Avada's combination features when Remove Unused CSS is active (>3.11)

When customers who are using Avada have its CSS/JavaScript combination features active, it won't be possible for them to benefit from the compatibility additions we are making in the RUCSS back-end.

This is because it depends on detecting specific files (CSS or JavaScript) to add the CSS selectors we found.

Disable Avada's CSS/JavaScript combination when the Remove Unused CSS is active for WP Rocket > 3.10.

**Additional context**

This came up after an internal discussion with @piotrbak.

We can also document why enabling these features isn't helpful towards performance.

CrochetFeve0251:
## Description

Fix an issue where Avada combined CSS file and prevent RUCSS to work sucessfully.
For that a constant `FUSION_DISABLE_COMPILERS` is set to prevent Avada from combining files.

Fixes #4881 
## Type of change 

No

## How Has This Been Tested?

- [x] Automated Test
- [x] On my local env

# Checklist:

### DIFF
--- a/inc/ThirdParty/Themes/Avada.php
+++ b/inc/ThirdParty/Themes/Avada.php
@@ -33,7 +33,7 @@ class Avada implements Subscriber_Interface {
 			'fusion_cache_reset_after'             => 'clean_domain',
 			'update_option_fusion_options'         => [ 'maybe_deactivate_lazyload', 10, 2 ],
 			'rocket_wc_product_gallery_delay_js_exclusions' => 'exclude_delay_js',
-			'init' => 'disable_compilers',
+			'init'                                 => 'disable_compilers',
 		];
 	}
 
--- a/inc/ThirdParty/Themes/Avada.php
+++ b/inc/ThirdParty/Themes/Avada.php
@@ -26,7 +26,6 @@ class Avada implements Subscriber_Interface {
 		if ( ! self::is_avada() ) {
 			return [];
 		}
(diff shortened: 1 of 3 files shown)


remyperona (approved):

remyperona (changes requested):
Shouldn't we only set the constant if the RUCSS option is enabled?

remyperona (approved):

vmanthos (approved):
Working as expected.

@piotrbak @CrochetFeve0251 

When disabling Remove Unused CSS on a page using WP Rocket's meta-box, Avada's CSS combination is disabled there as well, i.e. the same as with the rest of the site.

Is this something we'd like to address?

piotrbak:
@vmanthos Then it seems to be working as expected. I think we're fine here, you agree? 🆗

vmanthos:
> I think we're fine here, you agree? 

If what we mean to do is globally disable their combination when Remove Unused CSS is enabled, then yes.

Otherwise, if their combination should be active on pages where Remove Unused CSS is disabled, then we need to work on this.

piotrbak:
Ah! I got your point now, good catch! 

I believe that it's going to be an edge case and we should be fine here.

Issue #2264, opened 2020-01-17
Improve varnish cache support to clear only what is necessary


cdvv7788:
Whenever I try to `clear cache`, a PURGE request with `/.*` is sent to varnish. This has some a little issue: PURGE in varnish does not handle regexes.
If we filter the request method in VCL, and change it to use BAN, it works, however we get a troublesome situation when we update a single element in the site:
- A PURGE request with `/.*` is sent to varnish, so if we implement the BAN action as previously described, the WHOLE cache will be BANNED. This is the main issue, clearing the whole cache when updating a single post.
- There are a lot more PURGE requests sent to specific paths. They have a header with the method, so those could be used to decide which method to use (BAN or PURGE) if the previous problem didn't exist.

I am not sure about how wp-rocket works in this case (if it should clear the whole cache every time I update a single entry), but it sounds like it is not the way varnish cache should be invalidated.

trendymail:
Hi!

Same problem here.

When WP Rocket sends a PURGE request to Varnish with URL "/" or "/*", I have to ignore these requests because I do not want the whole Varnish cache invalidated (images, css, js, etc).

PURGE requests should be tied to pages/category/posts (if this is possible).

Have a nice day!

GeekPress:
Hi @cdvv7788 🤚

Thanks for your contribution!

@Tabrisrp Do you have any ideas about this issue?

Tabrisrp:
To give more context on how the Varnish sync works:

With our purge request we are following the recommended configuration from Varnish: https://www.varnish-software.com/wiki/content/tutorials/wordpress/wp_step_by_step.html#handling-purge-requests

- If the purge request contains the header X-purge-method: regex, it uses ban
- If not, it uses a normal purge

The purge requests are sent right before a purge event happend in WP Rocket:
- The whole cache was cleared
- The homepage was cleared
- A content was modified, so a partial clear happened

See: https://github.com/wp-media/wp-rocket/blob/8cf329ca6f1e7ad9145c80499cfd4be98e702e3b/inc/Addon/Varnish/Subscriber.php#L44

trendymail:
Hi!

I could be wrong but, AFAIK, header "X-purge-method" is not handled by the default builtin VCL file of Varnish.

Instead, you have to handle this specific header manually. So you need to write a "custom" VCL rule.

Nevertheless, if WP Rocket could send regex compatible ban requests, it would be great!

Have a nice day!

GeekPress:
[...]

PR #4929, opened 2022-04-14
Fixes #4875 Clean Varnish cache when used CSS is completed or purged

### LINKED ISSUE #4875: Varnish's cache isn't cleared when clearing page-specific used CSS
- Made sure you’re on the latest version ✅ `3.11.0.1`

**Describe the bug**

When clearing the used CSS for a specific URL, the cache of the URL is properly cleared:
https://github.com/wp-media/wp-rocket/blob/37eaa0016d79c21fbdc3bfc32102afd5eb750fdd/inc/Engine/Cache/PurgeActionsSubscriber.php#L54

However, if Varnish is used, e.g. on Cloudways, the cache of the URL won't be cleared at that time, and:
- the page won't be preloaded - the `wp_remote_get()` request will never reach WordPress
- Remove Unused CSS won't be created/applied there, until the specific URL's cache is cleared using the admin bar menu item, or the whole cache is cleared

**To Reproduce**

1. In an environment where Varnish is used, clear page-specific used CSS using the admin bar menu item.
2. Check the `age` header of the Varnish. That will be larger than `0`.

**Expected behavior**

 Varnish's cache should be cleared when using the "Clear used CSS of this URL" button.

**Additional context**

We should hook `clean_file` to `rocket_rucss_after_clearing_usedcss` here:
https://github.com/wp-media/wp-rocket/blob/37eaa0016d79c21fbdc3bfc32102afd5eb750fdd/inc/Addon/Varnish/Subscriber.php#L41-L47

**Ticket:** https://secure.helpscout.net/conversation/1828951060/334087/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description
Fix a problem with cache from Varnish that wasn't cleaned.
For that we hooked `clean_file` method to `rocket_rucss_after_clearing_usedcss` hhook.

Fixes #4875

## Type of change

No

## How Has This Been Tested?

- [ ] Automated testing

# Checklist:


remyperona (approved):

remyperona:
@CrochetFeve0251 Can you also add the clearing on `rocket_rucss_complete_job_status`

CrochetFeve0251:
@Tabrisrp I added the hook

remyperona (approved):

Mai-Saad:
@piotrbak please find exploratory test notes/Qs below: (using the test site http://wprocket-test.odns.fr/)
1- Switching theme while the separate mobile cache is enabled, only recreating the desktop version in the used CSS table. although, the new labs test site, is generating both desktop and mobile versions. is this something that we need to handle here or on another ticket?
2- When we change the safelist, the Cache of any page other than the home page is cleared (age header set to 0) while it is not cleared for the home page. Do we need a Separate GH for this?

piotrbak:
@Mai-Saad 

1. Was it happening when the server-side cache was enabled (Varnish or similar)?
2. Is the `age` on homepage reduced or it's increased after changing the safelist?

Mai-Saad:
> @Mai-Saad
> 
> 1. Was it happening when the server-side cache was enabled (Varnish or similar)?
> 2. Is the `age` on homepage reduced or it's increased after changing the safelist?

@piotrbak 
1. The problem is happening here http://wprocket-test.odns.fr/ and Yes Varnish is used
2. Age is reduced but does not reach 0 even after UCSS is completed in this case

piotrbak:
@Mai-Saad I think we are fine then. 

For the first one:
I believe that this Varnish setup doesn't have the `Separate Cache for Mobiles` feature and serves the same set of cache for both options. Here's the scenario:
1. We're switching the theme
2. Request is being done on the desktop, Varnish saves the cache
3. Request is being done on the mobile, Varnish serves the same cache

For the second one:
If the `age` header is decreased, it means that it was cleared in the meantime. We could confirm that by changing the safelist that would make a change in the Used CSS of the homepage.

Mai-Saad (approved):
Working as expected then.
Note: Dreampress tests were run on o2switch env and failed tests due to https://github.com/wp-media/wp-rocket/issues/2264
[testrail-report-389.pdf](https://github.com/wp-media/wp-rocket/files/8804494/testrail-report-389.pdf)

PR #4941, opened 2022-04-18
Bugfix: Remove last error check

engahmeds3ed:
## Description

Here
https://github.com/wp-media/wp-rocket/blob/86260dd09afd4ac15f44ad64b0219baf723a5f1c/inc/Engine/Optimization/RUCSS/Admin/Subscriber.php#L170-L177
we used this check to delete the transient that has the number of Action Scheduler's tables not to make this query with each `init` but now disabled this caching in admin side so no need to use this code.

Also @vmanthos told me today the following (Many thanks Boss):
> I've seen at least two tickets where no used CSS was generated because there was some kind of error: ` Constant WP_CACHE already defined`

So with any error in the log, we are bailing out there and this may lead the CRON JOB not to be scheduled especially when the customer has `php8` so many errors are in the log.

## Type of change

## How Has This Been Tested?

1. Define `WP_CACHE` twice into `wp-config.php` before enabling RUCSS
2. Enable RUCSS
3. Check the CRON JOB if it's scheduled or not inside action scheduler tools page

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
+++ b/inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
@@ -109,7 +109,7 @@ class Subscriber implements Subscriber_Interface {
 			],
 			'wp_ajax_rocket_spawn_cron'              => 'spawn_cron',
 			'rocket_deactivation'                    => 'cancel_queues',
-			'shutdown'                               => 'schedule_rucss_pending_jobs_cron',
+			'admin_init'                             => 'schedule_rucss_pending_jobs_cron',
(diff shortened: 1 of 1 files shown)


engahmeds3ed:
Here also I changed the way we detect missing Action Scheduler tables to use try catch instead of adding our code in a callback for the hook `shutdown`.

mostafa-hisham (approved):

Mai-Saad (approved):
Working as expected with both update/fresh install when `wp_cache` defines twice or using PHP 8.1+, and when actions table is deleted but [this fix](https://github.com/wp-media/wp-rocket/issues/4728) exists on current branch

PR #4965, opened 2022-04-25
Fix problem with action scheduler

### LINKED ISSUE #4960: Improve the control over rocket_rucss_pending_jobs_cron action
**Describe the bug**
Currently, we improved the behaviour described in the https://github.com/wp-media/wp-rocket/issues/4940 based on the https://github.com/woocommerce/action-scheduler/issues/698 suggestions. However, we can improve it even further by adding a health check cron.

**Expected behavior**
We'll schedule a permanent action/cron that will check the number of `pending` or `in-progress` `rocket_rucss_pending_jobs_cron` actions:
1. 0 actions - we need to schedule it
2. 1 action - we're just fine, no further action needed
3. 1+ actions - we need to reduce it to 1

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description

We had a problem with multiple actions running at the same time.
We fixed this by cancelling all running and pending action for the hook if multiple actions are launched.

Fixes #4960

## Type of change

No.

## How Has This Been Tested?

- [x] Tested on my testing environment

# Checklist:

### DIFF
--- a/inc/Engine/Common/Queue/AbstractASQueue.php
+++ b/inc/Engine/Common/Queue/AbstractASQueue.php
@@ -48,13 +48,16 @@ abstract class AbstractASQueue implements QueueInterface {
 	 */
 	public function schedule_recurring( $timestamp, $interval_in_seconds, $hook, $args = [] ) {
 		if ( $this->is_scheduled( $hook, $args ) ) {
-			$scheduled_actions = $this->search( [
-				'hook' => $hook,
-				'status' => [
-					ActionScheduler_Store::STATUS_RUNNING,
-					ActionScheduler_Store::STATUS_PENDING
-				]
-			], 'ids' );
+			$scheduled_actions = $this->search(
+				 [
+					 'hook'   => $hook,
+					 'status' => [
+						 ActionScheduler_Store::STATUS_RUNNING,
+						 ActionScheduler_Store::STATUS_PENDING,
+					 ],
+				 ],
+				 'ids'
+				);
 
 			if ( 1 === count( $scheduled_actions ) ) {
 				return '';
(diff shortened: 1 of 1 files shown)


inc/Engine/Common/Queue/AbstractASQueue.php
```diff
@@ -46,8 +48,13 @@ public function schedule_single( $timestamp, $hook, $args = [] ) {
 	 */
 	public function schedule_recurring( $timestamp, $interval_in_seconds, $hook, $args = [] ) {
 		if ( $this->is_scheduled( $hook, $args ) ) {
+			$all = $this->search( [ 'status' => [ ActionScheduler_Store::STATUS_RUNNING, ActionScheduler_Store::STATUS_PENDING ] ] );
+			if ( count( $all ) > 1 ) {
+				$this->cancel_all( $hook, $args );
+			}
```
engahmeds3ed: Here I have some comments as follows:-
1. `search` method's second parameter is `$return_format` and its default value is `OBJECT` and we don't need those objects so I believe we can use `ids` instead so we just get the actions IDs and that's good from performance perspective.
2. You need to use Yoda condition like that
    ```if ( 1 < count( $all ) ) {```
3. Use descriptive variable names like `$scheduled_actions`
4. You need to apply the concept of `return early` instead of using nested if conditions to be like that:
```suggestion
			$scheduled_actions = $this->search( [ 'status' => [ ActionScheduler_Store::STATUS_RUNNING, ActionScheduler_Store::STATUS_PENDING ] ] );
			if ( 1 === count( $scheduled_actions ) ) {
				return '';
			}
			
			// Here means that it's greater than 1 as this is enclosed into the condition of is_scheduled.
			$this->cancel_all( $hook, $args );
			return '';
```
5. Add a new condition to the `search` method call for the group to search only for our group (Small optional enhancement)

What do u think about those points?

engahmeds3ed (changes requested):

engahmeds3ed (approved):

engahmeds3ed (approved):

PR #4978, opened 2022-04-27
Fix #4231 Remove params from url to prevent JS to be interpeted as internal instead of external

### LINKED ISSUE #4231: JS minification exclusion isn't working when an external URL contains the site's TLD
- Made sure you’re on the latest version ✅ `3.9.1.1`

**Describe the bug**

On a site, e.g. `www.example.com`, when an external JavaScript file's URL contains the site's TLD, e.g. in a query string, like so:
```HTML
<script type="text/javascript" src="https://tarteaucitron.io/load.js?locale=fr&iswordpress=true&domain=www.example.com&uuid=a334ff7567f29a1047ddb578baf4dc230689285d"></script>
```
WP Rocket is treating the file as internal, because of how we are checking it here:
https://github.com/wp-media/wp-rocket/blob/7bc0dd8dcc0ef5e2af156d978c8a437ec8bb1bd1/inc/Engine/Optimization/AbstractOptimization.php#L133

Because of this, if the customer wants to exclude the file, using `tarteaucitron.io/load.js` or `tarteaucitron.io` will not work. Instead, `/load.js` will.

**To Reproduce**

1. Use the script I added earlier, making sure the `domain` query string matches the TLD of the test site.
2. Add `https://tarteaucitron.io/load.js` to the JavaScript minification exclusion text area.
3. Check the source code of a page. The exclusion will not work, i.e. the file will be served from the local cache.

**Expected behavior**

Exclusions for such URLs should work the same with all external URLs.

**Additional context**

This is definitely an edge case, but I'm documenting it in case we want to deal with this.

**Ticket:** https://secure.helpscout.net/conversation/1584764409/283487/
**Slack conversation:** https://wp-media.slack.com/archives/C08N8J6VC/p1628096218053100

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description
This fix an issue where some JS had the domain had in their params resulting being interpreted as internal files instead of external.
To fix that we remove params from the url of the script before the check.

Fixes #4231

## Type of change

No.

## How Has This Been Tested?

- [x] Automated Tests

# Checklist:

### DIFF
--- a/tests/Unit/inc/Engine/Optimization/AbstractOptimization/isExternalFile.php
+++ b/tests/Unit/inc/Engine/Optimization/AbstractOptimization/isExternalFile.php
@@ -3,7 +3,7 @@ namespace WP_Rocket\Tests\Unit\inc\Engine\Optimization\AbstractOptimization;
 
 use ReflectionClass;
 use WP_Rocket\Engine\Optimization\AbstractOptimization;
-use function Brain\Monkey\Functions;
+use Brain\Monkey\Functions;
 
 class Test_IsExternalFile extends \WP_Rocket\Tests\Unit\TestCase {
 
@@ -19,8 +19,28 @@ class Test_IsExternalFile extends \WP_Rocket\Tests\Unit\TestCase {
 	 * @dataProvider configTestData
 	 */
 	public function testShouldReturnAsExpected($config, $expected) {
-		Functions\expect('get_rocket_parse_url')->with($config['url'])-->andReturn($config['file']);
+		Functions\expect('get_rocket_parse_url')->with($config['url'])->andReturn($config['file']);
+		$this->configureParseContent($config);
+		$this->configureCollectHosts($config);
 		$this->assertEquals($expected, self::callProtectedMethod($this->optimization, 'is_external_file', array($config['url'])));
+	}
+
+	protected function configureParseContent($config) {
+		if(! array_key_exists('parse_content', $config)) {
+			return;
+		}
+		Functions\expect('content_url')->with()->andReturn($config['content_url']);
+		Functions\expect('wp_parse_url')->with($config['url'])->andReturn($config['url_parsed']);
+	}
+
+	protected function configureCollectHosts($config) {
+		if(! array_key_exists('collect_hosts', $config)) {
+			return;
+		}
+		$this->optimization->expects(self::once())->method('get_zones')->willReturn($config['zones']);
+		Functions\expect('apply_filters')->with('rocket_cdn_hosts', [], $config['zones'])->andReturn($config['cdn_hosts']);
+		Functions\expect('get_rocket_i18n_uri')->with()->andReturn($config['lang_hosts']);
+		Functions\expect('wp_parse_url')->with($config['lang_url'], PHP_URL_HOST)->andReturn($config['url_host']);
 	}
 
 	public static function callProtectedMethod($object, $method, array $args=array()) {
--- a/tests/Unit/inc/Engine/Optimization/AbstractOptimization/isExternalFile.php
+++ b/tests/Unit/inc/Engine/Optimization/AbstractOptimization/isExternalFile.php
@@ -1,0 +1,32 @@
+<?php
+namespace WP_Rocket\Tests\Unit\inc\Engine\Optimization\AbstractOptimization;
+
+use ReflectionClass;
+use WP_Rocket\Engine\Optimization\AbstractOptimization;
+use function Brain\Monkey\Functions;
+
+class Test_IsExternalFile extends \WP_Rocket\Tests\Unit\TestCase {
+
+	protected $optimization;
+
+	public function setUp(): void
+	{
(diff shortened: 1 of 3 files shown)


tests/Unit/inc/Engine/Optimization/AbstractOptimization/isExternalFile.php
```diff
@@ -0,0 +1,52 @@
...
+		Functions\expect('get_rocket_i18n_uri')->with()->andReturn($config['lang_hosts']);
+		Functions\expect('wp_parse_url')->with($config['lang_url'], PHP_URL_HOST)->andReturn($config['url_host']);
+	}
+
+	public static function callProtectedMethod($object, $method, array $args=array()) {
+		$class = new ReflectionClass(get_class($object));
+		$method = $class->getMethod($method);
+		$method->setAccessible(true);
+		return $method->invokeArgs($object, $args);
+	}
```
remyperona: You can use the `get_reflective_method()` from the wp-media/phpunit `TestCaseTrait` to do this

tests/Unit/inc/Engine/Optimization/AbstractOptimization/isExternalFile.php
```diff
@@ -0,0 +1,52 @@
+<?php
+namespace WP_Rocket\Tests\Unit\inc\Engine\Optimization\AbstractOptimization;
+
+use ReflectionClass;
+use WP_Rocket\Engine\Optimization\AbstractOptimization;
+use Brain\Monkey\Functions;
+
+class Test_IsExternalFile extends \WP_Rocket\Tests\Unit\TestCase {
```
remyperona: import the TestCase before using it

remyperona (changes requested):

tests/Unit/inc/Engine/Optimization/AbstractOptimization/isExternalFile.php
```diff
@@ -43,9 +44,8 @@ protected function configureCollectHosts($config) {
 		Functions\expect('wp_parse_url')->with($config['lang_url'], PHP_URL_HOST)->andReturn($config['url_host']);
 	}
 
-	public static function callProtectedMethod($object, $method, array $args=array()) {
-		$class = new ReflectionClass(get_class($object));
-		$method = $class->getMethod($method);
+	public function callProtectedMethod($object, $method, array $args=array()) {
+		$method = self::get_reflective_method($method, get_class($object));
 		$method->setAccessible(true);
```
remyperona: `$method = $this->get_reflective_method($method, get_class($object));`

You don't need `$method->setAccessible(true);` anymore, it's already in `get_reflective_method()`

tests/Unit/inc/Engine/Optimization/AbstractOptimization/isExternalFile.php
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	/**
+	 * @dataProvider configTestData
+	 */
+	public function testShouldReturnAsExpected($config, $expected) {
+		Functions\expect('get_rocket_parse_url')->with($config['url'])->andReturn($config['file']);
+		$this->configureParseContent($config);
+		$this->configureCollectHosts($config);
+		$this->assertEquals($expected, $this->callProtectedMethod($this->optimization, 'is_external_file', array
```
remyperona:
```suggestion
		$this->assertSame($expected, $this->callProtectedMethod($this->optimization, 'is_external_file', array
```

remyperona (changes requested):

tests/Unit/inc/Engine/Optimization/AbstractOptimization/isExternalFile.php
CrochetFeve0251: Oh ok

remyperona (approved):

Mai-Saad:
@piotrbak the same problem occurs in CSS files with Minify CSS, Do we need to handle this in the current PR?
i.e 
`<link id='styles-aqua-bg' rel='stylesheet' href="//maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.css?test=123&test2=123&domain=https://new.rocketlabsqa.ovh">`

when we exclude `maxcdn.bootstrapcdn.com/font-awesome/`, it won't be excluded and will be minified.

piotrbak:
@Mai-Saad @CrochetFeve0251 Yes, let's do it in the same way for CSS Minification

CrochetFeve0251:
@piotrbak @Mai-Saad the CSS minifer extends from the class I modified can you check if the change also worked for CSS urls?

Mai-Saad (approved):
As per discussion with @CrochetFeve0251 for external exclusion, we need either full URL or domain name and currently, both are working fine on the branch with JS/CSS while having/not having DN in the URL

PR #4996, opened 2022-05-02
Changed Divi thirdparty class to disable dynamic js on RUCSS active and added its tests

### LINKED ISSUE #4792: Divi's Dynamic CSS feature can potentially break the layout when RUCSS is active
- Made sure you’re on the latest version ✅ `3.10.9`

**Describe the bug**

Divi's Dynamic CSS feature (https://i.vgy.me/wFqnNA.jpg) will load a CSS file using JavaScript:
```JavaScript
<script type="application/javascript">
			(function() {
				var file     = ["https:\/\/www.example.com\/wp-content\/et-cache\/taxonomy\/product_cat\/2053\/et-divi-dynamic-late.css"];
				var handle   = document.getElementById('divi-style-parent-inline-inline-css');
				var location = handle.parentNode;

				if (0===document.querySelectorAll('link[href="' + file + '"]').length) {
					var link  = document.createElement('link');
					link.rel  = 'stylesheet';
					link.id   = 'et-dynamic-late-css';
					link.href = file;

					location.insertBefore(link, handle.nextSibling);
				}
			})();
</script>
```

This may contain rules that override rules in the used CSS, thus breaking the layout.

In both cases I checked, the only solution was to disable Divi's feature.

**To Reproduce**

I'm trying to figure out when the specific files will be loaded on the front-end. I'll update this as soon as I find a way.

**Expected behavior**

The layout shouldn't break with Divi.

**Additional context**

**Tickets:** 
https://secure.helpscout.net/conversation/1788130755/326288
https://secure.helpscout.net/conversation/1794925397/327633

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
…

## Description
Fix a problem where RUCSS has incompatibilities with a feature from Divi theme.
For that we disabled the feature when RUCSS is activated inside the thirdparty class.

Fixes #4792 

## Type of change

No.

## How Has This Been Tested?

- [x] Automated tests
- [x] Tested on local env

# Checklist:

### DIFF
--- a/inc/ThirdParty/Themes/Divi.php
+++ b/inc/ThirdParty/Themes/Divi.php
@@ -207,8 +207,7 @@ class Divi implements Subscriber_Interface {
 	 * @return void
 	 */
 	public function disable_dynamic_css_on_rucss() {
-
-		if ( ! apply_filters( 'pre_get_rocket_option_remove_unused_css', false ) ) { // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
+		if (! $this->options->get('remove_unused_css', false) ) {
 			return;
 		}
 		add_filter( 'et_use_dynamic_css', '__return_false' );
--- a/inc/ThirdParty/Themes/Divi.php
+++ b/inc/ThirdParty/Themes/Divi.php
@@ -207,7 +207,7 @@ class Divi implements Subscriber_Interface {
 	 * @return void
 	 */
 	public function disable_dynamic_css_on_rucss() {
-		if (! $this->options->get('remove_unused_css', false) ) {
+		if ( ! $this->options->get( 'remove_unused_css', false ) ) {
 			return;
 		}
 		add_filter( 'et_use_dynamic_css', '__return_false' );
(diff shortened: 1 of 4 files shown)


inc/ThirdParty/Themes/Divi.php
```diff
@@ -198,4 +200,18 @@ public function exclude_css_from_combine( $exclude_css ) {
...
 	}
+
+	/**
+	 * Disable Divi dynamic CSS when RUCSS is activated
+	 *
+	 * @return void
+	 */
+	public function disable_dynamic_css_on_rucss() {
+
+		if ( ! apply_filters( 'pre_get_rocket_option_remove_unused_css', false ) ) { // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound
```
remyperona: Is there a reason to use the filter instead of using the option value from the injected `$options` instance?

remyperona (changes requested):

inc/ThirdParty/Themes/Divi.php
CrochetFeve0251: @Tabrisrp no I just didn't knew we can get it this way

remyperona (approved):

vmanthos (approved):
Divi's Dynamic CSS was disabled as expected.

This is also the case for pages where Remove Unused CSS is disabled using our meta-box.

PR #4897, opened 2022-04-06
Fix missing AS missing DB tables issue

engahmeds3ed:
## Description

In some cases when client updates to our latest version 3.11.0.2, they face fatal error because one or more tables related to Action Scheduler isn't there, the error is like the following:-

```
[05-Apr-2022 12:36:29 UTC] PHP Fatal error:  Uncaught RuntimeException: Error saving action: Table 'wordpress.wp_actionscheduler_actions' doesn't exist in /var/www/html/wordpress/wp-content/plugins/wp-rocket/inc/Dependencies/ActionScheduler/classes/data-stores/ActionScheduler_DBStore.php:86
Stack trace:
#0 /var/www/html/wordpress/wp-content/plugins/wp-rocket/inc/Dependencies/ActionScheduler/classes/ActionScheduler_ActionFactory.php(177): ActionScheduler_DBStore->save_action()
#1 /var/www/html/wordpress/wp-content/plugins/wp-rocket/inc/Dependencies/ActionScheduler/classes/ActionScheduler_ActionFactory.php(105): ActionScheduler_ActionFactory->store()
#2 /var/www/html/wordpress/wp-content/plugins/wp-rocket/inc/Dependencies/ActionScheduler/functions.php(54): ActionScheduler_ActionFactory->recurring()
#3 /var/www/html/wordpress/wp-content/plugins/wp-rocket/inc/Engine/Common/Queue/AbstractASQueue.php(51): as_schedule_recurring_action()
#4 /var/www/html/wordpress/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/Controller/Queue.php(57): WP_Rocket\Eng in /var/www/html/wordpress/wp-content/plugins/wp-rocket/inc/Dependencies/ActionScheduler/classes/data-stores/ActionScheduler_DBStore.php on line 86
```

## Type of change

Please delete options that are not relevant.

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Enhancement (non-breaking change which improves an existing functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Is the solution different from the one proposed during the grooming?

Not groomed.

## How Has This Been Tested?

I need to test the queue runner by sending too much jobs from different groups and see if it'll handle our jobs with priority.

[...]

PR #5014, opened 2022-05-05
Fixes #5005 Use WP cron for RUCSS pending jobs

### LINKED ISSUE #5005: [WIP] PHP Fatal error: Uncaught InvalidArgumentException: Unidentified action ... in ActionScheduler_DBStore.php on line 451

**Describe the bug**

We have reports of PHP Fatal errors like the following one:
```
PHP Fatal error: Uncaught InvalidArgumentException: Unidentified action 222898 in /home/path/wp-content/plugins/wp-rocket/inc/Dependencies/ActionScheduler/classes/data-stores/ActionScheduler_DBStore.php:451
Stack trace:
#0 /home/path/wp-content/plugins/wp-rocket/inc/Dependencies/ActionScheduler/functions.php(120): ActionScheduler_DBStore->cancel_action()
#1 /home/path/wp-content/plugins/wp-rocket/inc/Dependencies/ActionScheduler/functions.php(148): as_unschedule_action()
#2 /home/path/wp-content/plugins/wp-rocket/inc/Engine/Common/Queue/AbstractASQueue.php(152): as_unschedule_all_actions()
#3 /home/path/wp-content/plugins/wp-rocket/inc/Engine/Common/Queue/AbstractASQueue.php(61): WP_Rocket\Engine\Common\Queue\AbstractASQueue->cancel_all()
#4 /home/path/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/Controlle in /home/path/wp-content/plugins/wp-rocket/inc/Dependencies/ActionScheduler/classes/data-stores/ActionScheduler_DBStore.php on line 451
```

This is likely happening because the action we are trying to cancel doesn't exist.

**To Reproduce**
N/A

**Expected behavior**

No PHP fatal errors should appear in the log.

**Additional context**

**Tickets:**
https://secure.helpscout.net/conversation/1869827503/340859/
https://secure.helpscout.net/conversation/1869687348/340842
https://secure.helpscout.net/conversation/1871435728/341107/
https://secure.helpscout.net/conversation/1871414605/341105/

**Slack Convo:** https://wp-media.slack.com/archives/C43T1AYMQ/p1651558351479639

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

In this PR, we use the WP cron instead of Action Scheduler for the RUCSS pending jobs event to run every minute.

Some of the existing code has been moved to a new Cron Subscriber class to regroup code related to the cron events, and reduce complexity in other classes.

Fixes #5005

## Type of change

## How Has This Been Tested?

- [x] Automated tests

# Checklist:


inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
```diff
@@ -4,8 +4,7 @@
 namespace WP_Rocket\Engine\Optimization\RUCSS\Admin;
 
 use WP_Rocket\Engine\Admin\Settings\Settings as AdminSettings;
-use WP_Rocket\Engine\Common\Queue\QueueInterface;
-use WP_Rocket\Engine\Common\Queue\RUCSSQueueRunner;
+use WP_Rocket\Engine\Optimization\RUCSS\Controller\Queue;
 use WP_Rocket\Engine\Optimization\RUCSS\Controller\UsedCSS;
 use WP_Rocket\Event_Management\Subscriber_Interface;
 use WP_Rocket\Logger\Logger;
```
engahmeds3ed:
```
use WP_Rocket\Logger\Logger;
```
Not used anymore

engahmeds3ed (changes requested):
I think we need also to cancel the old pending Action Scheduler recurring action with update.

Also I believe we need to unschedule the CRONs with update also as maybe the CRON runs when we still upgrade the code and this may cause errors.

What do you think?

remyperona:
> Also I believe we need to unschedule the CRONs with update also as maybe the CRON runs when we still upgrade the code and this may cause errors.

The WP cron events can't be triggered during update, as when you start an update there is the maintenance mode enabled, which prevents access to the website .

engahmeds3ed:
@Tabrisrp Good point.

What about:
> I think we need also to cancel the old pending Action Scheduler recurring action with update.

not to have one hook that is fired from two locations, what do u think?

remyperona:
my only worry about that would be to trigger the DB error by doing this, what do you think?

engahmeds3ed:
Here: https://github.com/wp-media/wp-rocket/blob/22b9aafffe55920aa40017007169fd3a3ab0849c/inc/Dependencies/ActionScheduler/classes/data-stores/ActionScheduler_DBStore.php#L449-L452

when the `wpdb` can't update the action to make its status `CANCELLED` it throws this error.

With update, we need to call [this method](https://github.com/wp-media/wp-rocket/blob/b1d16b7bd8ddfb0a998330fea09cd3fd687fb92f/inc/Engine/Optimization/RUCSS/Controller/Queue.php#L40-L47) but add it inside `try catch` so we can catch such this error, what do u think @Tabrisrp ?

remyperona:
Looks good with the try/catch 👍🏼

engahmeds3ed (approved):

Mai-Saad:
@Tabrisrp Currently user may be stuck in the in-progress status till entries are manually deleted from database in the following scenario:
1- On the PR, while preload is enabled, enable RUCSS => make sure to have some pending status in the used CSS table
2- Delete the Actions table of AS => make sure to have some in-progress items in the used CSS table
3- In the admin dashboard, refresh then click on the URL in AS error msg => Actions table is recreated but empty
4- refresh admin or access FE => the in-progress status in used CSS won't change
Can you please check this case?
Screenshot from 2022-05-17 22-20-08

remyperona:
How do you delete the AS tables? Manually? This is an action that is not done by the plugin automatically in any case as far as I know

piotrbak:
Yes, if it's not done automatically, then we're fine and we can treat it as an edge case

Mai-Saad:
> How do you delete the AS tables? Manually? This is an action that is not done by the plugin automatically in any case as far as I know

@Tabrisrp Yes, it is deleted manually in the current scenario ( before there was a reported [issue](https://github.com/wp-media/wp-rocket/pull/4897), where AS was deleted for some reason and caused an error => not sure how this can be reproduced automatically though)

engahmeds3ed:
The only case we found with real customers is that when installing wpr the AS tables are not created (whole tables).
We have transient that has AS tables count, we can check on it before changing the status to in-progress.
But yes, it's very edge case.

Mai-Saad (approved):
The main error was not reproducible, and as no further changes are needed for the edge case then we are good here.
[testrail-report-386.pdf](https://github.com/wp-media/wp-rocket/files/8715105/testrail-report-386.pdf)

PR #5031, opened 2022-05-10
Closes #5018 RUCSS compatibility with jevelin theme

### LINKED ISSUE #5018: RUCSS compatibility with Jevelin theme

**Describe the bug**
We need a comptibility with Jevelin theme. When it's activated, dynamic inline stylesheets containing `#heading-` should be preserved in the source.

Jevelin code can be found for example [here](https://github.com/michaelnas84/renatastreit.com.br/trunk/wp-content/themes/jevelin).

**Screenshots**
image

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description

Added ThirdParty Class for RUCSS Compatibility with jevelin theme to preserve dynamic selectors

Fixes #5018 

## Type of change

# Checklist:


inc/ThirdParty/Themes/Jevelin.php
```diff
@@ -0,0 +1,51 @@
...
+	public function preserve_patterns( $patterns ): array {
+		if ( ! self::is_jevelin() ) {
+			return $patterns;
+		}
+
+		$preserve = [
+			'#heading-',
+		];
+
+		return array_merge( $patterns, $preserve );
```
remyperona:
```suggestion
		$patterns[] = '#heading-';

		return $patterns;
```

Better to avoid using array_merge when it's not absolutely necessary

remyperona (changes requested):

inc/ThirdParty/Themes/Jevelin.php
jeawhanlee: @Tabrisrp True, I gave this a thought but then I also thought about a scenario where we might be preserving more than just one. what do you think?

remyperona: We can always update the code later if needed 👍🏼

jeawhanlee: ok Awesome!

remyperona (approved):

Mai-Saad (commented):
Please find notes below:
- In order to see `#heading-` => used Jevelin theme and imported one of their demo sites
- `#heading-` is now preserved in the source, however, UI is still corrupted than?nowprocket
- The nowprocket version contains other dynamic styles like `'#divider-','#button-','#icon-','#partners-','#image-','#text-'` => adding those to patterns still didn't fix the UI issue
- Finally, that kind of exclusions can exist in different plugins/themes, so shouldn't we have like shared exclusion list instead of duplication in every theme?
- Tested on https://rocketlabsqa.ovh/
@piotrbak @Tabrisrp @jeawhanlee What do you think?

CrochetFeve0251:
@Mai-Saad  @piotrbak  For the moment i didn't found any dynamic css on that theme.
However I found with the help of @mostafa-hisham that a bug in the Saas was happening due to broken CSS inside the theme.
The error is inside the file `/wp-content/themes/jevelin/css/plugins.css` at the line 50

Removing the block 
```css
//HgNwwMDDbYTJuGQ8MHBgYGJ1xOYGNgYJiBpuEpAwODHSF/siDZ+ISBgcGClEDqZ2Bg8B6CkQsAPRga0cpRtDEAAAAASUVORK5CYII=);background-position:center center;width:16px;height:16px;display:block;position:absolute;top:0;left:0;right:0;bottom:0;margin:auto;background-repeat:no-repeat;opacity:.8}
```
Fixes the issue with the RUCSS.

CrochetFeve0251:
> @Mai-Saad @piotrbak For the moment i didn't found any dynamic css on that theme. However I found with the help of @mostafa-hisham that a bug in the Saas was happening due to broken CSS inside the theme. The error is inside the file `/wp-content/themes/jevelin/css/plugins.css` at the line 50
> 
> Removing the block
> 
> ```css
> //HgNwwMDDbYTJuGQ8MHBgYGJ1xOYGNgYJiBpuEpAwODHSF/siDZ+ISBgcGClEDqZ2Bg8B6CkQsAPRga0cpRtDEAAAAASUVORK5CYII=);background-position:center center;width:16px;height:16px;display:block;position:absolute;top:0;left:0;right:0;bottom:0;margin:auto;background-repeat:no-repeat;opacity:.8}
> ```
> 
> Fixes the issue with the RUCSS.

With @piotrbak we listed and added dynamic classes.

inc/ThirdParty/Themes/Jevelin.php
```diff
@@ -0,0 +1,84 @@
...
+		$patterns[] = '.sh-footer-builder-widgets-';
+		$patterns[] = '.sh-image-gallery-simple-';
+		$patterns[] = '.header-navigation-';
+		$patterns[] = '.sh-google-maps-';
+		$patterns[] = '.sh-element-titlebar-breadcrumbs-';
+		$patterns[] = '.sh-header-builder-';
+		$patterns[] = '.sh-text-group-';
+		$patterns[] = '.sh-footer-builder-title-';
+
+		return $patterns;
```
remyperona: maybe there with a big list like that we can do an array merge

inc/ThirdParty/Themes/Jevelin.php
```diff
@@ -0,0 +1,84 @@
...
+	 *
+	 * @since 3.11.3
+	 *
+	 * @param WP_Theme $theme Instance of the theme.
+	 */
+	private static function is_jevelin( $theme = null ) {
+		$theme = $theme instanceof \WP_Theme ? $theme : wp_get_theme();
+
+		return 'jevelin' === strtolower( $theme->get( 'Name' ) ) || 'jevelin' === strtolower( $theme->get_template() );
+	}
```
remyperona, line 86: Would be good to use the same method that was introduced in another PR to check for the current theme

remyperona (changes requested):

inc/ThirdParty/Themes/Jevelin.php
CrochetFeve0251: @Tabrisrp I searched it but look like we didn't merged the branch yet.

remyperona (approved):

vmanthos (approved):
The PR is working as expected. 👍 

**Note:**

Used CSS isn't generated correctly because of parse errors in the CSS of the theme (v5.0.2).
@NataliaDrause has reported these issues in the tree-shaker's repo:
* https://github.com/wp-media/nodejs-treeshaker/issues/73#issuecomment-1101278623
* https://github.com/wp-media/nodejs-treeshaker/issues/73#issuecomment-1102399902

PR #5043, opened 2022-05-12
Closes #4802 Use filesystem to store the used CSS content

### LINKED ISSUE #4802: RUCSS - the wpr_rucss_used_css table can grow big on large websites

**Describe the bug**

The  `wpr_rucss_used_css` table's size depends on the size of the used CSS for each page, and also the number of pages on a site.

On large websites that have a lot of URLs, it can grow out of proportion. In one case, that table's size reached `18.4Gb`.

After internal discussion, I'm creating this GitHub issue to monitor these cases, and see if there is anything we can do in the future.

**To Reproduce**
Not relevant.

**Additional context**

This is different from #4161 which will be resolved with the new implementation of the feature.

**Tickets:** 
https://secure.helpscout.net/conversation/1810416819/330445/
https://secure.helpscout.net/conversation/1657346732/300522
https://secure.helpscout.net/conversation/1670431889/303126
https://secure.helpscout.net/conversation/1665033380/301876
https://secure.helpscout.net/conversation/1645065559/297803

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

This PR updates the RUCSS code to store the used CSS content on the filesystem instead of in the Database, to avoid table size issues on sites with a lot of URLs or with low database storage available.

Fixes #4802

## Type of change

No

## How Has This Been Tested?

- [x] Automated tests
 
# Checklist:


inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
```diff
@@ -317,6 +318,8 @@ private function delete_used_css_rows() {
 			$this->database->truncate_used_css_table();
 		}
 
+		$this->used_css->delete_all_used_css();
```
engahmeds3ed: Here we will delete all files just after deleting the DB rows, so let's think of the case that we already have pending rows so we will delete them one by one so while deleting from DB and just before deleting the files from filesystem we may find a row that its status changed to completed so it'll be deleted from the filesystem but not from the DB, right?

inc/Engine/Optimization/RUCSS/Controller/Filesystem.php
```diff
@@ -0,0 +1,146 @@
...
+		 *
+		 * @param int $levels Number of levels.
+		 */
+		$levels = apply_filters( 'rocket_used_css_dir_level', 3 );
+
+		$base   = substr( $hash, 0, $levels );
+		$remain = substr( $hash, $levels );
+
+		$base_array = str_split( $base );
+		$path_array = array_merge( $base_array, [ $remain ] );
```
CrochetFeve0251: @Tabrisrp as I told you on Slack but I think you missed it, I did a mistake there and `[]=` operator would be better.
```suggestion
		$path_array = str_split( $base );
		$path_array[] = $remain;
```

CrochetFeve0251 (changes requested):

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -245,13 +254,19 @@ public function treeshake( string $html ): string {
...
 		}
 
-		if ( 'completed' !== $used_css->status || empty( $used_css->css ) ) {
+		if ( 'completed' !== $used_css->status || empty( $used_css->hash ) ) {
+			return $html;
+		}
+
+		$used_css_content = $this->filesystem->get_used_css( $used_css->hash );
+
+		if ( empty( $used_css_content ) ) {
```
engahmeds3ed, line 273: do we need to delete the DB row it as a safeguard to make all rows synced with the DB?

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
remyperona: We could start by deleting the files, then the DB?

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
remyperona: I think that's a good idea

inc/Engine/Optimization/RUCSS/Admin/Subscriber.php
```diff
@@ -753,6 +756,24 @@ public function delete_as_tables_transient_on_tools_page() {
...
+	 *
+	 * @since 3.11.3
+	 *
+	 * @param string $new_version New plugin version.
+	 * @param string $old_version Previous plugin version.
+	 *
+	 * @return void
+	 */
+	public function delete_used_css( $new_version, $old_version ) {
+		if ( version_compare( $old_version, '3.11.3', '>=' ) ) {
```
mostafa-hisham: i think we need to change 3.11.3 to 3.11.4

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -388,25 +422,26 @@ private function add_used_css_to_html( string $html, UsedCSS_Row $used_css ): st
...
-	private function get_used_css_markup( UsedCSS_Row $used_css ): string {
+	private function get_used_css_markup( string $used_css ): string {
 		/**
-		 * Filters Used CSS content before saving into DB.
+		 * Filters Used CSS content before output.
 		 *
 		 * @since 3.9.0.2
 		 *
-		 * @param string $usedcss Used CSS.
+		 * @param string $usedcss Used CSS content.
```
mostafa-hisham: param name should be `used_css`

Mai-Saad:
@Tabrisrp @piotrbak please find exploratory test notes below:
1- If we have 2 posts having the same Used CSS then we clear used CSS for URL 1 and Url 2, and the used CSS hash is removed from the table while the file is still there in the file system => shouldn't the used CSS file be deleted in this case?

2- When update/rollback currently will result in a mix of new approach + old approach i.e used CSS table will have some URLs with hash while others with used CSS => shouldn't truncate happen here and only have either hash or used CSS based on used version?
Screenshot from 2022-06-13 12-50-08

3- If the used CSS file is read-only, we don't display a notice to the user when enabling RUCSS => shouldn't notice be displayed?
Note: in this case, the hash is created in the used CSS table and even after permission are set, the corresponding files aren't added to the file system when accessing the FE => do we need to make modification here so either not to create a hash in this case or add files once permissions are set? or Maybe we need to make it clear in the doc that we need to clear used CSS after permissions are set, what do you think?

remyperona:
1. This is what should happen, but I noticed an error in the code that would prevent it from happening. I'm going to push a fix
2. Since we rollback to a previous version of the code, we can't do any modification at that point
3. We can add a notice if the `used-css` folder is not writable. We can also avoid adding the DB row if the file writing failed.

piotrbak:
@Tabrisrp the 3rd point sounds like a good enhancement.

remyperona:
@Mai-Saad code is updated for points 1 and 3

CrochetFeve0251 (approved):

Mai-Saad:
1. Is fixed now

3. for the 3rd point, notice is displayed as expected but has some qs here @Tabrisrp @piotrbak :
- if permissions were read-only for the cache folder, there are warning logs for caching inner folders except for used CSS, do we need to add logs there?
- if we access pages while read-only permission on used CSS, we still add entries in the used CSS table, is this ok or we wouldn't add entries in this case? (note: those pages may be stuck in in-progress state unless we  manually clear them from the database) while if we clear used CSS for URL in this case, the entry will be removed from the database while it is still there in file sys even after permissions are back => do we need to do anything here?

4. We need to handle clear file sys if not accessed for last month as we do in used CSS table  :pray: (currently only used CSS table is affected). @Tabrisrp what do you think?

5. When updating from trunk to the current branch, we aren't removing existing rows from the used CSS table, can you please check?

piotrbak:
@Tabrisrp I think that all points are valid. What's your opinion?

remyperona:
> if permissions were read-only for the cache folder, there are warning logs for caching inner folders except for used CSS, do we need to add logs there?

What do you mean by warning logs? An admin notice?

> if we access pages while read-only permission on used CSS, we still add entries in the used CSS table, is this ok or we wouldn't add entries in this case? 

I can make the status `failed` if we couldn't write the used CSS to the filesystem to avoid the status to be stuck `in-progress`

> while if we clear used CSS for URL in this case, the entry will be removed from the database while it is still there in file sys even after permissions are back

Can't really do anything about that at this point. If permissions are set correctly after, the old file will be overwritten by the new anyway

> We need to handle clear file sys if not accessed for last month as we do in used CSS table

This is difficult, we don't have a way to know which URLs have been removed by the DB query. Also a file could be shared by multiple URLs, some which might have been accessed, and some which might have not.

Mai-Saad (approved):
Working as expected
[testrail-report-396.pdf](https://github.com/wp-media/wp-rocket/files/8915879/testrail-report-396.pdf)

notes:
- We currently truncate all used CSS entries with the update action
- Testing with WP on Linux (not checked on other OS as env not available)

jeawhanlee:
Unit test failed after merging develop into branch to resolve conflicts. so I'll wait for @Tabrisrp

PR #5048, opened 2022-05-15
Added Umami to JS Minify exclusion list

### LINKED ISSUE #5024: Umami.js needs to be excluded from Minification by default
**Describe the bug**
When Minification is enabled, Umami.js (an analytics tracker, alternative to Google Analytics & Matomo) fails to run properly. I must exclude it on each site from WP Rocket Minification in order for it to work properly. Requesting that the umami.js script be excluded by default as this will surely affect all customers using Umami for data collection / website analytics.

Umami script looks something like this:

<script async defer data-website-id="<UUID>" src="https://<UmamiHostname>/umami.js"></script>

When minification is enabled, it of course adds the data-minify="1" to it too.

Can learn more about Umami here: https://umami.is/docs/collect-data

**To Reproduce**
Steps to reproduce the behaviour:
1. Add a Umami tracker script to a website2. 
2. Run WP Rocket to clear cache and enable Minification
3. Preload cache, and test if script is running (can confirm if analytics show up in Umami server)
4. You should see that no data is collected using the script.
5. Exclude umami.js from Minification, then clear and preload cache again and test - you should now see it works correctly

**Expected behaviour**
Umami analytics script should work without needing to modify WP Rocket, as it'd require every site under management by an admin to have to exclude that script, and any mistakes in missing it will be a loss of important website metrics.

**Screenshots**
N/A

**Additional context**
N/A

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

d19dotca:
Resolves #5024

## Description

Adds umami to the list of minify JS default exclusions.

Fixes #5024 

## Type of change

N/A

## How Has This Been Tested?

N/A

# Checklist:


remyperona (approved):

vmanthos:
Hi @d19dotca.

You've added the exclusion to the array that holds external files. Is the `umami.js` always going to be served from a URL that's different from that of the site? For example:
Site: www.example.com
`umami.js`: app.something.com/umami.js

If not, are there issues when `umami.js` is served from the same TLD as that of the site?

d19dotca:
Hi @vmanthos - Umami is a self-hosted [analytics server](https://umami.is) so it won't be always be included in the same domain as the website referencing it and at closest will be a different subdomain still. In my case, I run the server and have it on my client sites, so beyond my own website which will talk to Umami on a subdomain, the rest will be entirely different domains.

vmanthos:
Thank you for your reply, @d19dotca! 🙏

vmanthos (approved):

**Note:** If the `umami.js` file is served from the same TLD as that of the site, e.g. for `example.com` that would be `example.com/(.*)umami.js`, the specific exclusion won't be applied.

PR #5061, opened 2022-05-19
Closes #5003 on respecting site trailing slash

### LINKED ISSUE #5003: Respect site's trailing slash
**Describe the bug**
Closing https://github.com/wp-media/wp-rocket/issues/1392 in favour of this one, since we have  abetter view of what's needed and what risks are.

When serving the cache page using PHP, we're not respecting the `/` settings in permalinks. We should respect that and the solution will solve this problem except the scenarios when:
1. Rewrite rules are present and working
2. Server-side cache is being used

**Expected behavior**
Respect the trailing slash rules when serving cache using PHP.

**Additional context**
When serving the cache using PHP WordPress is still not loaded. It's possible that we'll need to use WP Rocket configuration file to store the data related to permalinks settings.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description

Closes issue on respect site trailing slash.

Fixes #5003 

## Type of change

There are some differences from the proposed.

## How Has This Been Tested?

Tested locally.

# Checklist:


inc/classes/Buffer/class-cache.php
```diff
@@ -669,4 +670,42 @@ private function maybe_dynamic_cookies_filename( $filename, $cookies ) {
...
+		// If permalink_structure is 'plain'.
+		$clean_request_uri_last_char = '' === $permalink_structure ? '' : $clean_request_uri_last_char;
+
+		// Bail out if permalink structure and url are the same.
+		if ( $permalink_last_char === $clean_request_uri_last_char ) {
+			return;
+		}
+
+		$protocol = $this->tests->is_ssl() ? 'https://' : 'http://';
+		$url      = $protocol . $host . rtrim( $clean_request_uri, '/' ) . $permalink_last_char . $this->tests->get_query_string();
```
remyperona: The clean request uri variable doesn't contain the full original URL requested

remyperona (changes requested):
It is unnecessary to store the permalink structure into the WP Rocket options array. In `get_rocket_config_file()`, we can directly use `get_option( 'permalink_structure' )` to store the current permalink structure.

And we can use the `permalink_structure_changed` [hook](https://developer.wordpress.org/reference/hooks/permalink_structure_changed/) to update the config file when the structure is changed.

This hook and callback can be added in a new `Engine\Cache\Config\Subscriber` class, instead of adding it to the `Settings` class which is unrelated.

The function `rocket_generate_config_file()` can be used to re-generate the config file.

inc/classes/Buffer/class-cache.php
```diff
@@ -669,4 +670,42 @@ private function maybe_dynamic_cookies_filename( $filename, $cookies ) {
...
+	 */
+	private function maybe_redirect_with_trailing_slash() {
+		$permalink_structure = $this->config->get_config( 'permalink_structure' );
+		$host                = $this->config->get_host();
+
+		// Last character of permalink.
+		$permalink_last_char = substr( $permalink_structure, -1 );
+
+		// Request uri without protocol & TLD.
+		$request_uri = $this->tests->get_raw_request_uri();
```
remyperona:
```suggestion
		$request_uri = $this->tests->get_request_uri_base();
```

remyperona (changes requested):

inc/classes/Buffer/class-cache.php
```diff
@@ -678,13 +678,13 @@ protected function reset_lowercase( $matches ) {
 	 */
 	private function maybe_redirect_with_trailing_slash() {
 		$permalink_structure = $this->config->get_config( 'permalink_structure' );
-		$host                = $this->config->get_host();
+		$host                = $this->config->get_host() . ':10003';
```
remyperona:
```suggestion
		$host                = $this->config->get_host();
```

remyperona (changes requested):

remyperona (approved):

vmanthos:
@jeawhanlee This is working as expected for URLs with paths, e.g. `example.com/hello-world/` but for the homepage, it causes a redirect loop when the permalinks don't have a trailing `/`:


In the following check:
https://github.com/wp-media/wp-rocket/blob/af547d77ef69a773c62c6113f38757c4983bff96/inc/classes/Buffer/class-cache.php#L701-L704

the `$request_uri_last_char` will always be `/` and the `$permalink_last_char` will be empty.

The two characters that are compared are never equal and we don't bail out and the redirect takes place every time causing the loop.

PS: Thanks to @Mai-Saad for bringing this to my attention.

vmanthos:
Also, the following scenarios are causing issues:

**Custom permalink without a trailing `/`**

1. Usage of a query string for which we create cache, e.g. `location` added to the **Cache Query strings** text area in the Advanced Rules tab.
2. Visiting the homepage `https://example.com/?location=Paris` causes a redirect to `https://example.comlocation=Paris` and that request is canceled/failed.
3. Visiting a page with a path `https://example.com/?location=Paris` results in `error 404`: `https://example.com/hello-worldlocation=`

**Custom permalink with a trailing `/`**

1. Usage of a query string for which we create cache, e.g. `location` added to the **Cache Query strings** text area in the Advanced Rules tab.
2. Visiting a page with a path `https://example.com/hello-world?location=Paris` causes a redirect to `https://example.com/hello-world/location=Paris` and this results in an `error 404`.

CrochetFeve0251:
@jeawhanlee looks like the query 

> Also, the following scenarios are causing issues:
> 
> **Custom permalink without a trailing `/`**
> 
>     1. Usage of a query string for which we create cache, e.g. `location` added to the **Cache Query strings** text area in the Advanced Rules tab.
> 
>     2. Visiting the homepage `https://example.com/?location=Paris` causes a redirect to `https://example.comlocation=Paris` and that request is canceled/failed.
> 
>     3. Visiting a page with a path `https://example.com/?location=Paris` results in `error 404`: `https://example.com/hello-worldlocation=`
> 
> 
> **Custom permalink with a trailing `/`**
> 
>     1. Usage of a query string for which we create cache, e.g. `location` added to the **Cache Query strings** text area in the Advanced Rules tab.
> 
>     2. Visiting a page with a path `https://example.com/hello-world?location=Paris` causes a redirect to `https://example.com/hello-world/location=Paris` and this results in an `error 404`.

Look like the method `$this->tests->get_query_string()` doesn't returns the `?` I think adding it before the call of the function will solve the problem in this comment.

jeawhanlee:
> Also, the following scenarios are causing issues:
> 
> **Custom permalink without a trailing `/`**
> 
> 1. Usage of a query string for which we create cache, e.g. `location` added to the **Cache Query strings** text area in the Advanced Rules tab.
> 2. Visiting the homepage `https://example.com/?location=Paris` causes a redirect to `https://example.comlocation=Paris` and that request is canceled/failed.
> 3. Visiting a page with a path `https://example.com/?location=Paris` results in `error 404`: `https://example.com/hello-worldlocation=`
> 
> **Custom permalink with a trailing `/`**
> 
> 1. Usage of a query string for which we create cache, e.g. `location` added to the **Cache Query strings** text area in the Advanced Rules tab.
> 2. Visiting a page with a path `https://example.com/hello-world?location=Paris` causes a redirect to `https://example.com/hello-world/location=Paris` and this results in an `error 404`.

Thanks @vmanthos will work on these fixes.

jeawhanlee:
> @jeawhanlee looks like the query
> 
> > Also, the following scenarios are causing issues:
> > **Custom permalink without a trailing `/`**
> > ```
> > 1. Usage of a query string for which we create cache, e.g. `location` added to the **Cache Query strings** text area in the Advanced Rules tab.
> > 
> > 2. Visiting the homepage `https://example.com/?location=Paris` causes a redirect to `https://example.comlocation=Paris` and that request is canceled/failed.
> > 
> > 3. Visiting a page with a path `https://example.com/?location=Paris` results in `error 404`: `https://example.com/hello-worldlocation=`
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **Custom permalink with a trailing `/`**
> > ```
> > 1. Usage of a query string for which we create cache, e.g. `location` added to the **Cache Query strings** text area in the Advanced Rules tab.
> > 
> > 2. Visiting a page with a path `https://example.com/hello-world?location=Paris` causes a redirect to `https://example.com/hello-world/location=Paris` and this results in an `error 404`.
> > ```
> 
> Look like the method `$this->tests->get_query_string()` doesn't returns the `?` I think adding it before the call of the function will solve the problem in this comment.

Correct @CrochetFeve0251

remyperona (approved):

vmanthos:
@jeawhanlee I tested the branch after the latest changes and found that the query strings redirects are working and the redirect loop is fixed. 👏

However, when the permalinks don't have a trailing `/`, visiting a URL that has one, e.g. `https://vasilis.rocketlabsqa.ovh/hello-world/` doesn't redirect to `https://vasilis.rocketlabsqa.ovh/hello-world` which would be the expected outcome. (https://youtu.be/VdfwV1PdOq8)

@piotrbak Should we consider anchor links as well when respecting the trailing `/`?
Currently, we aren't: https://youtu.be/wDKHP8fQZ-E

piotrbak:
@vmanthos I think we can skip redirecting anchors. We need to make sure that redirect loop won't happen there though

piotrbak:
@jeawhanlee @vmanthos What's the state of this PR? From what you mentioned, it still need some love 🤔 Is that right?
```
However, when the permalinks don't have a trailing /, visiting a URL that has one, e.g. https://vasilis.rocketlabsqa.ovh/hello-world/ doesn't redirect to https://vasilis.rocketlabsqa.ovh/hello-world which would be the expected outcome. (https://youtu.be/VdfwV1PdOq8)
```

jeawhanlee:
@piotrbak I was with @vmanthos on this last week and I found out the branch on his test site was not updated. So I guess he'll check this and revert when he has some time.

piotrbak:
Thank you!

vmanthos:
@piotrbak @jeawhanlee 

This is working as expected. Please find my notes below.

When testing [ignored query strings](https://docs.wp-rocket.me/article/971-caching-query-strings#ignored-parameters), e.g. `utm_source` and the following conditions apply:
* Permalinks include a trailing `/`
* The visited URL doesn't have a trailing slash:
`https://domain.con/hello-world?utm_source=1&location=Paris`

the redirection URL doesn't include the ignored parameter `utm_source`:
`https://domain.con/hello-world/?location=Paris`

When visiting the URL with a trailing slash `utm_source`  is preserved:
https://vasilis.rocketlabsqa.ovh/hello-world/?utm_source=1&location=Thessaloniki

This is happening with `trunk` as well. We should preserve the ignored parameters in case these are handled by JavaScript after the page loads.

This is a new issue. Do you want it handled in the current PR or do you want me to create a new GitHub issue?

piotrbak:
@vmanthos If that's not a regression, we can create an issue and see the feedback. I think we're fine then.

vmanthos (approved):
Working as expected!

PR #5066, opened 2022-05-20
Closes #5028 on RUCSS not rewriting prloaded fonts url to CDN

### LINKED ISSUE #5028: RUCSS - Preloaded fonts are not rewritten to the CDN
- Made sure you’re on the latest version ✅ `3.11.2`

**Describe the bug**

During tree-shaking we preload fonts that we find in the used CSS:
https://github.com/wp-media/wp-rocket/blob/3a5b9b11dbda004854d54a4f07dc4ddc95bf2dbc/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L254

If CDN rewriting is activated, we are not rewriting the URLs of the preloaded fonts. This leads to duplicate loading of font files:

Full-Scale image > https://i.vgy.me/IfVqhy.jpg

**To Reproduce**

1. Enable Remove Unused CSS.
2. Enable CDN rewriting.
3. Check the source code of the page. The preloaded fonts will not point to the CDN.

**Expected behavior**

Preloaded fonts should be rewritten to the CDN if that's activated.

**Additional context**

**Ticket:** https://secure.helpscout.net/conversation/1874123205/341463?folderId=2135277

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description

Added filter to rewrite fonts to cdn.

Fixes #5028 

## Type of change

Slightly different from the proposed.

## How Has This Been Tested?

Tested online
- Enable RUCCSS
- Enable CDN Rewriting
- Inspect element of page and see preloaded fonts url rewritten.

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -655,7 +655,7 @@ class UsedCSS {
 
 			$font_url = $this->extract_first_font( $font_face['content'] );
 
-			$font_url = apply_filters( 'rocket_asset_url', $font_url, [ 'all' ] );
+			$font_url = apply_filters( 'rocket_font_url', $font_url );
 
 			if ( empty( $font_url ) ) {
 				continue;
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -655,6 +655,8 @@ class UsedCSS {
 
 			$font_url = $this->extract_first_font( $font_face['content'] );
 
+			$font_url = apply_filters( 'rocket_asset_url', $font_url, [ 'all' ] );
+
 			if ( empty( $font_url ) ) {
 				continue;
 			}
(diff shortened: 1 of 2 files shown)


CrochetFeve0251:
@Tabrisrp looks like we are having a problem with the replacement due to a [bail out](https://github.com/wp-media/wp-rocket/blob/42cb23f01f5ee7def30b8827b5e77e7ae913b391/inc/Engine/CDN/Subscriber.php#L227) in the `maybe_replace_url` when the host from the url is equals to the one from the website.

Do you still remember why this coding is here? (I know it is two years ago :/)

remyperona:
The method `maybe_replace_url()` replaces the CDN URL by the site URL, not the other way around, so it can't be used for this.

You could create a new filter and hook `add_cdn_url()` on it instead for example.

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -655,6 +655,8 @@ private function add_used_fonts_preload( string $html, string $used_css ): strin
 
 			$font_url = $this->extract_first_font( $font_face['content'] );
 
+			$font_url = apply_filters( 'rocket_font_url', $font_url );
```
remyperona, line 665: Can you add inline doc for the filter, following this format: https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters

remyperona (approved):

vmanthos (approved):

Luckyazimie391 (approved):

PR #5077, opened 2022-05-25
Closes #5041 on rucss compatibility with inline related posts dynamic selectors

### LINKED ISSUE #5041: RUCSS - Inline Related Posts widget breaks due to dynamic inline selector

**Describe the bug**
The Inline Related Posts plugin adds inline style with a selector like this `.u7cb7434f55ddbc61829903e3a7b5fb85`. 
The numeric-letter value is dynamic and the layout breaks when cache is refreshed, but Used CSS is still referencing an old selector.

There is unfortunately no unique `content` we can trigger. The style tag doesn't have any `id` or `attribute`.

The inline style looks like this:
image
Screenshot: https://i.imgur.com/EtuGf2Q.png

`.ctaText` is probably the only `content` we could use, but it is not unique.

**To Reproduce**
1. Install and activate **Inline Related Posts** plugin;
2. Save settings for the plugin - can be used default;
3. Create a post and add Related Posts module with a block;
4. Enable Remove Unused CSS option;
5. Used CSS will include the initial dynamic selector.
6. After cache is cleared the selector will change and styling will be missing.

**Expected behavior**
The style for the Related Posts module needs to be preserved.

**Additional context**
Plugin: Inline Related Posts https://wordpress.org/plugins/intelly-related-posts/
80 000 installations

Ticket: https://secure.helpscout.net/conversation/1875506582/341695/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description

Preserve dynamic inline styles for inline related posts plugin.

Fixes #5041 

## Type of change

This solution is in no way different from the proposed.

## How Has This Been Tested?

- Install & Activate inline related posts.
- Activate RUCSS
- Created Identical Posts.
- Then inspect page to see that dynamic inline styles are preserved

# Checklist:

### DIFF
--- a/inc/ThirdParty/Plugins/InlineRelatedPosts.php
+++ b/inc/ThirdParty/Plugins/InlineRelatedPosts.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace WP_Rocket\ThirdParty\Plugins;
+
+use WP_Rocket\Event_Management\Subscriber_Interface;
+
+/**
+ * Subscriber for compatibility with Inline Related Posts.
+ */
+class InlineRelatedPosts implements Subscriber_Interface {
+
+
+	/**
+	 * Subscriber for Inline Related Posts.
+	 *
+	 * @return array
+	 */
+	public static function get_subscribed_events() {
+		if ( ! defined( 'IRP_PLUGIN_SLUG' ) ) {
+			return [];
+		}
+
+		$events['rocket_rucss_inline_content_exclusions'] = 'excluded_inline_from_rucss';
+
+		return $events;
+	}
+
+	/**
+	 * Exclude inline style from RUCSS.
+	 *
+	 * @param array $excluded excluded css.
+	 * @return array
+	 */
+	public function excluded_inline_from_rucss( $excluded ) {
+		$excluded[] = '.centered-text-area';
+		$excluded[] = '.ctaText';
+
+		return $excluded;
+	}
+}
--- a/inc/ThirdParty/Plugins/InlineRelatedPosts.php
+++ b/inc/ThirdParty/Plugins/InlineRelatedPosts.php
@@ -20,9 +20,7 @@ class InlineRelatedPosts implements Subscriber_Interface {
 			return [];
 		}
 
-		$events['rocket_rucss_inline_content_exclusions'] = 'exclude_inline_from_rucss';
-
-		return $events;
+		return [ 'rocket_rucss_inline_content_exclusions' => 'exclude_inline_from_rucss' ];
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


inc/ThirdParty/Plugins/InlineRelatedPosts.php
```diff
@@ -0,0 +1,40 @@
...
+	 * @return array
+	 */
+	public static function get_subscribed_events() {
+		if ( ! defined( 'IRP_PLUGIN_SLUG' ) ) {
+			return [];
+		}
+
+		$events['rocket_rucss_inline_content_exclusions'] = 'exclude_inline_from_rucss';
+
+		return $events;
```
remyperona:
```suggestion
		return [
		    'rocket_rucss_inline_content_exclusions' => 'exclude_inline_from_rucss',
		];
```

tests/Integration/inc/ThirdParty/Plugins/InlineRelatedPosts/excludeInlineFromRucss.php
```diff
@@ -0,0 +1,27 @@
...
+ * @covers  \WP_Rocket\ThirdParty\Plugins\InlineRelatedPosts::exclude_inline_from_rucss
+ * @group   InlineRelatedPosts
+ */
+class Test_ExcludeInlineFromRucss extends TestCase {
+
+	public function tear_down() {
+		remove_filter( 'rocket_rucss_inline_content_exclusions', [ $this, 'return_false' ] );
+
+		parent::tear_down();
+	}
```
remyperona: This is not necessary

tests/Integration/inc/ThirdParty/Plugins/InlineRelatedPosts/excludeInlineFromRucss.php
```diff
@@ -0,0 +1,27 @@
...
+
+	/**
+	 * @dataProvider configTestData
+	 */
+	public function testShouldReturnExpected( $config, $expected ) {
+        $inline_related_posts = new InlineRelatedPosts();
+		$this->assertSame( $expected, apply_filters( 'rocket_rucss_inline_content_exclusions', $inline_related_posts->exclude_inline_from_rucss($config['excluded'] ) ) );
```
remyperona: You need to either use the filter, or the method call, but not both at the same time in the integration test.

remyperona (changes requested):

remyperona (approved):

Mai-Saad (approved):
Working as expected

PR #5095, opened 2022-05-27
RUCSS: preserve inline style tags on the page with a filter and only delete the CSS content

### LINKED ISSUE #4969: Allow filtering the CSS files added with <link> before they are removed by RUCSS
In some cases, we might need to exclude CSS files from being removed by RUCSS. 
When for any reason, removing them causes issues and we cannot properly safelist them.
This might simplify the troubleshooting of some edge cases (safelisting doesn't work, CSS links are using media queries that we cannot cover, etc)

Add a filter to manipulate the array of stylesheet files we grab at [$link_styles](https://github.com/wp-media/wp-rocket/blob/151cb6257e69c6832c4354b3a09be1a32eb1c521/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L298-L302) 

**Additional context**
related ticket: https://secure.helpscout.net/conversation/1853433495/338231?folderId=2683093
slack thread: https://wp-media.slack.com/archives/C029G1B5HD2/p1650551460727419

### LINKED ISSUE #5064: RUCSS compatibility with the BoomBox theme dark mode
- [x] Made sure you’re on the latest version `3.11.2`

**Describe the bug**
BoomBox dark mode dynamically injects a stylesheet that overrides the light mode style. 

It's done using inline JavaScript that looks for a specific inline style ID: `boombox-primary-style-inline-css`

```
...
if ( !!getCookie( 'boombox-theme-mode' ) ) {

var appendInterval = setInterval( function() {
	if ( document.getElementById( 'boombox-primary-style-inline-css' ) ) {
		clearInterval( appendInterval );
		appendThemeModeLink();
	}
}, 25 );
}
...
</script>
```

When the script doesn't find the target selector, the dark mode CSS file won't be added resulting in the following console error: 

```
Uncaught TypeError: Cannot read properties of null (reading 'parentNode')
    at appendThemeModeLink (<anonymous>:37:16)
    at HTMLAnchorElement.<anonymous> (<anonymous>:57:9)
appendThemeModeLink @ VM281:37
(anonymous) @ VM281:57
```

Adding `boombox-primary-style-inline-css` to `rocket_rucss_inline_atts_exclusions` fixes the issue. 

We might consider a possible leaner approach by simply changing the target selector on the JavaScript code to our own `wpr-usedcss`. 

In this case, we will be hooking to the `rocket_buffer` and looking for: 
```
document.getElementById( 'boombox-primary-style-inline-css' )
```
and replacing it by: 
```
document.getElementById( 'wpr-usedcss' )
```

That way, we don't have to preserve the whole CSS.

**To Reproduce**
1. Use the BoomBox theme
2. Enable the dark mode + RUCSS
3. See error when trying to change to dark mode

**Expected behavior**
BoomBox dark mode should work out of the box when RUCSS is used.

**Screenshots**
N/A

**Additional context**
Ticket - https://secure.helpscout.net/conversation/1889250265/343784

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
## Description

Here we want to preserve the style tag with its attributes on the page when this filter is used as some plugins/themes are depending on the existence of those styles so instead of keeping the whole styles we want only to keep the style tags.

Fixes #5064
Fixes #4969

## Type of change

No, we combined both issues in one PR as both are related to each others.

## How Has This Been Tested?

- [ ] Use the following filters to be able to test
```
add_filter( 'rocket_rucss_external_exclusions', function ( $exclusions ){
	$exclusions[] = 'wp-includes/css/dist/block-library/style.min.css';//here u can add any string from the link style url or attributes like below.
	$exclusions[] = 'wc-blocks-vendors-style-css';

	return $exclusions;
} );

add_filter( 'rocket_rucss_preserve_inline_style_tags', function ( $excluded, $style ){
	//If you want, $style[0] has the full style tag so u can search for a thing on the style to decide which styles to preserve its style tag.

	return true;
}, 10, 2 );
```

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -53,6 +53,13 @@ class UsedCSS {
 	private $queue;
 
 	/**
+	 * External exclusions list, can be urls or attributes.
+	 *
+	 * @var array
+	 */
+	private $external_exclusions = [];
+
+	/**
 	 * Inline CSS attributes exclusions patterns to be preserved on the page after treeshaking.
 	 *
 	 * @var string[]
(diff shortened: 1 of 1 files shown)


remyperona (approved):

vmanthos (approved):
This is workings as expected. 👍 

**Notes:**
1. If the same inline style tag is preserved using the `rocket_rucss_inline_atts_exclusions` filter, the CSS selectors are preserved, i.e. that filter prevails over `rocket_rucss_preserve_inline_style_tags`. @piotrbak is this something we need to safeguard on our side?
2. With the `rocket_rucss_preserve_inline_style_tags` filter in place we may want to consider moving some of the following into a new array that will hold attributes that are handled by JavaScript, e.g. the Divi ones:
https://github.com/wp-media/wp-rocket/blob/c8727558936fc0c4262942a7e1faf16d067cb321/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L60-L69
3. For #5064 to be fixed we need to preserve `boombox-primary-style-inline-css` in a new array, a change in the existing PR would be required, or use a compatibility class. @piotrbak @engahmeds3ed What do you think?

engahmeds3ed:
Thanks @vmanthos 

1. Yes `rocket_rucss_inline_atts_exclusions` is happening before the other one so if we added inline style to both, we will keep the full style tag with the selectors, is this what is happening?
2. Good point, do u know which ones exactly?
3. Yes, will do that in this PR in a bit.

engahmeds3ed:
I'll create another PR for compatibility with boombox theme now

PR #5096, opened 2022-05-30
Closes #5085 on RUCSS compatibility with unlimited elements dynamic inline styles

### LINKED ISSUE #5085: RUCSS to preserve inline CSS for the Unlimited Elements plugin
- Made sure you’re on the latest version - 3.11.2

**Describe the bug**
I am creating this GH issue to collect all the dynamic selectors issues related to the [Unlimited Elements](https://nl.wordpress.org/plugins/unlimited-elements-for-elementor/) plugin. Possibly we can see a common pattern in these cases.

Current case has Post Carousel widget which styles are added with a dynamic id selector like `#uc_uc_card_post_carousel_elementor13280`:
image

Unlimited Elements plugin may have many other widgets that use dynamic selectors.

This case has a trigger: `/wp-content/plugins/unlimited-elements-for-elementor/assets_libraries/owl-carousel-new/owl.carousel.min.js`

Another case [here](https://github.com/wp-media/wp-rocket/issues/5083) is similar but has no file trigger, there is only inline script.

We had more reports of the dynamic selector here:
https://github.com/wp-media/wp-rocket/issues/4166#issuecomment-901855725
https://github.com/wp-media/wp-rocket/issues/4166#issuecomment-999334465

**To Reproduce**
Reproduced on customer's site that uses the widget.
Example in the ticket: https://secure.helpscout.net/conversation/1896704647/344944/#thread-5610986172

**Expected behavior**
Preserved dynamic styles.

**Additional context**
Issue is resolved by preserving `#uc_uc_card_post_carousel_elementor` with `rocket_rucss_inline_content_exclusions` filter.

Ticket: https://secure.helpscout.net/conversation/1896704647/344944/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description

Added new third party class for unlimited elements plugin to preserve dynamic inline styles in RUCSS

Fixes #5085 

## Type of change

## How Has This Been Tested?

- Install and activate unlimited elements.
- install any widget from the plugin and use on elementor.
- check page source to see that inline styles are preserved.

# Checklist:


remyperona (approved):

Mai-Saad (approved):
Working as expected

PR #5099, opened 2022-05-30
Closes #5038 Add dynamic lists for RUCSS exclusions

### LINKED ISSUE #5038: Backend safelist for RUCSS exclusions
It's a plugin part of the following issue:
https://github.com/wp-media/rucss-backend/issues/5

WP Rocket needs to fetch specific exclusions from the remote server. It's about `inline_atts_exclusions`, `inline_content_exclusions` arrays.

**Additional context**
We need to make sure:
1. Fallback is used when there's no connection between site and our servers
2. Make sure that the exclusions are updated when they were fetched from the servers and then the connection is lost
3. Easy way to refetch the data from server (cleaning transients?)

**EDIT**
@Tabrisrp is right, Jo prepared the epic card for it before, it can be seen here:
https://github.com/wp-media/wp-rocket/issues/3602

remyperona:
## Description

Add dynamic lists system for RUCSS exclusions.

Closes #5038

## Type of change

- [x] New feature

No

## How Has This Been Tested?

- [x] Automated tests

# Checklist:


inc/Engine/Optimization/DynamicLists/APIClient.php
```diff
@@ -0,0 +1,134 @@
...
+namespace WP_Rocket\Engine\Optimization\DynamicLists;
+
+use WP_Error;
+use WP_Rocket\Admin\Options_Data;
+
+class APIClient {
+	/**
+	 * API URL.
+	 */
+	const API_URL = 'https://b.rucss.wp-rocket.me/api/';
```
CrochetFeve0251, line 13: @Tabrisrp is it normal this endpoint is different from usual RUCSS endpoint?

inc/Engine/Optimization/DynamicLists/DataManager.php
```diff
@@ -0,0 +1,136 @@
...
+	private $filesystem;
+
+	/**
+	 * Instantiate the class
+	 *
+	 * @param WP_Filesystem_Direct $filesystem Filesystem instance.
+	 */
+	public function __construct( $filesystem = null ) {
+		$this->filesystem = is_null( $filesystem ) ? rocket_direct_filesystem() : $filesystem;
+	}
```
CrochetFeve0251, line 21: That wouldn't be cleaner to inject it from the Container?

inc/Engine/Optimization/DynamicLists/ServiceProvider.php
```diff
@@ -0,0 +1,44 @@
...
+	public function register() {
+		$this->getContainer()->add( 'dynamic_lists_data_manager', DataManager::class );
+		$this->getContainer()->add( 'dynamic_lists_api_client', APIClient::class )
+			->addArgument( $this->getContainer()->get( 'options' ) );
+		$this->getContainer()->add( 'dynamic_lists', DynamicLists::class )
+			->addArgument( $this->getContainer()->get( 'dynamic_lists_api_client' ) )
+			->addArgument( $this->getContainer()->get( 'dynamic_lists_data_manager' ) )
+			->addArgument( $this->getContainer()->get( 'template_path' ) );
+		$this->getContainer()->share( 'dynamic_lists_subscriber', Subscriber::class )
+			->addArgument( $this->getContainer()->get( 'dynamic_lists' ) );
```
CrochetFeve0251: We made the change from string to ::class in the ServiceProvider? (If yes I need to add an issue for uniformising the code base)

CrochetFeve0251 (changes requested):

inc/Engine/Optimization/DynamicLists/APIClient.php
remyperona: @mostafa-hisham Is this the prod URL for the endpoint or the test one?

inc/Engine/Optimization/DynamicLists/DataManager.php
remyperona: It would, but the way our testing suite is set up right now, It's easier to have this configured that way for testing

inc/Engine/Optimization/DynamicLists/ServiceProvider.php
remyperona: It does look cleaner and easier to read that way, we could update the other service providers when we touch them

inc/Engine/Optimization/DynamicLists/APIClient.php
mostafa-hisham: porduction

CrochetFeve0251 (approved):

inc/Engine/Optimization/DynamicLists/DataManager.php
```diff
@@ -0,0 +1,136 @@
...
+
+		return $result;
+	}
+
+	/**
+	 * Gets the path to the dynamic lists JSON file
+	 *
+	 * @return string
+	 */
+	private function get_json_filepath(): string {
```
mostafa-hisham, line 86: @Tabrisrp 
if we are going to use GitHub actions to update the dynamic-lists.json we need to 
- setup the GitHub action
- change the path of the file to be in the plugin directory
- remove fallback file stuff
- update related tests

remyperona: The dynamic list file in the plugin directory is the fallback one. We're still going to store and update the one in `wp-rocket-config` separately on a weekly basis with the cron.

What we're missing in this PR for now is the GH action

vmanthos (commented):
This is working in accordance with what's described in the [Dynamic Inclusion and Exclusion Lists For WP Rocket](https://docs.google.com/document/d/1LCzqwCduwlu2xlddZ4o_A6kZhVf83jDUCYq4pM900L4/edit#) except for these:
1. the cache isn't cleared when the dynamic lists are updated
2. the lists can be updated for expired licenses - this should be handled server-side

Notes:
1. Updating from a version `<=3.11.5` and:
* the `/wp-rocket-config/` folder isn't writeable
* the `wpr_dynamic_lists` transient is deleted
no exclusions will be there until the `rocket_update_dynamic_lists` is run which would be in one week.
2. If the fallback `dynamic-lists.json` has been removed and external requests are blocked, clicking the **UPDATE LISTS** doesn't display any error.
3. When we cannot write to the `/wp-rocket-config/dynamic-lists.json` file, e.g. because of permissions/ownership we are displaying the generic message `Could not update lists.` Maybe change the message to something more descriptive like:
> We could not update the dynamic lists because we cannot write in the `/wp-content/wp-rocket-config/dynamic-lists.json` file. Check [the doc](https://docs.wp-rocket.me/article/626-how-to-make-system-files-htaccess-wp-config-writeable)* for troubleshooting steps.

 \* The doc will have to be updated or a new doc will be written. 
4. The `wpr_dynamic_lists` transient isn't deleted when uninstalling WP Rocket.
5. Currently, the exclusion categories aren't prefixed, e.g. for Remove Unused CSS inline attribute exclusions it's `inline_atts_exclusions`. Should these be prefixed, e.g. `rucss_inline_atts_exclusions`? 🤔

remyperona:
I pushed an update for:
- the clean cache
-  the license expiration check
- the transient deletion on uninstall

for note (3), we probably should check all messages to improve them. @piotrbak 

Current messages:
- license expired: `You need an active license to get the latest version of the lists from our server.`
- Unexpected return code from server: `Could not get updated lists from server.`
- No changes on the lists: `Lists are up to date.`
- Could not write file: `Could not update lists.`
- Success: `Lists are successfully updated.`

For note (5) I agree, but this needs to be changed on the server side. @mostafa-hisham

mostafa-hisham:
@Tabrisrp 
- I updated the backend to add `rucss_` as a prefix
- Also update this branch to use the new rucss exclusion names with the `rucss_` prefix

**FYI**: couldn't run integration tests i am having [problems](https://wp-media.slack.com/archives/GUT7FLHF1/p1661853241303329?thread_ts=1661788099.614609&cid=GUT7FLHF1) with integration tests on my machine, but Unit testing works fine and it was successful

vmanthos (approved):
This is working as expected.

About point 1 of [my comment](https://github.com/wp-media/wp-rocket/pull/5099/#pullrequestreview-1086772270) is this a case we'd like to address @piotrbak?

**Notes**

This has been tested using only Remove Unused CSS given that none of the rest inclusions/exclusions have been implemented on the SaaS yet.

piotrbak:
@vmanthos Regarding the 1st point, I think we're good here.

PR #5101, opened 2022-05-31
Closes #5065 on RUCSS wrong home url detection with wpml

### LINKED ISSUE #5065: RUCSS wrong home_url detection with WPML enabled

**Describe the bug**
With RUCSS we have SaaS dedicated queue for home pages so we need to detect if the current page is the home page and send this info to SaaS to operate it quickly.
We send this info in two requests:-
1. POST request to add the url to the queue and get the job details here: https://github.com/wp-media/wp-rocket/blob/3a5b9b11dbda004854d54a4f07dc4ddc95bf2dbc/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L215-L222
2. GET request to get the job/url details here: https://github.com/wp-media/wp-rocket/blob/3a5b9b11dbda004854d54a4f07dc4ddc95bf2dbc/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L500

When WPML is enabled, let's assume that we have two languages (EN/DE) with the following urls:-
**Default EN home:** https://example.org/
**DE home:** https://example.org/de

The problem will happen with the `DE home`, so when visiting this page for the first time, `home_url()` will return true as the page is opened and WPML hooks are applied on it so this condition will return true https://github.com/wp-media/wp-rocket/blob/3a5b9b11dbda004854d54a4f07dc4ddc95bf2dbc/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L434

But with the GET request, this check returns `false` and to make it works we need to use the following filter
```
apply_filters( 'wpml_home_url', home_url() )
```

**To Reproduce**
1. Activate WPML
4. Add two languages as mentioned above
5. Visit the secondary language home page.
6. This page will not have the used_css and will reach the failed status after three retries

**Expected behavior**
Used CSS should be available for this page.

**Screenshots**

**Additional context**

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

jeawhanlee:
## Description

This PR fixes the issue of RUCSS not generating used css for secondary homepage with wpml and also creates a thirdparty class for wpml.

Fixes #5065 

## Type of change

## How Has This Been Tested?

Followed steps in issue submission but this time used css is generated for secondary homepage.

# Checklist:


inc/3rd-party/plugins/i18n/wpml.php
```diff
@@ -1,11 +0,0 @@
-<?php
```
remyperona, line 1: Need to also remove the `require` for this file in `inc/3rd-party/3rd-party.php`

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -439,7 +439,15 @@ private function is_mobile(): bool {
...
-		return untrailingslashit( $url ) === untrailingslashit( home_url() );
+		/**
+		 * Filters the home url.
+		 *
+		 * @since 3.11.4
+		 *
+		 * @param bool  $is_home boolean return if current url is_home.
+		 * @param string  $url url of current page.
+		 */
+		return apply_filters( 'rocket_rucss_is_home_url', ( untrailingslashit( $url ) === untrailingslashit( home_url() ) ), $url );
```
remyperona: We should just filter the `home_url()` value here, and then apply the rest of the logic. Something like:
```
$home_url = apply_filters( 'rocket_rucss_home_url', home_url() );

return untrailingslashit( $url ) === untrailingslashit( $home_url );
```

remyperona: We should just filter the `home_url()` value here, and then apply the rest of the logic. Something like:
```
$home_url = apply_filters( 'rocket_rucss_home_url', home_url() );

return untrailingslashit( $url ) === untrailingslashit( $home_url );
```

inc/ThirdParty/Plugins/I18n/WPML.php
```diff
@@ -0,0 +1,50 @@
...
+	 *
+	 * @return array
+	 */
+	public static function get_subscribed_events() {
+		if ( ! defined( 'ICL_SITEPRESS_VERSION' ) ) {
+			return [];
+		}
+
+		$events = [
+			'wcml_is_cache_enabled_for_switching_currency' => '__return_true',
```
remyperona:
```suggestion
			'wcml_is_cache_enabled_for_switching_currency' => 'return_true',
```

You can't use `__return_true` directly here, but you can use the above by also using the `ReturnTypesTrait` available

remyperona:
```suggestion
			'wcml_is_cache_enabled_for_switching_currency' => 'return_true',
```

You can't use `__return_true` directly here, but you can use the above by also using the `ReturnTypesTrait` available

inc/ThirdParty/Plugins/I18n/WPML.php
```diff
@@ -0,0 +1,50 @@
...
+		 * Filters the home url.
+		 *
+		 * @since 3.11.4
+		 *
+		 * @param string  $url url of homepage.
+		 */
+		$home_url = apply_filters( 'wpml_home_url', home_url() ); // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals
+
+		return untrailingslashit( $url ) === untrailingslashit( $home_url );
+	}
```
remyperona: This needs to be changed to match with the suggested changes in the rucss controller

remyperona: This needs to be changed to match with the suggested changes in the rucss controller

remyperona (changes requested):

remyperona (changes requested):
I don't understand the latest change to the method in the compatibility file, can you explain what it does and why use this instead of `apply_filters( 'wpml_home_url', home_url() );`

jeawhanlee:
@Tabrisrp When I implemented the changes requested I found out that the the used css for the secondary homepage (example.org/es) wasn't still generated then I thought it must have been the changes I made then I rolled back to the initial state of the PR and it still didn't work.

I then found out at this point when the `wpml_home_url` filter is applied it returns as expected which is appending the language code to the `home_url()`
https://github.com/wp-media/wp-rocket/blob/95d626766fe72bdb1a49f39070bd9846598181d9/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L222

but here when we try to get the job status it returns just the home url, so I think this is happening probably because there is no real visit to the secondary home page either organic or inorganically then it returns just the home url.
https://github.com/wp-media/wp-rocket/blob/95d626766fe72bdb1a49f39070bd9846598181d9/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L508
so everything comes back to the way it was again..
Then I approached it differently by using the `wpml_language_is_active` filter.
I get the current url to be processed and I get the second segment of the url and pass it as the language code here
https://github.com/wp-media/wp-rocket/blob/ce4c916625f2a4e9adf514756059ff5236177597/inc/ThirdParty/Plugins/I18n/WPML.php#L65
so for example if we have a current url to be processed as example.org/es we check the last segment `es` against the `wpml_language_is_active` filter and if it returns true, then we know that is a valid language for the site and the url to be processed is a secondary homepage then return that url back so as to also match with the initial request in the treeshake method.

We return the home url early if url to be processed has no segment which means it is the default homepage and we also return the home url early if the expected second segment of the url is not a valid language code. from what I see WPML accommodates up to 7 chars for language code.
https://github.com/wp-media/wp-rocket/blob/ce4c916625f2a4e9adf514756059ff5236177597/inc/ThirdParty/Plugins/I18n/WPML.php#L45-L55

remyperona:
So `wpml_home_url` is not returning the expected URL when used to check the job status?

The issue with the new code is it expect the language to be added in the path, but that's not the only possible way, it can be a query string or a sub-domain too.

jeawhanlee:
> So `wpml_home_url` is not returning the expected URL when used to check the job status?

Yes it doesn't

> The issue with the new code is it expect the language to be added in the path, but that's not the only possible way, it can be a query string or a sub-domain too.

Can you please share instances that can generate both the query string and the subdomain so I can modify @Tabrisrp @engahmeds3ed 
🙏

engahmeds3ed:
I'm checking that now with @jeawhanlee

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -439,7 +439,16 @@ private function is_mobile(): bool {
...
 	 */
 	private function is_home( string $url ): bool {
-		return untrailingslashit( $url ) === untrailingslashit( home_url() );
+		/**
+		 * Filters the home url.
+		 *
+		 * @since 3.11.4
+		 *
+		 * @param bool  $home_url home url.
+		 * @param string  $url url of current page.
```
```suggestion
		 * @param string  $home_url home url.
		 * @param string  $url url of current page.
```

inc/ThirdParty/Plugins/I18n/WPML.php
```diff
@@ -0,0 +1,58 @@
...
+	 *
+	 * @param string $home_url home url.
+	 * @param string $url url of current page.
+	 * @return string
+	 */
+	public function is_secondary_home( string $home_url, string $url ): string {
+		global $sitepress;
+
+		// Get active languages on site.
+		$languages = $sitepress->get_active_languages();
```
remyperona, line 43: can the `get_active_languages()` method return something other than an array?

remyperona (changes requested):

inc/ThirdParty/Plugins/I18n/WPML.php
jeawhanlee: @Tabrisrp from what I see on the WPML core, it will return only an array. like something below:

```
Array
(
 [en] => Array
  (
   [id] => 1
   [active] => 1
   [default_locale] => en_US
   [native_name] => English
   [missing] => 0
   [translated_name] => English
   [language_code] => en
   [country_flag_url] => http://yourdomain/wpmlpath/res/flags/en.png
   [url] => http://yourdomain/about
  )

 [fr] => Array
  (
   [id] => 4
   [active] => 0
   [default_locale] => fr_FR
   [native_name] => Français
   [missing] => 0
   [translated_name] => French
   [language_code] => fr
   [country_flag_url] => http://yourdomain/wpmlpath/res/flags/fr.png
   [url] => http://yourdomain/fr/a-propos
  )

 [it] => Array
  (
   [id] => 27
   [active] => 0
   [default_locale] => it_IT
   [native_name] => Italiano
   [missing] => 0
   [translated_name] => Italian
   [language_code] => it
   [country_flag_url] => http://yourdomain/wpmlpath/res/flags/it.png
   [url] => http://yourdomain/it/circa
  )
)
```

remyperona (approved):

vmanthos (approved):
Working as expected.

PR #5116, opened 2022-06-06
Closes #5076 Preserve RUCSS clean rows event if RUCSS option is disabled

### LINKED ISSUE #5076: Clear RUCSS after a while when the feature is deactivated

Currently, the only auto-clearing mechanism we have is `rocket_rucss_clean_rows_time_event` which we turn off after the feature is deactivated. I assume the same logic will remain when we move to the filesystem (data will still be kept in that scenario as well).

We are not clearing RUCSS data to make sure the Used CSS will still be there if the feature is re-used after being deactivated. 

We don't cover the scenario where the feature will be turned off long-term or if the customer decides to stop using it (low CSS on original files, seeing that CPCSS is enough, issues that will need time to be fixed).

We could have a background process that prompts a message after let's say 3 days to confirm why the feature was deactivated:
- If it's temporary, we can ask them to open a ticket if there is an issue or need help
- If they don't need/feel like using the feature, we can ask them to clear RUCSS stored data (database/filesystem), and maybe collect their reasons

Alternatively, we can keep the `rocket_rucss_clean_rows_time_event` alive when the RUCSS is disabled by: 
- Adding conditions to check if the database still has Used CSS on it.
- Shortening the `delete_old_items()` life cycle (set to 1 month) as it makes sense to get clear the data sooner. A Week for example.

N/A

**Additional context**
Slack convo - https://wp-media.slack.com/archives/G01AMD70N7P/p1653459048671429

remyperona:
## Description

Preserve the RUCSS clean rows event if RUCSS option is disabled, to avoid stale data in the database.

Fixes #5076 

## Type of change

## How Has This Been Tested?

- [x] Automated tests

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Cron/Subscriber.php
+++ b/inc/Engine/Optimization/RUCSS/Cron/Subscriber.php
@@ -61,16 +61,6 @@ class Subscriber implements Subscriber_Interface {
 	 * @return void
 	 */
 	public function schedule_clean_not_commonly_used_rows() {
-		if (
-			! $this->used_css->is_enabled()
-			&&
-			wp_next_scheduled( 'rocket_rucss_clean_rows_time_event' )
-		) {
-			wp_clear_scheduled_hook( 'rocket_rucss_clean_rows_time_event' );
-
-			return;
-		}
-
 		if ( ! $this->used_css->is_enabled() ) {
 			return;
 		}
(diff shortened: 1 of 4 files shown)


mostafa-hisham (commented):
@Tabrisrp should we clear the scheduled event if there are no more rows to delete?

remyperona:
@mostafa-hisham I think it would complexify the code too much compared to the benefit

mostafa-hisham (approved):

piotrbak (approved):
It's working as expected. Not attaching the TR as TestRail is not working now.

PR #5143, opened 2022-06-14
Closes #5126 - Fixes background image breaking for lazyload with native lazyload

### LINKED ISSUE #5126: Usage of rocket_use_native_lazyload filter breaks some images when LazyLoad is enabled
- Made sure you’re on the latest version - 3.11.3

**Describe the bug**
In some cases when our filter `rocket_use_native_lazyload` is used and our LazyLoad is enabled, we apply double lazy-load on images.
The attribute `loading=lazy` is present as well as LazyLoad attributes.

Screenshot from the customer's site:
image

This breaks some images:
image

and some cases are already mentioned here: https://github.com/wp-media/wp-rocket/issues/4961

**To Reproduce**
It was reproduced on a customer's site where theme was using `rocket_use_native_lazyload` filter and LazyLoad for images was enabled. 
I was not able to reproduce the issue on my test site.

**Expected behavior**
Images should not break. There should not be a duplicative lazy-load applied to images.

**Additional context**
Slack thread with more details and screenshots: https://wp-media.slack.com/archives/C43T1AYMQ/p1654777415067329
Ticket: https://secure.helpscout.net/conversation/1913047242/347171/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description

This PR fixes the issue of adding rocket lazyload attributes on images that lazyloaded natively.

Fixes #5126

## Type of change

There is no grooming.

## How Has This Been Tested?
- Add add_filter( 'rocket_use_native_lazyload', '__return_true' ) to current theme functions.
- Add an element to a template with inline background-image style and check that image displays when lazyload is applied.

# Checklist:


CrochetFeve0251 (approved):

remyperona (changes requested):
This is not the actual issue, and we don't want to prevent JS lazyload being applied when the loading attribute is present. I added a comment on the related issue to explain.

CrochetFeve0251 (approved):

inc/Dependencies/RocketLazyload/Assets.php
```diff
@@ -100,7 +100,28 @@ class_loaded: "lazyloaded",
...
+                    if ( element.tagName === "IFRAME" && element.dataset.rocketLazyload == "fitvidscompatible" ) {
+                        if (element.classList.contains("lazyloaded") ) {
+                            if (typeof window.jQuery != "undefined") {
+                                if (jQuery.fn.fitVids) {
+                                    jQuery(element).parent().fitVids();
+                                }
+                            }
+                        }
+                    }
+                }
```
remyperona: it's not necessary for the background images

inc/Dependencies/RocketLazyload/Assets.php
```diff
@@ -100,7 +100,28 @@ class_loaded: "lazyloaded",
 			$script = rtrim( $script, ',' );
 		}
 
-		$script .= '};';
+		$script .= '
+		},
+		{
+			elements_selector: ".rocket-lazyload",
```
remyperona: The value for the selector should be the one passed to the method as part of the `$args` instead of being hardcoded

remyperona (changes requested):

remyperona (approved):

remyperona:
@jeawhanlee can you also backport this change to https://github.com/wp-media/rocket-lazyload-common

vmanthos (approved):
Working as expected.

PR #5157, opened 2022-06-17
Close #3444 Added code to delete params files

### LINKED ISSUE #3444: Query string version of the homepage is not cleared with rocket_clean_home

**Describe the bug**
When `rocket_clean_home` is triggered, the cache of the query string is not removed.

**To Reproduce**
1. Add a query string to cache in the settings
2. Cache the homepage with query string `domain.com/?test=123`
3. Make sure that it's cached on the SFTP
4. Update any post to trigger the `rocket_clean_home`
5. Confirm on SFTP that the homepage is cleared but the `#test=123` is not

**Expected behavior**
We should clear also the query strings versions, including the cache and ignored query strings. Probably just looking for directories starting with `#` in the root cache directory and passing them to `rrmdir` will be fine.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description

In this PR we fix a problem in the cache cleaning where params cache from homepage are not cleared when a post is updated.
This has been fixed by adding a check on folders starting with # and if some exist we delete them.

Fixes #3444

## Type of change

No.

## How Has This Been Tested?

- [ ] Manual Test

# Checklist:

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -628,6 +628,14 @@ function rocket_clean_home( $lang = '' ) {
 		}
 	}
 
+	$param_dirs = glob( $root . '/#*', GLOB_NOSORT );
+
+	if ( $param_dirs ) {
+		foreach ( $param_dirs as $dir ) {
+			rocket_rrmdir( $dir );
+		}
+	}
+
 	// Remove the hidden empty file for mobile detection on NGINX with the Rocket NGINX configuration.
 	$nginx_mobile_detect_files = glob( $root . '/.mobile-active', GLOB_NOSORT );
 	if ( $nginx_mobile_detect_files ) {


remyperona (approved):

Mai-Saad:
@CrochetFeve0251 Thanks for the PR. Can see the qs cache for home is deleted with cache qs, dedicated cache qs when edit home, adding a new post, updating a post, deleting a post,  clear the cache of this URL in page list or purge this URL from page view. But clear used CSS for this URL will clear the cache of home but not the home qs cache. I think it shall be deleted in this case too @piotrbak @CrochetFeve0251 what do you think?

piotrbak:
Yes, it should be included.

CrochetFeve0251:
Closed in favor of #6234. (I moved the modifications to the new code as I spotted some problems due to old code)

PR #5166, opened 2022-06-21
Closes #4105 Allow usage of Monolog v2 while using v1 for deploy

### LINKED ISSUE #4105: Upgrade Monolog to version 2

jeawhanlee:
## Description

This PR closes the issue on monolog version update.

Fixes #4105

## Type of change

### DIFF
--- a/.github/workflows/deploy_plugin.yml
+++ b/.github/workflows/deploy_plugin.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '7.4'
+          php-version: '7.1'
           tools: composer:v2
 
       - name: Get composer cache directory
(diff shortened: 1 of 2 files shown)


CrochetFeve0251:
@Tabrisrp I am wondering why we don't use Mozart on Monolog to prevent incompatibilities?

CrochetFeve0251 (approved):

remyperona:
@CrochetFeve0251 There is too much classes in Monolog, and we use only a small subset, that would really pollute our repository.

PR #5167, opened 2022-06-21
Guard RUCSS filter returns against array of null

engahmeds3ed:
## Description

Guarding RUCSS `preg_quote` first argument in three places against being not string.

## Type of change

Not groomed, just got a warning from @Mai-Saad that this may return fatal errors.

## How Has This Been Tested?

```
add_filter( 'rocket_rucss_external_exclusions', function ( $exclusions ){
        $exclusions[] = null;
	return $exclusions;
} );
```

OR

```
add_filter( 'rocket_rucss_inline_atts_exclusions', function ( $exclusions ){
        $exclusions[] = null;
	return $exclusions;
} );
```

OR

```
add_filter( 'rocket_rucss_inline_content_exclusions', function ( $exclusions ){
        $exclusions[] = null;
	return $exclusions;
} );
```

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -365,6 +365,10 @@ class UsedCSS {
 
 		$external_exclusions = (array) array_map(
 			function ( $item ) {
+				if ( ! is_string( $item ) ) {
+					return $item;
+				}
+
 				return preg_quote( $item, '/' );
 			},
 			/**
@@ -415,6 +419,10 @@ class UsedCSS {
 
 		$inline_atts_exclusions = (array) array_map(
 			function ( $item ) {
+				if ( ! is_string( $item ) ) {
+					return $item;
+				}
+
 				return preg_quote( $item, '/' );
 			},
 			/**
@@ -429,6 +437,10 @@ class UsedCSS {
 
 		$inline_content_exclusions = (array) array_map(
 			function ( $item ) {
+				if ( ! is_string( $item ) ) {
+					return $item;
+				}
+
 				return preg_quote( $item, '/' );
 			},
 			/**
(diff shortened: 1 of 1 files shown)


remyperona (commented):
Maybe we could remove items that are not string before passing it to array_map, using array_filter() with is_string?

With the current solution, values that are not string are still preserved in the array

engahmeds3ed:
> Maybe we could remove items that are not string before passing it to array_map, using array_filter() with is_string?
> 
> With the current solution, values that are not string are still preserved in the array

I was thinking of that, u r totally right, will adjust it to be like that.

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -927,4 +918,23 @@ public function notice_write_permissions() {
...
+	 *
+	 * @param array $items Array to be validated and quoted.
+	 *
+	 * @return array|string[]
+	 */
+	private function validate_array_and_quote( array $items ) {
+		$items_array = array_filter( $items, 'is_string' );
+
+		return array_map(
+			function ( $item ) {
```
CrochetFeve0251: You can consider using a static anonymous function as you don't need the context ^^'
```suggestion
			static function ( $item ) {
```

CrochetFeve0251 (changes requested):

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
engahmeds3ed: Good point, Yes this is [micro-optimization](https://www.designcise.com/web/tutorial/what-are-static-anonymous-functions-in-php#when-to-use-static-closures) so why not.
Thanks.

remyperona (approved):

CrochetFeve0251 (approved):

Mai-Saad (approved):
Working as expected

PR #5171, opened 2022-06-21
Closes #5169 - Hide related RUCSS notices on denied folder permission

### LINKED ISSUE #5169: RUCSS progress notices are displayed even when the used-css folder cannot be created
- Made sure you’re on the latest version ✅`3.11.4`

**Describe the bug**

When permissions aren't correct and we can't create the `used-css` folder we display the "countdown" notice:

and then the "success" notice:

although the actual `status` is `failed`: 

**To Reproduce**

1. Install WP Rocket `>=3.11.4`.
2. Change the permissions of the `cache` folder to `444`.
3. Activate Remove Unused CSS.
4. See the notices being displayed.

**Expected behavior**

We shouldn't display the two notices when we cannot create/write to the `used-css` folder.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description

This PR removes the related RUCSS notices when we have a denied permission on the `cache` folder.

Fixes #5169 

## Type of change

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Admin/Settings.php
+++ b/inc/Engine/Optimization/RUCSS/Admin/Settings.php
@@ -181,7 +181,7 @@ class Settings {
 	 * @return void
 	 */
 	public function display_processing_notice() {
-		if ( ! rocket_direct_filesystem()->is_writable( WP_ROCKET_CACHE_PATH ) ) {
+		if ( ! rocket_direct_filesystem()->is_writable( rocket_get_constant( 'WP_ROCKET_CACHE_PATH' ) ) ) {
 			return;
 		}
 
@@ -227,7 +227,7 @@ class Settings {
 	 * @return void
 	 */
 	public function display_success_notice() {
-		if ( ! rocket_direct_filesystem()->is_writable( WP_ROCKET_CACHE_PATH ) ) {
+		if ( ! rocket_direct_filesystem()->is_writable( rocket_get_constant( 'WP_ROCKET_CACHE_PATH' ) ) ) {
 			return;
 		}
 
--- a/inc/Engine/Optimization/RUCSS/Admin/Settings.php
+++ b/inc/Engine/Optimization/RUCSS/Admin/Settings.php
@@ -181,10 +181,6 @@ class Settings {
 	 * @return void
 	 */
 	public function display_processing_notice() {
-		if ( ! rocket_direct_filesystem()->is_writable( rocket_get_constant( 'WP_ROCKET_CACHE_PATH' ) ) ) {
-			return;
-		}
-
 		if ( ! $this->can_display_notice() ) {
 			return;
 		}
@@ -227,10 +223,6 @@ class Settings {
 	 * @return void
 	 */
 	public function display_success_notice() {
-		if ( ! rocket_direct_filesystem()->is_writable( rocket_get_constant( 'WP_ROCKET_CACHE_PATH' ) ) ) {
-			return;
-		}
-
 		if ( ! $this->can_display_notice() ) {
 			return;
 		}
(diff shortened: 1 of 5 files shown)


inc/Engine/Optimization/RUCSS/Admin/Settings.php
```diff
@@ -181,6 +181,10 @@ public function set_optimize_css_delivery_method_value( $field_args ): array {
 	 * @return void
 	 */
 	public function display_processing_notice() {
+		if ( ! rocket_direct_filesystem()->is_writable( WP_ROCKET_CACHE_PATH ) ) {
```
remyperona:
```suggestion
		if ( ! rocket_direct_filesystem()->is_writable( rocket_get_constant( 'WP_ROCKET_USED_CSS_PATH', '' ) ) ) {
```

inc/Engine/Optimization/RUCSS/Admin/Settings.php
```diff
@@ -223,6 +227,10 @@ public function display_processing_notice() {
 	 * @return void
 	 */
 	public function display_success_notice() {
+		if ( ! rocket_direct_filesystem()->is_writable( WP_ROCKET_CACHE_PATH ) ) {
```
remyperona:
```suggestion
		if ( ! rocket_direct_filesystem()->is_writable( rocket_get_constant( 'WP_ROCKET_USED_CSS_PATH', '' ) ) ) {
```

remyperona (changes requested):

remyperona:
The tests need to be updated too

inc/Engine/Optimization/RUCSS/Admin/Settings.php
engahmeds3ed: Also why not adding this logic inside the method `can_display_notice` after checking the screen? what do u think?

jeawhanlee: True Sir

remyperona (approved):

vmanthos (approved):
This is working as expected.

The notices won't be displayed if we cannot create the `used-css` folder. 

After restoring the folder's permissions the "processing" notice gets displayed:
```
WP Rocket: Please wait 10 seconds. The Remove Unused CSS service is processing your pages.
```

If `rocket_rucss_pending_jobs` is executed before the issue is resolved, any preexisting jobs will have a `status` of `failed` and will require manually clearing of the used CSS or wait for anything that will trigger automatic clearing, for the used CSS to be generated for those URLs.

@piotrbak Maybe we can stop the creation of used CSS jobs _until_ we can write to the disk. Also, can we consider using a cron to clear and re-try `failed` jobs?

piotrbak:
@vmanthos I like the idea of `stopping creation of used CSS jobs until we can write to the disk.`

I believe it needs some discussion on how to implement this properly and when bail out. Could you create an issue?

vmanthos:
@piotrbak I'll create the issue tomorrow.

vmanthos:
@piotrbak @jeawhanlee When `cache` is writeable but `used-css` isn't we are still displaying the notices.
This is likely an edge case but could be resolved if, as @Tabrisrp suggested, we check if `WP_ROCKET_USED_CSS_PATH` is writeable instead of `WP_ROCKET_CACHE_PATH`.

piotrbak:
@vmanthos @jeawhanlee yes, we should check the Used CSS directory

vmanthos:
@jeawhanlee This is working as expected. 🙏

vmanthos (approved):

PR #5172, opened 2022-06-21
Closes #3837 Notices for expired users with optimize CSS delivery option

### LINKED ISSUE #3837: Display notices to expired users using the remove unused CSS option
### What are we proposing to do? 
With this enhancement, we will notify expired users who have enabled the RUCSS feature that they need an active license to continue using the service.

We will leave the option enabled and working until 90 days after expiration date. Once this time has passed, the feature will be automatically deactivated.

### Why are we doing this?
The RUCSS feature generates server costs that we cannot handle for users who don’t use a valid license. This is still GPL compliant as RUCSS is an external service.

### Specifications
Full doc: https://docs.google.com/document/d/18ZZTohcN7YAjgU6jN37ZDPAreO6WPAQeRCM8MMKNKug/edit?usp=sharing

remyperona:
## Description

Display custom notices for expired users with optimize CSS delivery option enabled.

Fixes #3837

## Type of change

## How Has This Been Tested?

- [x] Manual testing
- [x] Automated tests

# Checklist:


assets/css/wpr-admin.css
```diff
@@ -1 +1 @@
...
...r-multiple{display:flex;align-items:center;flex-wrap:wrap}@media(max-width: 783px){.wpr-multiple{align-items:center;flex-direction:column}}.wpr-multiple .wpr-text{flex:1 1 auto;position:relative;top:-2px}@media(max-width: 783px){.wpr-multiple .wpr-text{width:100%}}.wpr-multiple .wpr-button{margin-left:16px;width:auto;min-width:inherit;padding-right:30px}@media(max-width: 783px){.wpr-multiple .wpr-button{margin-left:0}}.wpr-multiple input[type=text]{flex-grow:2}.wpr-multiple select{height:30px}.wpr-multiple-default{margin-right:20px}.wpr-multiple-list{display:none;padding:8px 0;margin:16px 0 0;background:#f2f3f6;border-radius:2px}.wpr-multiple-list li{margin-bottom:0;padding:4px 16px;font-size:0.8125rem;line-height:1.2307692308;font-family:Monaco}.wpr-multiple-list li span{display:inline-block;transition:all 150ms ease-out;-webkit-transition:all 150ms ease-out}.wpr-multiple-close{position:relative;top:3px;font-size:1rem;line-height:1;transition:color 200ms ease-out;-webkit-transition:color 200ms ease-out}.wpr-multiple-close:focus{outline:none}.wpr-multiple-close:hover,.wpr-multiple-close:focus{color:#d33f49}.wpr-multiple-close:hover+span,.wpr-multiple-close:focus+span{color:#d33f49}
```
engahmeds3ed: image

I can see the title divided into two rows, we faced this issue before so what do u think about it?

inc/Engine/License/Renewal.php
```diff
@@ -85,7 +96,43 @@ public function display_renewal_expired_banner() {
 			return;
 		}
 
-		echo $this->generate( 'renewal-expired-banner', [ 'renewal_url' => $this->user->get_renewal_url() ] ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+		$expiration    = $this->user->get_license_expiration();
+		$expired_since = time() - $expiration;
```
engahmeds3ed: we can divide this value by `DAY_IN_SECONDS` so in the conditions below we can use the numbers directly without the calculation inside the condition, what do u think?

inc/Engine/License/views/renewal-expired-banner-ocd-disabled.php
```diff
@@ -0,0 +1,42 @@
+<?php
+/**
+ * Renewal expired banner.
+ *
+ * @since 3.7.5
```
engahmeds3ed:
```suggestion
 * @since 3.11.5
```

inc/Engine/License/views/renewal-expired-banner-ocd.php
```diff
@@ -0,0 +1,52 @@
+<?php
+/**
+ * Renewal expired banner.
+ *
+ * @since 3.7.5
```
engahmeds3ed:
```suggestion
 * @since 3.11.5
```

engahmeds3ed (changes requested):
Small comments

engahmeds3ed (approved):

vmanthos (commented):
@Tabrisrp 

I found that Optimize CSS Delivery is still working for expired users even when the feature has been disabled in the UI. 
Also, the option is available for 15 days after the license has expired. In the original issue, the acceptance criterion was 90 days.

In all of the following cases where license had expired:
* more than 15 days but less than 30 days ago 
* more than 30 days ago
* more than 180 days ago

the Optimize CSS Delivery feature was disabled in the UI:


however, it was actually working. I could clear the used CSS, regenerate it, etc.

Also, the UI elements in the admin bar menu were displayed and were functional both for Remove Unused CSS and Load CSS Asynchronously depending on which of the two was activated:


I'm guessing we should add a check about whether a user's license is expired, e.g. in `UsedCSS::is_allowed()` for Remove Unused CSS and do something similar for Load CSS Asynchronously.

PS: The fixtures in the tests are using `extra_days => 90`. Should these be changed to 15 days?

remyperona:
@vmanthos You're right, I updated the code to also disable the option completely when expired.

In the current doc, it says 15 days for the delay, I think it was updated in between compared to the moment the issue itself was published.

vmanthos:
@Tabrisrp Optimize CSS Delivery/Load CSS Asynchronously still works for expired users after the latest changes and despite the fact that `maybe_disable_option()` `return 0;`.

Could this be an issue of the `priority` `maybe_disable_option` is hooked to `pre_get_rocket_option_remove_unused_css` and `pre_get_rocket_option_async_css`? 🤔

remyperona:
@vmanthos I updated the priority to the max for both, as we don't want anything to override it anyway

vmanthos:
@Tabrisrp Now the feature is disabled for _every_ user with an `expired license`, even for those whole license has expired less than 15 days ago.

I think we should check for that in the `maybe_disable_option()` method:
```PHP
$expired_since = ( time() - $this->user->get_license_expiration() ) / DAY_IN_SECONDS;

		if ( 15 > $expired_since ) {
			return $value;
		}

```

Also, when selecting _Optimize CSS Delivery > Load CSS Asynchronously_ we display the notice about the CPCSS generation progress although that doesn't take place:


Another issue is that users who:
1. have **autorenewal enabled**
2. their **license has expired**, e.g. because the payment didn't get through ([Slack discussion](https://wp-media.slack.com/archives/C08N8J6VC/p1658389747598749) with @adinaWP 🙏 )

will be able to "check" the checkbox because the `wpr-isDisabled` class isn't added: 
https://youtu.be/znmdNc229Eg

They shouldn't be able to do that.

Can you please look into these?

remyperona:
@vmanthos I updated the code for the first 2 points.

For the renewal part, I'm wondering if we can consider this ok, as the auto-renew will go through at some point anyway?

vmanthos:
@Tabrisrp Sometimes customers never renew their license so they will be stuck in this state.

@piotrbak What do you think about this?

piotrbak:
@vmanthos `Sometimes customers never renew their license so they will be stuck in this state.`
It means that they won't renew and we should block it, right?

Could you confirm that we respect the 15 days of delay for the auto-renewals that didn't go through?

vmanthos:
Thank you, Remy!

Both of the following issues are fixed:
* Optimize CSS Delivery is disabled even for licenses that have expired less than 15 days ago
* Load CSS Asynchronously notice is displayed even though the feature is disabled

> Could you confirm that we respect the 15 days of delay for the auto-renewals that didn't go through?

@piotrbak Yes, Optimize CSS Delivery is working fine for users with:
* a license that has expired less than 15 days ago `AND`
* have autorenewal enabled

piotrbak:
@Tabrisrp Is it a big effort to solve the followign scenario?
1. User is set to autorenewal + expired for more than 15 days
2. He's able to tick the checkbox and save settings
3. The page is reloaded with option disabled

We should not allow users to tick it.

vmanthos:
Another small issue is the lack of a linking doc when the white label feature is enabled:
`define ('WP_ROCKET_WHITE_LABEL_ACCOUNT', true);`
**screencast:** https://youtu.be/4PwJNWPwY0E

This is from here:
https://github.com/wp-media/wp-rocket/blob/791bcbe4dd4d15d51a9b0a3d2d0a929f6e73a6ae/inc/Engine/License/Renewal.php#L421

We do offer info about the reason we disable Optimize CSS Delivery for expired users but that's in banners in WP Rocket's dashboard. When white-labeling those won't be available.

Writing a doc about what led to this decision may come in handy for support as well. For example, to share with customers who got the warning that they won't be able to use the feature soon (the license has expired less than 15 days ago) or for those where it has already been disabled. It can be useful for those who will search our documentation before reaching out to support.

If that's OK with @wp-media/productrocket, @webtrainingwheels can you please create such a doc?

GeekPress:
The 15 days window is for all users, not only for auto-renewal. The check of auto-renewal turned on isn't necessary.

GeekPress:
> If that's OK with https://github.com/orgs/wp-media/teams/productrocket, @webtrainingwheels can you please create such a doc?

I will suggest keeping the default doc we are using without the white label.

Why?
In general, white label is used by agencies that don't want to share their license info with their clients. Having the default doc to provide them the same info will be enough.

vmanthos:
@GeekPress 

> I will suggest keeping the default doc we are using without the white label.

Without the white label the text is "Renew now" and clicking on that takes the user to our checkout page so they can renew the license.

With white label enabled, the text is "More info" and the URL is not pointing to any doc. Clicking there just refreshes the page. You can see this here: https://www.youtube.com/watch?v=4PwJNWPwY0E

We will have to remove the link, or point it to an existing doc or create a new one. What will it be?

GeekPress:
Ah ok. Don't we have any existing doc where we are talking about a valid license is necessary to use RUCSS?

vmanthos:
> Don't we have any existing doc where we are talking about a valid license is necessary to use RUCSS?

We don't have a specific doc about this. 

~~Maybe instead of creating a new one, we can add the info in the [Basic requirements](https://docs.wp-rocket.me/article/1529-remove-unused-css#basic-requirements) section of the main Remove Unused CSS doc.~~

**Edit:**
Piotr reminded me this is also about CPCSS, so creating a new doc should be the way to go.

remyperona:
I removed the auto-renew check from the code. Once we have the doc link, I'll add it to finalize the PR

piotrbak:
@Tabrisrp We'll have the doc here:
https://docs.wp-rocket.me/article/1711-what-happens-if-my-license-expires

PR #5200, opened 2022-06-29
For RUCSS check zlib library before encoding/decoding the used CSS content

engahmeds3ed:
## Description

Here we are guarding our RUCSS filesystem code if the host disabled zlib or both functions `gzdecode` and `gzencode` are disabled.

## Type of change

## How Has This Been Tested?

I'm discussing with @piotrbak now how will we test it but seems like we will put the following line in php.ini and see how it goes
```
disable_functions="gzdecode, gzencode"
```

# Checklist:


CrochetFeve0251 (approved):

Mai-Saad (approved):
Working as expected for the happy case (creating used CSS, deleting used CSS)
Note: When switching from with zlib to without 
- if we enable preload after switching from having Zlib  to not having zlib, the used CSS in the table will be deleted while the entry is still there in the file sys
- if we accessed a cached page that has used CSS before, Used CSS will be deleted from table when access the page and won't be created till we manually clear cache

PR #5209, opened 2022-07-01
Closes #5207 on redirect loop with WP on sub-dir

### LINKED ISSUE #5207: Potential redirect loop when using 3.11.4, trailing slash redirection
- Made sure you’re on the latest version - 3.11.4

**Describe the bug**
When permalinks are set to be without `/` and website is in the subdirectory we might experience redirect loop. 

**To Reproduce**
1. Use website placed in subdirectory (could be multisite)
2. Change the permalinks to be without `/`
3. Visit frontend in incognito

**Expected behavior**
No redirect loop

**Additional context**
Still not sure if there are any other factors needed to reproduce.

**Temporary solution**
Change the permalinks settings to be with `/`

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description

This PR fixes the issue with a redirect loop on sub-dirs

Fixes #5207 

## Type of change

## How Has This Been Tested?

- Tested on TLD and Sub-dir installation on Apache.
- Tested only on TLD installation and multilanguage on NGINX.

# Checklist:

### DIFF
--- a/inc/classes/Buffer/class-cache.php
+++ b/inc/classes/Buffer/class-cache.php
@@ -80,7 +80,7 @@ class Cache extends Abstract_Buffer {
 
 		// Full request uri path.
 		$request_uri = $this->config->get_server_input( 'REQUEST_URI' );
-
+		$request_uri = strtok($request_uri, '?') [0];
 		// Path to WP index processing file and remove index.php from path.
 		$php_self = str_replace( 'index.php', '', $this->config->get_server_input( 'PHP_SELF' ) );
 
--- a/inc/classes/Buffer/class-cache.php
+++ b/inc/classes/Buffer/class-cache.php
@@ -701,6 +701,11 @@ class Cache extends Abstract_Buffer {
 	 */
 	private function maybe_redirect_with_trailing_slash() {
 
+
+		if(rocket_get_constant('MULTISITE', false) && ! rocket_get_constant('SUBDOMAIN_INSTALL',  true) ) {
+			return;
+		}
+
 		$permalink_structure = $this->config->get_config( 'permalink_structure' );
 		$host                = $this->config->get_host();
 		$protocol            = $this->tests->is_ssl() ? 'https://' : 'http://';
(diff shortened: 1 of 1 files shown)


CrochetFeve0251 (approved):

piotrbak:
I can see this solution failing for subdirectory multisite on the sites that are not main. Example:
https://subdir.rocketlabsqa.ovh/test2/

Moreover, we should take care of not redirecting 404 errors of static assets. Example:
https://domain.com/idontexist.txt

It should not be redirected to: https://domain.com/idontexist.txt/

CrochetFeve0251 (approved):

remyperona (approved):

vmanthos:
@jeawhanlee 

While testing the PR I found and @Mai-Saad confirmed that there is a **redirect loop** for the homepage on a **Single site** installed in a **subfolder** where the permalinks don't have a trailing `/` and the URL includes:
* a cached `OR`
* an ignored query string `OR` 
* a combination of a cached and ignored query string

inc/classes/Buffer/class-cache.php
```diff
@@ -677,15 +700,33 @@ protected function reset_lowercase( $matches ) {
 	 * @return void
 	 */
 	private function maybe_redirect_with_trailing_slash() {
+
+		if ( defined( 'MULTISITE' ) && true === MULTISITE ) {
```
CrochetFeve0251, line 704: Based on @piotrbak on Slack we need also too check `SUBDOMAIN_INSTALL` is set to false

CrochetFeve0251 (changes requested):

inc/classes/Buffer/class-cache.php
piotrbak: We seem to be fine on subdomains. We should let QA confirm thag

CrochetFeve0251 (approved):

piotrbak:
@jeawhanlee @CrochetFeve0251 this one can be closed in favour of https://github.com/wp-media/wp-rocket/pull/5215 right?

PR #5214, opened 2022-07-04
Fixes #5213 Improve removal of inline style content when using RUCSS

### LINKED ISSUE #5213: RUCSS - Page breaks when it includes <style></style> element that contains only a space
- Made sure you’re on the latest version ✅ `3.11.4`

**Describe the bug**

In `3.11.4` we preserve `<style></style` elements while removing their content:
https://github.com/wp-media/wp-rocket/blob/391bc8b0a00a30bc036454c0336c45a8378e99c1/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L452-L456

When the source code of a page contains such an element:
```HTML
<style> </style>
```

we are removing every space in the buffer and this breaks the page.

**To Reproduce**

1. Add a `<style> </style>` to a page.
2. Enable Remove Unused CSS.
3. When used CSS is ready for a URL, visit it on the front-end.
4. The error will be broken.

**Expected behavior**

Even if `<style></style>` contains only a space Remove Unused CSS should be applied without breaking the page.

**Additional context**

**Related ticket:** https://secure.helpscout.net/conversation/1939185938/352141?folderId=273761

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Only remove inline style content if it's not empty, and only do it on the current style instead of the whole HTML.

Fixes #5213

## Type of change

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -450,7 +450,14 @@ class UsedCSS {
 			 * @param array $style Full match style tag.
 			 */
 			if ( apply_filters( 'rocket_rucss_preserve_inline_style_tags', true, $style ) ) {
-				$html = str_replace( $style['content'], '', $html );
+				$content = trim( $style['content'] );
+
+				if ( empty( $content ) ) {
+					continue;
+				}
+
+				$empty_tag = str_replace( $style['content'], '', $style[0] );
+				$html      = str_replace( $style[0], $empty_tag, $html );
 
 				continue;
 			}


engahmeds3ed (approved):
Awesome approach, will stop interference and not needed replacements.

vmanthos (approved):
This fixes the issue.

**Note**

When using:
```HTML
<style id="space"> </style>
```
the space wasn't removed:


which is expected the way this is coded:
https://github.com/wp-media/wp-rocket/blob/fb0dad50455ed4f4b9c77c5dece1ea97d5be551b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L453-L457

Issue #3477, opened 2021-01-12
Multisite - Cache Query String(s) & other features not working on subsites


vmanthos:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version ✅
- Used the search feature to ensure that the bug hasn’t been reported before ✅

**Describe the bug**

**Cache Query String(s)**, as well as any other option we save in the configuration file in `/wp-content/wp-rocket-config/` will not work on multisite.

Changes made on a subsite will be written in the subsite-specific configuration file, e.g. `example.com.subsite.php`.

However, the configuration file that we are using is that of the main site:
https://github.com/wp-media/wp-rocket/blob/df26acdfbe015d056fc442cfbf28734f9635aa9f/inc/classes/Buffer/class-tests.php#L192

which doesn't have the changes applied to the subsite.

**To Reproduce**
1. Go to MEGA/Multisite.
2. On a subsite add a query string in the **Cache Query String(s)** text area in the **Advanced Rules** tab.
3. Visit a URL after appending the query string from step 2.
4. Check the source code of the page. It won't be cached/optimized.

**Expected behavior**

The features should work when they are set up on subsites.

**Additional context**

A workaround is to use the main site's advanced rules tab to add the query strings that need to be cached.
However, this can be problematic if the specific behavior is only needed on subsites.

**Related ticket:** https://secure.helpscout.net/conversation/1393512095/230531?folderId=2135277

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

piotrbak:
@vmanthos @arunbasillal This is related to https://github.com/wp-media/wp-rocket/issues/1424

Since your issue is using the template, I guess we can close the above one.

arunbasillal:
Notes from #1424 

> When WP Rocket is active on the main site (main network site), exclude pages from being cached feature is not working on subsites (excluded pages are being cached).

> After deactivating WP-Rocket from the main site, everything is working fine on each subsite.

> This is related to below ticket:
https://secure.helpscout.net/conversation/716301944/87882?folderId=2415573#thread-2009772070

> Made some further tests. Blogs are fetching the content of the main site WP Rocket config file instead of their own.

> Thanks to Alfonso we have our own Multisite environment where I was able to recreate the problem.

> Related ticket:
[...]

PR #5215, opened 2022-07-05
Closes #5207 on redirect loop 

### LINKED ISSUE #5003: Respect site's trailing slash

### LINKED ISSUE #5207: Potential redirect loop when using 3.11.4, trailing slash redirection

jeawhanlee:
## Description

This PR will fix the redirect loop by redirecting with wordpress when permalink settings and url do not match.

Fixes #5207

## Type of change

# Checklist:


remyperona (approved):

Mai-Saad:
@jeawhanlee Thanks for the update.
Please check tabs **NGINX round 2** and **Apache** [here](https://docs.google.com/spreadsheets/d/1TPCAXKSg1fDKWN8hE0LpS37DYWkboOi3N8anFyQ5iHY/edit?usp=sharing), we will need to handle red cells. Orange cell(s) needs to be discussed with @piotrbak to see if anything is needed there. 

Notes: 
- Checking on Apache multisite with subdomain is out of scope till test env is set up.
- For enforcing/remove trailing slash plugins, we already have [bug](https://github.com/wp-media/wp-rocket/issues/5310) on the trunk which shall be fixed before checking it with the current PR

jeawhanlee:
@Mai-Saad @piotrbak From my findings on Apache, I believe single site test cases fails after first visit of the page because the page cache is no longer served via php but `.htaccess` is at play already so the cache files are served using the rewrite rules.
- However Clearing cache from the admin or manually deleting the gzip file in the cache folder or the entire related cache folder makes the page served again via php and the redirect is triggered once again.
according to @piotrbak and @vmanthos we don't need to handle this case if `.htacess` is at work.
@Mai-Saad Can you please confirm the redirections are ok when the cache is cleared from admin. 🙏

Also for failed test cases on Muti-site with sub-folder setup for both Apache & Nginx, when this PR went through QA, Multi-site with sub-folder setup were by-passed so we don't redirect. but having removed the by-pass in the recent commits, Main site works fine but sub-sites fail, I think we might need to handle this from the area of our config file. As it is, we are not loading the corresponding config file for sub-sites of Multi-sites set up in a sub-folder as discussed with @vmanthos 

@piotrbak Can you please check the orange cells [here](https://docs.google.com/spreadsheets/d/1TPCAXKSg1fDKWN8hE0LpS37DYWkboOi3N8anFyQ5iHY/edit?usp=sharing) in the Apache tabs if we need to handle them 🙏

piotrbak:
@jeawhanlee You can enable Separate Cache for Mobile Devices to remove rewrite rules from htaccess

piotrbak:
@jeawhanlee Are those red cells still failing? There are lots of them 🤔

inc/classes/Buffer/class-cache.php
```diff
@@ -669,4 +673,39 @@ private function maybe_dynamic_cookies_filename( $filename, $cookies ) {
...
+			// Check for root installation.
+			$request_uri_last_char = '/' === $request_uri ? '' : $request_uri_last_char;
+
+			// Check for subdir installation.
+			if ( '/' !== $request_uri && $php_self === $request_uri ) {
+				$request_uri_last_char = '';
+			}
+		}
+
+		return $permalink_last_char === $request_uri_last_char;
```
Kilbourne: Probably you should return the inverse boolean ( and remove the negation on line81) to be consistent with the function name: currently when _maybe_allow_wp_redirect_ returns true the redirect doesn't happen, while when  it returns false, it redirects

piotrbak: @wp-media/php What do you think about this suggestion?

jeawhanlee: Good Point @Kilbourne. Thanks🙏

CrochetFeve0251 (approved):

inc/classes/Buffer/class-cache.php
```diff
@@ -706,6 +706,6 @@ private function maybe_allow_wp_redirect(): bool {
 			}
 		}
 
-		return $permalink_last_char === $request_uri_last_char;
+		return ! $permalink_last_char === $request_uri_last_char;
```
Kilbourne: Hi, here there's a bug: you are coercing $permalink_last_char to a boolean and then inverting it, instead of checking if the variables are different(`$permalink_last_char !== $request_uri_last_char;`). In current state it returns false also when the variables are different.
@jeawhanlee  @piotrbak

inc/classes/Buffer/class-cache.php
```diff
@@ -674,8 +678,7 @@ protected function reset_lowercase( $matches ) {
 		return strtolower( $matches[0] );
 	}
 
-	/**
-	 * Sanitizes a string key.
+	/** Sanitizes a string key.
```
remyperona: why was this changed?

remyperona (changes requested):

inc/classes/Buffer/class-cache.php
jeawhanlee: I think that was a mistake

remyperona (approved):

inc/classes/Buffer/class-cache.php
jeawhanlee: Thanks once again @Kilbourne

vmanthos:
@jeawhanlee While doing QA we found that on multisites where:
1. the subsite is in a subdirectory `&&`
2. the permalinks there don't have a trailing `/`

caching/optimizations are not applied. This is because `maybe_allow_wp_redirect()` returns here:
https://github.com/wp-media/wp-rocket/blob/aa66a000de1602f0e1d2b1a0587c5e0d879e1276/inc/classes/Buffer/class-cache.php#L79

Logging the variables there, I got these:
```php
permalink_structure: /blog/%postname%
permalink_last_char: 
request_uri: /multisite/green/
request_uri_last_char: /
request_uri_last_char (line 719): /
```

In this scenario, `$permalink_last_char !== $request_uri_last_char` evaluates to `true`:
https://onlinephp.io?s=bcxBCoAgEEDRfdAdJggsCDxASavOIWIDVqY2aeDtq2XQ9j_4wxhMKAvOYXIRCbJPBNrPCAYJO0C3-lyVRR2QdmUXt0mrzii1UQQCGOsfIzwSPjHR8lX-8qVIzmkPze-jEgL-B21_Aw%2C%2C&v=8.1.12

PS: You can check this on [MEGA](https://mega.wp-rocket.me/multisite/green/).

piotrbak:
Issue is blocked by https://github.com/wp-media/wp-rocket/issues/3477.

If we won't come up with any solution to the above, we should bypass this enhancement for multisites at all.

jeawhanlee:
Open now as https://github.com/wp-media/wp-rocket/issues/3477 is fixed

CrochetFeve0251 (approved):

vmanthos (commented):
This is working as expected except for one case.

On a subfolder multisite network, when the permalinks don't contain a trailing `/`, visiting the homepage of a subsite, e.g. https://mega.wp-rocket.me/multisite/green won't add a trailing `/` as it's happening when WP Rocket is not enabled.

piotrbak:
@vmanthos We can move this issue to `QA done` then, right?

vmanthos:
> We can move this issue to QA done then, right?

If you decide that the issue I've mentioned previously (orange cells [here](https://docs.google.com/spreadsheets/d/1TPCAXKSg1fDKWN8hE0LpS37DYWkboOi3N8anFyQ5iHY/edit?pli=1#gid=615400320)) isn't something we should fix, then yes.

PR #5218, opened 2022-07-06
Load the correct config file in multisite

### LINKED ISSUE #3477: Multisite - Cache Query String(s) & other features not working on subsites

engahmeds3ed:
## Description

When we install WP Rocket in site level in a multisite which uses the sub directory pattern, we always load the main site config file and this is wrong as we need to load its corresponding config file instead.

Fixes #3477

## Type of change

We did it as a team (thanks @CrochetFeve0251 and @jeawhanlee ).
This change may affect the performance as we load the config file in each page load, @Tabrisrp what do u think?

## How Has This Been Tested?

- [x] Locally

# Checklist:

### DIFF
--- a/inc/classes/Buffer/class-config.php
+++ b/inc/classes/Buffer/class-config.php
@@ -194,18 +194,6 @@ class Config {
 
 		$host = $this->get_host();
 
-		if ( realpath( self::$config_dir_path . $host . '.php' ) && 0 === stripos( realpath( self::$config_dir_path . $host . '.php' ), $config_dir_real_path ) ) {
-			$config_file_path = self::$config_dir_path . $host . '.php';
-			return self::memoize(
-				__FUNCTION__,
-				[],
-				[
-					'success' => true,
-					'path'    => $config_file_path,
-				]
-			);
-		}
-
 		$path = str_replace( '\\', '/', strtok( $this->get_server_input( 'REQUEST_URI', '' ), '?' ) );
 		$path = preg_replace( '|(?<=.)/+|', '/', $path );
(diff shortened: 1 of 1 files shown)


remyperona (approved):

remyperona:
I think the performance impact is going to be negligeable, as the path will be empty so we won't enter in the loop anyway.

vmanthos (approved):
Working as expected. 👍

Issue #3077, opened 2020-09-09
Rollback to more than one version backward leads to upgrade to latest version


engahmeds3ed:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version **Yes**
- Used the search feature to ensure that the bug hasn’t been reported before **Yes**

**Describe the bug**
When trying to rollback for more than one time this leads to upgrade instead of multiple rollbacks.

**To Reproduce**
1. I have 3.7.0.1 installled.
2. I rollback to 3.6.4 and this happened successfully
3. from that I try to rollback to 3.5.5.1.
4. Instead it updates to 3.7.0.1

**Expected behavior**
we have two scenarios here:-
1. not to rollback more than one time and here we need to adjust the rollback button text to express that it will update or show error message for example.
2. Open the number of rollbacks and adjust the functionality of the button when user rollback more than one time.

**Screenshots**
First rollback from 3.7.0.1 to 3.6.4
![image](https://user-images.githubusercontent.com/15707971/92628172-b27ce080-f2cc-11ea-9431-2f303e08fe98.png)

![image](https://user-images.githubusercontent.com/15707971/92629838-3041eb80-f2cf-11ea-9b81-1afc8b336abc.png)

Second rollback from 3.6.4 to 3.5.5.1
![image](https://user-images.githubusercontent.com/15707971/92628393-0687c500-f2cd-11ea-80e7-9c567f47a245.png)

![image](https://user-images.githubusercontent.com/15707971/92629853-3637cc80-f2cf-11ea-9e5b-475ae1685e07.png)

**Additional context**
We need product input here

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

arunbasillal:
> Open the number of rollbacks and adjust the functionality of the button when user rollback more than one time.

I always thought that the rollback would keep on rolling back. So this is what I think should happen. 

- Let's say a user rolls back to 3.6.4. This WP Rocket should look and behave exactly like when the user installed 3.6.4 the first time. 
- The last stable is [hardcoded within WP Rocket](https://github.com/wp-media/wp-rocket/blob/8f7f67b621e5085ccf6424a08a15ce636e4fbf2f/wp-rocket.php#L77). A user on 3.6.4 would expect to rollback to the stable one before that and so on. 
[...]

PR #5225, opened 2022-07-07
Closes #5198 Add RUCSS warning modal when rollback

### LINKED ISSUE #5198: RUCSS Deactivation  for Non Expired Users Rollingback
The RUCSS feature available in versions prior to 3.11 has been shut down in May 2022. 
When non-expired users are rolling back to a previous version, we need to inform them that the RUCSS feature won’t be available.

[Details are available here](https://docs.google.com/document/d/1jlyfrU4hYqhFxT_Ol9f-sNJp8h8_ZRo7R4o_5IOjjV8/edit?usp=sharing)

remyperona:
## Description

Add a modal to warn about rollback when remove unused CSS is active since the previous version is not available anymore

The patch versions containing the updated RUCSS notices:
- https://github.com/wp-media/wp-rocket/tree/v3.9.5.1
- https://github.com/wp-media/wp-rocket/tree/v3.10.10.1

Fixes #5198

## Type of change

## How Has This Been Tested?

- [x] Manual tests

# Checklist:


jeawhanlee (approved):

vmanthos:
@Tabrisrp Please find my notes about the PR:

1. The modal is displayed but its style:


is different compared to the one in the [details doc](https://docs.google.com/document/d/1jlyfrU4hYqhFxT_Ol9f-sNJp8h8_ZRo7R4o_5IOjjV8/edit?usp=sharing):



2. In `3.10.10.1`, clicking on the `x` close button doesn't close the banner. If it's not meant to be dismissed, the `x` button should not be displayed.

3. In `3.9.5.1`, clicking on the `x` close button dismisses the dashboard "Congratulations" banner but not the actual RUCSS notice: https://youtu.be/UBfWqOqk5ag

Refreshing the page brings the "Congratulations" banner back.

4. For expired users on `3.9.5.1` we display the banner in WP Rocket's dashboard explaining that users must renew their license:


but only the following is displayed in the file optimization tab:


There isn't any info about the fact that they need to renew their license.

This is how the same is displayed in `3.10.10.1`:


 @Agathemed is this the way you want this displayed?

Lastly, because of #3077 it's not really possible to roll back to these versions. A customer seeing that modal will get a version, e.g. `3.11.4`, where Remove Unused CSS is actually enabled and this can be confusing.

@wp-media/productrocket maybe #3077 should be addressed with priority?

remyperona:
@vmanthos For point 1. can you try with the browser cache refreshed/disabled?

vmanthos:
@Tabrisrp Thank you for the suggestion.

It was indeed an issue coming from the browser caching the previous version of the CSS files. Now the modal is displayed as expected:

vmanthos:
@Tabrisrp Agathe confirmed that for expired users we should display the [same banner](https://camo.githubusercontent.com/909726b7ef882ef210d2031301ccd6fea2aedb61181e8f49c11c18b707f172ee/68747470733a2f2f692e7667792e6d652f523530625a412e6a7067) in the file optimization tab both on `3.9.5.1` and `3.10.10.1`

Will you take care of it in the current PR or do you want me to create a separate GitHub issue?

PS: It's number 4 on the list in [my comment](https://github.com/wp-media/wp-rocket/pull/5225#issuecomment-1180257326).

remyperona:
I pushed a commit to add the warning message when trying to enable the option as an expired user

vmanthos (commented):
@Tabrisrp I can see the banner for expired `3.9.5.1` users but the old message is displayed as well. Also, it's not possible to dismiss the banner unless the page is refreshed:


Can this be changed to work as it does on `3.10.10.1`?


Lastly, depending on the screen's resolution, on mobile devices, the tools tab is larger than the viewport, and so is the modal:


Can you please look into these topics? 🙏

remyperona:
I updated the CSS to avoid the display issue and better display the modal on small screens too.

For the `3.9.5.1` version:
> the old message is displayed as well.

Shouldn't it be displayed still to communicate that it's not available?

> Also, it's not possible to dismiss the banner unless the page is refreshed

Yes this warning has never been dismissible.

> Can this be changed to work as it does on 3.10.10.1?

It's not the same type of field, so we can't.

vmanthos:
> I updated the CSS to avoid the display issue and better display the modal on small screens too.

🙏

This is working as expected now:


> Shouldn't it be displayed still to communicate that it's not available?

Although GitHub showed the .gif I added when I previewed my reply I'm not seeing it anymore. ¯\_(ツ)_/¯

Here is a screenshot of what this looks like for `expired` users in `3.9.5.1`
 

Both the banner and the message are displayed. We should hide the message for expired licenses and instead display the banner.

remyperona:
@vmanthos This should be updated now

vmanthos (approved):
This is working as expected! 👍

PR #5237, opened 2022-07-12
Fix metadata cleanup error in rocket_renew_all_boxes function

tkalimullin-coludlinux:
## Description

The rocket_renew_all_boxes function didn't delete user metadata due to an error in the parameters passed to delete_metadata.

## Type of change

# Checklist:

### DIFF
--- a/inc/functions/admin.php
+++ b/inc/functions/admin.php
@@ -39,7 +39,7 @@ function rocket_need_api_key() {
  */
 function rocket_renew_all_boxes( $uid = null, $keep_this = [] ) {
 	// Delete a user meta for 1 user or all at a time.
-	delete_metadata( 'user', $uid, 'rocket_boxes', null === $uid );
+	delete_metadata( 'user', $uid, 'rocket_boxes', '', null === $uid );
 
 	// $keep_this works only for the current user.
 	if ( ! empty( $keep_this ) && null !== $uid ) {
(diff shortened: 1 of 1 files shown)


remyperona (approved):

inc/functions/admin.php
```diff
@@ -39,7 +39,7 @@ function rocket_need_api_key() {
  */
 function rocket_renew_all_boxes( $uid = null, $keep_this = [] ) {
 	// Delete a user meta for 1 user or all at a time.
-	delete_metadata( 'user', $uid, 'rocket_boxes', null === $uid );
+	delete_metadata( 'user', $uid, 'rocket_boxes', '', null === $uid );
```
engahmeds3ed:
```suggestion
	delete_metadata( 'user', $uid, 'rocket_boxes', '', ! $uid );
```

because we are calling it with zero value here:
https://github.com/wp-media/wp-rocket/blob/d89551a8af8d21e90b74861f97f5317b326c957d/inc/admin/upgrader.php#L28

What do u think?

engahmeds3ed (changes requested):
Thanks, just small code adjustment to make it works.

inc/functions/admin.php
piotrbak: @engahmeds3ed Can you push that change on your own?

piotrbak: @engahmeds3ed Is it possible to push that on our own? 🙏

vmanthos (approved):
This is working as expected.

PR #5239, opened 2022-07-12
Closes #4357 on automate generating WPR css and js file

### LINKED ISSUE #4357: automate generating wp-rocket CSS and JS files after deployment
**Description**
If we are doing modifications in styling or js we generate the minified version of the files locally first with gulp then pushed
which leads to creating conflict between branches and it is not easy to resolve because we dealing with minified versions of files

**Expected behavior**
Generate CSS and JS files after deployment to avoid creating conflicts

jeawhanlee:
## Description
This PR tries to close the initiative of automating the generation of WPR CSS and js files.

Fixes #4357 

## Type of change

Slightly different from the proposed.

# Checklist:


remyperona (changes requested):
That's a good start, but to avoid the issue with merge conflicts and still be able to review/test changes before deployement, we should modify the gulp tasks (or create new one), to generate unminified versions of the `wpr-admin.css`, `wpr-admin-rtl.css` and `wpr-admin.js`

remyperona (commented):
A couple of things:
- Can you run the tasks so that the unminified build are sync'ed to git
- Shouldn't the deploy build also use the `.min` suffix?
- Could you update the enqueue code to be compatible with both no suffix and suffix, depending on the value of `SCRIPT_DEBUG`

inc/admin/ui/enqueue.php
```diff
@@ -8,9 +8,13 @@
  * @since 1.0.0
  */
 function rocket_add_admin_css_js() {
-	wp_enqueue_style( 'wpr-admin', WP_ROCKET_ASSETS_CSS_URL . 'wpr-admin.css', null, WP_ROCKET_VERSION );
+	$admin_style     = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? 'wpr-admin.css' : 'wpr-admin.min.css';
```
remyperona: I'd suggest something like this instead:

```
$suffix = defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min';
```

and then use `$suffix` in the enqueue URL like
```
WP_ROCKET_ASSETS_CSS_URL . 'wpr-admin' . $suffix . '.css'
```

remyperona (changes requested):
The minified files are added to gitignore but still commited, you can use `git rm --cached <file>` to remove them

remyperona (changes requested):
only the map file remaining to be removed

remyperona (approved):

engahmeds3ed:
Do we need QA for this one?

engahmeds3ed:
@piotrbak 
@wp-media/qa 
do we need QA for this one?

piotrbak:
@engahmeds3ed How we could test this PR?

engahmeds3ed:
> @engahmeds3ed How we could test this PR?

I don't think we can test that!

@jeawhanlee @Tabrisrp what do u think?

jeawhanlee:
> > @engahmeds3ed How we could test this PR?
> 
> I don't think we can test that!
> 
> @jeawhanlee @Tabrisrp what do u think?

I don't think QA will be able to test this, we might have to run a fake deploy to test this, though I tested this using the **on:push and pull rquest** git action, so I think @Tabrisrp would know a better way to test this.

remyperona:
Yeah we don't need QA, if it was tested using different triggers, it should be fine on the release tag too.

mostafa-hisham:
@jeawhanlee @Tabrisrp
do we need to update the GitHub action to run the gulp task to generate unminified files too?

PR #5242, opened 2022-07-13
Fixes #3077 Correctly rollback multiple versions consecutively

### LINKED ISSUE #3077: Rollback to more than one version backward leads to upgrade to latest version

remyperona:
## Description

This PR fixes the issue where rolling back multiple versions consecutively was re-installing the latest version of the plugin instead.

Fixes #3077 

## Type of change

This solution removes the callback on the `pre_site_transient_update_plugins` filter, which is the root cause of the issue. It also moves the rollback-related code to the new architecture as part as the Update subscriber.

# Checklist:


jeawhanlee (approved):

vmanthos (approved):
I tested rolling back from `3.11.4.1` to `3.10.9`. It works as expected. 👍 

I cannot test rolling back to previous versions because `3.10.9` doesn't include the fix.

@Tabrisrp @wp-media/productrocket Should this PR be backported to `3.10.10.1` and `3.9.5.1`?

If/when it is, we'll re-test this to see if successive rollbacks work.

**Note**

With PHP `8.1.8` the following pops up in the `debug.log` when rolling back:
```
PHP Deprecated:  Constant FILTER_SANITIZE_STRING is deprecated in /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/admin/upgrader.php on line 41
```
It comes from the `sanitize_key()` function we are using here:
https://github.com/wp-media/wp-rocket/blob/ae10bfa280687698ed7330dceec1087bf55e524d/inc/admin/upgrader.php#L41

This is a core function so we can't do anything on our part

remyperona:
Yes I'm going to backport this change to the patch versions

PR #5280, opened 2022-07-26
Closes #4710 RUCSS compatibility with cloudflare server push

### LINKED ISSUE #4710: RUCSS compatibility with ​CloudFlare Server Push
- [x] Made sure you’re on the latest version `3.10.7`

**Describe the bug**
When using the CloudFlare plugin and enabling Server Push using the following constant, CloudFlare will add all resources with the `rel=preload` hint on the page. Including CSS. 

~~~
define('CLOUDFLARE_HTTP2_SERVER_PUSH_ACTIVE', true);
~~~

We will end up with resources loaded by the page. Example: 

~~~
<link rel="preload" href="/wp-content/themes/astra/assets/css/minified/main.min.css?ver=3.7.7" as="style">
<link rel="preload" href="/wp-includes/css/dist/block-library/style.min.css?ver=5.9" as="style">
~~~

What this does is load the CSS resources even if Remove Unused CSS is used.

It will give a console warning: 

~~~
The resource /wp-content/themes/astra/assets/css/minified/main.min.css?ver=3.7.7 was preloaded using link preload but not used within a few seconds from the window's load event. Please make sure it has an appropriate `as` value and is preloaded intentionally.
~~~

This warning might show on the PageSpeed Insights results as well.

The customer might think that the Remove Unused CSS is not operating properly. Or think that we are preloading the CSS files. 

**To Reproduce**
1. Using the official CloudFlare plugin
2. Add `define('CLOUDFLARE_HTTP2_SERVER_PUSH_ACTIVE', true);` to the `wp-config.php` file
3. See error

**Expected behavior**
Ideally, we should remove the resources hints related to CSS from the page after adding the Used CSS. 

Or at least warn the user using a notice that `CLOUDFLARE_HTTP2_SERVER_PUSH_ACTIVE` is set to `true`, and what it will implicate (loading CSS files + getting the warning about the preloaded assets).

**Screenshots**
image
image

**Additional context**
Ticket - https://secure.helpscout.net/conversation/1763607803/320562/

Potentially linked to the effort for https://github.com/wp-media/wp-rocket/issues/3180 if running CF APO compatibility keeps Cloudflare's official plugin enabled. 

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description

This PR displays a notice when cloudflare server push mode is active.

Fixes #4710 

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:


inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -1,5 +1,4 @@
 <?php
-declare( strict_types=1 );
```
remyperona: why this change?

inc/ThirdParty/Plugins/CDN/Cloudflare.php
```diff
@@ -0,0 +1,98 @@
...
+			return;
+		}
+
+		// if current user has required capapabilities.
+		if ( ! current_user_can( 'rocket_manage_options' ) ) {
+			return;
+		}
+
+		// If RUCSS is enabled.
+		if ( ! (bool) $this->options->get( 'remove_unused_css', 0 ) && ! (bool) $this->options->get( 'minify_concatenate_css', 0 ) ) {
```
Tabrisrp, line 77: Shouldn't the condition be an OR (||) instead of an AND?

inc/ThirdParty/Plugins/CDN/Cloudflare.php
```diff
@@ -0,0 +1,98 @@
...
+
+		if ( in_array( $notice_name, (array) $boxes, true ) ) {
+			return;
+		}
+
+		rocket_notice_html(
+			[
+				'status'         => 'warning',
+				'dismissible'    => '',
+				'message'        => __( 'Cloudflare server pushing mode can create incompatiblities with Remove Unused CSS and Combine CSS features', 'rocket' ),
```
remyperona: The message should be prefixed with `WP Rocket:` in bold like we do for other notices. You can check in the source code for references.

remyperona (changes requested):

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
jeawhanlee: Test were failing [here](https://github.com/wp-media/wp-rocket/blob/215bb7b7f8c0cbb232b5bbadfc058c7c428e3e6b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L922) for php 8 about param 1 expecting string|array given bool.

remyperona: On which function? I'd rather adapt the code or the fixtures to be more strict than removing the strict type checking

tests/Unit/inc/ThirdParty/Plugins/CDN/Cloudflare/displayServerPushingModeNotice.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * @dataProvider configTestData
+     */
+    public function testShouldReturnExpected( $config, $expected ) {
+
+        Functions\stubs([
+            'get_current_screen'   => $config['current_screen'],
+            'current_user_can'    => $config['capability'],
+            'rocket_get_constant' => $config['server_push'],
```
remyperona: Instead of stubbing `rocket_get_constant`, you can use `$this->constants['CLOUDFLARE_HTTP2_SERVER_PUSH_ACTIVE'] = $config['server_push']` at the start of the test method, it will be automatically manage by the StubTrait that way

tests/Integration/inc/ThirdParty/Plugins/CDN/Cloudflare/displayServerPushingModeNotice.php
```diff
@@ -0,0 +1,85 @@
...
+	/**
+	 * @dataProvider configTestData
+	 */
+	public function testShouldDoAsExpected($config, $expected) {
+		
+		Functions\when( 'rocket_get_constant' )->alias(function($value) use($config) {
+			if ( $value == 'CLOUDFLARE_PLUGIN_DIR' || $value == 'CLOUDFLARE_HTTP2_SERVER_PUSH_ACTIVE' ) {
+				return $config['server_push'];
+			}
+		});
```
remyperona: Instead of stubbing `rocket_get_constant`, you can use `$this->constants['CLOUDFLARE_HTTP2_SERVER_PUSH_ACTIVE'] = $config['server_push']` at the start of the test method, it will be automatically manage by the StubTrait that way

remyperona (changes requested):

remyperona (approved):

vmanthos (approved):
@jeawhanlee The notice is properly displayed in the admin area. 👍 

@piotrbak I'd suggest we replace the current message:
> Cloudflare server pushing mode can create incompatiblities with Remove Unused CSS and Combine CSS features

 with this or similar:
> Cloudflare's HTTP/2 Server Push is incompatible with the features of Remove Unused CSS and Combine CSS files. We strongly recommend disabling it.

piotrbak:
@vmanthos The message looks better, you're right!

@jeawhanlee Would you change it?

jeawhanlee:
> @jeawhanlee Would you change it?

Yes I will

vmanthos:
@jeawhanlee I took another look into this.

Currently, the message will be displayed even if Cloudflare isn't used on a site as long as:
1. The Cloudflare plugin is installed/activated.
2. The `'CLOUDFLARE_HTTP2_SERVER_PUSH_ACTIVE` is set to `true`.
3. Combine CSS or Remove Unused CSS is enabled.

I believe we shouldn't display the message unless Cloudflare is actually used on the site.

Maybe only check `if ( ! rocket_is_cloudflare() )` here: 
https://github.com/wp-media/wp-rocket/blob/d4c2f2e942c9a9cf31bf79c147f43d9061d664d2/inc/ThirdParty/Plugins/CDN/Cloudflare.php#L52

and checking if Cloudflare's plugin is active or the constant is `true` here:
[! rocket_get_constant( 'CLOUDFLARE_PLUGIN_DIR' )](https://github.com/wp-media/wp-rocket/blob/d4c2f2e942c9a9cf31bf79c147f43d9061d664d2/inc/ThirdParty/Plugins/CDN/Cloudflare.php#L52)

vmanthos (approved):
Thank you for the changes @jeawhanlee.

This is working as expected.

PR #5297, opened 2022-08-09
Closes #5189 - Compatibility class with one dot com

### LINKED ISSUE #5189: Compatibility with One.com (CDN and page cache)
- [x] Made sure you’re on the latest version `3.11.3`

**Describe the bug**
When One.com is used with WP Rocket's minification, none of the assets are minified (not created). 

The debugged is returning: 

```
Couldn’t get the file path from the URL.
```

Tracking back the error, `$file_path` is returning empty: 

https://github.com/wp-media/wp-rocket/blob/635769c38d685916b97f8184bd7f9219baddefca/inc/Engine/Optimization/Minify/JS/Minify.php#L213-L224

The remaining possible compatibilities are listed in the following comments - https://github.com/wp-media/wp-rocket/issues/5189#issuecomment-1167274538

It would be good to implement compatibility class with `One.com` hosting:
1. If their CDN is enabled
- Grey out our CDN input and fill it with the CNAME value
- Either exclude /wp-includes/(.*) or confirm with One.com why it's handled like this on their end
2. If their Page Caching is enabled
- Clear cache when needed
- Clear our cache when action happens in One.com plugin
- Grey out our Varnish add-on

```
X-VC-Enabled: true | false
X-VC-Cacheable: NO | Yes
X-VC-TTL: seconds (30 days)
X-VC-Purge-Key : timestamp
X-VC-Purge-Host : $_SERVER['SERVER_NAME']
X-VC-Purge-Method:
```
**To Reproduce**
1. Use One.com
2. Use JS or CSS minfication
3. See that all links using `usercontent.one` are not minified

**Expected behavior**
We should minify all assets served by One.com. 

**Screenshots**
N/A

**Additional context**
Ticket - https://secure.helpscout.net/conversation/1929106271/350176

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description

This PR adds compatibility class for one.com

Fixes #5189 

## Type of change

## How Has This Been Tested?

Automated & Manual Tests

# Checklist:


assets/css/wpr-admin-common.css
```diff
@@ -23,3 +23,6 @@
 	text-align: center;
 	z-index: 26;
 }
+.wp-core-ui select:disabled {
+	background-image: none;
+}
```
remyperona: what is this for?

remyperona (commented):
thinking out loud, could we use the `pre_get_rocket_option_` filter instead of  changing the value on init?

inc/ThirdParty/Hostings/OneCom.php
```diff
@@ -0,0 +1,269 @@
...
+		return $settings;
+	}
+
+	/**
+	 * Purge varnish cache.
+	 *
+	 * @return boolean
+	 */
+	public function should_display_varnish_auto_purge_input(): bool {
+		return $this->is_varnish_active() ? false : true;
```
remyperona:
```suggestion
		return ! $this->is_varnish_active();
```

The value returned by the method is already a boolean, you can use it directly

inc/ThirdParty/Hostings/OneCom.php
```diff
@@ -0,0 +1,269 @@
...
+	 *
+	 * @return string
+	 */
+	public function build_cname(): string {
+		if ( ! isset( $_SERVER['ONECOM_DOMAIN_NAME'] ) && ! isset( $_SERVER['HTTP_HOST'] ) ) {
+			return '';
+		}
+
+		$domain_name = $_SERVER['ONECOM_DOMAIN_NAME']; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.ValidatedSanitizedInput.MissingUnslash
+		$http_host   = $_SERVER['HTTP_HOST']; // phpcs:ignore WordPress.Security.ValidatedSanitizedInput.InputNotSanitized, WordPress.Security.ValidatedSanitizedInput.MissingUnslash
```
remyperona: would be better to sanitize the values instead of ignoring the warnings

assets/css/wpr-admin-common.css
jeawhanlee: Screenshot 2022-08-12 at 11 01 40
The dropdown does not display properly when the disable attribute is added, this is a style coming from wp core when the disabled attribute is added.

inc/ThirdParty/Hostings/OneCom.php
jeawhanlee: To catch false positive when one.com varnish is active (`is_varnish_active()` returns true) we want to return false so as to deactivate the toggle switch for our varnish addon.

remyperona: You can reverse the condition with `!` in front. I updated the suggestion to reflect that

assets/css/wpr-admin-common.css
remyperona: it would be better to modify the CSS dedicated to the settings page instead of the one common to the whole admin

inc/ThirdParty/Hostings/HostSubscriberFactory.php
```diff
@@ -45,6 +47,9 @@ public function get_subscriber() {
 				return new Godaddy();
 			case 'kinsta':
 				return new Kinsta();
+			case 'onecom':
+				$options = new Options( 'wp_rocket_' );
+				return new OneCom( $options,  new Options_Data( $options->get( 'settings', [] ) ) );
```
engahmeds3ed: I believe it'd be cleaner to use the helper functions `get_rocket_option` and `update_rocket_option` declared here:
https://github.com/wp-media/wp-rocket/blob/075a6013f6377c4cad2f56a11f10211e7541b303/inc/functions/options.php#L8-L41

instead of passing the options instance.

inc/ThirdParty/Hostings/OneCom.php
```diff
@@ -0,0 +1,235 @@
...
+	 * @param array   $cdn_cnames CDN CNAMES.
+	 * @param array   $cdn_zones CDN Zones.
+	 * @return void
+	 */
+	private function update_cdn_options( int $enable_cdn, array $cdn_cnames, array $cdn_zones ) {
+		$this->options->set( 'cdn', $enable_cdn );
+		$this->options->set( 'cdn_cnames', $cdn_cnames );
+		$this->options->set( 'cdn_zone', $cdn_zones );
+
+		$this->options_api->set( 'settings', $this->options->get_options() );
```
engahmeds3ed: Every time any request happen, we will update those settings and we don't really need that.

You can easily use the filter `pre_get_rocket_option_OPTIONKEY` that is defined here: 
https://github.com/wp-media/wp-rocket/blob/5434080ccd89b36f3312846e4288a50c9d6b7c3e/inc/classes/admin/class-options-data.php#L58

to return the values u need without changing the database.
or if u continued in this update approach u should find a way to make this update only when needed, not every time.

what do u think?

engahmeds3ed (changes requested):

remyperona (approved):

Mai-Saad:
@jeawhanlee Thanks for the PR. Please find exploratory test notes below:

1- (Regression) Currently activating WPR while One's CDN is enabled will result in the following warning:
`Warning: Undefined array key 0 in /customers/2/5/2/2ndlevel-wordpresspod1-cph3.1prod.one/httpd.www/stg/wp-content/plugins/wp-rocket/inc/Engine/CDN/CDN.php on line 208` => expected: No warnings

Screenshot from 2022-11-04 11-47-51


2- After installing WPR, we have this warning (same on the trunk)
`Warning: set_time_limit(): Cannot set max execution time limit due to system policy in /customers/2/5/2/2ndlevel-wordpresspod1-cph3.1prod.one/httpd.www/stg/wp-admin/includes/class-wp-upgrader.php on line 473`
2 installation



3- When activating/deactivating WPR (same on the trunk), we have this in debug.log
`[04-Nov-2022 09:06:52 UTC] Cron reschedule event error for hook: action_scheduler_run_queue, Error code: invalid_schedule, Error message: Event schedule does not exist., Data: {"schedule":"every_minute","args":["WP Cron"],"interval":60}`

on trunk:
Screenshot from 2022-11-04 11-03-10

4- Reserved for UI problem
Screenshot from 2022-11-04 11-26-12

5- While WPR is disabled, Activating One's CDN will result in changing URLs from
`<script type='text/javascript' src='https://stg.2ndlevel-wordpresspod1-cph3.1prod.one/wp-content/print-helloworld-v4.js'></script>`
to 
`<script type='text/javascript' src='https://usercontent.one/wp/stg.2ndlevel-wordpresspod1-cph3.1prod.one/wp-content/print-helloworld-v4.js?media=1667550848'></script>` => this URL won't be minified when WPR and minification are enabled, Shouldn't we minify it?


@piotrbak

-  Do we need to do changes for 2,3?
-  What is expected at WPR UI if there were excluded URLs from CDN at One (currently, nothing is displayed on WPR UI)?
-  What if WPR CDN was enabled with exclusions before enabling One' CDN => once One's CDN is enabled rewritten to CNAME is done but exclusions are still ours, shouldn't it be removed?

piotrbak:
@Mai-Saad No changes for 2,3 are needed

We need to fix 1 and 4.

remyperona (approved):

Mai-Saad:
Working with the following notes (if all are ok for now then we can merge after fixing conflicts + re-smoke)

1. We apply One's CNAME to the relative URL without query string => While ?nowprocket is not writing relative URL to CNAME
2. Exclusions at ONE's CDN are not displayed on our UI
3. Any CDN exclusion added at our UI while ONE's CDN is enabled won't be applied to the page source
4. Any changes at One's settings won't be applied at the page source except after clearing our cache (even if our UI is affected without cache clear while enabling/disabling CDN)
5. There is still a problem with CSS Minification while One's CDN is enabled due to this [issue](https://github.com/wp-media/wp-rocket/issues/5551)
6. One is not caching posts, thus verification of Cache partial/full clear effect on One's cache was done only on home and page type
7. Sometimes our timestamp appears/disappears from the page source at home (mainly appears with 1st visit after Clear One's cache, if we then clear our cache, it won't appear with the next visits till clear One's cache again)
8. If our Cloudflare Addon is enabled (without filling any data there) and WPR settings saved then clear our cache won't clear One's cache
9. Varnish will be dimmed only if One's cache is enabled
10. Exported settings while One's CDN is enabled won't contain CNAME Value and CDN will be 0
11. Enable Safe mode while deactivating WPR won't deactivate CDN

[testrail-report-423.pdf](https://github.com/wp-media/wp-rocket/files/10010889/testrail-report-423.pdf)

PR #5322, opened 2022-08-18
Handle hidden field array value

remyperona:
## Description

- Force an array type for the sitemaps option to prevent type error

## Type of change

# Checklist:


engahmeds3ed (approved):

piotrbak:
@Tabrisrp @engahmeds3ed We had problems with this functionality with the following scenario:
1. Have a 3.11.5 install with custom sitemaps
2. Update to 3.12
3. Disable preload and enable it once again
4. All the sitemaps disappear from the database

Since the solution wasn't perfect since beginning and it doesn't make sense to increase the complexity here, we'll need to get back to this topic during the staggered or in the next minors.

PR #5364, opened 2022-08-29
Fixes #5363 Guard against empty value before passing it to the delete url method in preload

### LINKED ISSUE #5363: Fatal error when attachment_link filter incorrectly changing return of the function get_permalink
- Made sure you’re on the latest version - 3.12.0.3

**Describe the bug**

This issue is created in order to monitor similar cases caused by 3rd party plugins.

Fatal error occurs when `attachment_link` filter incorrectly changing `return` of the function `get_permalink`.
The issue happens when _Preload_ is enabled and an image is being deleted from the Media library.

```
PHP Fatal error:  Uncaught TypeError: WP_Rocket\Engine\Preload\Controller\ClearCache::delete_url(): Argument #1 ($url) must be of type string, null given, called in /wp-content/plugins/wp-rocket/inc/Engine/Preload/Subscriber.php on line 307 and defined in /wp-content/plugins/wp-rocket/inc/Engine/Preload/Controller/ClearCache.php:59
Stack trace:
#0 /wp-content/plugins/wp-rocket/inc/Engine/Preload/Subscriber.php(307): WP_Rocket\Engine\Preload\Controller\ClearCache->delete_url(NULL)
#1 /wp-includes/class-wp-hook.php(309): WP_Rocket\Engine\Preload\Subscriber->delete_post_preload_cache(30828)
#2 /wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters(NULL, Array)
#3 /wp-includes/plugin.php(476): WP_Hook->do_action(Array)
#4 /wp-includes/post.php(6399): do_action('delete_post', 30828, Object(WP_Post))
#5 /wp-admin/post.php(321): wp_delete_attachment(30828, true)
#6 {main}
  thrown in /wp-content/plugins/wp-rocket/inc/Engine/Preload/Controller/ClearCache.php on line 59
```

In this case, the issue is caused by the [Disable Media Permalink by Hardweb.it](https://wordpress.org/plugins/disable-media-permalink-by-hardweb-it/) plugin.

**To Reproduce**
1. Install and activate [Disable Media Permalink by Hardweb.it](Disable Media Permalink by Hardweb.it) plugin;
2. Enable Preload in WP Rocket;
3. Navigate to the Media library;
4. Click "Delete Permanently" for an image;
5. The error will occur.

**Expected behavior**
No error should occur. Possibly we can guard our code.

**Additional context**

As per @engahmeds3ed 
> we are using the documented return type from WP core, they are wrong.

Current workaround is to add `add_filter( 'attachment_link', '__return_false', 11 );` to the `functions.php`.

Ticket: https://secure.helpscout.net/conversation/1990868788/364344/
Slack: https://wp-media.slack.com/archives/C43T1AYMQ/p1661766254643049

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

Use `empty()` function to check the `$url` value instead of a strict comparison with `false`, to be able to catch other empty values like `null`.

Fixes #5363

## Type of change

# Checklist:

### DIFF
--- a/inc/Engine/Preload/Subscriber.php
+++ b/inc/Engine/Preload/Subscriber.php
@@ -300,7 +300,7 @@ class Subscriber implements Subscriber_Interface {
 
 		$url = get_permalink( $post_id );
 
-		if ( false === $url ) {
+		if ( empty( $url ) ) {
 			return;
 		}
 
@@ -320,7 +320,7 @@ class Subscriber implements Subscriber_Interface {
 
 		$url = get_term_link( (int) $term_id );
 
-		if ( false === $url ) {
+		if ( empty( $url ) ) {
 			return;
 		}
 


CrochetFeve0251 (approved):

Mai-Saad (approved):
Working as expected

PR #5377, opened 2022-08-31
Fix warning from integration test due to preload table (Thirdparty tests)

CrochetFeve0251:
## Description
Fix warning due to the cache table from the new preload functionality.

For that I added the table in the bootstrap and added it on every tests needing it.
Note: I could fix warnings on the initial load from the bootstrap.php.
## Type of change

## How Has This Been Tested?

- [ ] Automated Test

# Checklist:


remyperona:
Looking at the tests run on GH there is still a lot if popping-up, it seems to happen even outside of the tests themselves?

CrochetFeve0251:
> Looking at the tests run on GH there is still a lot if popping-up, it seems to happen even outside of the tests themselves?

@Tabrisrp Yes problem is this function `_delete_all_posts` called inside `/tmp/wordpress-tests-lib/includes/bootstrap.php` that call the cache clearing and so the preload.

CrochetFeve0251:
@Tabrisrp look like guarding the table fix warnings outside the tests: https://github.com/wp-media/wp-rocket/runs/8074979289?check_suite_focus=true

PR #5380, opened 2022-08-31
Update action scheduler to v3.5

remyperona:
## Description

Update included action scheduler version to 3.5 (released last week)

## Type of change


mostafa-hisham (approved):
looks ok to me
@engahmeds3ed @CrochetFeve0251 
if you can have another look in case i missed something

vmanthos (approved):

piotrbak:
@Tabrisrp Seems like AS released 2 minor hotfix versions in a short time, the current version is 3.5.2, should we go there directly?

PR #5419, opened 2022-09-13
Closes #5397 Update license renewal banners content

### LINKED ISSUE #5397: Update texts and banners for users with an expired license
The  messages for users with an expired license need to be updated:

- to inform users about the optimize CSS features that will/have disappear/ed
- to display the correct renewal price

Conditions and messages are available [here.](https://docs.google.com/spreadsheets/d/1Zu1MbsVNkOWCsW0CmVRuY69I0t3vtTLG4ykhuOI3YVI/edit?usp=sharing)

remyperona:
## Description

Update the license renewal banners to display the accurate price for not grandfathered users, with no discount.

Fixes #5397

## Type of change

No

## How Has This Been Tested?

- [x] Automated tests

# Checklist:


jeawhanlee (approved):

vmanthos:
@Tabrisrp 

1. According to the [spreadsheet](https://docs.google.com/spreadsheets/d/1Zu1MbsVNkOWCsW0CmVRuY69I0t3vtTLG4ykhuOI3YVI/edit?usp=sharing) the text that should be displayed in the expired banner is:
> Your WP Rocket license is expired!
>
>Your website could be much faster if it could take advantage of our new features and enhancements. 🚀

which is different from the one in the plugin:
https://github.com/wp-media/wp-rocket/blob/94c76535083e13e707d732d0bd70e4b2b5b16ce7/inc/Engine/License/views/renewal-expired-banner.php#L18

Although the differences are small, I've confirmed with @Agathemed that we should be using the one in the spreadsheet.

2. Also, when displaying the prices of the discounted licenses the number are rounded, e.g. 199.2 becomes 199. 
 
This is happening after passing `$this->get_discount_price()` to `umber_format_i18n()` here:

https://github.com/wp-media/wp-rocket/blob/94c76535083e13e707d732d0bd70e4b2b5b16ce7/inc/Engine/License/Renewal.php#L119

Can you please look into these?

remyperona:
@vmanthos Thank you I missed that, it should be good now

Mai-Saad:
@Tabrisrp Please find the test note/question so far below:
- Discount % is displayed to grandfather while expired > 15 days => we shall display regular price in this case
- Shouldn't this transient `wpr_dashboard_seen` expire based on expired since when not once it's set?
- While testing non-grandfather Scenario, by adding  `return false` here after this line:
https://github.com/wp-media/wp-rocket/blob/94c76535083e13e707d732d0bd70e4b2b5b16ce7/inc/Engine/License/Renewal.php#L304 => The price displayed in the banner is as if the discount was applied while discount text not displayed (here no discount and regular price shall be there, not sure if we have API problem or we aren't simulating that type of user right, can you please check that at any duration?)

remyperona:
> Discount % is displayed to grandfather while expired > 15 days => we shall display regular price in this case

Should be fixed with the latest commit

> Shouldn't this transient wpr_dashboard_seen expire based on expired since when not once it's set?

I don't really understand what you mean here. Do we have any issue with the current values?

> While testing non-grandfather Scenario, by adding return false here after this line:

The price discount is calculated separately and doesn't use `is_grandfathered()` method, so you won't get the expected result doing that

Mai-Saad:
> I don't really understand what you mean here. Do we have any issue with the current values?

Well, if we have persona 1 mentioned in the sheet (this persona shall see the bubble as long as popup change or settings page not accessed yet) 
- user open dashboard after 7 days from wpr expired => user saw the notification bubble and related transient set to 2 weeks
- one week later, the popup msg will change, since that transient has not expired and the user already accessed the dashboard, the bubble won't be displayed till `wpr_dashboard_seen` expired, right?
(I am fine with that as I barely notice the bubble anyway. But as per requirements, the bubble shall be displayed when expired > 15) @piotrbak @Tabrisrp what do you think?

piotrbak:
@Mai-Saad I think we're okay here. That's a small and edgy case.

Mai-Saad (approved):
Working as expected
Note: remain issue at website side

PR #5422, opened 2022-09-13
Fixes #5416 Reduced UsedCss index key size

### LINKED ISSUE #5416: Limit the index column size for queue_name

**Describe the bug**
For specific table encoding we might encounter the following error:
```
[08-Sep-2022 18:02:15 UTC] WordPress database error Index column size too large. The maximum column size is 767 bytes. for query CREATE TABLE wp_wpr_rucss_used_css ( 
            id               bigint(20) unsigned NOT NULL AUTO_INCREMENT,
            url              varchar(2000)       NOT NULL default '',
            css              longtext                     default NULL,
            hash             varchar(32)                  default '',
            unprocessedcss   longtext                NULL,
            retries          tinyint(1)          NOT NULL default 1,
            is_mobile        tinyint(1)          NOT NULL default 0,
            job_id           varchar(255)        NOT NULL default '',
            queue_name       varchar(255)        NOT NULL default '',
            status           varchar(255)        NOT NULL default '',
            modified         timestamp           NOT NULL default '0000-00-00 00:00:00',
            last_accessed    timestamp           NOT NULL default '0000-00-00 00:00:00',
            PRIMARY KEY (id),
            KEY url (url(150), is_mobile),
            KEY modified (modified),
            KEY last_accessed (last_accessed),
            INDEX `queue_name_index` (`queue_name`),
            KEY hash (hash) ) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_520_ci made by require_once('wp-admin/admin.php'), do_action('admin_init'), WP_Hook->do_action, WP_Hook->apply_filters, WP_Rocket\Dependencies\Database\Table->maybe_upgrade, WP_Rocket\Dependencies\Database\Table->install, WP_Rocket\Dependencies\Database\Table->create
```

**Expected behavior**
We should limit `INDEX 'queue_name_index' ('queue_name'),` to be maximum 191 characters. 

**Additional context**
We need to decide what's maximum expected size there and alter tables for existing installations.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description

Fix a bug at the creation from the RUCSS table due to the index key being too long.
For that we reduced to 191.

Fixes #5416

## Type of change

No.

## How Has This Been Tested?

- [ ] Manually

# Checklist:


remyperona (commented):
don't we also need to update the schema for existing installs?

piotrbak:
@Tabrisrp What do you think about reducing the size here:
https://github.com/wp-media/wp-rocket/blob/c6b74a8ad463158511a7d10352554379d4731099/inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php#L68

That's what @engahmeds3ed suggested at some point, not sure if that works and which option is better/cleaner.

remyperona:
It really depends on the possible length of the queue_name values, do we know how long it can be?

piotrbak:
From what I can see, we should not expect this value to be long:
image

remyperona:
Makes me wonder why the queue name is indexed in the first place, we don't query based on it in the plugin. Looking at the code, I think there was a mistake at some point, making this an index instead of the status column.

CrochetFeve0251:
> Makes me wonder why the queue name is indexed in the first place, we don't query based on it in the plugin. Looking at the code, I think there was a mistake at some point, making this an index instead of the status column.

Ok I think for that I will have to update the schema for all customers but I agree it is weird to index on that value.

piotrbak:
Since it's scheduled for 3.12.2 we have some time for discussion. 
```
making this an index instead of the status column.
```
Which queries would take advantage of this change? 🤔

remyperona:
We have some queries that are based on the current status, so if there is a lot of entries, they could be faster

remyperona (changes requested):
I think we need to add the method to update the existing schema on update

remyperona (changes requested):
We should also drop the index for ` the queue_name` in the update method

inc/Dependencies/RocketLazyload/Iframe.php
```diff
@@ -106,7 +106,6 @@ private function getExcludedPatterns() {
 				'loading="eager"',
 				'data-skip-lazy',
 				'skip-lazy',
-				'google_ads_iframe_',
```
remyperona: This shouldn't be modified

remyperona (changes requested):

remyperona (approved):

remyperona (changes requested):
getting the following error locally:  `You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'DELETE INDEX `queue_name_index` (`status`)' at line 1`

remyperona (approved):

vmanthos:
@CrochetFeve0251 When installing WP Rocket from scratch indexes look fine.

However, updating WP Rocket from a previous version changes the `queue` index to `status` but it doesn't have the `191` length restriction.

I believe this is coming from here:

https://github.com/wp-media/wp-rocket/blob/35a401ed45acb41d42e19943f1030d3b5cb618ae/inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php#L216

vmanthos (approved):
Working as expected!
[testrail-report-416.pdf](https://github.com/wp-media/wp-rocket/files/9754572/testrail-report-416.pdf)

PR #5427, opened 2022-09-14
Closes 5382: Excluded url with extensions

### LINKED ISSUE #5382: Exclude pages ending with `.php` from getting added to the new preload module
- [x] Made sure you’re on the latest version `3.12.0.4`

**Describe the bug**
WP Rocket's Buffer skips processing pages ending with `.php`. 

https://github.com/wp-media/wp-rocket/blob/3a1dfedbf8937d084351be7a7e3f1b36b0b853f9/inc/classes/Buffer/class-tests.php#L449-L453

But they are still added to the new preload module table.

image

They will be stuck there as `in-progress` since their respective cache will never be created. 

It's best to simply not add them in the first place to avoid any unnecessary load trying to process them. 

In addition to that, we should notify the user that they have URLs ending with `.php` and that consequently, they won't be cached or optimized. 

In most cases, it's the permalink's structure that is at fault. 

**To Reproduce**
1. Add a permalink structure using `.php` as an extension to post names
2. See the issue on the `wp_wpr_rocket_cache` table

**Expected behavior**
`.php` URLs should not be added to preload.

**Screenshots**
Added to the description. 

**Additional context**
Ticket - https://secure.helpscout.net/conversation/1995216159/365250

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description
Fix a problem due to extensions URLs blocked in `in-progress` when they are added to the preload table.

For that we guarded the `create_or_nothing` and the `create_or_update` from [`Cache`](https://github.com/wp-media/wp-rocket/blob/develop/inc/Engine/Preload/Database/Queries/Cache.php) class against urls with file extensions.

Fixes #5382

## Type of change

## How Has This Been Tested?

- [ ] Automated Tests

# Checklist:

### DIFF
--- a/inc/Engine/Preload/Database/Queries/Cache.php
+++ b/inc/Engine/Preload/Database/Queries/Cache.php
@@ -516,7 +516,7 @@ class Cache extends Query {
 	 * @param string $url url to check.
 	 * @return bool
 	 */
-	protected function is_rejected( string $url ) {
+	protected function is_rejected( string $url ): bool {
 		$extensions = [
 			'php' => 1,
 			'xml' => 1,
--- a/inc/Engine/Preload/Database/Queries/Cache.php
+++ b/inc/Engine/Preload/Database/Queries/Cache.php
@@ -124,6 +124,10 @@ class Cache extends Query {
 	public function create_or_update( array $resource ) {
 		$url = untrailingslashit( strtok( $resource['url'], '?' ) );
 
+		if ( $this->is_rejected( $resource['url'] ) ) {
+			return false;
+		}
+
 		// check the database if those resources added before.
 		$rows = $this->query(
 			[
@@ -182,6 +186,11 @@ class Cache extends Query {
 	 * @return bool
 	 */
 	public function create_or_nothing( array $resource ) {
+
+		if ( $this->is_rejected( $resource['url'] ) ) {
+			return false;
+		}
+
 		$url = strtok( $resource['url'], '?' );
 
 		// check the database if those resources added before.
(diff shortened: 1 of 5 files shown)


jeawhanlee (approved):

inc/Engine/Preload/Database/Queries/Cache.php
```diff
@@ -500,4 +509,22 @@ public function remove_all() {
...
+	 * Check if the url is rejected.
+	 *
+	 * @param string $url url to check.
+	 * @return bool
+	 */
+	protected function is_rejected( string $url ) {
+		$extensions = [
+			'php' => 1,
+			'xml' => 1,
+			'xsl' => 1,
```
remyperona, line 523: I'd suggest to add kml extension too, as there was a comment in the issue about this type

remyperona (changes requested):

remyperona (approved):

Mai-Saad:
@CrochetFeve0251 Thanks for the PR. Currently can see these types of URLs in the cache table while permalinks are set to `/%postname%.php/`, can you please check?
Screenshot from 2022-09-29 16-52-19

Mai-Saad (approved):
Working as expected with the following notes:
- This kind of URL is cached and preloaded https://new.rocketlabsqa.ovh/open-book.xsl/project2 while https://new.rocketlabsqa.ovh/open-book.xsl is not cached nor preloaded
- If permalinks are set as `%post-name%.pdf` (another extension than pdf, XML, kml, xsl), Url domain/post.pdf will be cached and preloaded. @piotrbak  what do you think about this?
- If the user was using a trunk, having .PHP URLs in the database then update, those URLs won't automatically be removed nor with clear and preload nor re-enable preload but if permalinks were changed , .PHP URLs won't be added to the database anymore => @piotrbak  what do you think? shouldn't we remove them with the update?

These points are fine based on a discussion with @piotrbak to avoid complexity.

PR #5433, opened 2022-09-15
Closes #5394 Exclude URLs from preload

### LINKED ISSUE #5394: Allow excluding part of the website from the preload feature

**Describe the bug**
In some specific cases, customers might want to preload only part of the website. That's useful when the site owner is sure that specific URLs are not being visited at all.

**Expected behavior**
We should allow excluding patterns from the preload feature.

**Additional context**
There are a couple of approaches possible there. One of them is described below.

First, we need to make sure that **only** desired sitemap is loaded and fetched. We can use existing filter to achieve that:
https://github.com/wp-media/wp-rocket/blob/b39567786ccc5c321ace147e1ff9dcfaa6fa2eed/inc/Engine/Preload/Controller/LoadInitialSitemap.php#L74

Later, we need to add specific filter to exclude specific regex URL patterns from being preloaded/added to database. If pattern matches current URL, we'd check if URL exists in db:
- if so, it'll get removed
- if not, we won't add it

In the UI we'll introduce a text area for exclusions just below the Activate Preload option.

With the above solution, the only possibility that we'll preload not desired URL is cache lifespan expiration. We could also consider using the pattern there, but it would require us to use it directly on filepath or change the filepath to URL.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

### LINKED ISSUE #5445: URLs excluded from caching are still using the Action Scheduler
- Made sure you’re on the latest version ✅ `3.12.1`

**Describe the bug**

When a URL is excluded from caching, e.g. it has been added to the **Never Cache URL(s)** (Advanced Rules tab) it will not be cached but still be processed by the Action Scheduler, and an entry will be added to the `wp_actionscheduler_actions` table.

That URL will have the `in-progress` status and we'll be reprocessed when `rocket_revert_old_in_progress_rows` runs.

**To Reproduce**

1. Add a URL that's already in the `wpr_rocket_cache` table to the **Never Cache URL(s)** text area.
2. Let the preload process be completed.
3. Check the `wp_actionscheduler_actions` table.
4. A `rocket_preload_job_preload_url` entry will have been added there.

**Expected behavior**

Do not process such URLs.

**Additional context**

**Ticket:** https://secure.helpscout.net/conversation/2011973090/369100?folderId=273766

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description
Added a way to exclude uris from the preload.

For that we added a new check with a filter that check with the url to determine if it is excluded from the preload.
Then we used that filter where we update and add the urls to exclude them or delete them.

Fixes #5394

## Type of change

No.

## How Has This Been Tested?

- [ ] Automated Test
- [ ] Manual Test

# Checklist:


inc/Engine/Admin/Settings/Page.php
```diff
@@ -1161,15 +1161,48 @@ private function preload_section() {
...
+					],
+					'description'       => __( 'Specify XML sitemap(s) to be used for preloading', 'rocket' ),
+					'placeholder'       => 'http://example.com/sitemap.xml',
+					'parent'            => 'sitemap_preload',
+					'section'           => 'preload_section',
+					'page'              => 'preload',
+					'default'           => [],
+					'sanitize_callback' => 'sanitize_textarea',
+				],
+			]
```
remyperona: Why was this added back?

inc/Engine/Admin/Settings/Page.php
```diff
@@ -1161,15 +1161,48 @@ private function preload_section() {
...
 					'label'             => __( 'Activate Preloading', 'rocket' ),
 					'section'           => 'preload_section',
 					'page'              => 'preload',
 					'default'           => 1,
 					'sanitize_callback' => 'sanitize_checkbox',
 				],
-				'dns_prefetch'   => [
+				'preload_excluded_uri' => [
+					'type'              => 'textarea',
+					'label'             => __( 'Sitemaps for preloading', 'rocket' ),
```
remyperona: Label is not correct

inc/Engine/Preload/Controller/CheckExcludedTrait.php
```diff
@@ -14,4 +14,25 @@ protected function is_excluded( string $url ) {
...
 	}
+
+	/**
+	 * Check if the url is excluded by using a filter.
+	 *
+	 * @param string $url url to check.
+	 * @return bool
+	 */
+	protected function is_excluded_by_filter( string $url ) {
+		$regexes = apply_filters( 'rocket_preload_exclude_urls', [] );
```
remyperona: Add docblock for filter

inc/Engine/Preload/Controller/CheckExcludedTrait.php
```diff
@@ -14,4 +14,25 @@ protected function is_excluded( string $url ) {
...
 		return (bool) preg_match( "/$excluded/", $url );
 	}
+
+	/**
+	 * Check if the url is excluded by using a filter.
+	 *
+	 * @param string $url url to check.
+	 * @return bool
+	 */
+	protected function is_excluded_by_filter( string $url ) {
```
remyperona:
```suggestion
	protected function is_excluded_by_filter( string $url ): bool {
```

inc/Engine/Preload/Controller/CheckExcludedTrait.php
```diff
@@ -14,4 +14,25 @@ protected function is_excluded( string $url ) {
...
+		$regexes = apply_filters( 'rocket_preload_exclude_urls', [] );
+
+		if ( ! $regexes ) {
```
remyperona:
```suggestion
		if ( empty( $regexes ) ) {
```

inc/Engine/Preload/Subscriber.php
```diff
@@ -375,4 +379,22 @@ public function preload_after_automatic_cache_purge( $deleted ) {
...
 		}
 	}
+
+	/**
+	 * Add the excluded uri from the preload to the filter.
+	 *
+	 * @param array $regexes regexes containing excluded uris.
+	 * @return array|false
+	 */
+	public function add_preload_excluded_uri( array $regexes ) {
```
remyperona:
```suggestion
	public function add_preload_excluded_uri( $regexes ): array {
```

inc/Engine/Preload/Subscriber.php
```diff
@@ -375,4 +379,22 @@ public function preload_after_automatic_cache_purge( $deleted ) {
...
+		$preload_excluded_uri = $this->options->get( 'preload_excluded_uri', false );
```
remyperona:
```suggestion
		$preload_excluded_uri = $this->options->get( 'preload_excluded_uri', [] );
```

inc/Engine/Preload/Subscriber.php
```diff
@@ -375,4 +379,22 @@ public function preload_after_automatic_cache_purge( $deleted ) {
...
+
+		if ( ! $preload_excluded_uri ) {
```
remyperona:
```suggestion
		if ( empty( $preload_excluded_uri ) ) {
```

inc/Engine/Preload/Subscriber.php
```diff
@@ -375,4 +379,22 @@ public function preload_after_automatic_cache_purge( $deleted ) {
...
+			return false;
```
remyperona:
```suggestion
			return $regexes;
```

inc/Engine/Preload/Subscriber.php
```diff
@@ -375,4 +379,22 @@ public function preload_after_automatic_cache_purge( $deleted ) {
...
+		}
+
+		$preload_excluded_uri = explode( "\n", $preload_excluded_uri );
```
remyperona:
```suggestion
		$preload_excluded_uri = explode( "\n", $preload_excluded_uri );
```

The value of the option should be sanitized on saving to convert it to an array instead of a string, as we do for other textarea fields

remyperona (changes requested):

inc/Engine/Preload/Controller/CheckExcludedTrait.php
```diff
@@ -14,4 +14,30 @@ protected function is_excluded( string $url ) {
...
+		 * Regex to exclude URI from preload.
+		 *
+		 * @param string[] regexes to check
+		 */
+		$regexes = apply_filters( 'rocket_preload_exclude_urls', [] );
+
+		if ( empty( $regexes ) ) {
+			return false;
+		}
+
```
engahmeds3ed: U need casting here and make sure that this array contains strings only
```suggestion
		$regexes = (array) apply_filters( 'rocket_preload_exclude_urls', [] );

		if ( empty( $regexes ) ) {
			return false;
		}
		
		$regexes = array_filter( $items, 'is_string' );
```

inc/Engine/Preload/Controller/CheckExcludedTrait.php
```diff
@@ -14,4 +14,30 @@ protected function is_excluded( string $url ) {
...
+		foreach ( $regexes as $regex ) {
+			if ( preg_match( '/' . $regex . '/', $url ) ) {
```
engahmeds3ed: @piotrbak will we apply all regexes or only allow (.*) ?

inc/Engine/Preload/Controller/ClearCache.php
```diff
@@ -31,12 +31,16 @@ public function __construct( Cache $query ) {
 	public function partial_clean( array $urls ) {
 		foreach ( $urls as $url ) {
 			if ( ! $this->is_excluded( $url ) ) {
```
engahmeds3ed: multiple nested levels are not recommended, try bailout early

engahmeds3ed (changes requested):

inc/Engine/Preload/Controller/CheckExcludedTrait.php
piotrbak: Let's keep it with `(.*)` only for now.

piotrbak:
@CrochetFeve0251 Would it be possible to include the following issue directly in this PR?
https://github.com/wp-media/wp-rocket/issues/5445

It sounds like they're touching exactly the same part of the code and I believe we could solve it by just adding the `Never Cache URLs` values to the new filter.

inc/Engine/Preload/Admin/Subscriber.php
```diff
@@ -57,4 +59,21 @@ public static function get_subscribed_events() {
...
+	/**
+	 * Sanitizes Preload Excluded URI option when saving the settings
+	 *
+	 * @param array $input Array of values submitted from the form.
+	 *
+	 * @return array
+	 */
+	public function sanitize_options( $input ) : array {
+		if ( empty( $input['preload_excluded_uri'] ) ) {
+			$input['preload_excluded_uri'] = [];
```
remyperona, line 84:
```suggestion
			$input['preload_excluded_uri'] = [];

			return $input;
```

inc/Engine/Preload/Controller/CheckExcludedTrait.php
remyperona: This still needs to be updated

remyperona (changes requested):

remyperona (approved):

Mai-Saad:
Moved to in-progress till we handle [slack](https://wp-media.slack.com/archives/GUT7FLHF1/p1664283914314709) discussion

piotrbak:
@CrochetFeve0251 Is this one ready to be reviewed again?

CrochetFeve0251:
> @CrochetFeve0251 Is this one ready to be reviewed again?
 
@piotrbak it is

vmanthos:
@CrochetFeve0251 Currently, when any URL is added in the Preload exclusion text area the whole cache is cleared which is not something that should be happening.

This also affects QA testing, for example when trying to check if partial cache clearing will result in the removal of an excluded URL from the database.

PS: The PR is in "Ready for review". So whoever reviews it could also check this change as well.

inc/Engine/Preload/Controller/ClearCache.php
```diff
@@ -30,13 +30,19 @@ public function __construct( Cache $query ) {
 	 */
 	public function partial_clean( array $urls ) {
 		foreach ( $urls as $url ) {
-			if ( ! $this->is_excluded( $url ) ) {
+			if ( $this->is_excluded( $url ) ) {
+				continue;
+			}
```
engahmeds3ed: No need to use this check as we already hooking to the filter here: 
https://github.com/wp-media/wp-rocket/pull/5433/files#diff-13bb6ae74d8bc004dd6fbdb9a3d705ceb0c263bb9ffa6e4df10ebc19190589afR116-R119
and add the callback `add_cache_reject_uri` which gets the option values and add to the filter return.

inc/Engine/Preload/Subscriber.php
```diff
@@ -185,7 +192,12 @@ public function update_cache_row() {
 			do_action( 'rocket_preload_completed', $url, $detected );
 		}
 
-		if ( ( isset( $_GET ) && is_array( $_GET ) ) && 0 < count( $_GET ) || ( $this->query->is_pending( $url ) && $this->options->get( 'do_caching_mobile_files', false ) ) ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		if ( ( isset( $_GET ) && is_array( $_GET ) ) && 0 < count( $_GET ) || ( $this->query->is_pending( $url ) && $this->options->get( 'do_caching_mobile_files', false ) ) ) { //phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
engahmeds3ed: Simplify it like:

```suggestion
		if ( ! empty( (array) $_GET ) || ( $this->query->is_pending( $url ) && $this->options->get( 'do_caching_mobile_files', false ) ) ) { //phpcs:ignore WordPress.Security.NonceVerification.Recommended
```

inc/Engine/Preload/Controller/CheckExcludedTrait.php
```diff
@@ -14,4 +14,33 @@ protected function is_excluded( string $url ) {
...
+		 * Regex to exclude URI from preload.
+		 *
+		 * @param string[] regexes to check
+		 */
+		$regexes = (array) apply_filters( 'rocket_preload_exclude_urls', [] );
+
+		if ( empty( $regexes ) ) {
+			return false;
+		}
+
```
engahmeds3ed, line 36: can we plz make this array unique, we are adding into this filter from two places so we need to make sure that it's unique  and no duplication.
something like
```
$regexes = array_flip( array_flip( $regexes ) );
```

inc/Engine/Preload/Controller/CheckExcludedTrait.php
```diff
@@ -14,4 +14,33 @@ protected function is_excluded( string $url ) {
...
+		$regexes = array_filter( $regexes, 'is_string' );
```
engahmeds3ed: Here we are looping on the items and get only strings from inside and in the following line we are looping again, why not using the second loop and skip the iteration if it's not a string, that's good from performance wise, what do u think?

engahmeds3ed (changes requested):

inc/Engine/Preload/Subscriber.php
```diff
@@ -375,4 +387,20 @@ public function preload_after_automatic_cache_purge( $deleted ) {
...
 	}
+
+	/**
+	 * Add the excluded uri from the preload to the filter.
+	 *
+	 * @param array $regexes regexes containing excluded uris.
+	 * @return array|false
+	 */
+	public function add_preload_excluded_uri( $regexes ): array {
+		$preload_excluded_uri = $this->options->get( 'preload_excluded_uri', [] );
```
engahmeds3ed:
```suggestion
		$preload_excluded_uri = (array) $this->options->get( 'preload_excluded_uri', [] );
```

Guard against failure in array_merge if that's not an array.

engahmeds3ed (approved):
Just a small minor change, Already approved but would be better to have it.

vmanthos (commented):
@wp-media/php Exclusions are not taken into consideration when they contain a `/`, e.g. `/path`.

This is because we are using `/` as the delimiter of the `preg_match()` and we are not escaping the `$regex` here :
https://github.com/wp-media/wp-rocket/blob/13208ce736b2f51b9c26d798a0c10c7e30222f23/inc/Engine/Preload/Controller/CheckExcludedTrait.php#L43

Using `@` as a delimiter fixes that. 

If a URL has a trailing `/` it will not be excluded from preloading because it is checked against the URL without it. This is at least when `update_cache_row()` is executed here:
https://github.com/wp-media/wp-rocket/blob/13208ce736b2f51b9c26d798a0c10c7e30222f23/inc/Engine/Preload/Subscriber.php#L101

`is_excluded_by_filter()` is used in several places. I'm not sure if in all places the `$url` will be without a `/`. These should be checked.

We should make sure that we cover cases like the following:
* `example.com/path?query_string`
*  `example.com/path#anchor`
*  `/path?query_string`
*  `/path#anchor`
* `/path/`
*  `example.com/path/`
* URIs with `http/https`
* the use of `(.*)` in all previous cases

Lastly, while running the `preg_match()` we should take into consideration that the `$regexes` except for what's added in the "Never Cache URL(s)" text area also includes:
```PHP
\/(?:.+\/)?feed(?:\/(?:.+\/?)?)?$|\/(?:.+\/)?embed\/|\/(index\.php\/)?wp\-json(\/.*|$)
```
when caching of the feeds isn't enabled.

remyperona:
@vmanthos I updated the code a bit following your comment, can you test again?

vmanthos:
@Tabrisrp When permalinks have a trailing `/` the change works. 

When their structure doesn't have a trailing `/` excluded URLs with a trailing `/` are preloaded.

I'm getting mixed results for URLs added in the **Never Cache URL(s)** text area, i.e. some of them are still preloaded although they shouldn't. 

This has also to do with how we currently handle URLs added there. For them to be excluded, a trailing `/` has to be added or not depending on whether one exists or not in the permalinks.

I've created a spreadsheet to depict my findings:
https://docs.google.com/spreadsheets/d/1oMq-k0PqYeIriS1DdfJ5p99ayGNhs1mJZ4VjYYP1DlQ/edit?usp=sharing

From a user's perspective, exclusions should work regardless of whether they add a trailing `/` or not in both the preload exclusion area and the Never Cache URL(s).

---

Also, URLs like the following are not currently excluded:
1. `https://vasilis.rocketlabsqa.ovh/pariatur-ipsam-reiciendis-aliquam-perferendis-eos-corporis?source=fb_campaign`
2. `https://vasilis.rocketlabsqa.ovh/nisi-quos-repudiandae-magni#contact`

@piotrbak Should we handle these? 

About URLs containing query strings (example 1), once we allow preloading them using custom sitemaps, we should consider if we'll offer the ability to exclude them using the UI.  Of course, customers can edit the sitemaps to remove the ones they don't want to be preloaded.

About 2, it may make sense to not preload the URL without the anchor link.

piotrbak:
@vmanthos 
```
About URLs containing query strings (example 1), once we allow preloading them using custom sitemaps, we should consider if we'll offer the ability to exclude them using the UI. Of course, customers can edit the sitemaps to remove the ones they don't want to be preloaded.
```
Yes, I believe we should allow the exclusions there, but I think only when they're specified directly like:
`/pariatur-ipsam-reiciendis-aliquam-perferendis-eos-corporis?(.*)`
What do you think?

```
About 2, it may make sense to not preload the URL without the anchor link.
```
Not sure if I understood correctly. Do you think it makes sense to save URLs with the anchor to the DB at all? 🤔

vmanthos:
> I believe we should allow the exclusions there, but I think only when they're specified directly

This seems good.

> Do you think it makes sense to save URLs with the anchor to the DB at all?

If a URL having an anchor is added to the exclusion area, we should exclude it from preload, i.e. treat it as though it didn't have the anchor link.

remyperona:
Can we summarize what needs to happen here to get the PR ready?

piotrbak:
@Tabrisrp The only question here is about how this PR will work together with https://github.com/wp-media/wp-rocket/pull/5491.
This type of exclusions should match, right? `/pariatur-ipsam-reiciendis-aliquam-perferendis-eos-corporis?(.*)` 

I'll send you the description + documentation URLs once confirmed with Lucy and we're 👌 here.

piotrbak:
@Tabrisrp Here's the detailed wording for this textarea:
**Title:** Exclude URLs (it should be URLs not URI, what do you think?)
**Description:** Specify URLs to be excluded from the preload feature (one per line). More info
**More info links:** 
```
https://docs.wp-rocket.me/article/1721-exclude-urls-from-being-preloaded
6349682bde258f5018eb456d
https://fr.docs.wp-rocket.me/article/1722-exclure-url-du-prechargement
63496d5b8a552811521e52d2
```
**Placeholder:** /author/(.*)
**Below the textarea:** Use (.*) wildcards to address multiple URLs under a given path.
image

remyperona:
setting field is updated with all the info. For the query string, we can test that in the other PR once we have this merged

vmanthos:
@Tabrisrp Still, when permalinks don't have a trailing `/` and the exclusion has, the URL is preloaded.

Logging `$regex` and `$url` here:
https://github.com/wp-media/wp-rocket/blob/324bf97608d7dcd5aa92ea3374838256abca508e/inc/Engine/Preload/Controller/CheckExcludedTrait.php#L45

I'm getting:
```PHP
Regex: /consectetur-bibendum-eu/
URL: https://vasilis.rocketlabsqa.ovh/consectetur-bibendum-eu
```

which is why the RegEx doesn't exclude the URL.

`untrailingslashit()` the `$regex` here:
https://github.com/wp-media/wp-rocket/blob/324bf97608d7dcd5aa92ea3374838256abca508e/inc/Engine/Preload/Controller/CheckExcludedTrait.php#L40

works but then we'll have issues if, e.g. a `(.*)` is added after the path.

Can you please take a look into this? 🙏

vmanthos:
After the latest changes this is working as expected except for this scenario:
* Permalinks settings without a trailing `/`
* Never Cache URLs with a trailing `/` -> preloaded & cached as per #5509.

@piotrbak Is this something that should be handled now?

piotrbak:
@vmanthos I think we're okay here for now, it's a bit edgy and we already have an issue for that

PR #5480, opened 2022-09-29
Fix/5400 handling RUCSS errors

### LINKED ISSUE #5400: Better handling RUCSS errors

**Describe the bug**
1. If job wasn't successful, we don't store any data why it failed
2. When SaaS receives timeout, we're increasing the `retries` column up to 3 and then change the status to `failed`. For each retry we're fetching the same Job Details

**Expected behavior**
1. If job wasn't successful, we should save the last status code and the last error message to the database, erasing the previous messages. We'll need a new column to achieve that. 
2. We should create a new `job` in RUCSS for each retry (3 retries) instead of fetching the same `job details`.

**Additional context**
Additional context:
https://wp-media.slack.com/archives/C029G1B5HD2/p1662458337836339?thread_ts=1662099835.213589&cid=C029G1B5HD2

@engahmeds3ed @mostafa-hisham since you were discussing this problem, I believe that you should do grooming there. 

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

mostafa-hisham:
## Description
Save errors we receive from SAAS in the database and create new jobs for timeout pages

Fixes #5400 

No
## How Has This Been Tested?

- [x] Tested locally
- [x] Unit and integration tests

# Checklist:


inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -612,13 +623,20 @@ public function check_job_status( int $id ) {
...
-			$this->used_css_query->increment_retries( $id, $row_details->retries );
+			// on timeout errors with code 504 create new job.
+			if ( 504 === $job_details['code'] ) {
+				$add_to_queue_response = $this->add_url_to_the_queue( $row_details->url, $row_details->is_mobile );
+				if ( false !== $add_to_queue_response ) {
+					$new_job_id = $add_to_queue_response['contents']['jobId'];
+				}
+			}
+			// Increment the retries number with 1 , Change status to pending again and change job id on timeout.
+			$this->used_css_query->increment_retries( $id, $row_details->retries, $new_job_id );
```
engahmeds3ed: We need to apply single responsibility concept here, so the method `increment_retries` is used to increment the number of retries, I know that you need to only make one query to handle both cases, but think of it, the normal case, this new_job_id would be false.

Also, timeout status code is 408.

What do u think?

```suggestion
			// on timeout errors with code 408 create new job.
			switch( $job_details['code'] ){
				case 408:
					$add_to_queue_response = $this->add_url_to_the_queue( $row_details->url, $row_details->is_mobile );
					if ( false !== $add_to_queue_response ) {
						// Create a new method in the query class to update the job_id column with $add_to_queue_response['contents']['jobId']
					}
				break;
			}
			// Increment the retries number with 1 , Change status to pending again and change job id on timeout.
			$this->used_css_query->increment_retries( $id, $row_details->retries );
```

inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php
```diff
@@ -230,4 +232,21 @@ public function maybe_trigger_recreate_table() {
...
+	 *
+	 * @return bool
+	 */
+	protected function add_error_message_column() {
+		$error_message_column_exists = $this->column_exists( 'error_message' );
+
+		$created = true;
+
+		if ( ! $error_message_column_exists ) {
+			$created &= $this->get_db()->query( "ALTER TABLE {$this->table_name} ADD COLUMN error_message longtext NULL default NULL AFTER hash" );
```
engahmeds3ed: Enclose the table name between `` like.

```suggestion
			$created &= $this->get_db()->query( "ALTER TABLE `{$this->table_name}` ADD COLUMN error_message longtext NULL default NULL AFTER hash" );
```

engahmeds3ed (changes requested):
@piotrbak so here we will only handle the timeout error and can't write the used CSS file into filesystem and then we can extend that to log the other errors, OK with that?

piotrbak:
@engahmeds3ed 
```
so here we will only handle the timeout error and can't write the used CSS file into filesystem and then we can extend that to log the other errors, OK with that?
```
Can't we log everything that is not 200? Later we will only adjust things on the SaaS side (to respond on errors with either status code or message) and eventually implement automatic reactions on the plugin side.

engahmeds3ed:
Totally agree with you, @mostafa-hisham can u plz completely remove the check for the status code and only log all codes?
I know that we only have 200 and 500 now from SaaS side but as Piotr mentioned, we need to make it more generic from the plugin side and control it from the SaaS whenever we want.

piotrbak:
@engahmeds3ed here we need @mostafa-hisham to continue the work on this right? If so, we'll move it to the next minor

mostafa-hisham:
@piotrbak @engahmeds3ed 
We log (save)  the error message in all status codes but only create a new job if the page  timeout or if we couldn't open it with puppeteer (status 408)

piotrbak:
@mostafa-hisham thank you!

remyperona (approved):

vmanthos (approved):
This is working as expected.

**Note**
Although I simulated a timeout, there isn't a way to force Puppeteer on the SaaS side to not open a page.

However, I randomly run into it:

vmanthos:
@mostafa-hisham I'm getting the following database error, which prevents the creation of the `wpr_rucss_used_css` table during a fresh installation:

```PHP
 WordPress database error Incorrect prefix key; the used key part isn't a string, the used length is longer than the key part, or the storage engine doesn't support unique prefix keys for query CREATE TABLE wp_wpr_rucss_used_css ( 
			id               bigint(20) unsigned NOT NULL AUTO_INCREMENT,
			url              varchar(2000)       NOT NULL default '',
			css              longtext                     default NULL,
			hash             varchar(32)                  default '',
			error_code       varchar(32)                NULL default NULL,
			error_message    longtext                NULL default NULL,
			unprocessedcss   longtext                NULL,
			retries          tinyint(1)          NOT NULL default 1,
			is_mobile        tinyint(1)          NOT NULL default 0,
			job_id           varchar(255)        NOT NULL default '',
			queue_name       varchar(255)        NOT NULL default '',
			status           varchar(255)        NOT NULL default '',
			modified         timestamp           NOT NULL default '0000-00-00 00:00:00',
			last_accessed    timestamp           NOT NULL default '0000-00-00 00:00:00',
			PRIMARY KEY (id),
			KEY url (url(150), is_mobile),
			KEY modified (modified),
			KEY last_accessed (last_accessed),
			INDEX `status_index` (`status`(191)),
			INDEX `error_code_index` (`error_code`(191)),
			KEY hash (hash) ) DEFAULT CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_520_ci made by require_once('wp-admin/admin.php'), do_action('admin_init'), WP_Hook->do_action, WP_Hook->apply_filters, WP_Rocket\Dependencies\Database\Table->maybe_upgrade, WP_Rocket\Dependencies\Database\Table->install, WP_Rocket\Dependencies\Database\Table->create
```

PS: I've tested this on 135.125.216.230 with PHP `8.1`.

inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php
```diff
@@ -28,8 +28,7 @@ class UsedCSS extends Table {
 	 *
 	 * @var int
 	 */
-	protected $version = 20220920;
-
+	protected $version = 20220926;
```
engahmeds3ed, line 31:
```suggestion
	protected $version = 20221104;
```

this need to be changed to be the last one

inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php
```diff
@@ -41,6 +40,9 @@ class UsedCSS extends Table {
 		20220131 => 'make_status_column_index',
 		20220513 => 'add_hash_column',
 		20220920 => 'make_status_column_index_instead_queue_name',
+		20220926 => 'add_error_message_column',
+		20221103 => 'add_error_code_column',
+		20221104 => 'make_error_code_column_index',
```
engahmeds3ed: I believe we need to make all of those changes in one function, because of the following reasons:
1. This migrations array is growing too fast
2. Those changes are still not shipped to any customer so no need to make separate function for each one.

Just make it one method called `add_error_columns`

What do u think?

inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php
```diff
@@ -62,6 +64,8 @@ protected function set_schema() {
 			url              varchar(2000)       NOT NULL default '',
 			css              longtext                     default NULL,
 			hash             varchar(32)                  default '',
+			error_code       varchar(32)                NULL default NULL,
```
engahmeds3ed:
```suggestion
			error_code       varchar(32)             NULL default NULL,
```

remove three spaces to be correctly aligned (small and silly change I know :D )

inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php
```diff
@@ -75,6 +79,7 @@ protected function set_schema() {
 			KEY modified (modified),
 			KEY last_accessed (last_accessed),
 			INDEX `status_index` (`status`(191)),
+			INDEX `error_code_index` (`error_code`(191)),
```
engahmeds3ed:
```suggestion
			INDEX `error_code_index` (`error_code`),
```

we don't need to worry about this 191 utf8 and multibyte issue because the column size is already less than 191.

engahmeds3ed (changes requested):

engahmeds3ed (approved):

vmanthos (approved):

**Note**

The `error_code` and `error_message` will be added after the third failed retry. For troubleshooting purposes, e.g. during support, it would be good to have them after every failed retry.

PR #5483, opened 2022-09-29
Fixes Fatal error when get_template return Null

### LINKED ISSUE #5481: PHP Fatal error when Template is used in a parent theme
- Made sure you’re on the latest version ✅ `3.12.1.1`

**Describe the bug**

When the header of a parent theme incorrectly includes a `Template` a PHP Fatal error occurs:
```PHP
PHP Fatal error:  Uncaught TypeError: strtolower() expects parameter 1 to be string, null given in /home/example.com/wp-content/plugins/wp-rocket/inc/ThirdParty/Themes/Flatsome.php:75
Stack trace:
#0 /home/example.com/wp-content/plugins/wp-rocket/inc/ThirdParty/Themes/Flatsome.php(75): strtolower(NULL)
#1 /home/example.com/wp-content/plugins/wp-rocket/inc/ThirdParty/Themes/Flatsome.php(15): WP_Rocket\ThirdParty\Themes\Flatsome::is_flatsome()
#2 /home/example.com/wp-content/plugins/wp-rocket/inc/classes/event-management/class-event-manager.php(38): WP_Rocket\ThirdParty\Themes\Flatsome::get_subscribed_events()
#3 /home/example.com/wp-content/plugins/wp-rocket/inc/Plugin.php(144): WP_Rocket\Event_Management\Event_Manager->add_subscriber(Object(WP_Rocket\ThirdParty\Themes\Flatsome))
#4 /home/example.com/wp-content/plugins/wp in /home/example.com/wp-content/plugins/wp-rocket/inc/ThirdParty/Themes/Flatsome.php on line 75
```
here:
https://github.com/wp-media/wp-rocket/blob/103abf7be9d8acd6ea3abae742b533bfd9cd4c32/inc/ThirdParty/Themes/Flatsome.php#L72

This is because `$theme->get_template()` returns `NULL`.

**To Reproduce**

1. Add `Template: folder-name-of-the-theme` in the header of the `style.css` of the active theme.
2. Try to activate WP Rocket.
3. Look into the `debug.log` for the fatal error.

**Expected behavior**

No fatal errors should occur even if the `Template` is added to a parent theme.

**Additional context**

We should guard the check for the template and do the same in other theme compatibility files, e.g. Avada, Jevelin, Xstore where we are using the same check.

**Ticket:** https://secure.helpscout.net/conversation/2019395931/370792/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

mostafa-hisham:
## Description

covert null to empty string when use `strlower` with `get_template`

Fixes #5481

## Type of change

no

## How Has This Been Tested?
- [x] Tested locally
- [x] Unit and integration

# Checklist:


vmanthos:
@mostafa-hisham There is this as well:
https://github.com/wp-media/wp-rocket/blob/103abf7be9d8acd6ea3abae742b533bfd9cd4c32/inc/ThirdParty/Themes/Avada.php#L47

jeawhanlee (approved):

Mai-Saad:
@mostafa-hisham  @Tabrisrp As per discussion with @vmanthos regarding the Uncode theme here `/inc/3rd-party/themes`

> For Uncode, we will most likely face the same issue because of the way we are checking for the child theme which is the same as for the Flatsome issue:
> https://github.com/wp-media/wp-rocket/blob/962863d6e91ea054fc904202a478ff668c71d1a1/inc/3rd-party/themes/uncode.php#L7
> The Uncode compatibility file should be moved to /inc/ThirdParty/Themes and use the is_current_theme() method that Remy added to fix the fatal error:
> https://github.com/wp-media/wp-rocket/blob/962863d6e91ea054fc904202a478ff668c71d1a1/inc/ThirdParty/Themes/ThirdpartyTheme.php#L22-L31

What do you think?

remyperona:
Uncode theme compatibility code has been updated to use the new system too

jeawhanlee (approved):

Mai-Saad (approved):
Working as expected

PR #5488, opened 2022-09-30
Fixes warning in unit tests related to expectation after Mockery v1.5.1 update

remyperona:
## Description

Fixes unit tests warning which appeared after Mockery v1.5.1 update, in which they remove the default expectation of once when you stub a method. Now we need to explicitly set the method count if it's called one time.

## Type of change

# Checklist:


tests/Fixtures/inc/Engine/Optimization/RUCSS/Cron/Subscriber/cronCleanRows.php
```diff
@@ -32,20 +32,18 @@
 	'test_data' => [
 		'shouldDeleteOnUpdate' => [
 			'input' => [
-				'remove_unused_css' => true,
```
CrochetFeve0251, line 35: @Tabrisrp I think we need to keep that we use it in the tests

tests/Fixtures/inc/Engine/Optimization/RUCSS/Cron/Subscriber/cronCleanRows.php
```diff
@@ -32,20 +32,18 @@
...
-				'used_css'          => $used_css,
+				'used_css' => $used_css,
 				'has_delay' => true,
 				'delay' => 1,
 				'deletion_activated' => true,
 			]
 		],
 		'shouldNotDeleteOnDisabled' => [
 			'input' => [
-				'remove_unused_css' => false,
```
CrochetFeve0251, line 44: @Tabrisrp same here

CrochetFeve0251 (changes requested):

tests/Fixtures/inc/Engine/Optimization/RUCSS/Cron/Subscriber/cronCleanRows.php
remyperona: the option is never used for this method

PR #5494, opened 2022-10-03
Fixed the issue with the url

### LINKED ISSUE #5482: Duplication in preload DB table when sitemap contains query strings

**Describe the bug**
If the sitemap contains urls with query strings, for example: https://example.org/?post_type=test&id=5 we save this url into the database without query strings to be like https://example.org/ and when we have multiple urls like that, we insert those with the same url so the DB has duplication.

**To Reproduce**
1. Install and activate RankMath SEO plugin.
2. Install and activate Thrive Theme Builder Version: 3.11.1
3. Make sure that those post types are enabled in sitemap
thrive

4. Enable the preload and start preloading from sitemap.
5. Check the cache database table, you will find the duplication there.

**Expected behavior**
We should make the cache table rows unique by urls.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description

Fix a problem of duplication with urls having query params due to wrong variable.

Fixes #5482

## Type of change

No.

## How Has This Been Tested?

- [ ] Manual Test

# Checklist:

### DIFF
--- a/inc/Engine/Preload/Database/Queries/Cache.php
+++ b/inc/Engine/Preload/Database/Queries/Cache.php
@@ -187,7 +187,7 @@ class Cache extends Query {
 		// check the database if those resources added before.
 		$rows = $this->query(
 			[
-				'url' => untrailingslashit( $resource['url'] ),
+				'url' => untrailingslashit( $url ),
 			],
 			false
 		);


jeawhanlee (approved):

vmanthos (approved):
This is working as expected.

PR #5552, opened 2022-11-08
Get remote delay JS scripts list

### LINKED ISSUE #5525: [Task] Refactor Delay JavaScript - Get remote scripts list (1/3)
Sub Task from the epic: #5493

# CRON Job

We will have a CRON job that runs weekly to get the latest changes from the API: 
https://staging-sass-backend.wp-rocket.me/api/delay-js-exclusions/list 

Try using the code from here: https://github.com/wp-media/wp-rocket/blob/029c4de9879a7f235cffa8ddd881828934149c9f/inc/Engine/Optimization/DynamicLists/DynamicLists.php#L148-L152
With the same CRON (Don't create a new CRON)
  
We will save the data into the json file and also cache it for one week in transient, let's save them with the following structure
- plugins (type in API)
  - Plugin ID (condition in API)
    - Plugin Name (title in API)
    - Exclusions (exclusions in API)
    - is_default
- themes (type in API)
  - Theme ID (condition in API)
    - Theme Name (title in API)
    - Exclusions (exclusions in API)
    - is_default
- scripts
  - Script Name (title in API)
  - Exclusions (exclusions in API)
  - is_default

We can check with @mostafa-hisham if we can adjust the data in SaaS to return like that, because this will be easier to match with current themes/plugins and also to show them in settings page.

# DevOps

- With deployment we will get and save the json file into the plugin's code, so we need to edit the old GitHub workflow here to be more generic and deals with both lists:
  https://github.com/wp-media/wp-rocket/blob/d5504dbe87397f15711a30761ece630f97af8920/.github/workflows/deploy_dynamic_lists_fallback.yml#L1

- Create a dedicated json file for this delay JS exclusion list.

- I hope we can make a dedicated directory for those json files to be more organized

engahmeds3ed:
## Description

As mentioned in the issue, I refactored the dynamic lists part to have both default options' lists and the new delay JS lists.

Fixes #5525 

## Type of change

No

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DynamicLists/AbstractAPIClient.php
+++ b/inc/Engine/Optimization/DynamicLists/AbstractAPIClient.php
@@ -49,6 +49,11 @@ abstract class AbstractAPIClient {
 		$this->options = $options;
 	}
 
+	/**
+	 * Specify API endpoint path.
+	 *
+	 * @return string
+	 */
 	abstract protected function get_api_path();
 
 	/**
@@ -82,7 +87,6 @@ abstract class AbstractAPIClient {
 	/**
 	 * Handle the request.
 	 *
-	 * @param string $request_path request path.
 	 * @param array  $args Passed arguments.
 	 *
 	 * @return bool
--- a/inc/Engine/Optimization/DynamicLists/AbstractAPIClient.php
+++ b/inc/Engine/Optimization/DynamicLists/AbstractAPIClient.php
@@ -6,7 +6,7 @@ namespace WP_Rocket\Engine\Optimization\DynamicLists;
 use WP_Error;
 use WP_Rocket\Admin\Options_Data;
 
-class APIClient {
+abstract class AbstractAPIClient {
 	/**
 	 * API URL.
 	 */
@@ -17,28 +17,28 @@ class APIClient {
 	 *
 	 * @var int
 	 */
-	protected $response_code = 200;
+	private $response_code = 200;
 
 	/**
 	 * Error message.
 	 *
 	 * @var string
 	 */
-	protected $error_message = '';
+	private $error_message = '';
 
 	/**
 	 * Response Body.
 	 *
 	 * @var string
 	 */
-	protected $response_body = '';
+	private $response_body = '';
 
 	/**
 	 * Plugin options instance.
 	 *
 	 * @var Options_Data
 	 */
-	protected $options;
+	private $options;
 
 	/**
 	 * Instantiate the class.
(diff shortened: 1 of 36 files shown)


inc/Engine/Optimization/DynamicLists/AbstractDataManager.php
```diff
@@ -20,13 +29,27 @@ public function __construct( $filesystem = null ) {
...
+	 *
+	 * @return string
+	 */
+	abstract protected function get_json_filename();
+
 	/**
 	 * Gets the lists content
 	 *
-	 * @return object
+	 * @return object|string
```
remyperona: we should avoid mixed return values, why do we need this change?

inc/Engine/Optimization/DynamicLists/AbstractAPIClient.php
```diff
@@ -100,7 +106,7 @@ private function handle_request( string $request_path, array $args ) {
 		];
 
 		$response = wp_remote_get(
-			$api_url . $request_path,
+			$api_url . $this->get_api_path(),
```
remyperona, line 109: get_api_path() should be added as an astract method in this class before being used in the child classes

remyperona (changes requested):

inc/Engine/Optimization/DynamicLists/AbstractAPIClient.php
engahmeds3ed: it's already abstract here: https://github.com/wp-media/wp-rocket/blob/dbebfe38485d7dda81555aeccda08fffda2d2bbb/inc/Engine/Optimization/DynamicLists/AbstractAPIClient.php#L57

inc/Engine/Optimization/DynamicLists/AbstractDataManager.php
engahmeds3ed: because here:
https://github.com/wp-media/wp-rocket/blob/dbebfe38485d7dda81555aeccda08fffda2d2bbb/inc/Engine/Optimization/DynamicLists/AbstractDataManager.php#L63

we return empty string.
What do u suggest?

inc/Engine/Optimization/DynamicLists/AbstractAPIClient.php
remyperona: ah I searched for it but missed it

inc/Engine/Optimization/DynamicLists/AbstractDataManager.php
remyperona: We can return a stdClass object instead of an empty string

CrochetFeve0251 (approved):

PR #5563, opened 2022-11-11
Refactor Delay JavaScript - Settings part

### LINKED ISSUE #5529: [Task] Refactor Delay JavaScript - Settings part (2/3)
Sub Task from the epic: #5493

# Setting view

- We need to create a new field type called `delay_js_exclusion` or we can make it more generic `categorized_multiselect`, start with a method like that: https://github.com/wp-media/wp-rocket/blob/d833ad2e34dab92e3e58838fedf4b4d643d84ce8/inc/Engine/Admin/Settings/Render.php#L425-L434
- This field type will accept a new argument called `items` which will have the items from the exclusion list mentioned here #5525 also bear in mind that the value attribute in this case is array.
- For the plugin icons we can get them from wp.org https://ps.w.org/woocommerce/assets/icon-128x128.png
- For the theme icon (if needed), get screenshot.png from the active theme
- We need also a way to cache those icons and always fallback to a default icon if the plugin is premium and isn't on the wp.org
- We need the figma file to apply the CSS from there easily and precisely @piotrbak @GeekPress 

# Settings Save

- Save the selected items to be excluded from delay JS (just to show the selection on the settings page)
- We will have another hidden field to save the exclusions for those selected items or at least cache it in transient so it can be easily called in frontent part while generating the cache file
- When we have some changes in the API for the exclusions we will need to regenerate the exclusions that are mentioned in the previous point based on the selected items and the data from the API, for example:
If u have plugin_1 that has an item that is selected which has three exclusions, the next time we get the list from the API we got that it has only one exclusion (or four exclusions) so here we will regenerate this list as the saved one is obsolete.

engahmeds3ed:
## Description

Created UI and logic to save inputs for the new Delay JS option.
For that we created a new input type and added logic to link it to the list.
On another end we added logic to Delay JS to save 3 fields:
- `delay_js_exclusions`: To save manual exclusion.
- `delay_js_exclusions_selected`: To save name of exclusions excluded by the user.
- `delay_js_exclusions_selected_exclusions`: To save exclusions generated according to the plugins selected by the user.
/!\ : Icons for plugins, scripts and themes are still missing.
 
Fixes #5529 

## Type of change

Yes.

## How Has This Been Tested?

- [ ] Local tests.

# Checklist:


views/settings/fields/categorized_multiselect.php
```diff
@@ -0,0 +1,204 @@
...
+}
+
+?>
+
+<div id = '<?php echo esc_attr( $data['id'] ); ?>' class="wpr-field wpr-multiple-select <?php echo esc_attr( $data['container_class'] ); ?>" data-default="<?php echo ( ! empty( $data['default'] ) ? 'wpr-radio-' . esc_attr( $data['default'] ) : '' ); ?>" <?php echo $data['parent']; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped -- $data['parent'] escaped with esc_attr. ?>>
+
+	<h2><?php echo esc_html( __( 'Excluded JavaScript Files', 'rocket' ) ); ?></h2>
+	<p>
+	<?php
+	echo esc_html( __( 'Add JavaScript files to be excluded from delaying execution by selected them / or by add in “My scripts”.', 'rocket' ) );
```
mostafa-hisham: `My scripts` looks strange I think we need to change it to `delayed scripts` or find a better title

views/settings/fields/categorized_multiselect.php
```diff
@@ -0,0 +1,204 @@
...
+	<?php
+	render_list(
+		esc_html( __( 'Themes', 'rocket' ) ),
+		esc_attr( $data['id'] ) . '_themes',
+		$wp_rocket_plugins,
+		$wp_rocket_state
+	);
+	?>
+
+	<p><?php echo esc_html( __( 'Specify URLS or keywords in “My scripts”, which are not in the above categories.', 'rocket' ) ); ?></p>
```
mostafa-hisham: My scripts again

views/settings/fields/categorized_multiselect.php
```diff
@@ -0,0 +1,204 @@
...
+	<div class="wpr-list open">
+		<div class="wpr-list-header">
+			<h3><?php echo esc_html( __( 'My scripts', 'rocket' ) ); ?></h3>
```
mostafa-hisham: My scripts again

views/settings/fields/categorized_multiselect.php
GeekPress: `My scripts` is the label that has been recommended by the UX agency regarding the new UI.

CrochetFeve0251: @GeekPress to be fair I agree with @mostafa-hisham it took me time to understand what was `My scripts`

vmanthos:
> Add JavaScript files to be excluded from delaying execution by selected them / or by add in “My scripts”.

I agree that "My scripts" is a bit confusing. 

Additionally, if we opt in for this text it should be:

> Exclude JavaScript files from delayed execution by selecting them or by adding them to “My scripts”.

inc/Engine/Optimization/DynamicLists/DynamicLists.php
```diff
@@ -232,4 +235,45 @@ public function get_combine_js_excluded_inline(): array {
...
+	}
+
+	/**
+	 * Get all exclusions (merged together) for a list of item IDs.
+	 *
+	 * @param array $items List of items.
+	 *
+	 * @return array
+	 */
+	public function get_delayjs_items_exclusions( array $items ) {
```
mostafa-hisham: This function is not used anywhere in the code
@engahmeds3ed please have a look

inc/Engine/Optimization/DynamicLists/DynamicLists.php
```diff
@@ -232,4 +235,45 @@ public function get_combine_js_excluded_inline(): array {
...
 	}
+
+	/**
+	 * Get list of exclusions from the API list.
+	 *
+	 * @param string $item_id Item ID to get exclusions for (plugin slug, theme slug, script ID).
+	 *
+	 * @return array
+	 */
+	public function get_delayjs_exclusions_by_id( string $item_id ) {
```
mostafa-hisham: This function is not used anywhere in the code

assets/css/wpr-admin-common.css
```diff
@@ -1,25 +0,0 @@
...
-	min-width: 18px;
-	height: 18px;
-	border-radius: 9px;
-	background-color: #00A66B;
-	color: #fff;
-	font-size: 11px;
-	line-height: 1.6;
-	text-align: center;
-	z-index: 26;
-}
```
remyperona: why is this deleted?

assets/js/wpr-admin.min.js
```diff
@@ -1,2 +1,2 @@
...
...this.skipX=!0,null!=e.y?(this._addTween(this,"y",this.y,"max"===e.y?n(t,"y"):e.y,"scrollTo_y",!0),this._overwriteProps.push("scrollTo_y")):this.skipY=!0,!0},set:function(t){this._super.setRatio.call(this,t);var e=this._wdw||!this.skipX?this.getX():this.xPrev,i=this._wdw||!this.skipY?this.getY():this.yPrev,s=i-this.yPrev,t=e-this.xPrev;this._autoKill&&(!this.skipX&&(7<t||t<-7)&&n(this._target,"x")>e&&(this.skipX=!0),!this.skipY&&(7<s||s<-7)&&n(this._target,"y")>i&&(this.skipY=!0),this.skipX&&this.skipY&&this._tween.kill()),this._wdw?a.scrollTo(this.skipX?e:this.x,this.skipY?i:this.y):(this.skipY||(this._target.scrollTop=this.y),this.skipX||(this._target.scrollLeft=this.x)),this.xPrev=this.x,this.yPrev=this.y}}),e=t.prototype;t.max=n,e.getX=function(){return this._wdw?null!=a.pageXOffset?a.pageXOffset:(null!=r.scrollLeft?r:document.body).scrollLeft:this._target.scrollLeft},e.getY=function(){return this._wdw?null!=a.pageYOffset?a.pageYOffset:(null!=r.scrollTop?r:document.body).scrollTop:this._target.scrollTop},e._kill=function(t){return t.scrollTo_x&&(this.skipX=!0),t.scrollTo_y&&(this.skipY=!0),this._super._kill.call(this,t)}}),window._gsDefine&&window._gsQueue.pop()()},{}]},{},[2]);
```
remyperona, line 1: what are the changes in the unminified file?

inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -193,4 +211,21 @@ public static function exclusion_list_has_default(): bool {
 		return true;
 	}
 
+	/**
+	 * Refresh exclusions option based on selected items option.
+	 *
+	 * @return void
+	 */
+	public function refresh_exclusions_option() {
+		$current_selected_exclusions_selected_exclusion = get_rocket_option( 'delay_js_exclusions_selected_exclusions', '' );
```
remyperona: is there a reason to use `get_rocket_option()` instead of the `$options` instance in the class?

views/settings/fields/categorized_multiselect.php
```diff
@@ -0,0 +1,198 @@
...
+$wp_rocket_plugins = property_exists( $wp_rocket_items, 'plugins' ) ? $wp_rocket_items->plugins : new stdClass();
+
+$wp_rocket_textarea = get_rocket_option( esc_attr( $data['id'] ) );
+if ( is_array( $wp_rocket_textarea ) ) {
+	$wp_rocket_textarea = implode( "\n", $wp_rocket_textarea );
+}
+
+$wp_rocket_select            = get_rocket_option( esc_attr( $data['id'] ) . '_selected' );
+$wp_rocket_select_exclusions = get_rocket_option( esc_attr( $data['id'] ) . '_selected_exclusions' );
+$wp_rocket_state             = json_decode( $wp_rocket_select );
```
remyperona: that's a lot of processing in a view template, this could be in the main class and the final data passed directly

views/settings/fields/categorized_multiselect.php
```diff
@@ -0,0 +1,198 @@
...
+ * Fetch the icon.
+ *
+ * @param stdClass $item item from the list.
+ * @return string
+ */
+function fetch_icon( stdClass $item ) {
+	if ( property_exists( $item, 'icon_url' ) && $item->icon_url !== '' ) {
+		return esc_url( $item->icon_url );
+	}
+	return esc_url( WP_ROCKET_ASSETS_IMG_URL . 'default-icon.png' );
```
remyperona: we could use the dashicon instead of an image

remyperona (changes requested):

views/settings/fields/categorized_multiselect.php
```diff
@@ -0,0 +1,198 @@
...
+$wp_rocket_state = $wp_rocket_state ?: [];
+
+/**
+ * Fetch the icon.
+ *
+ * @param stdClass $item item from the list.
+ * @return string
+ */
+function fetch_icon( stdClass $item ) {
+	if ( property_exists( $item, 'icon_url' ) && $item->icon_url !== '' ) {
```
remyperona:
```suggestion
	if ( property_exists( $item, 'icon_url' ) && ! empty( $item->icon_url ) ) {
```

inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -77,6 +90,11 @@ public function sanitize_options( $input, $settings ) : array {
 		$input['delay_js']            = $settings->sanitize_checkbox( $input, 'delay_js' );
 		$input['delay_js_exclusions'] = ! empty( $input['delay_js_exclusions'] ) ? rocket_sanitize_textarea_field( 'delay_js_exclusions', $input['delay_js_exclusions'] ) : [];
 
+		if ( empty( $input['delay_js_exclusions_selected'] ) ) {
+			$input['delay_js_exclusions_selected']            = '';
+			$input['delay_js_exclusions_selected_exclusions'] = '';
```
remyperona: shouldn't it be an array instead of an empty string?

assets/css/wpr-admin-common.css
CrochetFeve0251: @Tabrisrp I guess this is deleted as we weren't compiling non minified assets before they were a bit outdated.

views/settings/fields/categorized_multiselect.php
CrochetFeve0251: We use Image sent by the server so we can't change it to an icon

assets/js/wpr-admin.min.js
CrochetFeve0251: @Tabrisrp JS used to create interaction inside lists.

inc/Engine/Optimization/DynamicLists/DynamicLists.php
CrochetFeve0251: @engahmeds3ed will add the part where it is used. (I left it there as I was thinking this part was a part of the last issue and that seemed total waste of time to drop it to readd it later)

assets/css/wpr-admin-common.css
remyperona: it's still css that we need it should not be removed

remyperona: locally it shows the file with content, not sure what GH is on, but we can ignore that for now

inc/Engine/Admin/Settings/Render.php
```diff
@@ -314,6 +332,21 @@ public function number( $args ) {
...
 	 */
 	public function categorized_multiselect( $args ) {
+		if ( is_array( $args['value'] ) ) {
+			$args['value'] = implode( "\n", $args['value'] );
+		}
+
+		$args['value'] = empty( $args['value'] ) ? '' : $args['value'];
+		$args['items'] = empty( $args['items'] ) ? new stdClass() : $args['items'];
+
+		$wp_rocket_textarea = get_rocket_option( esc_attr( $args['id'] ) );
```
remyperona:
```suggestion
		$wp_rocket_textarea = get_rocket_option( sanitize_key( $args['id'] ) );
```

inc/Engine/Admin/Settings/Render.php
```diff
@@ -314,6 +332,21 @@ public function number( $args ) {
...
+		if ( is_array( $wp_rocket_textarea ) ) {
+			$wp_rocket_textarea = implode( "\n", $wp_rocket_textarea );
+		}
+
+		$args['wp_rocket_textarea'] = $wp_rocket_textarea;
+		$args['selected']           = get_rocket_option( esc_attr( $args['id'] ), [] );
```
remyperona:
```suggestion
		$args['selected']           = get_rocket_option( sanitize_key( $args['id'] ), [] );
```

inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -10,19 +10,28 @@
...
 	 */
-	private $dynamic_lists;
+	private $site_list;
+
+	/**
+	 * Options instance.
+	 *
+	 * @var Options_Data
+	 */
+	protected $option;
```
remyperona:
```suggestion
	protected $options;
```

inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -10,19 +10,28 @@
...
 	/**
 	 * Constructor.
 	 *
-	 * @param DynamicLists $dynamic_lists DynamicLists instance.
+	 * @param SiteList     $site_list DelayJS Site List instance.
+	 * @param Options_Data $option Options instance.
 	 */
-	public function __construct( DynamicLists $dynamic_lists ) {
-		$this->dynamic_lists = $dynamic_lists;
+	public function __construct( SiteList $site_list, Options_Data $option ) {
```
remyperona:
```suggestion
	public function __construct( SiteList $site_list, Options_Data $options ) {
```

inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -10,19 +10,28 @@
...
+		$this->site_list = $site_list;
+		$this->option    = $option;
```
remyperona:
```suggestion
		$this->options   = $options;
```

inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -61,7 +70,7 @@ public function set_option_on_update( $old_version ) {
 			return;
 		}
 
-		$options = get_option( 'wp_rocket_settings', [] );
+		$options = $this->option->get( 'wp_rocket_settings', [] );
```
remyperona:
```suggestion
		$options = $this->options->get( 'wp_rocket_settings', [] );
```

inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -91,8 +100,10 @@ public function sanitize_options( $input, $settings ) : array {
 		$input['delay_js_exclusions'] = ! empty( $input['delay_js_exclusions'] ) ? rocket_sanitize_textarea_field( 'delay_js_exclusions', $input['delay_js_exclusions'] ) : [];
 
 		if ( empty( $input['delay_js_exclusions_selected'] ) ) {
-			$input['delay_js_exclusions_selected']            = '';
-			$input['delay_js_exclusions_selected_exclusions'] = '';
+			$input['delay_js_exclusions_selected']            = [];
+			$input['delay_js_exclusions_selected_exclusions'] = [];
+		} else {
+			$input['delay_js_exclusions_selected_exclusions'] = $this->site_list->get_delayjs_items_exclusions( $input['delay_js_exclusions_selected'] );
 		}
```
remyperona: I'd suggest to move this to its own sanitize method in the SiteList class, and we can use the filter to hook the sanitize method

inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -217,15 +228,14 @@ public static function exclusion_list_has_default(): bool {
 	 * @return void
 	 */
 	public function refresh_exclusions_option() {
-		$current_selected_exclusions_selected_exclusion = get_rocket_option( 'delay_js_exclusions_selected_exclusions', '' );
-
-		if ( empty( $current_selected_exclusions_selected_exclusion ) ) {
-			update_rocket_option( 'delay_js_exclusions_selected_exclusions', '' );
-
+		$selected_items = $this->option->get( 'delay_js_exclusions_selected' );
```
remyperona:
```suggestion
		$selected_items = $this->options->get( 'delay_js_exclusions_selected', [] );
```

inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -217,15 +228,14 @@ public static function exclusion_list_has_default(): bool {
...
+		if ( empty( $selected_items ) ) {
+			$this->option->set( 'delay_js_exclusions_selected_exclusions', [] );
```
remyperona:
```suggestion
			$this->options->set( 'delay_js_exclusions_selected_exclusions', [] );
```

The Options_Data is not enough, we also need the options api class instance to save the settings

inc/Engine/Optimization/DelayJS/Admin/Settings.php
```diff
@@ -217,15 +228,14 @@ public static function exclusion_list_has_default(): bool {
...
 			return;
 		}
 
-		update_rocket_option( 'delay_js_exclusions_selected_exclusions', $current_selected_exclusions_selected_exclusion );
+		$selected_exclusions = $this->site_list->get_delayjs_items_exclusions( $selected_items );
+		$this->option->set( 'delay_js_exclusions_selected_exclusions', $selected_exclusions );
```
remyperona:
```suggestion
		$this->options->set( 'delay_js_exclusions_selected_exclusions', $selected_exclusions );
```

The Options_Data is not enough, we also need the options api class instance to save the settings

inc/Engine/Optimization/DelayJS/Admin/SiteList.php
```diff
@@ -0,0 +1,209 @@
...
+	 * @param array $items List of items.
+	 *
+	 * @return array
+	 */
+	public function get_delayjs_items_exclusions( array $items ) {
+		$exclusions = [];
+
+		foreach ( $items as $item ) {
+			$exclusions = array_merge( $exclusions, $this->get_delayjs_exclusions_by_id( $item ) );
+		}
```
remyperona, line 134: It would be more performant to create 3 different arrays for each item, and then do one array_merge with all of them instead of doing 3 array_merge

remyperona (changes requested):

inc/Engine/Admin/Settings/Render.php
engahmeds3ed: Oh, I believe we don't need this part anyway because I split them into two fields

inc/Engine/Optimization/DelayJS/Admin/SiteList.php
```diff
@@ -206,4 +229,40 @@ private function get_icon( $item ) {
...
+	public function sanitize_options( $input ) : array {
+		if ( empty( $input['delay_js_exclusions_selected'] ) ) {
+			$input['delay_js_exclusions_selected']            = [];
+			$input['delay_js_exclusions_selected_exclusions'] = [];
+		} else {
+			$input['delay_js_exclusions_selected_exclusions'] = $this->get_delayjs_items_exclusions( $input['delay_js_exclusions_selected'] );
+		}
+
+		return $input;
+	}
```
remyperona: You can return early instead of using an else

remyperona (changes requested):

remyperona:
Tests need to be updated

inc/Engine/Optimization/DelayJS/Admin/SiteList.php
engahmeds3ed: it's not 3 array_merge, we are passing here the selected exclusions from our options so we don't know the type of each selected exclusion we only have their IDs.
what do u think?

PR #5565, opened 2022-11-14
Closes #5509 never cache url exclusions with permanlinks trailing slash

### LINKED ISSUE #5509: Never Cache URL(s) exclusions work only when they match the existence or not of a trailing slash in permalinks
- Made sure you’re on the latest version ✅`3.12.1.1`

**Describe the bug**

When adding URLs to the "Never Cache URL(s)" text area, for them to be excluded they will have to have a trailing slash or don't have one depending on whether one exists or not in the permalinks.

There is a **Should you use the trailing slash?** section in [Exclude Pages from the Cache and Optimizations](https://docs.wp-rocket.me/article/54-exclude-pages-from-the-cache) which describes how URLs should be added to the specific text area for the exclusions to work.

Users will add a trailing `/` or not, primarily based on habit, i.e. the way they write or paste URLs. Unless they read our doc beforehand, this may end up in URLs not being excluded which can be confusing/frustrating both for the customers and support.

**To Reproduce**
1. When permalinks have a trailing `/` exclude a URL without one.
2. The URL will be cached.
3. Remove the trailing `/` from the permalinks and exclude a URL that has one.
4. The URL will be cached.

**Expected behavior**

A URL added in the "Never Cache URL(s)" text area should be excluded regardless of whether it has a trailing `/` or not and the way the permalinks are set up.

**Additional context**

With WP Rocket `3.12.2` we are going to allow users to exclude URLs from preloading. "Never cache URLs" should be excluded from preloading as well but this won't happen given that the URLs may not be excluded from caching if they aren't formed correctly as far as the trailing `/` goes.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description
Fixes the issue of not excluding some uri's in the Never Cache Urls due to non matching permalink structure

Fixes #5509 

## Type of change

## How Has This Been Tested?
Automated and Manual Test

# Checklist:


inc/functions/options.php
```diff
@@ -235,6 +235,15 @@ function get_rocket_cache_reject_uri( $force = false, $show_safe_content = true
 		return '';
 	}
 
+	// Get permalink structure.
+	$trailing_slash = '/' === substr( get_option( 'permalink_structure' ), -1 ) ? '/' : '';
+	$uris           = array_map(
+		function ( $uri ) use ( $trailing_slash ) {
+			return rtrim( $uri, '/' ) . $trailing_slash;
```
mostafa-hisham: I think we can use `user_trailingslashit` function here instead

mostafa-hisham (approved):

vmanthos (approved):
Working as expected. 👍

PR #5579, opened 2022-11-17
Fix DB not found columns error in cache table with upgrade

engahmeds3ed:
## Description

When the user upgrades and before he refreshes any admin page, and the preload is triggered (automatically or by a frontend visit), so the new code after upgrade will be working with preload but the database still old because we trigger the database changes with `admin_init` hook, so here we will trigger this again with upgrade, so no need to refresh the admin or visit any admin page.

## Type of change

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

### DIFF
--- a/inc/Engine/Cache/PurgeExpired/Subscriber.php
+++ b/inc/Engine/Cache/PurgeExpired/Subscriber.php
@@ -54,11 +54,12 @@ class Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'init'                => 'schedule_event',
-			'rocket_deactivation' => 'unschedule_event',
-			static::EVENT_NAME    => 'purge_expired_files',
-			'cron_schedules'      => 'custom_cron_schedule',
-			'wp_rocket_upgrade'   => [ 'update_lifespan_option_on_update', 13, 2 ],
+			'init'                                   => 'schedule_event',
+			'rocket_deactivation'                    => 'unschedule_event',
+			static::EVENT_NAME                       => 'purge_expired_files',
+			'cron_schedules'                         => 'custom_cron_schedule',
+			'wp_rocket_upgrade'                      => [ 'update_lifespan_option_on_update', 13, 2 ],
+			'rocket_after_automatic_cache_purge_dir' => 'run_upgrade_routine',
 		];
 	}
 
@@ -180,5 +181,14 @@ class Subscriber implements Subscriber_Interface {
 			$this->options->get( 'purge_cron_interval' ),
 			$this->options->get( 'purge_cron_unit' )
 		);
+	}
+
+	/**
+	 * With automatic cache purge we need to run the upgrade routine.
+	 *
+	 * @return void
+	 */
+	public function run_upgrade_routine() {
+		rocket_upgrader();
 	}
 }
--- a/inc/Engine/Cache/PurgeExpired/Subscriber.php
+++ b/inc/Engine/Cache/PurgeExpired/Subscriber.php
@@ -54,12 +54,11 @@ class Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'init'                                   => 'schedule_event',
-			'rocket_deactivation'                    => 'unschedule_event',
-			static::EVENT_NAME                       => 'purge_expired_files',
-			'cron_schedules'                         => 'custom_cron_schedule',
-			'wp_rocket_upgrade'                      => [ 'update_lifespan_option_on_update', 13, 2 ],
-			'rocket_after_automatic_cache_purge_dir' => 'run_upgrade_routine',
+			'init'                => 'schedule_event',
+			'rocket_deactivation' => 'unschedule_event',
+			static::EVENT_NAME    => 'purge_expired_files',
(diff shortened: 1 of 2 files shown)


remyperona (approved):

inc/Engine/Cache/PurgeExpired/Subscriber.php
```diff
@@ -54,11 +54,12 @@ public function __construct( Options_Data $options, PurgeExpiredCache $purge ) {
...
-			'rocket_deactivation' => 'unschedule_event',
-			static::EVENT_NAME    => 'purge_expired_files',
-			'cron_schedules'      => 'custom_cron_schedule',
-			'wp_rocket_upgrade'   => [ 'update_lifespan_option_on_update', 13, 2 ],
+			'init'                                   => 'schedule_event',
+			'rocket_deactivation'                    => 'unschedule_event',
+			static::EVENT_NAME                       => 'purge_expired_files',
+			'cron_schedules'                         => 'custom_cron_schedule',
+			'wp_rocket_upgrade'                      => [ 'update_lifespan_option_on_update', 13, 2 ],
+			'rocket_after_automatic_cache_purge_dir' => 'run_upgrade_routine',
```
remyperona: that won't work for everyone because you can disable automatic cache clearing when setting cache clean lifespan to zero

remyperona (changes requested):

inc/Engine/Cache/PurgeExpired/Subscriber.php
engahmeds3ed: so do u think that this may work if added to `after_rocket_clean_domain` for example or what do u suggest?

engahmeds3ed: Also what I'm trying to achieve is to fix this for most of customers, for sure we may miss some users with some configurations and for those, the main upgrade routine will happen when they open the admin page.

inc/Engine/Preload/Database/Tables/Cache.php
```diff
@@ -14,6 +14,7 @@ class Cache extends Table {
 	public function __construct() {
 		parent::__construct();
 		add_action( 'rocket_preload_activation', [ $this, 'maybe_upgrade' ] );
+		add_action( 'wp_rocket_upgrade', [ $this, 'maybe_upgrade' ] );
```
remyperona: why is this one not enough to fix the issue? This should be triggered on every load, front and back, right?

engahmeds3ed: No, `wp_rocket_upgrade` is fired with `admin_init` here:
https://github.com/wp-media/wp-rocket/blob/d89551a8af8d21e90b74861f97f5317b326c957d/inc/admin/upgrader.php#L56

and here:

https://github.com/wp-media/wp-rocket/blob/d89551a8af8d21e90b74861f97f5317b326c957d/inc/admin/upgrader.php#L21

remyperona: how about hooking it on init in there instead of wp_rocket_upgrade? That way it will run on front too

engahmeds3ed: Yes this is one of the solutions that I mentioned here: https://wp-media.slack.com/archives/GUT7FLHF1/p1668776723731869

this will work but do u think it'll cause any performance issues?

remyperona: it won't be noticeable, the upgrade method bails out early if there is nothing to do

inc/Engine/Cache/PurgeExpired/Subscriber.php
remyperona:
```suggestion
```

inc/Engine/Cache/PurgeExpired/Subscriber.php
```diff
@@ -181,4 +182,13 @@ public function update_lifespan_option_on_update( $new_version, $old_version ) {
...
 	}
+
+	/**
+	 * With automatic cache purge we need to run the upgrade routine.
+	 *
+	 * @return void
+	 */
+	public function run_upgrade_routine() {
+		rocket_upgrader();
+	}
```
remyperona:
```suggestion
```

inc/Engine/Preload/Database/Tables/Cache.php
remyperona:
```suggestion
		add_action( 'init', [ $this, 'maybe_upgrade' ] );
```

remyperona (changes requested):

inc/Engine/Preload/Database/Tables/Cache.php
engahmeds3ed: Awesome, let's do it like that and see how it goes with testing

engahmeds3ed:
@wp-media/qa I think we are good to test and validate this.

vmanthos (commented):
With the latest changes, the `is_locked` column will be added if a visitor accesses a non-cached page.

But still, it's not added in these cases:
* no admin action takes place after the update (the `plugins.php` page isn't focused or is closed)
* visitors visit pages that are already cached

@wp-media/productrocket Is this OK?

remyperona (approved):

Issue #491, opened 2017-07-28
The cache of URLs with query strings is not cleared when permalinks don't have a trailing slash


foliovision:
We use Advanced -> Cache query strings to cache some URLs with query strings. I enter "test" into that and then not only https://foliovision.com/2017/05/improved-video-loading-code is cached, but also https://foliovision.com/2017/05/improved-video-loading-code?test=1

Problem is that when I update the post the cache for https://foliovision.com/2017/05/improved-video-loading-code is cleared, but for https://foliovision.com/2017/05/improved-video-loading-code?test=1 it's not cleared.

Clearly a bug. Please check our pull request: #490

Tabrisrp:
I tested the issue on our test installation, and I can't replicate.

When updating the post, the caches for both the default version and the query string are correctly cleared.

Please test on a clean installation. If you're having this issue on this particular website, you can open a support ticket to understand why it's happening in this case.

GeekPress:
Hi @foliovision 🤚

We haven't back from you since @Tabrisrp asked you a question, so I'm closing this issue. However, if you still need our help, just reply here and we will re-open the issue.

Best

vmanthos:
I'm reopening this one as I found the conditions under which this issue occurs.

**To reproduce:**
1. **Permalink settings:** remove the trailing `/`
![](https://i.vgy.me/87Ti2b.jpg)

2. Add `location` in the _Advanced Rules tab > Cache Query String(s)_ .

Visiting https://example.com/hello-world

the cache is created in `/cache/wp-rocket/example.com/hello-world/`.

Visiting:
* https://example.com/hello-world?location=Athens
* https://example.com/hello-world?location=Thessaloniki
the cache is created respectively in:
```
/cache/wp-rocket/example.com/hello-world#location=Athens
/cache/wp-rocket/example.com/hello-world#location=Thessaloniki
```
![](https://i.vgy.me/9CWu5U.jpg)

That is _NOT_ in a subfolder of the main `/hello-world/` post.
When clearing the cache of `/hello-world/` the cache of the other folders isn't cleared.

The following is happening when permalink settings _have_ a trailing `/`:

Visiting:
* https://example.com/hello-world/?location=Athens
* https://example.com/hello-world/?location=Thessaloniki

the cache is created respectively in:
```
/cache/wp-rocket/example.com/hello-world/#location=Athens
/cache/wp-rocket/example.com/hello-world/#location=Thessaloniki
```
That is in subfolders of the main post.

[...]

PR #5581, opened 2022-11-18
Closes #4773 Clear cache partially for never cache urls

### LINKED ISSUE #4773: Clear cache partially instead of globally when Never cache This Page is changed in the Edit screen metabox
- Made sure you’re on the latest version - 3.10.8

**Describe the bug**
When Never Cache This Page option is changed in Edit screen metabox, we clear cache for the entire domain. This is unnecessary and not desired.

**To Reproduce**
1. Go to Edit Screen setting
2. Change metabox value "Never cache this page" from true to false
3. Click Update
4. Entire cache will be cleared

**Expected behavior**
Cache should be cleared only for the related URL where the setting is being changed.

"Never Cache URL(s)" should be handled the same way, i.e. clear the cache only of the new URL(s) that's added in the field.

**Additional context**
By @engahmeds3ed 

> The journey starts [here](https://github.com/wp-media/wp-rocket/blob/d833ad2e34dab92e3e58838fedf4b4d643d84ce8/inc/admin/ui/meta-boxes.php#L134-L138) when saving the post metabox values, when never cache this page option is changed (from true to false or from false to true) this will update our options [here](https://github.com/wp-media/wp-rocket/blob/d833ad2e34dab92e3e58838fedf4b4d643d84ce8/inc/admin/ui/meta-boxes.php#L146) specifically the option named Never Cache URL(s) to have this page url and then this triggers [this callback](https://github.com/wp-media/wp-rocket/blob/bbadfba54d3a43501305659207b92840e480b417/inc/admin/options.php#L104) .

Slack: https://wp-media.slack.com/archives/C43T1AYMQ/p1646124958398979
Ticket: https://secure.helpscout.net/conversation/1787123848/326047/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description
Closes issue of global cache clearing for Never Cache URL(s)

Fixes #4773 

## Type of change

Slightly

We also needed to consider patterns and also get the full url using the path.

## How Has This Been Tested?

Automated & Manual Test

# Checklist:


remyperona (changes requested):
I think that's a good start, but I would recommend 2 things to improve:
- Add this to a class instead of creating a new function (could be on the settings class and subscriber for example)
- Use BerlinDB (our DB dependency) to do the request to the cache table

inc/Engine/Preload/Database/Queries/Cache.php
```diff
@@ -250,6 +250,22 @@ public function get_rows_by_url( string $url ) {
...
+	 * @param string $regex Regular Expression.
+	 * @return array
+	 */
+	public function get_urls_by_regex( string $regex ): array {
+		$db = $this->get_db();
+
+		$prefixed_table_name = $db->prefix . $this->table_name;
+		$query               = "SELECT `url` FROM `$prefixed_table_name` WHERE `status` = 'completed' AND `url` REGEXP '$regex'";
+		$results             = $db->get_results( $query, ARRAY_N );
+
```
remyperona: @engahmeds3ed Do you think it's the best way to do this query using BerlinDB? You have more experience with it than me

inc/Engine/Cache/Purge.php
```diff
@@ -246,4 +256,50 @@ private function get_post_terms_urls( WP_Post $post ) {
...
 	}
+
+	/**
+	 * Purge single cache file(s) added in the Never Cache URL(s).
+	 *
+	 * @param array $old_value An array of previous values for the settings.
+	 * @param array $value An array of submitted values for the settings.
+	 * @return void
+	 */
+	public function purge_cache_reject_uri_partially( $old_value, $value ): void {
```
remyperona: In the subscriber you typecast the parameters, and here you don't, both should be the same for consistency

inc/Engine/Cache/Purge.php
```diff
@@ -246,4 +256,50 @@ private function get_post_terms_urls( WP_Post $post ) {
...
+		// Bail out if setting is not array.
+		if ( ! is_array( $old_value ) || ! is_array( $value ) ) {
+			return;
+		}
```
remyperona: if you typecast the parameters to array, this is not needed

remyperona (changes requested):

remyperona (approved):

vmanthos:
@jeawhanlee Doing exploratory testing I found that when in the **Never Cache URLs** text area I added `/hello(.*)`, although the correct URL was found here:
https://github.com/wp-media/wp-rocket/blob/e50db622d4b53767d2620a5d510018acfc4bea16/inc/Engine/Cache/Purge.php#L303

the whole cache was cleared.

remyperona (approved):

vmanthos:
This is working as expected in terms of not clearing the whole cache when adding URL(s) in the Never Cache URL(s) text area.

Because of [this issue](https://github.com/wp-media/wp-rocket/issues/491#issuecomment-1162916459) when:
1. Permalinks don't have a trailing `/`.
2. There are query strings, e.g. `location`, in the "Cache Query String(s)" text area.
3. `/hello-world?location=Paris` is cached (in `/cache/wp-rocket/example.com/hello-world#location=Paris`).

Excluding, e.g. `/hello-world` will not clear the cache file from step 3.

Regardless of this, visiting `/hello-world?location=Paris` will fetch the non-cached page. 

Unless the whole cache is cleared, removing `/hello-world` from the exclusion text area will result in serving the old cache file when visiting `/hello-world?location=Paris`.

@piotrbak Should we fix #491 before releasing this one?

piotrbak:
@vmanthos I think this is a bit edgy, we can move forward without that fix. For the query strings and clearing cache we already more issues, we should probably tackle all of them at once.

vmanthos (approved):
Following Piotr's comment, this PR is :+1: .

Issue #678, opened 2017-12-28
index.php file not processed correctly


pcfreak30:
There appears to be a bug that when you access `/index.php` it creates a `cache/wp-rocket/index.php/` cache folder instead of serving the root index cache file. this has a cascade effect on the rewrite fallback for when the htaccess rewrite for cache fails and go to `index.php`. the rewrite rules start over and reapply it, matching the bugged file that exists.

There needs to be a very early check to catch this and correct it.

arunbasillal:
Hey @pcfreak30 

Thanks for reporting this, I was able to spot this on couple of websites and confirm that this is indeed something to work on. 

The quick fix is to add `/index.php` to the "Never Cache URLs" in the Advanced Rules tab of WP Rocket. This way the `/cache/wp-rocket/domain.com/index.php` is never created and this doesn't happen. 

> this has a cascade effect on the rewrite fallback for when the htaccess rewrite for cache fails and go to index.php. the rewrite rules start over and reapply it, matching the bugged file that exists.

Can you please explain how this works and why this happens? Is this how Apache works?

(rest of the comment is to document the findings and not necessarily a direct reply to you. You already know what is happening.)

We have an issue where if `domain.com/index.php` is not redirected to `domain.com` (as with a standard WordPress install), then:

- WP Rocket creates a cache for index.php
- This is in `/wp-content/cache/wp-rocket/domain.com/index.php/index-https.html`
- Inner uncached pages on `domain.com` (like `domain.com/about` ) will display the homepage content.
- Our rewrite rule is the cause, not sure why -> `RewriteRule .* "/wp-content/cache/wp-rocket/%{HTTP_HOST}%{REQUEST_URI}/index%{ENV:WPR_SSL}.html%{ENV:WPR_ENC}" [L]`
- In theory for `domain.com/about`, the rewrite rule should translate to: `/wp-content/cache/wp-rocket/domain.com/about/index-https.html`
- But in effect it becomes `/wp-content/cache/wp-rocket/domain.com/index.php/index-https.html`

pcfreak30:
The [L] cause it to stop the rewrite logic (LAST) and repeat.

pcfreak30:
The rest I have forgotten the details sorry :P

arunbasillal:
I think I have a grip on this, thanks to your tip in your original comment @pcfreak30 !

[...]

PR #5604, opened 2022-12-01
Closes #5599 The /feed/ url is cached and optimized by default on 3.12.3.2

### LINKED ISSUE #5599: The /feed/ url is cached and optimized by default
yes- Made sure you’re on the latest version
yes - Used the search feature to ensure that the bug hasn’t been reported before

**Describe the bug**
On WP Rocket 3.12.3.2 `/feed/` is cached and optimized by default. rolling back to 3.11.5 fixes the issue.
this might be affecting `/wp-json/` too and probably other paths. (untested)

**To Reproduce**
1. Install 3.12.3.2
2. Visit the /feeds URL and see caching and optimization applied
3. Rollback to 3.11.5 and clear the cache
4. Visit the /feeds URL again,  caching and optimization are NOT applied

**Expected behavior**
Feeds should not be cached unless the user opts for this. And optimizations shouldn't be applied in any case.

**Screenshots**
IM3zjAV

https://i.imgur.com/IM3zjAV.jpg

**Additional context**
ticket where /feed/ is cached https://secure.helpscout.net/conversation/2079965213/385018/
probably related, about /wp-json/: https://secure.helpscout.net/conversation/2082507658/385784?folderId=273761

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

jeawhanlee:
## Description
Fixes the issue where trailing slash is appended to non user defined patterns for cache_reject_uri.

Fixes #5599 

## Type of change

No

## How Has This Been Tested?
Automated and Manual Tests.

# Checklist:


remyperona (changes requested):
The value in the option should be updated before saving it to the DB, not when serving it from the DB.

We want the change to be reflected in the UI for this field

inc/Engine/Cache/Config/ConfigSubscriber.php
```diff
@@ -47,6 +47,29 @@ function ( $uri ) {
 	 * @return void
 	 */
 	public function regenerate_config_file() {
+		$cache_reject_uri = $this->match_pattern_with_permalink_structure( get_rocket_option( 'cache_reject_uri', [] ) );
+		update_rocket_option( 'cache_reject_uri', $cache_reject_uri );
```
remyperona: We could use our Options and Options_Data classes instead of the helper functions

inc/Engine/Cache/Config/ConfigSubscriber.php
```diff
@@ -47,6 +47,29 @@ function ( $uri ) {
...
+
 		rocket_generate_config_file();
 	}
+
+	/**
+	 * Modify cache_reject_uri values.
+	 *
+	 * @param array|mixed $value New values.
+	 * @param array|mixed $old_value Old values.
```
remyperona:
```suggestion
	 * @param mixed $value The new, unserialized option value.
	 * @param mixed $old_value The old option value.
```

remyperona (changes requested):

remyperona (approved):

vmanthos:
@jeawhanlee I run tests and found that when:
1. the permalinks don't have a `/` and
2. in the **Never Cache URLs** text area `/` was added to prevent caching of the homepage

the `/` is removed so caching of the homepage takes place. This isn't happening with `trunk` or when the permalinks include a trailing `/`.

**Note**

This is to have a reference of the [the index.php issue](https://github.com/wp-media/wp-rocket/issues/678) reported [here](https://github.com/wp-media/wp-rocket/issues/5599#issuecomment-1335678072).

When permalinks:
1. have a trailing `/`and 
2. the `/cache/wp-rocket/example.com/index.php/` folder  is created manually, the non-cached pages are redirected to the homepage.

**Slack discussion:** https://wp-media.slack.com/archives/GUT7FLHF1/p1670251431739879

mostafa-hisham (approved):

inc/Engine/Cache/Config/ConfigSubscriber.php
```diff
@@ -2,29 +2,107 @@
...
 		];
 	}
 
+	/**
+	 * Returns a matching user added patterns with permalink structure
+	 *
+	 * @param array $patterns user pattern.
+	 * @return array
+	 */
+	public function match_pattern_with_permalink_structure( array $patterns ): array {
```
engahmeds3ed:
```suggestion
	private function match_pattern_with_permalink_structure( array $patterns ): array {
```

inc/Engine/Cache/Config/ConfigSubscriber.php
```diff
@@ -2,29 +2,107 @@
...
 	/**
 	 * Return an array of events that this subscriber wants to listen to.
 	 *
 	 * @return array
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'permalink_structure_changed' => 'regenerate_config_file',
+			'permalink_structure_changed'         => 'regenerate_config_file',
+			'pre_update_option_' . WP_ROCKET_SLUG => [ 'change_cache_reject_uri_with_permalink', 10, 2 ],
```
engahmeds3ed:
```suggestion
			'pre_update_option_' . rocket_get_constant( 'WP_ROCKET_SLUG', 'wp_rocket_settings' ) => [ 'change_cache_reject_uri_with_permalink', 10, 2 ],
```
what do u think

tests/Fixtures/inc/Engine/Cache/Config/ConfigSubscriber/changeCacheRejectUriWithPermalink.php
```diff
@@ -0,0 +1,101 @@
+<?php
```
engahmeds3ed, line 1: (That's a dev matter so it won't affect the PR itself and not a blocker)
I know that you need to write less code and not to duplicate ourselves but this fixture isn't readable for me, I believe you need to make it simpler by writing the pattern manually for each case below, so it'd be easier for anyone to add cases whenever u want

engahmeds3ed (changes requested):

vmanthos (approved):
This is working as expected now.

**Note**
[The exclusion](https://github.com/wp-media/wp-rocket/issues/5599#issuecomment-1335678072) of `index.php` will not work in [some cases](https://github.com/wp-media/wp-rocket/pull/5604#issuecomment-1338964373).

After internal discussion, we decided that that's fine since we recommend excluding `/index.php` in our docs.

PR #5609, opened 2022-12-05
Possible fix for issue #5608 (ImageDimensions::is_external_file returns false for external files...)

### LINKED ISSUE #5608: ImageDimensions::is_external_file returns false for external files...
**Describe the bug**
When the image URL contains the site domain somewhere in the middle - ImageDimensions::is_external_file() returns false, even though it's actually is external (e.g., when using another CDN URL that wraps the URLs with it's own URL).
As a result, wp-rocket fails to add the image dimensions on it.

**To Reproduce**
1. Enable "Image Dimensions" on wp-rocket.
2. Add somewhere on the theme an image tag with a URL that looks like:
``
3. Note that on the website front-end, wp-rocket doesn't add the image dimensions to this image.

**Expected behavior**
Image dimensions are being added to this image.

**Additional context**
On the `is_external_file` method, there's [this](https://github.com/wp-media/wp-rocket/blob/develop/inc/Engine/Media/ImageDimensions/ImageDimensions.php#L214) code:
```
// URL has domain and domain is part of the internal domains.
if ( ! empty( $file['host'] ) ) {
	foreach ( $hosts as $host ) {
		if ( false !== strpos( $url, $host ) ) {
			return false;
		}
	}

	return true;
}
```
Here it returns false, because strpos is looking for the host *anywhere*  inside the given URL and not only in the beginning of it (or it's host).
Probably something like `if ( false !== strpos( $file['host'], $host ) ) {` would eventually return true and this URL would have treated as it should have - external.

*Opened a PR with a fix that solved it in my local environment: #5609

pniel-cohen:
A fix that should solve #5608.

### DIFF
--- a/inc/Engine/Media/ImageDimensions/ImageDimensions.php
+++ b/inc/Engine/Media/ImageDimensions/ImageDimensions.php
@@ -214,7 +214,7 @@ class ImageDimensions {
 		// URL has domain and domain is part of the internal domains.
 		if ( ! empty( $file['host'] ) ) {
 			foreach ( $hosts as $host ) {
-				if ( false !== strpos( $url, $host ) ) {
+				if ( false !== strpos( $file['host'], $host ) ) {
 					return false;
 				}
 			}


piotrbak:
@pniel-cohen Thanks for creating the PR, we'll review and test it. If things go correctly, it'll be included in one of the upcoming minor versions. 🙏

remyperona (approved):

vmanthos (approved):

PR #5615, opened 2022-12-08
Closes #5519 Never cache this page is not applied immediately after saving a post

### LINKED ISSUE #5519: Never cache this page is not applied immediately after saving a post

**Describe the bug**

When activating the "Never cache this page" option in WP Rocket's meta-box, e.g. for `/page-1/`, the change isn't added to the configuration file after saving the post despite the file being regenerated. This means the page can still be cached.

It will be added to the config file the next time it is regenerated, e.g. when the option is selected for another post/page, e.g. `/page-2/`. In this case, only `/page-1/` will be added, but not `/page-2/`.

I've confirmed the URI is added at this point:
https://github.com/wp-media/wp-rocket/blob/c417af78b06be599a0be626fa4586a04c105133c/inc/admin/ui/meta-boxes.php#L148

but when `rocket_generate_config_file()` runs it doesn't include it.
 
[get_rocket_cache_reject_uri()](https://github.com/wp-media/wp-rocket/blob/c417af78b06be599a0be626fa4586a04c105133c/inc/functions/options.php#L247), which is used when [regenerating the config file](https://github.com/wp-media/wp-rocket/blob/c417af78b06be599a0be626fa4586a04c105133c/inc/functions/files.php#L126), runs several times after updating the post. In [one of them](https://github.com/wp-media/wp-rocket/blob/c417af78b06be599a0be626fa4586a04c105133c/inc/Engine/Preload/Links/Subscriber.php#L177), the newly excluded URI exists, but when the `save_post` hook fires and we regenerate the config file it doesn't.

Although I've debugged this extensively I wasn't able to figure out why this is happening.

**To Reproduce**

1. Enable "Never cache this page" for a post.
2. Visit the post on the front end.
3. WP Rocket's signature will be there / a cached file will be in `/cache/wp-rocket/`.

**Expected behavior**

The URI should be excluded immediately after saving the post/page.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

jeawhanlee:
## Description
Fixes the issue where posts added to never cache uri from the metabox was not applied immediately.

Fixes #5519 

## Type of change

No

## How Has This Been Tested?
Manual Test

# Checklist:

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -123,7 +123,7 @@ function get_rocket_config_file() { // phpcs:ignore WordPress.NamingConventions.
 		}
 
 		if ( 'cache_reject_uri' === $option ) {
-			$buffer .= '$rocket_' . $option . ' = \'' . get_rocket_cache_reject_uri() . "';\n";
+			$buffer .= '$rocket_' . $option . ' = \'' . get_rocket_cache_reject_uri( true ) . "';\n";
 		}
 
 		if ( 'cache_query_strings' === $option ) {


mostafa-hisham (approved):

vmanthos (approved):
Working as expected! 👍

Issue #2974, opened 2020-08-08
Regex gets stripped as comment in minified JS file


girlie:
Two double slashes in Regex can be interpreted as comment markers -- even though the first one is escaped -- and subsequent JS is then stripped from the resulting minified file.

i.e., it starts out like this:

```
if(b)return w.IGNORE;var c=B(fb(a,"href"));b=R(fb(a,"hostname"));c=c?/^https?:\/\//.test(c)||/^\/\//.test(c):!1;if(c&&!lc(a)){if(la)return w.TRACK;
if(!0===p.aff_domains[b]||H(lb,b))return w.AFFILIATE;if(Ka(b))return w.PRE_AFFILIATED;
```

but it ends up like this:

```
if(b)return w.IGNORE;var c=B(fb(a,"href"));b=R(fb(a,"hostname"));c=c?/^https?:\/\
if(!0===p.aff_domains[b]||H(lb,b))return w.AFFILIATE;if(Ka(b))return w.PRE_AFFILIATED;
```

(Note that in the original ticket, this code was contained within an external file, so the problem didn't become apparent until Combine JS was enabled.) 

Steps to reproduce:

1. Upload the attached _.txt_ file to your theme folder, then rename it to have a _.js_ extension.
2. Enqueue the file via your theme's _functions.php_.
3. Enable JS minification in WP Rocket (but not Combine JS, to make the next step easier).
4. Inspect the final code in the resulting minified file, to see how it's changed.

Related issues from the third party minify library repo:
* https://github.com/matthiasmullie/minify/issues/323
* https://github.com/matthiasmullie/minify/issues/329

Pull request which might be the fix:
https://github.com/matthiasmullie/minify/pull/330

Original ticket: https://secure.helpscout.net/conversation/1241574590/183879
Slack discussion: https://wp-media.slack.com/archives/C43T1AYMQ/p1596652561163300

[166124X1633667.skimlinks.txt](https://github.com/wp-media/wp-rocket/files/5045415/166124X1633667.skimlinks.txt)

enricodias:
I just updated the PR matthiasmullie/minify#330 to include this case, but it I don't know when (and if) it will be merged.

jazir555:
I have a feeling this is deeply related to the Elementor JS file loading issue. I'm not a coder, but it likely that this could be a reason WP-Rocket is ignoring those files being dequeued with the new elementor asset loading experiment enabled. The feature is explicitly for JS files.

[...]

PR #5627, opened 2022-12-14
Closes #5617 Update Minify dependency with latest changes

### LINKED ISSUE #2083: CSS Variables getting 0px replaced with 0

**Describe the bug**
When using CSS variables, if they're having `0px` it's trimmed to `0` causing css `calc()` or other operations to break.

**To Reproduce**
##### Uncompressed:
```
:root {
  --some-variable: 0px;
}
:root[data-custom-value="1"] {
  --some-variable: 20px;
}
.some-class {
  height: calc( 100vh - var(--some-variable) );
}
```
##### Compressed:
```
:root {
  --some-variable: 0;
}
...
```
In this example if the html tag doesn't have the `data-custom-value` attribute, the `some-class` height calc operation will fallback on the the first declaration, which had, which breaks because the `calc` operation needs the `0px` , not `0` without the unit. 
The magic of CSS variables is obvious in this example. I don't have to redeclare chained classes to meet the conditions.

**Expected behavior**
The patterns are obvious to trim px in 0px declarations (which do make a lot of sense). Would be great to ignore CSS variables, or to have some sort of flag, eg: !wprocket-ignore;

### LINKED ISSUE #5617: Update Minify dependency with latest changes
We should update the minification library. 

[It'll](https://github.com/matthiasmullie/minify/commit/d8993df3953c2db59c12a713e9e5c927fe4c5426) fix the following issues:
- https://github.com/wp-media/wp-rocket/issues/2083

**Additional context**
We need to make sure we won't introduce regression described here:
- https://github.com/wp-media/wp-rocket/issues/2974

jeawhanlee:
## Description
Pulls [change](https://github.com/matthiasmullie/minify/commit/d8993df3953c2db59c12a713e9e5c927fe4c5426) from the minify lib that will fix the `px` unit that is removed. This usually happened with styles on a single line like `:root {--test-margin-offset: 0px}`

Fixes #2083
Fixes #5617 

## Type of change

NO

## How Has This Been Tested?
Manual Test

# Checklist:

### DIFF
--- a/inc/Dependencies/Minify/CSS.php
+++ b/inc/Dependencies/Minify/CSS.php
@@ -740,7 +740,7 @@ class CSS extends Minify
 		// PHP only supports $this inside anonymous functions since 5.4
 		$minifier = $this;
 		$this->registerPattern(
-			'/(?<=^|[;}])\s*(--[^:;{}"\'\s]+)\s*:([^;{}]+)/m',
+			'/(?<=^|[;}{])\s*(--[^:;{}"\'\s]+)\s*:([^;{}]+)/m',
 			function ($match) use ($minifier) {
 				$placeholder = '--custom-'. count($minifier->extracted) . ':0';
 				$minifier->extracted[$placeholder] = $match[1] .':'. trim($match[2]);


remyperona (approved):

Mai-Saad:
@jeawhanlee Thanks for the PR. Working as expected with the following notes.
- Only working for` --variable`, if we used `margin-left:0px` it will be `margin-left:0`
- The issue [here](https://github.com/wp-media/wp-rocket/issues/2974) is still not fixed 
[testrail-report-433.pdf](https://github.com/wp-media/wp-rocket/files/10236748/testrail-report-433.pdf)

PR #5628, opened 2022-12-14
Fix 5619: Added new logic for the preload queue

### LINKED ISSUE #5619: Improve the logic of adding URLs to the preload queue
- Made sure you’re on the latest version 🆗 
- Used the search feature to ensure that the bug hasn’t been reported before 🆗 

**Describe the bug**
This logic is not correct:
https://github.com/wp-media/wp-rocket/blob/36f4b69222ac01aba6e267b1d950f5fb639f53d7/inc/Engine/Preload/Database/Queries/Cache.php#L320

We're looking for the entries in the database that are `in-progress`, if for any reason we reach the defined number, the preload process will be stuck until the cache is cleared or our 12h cron `rocket_preload_revert_old_in_progress_rows` runs.

**To Reproduce**
1. Enable preload
2. Change status of 45 URLs to `in-progress` in the database
3. The preload is stuck

**Expected behavior**
The preload should not be stuck then.

**Additional context**
We introduced this limit to make sure that we are not flooding AS with our actions and we allow 3rd-parties to run their own actions in timely manner.

Instead of [checking](https://github.com/wp-media/wp-rocket/blob/36f4b69222ac01aba6e267b1d950f5fb639f53d7/inc/Engine/Preload/Database/Queries/Cache.php#L320) the number of `in-progress` items in the database, we could check the number of `rocket_preload_job_preload_url` actions that are pending in the AS.

We could also consider other ways of fixing this problem, the discussion:
https://wp-media.slack.com/archives/GUT7FLHF1/p1670860166280739

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description
Fix an issue where the queue can be stuck by in-progress tasks that failed.

For that we added a new status that prevent urls from accumulating into the queue.

Fixes #5619

## Type of change

No.

## How Has This Been Tested?

- [ ] Automated Test 

# Checklist:


jeawhanlee (approved):

inc/Engine/Preload/Database/Queries/Cache.php
```diff
@@ -455,6 +457,22 @@ public function revert_old_in_progress() {
 		$db->query( "UPDATE `$prefixed_table_name` SET status = 'pending' WHERE status = 'in-progress' AND `modified` <= date_sub(now(), interval 12 hour)" );
 	}
 
+	/**
+	 * Revert old in-progress rows.
+	 */
+	public function revert_old_failed() {
```
remyperona: Maybe update the method name and description to `set_failed()` or something like that

remyperona (changes requested):

inc/Engine/Preload/Database/Queries/Cache.php
CrochetFeve0251: I updated the doc that wasn't up to date.

jeawhanlee (approved):

Mai-Saad:
We need to increase the 2mins and get feedback [here](https://wp-media.slack.com/archives/GUT7FLHF1/p1672757821546629).

piotrbak:
I'm thinking, maybe we could make the [value](https://github.com/wp-media/wp-rocket/pull/5628/files#diff-30a46cf1cdbb73aeb3f9cc7171e0088ad71307d47bd27d32e548e2aab0f1d540R650) of `2 minutes` auto-adjustable and also introduce the filter there. Let's say the value could be:

`rocket_preload_cache_pending_jobs_cron_rows_count` / `15` = [value](https://github.com/wp-media/wp-rocket/pull/5628/files#diff-30a46cf1cdbb73aeb3f9cc7171e0088ad71307d47bd27d32e548e2aab0f1d540R650) 

And the `value` could be filtered for advanced cases, it'd need to return the `int`, always. We could also probably use the same `value` as a schedule for this cron.

With the above example:
1. With default `rocket_preload_cache_pending_jobs_cron_rows_count` (45) value, we'd catch URLs that are in `in-progress` state for 3 minutes. Cron schedule would run each 3 minutes too.
2. With `rocket_preload_cache_pending_jobs_cron_rows_count` set to 450, we'd catch URLs that are in `in-progress` state for 30 minutes. Cron schedule would run each 30 minutes too.

Not sure what would be the best value, 15 sounds reasonable, but we might think about the different one. What do you think @CrochetFeve0251 @wp-media/php ?

CrochetFeve0251:
> I'm thinking, maybe we could make the [value](https://github.com/wp-media/wp-rocket/pull/5628/files#diff-30a46cf1cdbb73aeb3f9cc7171e0088ad71307d47bd27d32e548e2aab0f1d540R650) of `2 minutes` auto-adjustable and also introduce the filter there. Let's say the value could be:
> 
> `rocket_preload_cache_pending_jobs_cron_rows_count` / `15` = [value](https://github.com/wp-media/wp-rocket/pull/5628/files#diff-30a46cf1cdbb73aeb3f9cc7171e0088ad71307d47bd27d32e548e2aab0f1d540R650)
> 
> And the `value` could be filtered for advanced cases, it'd need to return the `int`, always. We could also probably use the same `value` as a schedule for this cron.
> 
> With the above example:
> 
>     1. With default `rocket_preload_cache_pending_jobs_cron_rows_count` (45) value, we'd catch URLs that are in `in-progress` state for 3 minutes. Cron schedule would run each 3 minutes too.
> 
>     2. With `rocket_preload_cache_pending_jobs_cron_rows_count` set to 450, we'd catch URLs that are in `in-progress` state for 30 minutes. Cron schedule would run each 30 minutes too.
> 
> 
> Not sure what would be the best value, 15 sounds reasonable, but we might think about the different one. What do you think @CrochetFeve0251 @wp-media/php ?

The filter and calculating the delay is a good idea.
However there is no CRON for this task as we run it with the preload.

piotrbak:
@CrochetFeve0251 Is it ready for QA/CR?

remyperona (approved):

Mai-Saad:
For CR and further test:
The last change [e3523dd](https://github.com/wp-media/wp-rocket/pull/5628/commits/e3523ddf08e3d26da1f682fb36470e80e0f82d13) was mainly to fix the following problem/scenario:
Brief description=> changing status is not changing modified nor last accessed then when we process URL, mostly the modified/last accessed > 3min (default time to switch in progress to failed if not succeeded) => This will cause many URLs to be switched to failed although we didn't send the request for the URL then will be completed
Steps:
1- we did clear and preload so the time of each entry now not changed => mostly > 3mins
2- We change the status to inprogress => time still not changed => why not change last modified in this case?
3- After 3min we do our check and send all inprogress things to failed even if no request is sent
4- After a while request is sent and the status change to completed => time change here
Screenshot from 2023-01-10 06-27-27
Screenshot from 2023-01-10 06-26-06
Screenshot from 2023-01-10 06-17-42
Screenshot from 2023-01-10 06-18-35

**Notes:** 
- The change above fixed this problem but currently we change the last accessed when changing the status from pending to inprogress or from inprogress to failed/completed
- As per team discussion => modified shall change whenever we change item(s) in the row while last accessed whenever we receive a response for URL (page got cached) => @engahmeds3ed can you please confirm?

inc/Engine/Preload/Controller/PreloadUrl.php
```diff
@@ -150,7 +150,15 @@ public function preload_url( string $url ) {
 
 			usleep( $delay_between );
 		}
+		$rows = $this->query->get_rows_by_url( $url );
 
+		if ( ! $rows || count( $rows ) === 0 ) {
```
engahmeds3ed: No Need to check the count here because we return false from `get_rows_by_url` method if there is no available rows.

inc/Engine/Preload/Controller/PreloadUrl.php
```diff
@@ -150,7 +150,15 @@ public function preload_url( string $url ) {
...
+			return;
+		}
+
+		$row = array_pop( $rows );
```
engahmeds3ed: Are we using `array_pop` to get only mobile row or what?
from what I know, `get_rows_by_url` returns two rows, one for desktop version and the other for mobile version, right?
So Which one do we need to update last accessed?

inc/Engine/Preload/Controller/PreloadUrl.php
```diff
@@ -150,7 +150,15 @@ public function preload_url( string $url ) {
...
+
+		$this->query->update_last_access( $row->id );
```
engahmeds3ed: Why not letting the cache creation process handle this last accessed column?
did u do that Because of hosts with their own cache or what?

inc/Engine/Preload/Controller/PreloadUrl.php
```diff
@@ -181,9 +189,13 @@ protected function get_mobile_user_agent_prefix() {
 	 * @return void
 	 */
 	public function process_pending_jobs() {
+		$stuck_rows = $this->query->get_outdated_in_progress_jobs();
+		foreach ( $stuck_rows as $row ) {
+			$this->query->make_status_failed( $row->id );
+		}
```
engahmeds3ed: I believe having a dedicated method hooked to CRON hook `rocket_preload_process_pending` would be better to reflect the proper method name like what we do now here:
https://github.com/wp-media/wp-rocket/blob/b1857bdb42d48b7c1aed2aaeec413bfdc3e5a3b0/inc/Engine/Preload/Cron/Subscriber.php#L56-L57

what do u think?

inc/Engine/Preload/Database/Queries/Cache.php
```diff
@@ -371,7 +371,8 @@ public function make_status_inprogress( int $id ) {
 		return $this->update_item(
 			$id,
 			[
-				'status' => 'in-progress',
+				'status'        => 'in-progress',
+				'last_accessed' => current_time( 'mysql', true ),
```
engahmeds3ed: I believe `last_accessed` should be set when we try to access this page but now it's set every time we change the status and this isn't reflecting its name so why do we need to set the last_accessed here?

inc/Engine/Preload/Database/Queries/Cache.php
```diff
@@ -585,4 +606,75 @@ protected function is_rejected( string $url ): bool {
...
+		$delay = (int) apply_filters(
+			'rocket_preload_outdated',
+			/**
+			 * Set the max number of rows in batches.
+			 *
+			 * @param int $count number of rows in batches.
+			 * @return int
+			 */
+			(int) ( apply_filters( 'rocket_preload_cache_pending_jobs_cron_rows_count', 45 ) / 15 )
+		);
```
engahmeds3ed: this is the query class so I don't like having WP hooks here, try sending this value as a method argument.
If you moved this code here:
https://github.com/wp-media/wp-rocket/blob/f20d0c189ca6ce6c099ae5bbde0ade2b9a420a7f/inc/Engine/Preload/Controller/PreloadUrl.php#L191

You will take benefit of having the filter `rocket_preload_cache_pending_jobs_cron_rows_count` not used twice:
https://github.com/wp-media/wp-rocket/blob/f20d0c189ca6ce6c099ae5bbde0ade2b9a420a7f/inc/Engine/Preload/Controller/PreloadUrl.php#L196

inc/Engine/Preload/Database/Queries/Cache.php
```diff
@@ -585,4 +606,75 @@ protected function is_rejected( string $url ): bool {
...
+
+		return $this->query(
+			[
+				'status'     => 'in-progress',
+				'is_locked'  => false,
+				'date_query' => [
+					[
+						'column' => 'last_accessed',
```
engahmeds3ed: why not dealing here with the modified column?

engahmeds3ed (changes requested):

inc/Engine/Preload/Controller/PreloadUrl.php
CrochetFeve0251: The idea there was to run it before processing pending jobs to make sure it sure we won't make `in-progress` stucked longer than expected.

engahmeds3ed: Yes, this can be achieved by hooks priorities as well.
Thanks Cyrille.

inc/admin/upgrader.php
```diff
@@ -275,6 +275,10 @@ function rocket_new_upgrade( $wp_rocket_version, $actual_version ) {
 		rocket_generate_advanced_cache_file();
 	}
 
+	if ( version_compare( $actual_version, '3.12.5', '<' ) ) {
+		wp_clear_scheduled_hook( 'rocket_preload_revert_old_in_progress_rows' );
+	}
+
```
engahmeds3ed: Why not moving this code to the subscriber with the hook `wp_rocket_upgrade` ?
Also before clearing this schedule, we need to check if it's scheduled or not.

inc/Engine/Preload/Database/Queries/Cache.php
```diff
@@ -585,4 +606,59 @@ protected function is_rejected( string $url ): bool {
...
+	 *
+	 * @param int $id id from the task.
+	 * @return bool
+	 */
+	public function make_status_failed( int $id ) {
+		return $this->update_item(
+			$id,
+			[
+				'status'   => 'failed',
+				'modified' => current_time( 'mysql', true ),
```
engahmeds3ed, line 621: if we are using berlinDB update_item method, no need to set the modified column, berlinDB will do that for us.

engahmeds3ed (changes requested):
Looks amazing, just small enhancements.

mostafa-hisham (approved):

Mai-Saad:
This cron is not deleted after deleting WPR `rocket_preload_revert_old_failed_rows` and modified only changes when the status is completed ( the same as last accessed)

inc/Engine/Preload/Database/Queries/Cache.php
CrochetFeve0251: @engahmeds3ed look like this is not working so I reverted back changes

remyperona (approved):

Mai-Saad:
Working fine now

Mai-Saad (approved):

PR #5632, opened 2022-12-16
Closes #3477 - Subsites on multisite loading incorrect config file

### LINKED ISSUE #3477: Multisite - Cache Query String(s) & other features not working on subsites

jeawhanlee:
## Description

This will fix the issue of subsites loading incorrect config file(the main site config file) on multisite.

Fixes #3477 

## Type of change

## How Has This Been Tested?

Manual Tests

# Checklist:

### DIFF
--- a/inc/classes/Buffer/class-config.php
+++ b/inc/classes/Buffer/class-config.php
@@ -206,7 +206,7 @@ class Config {
 		if ( ! empty( $path ) ) {
 			$config_file_paths = [];
 
-			// Loop through paths and store valid config paths matching the url current path.
+			// Loop through paths and store valid config file paths matching the url current path in an array.
 			foreach ( $path as $p ) {
 				static $dir;
 
--- a/inc/classes/Buffer/class-config.php
+++ b/inc/classes/Buffer/class-config.php
@@ -196,34 +196,48 @@ class Config {
 		$path = str_replace( '\\', '/', strtok( $this->get_server_input( 'REQUEST_URI', '' ), '?' ) );
 		$path = preg_replace( '|(?<=.)/+|', '/', $path );
 		$path = explode( '%2F', preg_replace( '/^(?:%2F)*(.*?)(?:%2F)*$/', '$1', rawurlencode( $path ) ) );
+		// Remove empty array values.
+		$path = array_filter( $path );
 
-		$config_file_paths = [];
+		/**
+		 * If path is not empty.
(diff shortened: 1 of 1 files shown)


inc/classes/Buffer/class-config.php
```diff
@@ -197,34 +197,33 @@ public function get_config_file_path() {
...
-				return self::memoize(
+				$config_file_paths[] = self::$config_dir_path . $host . '.' . $dir . $p . '.php';
+			}
+
+			$dir .= $p . '.';
+		}
+
+		$config_file_paths = array_reverse( $config_file_paths );
+
+		foreach ( $config_file_paths as $config_path ) {
```
remyperona: If I understand correctly, here we are reversing the array order and then using the first item of the array for the config path, is that right?

jeawhanlee: That's right. We can just return the first item alone right?

remyperona: In that case, you can use `array_shift()` to extract the first item of the array directly instead of starting a loop

remyperona (approved):

Mai-Saad:
Working as expected
[testrail-report-434.pdf](https://github.com/wp-media/wp-rocket/files/10259923/testrail-report-434.pdf)

PR #5656, opened 2022-12-30
Added a way to change lazyload regex and logic to change it for Avada

### LINKED ISSUE #5649: Lazyload compatibility with the new Avada background markup
- [x] Made sure you’re on the latest version `3.12.3.3`

**Describe the bug**
Avada changed their markup for backgrounds inside an element to look like: 
```
<div class="fusion-equal-height-columns" style="--awb-background-image:url("path/to/image.jpg");--awb-background-size:cover;" ></div>
```
Instead of: 
```
<div class="fusion-equal-height-columns" style="background-image:url("path/to/image.jpg");--awb-background-size:cover;" ></div>
```

When lazyload is enabled, what happens is that our regex will catch `background-image:url("path/to/image.jpg");` and remove it, leaving the `--awb-` to be glued to whatever comes after it. 

In the above example, the end markup added to the page when lazyload is triggered will look like: 
```
<div class="fusion-equal-height-columns" style="--awb---awb-background-size:cover;background-image:url("path/to/image.jpg");" ></div>
```
Which will alter how the `background-size` property operates. 

Changing the following regex:

```
background-image\s*:\s*(?<attr>\s*url\s*\((?<url>[^)]+)\))\s*;?
```

found here: https://github.com/wp-media/wp-rocket/blob/58784bd05e3919595ed453870f69aeb3f580e8b2/inc/Dependencies/RocketLazyload/Image.php#L70 to: 

```
(--awb-)?background-image\s*:\s*(?<attr>\s*url\s*\((?<url>[^)]+)\))\s*;?
```

should fix the issue.

The new regex will look for `--awb-` as an optional string. 

We can use the same logic to accommodate themes/plugins prefixing the CSS background property. 

**To Reproduce**
1. Use the latest Avada
2. Add a background to an element
3. Enabled lazyload

**Expected behavior**
Lazyload on inline background shouldn't affect how the element is styled.

**Screenshots**
N/A

**Additional context**
Ticket - https://secure.helpscout.net/conversation/2106293204/390879/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description

Fix an issue where new properties are present for Avada theme that includes the bg image.

Fixes #5649

## Type of change

No.

## How Has This Been Tested?

- [ ] Automated tests

# Checklist:


inc/Dependencies/RocketLazyload/Image.php
```diff
@@ -67,7 +67,13 @@ public function lazyloadBackgroundImages( $html, $buffer ) {
 				continue;
 			}
 
-			if ( ! preg_match( '#background-image\s*:\s*(?<attr>\s*url\s*\((?<url>[^)]+)\))\s*;?#is', $element['styles'], $url ) ) {
+			$regex = apply_filters( 'rocket_lazyload_bg_images_regex', 'background-image\s*:\s*(?<attr>\s*url\s*\((?<url>[^)]+)\))\s*;?' );
+
+			if(! is_string( $regex ) ) {
```
jeawhanlee: I think to be more precise, we should check if regex has a `background-image` pattern in it.

jeawhanlee:
```suggestion
			if ( false !== strpos( $regex, 'background-image' ) ) {
```
What do you think?

remyperona: Also add doc block for the new filter

CrochetFeve0251: @jeawhanlee I am wondering if checking for `background-image` is a good thing as the user can create a regex like that `background-imag(e)?`and that is still valid whereas it is not passing the check.

Maybe instead we could check if the regex is valid with the following code? 
```php
@preg_match($string, '') !== FALSE;
```

jeawhanlee: I see what you mean, I just think it would be good if we return on time if user pattern is not `background` related maybe we can do this with a preg_match.

piotrbak:
@CrochetFeve0251 Is it ready for the next review?

remyperona (approved):

vmanthos:
The PR is working for both the new and old `background-image` markup used by Avada.

  During exploratory testing, I found that background images with a filename like `(plaska)+ziemia_xD flat earth (lol).PNG` are not working. The regex picks up only this `(plaska` which results in an error `404`.

This is happening with `trunk` as well. 

@piotrbak Will we address this, and will it be in this PR?

vmanthos:
@piotrbak Pinging about the comment:
https://github.com/wp-media/wp-rocket/pull/5656#issuecomment-1398113306

If we aren't going to address this the PR is `approved`. :+1:

piotrbak:
@vmanthos We'll move forward there without additional fix. It's there for all images, not only for Avada, right?

vmanthos:
> It's there for all images, not only for Avada, right?

Exactly. I'll create a different GitHub issue about this.

vmanthos (approved):

PR #5664, opened 2023-01-05
Closes #5654 Update WebP detection for browsers missing Accept headers data

### LINKED ISSUE #5654: Improve webp detection on apple devices

**Describe the bug**
Currently, WebP support detection is not working well for Safari browsers. Even for MacOS/Safari versions that do support WebP (Big Sur and later), the WebP cache is not being served.

The issue is due to Safari no longer sending the needed headers so WP Rocket can serve the WebP cache properly.

Related to this issue as well: https://github.com/wp-media/wp-rocket/issues/5524

**To Reproduce**
View any page in MacOS Big Sur or later where WebP images are supported with the WebP add-on enabled. Pages that have WebP versions of images available will still not have WebP cache served.

The user did some extensive testing on their end, and here were their specific results (seems on iOS it's not only happening for Safari, but also for other browsers like FF and Chrome since they are required to use the Safari rendering engine)

Windows, Chrome: html with webp links is provided (correct)
Windows, Edge: html with webp links is provided (correct)
Ubuntu, Firefox: html with webp links is provided (correct)
MacbookPro, Chrome: html with webp links is provided (correct)
MacbookPro, Firefox: html with webp links is provided (correct)
MacbookPro, Safari: html with JPG links is provided (WRONG)
iPhone 13 mini, Chrome: html with JPG links is provided (WRONG)
iPhone 13 mini, Firefox: html with JPG links is provided (WRONG)
iPhone 13 mini, Safari: html with JPG links is provided (WRONG) - same for iPhone 14 iOS 16.2, iPhone 12 iOS 14.2, etc.
iPad Mini 5, Chrome: html with JPG links is provided (WRONG)
iPad Mini 5, Firefox: html with JPG links is provided (WRONG)
iPad Mini 5, Safari: html with JPG links is provided (WRONG)

User agents for FF/Chrome/Safari/etc on iOS will be different, so these should be taken into account for the solution as well.

**Expected behavior**
According to Piotr in the Slack discussion, WP Rocket will need to regex user agents to detect specific apple browsers and serve our WebP cache based on that.

**Screenshots**
Safari 16 on Ventura
image

Chrome on Windows
image

**Additional context**
Slack discussion: https://wp-media.slack.com/archives/C08N8J6VC/p1672325582266889
Ticket: https://secure.helpscout.net/conversation/2108559032/391269/
Test page where issue occurs: https://test.prodvomschaf.de/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

remyperona:
## Description

This PR refactors the WebP class to split it in multiple classes to reduce complexity and bring it in the new architecture.

It also add new detection for browsers missing Accept headers data, most notably Apple devices on newer versions of Safari.

Fixes #5654

## Type of change

No

# Checklist:


inc/Addon/WebP/AbstractWebp.php
```diff
@@ -0,0 +1,103 @@
...
+		 */
+		$webp_plugins = (array) apply_filters( 'rocket_webp_plugins', [] );
+
+		if ( ! $webp_plugins ) {
+			// Somebody probably messed up.
+			return [];
+		}
+
+		foreach ( $webp_plugins as $i => $plugin ) {
+			if ( ! is_a( $plugin, '\WP_Rocket\Subscriber\Third_Party\Plugins\Images\Webp\Webp_Interface' ) ) {
```
CrochetFeve0251: @Tabrisrp maybe we can use the ::class attribute here?

CrochetFeve0251 (approved):

vmanthos:
@Tabrisrp I merged `develop` to this PR locally and then I tested it.

We are not rewriting images to their WebP counterparts. This is because the path here is wrong:
https://github.com/wp-media/wp-rocket/blob/3c2bd70b49c1331b1ca1a535064ac937f321ea53/inc/Addon/WebP/Subscriber.php#L426

Instead of being `/var/www/example.com/wp-content/uploads/2022/12/01/image.jpg` that's `/var/www/example.com/wp-contentuploads/2022/12/01/image.jpg`, i.e. there is a missing `/` after `wp-content`.

This is coming from here:
https://github.com/wp-media/wp-rocket/blob/3c2bd70b49c1331b1ca1a535064ac937f321ea53/inc/Addon/WebP/Subscriber.php#L408

Can you please take a look into this?

You can see this happening here:
https://mega.wp-rocket.me/divi/webp-testing/

remyperona:
@vmanthos Thanks, it's fixed with the latest commit

vmanthos:
@Tabrisrp Thank you for the fix. WebP-containing cached pages are correctly generated now.
However, on Safari `jpg` images are still served. 

I saw that you created methods to detect if `is_browser_webp_compatible()` in `WP_Rocket\Addon\WebP`, but I didn't see any change related to how we are serving the `webp` file. Shouldn't that method be used here?
https://github.com/wp-media/wp-rocket/blob/eb7de8265f190f1efd4d14865b0da15f68c5303d/inc/classes/Buffer/class-cache.php#L589-L625

I'm surely missing something here. Can you please look into this?

piotrbak:
@Tabrisrp Is it ready for the next CR or should it go to QA directly?

inc/Addon/WebP/AbstractWebp.php
```diff
@@ -0,0 +1,104 @@
...
+	 * @param CDNSubscriber $cdn_subscriber CDNSubscriber instance.
+	 */
+	public function __construct( CDNSubscriber $cdn_subscriber ) {
+		$this->cdn_subscriber = $cdn_subscriber;
+	}
+
+	/**
+	 * Get a list of active plugins that convert and/or serve webp images.
+	 *
+	 * @since 3.4
```
engahmeds3ed:
```suggestion
	 * @since 3.12.6
```

inc/Addon/WebP/AbstractWebp.php
```diff
@@ -0,0 +1,104 @@
...
+		 * @param array $webp_plugins An array of Webp_Interface objects.
+		 */
+		$webp_plugins = (array) apply_filters( 'rocket_webp_plugins', [] );
+
+		if ( ! $webp_plugins ) {
+			// Somebody probably messed up.
+			return [];
+		}
+
+		foreach ( $webp_plugins as $i => $plugin ) {
```
engahmeds3ed:
```suggestion
		foreach ( $webp_plugins as $plugin_key => $plugin ) {
```

inc/Addon/WebP/AbstractWebp.php
```diff
@@ -0,0 +1,104 @@
...
+			if ( ! $plugin instanceof Webp_Interface ) {
+				unset( $webp_plugins[ $i ] );
+				continue;
+			}
+
+			if ( ! $this->is_plugin_active( $plugin->get_basename() ) ) {
+				unset( $webp_plugins[ $i ] );
+				continue;
```
engahmeds3ed: We don't need to use continue here as it's already the end of the loop iteration, what do u think?

inc/Addon/WebP/AbstractWebp.php
```diff
@@ -0,0 +1,104 @@
...
+			}
+		}
+
+		return $webp_plugins;
+	}
+
+	/**
+	 * Tell if a plugin is active.
+	 *
+	 * @since 3.4
```
engahmeds3ed:
```suggestion
	 * @since 3.12.6
```

inc/Addon/WebP/AbstractWebp.php
```diff
@@ -0,0 +1,104 @@
...
+	protected function is_plugin_active( string $plugin_basename ): bool {
+		if ( \doing_action( 'deactivate_' . $plugin_basename ) ) {
+			return false;
+		}
+
+		if ( \doing_action( 'activate_' . $plugin_basename ) ) {
+			return true;
+		}
+
+		return \rocket_is_plugin_active( $plugin_basename );
```
engahmeds3ed:
```suggestion
		if ( doing_action( 'deactivate_' . $plugin_basename ) ) {
			return false;
		}

		if ( doing_action( 'activate_' . $plugin_basename ) ) {
			return true;
		}

		return rocket_is_plugin_active( $plugin_basename );
```

inc/Addon/WebP/AbstractWebp.php
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	/**
+	 * Tell if WP Rocket uses a CDN for images.
+	 *
+	 * @since 3.4
```
engahmeds3ed:
```suggestion
	 * @since 3.12.6
```

inc/Addon/WebP/AdminSubscriber.php
```diff
@@ -0,0 +1,214 @@
...
+
+	/**
+	 * Modifies the WebP section description of WP Rocket settings.
+	 *
+	 * @since 3.4
+	 *
+	 * @param  array $cache_webp_field Section description.
+	 * @return array
+	 */
+	public function webp_section_description( $cache_webp_field ) {
```
engahmeds3ed, line 62: very complex method, I don't know if moving each block to its own private method is in the scope of this PR or not but I think this will help us understanding this part more and change it when needed, what do u think?

inc/Addon/WebP/AdminSubscriber.php
```diff
@@ -0,0 +1,214 @@
...
+		// 2.
+		$cache_webp_field['description'] = esc_html__( 'WP Rocket will create separate cache files to serve your WebP images.', 'rocket' );
+
+		return $cache_webp_field;
+	}
+
+	/**
+	 * Disable 'cache_webp' setting field if another plugin serves WebP.
+	 *
+	 * @since 3.4
```
engahmeds3ed, line 193:
```suggestion
	 * @since 3.12.6
```

inc/Addon/WebP/Subscriber.php
```diff
@@ -0,0 +1,570 @@
...
+
+		/**
+		 * Tell if the page contains webp files.
+		 *
+		 * @since 3.4
+		 *
+		 * @param bool   $has_webp True if the page contains webp files. False otherwise.
+		 * @param string $html     The page’s html contents.
+		 */
+		$has_webp = apply_filters( 'rocket_page_has_hebp_files', $has_webp, $html );
```
engahmeds3ed: I know that this filter was there before this refactoring but I believe there is a typo on the filter name and u solved it for variable name, I'm not sure we can change that and deprecate the old name, what do u think?
```suggestion
		$has_webp = apply_filters( 'rocket_page_has_webp_files', $has_webp, $html );
```

engahmeds3ed (changes requested):
Minor discussion points

inc/Addon/WebP/AdminSubscriber.php
remyperona: I don't like it but it's also out of scope

engahmeds3ed (approved):

vmanthos:
@Tabrisrp I tested the PR using browserstack.com and found that WebP images are served to the latest Safari version but they are not served to:
1. Previous Safari versions that support WebP images, i.e.  Safari 14-15.6 on MacOS 11 Big Sur or later.
2. Mobile Safari versions that support WebP Images, i.e. versions 14 onward.

For the 1, we don't have RegEx's to find their user agents. Here are some of them:
**Big Sur/Safari 14.1**
`Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1.2 Safari/605.1.15`

**Monterey/Safari 15.6**
`Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1.2 Safari/605.1.15`

For number 2, we do detect the user agent here:
https://github.com/wp-media/wp-rocket/blob/1d6e3dc00e6b91b0be65cc03da603bb5a8bfdd3c/inc/classes/Buffer/class-cache.php#L665-L669

but when we check this:
https://github.com/wp-media/wp-rocket/blob/1d6e3dc00e6b91b0be65cc03da603bb5a8bfdd3c/inc/classes/Buffer/class-cache.php#L671-L674

that matches the mobile user agent as well and bails out:
https://regex101.com/r/Dv7yqC/1

remyperona:
Don't those versions have the WebP hint in their Accept header?

vmanthos:
It seems they don't, at least not on browser stack.com where I checked. By the way, this wasn't working with trunk either.

remyperona:
I fixed case 2.

For case 1 we don't have a way to detect the OS, the user agent doesn't give the correct information (it's outdated).

piotrbak:
@vmanthos Sounds like we don't have a way there. We should be fine with it

vmanthos:
@Tabrisrp I believe we need to change the `Cache::is_user_agent_compatible()` method as well. 

Currently, if a Safari that supports WebP images accesses a non-cached URL, we are creating a cached file that includes `webp` images but doesn't have `-webp` in its filename.

remyperona:
Thank you, updated that too

vmanthos:
@Tabrisrp Thank you for the update.

With the latest changes the only thing preventing serving WebP images to `13 < Safari < 16` is this conditional:
https://github.com/wp-media/wp-rocket/blob/3c6a68b8e747021c1c1a74ea409db7e824fa46a9/inc/classes/Buffer/class-cache.php#L674

Setting that to `14` allowed WP Rocket to serve the `-webp` cache file to Safari both on MacOS and iPads which share the same user agents:
```
iPad 8th / Safari 14.0
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.0 Safari/605.1.15

iPad / Safari 15.3
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.3 Safari/605.1.15

Monterey / Safari 15.6
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.3 Safari/605.1.15

Bigsur / Safari 14.1.2
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1.2 Safari/605.1.15

Catalina / Safari 13.1.2 (not supporting WebP)
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_6) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.1.2 Safari/605.1.15
```
By the way, Safari 15.6.1 is available for [Catalina](https://www.macrumors.com/2022/08/18/safari-security-update-macos-big-sur-catalina/). This isn't available on browserstack.com to test, but that should work fine with the WebP cache.

Of course, this should be changed as well:
https://github.com/wp-media/wp-rocket/blob/562dd77c3b0af0d17ede87308213d91f586e751c/inc/Addon/WebP/Subscriber.php#L278

remyperona:
Isn't Safari compatible with WebP only from Safari 16 on macOS?

vmanthos:
> Isn't Safari compatible with WebP only from Safari 16 on macOS?

No. Safari started supporting WebP images in v14:
https://caniuse.com/webp

remyperona:
I'm failing to see what pattern we can use to detect Safari 14 or higher, only on Big Sur or higher

vmanthos:
@Tabrisrp I tested Safari 15.6.1 on Catalina which preceded Big Sur and WebP images are loaded there:
image

~From this, it's safe to assume that regardless of the MacOS version if Safari >= 14 is used WebP images are supported.~

**UPDATE:** Check comment:
https://github.com/wp-media/wp-rocket/pull/5664#issuecomment-1429977718

@piotrbak Should we apply the changes I mentioned here?
https://github.com/wp-media/wp-rocket/pull/5664#issuecomment-1429290574

piotrbak:
@vmanthos Was this image loaded and displayed correctly? If so, we can proceed.

vmanthos:
> Was this image loaded and displayed correctly?

It was the first time I checked. I double-checked to make sure I didn't miss anything and it's not displayed. :disappointed: 
I'm not sure what happened the first time around. ¯\_(ツ)_/¯

So, unfortunately, this is a no-go as it seems that Big Sur is indeed a requirement.

vmanthos (approved):

**NOTES**

The `-webp` cached files will _NOT_ be served to:
* Desktop Safari versions `< 16`
* iPad Safari versions `< 16`
* iPhone Safari versions `< 14`

PR #5666, opened 2023-01-09
Added $ to `cache_reject_uri` to prevent partial urls to be used.

### LINKED ISSUE #5588: Partial URL added to Never Cache URLs is excluded from preloading
- Made sure you’re on the latest version ✅ `3.12.3.1`

**Describe the bug**

When adding URLs to the Never Cache URLs text area, those are removed from the `wpr_rocket_cache` table in order to avoid preloading them.

When the partial path of a URL is added, WP Rocket will not exclude it from caching, i.e. caching/optimizations are applied. However, the URL is removed from `wpr_rocket_cache` table.

**To Reproduce**
1. Change the permalinks to not include a trailing `/` (necessary after [this fix](https://github.com/wp-media/wp-rocket/pull/5565)).
2. For a URL `/example-url-to-exclude` add `/example-url-to` to the "Never Cache URLs" text area.
3. After preload finishes check the `wpr_rocket_cache` table for the `/example-url-to-exclude` URL. It will not be there.

**Expected behavior**

We should not exclude the URL from preload. Ideally, although this may not be possible, we should detect such URLs and display a notice.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

CrochetFeve0251:
## Description
Added $ to all `cache_reject_uri` to prevent partial urls to be excluded from preload.

Fixes #5588

## Type of change

Yes after testing the second grooming doesn't work so I went back to the first one.

## How Has This Been Tested?
- [ ] Test on local env

# Checklist:


mostafa-hisham (approved):

inc/Engine/Preload/Controller/CheckExcludedTrait.php
```diff
@@ -11,7 +11,13 @@ trait CheckExcludedTrait {
 	 * @return array
 	 */
 	public function add_cache_reject_uri_to_excluded( array $regexes ): array {
-		$regexes[] = get_rocket_cache_reject_uri();
+		$user_added_cache_reject_uri         = (array) get_rocket_option( 'cache_reject_uri', [] );
```
jeawhanlee: We need to check if `$user_added_cache_reject_uri` is empty and return early else continue process

jeawhanlee (commented):
Need to update tests too

jeawhanlee (approved):

vmanthos (approved):

Issue #5647, opened 2022-12-27
Enhance the Handling of WWW/non-WWW Updates


camilamadronero-zz:
**Is your feature request related to a problem? Please describe.**
Let's say you change your site URL from non-WWW to WWW, and visit the non-WWW URL:

- **WordPress without WP Rocket:**  You'll be automatically redirected to the WWW URL. 
- **WordPress with WP Rocket:** WP Rocket will serve the non-WWW cache, **if the old WP Rocket configuration file exists**.
The non-WWW cache will be outdated, with broken links, and it's not purgeable.

While we have a [guide for domain changes](https://docs.wp-rocket.me/article/705-changing-domains-migrating-sites-with-wp-rocket), and having two configuration files is not always happening, making an automatic check+update will be useful.

**Describe the solution you'd like**
WP Rocket should detect the URL update or [General options](/wp-admin/options-general.php) update, and purge its paths and configurations accordingly.

**Describe alternatives you've considered**
- Unless the [WP Rocket | Multiple Config Files](https://github.com/wp-media/wp-rocket-helpers/tree/master/various/wp-rocket-multiple-config-files) helper is detected, WP Rocket should allow only one config file to exist.
- Detect the URL change and display an admin notice to purge config files.

**Additional context**
Related ticket: https://secure.helpscout.net/conversation/2097115953/388901?folderId=377611.
Slack thread: https://wp-media.slack.com/archives/G8L4EE8N9/p1671136542516399.

piotrbak:
This is not something that can be implemented in the nearest future.

PR #5688, opened 2023-01-17
Closes 5667 detect domain change

### LINKED ISSUE #5667: Display a notice after a site was cloned to disable/re-enable WP Rocket
Not all users are aware they have to deactivate WP Rocket before a migration/clone and if a user migrate/clone their site with WP Rocket active, it can leads to issues like a non cached/optimized site due to wrong config files. 

Deactivate/re-activate WP Rocket allows to recreate the right config files and fixes the issues.

Display a notice after a site was cloned to disable/re-enable WP Rocket. 

Auto-detect the new domain/site URL and deactivate/re-activate WP Rocket in background.
Woocommerce detects when a new domain/site URL is found and automatically disable functionality like Subscriptions.

**Additional context**
Ticket: https://secure.helpscout.net/conversation/2114325821/393025/
HS Doc: https://docs.wp-rocket.me/article/705-changing-domains-migrating-sites-with-wp-rocket

CrochetFeve0251:
## Description

This issue is related to the WP Rocket caching plugin for WordPress, where the user is requesting a feature that would display a notice after a site has been cloned to remind the user to deactivate and re-activate WP Rocket. The problem is that some users may not be aware that they need to deactivate WP Rocket before a migration or clone, and if they do not, it can lead to issues such as a non-cached or non-optimized site due to incorrect config files.
The solution proposed during grooming is to add a new subscriber class that detects a change of domain and launches the action `rocket_domain_changed` which will clear the cache, reset the preload cache table, and clear the config files.

Fixes  #5667

## Type of change

No.

## How Has This Been Tested?

- [ ] Tested locally.
- [ ] Test with automated tests.

# Checklist:


inc/Engine/Admin/DomainChange/Subscriber.php
```diff
@@ -0,0 +1,51 @@
...
+
+		$last_base_url = get_option( self::LAST_BASE_URL_OPTION );
+
+		if ( $base_url === $last_base_url ) {
+			return;
+		}
+
+		update_option( self::LAST_BASE_URL_OPTION, $base_url );
+
+		do_action( 'rocket_domain_changed' );
```
remyperona: add docblock for the action

remyperona (changes requested):

remyperona (approved):

vmanthos:
@CrochetFeve0251 I checked the PR on single (as in not multisite) sites. Both when changing www to non-www and vice versa, and when completely changing the domain name.

* The old configuration files are deleted and new ones are correctly created.
* The `wpr_rocket_cache` table is cleared and the new sitemap is loaded and preloaded.

What's not happening, and I believe it should - it will also resolve the issue mentioned by @camilamadronero-zz [here](https://github.com/wp-media/wp-rocket/issues/5647)), is deleting the old cache folders. 

Can you please look into this? Also, take into consideration multisites. :pray:

remyperona (approved):

vmanthos:
@CrochetFeve0251 After changing the domain name, e.g. adding or removing `www`,  when first logging into the admin area I'm getting the following fatal error:
```PHP
PHP Fatal error:  Uncaught TypeError: call_user_func_array(): Argument #1 ($callback) must be a valid callback, cannot access protected method WP_Rocket\Engine\Cache\AdminSubscriber::clear_cache() in /home/u923949912/domains/example.com/public_html/wp-includes/class-wp-hook.php:308
Stack trace:
#0 /home/u923949912/domains/example.com/public_html/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters()
#1 /home/u923949912/domains/example.com/public_html/wp-includes/plugin.php(517): WP_Hook->do_action()
#2 /home/u923949912/domains/example.com/public_html/wp-content/plugins/wp-rocket/inc/Engine/Admin/DomainChange/Subscriber.php(52): do_action()
#3 /home/u923949912/domains/example.com/public_html/wp-includes/class-wp-hook.php(308): WP_Rocket\Engine\Admin\DomainChange\Subscriber->maybe_launch_domain_changed()
#4 /home/u923949912/domains/example.com/public_html/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters()
#5 /home/u923949912/domains/example.com/public_html/wp-includes/plugin.php(517): WP_Hook->do_action()
#6 /home/u923949912/domains/example.com/public_html/wp-admin/admin.php(175): do_action()
#7 /home/u923949912/domains/example.com/public_html/wp-admin/plugins.php(10): require_once('/home/u92394991...')
#8 {main}
  thrown in /home/u923949912/domains/example.com/public_html/wp-includes/class-wp-hook.php on line 308
```

After this, WordPress and WP Rocket is loading.

Apart from the fatal error:
1. The cache isn't cleared.
2. The entries in the `wpr_rocket_cache` table aren't removed and are not preloaded. Note that this happened when `wp_rocket_last_base_url` was already in the database and `rocket_domain_changed` fired.
3. `wp_rocket_last_base_url` isn't removed when uninstalling WP Rocket.

CrochetFeve0251:
> @CrochetFeve0251 After changing the domain name, e.g. adding or removing `www`, when first logging into the admin area I'm getting the following fatal error:
> 
> ```
> PHP Fatal error:  Uncaught TypeError: call_user_func_array(): Argument #1 ($callback) must be a valid callback, cannot access protected method WP_Rocket\Engine\Cache\AdminSubscriber::clear_cache() in /home/u923949912/domains/example.com/public_html/wp-includes/class-wp-hook.php:308
> Stack trace:
> #0 /home/u923949912/domains/example.com/public_html/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters()
> #1 /home/u923949912/domains/example.com/public_html/wp-includes/plugin.php(517): WP_Hook->do_action()
> #2 /home/u923949912/domains/example.com/public_html/wp-content/plugins/wp-rocket/inc/Engine/Admin/DomainChange/Subscriber.php(52): do_action()
> #3 /home/u923949912/domains/example.com/public_html/wp-includes/class-wp-hook.php(308): WP_Rocket\Engine\Admin\DomainChange\Subscriber->maybe_launch_domain_changed()
> #4 /home/u923949912/domains/example.com/public_html/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters()
> #5 /home/u923949912/domains/example.com/public_html/wp-includes/plugin.php(517): WP_Hook->do_action()
> #6 /home/u923949912/domains/example.com/public_html/wp-admin/admin.php(175): do_action()
> #7 /home/u923949912/domains/example.com/public_html/wp-admin/plugins.php(10): require_once('/home/u92394991...')
> #8 {main}
>   thrown in /home/u923949912/domains/example.com/public_html/wp-includes/class-wp-hook.php on line 308
> ```
> 
> After this, WordPress and WP Rocket is loading.
> 
> Apart from the fatal error:
> 
>     1. The cache isn't cleared.
> 
>     2. The entries in the `wpr_rocket_cache` table aren't removed and are not preloaded. Note that this happened when `wp_rocket_last_base_url` was already in the database and `rocket_domain_changed` fired.
> 
>     3. `wp_rocket_last_base_url` isn't removed when uninstalling WP Rocket.
> 
> 
> Can you please look into these? pray

@vmanthos the fatal could explain  point 1. and 2.
I will fix it and look at 3.

vmanthos:
@CrochetFeve0251 Thank you for the fixes!

All issues have been fixed except for the cache clearing (1.). 
`rocket_clean_domain()` runs but the URL we pass to be cleared is the current one, i.e. the one after the change.

When you have time, please check this. In the meantime, I'll continue exploratory testing.

vmanthos:
@CrochetFeve0251 Additionally to my previous comment, the following part should be updated:
https://github.com/wp-media/wp-rocket/blob/60d5f5762639fae5fdf76f443ba41540aa13af5c/inc/Engine/Cache/AdminSubscriber.php#L212-L222

to match the logic here:
https://github.com/wp-media/wp-rocket/blob/0256f7421f3de35f1ace63c5526092fdfda56077/inc/functions/files.php#L250-L264

More info: [#5854](https://github.com/wp-media/wp-rocket/issues/5854)

vmanthos:
@CrochetFeve0251 I continued exploratory testing on my Hostinger test server. They have a staging site feature where both the URL and the files paths are different compared to the production site. 

I pushed changes from staging to production and I found that they are _not_ overwriting the value of `wp_rocket_last_base_url`. So in both environments, it remains the same after the push and this prevents `rocket_domain_changed` from firing. Consequently, anything that relies on this is failing.

I discovered that while they are overwriting the database, they won't do it for fields whose values contain the site's URL, e.g. `wpr_rocket_cache` rows. I'm guessing other hosts will do the same to prevent breaking the website, e.g. if replacing the `siteurl` or `home` options in the `wp_options` table.

To circumvent this we can save a hash of the URL of the site to `wp_rocket_last_base_url` instead of the actual URL.

What do you think?

vmanthos:
@CrochetFeve0251 On Hostinger, `.htaccess` isn't pushed from staging to production. This could be happening with other hosts as well and the reason is likely similar to the "why" they are not overwriting database entries that contain the URL of the site.

If there are changes to the staging site, e.g. a URLis add to the "Never Cache URLs", and our rewrite rules are in place, after the push those changes will not be reflected on the `.htaccess` file of the production site until the file is regenerated. Therefore the excluded URL, if previously cached, will still be served.  

We should `flush_rocket_htaccess()` when `rocket_domain_changed` fires. That can be added here:
https://github.com/wp-media/wp-rocket/blob/60d5f5762639fae5fdf76f443ba41540aa13af5c/inc/Engine/Cache/AdminSubscriber.php#L189-L191

---

To sum up my findings so far:
1. `rocket_clean_domain()` runs but the URL we pass to be cleared is the current one, i.e. the one after the change. ([comment](https://github.com/wp-media/wp-rocket/pull/5688#issuecomment-1490179299))
2. The logic about which `.php` files are cleared in the configuration folder must be changed. ([comment](https://github.com/wp-media/wp-rocket/pull/5688#issuecomment-1498965548))
3. We need to store a hash(?) of the site's URL in `wp_rocket_last_base_url`. ([comment](https://github.com/wp-media/wp-rocket/pull/5688#issuecomment-1500038879))
4. The `.htaccess` file should be regenerated when we detect a domain change.

remyperona (approved):

vmanthos:
@CrochetFeve0251 Thank you for the fixes. :pray: 

If possible, we should also handle the case where WP Rocket's settings were changed in one of the two environments before pushing, e.g. the staging to production, and clear the cache if necessary.

The scenario goes as follows:
1. On the staging site, the Minify CSS files feature is enabled.
2. The change is pushed to production.
3. The cache is cleared and preloaded if the preload feature is enabled.

As a reminder, we shouldn't clear the cache when some options are changed:
https://github.com/wp-media/wp-rocket/blob/0300a74643a97932214478df4a49fee39fc37b2f/inc/admin/options.php#L25-L54

inc/Engine/Admin/DomainChange/Subscriber.php
```diff
@@ -4,6 +4,7 @@
 
 use WP_Rocket\Event_Management\Subscriber_Interface;
 use WP_Filesystem_Direct;
+use function Patchwork\Config\get;
```
remyperona: This probably shouldn't be there

inc/Engine/Admin/DomainChange/Subscriber.php
```diff
@@ -21,7 +23,10 @@ class Subscriber implements Subscriber_Interface {
 	 */
 	public static function get_subscribed_events() {
 		return [
-			'admin_init' => 'maybe_launch_domain_changed',
+			'admin_init'                    => 'maybe_launch_domain_changed',
+			'rocket_configurations_changed' => 'configurations_changed',
+			'rocket_domain_changed'         => 'maybe_clean_cache_domain_change',
+			'update_option_' . rocket_get_constant( 'WP_ROCKET_SLUG' ) => ['save_hash_on_update_options', 10, 2],
```
remyperona: spacing issue inside array

inc/Engine/Admin/DomainChange/Subscriber.php
```diff
@@ -57,4 +62,61 @@ public function maybe_launch_domain_changed() {
...
+		update_option( self::LAST_OPTION_HASH, $hash );
+	}
+
+	/**
+	 * Maybe clean cache on domain change.
+	 *
+	 * @return void
+	 */
+	public function maybe_clean_cache_domain_change() {
+		if ( apply_filters( 'rocket_configurations_changed', false ) ) {
```
remyperona: missing docblock for filter

remyperona (changes requested):
There is issues in the tests too

inc/Engine/Admin/DomainChange/Subscriber.php
CrochetFeve0251: @Tabrisrp is there a way to add them to phpcs? I always forget them

remyperona (approved):

vmanthos:
@CrochetFeve0251 The cache isn't cleared on the production site when WP Rocket's settings have been changed on the staging site. So, they will not be applied there until the cache is cleared.

There is also the matter of adding new content, e.g. new posts, or making changes that affect the front-end, on staging. In this case, the cache of the production site will have to be cleared after the push. 

Unless there is a reliable way to detect this, maybe we should always clear the cache after the push. 

What do you think @piotrbak?

piotrbak:
@vmanthos Could you write the exact scenario here about the settings? I might be missing the point.

vmanthos:
It's the scenario I mentioned here:
https://github.com/wp-media/wp-rocket/pull/5688#issuecomment-1515894318
 where a setting that affects the front end is enabled or changed, e.g. new exclusions are added to fix an issue.

piotrbak:
@CrochetFeve0251 What do you think about this?

piotrbak:
@vmanthos @CrochetFeve0251 After reviewing the risks and what was discovered during the exploratory QA here, we had discussion related to the approach here.

We'll display a banner informing customer about the detected domain change and allow him to regenerate WP Rocket's files.

The AC for this are following (@vmanthos please validate):
- Non-dismissible warning notice will be displayed when the domain change is detected (wording at the bottom of the message)
- If customer decides to regenerate WP Rocket's files, we'll need to regenerate: configuration file, advanced-cache file, .htaccess. We'll also need to erase the cache table and remove Cache files, erase the Used CSS table and remove Used CSS files
- If preload was enabled, we should refetch the sitemaps.
- Multisites will be excluded from this process

**Wording:**
```
We detected that the website domain has changed. The configuration files must be regenerated for the page cache and all other optimizations to work as intended. Learn More (https://docs.wp-rocket.me/article/705-changing-domains-migrating-sites-with-wp-rocket)
```
**Button:** ` Regenerate WP Rocket configuration files now`

What if:
- A couple of configuration files are present in the configuration directory?

vmanthos:
@piotrbak The acceptance criteria look good. :+1:

vmanthos:
@CrochetFeve0251 The notice we display is dismissible:
image

According to the [acceptance criteria](https://github.com/wp-media/wp-rocket/pull/5688#issuecomment-1560915060) it should be non-dismissible.

vmanthos:
@CrochetFeve0251 The notice is still dismissible because of a typo here:
https://github.com/wp-media/wp-rocket/blob/9ae896f65e83e0017d07e715af81ed48d98dea11/inc/Engine/Admin/DomainChange/Subscriber.php#L170

It should be:
```PHP
'dismissible' => '',
```

Can you please fix that? :pray: 

**UPDATE**

There are some PHPCS errors as well:
https://github.com/wp-media/wp-rocket/actions/runs/5420026291/jobs/9853723073#step:5:41

vmanthos:
@CrochetFeve0251 The notice is non-dismissible now. :+1: 

I found the following issues:
1. The `wp_rocket_last_option_hash` option should be removed when uninstalling WP Rocket.
2. The same goes for the `rocket_domain_changed` transient which will remain if the "Regenerate WP Rocket configuration files now" isn't clicked before uninstalling the plugin.
3. On my local site, I am getting the following warnings:
```PHP
PHP Warning:  Cannot modify header information - headers already sent by (output started at /shared/httpd/wprocketest/htdocs/wp-content/plugins/wp-rocket/inc/admin/ui/notices.php:804) in /shared/httpd/wprocketest/htdocs/wp-admin/includes/misc.php on line 1431
PHP Warning:  Cannot modify header information - headers already sent by (output started at /shared/httpd/wprocketest/htdocs/wp-content/plugins/wp-rocket/inc/admin/ui/notices.php:804) in /shared/httpd/wprocketest/htdocs/wp-includes/functions.php on line 6896
PHP Warning:  Cannot modify header information - headers already sent by (output started at /shared/httpd/wprocketest/htdocs/wp-content/plugins/wp-rocket/inc/admin/ui/notices.php:804) in /shared/httpd/wprocketest/htdocs/wp-admin/admin-header.php on line 9
```
that cause various issues.

I wasn't able to reproduce this on the live website but I'm getting the notice loading before the page on both sites.
https://youtu.be/nheARMiB55I (watch this at 0.25x playback speed) 

Can you look into these, please? :pray:

inc/Engine/Admin/DomainChange/Subscriber.php
```diff
@@ -124,4 +145,100 @@ public function maybe_clean_cache_domain_change() {
...
+		$args = [
+			'status'         => 'warning',
+			'dismissible'    => '',
+			'dismiss_button' => false,
+			'message'        => sprintf(
+			// translators: %1$s = <strong>, %2$s = </strong>, %3$s = <a>, %4$s = </a>.
+				__( '%1$sWP Rocket:%2$s We detected that the website domain has changed. The configuration files must be regenerated for the page cache and all other optimizations to work as intended. %3$sLearn More%4$s', 'rocket' ),
+				'<strong>',
+				'</strong>',
+				'<a href="https://docs.wp-rocket.me/article/705-changing-domains-migrating-sites-with-wp-rocket">',
```
remyperona: Shouldn't we use beacon here to get the appropriate link in EN/FR?

inc/Engine/Admin/DomainChange/Subscriber.php
```diff
@@ -124,4 +145,100 @@ public function maybe_clean_cache_domain_change() {
...
+			return;
+		}
+
+		$last_base_url_encoded = get_transient( 'rocket_domain_changed' );
+
+		if ( ! $last_base_url_encoded ) {
+			return;
+		}
+
+		$last_base_url = base64_decode( $last_base_url_encoded ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_decode
```
remyperona, line 210: why are we using base64 here?

inc/Engine/Common/Ajax/AjaxHandler.php
```diff
@@ -0,0 +1,37 @@
...
+	 * Validate the referer before going futher.
+	 *
+	 * @param array $args Arguments to configure the validation.
+	 *
+	 * @return bool
+	 */
+	public function validate_referer( array $args ) {
+
+		if ( key_exists( 'referer', $args ) && is_string( $args['referer'] ) ) {
+			check_admin_referer( $args['referer'] );
```
remyperona: I think the index name should be `action` instead of `referer` to match with the parameter name of the function

inc/Engine/Common/Ajax/AjaxHandler.php
```diff
@@ -0,0 +1,37 @@
...
+class AjaxHandler {
+
+	/**
+	 * Validate the referer before going futher.
+	 *
+	 * @param array $args Arguments to configure the validation.
+	 *
+	 * @return bool
+	 */
+	public function validate_referer( array $args ) {
```
remyperona: I don't like using array much in method parameters, it's hiding what is expected and creates the need to validate the array key/values passed.

For an AJAX request we know what we are going to validate is the nonce/referer and the capability, so we can pass those as parameter directly.

remyperona (changes requested):

inc/Engine/Admin/DomainChange/Subscriber.php
CrochetFeve0251: Base 64 here is due to some hosts doing operation on site URL options when we change the domain

remyperona: why base64 instead of md5 for example?

inc/Engine/Common/Ajax/AjaxHandler.php
```diff
@@ -5,19 +5,19 @@
...
-	 *
-	 * @param array $args Arguments to configure the validation.
+	 * Validate the referer before going further.
 	 *
+	 * @param string $action Action to validate.
+	 * @param string $capacities User capacity to verify.
 	 * @return bool
 	 */
-	public function validate_referer( array $args ) {
+	public function validate_referer( string $action = '', string $capacities = '' ) {
```
remyperona: the parameters are not optionals, so I don't think it's necessary to give them a default empty value. That will avoid a couple of conditionals in the method too

remyperona (changes requested):

remyperona (approved):

vmanthos:
I checked the PR and found the following issues after clicking on the "Regenerate WP Rocket configuration files now" button:
- The configuration file isn't regenerated - this is necessary when the settings we store in that file have been changed.
- The cache of the site where the content is pushed isn't cleared if no changes to WP Rocket's settings have been made. If changes have been made to a page, or a new post is published (and the homepage is set to display the posts) the changes won't be visible.

@CrochetFeve0251 Can you please look into these? :pray:

piotrbak:
@CrochetFeve0251 @vmanthos Is this one ready for the QA?

vmanthos:
@piotrbak I'm currently testing the latest commit to see if the [reported issues](https://github.com/wp-media/wp-rocket/pull/5688#issuecomment-1658070902) are fixed.

vmanthos:
@piotrbak In multisites, the notice is still displayed but the button to regenerate the files and clear the cache is missing:
image

Is this the expected outcome or shouldn't we display the notice at all?

vmanthos:
@CrochetFeve0251 Can you please check Remy's comment?
https://github.com/wp-media/wp-rocket/pull/5688#discussion_r1262975246

Currently, the URL leads directly to the EN documentation, i.e. doesn't open in the Beacon, and will be the same regardless of the language the user is using, e.g. French.

vmanthos:
@CrochetFeve0251 The following issues have been fixed:
* Links aren't displayed in Beacon.
* Notice is displayed in multisites.

However, now the button to regenerate WP Rocket's files after a "push" isn't displayed. I believe it's related to the removal of `$args['action'] = 'regenerate_configuration';` in the following commit:
https://github.com/wp-media/wp-rocket/pull/5688/commits/7c67bb3ad102f340a2428432d1cc24ac21a4a6df#diff-9055161c7bbe500619682e326d32655472afc64ecaae55868d5e0225d330f53aL192-L195

vmanthos (approved):
Thank you @CrochetFeve0251 for your continuous effort on this PR.

 This is working as expected! 🎉

**TestRail report:** [testrail-report-511.pdf](https://github.com/wp-media/wp-rocket/files/12258683/testrail-report-511.pdf)

vmanthos:
@piotrbak @wp-media/php 

While testing another PR I found that the notice will be displayed when:
1. A translation plugin is used, e.g. WPML.
2. The setting to display or not the default language's code in the permalink is enabled/disabled.

For WPML, we are already handling cache clearing when the specific option changes. 

With the notice displayed, the WMPL users who will click the "Regenerate WP Rocket configuration files now" button will have the cache cleared again and additionally, the `advanced-cache.php` and `.htaccess` files will be regenerated.

How should we handle this?

piotrbak:
@vmanthos Are there any changes in our config file (new one generated when doing a visit) or .htaccess (change in rewrite rules after regenerating it) when changing this option in WPML?

vmanthos:
The only change I see is in the rewrite rules:
`RewriteBase /` becomes `RewriteBase /en/` and vice versa depending on whether the option to display `/en/` is enabled or disabled.

piotrbak:
@vmanthos So there's a value of doing it 🤔

vmanthos:
If there is, then we missed something when we added the fix for this:
https://github.com/wp-media/wp-rocket/issues/5761

In any case, this was a random finding and it would be good to investigate it further to make sure:
1. If it's indeed useful.
2. If it can cause side effects.

piotrbak:
Probably the home_url or site_url is changed when doing this operation. Now I'm thinking, rewrite rules might not be working correctly for different languages if they're set to `/en/`, but that's not a regression, since it'd happen when regenerating htaccess anyways.

piotrbak:
@CrochetFeve0251 Do you think it'd possible not to display the message based on the action that triggered it?

CrochetFeve0251:
@piotrbak what do you mean by action? I need some context.

piotrbak:
@CrochetFeve0251 It's about this one:
```
The setting to display or not the default language's code in the permalink is enabled/disabled.
```

So, we're on the WPML Languages Settings page and we're changing this one:
image

Pinging @vmanthos for confirmation

vmanthos:
Exactly.

PR #5704, opened 2023-01-20
Closes #5643 Added $_GET inside logs for cache

### LINKED ISSUE #5643: Revealing hidden query string on Debug
- [x] Made sure you’re on the latest version `3.12.3.3`

**Describe the bug**
If `$_GET` is customized for whatever reason, it will force WP Rocket to exclude pages from cache and optimizations. 

The only clue we get when `WP_ROCKET_RUCSS_DEBUG` is the cause of the rejection but no hint about what is the offending query string. 

Support will need that information to: 

- Help the customer to find where it comes from. 
- Make the found query string(s) cacheable by manually adding them to `Cache Query Strings` if the root cause can't be fixed.

This can be done by changing https://github.com/wp-media/wp-rocket/blob/58784bd05e3919595ed453870f69aeb3f580e8b2/inc/classes/Buffer/class-tests.php#L200 to: 

```
$this->set_error( 'Query string URL is excluded:' . PHP_EOL . print_r( $_GET, true ) );
```

This will output something like the following on the debugger: 

```
Query string URL is excluded:
Array
(
    [hidden-query-string] => query-string-value
)
```

**To Reproduce**
N/A

**Expected behavior**
Support should have more information about the possibly offending query strings. 

**Screenshots**
N/A

**Additional context**
Ticket - https://secure.helpscout.net/conversation/2104097765/390446/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description
This pull request is to fix an issue with the lack of information provided when WP Rocket (a caching plugin for WordPress) excludes pages from cache and optimizations due to a customized $_GET variable. This can make it difficult for support to help the customer find and fix the root cause, or manually add the query string(s) to the Cache Query Strings.

Fixes #5643

## Type of change

The solution is the same as the one proposed during grooming.

## How Has This Been Tested?

- [ ] Test automatically
- [ ] Test B

# Checklist:


remyperona (approved):

Mai-Saad (approved):
Working as expected

PR #5712, opened 2023-01-23
Change the way we detect One.com CDN option

### LINKED ISSUE #5708: Compatibility between One.com CDN and security plugins

**Describe the bug**
One.com's CDN is limited in rewriting directories different than `wp-content`. When using security plugins that are changing default `wp-content` path, files won't be hosted on the edges while we'll still rewrite them. It can cause broken layout.

**To Reproduce**
1. Go to One.com website 
2. Install WP Hide plugin or similar and change `wp-content` path
3. See error

**Expected behavior**
Until this is handled properly on One.com's end, we'll disable our CDN compatibility when detecting that `wp-content` directory path is changed.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
## Description

Here we changed the way we detect one.com CDN option and added a condition to get the CDN option value from the database only when their performance plugin is enabled by checking if the value of constant `vcaching` which is defined there like this:

image

Fixes #5708

## Type of change

[XS] effort, not groomed.

## How Has This Been Tested?

Checked it with @piotrbak 

# Checklist:

### DIFF
--- a/inc/ThirdParty/Hostings/OneCom.php
+++ b/inc/ThirdParty/Hostings/OneCom.php
@@ -142,7 +142,7 @@ class OneCom implements Subscriber_Interface {
 	 * @return boolean
 	 */
 	public function is_varnish_active() {
-		return rest_sanitize_boolean( get_option( 'varnish_caching_enable' ) );
+		return rocket_get_constant( 'vcaching', false ) && rest_sanitize_boolean( get_option( 'varnish_caching_enable' ) );
 	}
 
 	/**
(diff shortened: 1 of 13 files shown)


piotrbak:
For the @wp-media/qa team. In this PR we need to make sure that our auto-enabled CDN for One.com is activated **only** when:
1. Their `Performance Cache` plugin is enabled
2. In their settings CDN is set to `On`

In the previous version the first check didn't exist.

remyperona (approved):

vmanthos (approved):

**Notes**

* Enabling/disabling the CDN option in the Performance Cache plugin doesn't clear the cache. 
* If the CDN option is active and the Performance Cache is deactivated we don't clear the cache.
The same happens when activating the plugin while the CDN option is enabled. 

These result in either still serving content from the CDN when we shouldn't or the other way around.

PR #5715, opened 2023-01-25
Added a filter to change endpoint on RUCSS

CrochetFeve0251:
## Description

Add a filter on RUCSS endpoint to add params on wp-rocket?me

## Type of change

## How Has This Been Tested?

- [ ] Test in local

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
+++ b/inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
@@ -82,7 +82,13 @@ abstract class AbstractAPIClient {
 		$args['method'] = strtoupper( $type );
 
 		$response = wp_remote_request(
-			$api_url . $this->request_path,
+			/**
+			 * RUCSS SAAS endpoint.
+			 *
+			 * @param string $url endpoint url.
+			 * @returns string
+			 */
+			apply_filters( 'rocket_rucss_api_endpoint', $api_url . $this->request_path ),
 			$args
 		);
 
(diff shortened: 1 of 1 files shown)


remyperona (changes requested):
I don't think being able to modify the URL for this is something we want to have broadly available. Could it be possible to filter and use the `$args` data instead?

inc/Engine/Optimization/RUCSS/AbstractAPIClient.php
```diff
@@ -81,6 +81,14 @@ private function handle_request( array $args, string $type = 'post' ) {
 
 		$args['method'] = strtoupper( $type );
 
+		/**
+		 * RUCSS SAAS request arguments.
+		 *
+		 * @param array $args request arguments.
+		 * @returns array
+		 */
+		$args = apply_filters( 'rocket_rucss_api_endpoint', $args );
```
remyperona, line 87: I think that's not necessary, on the website side you can use the https://developer.wordpress.org/reference/hooks/http_request_args/ filter to modify the args for this specific request

CrochetFeve0251: @Tabrisrp I could but it would heavily couple the code from the other plugin with the RUCSS url.

remyperona (approved):

vmanthos:
@CrochetFeve0251 I believe we should safeguard the value returned by the filter. 

Also, if the `$args` don't include the `credentials` we'll disable Remove Unused CSS and display the warning:

> We couldn't generate the used CSS because you're using a nulled version of WP Rocket. You need an active license to use the Remove Unused CSS feature and further improve your website's performance.

remyperona:
That's why I don't really like this, we're adding a filter which only purpose is to be used for our website and not for a public usage. And it causes this kind of constraints where we have to check for correct usage of the filter, safeguard, while there is already a way to achieve the same thing without adding a new filter.

PR #5718, opened 2023-01-26
Set minimum Action Scheduler version

### LINKED ISSUE #5638: Set minimum Action Scheduler required version to avoid conflicts with older versions

**Describe the bug**
On sites running an older version of **Action Scheduler**, for example when using WooCommerce v3.5.4 which has AS v2.1.0, or they are using WooCommerce v3.9.5, which includes AS v2.2.5 the following fatal error will happen:

`
Fatal error: Uncaught Error: Class 'ActionScheduler_StoreSchema' not found in /public_html/wp-content/plugins/wp-rocket/inc/Engine/HealthCheck/ActionSchedulerCheck.php:45 Stack trace: #0 /public_html/wp-includes/class-wp-hook.php(308): WP_Rocket\Engine\HealthCheck\ActionSchedulerCheck->maybe_recreate_as_tables('') #1 /public_html/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters(NULL, Array) #2 /public_html/wp-includes/plugin.php(517): WP_Hook->do_action(Array) #3 /public_html/wp-content/plugins/wp-rocket/inc/Engine/Activation/Activation.php(73): do_action('rocket_activati...') #4 /public_html/wp-includes/class-wp-hook.php(308): WP_Rocket\Engine\Activation\Activation::activate_plugin('') #5 /home/ in /public_html/wp-content/plugins/wp-rocket/inc/Engine/HealthCheck/ActionSchedulerCheck.php on line 45
`

**To Reproduce**
1. Install an older version of WooCommerce
2. Activate WP Rocket
3. See error

**Expected behavior**
We should guard against this error by checking the minimum version of Action Scheduler 

**Additional context**

as a temporary fix, Installing the standalone Action Scheduler plugin solves the issue.

Woocommerce version: https://i.imgur.com/4B57apU.png
ticket https://secure.helpscout.net/conversation/2102275631/390086?folderId=2683093
slack https://wp-media.slack.com/archives/C43T1AYMQ/p1671562361670519

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
## Description

Given we have WC old version for example v 3.5.4 which has AS v2.1.0
While activating WPR, the activation code is called before loading our Action Scheduler's code (or the latest version) so only in this instance our activation code is loaded with the old version of AS but after enabling WPR, it loads our latest version.

So reproducing steps are like that:
1. Install and activate WC 3.5.4
2. Install and activate WPR
3. See fatal error with activation

I believe we have two options here:
1. Find a way to load our AS with activation
2. Add the guard into the activation code along with disabling preload with activation only if we detected older version of AS

First option is impossible because the hook `plugins_loaded` is already loaded with the old action scheduler code so the instance of `ActionScheduler` class are already there, also the new AS code will not instantiate those classes again.
everything is attached to this `plugins_loaded` hook like here

https://github.com/wp-media/wp-rocket/blob/103abf7be9d8acd6ea3abae742b533bfd9cd4c32/inc/Dependencies/ActionScheduler/action-scheduler.php#L36

Even if I called the functions manually before activation it'll not run because of this line:

https://github.com/wp-media/wp-rocket/blob/103abf7be9d8acd6ea3abae742b533bfd9cd4c32/inc/Dependencies/ActionScheduler/action-scheduler.php#L53

## What is the exact problem of having the old version of Action Scheduler?
We have two problems as below:

1. This [class](https://github.com/wp-media/wp-rocket/blob/41e3eddcaa34e679e495f3517745baa232232012/inc/Engine/HealthCheck/ActionSchedulerCheck.php#L45) isn't there, this part is responsible for creating AS's table again if they are not there => here we may skip this part if the action scheduler is old one.
2. This [function](https://github.com/wp-media/wp-rocket/blob/fbbc98f992c5aaf546ce581794aed90b63502966/inc/Engine/Common/Queue/AbstractASQueue.php#L28) is called in preload and it's not there in the old version of AS => here we solved this by checking if the function exists otherwise schedule single action because scheduling async is there from version 3.0.0.

And I checked the Action Scheduler history and found that this versioning feature (that always loads the most updated version of action scheduler) was there from version 1.0

So this PR solved the above two issues with exactly mentioned methods

Fixes #5638 

## Type of change

No

## How Has This Been Tested?

- [x] Tested it on my test site and all good.

# Checklist:


remyperona (approved):

vmanthos (approved):
Working as expected. :+1: 

NOTE
There is a PHP fatal error in one of the test cases I checked but according to @engahmeds3ed this is not related to us.
[testrail-report-445.pdf](https://github.com/wp-media/wp-rocket/files/10576309/testrail-report-445.pdf)

vmanthos (approved):

PR #5728, opened 2023-01-30
Added a way to change delete old cache interval from preload

### LINKED ISSUE #5657: Add a filter to disable the removal of non accessed URLs from the wpr_rocket_cache table
The case is a customer who wants to keep all the URLs in the cache, even if they are not accessed during the past 30 days

Add a boolean filter to turn on/off this feature here
https://github.com/wp-media/wp-rocket/blob/301916893739f108e87b181b52485a917df7abe2/inc/Engine/Preload/Cron/Subscriber.php#L227

or here:
https://github.com/wp-media/wp-rocket/blob/301916893739f108e87b181b52485a917df7abe2/inc/Engine/Preload/Database/Queries/Cache.php#L290

- A cron-triggered snippet to update the `last_accessed` row to the current DateTime to all `completed` URLs.
But this is an overkill and requires too much writing into the database 
- unhook rocket_preload_clean_rows_time_event / remove_old_rows
-
**Additional context**
ticket; https://secure.helpscout.net/conversation/2096924274/388876?folderId=2683093

CrochetFeve0251:
## Description
Gives the opportunity to change the delay before deletion of old not accessed row in the preload or totally disable it.

Fixes #5657

## Type of change

No.

## How Has This Been Tested?

- [ ] Automated test

# Checklist:

### DIFF
--- a/inc/Engine/Preload/Cron/Subscriber.php
+++ b/inc/Engine/Preload/Cron/Subscriber.php
@@ -234,7 +234,6 @@ class Subscriber implements Subscriber_Interface {
 		 * Delay before the not accessed row is deleted.
 		 *
 		 * @param string $delay delay before the not accessed row is deleted.
-		 * @returns string
 		 */
 		$delay = (string) apply_filters( 'rocket_preload_delay_delete_non_accessed', '1 month' );
 
--- a/inc/Engine/Preload/Cron/Subscriber.php
+++ b/inc/Engine/Preload/Cron/Subscriber.php
@@ -79,7 +79,6 @@ class Subscriber implements Subscriber_Interface {
 		 * Delay before the not accessed row is deleted.
 		 *
 		 * @param string $delay delay before the not accessed row is deleted.
-		 * @returns string
 		 */
 		$delay = (string) apply_filters( 'rocket_preload_delay_delete_non_accessed', '1 month' );
 
--- a/inc/Engine/Preload/Cron/Subscriber.php
+++ b/inc/Engine/Preload/Cron/Subscriber.php
@@ -251,7 +251,7 @@ class Subscriber implements Subscriber_Interface {
 			$unit   = $parts[1];
 		}
 
-		$this->query->remove_all_not_accessed_rows( $delay, $unit );
+		$this->query->remove_all_not_accessed_rows( $value, $unit );
 	}
 
 	/**
--- a/inc/Engine/Preload/Cron/Subscriber.php
+++ b/inc/Engine/Preload/Cron/Subscriber.php
@@ -246,7 +246,7 @@ class Subscriber implements Subscriber_Interface {
 		$value = 1;
 		$unit  = 'month';
 
-		if ( count( $parts ) === 2 ) {
+		if ( count( $parts ) === 2 || $parts[0] >= 0 ) {
 			$value = $parts[0];
 			$unit  = $parts[1];
 		}
--- a/inc/Engine/Preload/Cron/Subscriber.php
+++ b/inc/Engine/Preload/Cron/Subscriber.php
@@ -83,7 +83,7 @@ class Subscriber implements Subscriber_Interface {
 		 */
 		$delay = (string) apply_filters( 'rocket_preload_delay_delete_non_accessed', '1 month' );
 
-		if ( $delay === "" || wp_next_scheduled( 'rocket_preload_clean_rows_time_event' ) ) {
+		if ( '' === $delay || wp_next_scheduled( 'rocket_preload_clean_rows_time_event' ) ) {
 			return;
 		}
 
--- a/inc/Engine/Preload/Cron/Subscriber.php
+++ b/inc/Engine/Preload/Cron/Subscriber.php
@@ -246,7 +246,7 @@ class Subscriber implements Subscriber_Interface {
 		$value = 1;
 		$unit  = 'month';
 
-		if ( count( $parts ) === 2 || $parts[0] >= 0 ) {
+		if ( count( $parts ) === 2 && $parts[0] >= 0 ) {
 			$value = $parts[0];
 			$unit  = $parts[1];
 		}
--- a/inc/Engine/Preload/Cron/Subscriber.php
+++ b/inc/Engine/Preload/Cron/Subscriber.php
@@ -244,11 +244,11 @@ class Subscriber implements Subscriber_Interface {
 		}
 
 		$value = 1;
-		$unit   = 'month';
+		$unit  = 'month';
 
 		if ( count( $parts ) !== 2 ) {
 			$value = $parts[0];
-			$unit   = $parts[1];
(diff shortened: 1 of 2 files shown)


mostafa-hisham (approved):

inc/Engine/Preload/Cron/Subscriber.php
```diff
@@ -75,8 +75,15 @@ public static function get_subscribed_events() {
 	 * @return void
 	 */
 	public function schedule_clean_not_commonly_used_rows() {
+		/**
+		 * Delay before the not accessed row is deleted.
+		 *
+		 * @param string $delay delay before the not accessed row is deleted.
+		 * @returns string
```
remyperona:
```suggestion
```
not necessary to add the return value for filters and actions

inc/Engine/Preload/Cron/Subscriber.php
```diff
@@ -224,7 +231,19 @@ public function schedule_revert_old_in_progress_rows() {
 	 * @return void
 	 */
 	public function remove_old_rows() {
-		$this->query->remove_all_not_accessed_rows();
+		/**
+		 * Delay before the not accessed row is deleted.
+		 *
+		 * @param string $delay delay before the not accessed row is deleted.
+		 * @returns string
```
remyperona:
```suggestion
```

inc/Engine/Preload/Database/Queries/Cache.php
```diff
@@ -289,7 +291,7 @@ public function get_old_cache() : array {
 		}
 
 		$prefixed_table_name = $db->prefix . $this->table_name;
-		$query               = "SELECT id FROM `$prefixed_table_name` WHERE `last_accessed` <= date_sub(now(), interval 1 month)";
+		$query               = "SELECT id FROM `$prefixed_table_name` WHERE `last_accessed` <= date_sub(now(), interval $delay)";
```
remyperona: why not use the filter here also instead of adding a parameter to the method?

remyperona (changes requested):

inc/Engine/Preload/Database/Queries/Cache.php
CrochetFeve0251: @Tabrisrp is some previous CR @engahmeds3ed told me to not add filters inside filters inside the Query.

remyperona: I don't understand why it would be different from the current implementation 🤔

CrochetFeve0251: @Tabrisrp  You have an example of current implementation [here](https://github.com/wp-media/wp-rocket/blob/ef300cf2d46e837789f9ad42ff8166c832a47dc3/inc/Engine/Preload/Database/Queries/Cache.php#L666) where the filter is declared [here](https://github.com/wp-media/wp-rocket/blob/ef300cf2d46e837789f9ad42ff8166c832a47dc3/inc/Engine/Preload/Controller/PreloadUrl.php#L191).

remyperona:
```suggestion
               $delay = (string) apply_filters( 'rocket_preload_delay_delete_non_accessed', '1 month' );
		$query               = "SELECT id FROM `$prefixed_table_name` WHERE `last_accessed` <= date_sub(now(), interval $delay)";
```

I still don't understand why we can't do that?

engahmeds3ed: Let's discuss about it and at last we get a rule to be applied.

Sure filters can be added anywhere, what I wanted to say is this is the query class which deals directly with the database so those methods here should be more generic and deals with arguments instead of using WordPress hooks here.

I think this is to separate the business logic layer from database layer (Separation of concerns concept).

This is my proposed idea, plz let me know what do u think about it?

remyperona: if we do it like that we need to find a different way to pass the value to the parameter, a string like this is not great

CrochetFeve0251: @Tabrisrp I changed the method params to adapt to it.

inc/Engine/Preload/Cron/Subscriber.php
```diff
@@ -237,11 +237,21 @@ public function remove_old_rows() {
...
 		}
 
-		$this->query->remove_all_not_accessed_rows( $delay );
+		if ( count( $parts ) !== 2 ) {
+			$delay = $parts[0];
+			$unit  = $parts[1];
+		} else {
+			$delay = 1;
+			$unit  = 'month';
+		}
```
remyperona:
```suggestion
		$value = 1;
		$unit   = 'month';

		if ( count( $parts ) !== 2 ) {
			$value = $parts[0];
			$unit   = $parts[1];
		}
```

inc/Engine/Preload/Cron/Subscriber.php
```diff
@@ -237,11 +237,21 @@ public function remove_old_rows() {
...
+
+		$this->query->remove_all_not_accessed_rows( $delay, $unit );
```
remyperona:
```suggestion
		$this->query->remove_all_not_accessed_rows( $value, $unit );
```

inc/Engine/Preload/Database/Queries/Cache.php
```diff
@@ -300,11 +300,11 @@ public function get_old_cache( string $delay = '1 month' ) : array {
...
 	 *
-	 * @param string $delay delay before the not accessed row is deleted.
-	 *
+	 * @param int    $delay delay before the not accessed row is deleted.
+	 * @param string $unit unit from the delay.
 	 * @return void
 	 */
-	public function remove_all_not_accessed_rows( string $delay = '1 month' ) {
+	public function remove_all_not_accessed_rows( int $delay = 1, string $unit = 'month' ) {
 		$rows = $this->get_old_cache( $delay );
```
remyperona:
```suggestion
		$rows = $this->get_old_cache( $delay, $unit );
```

remyperona (changes requested):

remyperona (approved):

Mai-Saad:
@CrochetFeve0251 Thanks for the PR.
- If filter was set to '-1 year'  or '0.5 year'(probably shouldn't be a case , but if it happened accidentally, then we executed the cron => all cache will be cleared. Can we guard this?
- Note : If filter set to Null, ' ', false, we won't remove anything from cache. while if the filter was set to 0 , 1 , true , -1, it will behave as default (clear URLs with last accessed > 1 month)

Mai-Saad:
working as expected notice that.
`0, null, '', false => do nothing`
`<0 and >0 , true  ' year'=> use default`
`'1 year' => use 1 year`
[testrail-report-460.pdf](https://github.com/wp-media/wp-rocket/files/10969723/testrail-report-460.pdf)

PR #5730, opened 2023-01-31
Closes #5687 Remove multiple config files on deactivation

### LINKED ISSUE #5687: WP Rocket’s config files whose filenames don’t match the site’s domain are not removed upon plugin deactivation
- Made sure you’re on the latest version (tested on 3.12.4 and 3.12.5)
- Used the search feature to ensure that the bug hasn’t been reported before ✓

**Describe the bug**
The rules we add to the .htaccess file are not cleared when there is a PHP file left in the `/wp-rocket-config/` folder that has a different filename (i.e., it does not match the site’s domain).

This causes the cached files to still be served even after deactivating WP Rocket on the site (in those cases the rewrite rules are still present) causing the server to deliver the cached files.

**To Reproduce**
1. Visit a page while having WP Rocket enabled to see the cached file.
2. Duplicate the config file from the `/wp-rocket-config/ `folder and give it a filename different from the site’s domain.
3. Deactivate WP Rocket.
4. Check the .htaccess and see that our rules are astill there.
5. Visit the previoiusly visited page and see that the cached file is still served.

**Expected behavior**
The rules we place to the .htaccess file should be removed upon deactivating the plugin. 

We should make sure we remove all config files present in the /wp- rocket-config/ folder upon deactivating the plugin, no matter which filenames they have. Or find a way to remove the .htaccess rules if additional config files are kept, to avoid the issue.

**Screenshots**
Screencast: https://share.getcloudapp.com/E0uRXKbN

**Additional context**
Related HS ticket: https://secure.helpscout.net/conversation/2114970779/393122/
Props to @alfonso100

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description
This PR solves the issue where we do not clear out irrelevant config files when WP Rocket is deactivated.

Fixes #5687 

## Type of change

Slightly different with a helpful addition from @engahmeds3ed 

## How Has This Been Tested?
Manual Tests

# Checklist:

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -233,8 +233,8 @@ function rocket_delete_config_file() {
 		rocket_direct_filesystem()->delete( $config_file );
 	}
 
-	// Bail out if WP Rocket is not network active.
-	if ( is_multisite() && ! is_plugin_active_for_network( plugin_basename( WP_ROCKET_FILE ) ) ) {
+	// Bail out if WP Rocket is multisite.
+	if ( is_multisite() ) {
 		return;
 	}
 
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -232,6 +232,24 @@ function rocket_delete_config_file() {
 	foreach ( $config_files_path as $config_file ) {
 		rocket_direct_filesystem()->delete( $config_file );
 	}
+
+	// Bail out if WP Rocket is not network active.
+	if ( is_multisite() && ! is_plugin_active_for_network( plugin_basename( WP_ROCKET_FILE ) ) ) {
+		return;
+	}
+
+	try {
+		$config_dir = new FilesystemIterator( (string) rocket_get_constant( 'WP_ROCKET_CONFIG_PATH' ) );
+	} catch ( Exception $e ) {
+		return;
+	}
+
+	// Remove all files with php extension in the config folder.
+	foreach ( $config_dir as $file ) {
+		if ( $file->isFile() && 'php' === $file->getExtension() ) {
+			rocket_direct_filesystem()->delete( $file->getPathname() );
+		}
+	}
 }
 
 /**
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -246,7 +246,7 @@ function rocket_delete_config_file() {
 
 	// Remove all files with php extension in the config folder.
 	foreach ( $config_dir as $file ) {
-		if ( ! $file->isFile() || 'php' !== $file->getExtension() ) {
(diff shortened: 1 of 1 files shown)


inc/functions/files.php
```diff
@@ -232,6 +232,24 @@ function rocket_delete_config_file() {
 	foreach ( $config_files_path as $config_file ) {
 		rocket_direct_filesystem()->delete( $config_file );
 	}
+
+	// Bail out if WP Rocket is not network active.
+	if ( is_multisite() && ! is_plugin_active_for_network( plugin_basename( WP_ROCKET_FILE ) ) ) {
```
remyperona: WP Rocket is never network active, it's not compatible with it

remyperona (changes requested):

inc/functions/files.php
```diff
@@ -232,6 +232,24 @@ function rocket_delete_config_file() {
...
+		$config_dir = new FilesystemIterator( (string) rocket_get_constant( 'WP_ROCKET_CONFIG_PATH' ) );
+	} catch ( Exception $e ) {
+		return;
+	}
+
+	// Remove all files with php extension in the config folder.
+	foreach ( $config_dir as $file ) {
+		if ( $file->isFile() && 'php' === $file->getExtension() ) {
+			rocket_direct_filesystem()->delete( $file->getPathname() );
+		}
```
engahmeds3ed: Bailout early
```suggestion
		if ( ! $file->isFile() || 'php' !== $file->getExtension() ) {
			continue;
		}
		rocket_direct_filesystem()->delete( $file->getPathname() );
```

engahmeds3ed (approved):

Mai-Saad:
Working fine for the single site, however, on a multisite, deactivating WPR from the site won't remove the wrong .php and thus won't remove WPR from the htaccess. 

Steps:
1- install WPR on multisite (using Apache)
2- Activate WPR on the main site or subsite
3- Duplicate the config file from the /wp-rocket-config/ folder and give it a filename different from the site’s domain.
4- Deactivate WP Rocket.
5- Check the .htaccess and see that our rules are still there.

=> expected after step 4 is removing the wrong config file and removing our rules from htaccess. @piotrbak  @jeawhanlee  what do you think?

Note: if we cached a page after step 2 and then visited the same page after step 5, it will be served from the nocache which is correct

**Update:**

As per discussion with @piotrbak We will keep it as is for multisite

**Update**
Multisite isnot handled in certain case already existing even if no wrong config file is there https://github.com/wp-media/wp-rocket/issues/5831

Mai-Saad (approved):

PR #5739, opened 2023-02-03
Closes #5724 Clean up failed RUCSS after a short period of time

### LINKED ISSUE #5724: Cleanup failed RUCSS after a short period of time
- [x] Made sure you’re on the latest version `3.12.5.2`

**Describe the bug**
When RUCSS fails for a specific URL after 3 retries, the only scenario where it will be renewed is when: 
- RUCSS table is truncated in full
- When the page the page Used CSS is manually cleared
- When the old Used CSS is deleted after 1 month taking `last_accessed` as a reference

If none of the action is taken, the page will remain without RUCSS for a long time for no specific reason. 

Here is an example: 

image

The URL didn't have its Used CSS renewed after 7 days, and still going without taking any action.

Having a background process that clears failed RUCSS in a shorter period, like in 3 days, for example, will give a better RUCSS ratio.

The 3 days should be enough to get the reason why it failed for a specific URL when debugging. 

A temporary fix is creating a cron and make it trigger the following file on the root folder of a WordPress installation: 
https://gist.github.com/DahmaniAdame/3a2a550b4a201ad39516d842207c884e

**To Reproduce**
1. Push RUCSS to fail in a URL
2. See how long it takes for it to be cleared if no action was done

**Expected behavior**
We should clear failed RUCSS in a shorter period of time to get a better RUCSS ratio. 

**Screenshots**

**Additional context**
Ticket - https://secure.helpscout.net/conversation/2135056677/397038/

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

jeawhanlee:
## Description
This PR schedules a cron to clear RUCSS failed jobs every 3 days.

Fixes #5724 

## Type of change

Yes.

## How Has This Been Tested?
Automated Tests

# Checklist:


remyperona (approved):

piotrbak:
@jeawhanlee After talking with @Mai-Saad .

In this PR it'd be much better to set the row to `pending` and reset the `retries` column. Not sure if we have that kind of function in the codebase 🤔

remyperona (approved):

Mai-Saad:
@jeawhanlee @piotrbak Thanks for the update. Please find my notes/qs below: (smoke test is done after merging develop local to the PR)
1- Now whenever the cron runs, we will revert all failed URLs to pending. Shouldn't we only revert the ones which failed since >= 3days?
2- Reverted URLs still have the error msg in the database. I think it shall be cleared while reverting to pending as same as retries. can you please check?

piotrbak:
@Mai-Saad @jeawhanlee Agree with both QA points 🙏

remyperona (approved):

Mai-Saad:
@jeawhanlee Thanks for the update. Currently, the msg is cleared and we revert only if failed since >= 3d. 
One more thing we need to handle here, When reverting the status, we need to update the modified date. Can you please check that?

remyperona (approved):

Mai-Saad:
@jeawhanlee the last modified date is updated, however, if the filter was set to `null `or `' '`, we will have error in logs. Can we guard this?
`add_filter( 'rocket_remove_rucss_failed_jobs_cron_interval', function( $interval ){
return null;
} );`

`[20-Mar-2023 10:15:22 UTC] Cron reschedule event error for hook: rocket_remove_rucss_failed_jobs, Error code: invalid_schedule, Error message: Event schedule does not exist., Data: {"schedule":"rocket_remove_rucss_failed_jobs","args":[],"interval":null}`

inc/Engine/Optimization/RUCSS/Cron/Subscriber.php
```diff
@@ -165,7 +165,9 @@ public function add_interval( $schedules ) {
 		 *
 		 * @param int $interval Interval in seconds.
 		 */
-		$interval = apply_filters( 'rocket_remove_rucss_failed_jobs_cron_interval', 3 * rocket_get_constant( 'DAY_IN_SECONDS', 86400 ) );
+		$default_interval = 3 * rocket_get_constant( 'DAY_IN_SECONDS', 86400 );
+		$interval         = apply_filters( 'rocket_remove_rucss_failed_jobs_cron_interval', $default_interval );
```
remyperona: the filter comment needs to be next to the filter use

remyperona (changes requested):

remyperona (approved):

Mai-Saad:
@jeawhanlee Thanks for the update, no error now with null , ''. However, can see error with -ve  value. Do you think we need to handle this? cc @piotrbak 
`[24-Mar-2023 07:09:13 UTC] Cron reschedule event error for hook: rocket_remove_rucss_failed_jobs, Error code: could_not_set, Error message: The cron event list could not be saved., Data: {"schedule":"rocket_remove_rucss_failed_jobs","args":[],"interval":-1}`

piotrbak:
Thank you @Mai-Saad 

I believe we're good here, it's a bit edgy and not causing fatal errors. That's acceptable.

Mai-Saad (approved):
Working fine then, failed test due to other known issue.
[testrail-report-465.pdf](https://github.com/wp-media/wp-rocket/files/11076087/testrail-report-465.pdf)

DahmaniAdame:
@piotrbak currently, only the cron interval is controllable using the `rocket_remove_rucss_failed_jobs_cron_interval` filter: 

https://github.com/wp-media/wp-rocket/blob/3b6300ec350c3fa60bdae03c32bc8d5c4b9cb7f9/inc/Engine/Optimization/RUCSS/Cron/Subscriber.php#L169

While failed jobs we are looking for on the database have a static 3 days age: 

https://github.com/wp-media/wp-rocket/blob/3b6300ec350c3fa60bdae03c32bc8d5c4b9cb7f9/inc/Engine/Optimization/RUCSS/Database/Queries/UsedCSS.php#L423

Which nullifies the effect of `rocket_remove_rucss_failed_jobs_cron_interval`. Regardless of its value, it will still be a static 3 days minimum of unavailable Used CSS without manual interaction.

Both need to be controllable using a filter for cases where the 3 days interval failed lifetime are not tolerable. At least, until we have more accurate errors reporting from SaaS To know what is causing `40x` errors.

PR #5746, opened 2023-02-07
#5734 Incorrect renewal price on banner for users from 2022

### LINKED ISSUE #5734: Expiring soon users from 2022 see incorrect auto-renewal price on dashboard banner
**Bug details:** 
Customer has the latest version of the plugin installed - 3.12.5.2 . 
The license was purchased on February 10th 2022 and it will expire soon - on the 10th of February 2023. 

The auto-renewal banner with the countdown timer on the plugin dashboard is displaying an incorrect renewal rate for the customer: $59 instead of $49. 

Customer profile: https://wp-rocket.me/wp-admin/user-edit.php?user_id=230864

image

**To Reproduce**
Steps to reproduce the behavior (if it's a general issue - to be confirmed!):
1. Install a license registered in 2022 on a website; 
2. Adjust the license's expiration date to expire in the next few days days;
3. Check the expiring soon banner on the plugin dashboard to observe incorrect renewal price. 

**Expected behavior**
The renewal rates for customers registered in 2022 should be $49 for Single, $99 for Plus and $249 for Infinite. 

**Additional context and related issues**
- The translations for the pricing line are missing as well: https://github.com/wp-media/wp-rocket/issues/5735
- The banner should not be displayed if the customers have auto-renewal on: https://github.com/wp-media/wp-rocket/issues/4641

jeawhanlee:
## Description
This PR fixes incorrect price on the renewal banner for users from 2022.

Fixes #5734 

## Type of change

Slightly.

## How Has This Been Tested?
Manual & Automated Tests

# Checklist:


inc/Engine/License/API/User.php
```diff
@@ -120,4 +120,17 @@ public function get_renewal_url() {
...
+	 * Return the actual user creation date.
+	 *
+	 * @return integer
+	 */
+	public function get_actual_creation_date(): int {
+		if ( ! isset( $this->user->renewal_url ) ) {
+			return 0;
+		}
+
+		return $this->user->date_created;
```
remyperona:
```suggestion
		if ( ! isset( $this->user->date_created ) ) {
			return 0;
		}

		return $this->user->date_created;
```

inc/Engine/License/API/User.php
```diff
@@ -120,4 +120,17 @@ public function get_renewal_url() {
 
 		return $this->user->renewal_url;
 	}
+
+	/**
+	 * Return the actual user creation date.
+	 *
+	 * @return integer
+	 */
+	public function get_actual_creation_date(): int {
```
remyperona: what do you mean by actual there? Why not directly get_creation_date()?

remyperona (changes requested):

inc/Engine/License/API/User.php
jeawhanlee: According to @piotrbak comment [here](https://github.com/wp-media/wp-rocket/issues/5734#issuecomment-1415815577) we will needed to check if the `date_created` is zero, but the actual value of `date_created` is already modified [here](https://github.com/wp-media/wp-rocket/blob/60b76bc8bc9a366bf5b9644894df8203bba103fd/inc/Engine/License/API/User.php#L67) in `get_creation_date()` that's why I created another method to get it as it is.

remyperona: we could return 0 in `get_creation_date()` instead of returning `time()`, and use `get_creation_date()` directly

remyperona (changes requested):

inc/Engine/License/API/User.php
jeawhanlee: Thanks Boss 🙏

remyperona (approved):

Mai-Saad:
@jeawhanlee Thanks for the PR, works fine for a valid date of expiring soon user (single/plus/infinite) GF/GM/NGF. However, if the auto-renew is empty while the creation date is valid, we still display the banner => it shouldn't be displayed in this case as per discussion with @piotrbak . can you please check?

inc/Engine/License/API/User.php
```diff
@@ -72,13 +72,17 @@ public function get_creation_date() {
...
 	 *
-	 * @return boolean
+	 * @return mixed
 	 */
 	public function is_auto_renew() {
 		if ( ! isset( $this->user->has_auto_renew ) ) {
 			return false;
 		}
 
+		if ( '' === $this->user->has_auto_renew ) {
```
remyperona, line 82: Is this an actual use case that can happen? Can our API return an empty value instead of true/false?

jeawhanlee: Yes I think our API could return an empty value, on my local, I currently have an empty value for `has_auto_renew`

remyperona: For me this is something that should be fixed on the API side, it should always return a boolean, not an empty string

piotrbak: @Tabrisrp @GeekPress Currently `has_auto_renew` and `has_one-com_account` are returning empty values:
```
"has_auto_renew":"","has_one-com_account":""
```
The same for `is_blocked`.

piotrbak:
This one will be fixed on the website side 🎉

PR #5763, opened 2023-02-15
exclude uncode app and plugin js files from combined and delay js

### LINKED ISSUE #5743: Additional Delay JavaScript Execution exclusions for Uncode theme
When Delay JavaScript Execution is active, Uncode's AJAX product filter does not work, even after interaction.
There is no console error, but the overlay doesn't work when clicked. In the ticket we had, the issue was specifically on mobile. 

These are the additional files which need to be excluded: 
/wp-content/themes/uncode/library/js/app.min.js
/wp-content/themes/uncode/library/js/plugins.min.js

Ticket: https://secure.helpscout.net/conversation/2141854867/398426?folderId=377611
Example of how it should work: https://jumpshare.com/j/UA7amWkv4ojlmE5ioFxh
Without exclusions, the overlay did not appear when clicked.

mostafa-hisham:
## Description

exclude uncode app.js and plugin.js files from combined and delay js to fix issues with overlay in mobile

Fixes #5743 

## Type of change

Yes a little bit we need to exclude them from minify too

## How Has This Been Tested?

- [x] Unit and Integration tests

# Checklist:


jeawhanlee (approved):

vmanthos:
I've managed to reproduce the issue on the customer's site using Chrome's local overrides and using the Delay JS Execution script of the 3.12.4 version.

With the version of the script (from 3.12.6.1) currently on the customer's website, the issue isn't happening even without excluding the scripts of this PR.

Instead, on MEGA/Uncode I found that we need to exclude `jQuery` when WooCoomerce is installed/activated.

@piotrbak Will we handle this in the current PR or should I create a new GH issue and a new PR for that?

piotrbak:
@vmanthos The main question here is why `/themes/uncode/library/js/woocommerce-uncode.js` is excluded from the Delay JS 🤔

piotrbak:
@vmanthos The problem here is that the exclusions are not case-sensitive and we're matching this file with:
https://github.com/wp-media/wp-rocket/blob/16a72f43d2ef581a9b359cc4e7f1794f7ff96f28/inc/ThirdParty/Themes/Uncode.php#L101

piotrbak:
The solution might be changing the above line to:
`$exclude_delay_js[] = 'UNCODE\.(.*)\)\;';`

vmanthos:
Nice catch, Sir! :tophat:

piotrbak:
@vmanthos Could you check if we can safely change the exclusion there? Cause now it seems that we're breaking Uncode websites with our auto-exclusions.

vmanthos:
@piotrbak In my tests, removing `UNCODE\.` from the exclusions didn't cause any console errors or issues.

This was added to exclude inline JavaScript where `UNCODE.` is used, e.g.:
```html
<script type="text/javascript">UNCODE.initHeader();</script>
```
The methods that are used with it are:
```JavaScript
initHeader
initBox
fixMenuHeight
initRow
```

These are defined in `/uncode/library/js/init.js` which is automatically excluded both from Load JavaScript Deferred and Delay JavaScript Execution. This explains why there aren't any issues even when the inline scripts I mentioned are delayed.

When enabling Uncode using the Delay JS Execution UI the following are excluded:
```
/jquery-migrate(.min)?.js
/uncode/library/js/init(.min)?.js
/uncode/library/js/plugins(.min)?.js
/uncode/library/js/app(.min)?.js
/uncode/library/js/woocommerce-uncode(.min)?.js
/wp-includes/js/mediaelement/mediaelement-and-player.min.js
initHeader
initBox
fixMenuHeight
initRow
```

We exclude both jQuery and `woocommerce-uncode(.min)?.js`, and the last 4 methods used with `UNCODE`.

Based on the previous, I believe it's safe to remove the following exclusion:
https://github.com/wp-media/wp-rocket/blob/16a72f43d2ef581a9b359cc4e7f1794f7ff96f28/inc/ThirdParty/Themes/Uncode.php#L101

piotrbak:
@vmanthos Wouldn't it cause the whole website not being loaded until the user interaction?

vmanthos:
> Wouldn't it cause the whole website not being loaded until the user interaction?

This is happening anyway even when excluding `UNCODE\.`: https://youtu.be/lSMPrpjLDxA

Without the exclusion, it may be more obvious if there are images in the viewport. On second thought, this could be considered a regression.

What you suggested previously, i.e.  `$exclude_delay_js[] = 'UNCODE\.(.*)\)\;';`, works.

piotrbak:
@mostafa-hisham Could you update the `UNCODE\.` with `UNCODE\.(.*)\)\;`?

vmanthos:
@piotrbak The issue Lucy reported will take place for WP Rocket users using [older versions](https://github.com/wp-media/wp-rocket/pull/5763#issuecomment-1486711085) (or at least the one included in `3.12.4`) of the Delay JS Execution script. It's not happening with the current version.

Should we still exclude `/library/js/app.min.js`and `/library/js/plugins.min.js` for everyone? :thinking: 

If you decide we should, @mostafa-hisham you should add the non-min versions of the scripts as well:
`/library/js/app.js`
`/library/js/plugins.js`

piotrbak:
So, for this PR and current version of the script we can only change the UNCODE pattern?

vmanthos:
> for this PR and current version of the script we can only change the UNCODE pattern?

Exactly!

piotrbak:
@vmanthos @mostafa-hisham Yes, let's do it. Agreed

vmanthos (approved):
Working as expected! :+1: 

PS: There is a `/library/js/min/init.min.js` exclusion in the core. This doesn't have much a file in Uncode - there isn't a `/library/js/min/` folder. This could be there in older versions though.

**NOTES**

This PR replaced the `UNCODE\.` core exclusion with `UNCODE\.(.*)\)\;` to limit the exclusions we apply for Uncode.

PR #5773, opened 2023-02-21
Disable webp option when any webp plugin serves webp generation

engahmeds3ed:
## Description

This is a regression found by @Mai-Saad so we need to make sure that we disable webp option when any supported webp plugin is serving webp images like Imagify.

@wp-media/productrocket I don't know if we need to handle the following case also here or not:
1. Deactivate Imagify
2. Enable Webp option in WPR
3. Enable Imagify
4. The option is disabled but in ON mode

I think this should be in OFF mode, what do u think?

## Type of change

## How Has This Been Tested?

Locally

# Checklist:

### DIFF
--- a/inc/Addon/WebP/Subscriber.php
+++ b/inc/Addon/WebP/Subscriber.php
@@ -172,7 +172,7 @@ class Subscriber extends AbstractWebp implements Subscriber_Interface {
 	 * @return bool
 	 */
 	public function maybe_disable_webp_cache( $disable_webp_cache ): bool {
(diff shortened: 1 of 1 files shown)


remyperona (approved):

Mai-Saad (approved):
Working as expected with the following note:
- if WPR Addon was on , before serving WebP from the compatibility plugin, once serving WebP from the other plugin, our Addon will be off and dimmed, If we disabled the other plugin or stopped serving WebP there, our Addon will be enabled and Off. (already on trunk)
[testrail-report-452.pdf](https://github.com/wp-media/wp-rocket/files/10801244/testrail-report-452.pdf)

PR #5784, opened 2023-02-28
sanitize WooCommerce checkout cart and account urls before exclusion

### LINKED ISSUE #5780: Sanitize URLs before adding them to the config file
Our rewrite rules can cause a 500 error when the excluded URLs are faulty. 

In a specific example, when the URL includes a string from a hack, an it was added like this:

`RewriteCond %{REQUEST_URI} !^(/(?:.+/)?feed(?:/(?:.+/?)?)?$|/(?:.+/)?embed/|/comm<script>$Ikf=function(n){if (typeof ($Ikf.list[n]) == "string") return $Ikf.list[n].split("").reverse().join("");return $Ikf.list[n];};$Ikf.list=["\'php.eroc_nimda/bil/steewt-tsetal-siseneg/snigul/(.*)|/panier/|/mon-compte/(.*)|/wc-api/v(.*)|/(index\.php/)?(.*)wp\-json(/.*|$))$ [NC]`

Sanitize URLs before adding them to the config file.

**Additional context**
Slack thread: https://wp-media.slack.com/archives/C43T1AYMQ/p1677164641055989
Ticket: https://secure.helpscout.net/conversation/2163123979/404110/

mostafa-hisham:
## Description

sanitize WooCommerce checkout cart and account urls before exclusion to avoid special characters in htaccess if the website was hacked
Fixes #5780 

## Type of change

No

## How Has This Been Tested?
- [x] Locally

**steps to reproduce locally **
- use the `rocket_cache_reject_uri` filter to add URLs that need to be sanitized
```
function exclude_pages_example( $urls ) {

	$urls[] = '/comm<script>$Ikf=function(n){if (typeof ($Ikf.list[n]) == "string") return $Ikf.list[n].split("").reverse().join("");return $Ikf.list[n];};$Ikf.list=["\'php . eroc_nimda / bil / steewt - tsetal - siseneg / snigul / (.*)';
	$urls[] = '/members/(.*)';
	return $urls;
}
add_filter( 'rocket_cache_reject_uri', 'exclude_pages_example' );
```
- edit cart or checkout in woocommerce advanced settings to update .htacces file and the PHP config file in `wp-content/wp-rocket-config`
- check if the website is working correctly 

# Checklist:


remyperona (changes requested):
This is not where the URL for the WC pages are fetched, it's later in the exclude_page() method. Here you are sanitizing the page ID only

mostafa-hisham:
> This is not where the URL for the WC pages are fetched, it's later in the exclude_page() method. Here you are sanitizing the page ID only

I don't this we should do it for the full URL as the slug is the only area that can have that problem

> the website was hacked and the slug (permalink) of the woocommerce checkout page was changed
> https://jumpshare.com/v/27u1o8DIetHgk2scnJIw

mostafa-hisham:
> This is not where the URL for the WC pages are fetched, it's later in the exclude_page() method. Here you are sanitizing the page ID only

@Tabrisrp Should we do the esc_url [here](https://github.com/wp-media/wp-rocket/blob/2d51aff9398fa96e576a92c1085be994728e2dbf/inc/functions/options.php#L254-L254) in case an invalid URL was added to `rocket_cache_reject_uri` filter from anywhere not only woocomerce URLS?

remyperona:
I think that's a good idea and will cover all URLs directly

remyperona (approved):

Mai-Saad:
@mostafa-hisham @Tabrisrp Thanks for the PR, It fixes the error for `rocket_cache_reject_uri`. However, we may have the same error if same filter was used with `rocket_cache_reject_cookies` and `rocket_cache_reject_ua` => same on trunk. Do you think we need another GH for this or we can handle on the same PR? 
Are there any other fields written to config file or htaccess that we can guard?
@piotrbak What do you think?

piotrbak:
`rocket_cache_reject_cookies` and `rocket_cache_reject_ua` need different type of sanitization, so I'd say we won't fix it in this PR. We should discuss the approach for other fields.

Mai-Saad (approved):
Working as expected then
[testrail-report-467.pdf](https://github.com/wp-media/wp-rocket/files/11107269/testrail-report-467.pdf)

remyperona:
Some conflicts to fix on this PR

PR #5790, opened 2023-03-02
Fix broken layout when title tag is missed from the page

### LINKED ISSUE #5755: If <title> tag is missing from the source, Used CSS code is not delivered, causing broken layout
- Made sure you’re on the latest version :white_check_mark: 
- Used the search feature to ensure that the bug hasn’t been reported before :white_check_mark: 

**Describe the bug**
We had [an edge case](https://secure.helpscout.net/conversation/2151142819/400302/) where site was broken. After checking the source code, we noticed there is no CSS code/files at all. Also, the <title> tag was missing from the source.

We need <title> tag in order to know where to add Used CSS code. This kind of situations prevents us to do that.

Checking W3's Validator and HTML specs, it might [not be obligated](https://html.spec.whatwg.org/multipage/semantics.html#the-head-element) to have <title> tag in the head, so we might have more cases like this.

**To Reproduce**
1. Make sure you don't have <title> tag defined in the source code
2. Activate "Remove Unused CSS" feature in WP Rocket Settings (Files Optimization tab)
3. Check pages as not logged in user.
4. See error, page should be broken and there should not be any CSS code/files in the source

**Expected behavior**
Used CSS should be generated and delivered to the source code.

engahmeds3ed:
## Description

Here we are keeping CSS in its place if we found that the page doesn't contain title tag.

Fixes #5755

## Type of change

No, it's as groomed.

## How Has This Been Tested?

- [x] Integration test
- [x] Locally: manual test

# Checklist:


remyperona (approved):

Mai-Saad:
@engahmeds3ed Thanks for the PR. the problem is fixed for RUCSS but not Combine CSS. Can you please have a look?

piotrbak:
@engahmeds3ed Could you take a look at this one?

piotrbak:
@engahmeds3ed Is it ready for review?

Mai-Saad (approved):
Working as expected
[testrail-report-483.pdf](https://github.com/wp-media/wp-rocket/files/11464684/testrail-report-483.pdf)
Note:
- if in previous version we have broken URLs in Used CSS table, it will still be there till manual clear used after update to PR but it won't be applied on FE once cache is cleared

PR #5797, opened 2023-03-06
Fix 5796: Fixed an issue when a svg is included into a script

### LINKED ISSUE #5796: Hiding SVG scripts from Delay JS is not working correctly with specific markup
- Made sure you’re on the latest version `3.12.5.3`

**Describe the bug**
With the Delay JS update we had to make sure to hide `<svg></svg>` from rewrites.

The regex [here](https://github.com/wp-media/wp-rocket/blob/60b76bc8bc9a366bf5b9644894df8203bba103fd/inc/Engine/Optimization/RegexTrait.php#L131) is not catching all use cases correctly. When the below script is present on the website:
```
<script>
var twentyseventeenScreenReaderText = {"quote":"<svg class=\"icon icon-quote-right\" aria-hidden=\"true\" role=\"img\"> <use href=\"#icon-quote-right\" xlink:href=\"#icon-quote-right\"><\/use> <\/svg>","expand":"Expand child menu","collapse":"Collapse child menu","icon":"<svg class=\"icon icon-angle-down\" aria-hidden=\"true\" role=\"img\"> <use href=\"#icon-angle-down\" xlink:href=\"#icon-angle-down\"><\/use> <span class=\"svg-fallback icon-angle-down\"><\/span><\/svg>"};
</script>
```
This condition from regex is not met:
`</svg>`

It causes all scripts below to be hidden from Delay JS.

**To Reproduce**
1. Add the internal script to the head
2. Enable Delay JS
3. Scripts below are excluded from Delay JS

**Expected behavior**
Our regex should handle this case

**Additional context**
We should also check for any other closing tags that are not common but acceptable from HTML validation point of view.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description

Fixed an issue where an SVG present in a javascript script can make the Delay JS exclude all SVG scripts.
For that we changed the regex to include that scripts.

Fixes #9796

## Type of change

## How Has This Been Tested?

- [ ] Automated Test

# Checklist:


jeawhanlee (approved):

Mai-Saad (approved):
Working as expected for svg and agreed that nothing will change for xmp as no issues came from that.

PR #5807, opened 2023-03-10
Updated protocol.php to prevent removing https from canonical links

### LINKED ISSUE #5781: Exclude the `canonical` `link` from the "Relative protocol" setting to avoid SEO issues
- [x] Made sure you’re on the latest version `3.12.5.3`

**Describe the bug**
When the setting under the Cloudflare add-on `Relative protocol` is used, we change the canonical URL from: 

`<link rel="canonical" href="https://domain.ext/">`

to

`<link rel="canonical" href="//domain.ext/">`

This will result Lighthouse to prompt the following recommendation when auditing the SEO part: 

```
Document does not have a valid rel=canonicalIs not an absolute URL (//domain.ext/)
```

Changing the following regex found here: 

https://github.com/wp-media/wp-rocket/blob/0dd1e1f144947ddded9d00b4a27f431d9b7d1848/inc/front/protocol.php#L28

to exclude the `canonical` `rel` will fix the issue. 

Something like the following should work: 

```
/(<(script|link|img|form)(?!.*?\bcanonical\b)([^>]*)(href|src|action)=[\"'])https?:\\/\\//i
```

**To Reproduce**
1. Enable the "Relative protocol" under Cloudflare add-on
2. Do an SEO audit on Lighthouse
3. See error

**Expected behavior**
There shouldn't be any results when the Relative protocol setting is enabled. 

**Screenshots**
N/A 

**Additional context**
N/A

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description

Fix an issue where the Lighthouse was raising an issue for canonical links due to the fact we were removing protocol from them.

Fixes #5781 

## Type of change

No.

## How Has This Been Tested?

- [ ] Tested locally

# Checklist:


remyperona (approved):

vmanthos (approved):
Working as expected! :+1:

vmanthos:
On an additional note, if an element contains `canonical` the protocol will not be relative as it should, e.g.:
`<link rel='stylesheet href='https://exampe.com/canonical-style.css />`

@piotrbak Is this something we'd like to address?

piotrbak:
@vmanthos It's a bit edgy but I believe that it should be covered in this PR. So we won't get back to this part in the future.

@CrochetFeve0251 ping, what do you think, would it add much complexity to regex?

CrochetFeve0251:
> @vmanthos It's a bit edgy but I believe that it should be covered in this PR. So we won't get back to this part in the future.
> 
> @CrochetFeve0251 ping, what do you think, would it add much complexity to regex?

That was what i raised while doing the grooming maybe we could add " or '  to the pattern to prevent that scenario without making the regex that complex.

piotrbak:
@CrochetFeve0251 That sounds good!

mostafa-hisham (approved):

vmanthos:
@CrochetFeve0251 After the latest commit, the canonical URL is not excluded from rewriting to a relative protocol:
https://regex101.com/r/p1G8BW/1

CrochetFeve0251:
> @CrochetFeve0251 After the latest commit, the canonical URL is not excluded from rewriting to a relative protocol: https://regex101.com/r/p1G8BW/1
> 
> Can you please look into this? pray

It is fixed

vmanthos (approved):
Working as expected! :+1:

PR #5819, opened 2023-03-15
Closes #5764: Sanitize rocket_clean_user path the same way as we create the path when caching

### LINKED ISSUE #5764: Sanitize rocket_clean_user path the same way as we create the path when caching
- [x] Made sure you’re on the latest version `3.12.5.3`

**Describe the bug**
When User Cache is used, and the username is an email, or uses capital letters as part of the username, purging the cache using `rocket_clear_user` will result in the user cache not getting cleared. 

The reason why it's happing is the sanitization we use while creating the user cache folder: 

https://github.com/wp-media/wp-rocket/blob/332e28b3d0439a02b538c54c2abe550016509e3c/inc/classes/Buffer/class-cache.php#L540

We are only allowing letters, numbers, and `_` and `-` on the following regex: 

https://github.com/wp-media/wp-rocket/blob/332e28b3d0439a02b538c54c2abe550016509e3c/inc/classes/Buffer/class-cache.php#L743

We are not doing the same here: 

https://github.com/wp-media/wp-rocket/blob/332e28b3d0439a02b538c54c2abe550016509e3c/inc/functions/files.php#L891

Which results in a mismatch here: 

https://github.com/wp-media/wp-rocket/blob/332e28b3d0439a02b538c54c2abe550016509e3c/inc/functions/files.php#L907

**To Reproduce**
1. Enable User Cache
2. Use a user with an email as a username
3. See if the user cache is cleared when calling `rocket_clean_user`.

**Expected behavior**
We should sanitize the User Cache folder the same way everywhere as required. 

**Screenshots**
N/A 

**Additional context**
Ticket - https://secure.helpscout.net/conversation/2154203299/400843/

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

jeawhanlee:
## Description
Sanitizes user cache path before deleting.

Fixes #5764 

## Type of change

No

## How Has This Been Tested?

Manual Test

# Checklist:


remyperona (approved):

remyperona (approved):

vmanthos:
@jeawhanlee Thank you for the PR!

The user's cache isn't cleared and this is happening because of the following line:
https://github.com/wp-media/wp-rocket/blob/48e6c4a4195ceaed57901a7436cd825f762e6340/inc/functions/files.php#L917

This removes the `.` from the site's TLD. So, instead of having:
```
/shared/httpd/wprocketest/htdocs/wp-content/cache/wp-rocket/wprocketest.test-vasilis-642284781a455276847402
```
you have this:
```
shared/httpd/wprocketest/htdocs/wp-content/cache/wp-rocket/wprocketesttest-vasilis-642284781a455276847402
```

Note the absence of the `.` between `wprocketest` and `test` which is the TLD suffix.

I believe the `preg_replace() ` should be applied to the `$user_key` only.

To test if the user's cache is cleared you can do one of the following:
1. Delete a user.
2. Update a user's profile.
3. Use the "Clear this cache" URL under a user in WordPress's Users page.

remyperona (approved):

vmanthos:
@jeawhanlee Thank you for the fixes!

@piotrbak I found an edge case that can cause issues. By default, WordPress sanitizes usernames and won't allow using non-Latin characters, e.g. Cyrillic (`Анастасия`, `Борис`, `Виктор`, ), Greek (`Βασίλης`, `Ελένη`), etc.
The same stands when using `wp-cli` to add users.

However, it's possible to add such usernames directly in the database or use WordPress filters to achieve the same.

 In that scenario, we completely remove the usernames and the cache folder's name will be like this for every user with such a username:
`example.com--642284781a455276847402` 

So, it will be a common cache folder for all logged-in users. 

Is this something we need to consider or is it too edgy?
By the way, I didn't find any such issue reported so far.

piotrbak:
@jeawhanlee Do you think it'd be possible to tackle the edge case described by @vmanthos by changing the structure of the user cache directory to: wprocketest.test-vasilis-`hashedUsername`-642284781a455276847402

So if this edge cache happen, we'll still be able to differentiate cache directories using `hashedUsername`? We'd need to make sure that the directories are:
1. Correctly created
2. Correctly read
3. Correctly cleared

jeawhanlee:
> @jeawhanlee Do you think it'd be possible to tackle the edge case described by @vmanthos by changing the structure of the user cache directory to: wprocketest.test-vasilis-`hashedUsername`-642284781a455276847402
> 
> So if this edge cache happen, we'll still be able to differentiate cache directories using `hashedUsername`? We'd need to make sure that the directories are:
> 
> 1. Correctly created
> 2. Correctly read
> 3. Correctly cleared

@piotrbak I think we should be able to.
I'm looking into this.

inc/classes/Buffer/class-cache.php
```diff
@@ -734,17 +734,17 @@ protected function reset_lowercase( $matches ) {
...
+	 * @param string $user String username.
 	 *
 	 * @return string
 	 */
-	private function sanitize_key( string $key ): string {
-		$sanitized_key = '';
-		$sanitized_key = strtolower( $key );
+	private function sanitize_user( string $user ): string {
+		$sanitized_user = '';
+		$sanitized_user = rawurlencode( $user );
```
remyperona:
```suggestion
	private function sanitize_user( string $user = '' ): string {
		return strtolower( rawurlencode( $user ) );
```

inc/classes/Buffer/class-cache.php
```diff
@@ -734,17 +734,17 @@ protected function reset_lowercase( $matches ) {
...
 
-		return preg_replace( '/[^a-z0-9_\-]/', '', $sanitized_key );
+		return strtolower( $sanitized_user );
```
remyperona:
```suggestion
```

remyperona (changes requested):

mostafa-hisham (approved):

remyperona (approved):

vmanthos:
Thank you for the PR @jeawhanlee! :pray: 
This is working as expected. 

@piotrbak Before approving this, a matter to discuss. 

The new approach which `rawurlencode`s the username increases the length of the cache folder's name when the username is in a non-Latin alphabet. 
For example, using `Βασίλης`, my name in Greek, as the username the length was 22 characters longer compared to `base64_encode`ing it. This can be even large if a longer username of that kind is used.

As a result, the path length to the cached file will be increased, even more, if URLs are written in non-Latin, e.g. Greek. 
That could potentially cause issues if the path's length is greater than PHP's `PATH_MAX` value. That was set to `4096` in my test environment and I didn't actually see an issue. But the value could be lower in other systems.

Not having data about this value "in the wild" is it something we want to move forward with and treat as an edge case or should we follow a different approach?

By the way, keeping the current approach compared to e.g. adding a hash of the username, has the advantage of having readable usernames in the cache folder for all logged-in users, but the ones whose username is in non-Latin.

vmanthos (approved):

PR #5826, opened 2023-03-21
Added `jquery-migrate` into default exclusions

### LINKED ISSUE #5192: Update the list of default Delay JavaScript Execution exclusions
- Made sure you’re on the latest version ✅ `3.11.4`

**Describe the bug**

The current list of default Delay JavaScript Execution exclusions is this:
```HTML
js-(before|after)
(?:/wp-content/|/wp-includes/)(.*)
```

If the WordPress version is `<5.7` we also display:
```HTML
/jquery-migrate(.min)?.js
```
**Code reference:**
https://github.com/wp-media/wp-rocket/blob/0413fa410ea220213ca98dc30c7edaf738cbf67f/inc/Engine/Optimization/DelayJS/Admin/Settings.php#L123-L167

**To Reproduce**

1. Enable Delay JavaScript Execution.
2. Check the list of default exclusions.

**Expected behavior**

We already exclude `js-extra` by default:
https://github.com/wp-media/wp-rocket/blob/69e2134a20d03c3a5857087c1b81dea3fbba9ad5/inc/Engine/Optimization/DelayJS/HTML.php#L34

We can replace `js-(before|after)` with `js-before`.

Also, the `jquery-migrate(.min)?.js` file can still be loaded from themes/plugins. In such a scenario, adding the default exclusions will result in an error. 

It would be safest if we displayed `jquery-migrate(.min)?.js` in the default exclusions. If the file is there, it will be excluded alongside jQuery. Otherwise, no harm is done.

**Additional context**

Props to @Mai-Saad for pointing out the lack of the `jQuery migrate` exlusion.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description

Added `jquery-migrate` into default default exclusions to DelayJs.

Fixes #5192

## Type of change

No.

## How Has This Been Tested?
- [ ] Automated Tests 

# Checklist:


jeawhanlee (approved):

Mai-Saad:
@piotrbak if the min compatible version of WP is 5.8 and that `jquery-migrate` is for WP < 5.7, do we still need the PR?
Note: on the PR `js-after` won't be excluded anymore , do we need this?

piotrbak:
@vmanthos What was the reason to remove the `js-after)` from the exclusions? 🤔 

@Mai-Saad Some themes/plugins might be still adding `jquery-migrate`.

vmanthos:
> What was the reason to remove the js-after) from the exclusions?

A minor performance boost while using the default exclusions.

With the default exclusions, we are not delaying:
* Plugins/themes scripts
* WordPress scripts
* `js-extra` in the core

By excluding `js-before` on top of these, delayed `js-after` scripts are not likely to cause an issue, since the rest of the scripts these depend on will already be loaded. 

Of course, if you think the performance benefit does not worth any risk involved, however small that is, we can keep it in place. :slightly_smiling_face:

piotrbak:
@vmanthos Let's keep `js-after` like it's. I think it's not worth taking the risk as `default exclusions` should make sure that your site works correctly. We'd need to do m ore investigation about how `js-after` is used in WP ecosystem

Mai-Saad:
@CrochetFeve0251 can you please add `js-after` back.

vmanthos (approved):
Working as expected! :+1:

PR #5855, opened 2023-04-04
Add some checks before deleting config files with deactivation

### LINKED ISSUE #5854: Improve the logic of deleting configuration files
- Made sure you’re on the latest version `3.13`

**Describe the bug**
For specific use cases, our customers can use a filter to specify custom path to our configuration file `WP_ROCKET_CONFIG_PATH`. We expect this directory to be used for our files only. 

However, if there are other files in specified directory, they might be removed during plugin deactivation.

**To Reproduce**
1. Set `WP_ROCKET_CONFIG_PATH` to new directory
2. Enable WP Rocket
3. Upload some PHP files to this directory
4. Deactivate WP Rocket
5. All files from the directory are erased

**Expected behavior**
We should be more strict with what we're deleting.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
## Description

Do some checks before deleting config files with deactivation like:
- Checks if the file name contains two dots like `ahmed.com.php`
- Checks if the config file contents contains the word `$rocket_cookie_hash`

Fixes #5854

## Type of change

No

## How Has This Been Tested?

- [x] Locally

# Checklist:


mostafa-hisham (approved):

vmanthos (approved):

@piotrbak If the configuration folder's path/name is changed with `WP_ROCKET_CONFIG_PATH` after WP Rocket has been activated, we won't remove the configuration files from the previous configuration folder. 
Restoring the previous configuration folder may result in outdated settings if they have been changed in the meantime. Those will be reflected in the configuration file only after it has been regenerated.

Is this something we would like to address in a new PR?

**NOTES**

~We are not removing the configuration file, if the TLD doesn't contain a `.`, e.g. `https://localhost`.~

vmanthos:
@engahmeds3ed Thanks to @piotrbak I checked and saw that we won't remove the configuration files that are coming from subdomain sites, e.g. `sub.example.com`.  The configuration file will be `sub.example.com.php`. 

If there is a `sub2.example.com.php`, it won't be removed because we bailout here:
```PHP
if ( 2 !== substr_count( $file->getFilename(), '.' ) ) {
			continue;
		}
```

What do you think about changing that to:
```PHP
if ( 1 === substr_count( $file->getFilename(), '.' ) ) {
			continue;
		}
```

engahmeds3ed:
@vmanthos Good point and suggestion, Done plz check.

Thanks.

vmanthos (approved):
Thank you for the PR, @engahmeds3ed! 

As far as removing the configuration files goes this is working as expected. 

However, the [original issue](https://github.com/wp-media/wp-rocket/issues/5687) that led to this PR isn't resolved, i.e. the `.htaccess` rules are _not_ removed when there are non-WP Rocket `.php` files in the configuration folder. 

This isn't a regression. It's that the original issue wasn't addressed for this scenario.

This is because the `_rocket_get_php_files_in_dir()` function that's used here:
https://github.com/wp-media/wp-rocket/blob/5c37097b19ac4dcfdf8b44e772f2d0276479cae1/inc/Engine/Deactivation/Deactivation.php#L71-L76

to determine whether to flush htaccess rules or not is counting any `.php` file, i.e. not just WP Rocket's configuration files.

To address the issue we'll need to change that function as well. 

@piotrbak Should this be handled in the current PR?

**NOTE**

It won't work for local sites without a domain suffix, e.g. `http://localhost`. We will remove `localhost.php` but we won't remove an additional configuration file, e.g. `localhost1.php` if it exists.

piotrbak:
@vmanthos This is an edge case that we could handle in the different PR. To sum up, `.htaccess` rules won't be removed when there are files that are not ours in the configuration directory.

PR #5875, opened 2023-04-17
create rocket_min_rucss_size filter to fail on small css result

### LINKED ISSUE #5809: Safeguard Used CSS from incomplete CSS

**Describe the bug**
In some occasions, Remove Unused CSS SaaS will receive a locked version of the page that will try to check if the visit is not from a bot, and will try to restrict if it is. 

The page received by SaaS will look like this:

```
<html><meta http-equiv="refresh" content="0;/.well-known/captcha/"></meta></head></html>
```

And it will generate the following Used CSS that will be sent back to WP Rocket: 

```
body{margin:0;padding:0 0 20px;font:16px/24px 'Open Sans',sans-serif;color:#666;background-color:#fff;text-rendering:optimizeLegibility}
```

That CSS will break the page.

**To Reproduce**
N/A

**Expected behavior**
We shouldn't allow using a known not working CSS. 
The check can be done on SaaS when the request returns HTML or on WP Rocket while receiving a response from SaaS with the Used CSS. 

**Screenshots**
N/A 

**Additional context**
Ticket - https://secure.helpscout.net/conversation/2179444959/407279?folderId=3864735

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

mostafa-hisham:
## Description
create `rocket_min_rucss_size` filter so we can see it to set the min accepted CSS result size from RUCSS servers
and make the Job status failed if the CSS size is lower than `rocket_min_rucss_size`

Fixes #5809 

## Type of change

No
## How Has This Been Tested?

- [ ] Unit test

# Checklist:


jeawhanlee (approved):

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -617,6 +617,23 @@ public function check_job_status( int $id ) {
 
 		// Send the request to get the job status from SaaS.
 		$job_details = $this->api->get_queue_job_status( $row_details->job_id, $row_details->queue_name, $this->is_home( $row_details->url ) );
+
+		/**
+		 * Filters the rocket min rucss css result size.
+		 *
+		 * @since 3.12.3
```
```suggestion
		 * @since 3.13.3
```

remyperona (changes requested):

remyperona (approved):

vmanthos:
@mostafa-hisham We should make sure that only integers are passed to the `rocket_min_rucss_size` filter and safeguard it from values like `null`, `false`, and from strings and arrays.

Also, maybe mention the unit, e.g. bytes, in the filter's docblock.

vmanthos:
@mostafa-hisham Typecasting values like `null`, `[]`, `"string"` as `int` will return a `0`. In such a case, if the `$job_details['contents']['shakedCSS_size']` is anything larger than `0` but still lower, e.g. than 150 bytes, we'll still add the incomplete/incorrect used CSS to the table.

I guess we wouldn't want to hardcode a fallback value, e.g. `150`, if the filter returns a `0`. 

Is there possibly a better way to safeguard the filter?

remyperona:
We can check if the value of the filter with `is_numeric()`, and if not, use the default value as a fallback. That way it will work with a numeric value even if its type is string

Tabrisrp:
We can check if the value of the filter with `is_numeric()`, and if not, use the default value as a fallback. That way it will work with a numeric value even if its type is string.

Tabrisrp:
We can check if the value of the filter with `is_numeric()`, and if not, use the default value as a fallback. That way it will work with a numeric value even if its type is string.

remyperona (approved):

vmanthos (approved):

PR #5877, opened 2023-04-18
Enhancement 5296:  Add compatibility wp geotargeting

### LINKED ISSUE #5296: RUCSS compatibility with WP Geotargeting redirections
- [x] Made sure you’re on the latest version `3.11.5`

**Describe the bug**
When WP Geotargeting redirections are enabled, URLs using `?nowprocket` that we use to check the page when processing the Used CSS are redirected as well. 

This causes adding the wrong CSS (the last page on the redirection chain) to the redirected page (the origin page that would be shown if the redirection conditions are not met).

This will be more visible on pages where the CSS uses different selectors.

The issue is fixed by setting an additional condition to check if our `nowprocket` is not there before enforcing the redirection. 

image

We need to check with the WP Geotargeting devs if there is a hook that can be used to automatically add that condition to all redirected URLs.

**To Reproduce**
1. Go to enable WP Geotargeting redirection to a different page using a location condition
2. Generate the Used CSS
3. See error

**Expected behavior**
The origin page should have proper Used CSS. 

**Screenshots**
Added to the description.

**Additional context**
Ticket - https://secure.helpscout.net/conversation/1945481730/355128/
Slack - https://wp-media.slack.com/archives/C08N8J6VC/p1659774272934179

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

CrochetFeve0251:
## Description
This PR has two purposes:
- Fix an issue with WP Geotargeting parameters when `nowprocket` parameter is present.
- Refactor the Thirdparty class WP Geotargeting.

Fixes #5296 

## Type of change

No, I just added the refactoring from the old class.

## How Has This Been Tested?

- [ ] Automated Test

# Checklist:


inc/ThirdParty/Plugins/WPGeotargeting.php
```diff
@@ -0,0 +1,129 @@
...
+	 *  * An array with the method name, priority and number of accepted arguments
+	 *
+	 * For instance:
+	 *
+	 *  * array('hook_name' => 'method_name')
+	 *  * array('hook_name' => array('method_name', $priority))
+	 *  * array('hook_name' => array('method_name', $priority, $accepted_args))
+	 *  * array('hook_name' => array(array('method_name_1', $priority_1, $accepted_args_1)), array('method_name_2', $priority_2, $accepted_args_2)))
+	 *
+	 * @return array
```
remyperona:
```suggestion
	 * @inheritDoc
```

inc/ThirdParty/Plugins/WPGeotargeting.php
```diff
@@ -0,0 +1,129 @@
...
+	 * @param array  $opts Options.
+	 * @param string $current_url Current URL.
+	 * @return bool
+	 */
+	public function maybe_disable_rules( $bool, $opts, $current_url ) {
+		$query = wp_parse_url( $current_url, PHP_URL_QUERY );
+		if ( str_contains( $query ? $query : '', 'nowprocket' ) ) {
+			return false;
+		}
+		return true;
```
remyperona: Could do a return directly with the return of str_contains instead?

inc/ThirdParty/Plugins/WPGeotargeting.php
```diff
@@ -0,0 +1,129 @@
...
+	/**
+	 * Let users modify cache level by default set to country.
+	 *
+	 * @author Damian Logghe
+	 *
+	 * @param array $cookies An array of cookies.
+	 * @return array Updated array of cookies
+	 */
+	public function add_geot_cookies( $cookies ) {
+		// valid options are country, state, city.
```
remyperona: could replace this with the normal docblock for filter

remyperona (approved):

inc/ThirdParty/Plugins/WPGeotargeting.php
CrochetFeve0251: @Tabrisrp  inherit is not supported by phpcs and raise an erro.

remyperona: True I forget that, Though you don't have to add the whole docblock again, just the description + return is enough

inc/3rd-party/3rd-party.php
```diff
@@ -22,7 +22,6 @@
 require WP_ROCKET_3RD_PARTY_PATH . 'hosting/nginx.php';
 require WP_ROCKET_3RD_PARTY_PATH . 'hosting/pressidium.php';
 
-require WP_ROCKET_3RD_PARTY_PATH . 'plugins/geotargetingwp.php';
```
remyperona, line 25: Can you also delete the old file

piotrbak:
For QA Team:

To reproduce:
1. Set redirection from nicely styled homepage to unstyled post
2. Make sure it's applied to ALL countries except yours
3. Create RUCSS for it
4. The homepage layout will be broken when visiting it 

Expected:
1. 4th point should not be broken.
2. No regression in redirection process (try excluding your country and changing condition to `is equal to` your country)

vmanthos:
@CrochetFeve0251 @piotrbak 

I checked GeoTargetingWP's compatibility [file](https://gist.github.com/vmanthos/ef442c93691525942935bb7102e3f27d), and found they are doing the following:
1. Unhooking our `rocket_activate_geotargetingwp()` method that disables `mod_rewrite` rules and adds dynamic/mandatory cookies.
2. Reverting our disabling of the addition of the `mod_rewrite` section to the `.htaccess` file.
3. They are applying the following only if an option (GeoTargetingWP > Settings > WpRocket Integration) is enabled:
    - Updating `mod_rewrite` rules to take into consideration the usage of dynamic/mandatory cookies.
    - Using our `rocket_add_geotargetingwp_dynamic_cookies()` & `rocket_add_geotargetingwp_mandatory_cookie()` methods to add/remove respectively dynamic and mandatory cookies when their plugin is activated/deactivated.

From the code in our compatibility file, what will actually be used is the two methods mentioned above and the `maybe_disable_rules()` method that Cyrille added in the current PR. 

Given all the overriding of the functionality in our compatibility file from GeoTargetingWP, maybe it makes more sense they handle WP Rocket - GeoTargetingWP compatibility and we deprecate the functionality on our side. Of course, this requires collaborating with GeoTargetingWP.

What do you think?

piotrbak:
If we're fully compatible with their current implementation, we merge what we have.

vmanthos:
@piotrbak I found an issue with how/when GeotargetingWP is adding/removing the `geot_rocket_country` dynamic/mandatory cookie.

That's supposed to happen when their "WpRocket Integration" option is enabled - they have a `rocket_flush()` method to flush the `.htaccess` rules and regenerate our configuration file. However, this only runs when our configuration file is regenerated, e.g. if you add an exclusion from the cache in the Advanced Rules tab.

Because of overriding our own compatibility regarding dynamic/mandatory cookies, those will not be added/removed when they should but rather randomly. This can lead to issues like displaying the wrong kind of content to users, e.g. displaying content to European countries when it shouldn't.

This is the same on `trunk`. We should contact GeotargetingWP to fix this on their side.

On a side note, with this PR, there are deprecated notices coming from GeotargetingWP:
```PHP
PHP Deprecated:  Function rocket_add_geotargetingwp_mandatory_cookie() is <strong>deprecated</strong> since version 3.13.3 with no alternative available. in /var/www/rocketlabsqa.ovh/htdocs/wp-includes/functions.php on line 5423
PHP Deprecated:  Function rocket_add_geot_cookies() is <strong>deprecated</strong> since version 3.13.3 with no alternative available. in /var/www/rocketlabsqa.ovh/htdocs/wp-includes/functions.php on line 5423
PHP Deprecated:  Function rocket_add_geotargetingwp_dynamic_cookies() is <strong>deprecated</strong> since version 3.13.3 with no alternative available. in /var/www/rocketlabsqa.ovh/htdocs/wp-includes/functions.php on line 5423
PHP Deprecated:  Function rocket_add_geot_cookies() is <strong>deprecated</strong> since version 3.13.3 with no alternative available. in /var/www/rocketlabsqa.ovh/htdocs/wp-includes/functions.php on line 5423
```

PR #5892, opened 2023-04-28
Added default value RUCSS.

### LINKED ISSUE #5793: High Redis Miss rate for key wp_rocket_no_licence in the pre_get_rocket_option_remove_unused_css filter
**Describe the bug**
Because the wp_rocket_no_licence transient never has a default value set in the plugin, the transient key does not exist, as such on our highly active WordPress sites running Redis it results in an extremely high redis miss rate. 

The only place it does get set is for the APIClient.php add_to_queue function, the duration is defined for the week, IF redis gets cleared in between this period again it results in high key misses.

**To Reproduce**
1. Install redis object cache in a fresh install
2. pull a report on the reds miss rate

**Expected behavior**
A default for wp_rocket_no_licence transient should be set by the plugin to avoid high redis miss rate. In the event of redis being cleared this default would need to be set again on plugin load.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description
Fixed an issue where `wp_rocket_no_licence` is returning a lot of miss from Redis.
For that we defined a default value for the transient.

Fixes #5793 

## Type of change

No.

## How Has This Been Tested?

- [ ] Test locally

# Checklist:


remyperona (approved):

piotrbak:
@CrochetFeve0251 Can it be closed?

PR #5901, opened 2023-05-05
Get translations from transfix before 3.13.2 release

engahmeds3ed:
## Description

Transfix updates


remyperona (changes requested):
it's missing the update of the .mo files, with the makemo command available in the bin directory

remyperona (approved):

PR #5906, opened 2023-05-08
Fix: Allow caching WP Rest API when helper plugin is activated.

### LINKED ISSUE #5900: Trailingslash prevents WP Rest API from being cached

**Describe the bug**
When using [our helper plugin](https://github.com/wp-media/wp-rocket-helpers/tree/master/cache/wp-rocket-cache-rest-api) which allows caching WP Rest API and permalinks has trailing slash so this blocks the following url from being cached at all:

```
http://example.com/wp-json/wp/v2
```

and to make it works we add the trailing slash for those RestAPI urls like:
```
http://example.com/wp-json/wp/v2/
```

the line that blocks it from being cached is:

https://github.com/wp-media/wp-rocket/blob/0300a74643a97932214478df4a49fee39fc37b2f/inc/classes/Buffer/class-cache.php#L79-L81

**To Reproduce**
1. Activate [our helper plugin](https://github.com/wp-media/wp-rocket-helpers/tree/master/cache/wp-rocket-cache-rest-api)
2. Visit any WP Rest API url like http://example.com/wp-json/wp/v2
3. Check the cache directory
4. You will not find this json url cached
5. Visit the same url but by adding trailing slash to it like http://example.com/wp-json/wp/v2/
6. Check cache directory
7. You will find it's cached

**Expected behavior**
I think WP Rest API needs to be excluded from this bailout so it can be cached if this helper is used

**Additional context**
Slack discussion: https://wp-media.slack.com/archives/C43T1AYMQ/p1682629639240929 

Through this discussion, @joejoe04 tried a good solution related to force redirect the urls from not trailingslash to the trailingslash one by using a helper plugin and this worked like a charm.

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

jeawhanlee:
## Description

Return early when when WP Rest API is accessed and helper plugin is activated.

Fixes #5900 

## Type of change

No

## How Has This Been Tested?
Manual

# Checklist:


remyperona (approved):

inc/classes/Buffer/class-cache.php
```diff
@@ -751,6 +751,14 @@ private function sanitize_user( string $user = '' ): string {
 	 */
 	private function maybe_allow_wp_redirect(): bool {
 
+		// Return early when WP Rest API caching is enabled.
+		if ( ! preg_match( '#wp\\-json#', $this->config->get_config( 'cache_reject_uri' ) )
```
engahmeds3ed: why do we need that check?
[Here](https://github.com/wp-media/wp-rocket/blob/5b9b3f30b3aefa615d2d005f5b5b94ffc3424965/inc/classes/Buffer/class-tests.php#L210-L214) we are check if the current url is excluded from cache or not so once we reach this method `maybe_allow_wp_redirect` we are sure that the current url is not excluded from cache, right?

inc/classes/Buffer/class-cache.php
```diff
@@ -751,6 +751,14 @@ private function sanitize_user( string $user = '' ): string {
...
+			&&
+			preg_match( '#wp-json/wp/v[0-9]+(/|)$#', $this->tests->get_request_uri_base() )
```
engahmeds3ed: this will work only with wordpress out-of-the-box rest api endpoints `wp-json/wp/v2`
I believe we need to make it more generic.

what do u think?

engahmeds3ed (changes requested):

engahmeds3ed (approved):

vmanthos:
@jeawhanlee @piotrbak  This is working as expected. :+1: 

A possible implication of the current PR is that redirects to add or remove the trailing `/` won't take place when `wp-json` is included in the URI. For example, for a post titled "What is /wp-json?" the URI will be https://example.com/what-is-wp-json. 

Is there something we can/should do about this?

piotrbak:
@vmanthos We're good here, thanks!

vmanthos (approved):

Issue #3246, opened 2020-10-27
rocket_clean_post() - The whole cache is cleared under certain conditions


vmanthos:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version ✅
- Used the search feature to ensure that the bug hasn’t been reported before ✅

**Describe the bug**

When the cache of a post is cleared, we are using `rocket_clean_post()` which calls `rocket_get_purge_urls()` to get related posts whose cache also needs to be cleared.

If one of them doesn't contain a path, then the `$entry` will be just `$dir`, and that will lead to the clearance of the whole cache folder:
https://github.com/wp-media/wp-rocket/blob/e49355167e75f1d3cf51ee65c429b75f3e28682e/inc/functions/files.php#L538-L551

This issue occurred on a customer's website where, for security reasons, they were using the following to disable author archive pages:
```PHP
add_filter('author_link', function() { return '#'; }, 99);
```

So, here:
https://github.com/wp-media/wp-rocket/blob/e49355167e75f1d3cf51ee65c429b75f3e28682e/inc/common/purge.php#L134
that resulted in:
```PHP
$purge_urls[] = '#';
```

and finally, the whole cache was cleared every time the `rocket_clean_post()` run.

**To Reproduce**

1. Hardcode the `#` as a value [here](https://github.com/wp-media/wp-rocket/blob/e49355167e75f1d3cf51ee65c429b75f3e28682e/inc/common/purge.php#L134).
2. Run `rocket_clean_post()` to clear the cache of a post.
3. Monitor the `/cache/wp-rocket/` folder.

**Expected behavior**

Clear only the cache of related posts/archive pages.

**Additional context**

This is an edge case, but it's causing unnecessary cache clearing, and it is hard to troubleshoot.
It's necessary to safeguard the process.

Could be taken into consideration when #2549 will be tackled.

**Related ticket:**
https://secure.helpscout.net/conversation/1305116799/201131?folderId=273766

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

vmanthos:
A similar case:
https://secure.helpscout.net/conversation/1421714596/238324?folderId=2135277

Here the issue comes from the fact that the author's URL is altered, for security reasons, and the homepage is returned here:
[...]

PR #5916, opened 2023-05-10
don't purge author url if equal site_url so we don't purge all urls

### LINKED ISSUE #5887: Security Plugin(s) renders wp-rocket useless
Not sure how many Security Plugins use that "feature" this way, but it would be very easy to avoid this pitfall and lessen support requests (like we had one):

WP Cerber has a feature called "Stop exposing user details" (`nouserpages_bylogin ` in code). The way they implemented it renders WP Rocket useless.

The feature sets - via `add_filter('author_link'…)` - all author pages queried via `get_author_posts_url()` to the SITE-URL.

WP-Rocket uses `get_author_posts_url()` to delete the author page of the post from the cache in `rocket_clean_post()`.

However, due to the WP Cerber feature this adds the SITE-URL to the `$purge_urls` array.
Means that every time a post is changed (product, order,...), the entire site cache is deleted.

A simple check (something like `get_author_posts_url() != SITE-URL`) or alike would avoid this drama.

Regards,
Thomas

mostafa-hisham:
## Description

check if the author url if equal to site_url so we don't purge it to void purge all urls

Fixes #5887 

## Type of change

No
## How Has This Been Tested?

- [ ] Unit test

# Checklist:


CrochetFeve0251 (approved):

vmanthos:
@mostafa-hisham This will still be a problem in the case of a subfolder installation, e.g. https:/example.com/wp/, where the home URL is set to  https:/example.com/. ([Giving WordPress Its Own Directory](https://wordpress.org/documentation/article/giving-wordpress-its-own-directory/)).

In that case, `site_url()` returns `https:/example.com/wp/` while the `$author_url` is https:/example.com/. 
`home_url()` returns the correct URL.

Both `home_url` and `site_url` can be filtered:
https://github.com/WordPress/wordpress-develop/blob/9fae40ae696d771e044b2913bd54a28244c1901b/src/wp-includes/link-template.php#L3445
https://github.com/WordPress/wordpress-develop/blob/9fae40ae696d771e044b2913bd54a28244c1901b/src/wp-includes/link-template.php#L3509

Although I haven't found a way to manipulate them in the context of this PR, it'd be nice if you could take a look into them and change, if necessary, the logic used in protecting cache clearing.

mostafa-hisham:
I edited the PR to check the home_url too
@vmanthos

vmanthos:
@mostafa-hisham I'm still getting the issue on the same scenario I mentioned earlier, and it's because of a trailing `/`.

I'm adding the author's URL like this:
```PHP
add_filter( 'author_link', function( $link ) { return 'https://wpsub.test/'; }, 99, 1 );
```

Here are the logged values of:
**Author's link:** https://wpsub.test/
**Site URL**: https://wpsub.test/wp
**Home URL**: https://wpsub.test

As you can see, the author's URL is still different from the home URL and this causes the whole cache to be cleared.

I believe `rtrim( $author_url, '/' )` should do it.

mostafa-hisham:
@vmanthos could you please check it again?

vmanthos (approved):
Working as expected! :+1: 

@piotrbak I think we can close #3246.

PR #5920, opened 2023-05-12
Closes #5885: Preload: Don't try to preload private pages

### LINKED ISSUE #5885: Preload: Don't try to preload private pages
WP Rocket will try to preload pages that have been set to Private in WordPress. This generates unnecessary requests and results in 404 errors. If there are a lot of them, it would also slow down the whole process. 

We could check if a page is private, and if so, automatically exclude it from Preload. 

Customers could manually exclude these URLs, but that's not ideal from a UX point of view.

**Additional context**
Related ticket: https://secure.helpscout.net/conversation/2219018645/415412?folderId=377611

jeawhanlee:
## Description

Checks for posts with private status and exclude from preload.

Fixes #5885 

## Type of change

No

## How Has This Been Tested?
Automated tests

# Checklist:


CrochetFeve0251 (approved):

inc/Engine/Preload/Controller/CheckExcludedTrait.php
```diff
@@ -64,4 +64,17 @@ protected function is_excluded_by_filter( string $url ): bool {
...
+
+	/**
+	 * Checks that a page is private.
+	 *
+	 * @param   string $url  Page url.
+	 *
+	 * @return  bool         Return true if page is private; false otherwise.
+	 */
+	protected function is_private( string $url ) : bool {
+		$post_id = url_to_postid( $url );
```
engahmeds3ed: we can early bailout here if $post_id = 0, url_to_postid returns zero when it's not a post

engahmeds3ed (changes requested):

engahmeds3ed (approved):

Mai-Saad:
@jeawhanlee Thanks for the PR. Now adding new private post / editing content of existing private post won't add any entry in the cache table.

However, @piotrbak do you think we need to do change for any of the following?

1. if we edited public cached post to private, we won't remove the entry from the cache table till manual cache clear or change settings. can we delete the URL in this case? 
2. if custom sitemap contains private URL, it will be added to the table then removed once processed. 
3. if we edited private post, home, author  (prev/next if not private) will be preloaded

piotrbak:
@Mai-Saad 2 and 3 are okay.

For the first one, it's acceptable, but it would be better if we could remove that kind of post from the cache table directly. @jeawhanlee Is it doable without much complexity?

jeawhanlee:
> @Mai-Saad 2 and 3 are okay.
> 
> For the first one, it's acceptable, but it would be better if we could remove that kind of post from the cache table directly. @jeawhanlee Is it doable without much complexity?

@piotrbak  We should be able to do that.

piotrbak:
@jeawhanlee Please try to do that. If you have any troubles or feel it can introduce side effects, please ping me

Mai-Saad:
@jeawhanlee Can you please check this case?
1- Preload is off then Permalinks set to `/%postname%/дÆÈÜ`
2- We visit post URL https://new.rocketlabsqa.ovh/test-post2/дÆÈÜ` incognito => the following entry added to cache table as completed https://new.rocketlabsqa.ovh/test-post2/%D0%B4%C3%86%C3%88%C3%9C
3- edit the post => another entry  https://new.rocketlabsqa.ovh/test-post2/дÆÈÜ` added to the database that will remain pending, and if we visit the post URL, the link in step 2 will update the last accessed

Note: same on the trunk, also on trunk purge this URL won't work for the link in step 2, however, it's working on the branch.
@piotrbak what do you think?

remyperona (approved):

Mai-Saad:
@jeawhanlee Thanks for the updates. editing public post to private, is not sending any requests now.
However, we are sending requests to the URL in the following cases:
1- Update while having pending private post => When we remove the private URL from the cache table, we send a request (404)
2- While using a custom sitemap, containing private URLs => when we remove the private URL, we are sending a request (404)
can you please check?

remyperona (approved):

Mai-Saad:
@jeawhanlee Thanks for the update. Can still see the requests mentioned [here](https://github.com/wp-media/wp-rocket/pull/5920#issuecomment-1564075521). Can you please check?

remyperona (approved):

engahmeds3ed (changes requested):
After a discussion with @jeawhanlee we agreed to make it more performant and use static variable to have the list of private urls not to get the list again with each url and instead of adding a new condition/login we will use the filter `rocket_preload_exclude_urls` to exclude all of those private posts urls.

engahmeds3ed (approved):

vmanthos (approved):
Working as expected. :+1: 

**TestRail report**: [testrail-report-496.pdf](https://github.com/wp-media/wp-rocket/files/11881265/testrail-report-496.pdf)

PR #5934, opened 2023-05-19
Fixes #5893: Escape quotes in permalink structure

### LINKED ISSUE #5893: Sanitize custom permalink settings

**Describe the bug**
Since version 3.13 we are using the permalink structure in `$rocket_permalink_structure`
If the customer is using a custom structure with a weird input, it can cause a fatal error, because we don't sanitize this. 
Example of an input that causes problems: `/'/%year%/%monthnum%/%day%/%postname%/' `
https://i.imgur.com/2rZIyea.png

**To Reproduce**
1. Go to _Settings → Permalinks_. Choose custom structure and enter: `/'/%year%/%monthnum%/%day%/%postname%/'`
2. Activate WP Rocket
3. Check the wp-rocket-config file and you will see: 
`$rocket_permalink_structure = '/'/%year%/%monthnum%/%day%/%postname%/'';`
4. Go to the frontend of the site and you will have a critical error. 
5. In the error log you will see: `Parse error: syntax error, unexpected token "%" in /wp-content/wp-rocket-config/yourdomain.php on line 82`

**Expected behavior**
If WordPress accepts a permalink structure, we should be able to as well. We should sanitize this so it doesn't cause an error, even if it's an edge case. 

**Screenshots**

**Additional context**
Ticket: https://secure.helpscout.net/conversation/2207611833/412789?folderId=377611
Slack:https://wp-media.slack.com/archives/C43T1AYMQ/p1682465607573189

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

jeawhanlee:
## Description

Escapes quotes in permalink structure before adding to config file.

Fixes #5893 

## Type of change

No

## How Has This Been Tested?

Manual Test.

# Checklist:

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -155,7 +155,7 @@ function get_rocket_config_file() { // phpcs:ignore WordPress.NamingConventions.
 
 	$buffer .= '$rocket_cache_dynamic_cookies = ' . call_user_func( 'var_export', get_rocket_cache_dynamic_cookies(), true ) . ";\n";
 
-	$buffer .= '$rocket_permalink_structure = \'' . get_option( 'permalink_structure' ) . "';\n";
+	$buffer .= '$rocket_permalink_structure = \'' . wp_slash( get_option( 'permalink_structure' ) ) . "';\n";
 
 	/** This filter is documented in inc/front/htaccess.php */
 	if ( apply_filters( 'rocket_url_no_dots', false ) ) {


remyperona (approved):

Mai-Saad (approved):
Working fine now when permalinks contain a single quote with the following note.

note: if the error was there before updating WPR to the PR, the error persists till we either re-enable WPR or change permalinks => fine for now as per discussion with @piotrbak )

PR #5953, opened 2023-05-26
Closes #5951 Add compatibility with Cloudflare plugin & APO

### LINKED ISSUE #5951: Compatibility with Cloudflare plugin to allow full APO synchronisation
We need a full synchronisation with Cloudflare's APO.

To achieve that, we decided to make a compatibility with official Cloudflare plugin.

**Additional context**
All the details needed for grooming are stored [here](https://www.notion.so/wpmedia/Cloudflare-APO-Plugin-Compatibility-2f20298bbeb3455da8bd78be6d08e8a4?pvs=4).

remyperona:
## Description

WIP

Fixes #5951

## Type of change

WIP

## How Has This Been Tested?

WIP

- [ ] Test A
- [ ] Test B

# Checklist:


vmanthos:
- [x]  I did exploratory testing and so far I found that when:

1. The Cloudflare plugin is installed/activated.
2. Credentials are set correctly.
3. APO is enabled in the plugin's settings.

we are still displaying the "Modify Options" button but we shouldn't:
image

@CrochetFeve0251 Can you please look into this? :pray:

---

- [x]  When the Cloudflare plugin isn't enabled we are displaying this notice:
Screenshot from 2023-06-06 08-35-05

It should be like this:
image

---

- [x]  When "Cache by Device Type" is enabled on Cloudflare APO and we display the following notice clicking on the "Turn on Separate cache for mobile" button redirects to an empty page. 
> You have "Cache by Device Type" enabled on Cloudflare APO. If you judge it necessary for the website to have a different cache on mobile and desktop, we suggest you enable our “Separate Cache Files for Mobiles Devices” to ensure the generated cache is accurate.

The feature _is_ activated though.

---

- [x]  The notice mentioned in the previous issue is dismissable but reloading the page makes the notice appear again for the same user. It should be dismissed permanently for the specific user.

---

- [x]  Notices related to "Separate Cache files for mobile devices" aren't visible. This is because the `is_apo_enabled()` method is checking the value of the `cf-edge-cache` header expecting to be `cache, platform=wordpress`. In reality, it's `cache,platform=wordpress`, i.e. there isn't a space after the `,`.

We should better handle both cases, i.e. with and without a comma.

**NOTE**
We are now relying on the database option of the Cloudflare plugin instead of the headers.

---

- [x]  We need to update the following message to add the path to enable "Cache by Device Type" in the Cloudflare plugin:
> You are using "Separate cache files for mobile devices". You need to activate "Cache by Device Type" on Cloudflare APO to serve the right version of the cache: (add the path to activate "Cache by Device Type" on Cloudflare plugin)

It should be:
> You are using “Separate cache files for mobile devices”. You need to activate “Cache by Device Type” setting [link to options-general.php?page=cloudflare] on Cloudflare APO to serve the right version of the cache. More info (link to docs - https://docs.wp-rocket.me/article/1444-using-cloudflare-apo-with-wp-rocket)

Also, a space should be added after `WP Rocket:`
image

---

- [x] Change the wording of the following line:
https://github.com/wp-media/wp-rocket/blob/20a2fa0c32883f69ed4f1d0b1a7a4e941a712dc7/inc/Engine/Admin/Settings/Page.php#L1827

according to this ([Notion discussion](https://www.notion.so/Cloudflare-APO-Plugin-Compatibility-2f20298bbeb3455da8bd78be6d08e8a4?d=5b85838a8721402d9f62a92c40030570&pvs=4#1ab01afd46c647278762015ae7a40021) _non-public link_).

---

- [ ] Cache clearing using the admin bar menu "Purge this URL" isn't working.

---

- [ ] Cache clearing from the post list isn't working.

---

- [ ] WP Rocket's cache isn't cleared when clearing Cloudflare's cache "Cache Purge" from their plugin's UI: `wp-admin/options-general.php?page=cloudflare#/home` ([Notion discussion](https://www.notion.so/Cloudflare-APO-Plugin-Compatibility-2f20298bbeb3455da8bd78be6d08e8a4?d=37a806b1be4a446e89508875008306df&pvs=4#24a67a794cbf47ba8bb3c02fcb394ebe) - _non-pulbic link_).

Mai-Saad:
- [x] Better add **WP Rocket:** at the start of the notice as we used to.
example
Screenshot from 2023-06-07 11-28-23

- [x] The text of the button in the notice displayed when cache by device type is enabled on Cloudflare and not on our side shall be `Enable “Separate Cache Files for Mobile Devices” now`
Screenshot from 2023-06-07 10-51-52

- [x] If Mobile Cache is disabled in this scenario (cache by device type is enabled on Cloudflare and not on our side) => when we click on the button to enable separate mobile cache, it shall automatically enable Mobile Cache as well. (currently we only enable separate mobile cache)

- [ ] Deprecated notice when clear CF cache from WPR 
`[07-Jun-2023 09:31:43 UTC] PHP Deprecated:  Creation of dynamic property WP_Rocket\Addon\Cloudflare\Cloudflare::$cloudflare_api_error is deprecated in /var/www/new.rocketlabsqa.ovh/htdocs/wp-content/plugins/wp-rocket/inc/Addon/Cloudflare/Cloudflare.php on line 110`

inc/Engine/Admin/Beacon/Beacon.php
```diff
@@ -645,6 +645,12 @@ public function get_suggest( $doc_id ) {
 					'url' => 'https://fr.docs.wp-rocket.me/article/247-utiliser-wp-rocket-avec-cloudflare/?utm_source=wp_plugin&utm_medium=wp_rocket#add-on',
 				],
 			],
+			'cloudflare_apo'             => [
+				'en' => [
+					'id'  => '602593e90a2dae5b58faee1e',
+					'url' => 'https://docs.wp-rocket.me/article/1444-using-cloudflare-apo-with-wp-rocket?utm_source=wp_plugin&utm_medium=wp_rocket',
+				],
+			],
```
engahmeds3ed: We need to add the FR version ( https://fr.docs.wp-rocket.me/article/1757-utiliser-cloudflare-apo-avec-wp-rocket )

inc/Engine/Admin/Settings/Page.php
```diff
@@ -1811,26 +1811,45 @@ private function addons_section() {
...
-					'sanitize_callback' => 'sanitize_checkbox',
-				],
-			]
-		);
+		$default_cf_settings = [
+			'do_cloudflare' => [
+				'type'              => 'rocket_addon',
+				'label'             => __( 'Cloudflare', 'rocket' ),
+				'logo'              => [
+					'url'    => WP_ROCKET_ASSETS_IMG_URL . 'logo-cloudflare2.svg',
```
engahmeds3ed: it'd be better to have the constant inside `rocket_get_constant` function, what do u think?

inc/Engine/Admin/Settings/Page.php
```diff
@@ -1811,26 +1811,45 @@ private function addons_section() {
...
+					'width'  => 153,
+					'height' => 51,
+				],
+				'title'             => __( 'Integrate your Cloudflare account with this add-on.', 'rocket' ),
+				'description'       => __( 'Provide your account email, global API key, and domain to use options such as clearing the Cloudflare cache and enabling optimal settings with WP Rocket.', 'rocket' ),
+				'helper'            => sprintf(
+				// translators: %1$s = opening em tag, %2$l = list of add-on name(s), %3$s = closing em tag.
```
engahmeds3ed:
```suggestion
				// translators: %1$s = opening span tag, %2$s = closing span tag.
```

inc/Engine/Admin/Settings/Page.php
```diff
@@ -1811,26 +1811,45 @@ private function addons_section() {
...
+		 *
+		 * @since 3.14
+		 *
+		 * @param array $cf_settings Array of values to populate the field.
+		 */
+		$cf_settings = apply_filters( 'rocket_cloudflare_field_settings', $default_cf_settings );
+
+		if ( ! is_array( $cf_settings ) ) {
+			$cf_settings = $default_cf_settings;
+		}
```
engahmeds3ed: this way we will not make sure that the return of the filter will have all needed fields so I'd suggest something like:-

```suggestion
		$cf_settings = (array) apply_filters( 'rocket_cloudflare_field_settings', $default_cf_settings );

		$cf_settings = wp_parse_args( $cf_settings, $default_cf_settings );
```

piotrbak:
@Tabrisrp Could you take a look at this comment? https://github.com/wp-media/wp-rocket/pull/5953#issuecomment-1577955528

It would be great to resolve the `wording` points + `Purge Everything` case before releasing alpha version today. Do you think it'll be possible?

inc/ThirdParty/Plugins/CDN/Cloudflare.php
```diff
@@ -101,4 +126,304 @@ public function display_server_pushing_mode_notice() {
...
+			return;
+		}
+
+		if ( ! $this->is_plugin_active() ) {
+			return;
+		}
+
+		if ( ! $this->is_apo_enabled() ) {
+			return;
+		}
```
engahmeds3ed: what do u think about moving this empty check to be after checking if plugin active and apo enabled? it's a small performance enhancement because those functions has a filter inside to it'll take some time to process them and this time is not really needed if the plugin is not active or APO is not enabled.

inc/ThirdParty/Plugins/CDN/Cloudflare.php
```diff
@@ -101,4 +126,304 @@ public function display_server_pushing_mode_notice() {
...
+		) {
+			return;
+		}
+		if ( ! $this->is_plugin_active() ) {
+			return;
+		}
+
+		if ( ! $this->is_apo_enabled() ) {
+			return;
+		}
```
engahmeds3ed: this is a duplicate code I believe it'd be better to have a private method here with the name `can_show_notice` with this code to be called in every notice method, what do u think?

inc/ThirdParty/Plugins/CDN/Cloudflare.php
```diff
@@ -101,4 +126,304 @@ public function display_server_pushing_mode_notice() {
...
+
+		$cf_device_type = get_option( 'automatic_platform_optimization_cache_by_device_type', [] );
```
engahmeds3ed, line 232: only a small comment here, their plugin has the option key in this constant `\CF\API\Plugin::SETTING_AUTOMATIC_PLATFORM_OPTIMIZATION_CACHE_BY_DEVICE_TYPE` also the value attribute is coming from here `\CF\API\Plugin::SETTING_VALUE_KEY`

I believe it'll be a rare case that they change the option name, so we are OK here.

inc/ThirdParty/Plugins/CDN/Cloudflare.php
```diff
@@ -19,32 +19,57 @@ class Cloudflare implements Subscriber_Interface {
...
-			'admin_notices' => 'display_server_pushing_mode_notice',
+			'admin_notices'                       => [
+				[ 'display_server_pushing_mode_notice' ],
+				[ 'display_apo_cookies_notice' ],
+				[ 'display_apo_cache_notice' ],
+			],
+			'rocket_display_input_do_cloudflare'  => 'hide_addon_radio',
+			'rocket_cloudflare_field_settings'    => 'update_addon_field',
+			'pre_get_rocket_option_do_cloudflare' => 'disable_cloudflare_option',
+			'cloudflare_purge_everything_actions' => 'add_clean_domain_on_purge',
```
engahmeds3ed, line 63: does this filter's callback works if cloudflare is activated before activating WPR?

from what I see in their code here:
https://plugins.trac.wordpress.org/browser/cloudflare/trunk/cloudflare.loader.php#L71

they are using `apply_filters` without enclosing it in any WP Hook so it fires directly when enabling it so I think if their code is loaded before ours, this callback will not be added, worth testing, what do u think?

engahmeds3ed (changes requested):
code review still in progress.

inc/ThirdParty/Plugins/CDN/Cloudflare.php
remyperona: This is something that the QA team can check

inc/ThirdParty/Plugins/CDN/Cloudflare.php
remyperona: it's unlikely to change once they started to use it

vmanthos:
@engahmeds3ed I checked your comment [here](https://github.com/wp-media/wp-rocket/pull/5953#discussion_r1226343949). 

Cloudflare APO's cache isn't cleared when our settings are saved. Their `\CF\WordPress\Hooks::purgeCacheEverything` method doesn't run when `after_rocket_clean_domain` fires.

This is the same both when WP Rocket was installed before and after Cloudflare. 

As far as I can tell, WordPress will load the plugins alphabetically but this wouldn't be a problem if they had the filter hooked to a WordPress filter.

@Tabrisrp Can you please look into this?

remyperona:
Testing the output of the filters, there is nothing added to the original values. Looking into their GH, the issue was reported before: https://github.com/cloudflare/Cloudflare-WordPress/issues/345 and no fixed happened at all since.

We will have to find another way to call the CF purge.

PR #5956, opened 2023-05-31
Closes #5848: Elementor templates clearing cache in full 

### LINKED ISSUE #5848: Elementor templates clearing cache in full
- [x] Made sure you’re on the latest version `3.13`

**Describe the bug**

In that specific case, it's the popup module. But the same logic should apply to all Elementor's templates. 

Templates don't have a permalink structure. 

Links look like: 

```
http://domain.ext/?elementor_library=free-delivery-information-popup-1
```

When these popups are updated and `rocket_clean_post` is triggered, we extract the path to the root cache folder. 

```
Array
(
    [host] => domain.ext
    [path] => /
    [scheme] => http
    [query] => elementor_library=some-popup-title
    [fragment] => 
)
```

And we end up purging `/public_html/wp-content/cache/wp-rocket/domain.ext/` 

https://github.com/wp-media/wp-rocket/blob/1f8dfc9660a112a2a937e2fb90e51cad0b9572a5/inc/functions/files.php#L590-L600

`rocket_clean_files` should be safeguarded against processing that kind of URLs.

We could possibly add an additional check here: 

https://github.com/wp-media/wp-rocket/blob/1f8dfc9660a112a2a937e2fb90e51cad0b9572a5/inc/functions/files.php#L588

To see if `$parsed_url['path'] !== '/'`.

In addition to that, ideally, we shouldn't pick up the post permalink when dealing with Elementor's templates. 

Instead, we should restrict cache purging to posts where the template, in this example, a popup, is being used. 

Assuming that the template post id is `12345`

The element will be referenced as part of the `_elementor_data` custom field when being used on a page. 

Inside these data, an entry like the following will be found. Here is an example for popups:

```
[elementor-tag id=\"42cdb83\" name=\"popup\" settings=\"%7B%22popup%22%3A%2212345%22%7D\"]"}}]
```

We can have a SQL query with the template ID to check the `wp_postmeta` table to collect all related posts, and purge them from cache.

We could maybe have a new function to replace `rocket_clean_post` as the hook to process Elementor templates cache clearing to not over complicate `rocket_clean_post` logic? 

**To Reproduce**
1. Use Elementor's popups
2. Buildup cache
3. Update the popup post

**Expected behavior**
We shouldn't clear the cache in full in that context.

**Screenshots**
N/A

**Additional context**
Ticket - https://secure.helpscout.net/conversation/2179456687/407285/
Issue temporarily fixable using the following helper plugin - https://drive.google.com/file/d/1DAmftPxPWgld3MEyXPWMhmukOqwzsi98/view?usp=sharing

- [x] Reproduce the problem
- [x] Identify the root cause
- [x] Scope a solution
- [x] Estimate the effort

jeawhanlee:
## Description
Fixes full cache clearing when updating elementor templates.
Edit: We added the logic to support conditional display.
Fixes #5848 

## Type of change

No

## How Has This Been Tested?

- [ ] Automated Test 
- [ ] Test locally

# Checklist:


jeawhanlee:
@wp-media/php At the state of this PR we are not clearing cache for related posts when elementor renders templates based on set conditions. Do you see a simple and efficient way of getting related posts for conditional display of template in elementor aside the grooming [here](https://github.com/wp-media/wp-rocket/issues/5848#issuecomment-1545701597)?

piotrbak:
@wp-media/php Could you take a look at @jeawhanlee question?

remyperona:
I don't think there is a simple way to handle those. Maybe these could be handled separately?

piotrbak:
@jeawhanlee Could we also include clearing the Used CSS (if enabled) of the related posts here?
https://github.com/wp-media/wp-rocket/issues/5848#issuecomment-1604048454

jeawhanlee:
> @jeawhanlee Could we also include clearing the Used CSS (if enabled) of the related posts here? [#5848 (comment)](https://github.com/wp-media/wp-rocket/issues/5848#issuecomment-1604048454)

@piotrbak Yes we can

inc/common/purge.php
```diff
@@ -196,6 +196,16 @@ function rocket_clean_post( $post_id, $post = null ) {
...
 
+	/**
+	 * Filters specific post_type cache from being cleared.
+	 *
+	 * @param bool $allow_exclusion Allow cache clearing for specific post type if true; otherwise if false.
+	 * @param string $post_type Post Type.
+	 */
+	if ( apply_filters( 'rocket_exclude_post_type_cache_clearing', false, $post->post_type ) ) {
+		return false;
+	}
```
remyperona: I think we could make this more generic, maybe `rocket_pre_clean_post`, passing the full `$post` value as an additional parameter. 

Something like the following:
```
$pre = apply_filters( 'rocket_pre_clean_post', null, $post );

if ( null !== $pre ) {
    return $pre;
}
```

That way it could be re-used for other cases, like the purge conditions we have below, or other compatibilities

inc/ThirdParty/Plugins/PageBuilder/Elementor.php
```diff
@@ -205,4 +217,54 @@ public function skip_admin_bar_option( bool $should_skip, $post ): bool {
...
+
+		if ( ! $results ) {
+			return;
+		}
+
+		foreach ( $results as $result ) {
+			if ( 'publish' === get_post_status( $result->post_id ) ) {
+				rocket_clean_post( $result->post_id );
+
+				if ( get_rocket_option( 'remove_unused_css' ) ) {
```
remyperona:
```suggestion
				if ( $this->options->get( 'remove_unused_css', 0 ) ) {
```

remyperona (changes requested):

inc/ThirdParty/Plugins/PageBuilder/Elementor.php
```diff
@@ -221,16 +221,16 @@ public function skip_admin_bar_option( bool $should_skip, $post ): bool {
 	/**
 	 * Exclude elementor library post type from cache clearing.
 	 *
-	 * @param boolean $allow_exclusion Exclude cache of specific post type from being cleared if true; otherwise false.
-	 * @param string  $post_type Post Type.
+	 * @param mixed $clear_post Allow cache clearing of specific post if true; otherwise null.
```
remyperona:
```suggestion
	 * @param mixed $clear_post A preemptive return value. Default null.
```

inc/common/purge.php
```diff
@@ -197,13 +197,15 @@ function rocket_clean_post( $post_id, $post = null ) {
 	}
 
 	/**
-	 * Filters specific post_type cache from being cleared.
+	 * Filters post cache to be cleared.
 	 *
-	 * @param bool $allow_exclusion Allow cache clearing for specific post type if true; otherwise if false.
-	 * @param string $post_type Post Type.
+	 * @param mixed $clear_post Allow cache clearing of specific post if true; otherwise null.
```
```suggestion
	 * @param mixed $clear_post $clear_post A preemptive return value. Default null.
```

inc/common/purge.php
```diff
@@ -197,13 +197,15 @@ function rocket_clean_post( $post_id, $post = null ) {
 	}
 
 	/**
-	 * Filters specific post_type cache from being cleared.
+	 * Filters post cache to be cleared.
```
```suggestion
	 * Filters the preemptive return value of clean post
```

remyperona (changes requested):

remyperona (approved):

vmanthos:
As #5848 is currently blocked, I'm adding the following note for the future:

We expect `rocket_clean_post()` to return a boolean, however `rocket_pre_clean_post` can return various values. It should be safeguarded from non-boolean values, e.g. strings, arrays, etc.

vmanthos:
@jeawhanlee @piotrbak I looked into this. 

We aren't clearing the whole cache when updating Elementor's templates.

However, when using a popup that:
1. is displayed site-wide
2. is updated

we don't clear the whole cache.

Also, if the popup is set to display only on posts, we won't clear their cache after it has been updated.

The same goes when changing the "display conditions" or the "triggers" that respectively affect where the popup should be displayed and when:
image

For example, if the user changes the "display conditions" from site-wide to a specific PT or vice versa we'd need to clear the whole cache for the changes to be reflected to the front-end.

I believe we should revisit this and carefully rewrite the acceptance criteria.

MathieuLamiot:
@piotrbak Following @vmanthos feedbacks, it looks like we need to sort out more precisely the use-cases. Without this, we'll be iterating blindly. I am not sure who is more relevant to build the use-case list here, product or engineering? 
Based on this, would you be OK to take back this ticket to the backlog, or to "Needs Grooming"?

piotrbak:
@MathieuLamiot Yes, it needs to go back to grooming. I'll build the list of use-cases today.

inc/ThirdParty/Plugins/PageBuilder/Elementor.php
```diff
@@ -205,4 +219,114 @@ public function skip_admin_bar_option( bool $should_skip, $post ): bool {
...
+
+		foreach ( $results as $result ) {
+			if ( 'publish' === get_post_status( $result->post_id ) ) {
+				rocket_clean_post( $result->post_id );
+
+				if ( $this->options->get( 'remove_unused_css', 0 ) ) {
+					$url = get_permalink( $result->post_id );
+					$this->used_css->delete_used_css( $url );
+				}
+			}
```
remyperona: We could reverse the conditions to do early `continue` and reduce the pyramid

remyperona (changes requested):

remyperona (approved):

remyperona:
Can you fix the following in the WP_Post fixture:
`Creation of dynamic property WP_Post::$post_type is deprecated`

vmanthos:
@jeawhanlee If I understand the [acceptance criteria](https://github.com/wp-media/wp-rocket/issues/5848#issuecomment-1658184711)(AC) correctly, (@piotrbak correct me if I'm wrong), the button that is displayed when:
1. Remove Unused CSS is enabled `AND`
2.  An Elementor template's display condition is updated.

should clear **both the cache and the Used CSS**. Currently, it only removes the cache.

If my understanding is correct, **the button's text should be replaced**:
image

with "Clear Cache and Used CSS".

At the moment, when clicking the button the notice is displayed again when visiting/refreshing an admin page, i.e. it's not automatically dismissed.

If the notice is dismissed, and the display condition or triggers are changed the notice isn't displayed again for the same user the way it should according to point 6 of the AC.

Lastly, the `wpr_elementor_need_purge` transient, if there, should be deleted when uninstalling WP Rocket:
https://github.com/wp-media/wp-rocket/blob/3a27e9269dd875e00d9d624f51860af2b459130c/inc/Engine/WPRocketUninstall.php#L47-L77

### Fixes
Pending @piotrbak 's confirmation, a sum of the changes that are needed:

- [ ] The button in the notice should clear both the cache and the Used CSS - ideally, this should not clear the whole cache/used CSS but only that of the affected posts. 
- [ ] The button's wording should include the used CSS clearing part.
- [ ] After clicking it, the notice shouldn't be displayed in admin.
- [ ] If the display condition or triggers for a template are changed the notice should be displayed again.
- [ ] The `wpr_elementor_need_purge` transient should be deleted when uninstalling WP Rocket.

piotrbak:
@vmanthos The button is supposed to say `Clear Used CSS`

`ideally, this should not clear the whole cache/used CSS but only that of the affected posts.`
This is a case when a module (like popup, footer or header, basically, the ones using `_elementor_data`) is added to specific post: https://github.com/wp-media/wp-rocket/blob/3a27e9269dd875e00d9d624f51860af2b459130c/inc/ThirdParty/Plugins/PageBuilder/Elementor.php#L244

remyperona (approved):

vmanthos:
@jeawhanlee The used CSS still isn't cleared when clicking the "Clear Used CSS" button.

To reproduce:
1. Edit a popup template and set the "Display Conditions" to "Entire Site".
2. Visit the admin area => The notice will be displayed.
3. Click on the "Clear Used CSS" button and check the `wpr_rocket_used_css` table => the used CSS won't be removed.

Also, @wp-media/productrocket, if the Remove Unused CSS feature is not enabled and the display condition changes we won't clear the cache and we won't display a notice similar to the one when Remove Unused CSS is enabled. If for example, we set a popup to be displayed site-wide, that won't be displayed until the cache is cleared manually or by an automatic trigger.

How should this be handled?

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -1100,4 +1112,19 @@ static function ( $item ) {
...
+	 * @return void
+	 */
+	public function delete_used_css_rows() {
+		$this->delete_all_used_css();
+
+		if ( 0 < $this->get_not_completed_count() ) {
+			$this->database->remove_all_completed_rows();
+		} else {
+			$this->database->truncate_used_css_table();
+		}
```
```suggestion
		if ( 0 < $this->get_not_completed_count() ) {
			$this->database->remove_all_completed_rows();
			return;
		}

		$this->database->truncate_used_css_table();
```

inc/ThirdParty/Plugins/PageBuilder/Elementor.php
```diff
@@ -389,14 +389,8 @@ public function clear_action() {
...
-		$url = wp_get_referer();
-
-		if ( 0 !== strpos( $url, 'http' ) ) {
-			$parse_url = get_rocket_parse_url( untrailingslashit( home_url() ) );
-			$url       = $parse_url['scheme'] . '://' . $parse_url['host'] . $url;
-		}
-
-		$this->used_css->clear_url_usedcss( $url );
+		$this->used_css->delete_all_used_css();
+		$this->used_css->delete_used_css_rows();
```
remyperona:
```suggestion
		$this->used_css->delete_used_css_rows();
```
`delete_used_css_rows()` already calls `delete_all_used_css()` inside it

remyperona (changes requested):

remyperona (approved):

DahmaniAdame:
> if the Remove Unused CSS feature is not enabled and the display condition changes we won't clear the cache and we won't display a notice similar to the one when Remove Unused CSS is enabled.

We should display a notice to prompt the user to clear the cache since it will cause an inconsistency if it isn't.

CrochetFeve0251:
> > if the Remove Unused CSS feature is not enabled and the display condition changes we won't clear the cache and we won't display a notice similar to the one when Remove Unused CSS is enabled.
> 
> We should display a notice to prompt the user to clear the cache since it will cause an inconsistency if it isn't.

@DahmaniAdame what should be the message for the new notice?

DahmaniAdame:
@CrochetFeve0251 
```
Your Elementor template was updated. Clear the cache if the display conditions were changed.
```

remyperona (approved):

vmanthos:
@CrochetFeve0251 I can see the message displayed when an Elementor's template is changed while Remove Unused CSS is disabled.

Clicking the "Clear Cache" CTA will clear the whole cache even if the specific template was used for a subset of the site's pages, e.g., only posts or the front page.

@piotrbak I believe this is not the intended behavior, is it?

I'd expect:
1. We'd check the "Display Conditions" for the specific template to
2. Gather the URLs where the template is used and
3. Clear only the affected URLs cache.

By the way, the method you mentioned [here](https://github.com/wp-media/wp-rocket/pull/5956#issuecomment-1725243433) doesn't seem to be used anywhere in the code. :thinking:

MathieuLamiot:
Hello @CrochetFeve0251, @vmanthos, @piotrbak, @DahmaniAdame 👋 
This PR has been going on for too long now and maybe it is just me, but I am unable to understand its scope anymore. I am afraid we are chasing a perfect and optimized compatibility with Elementor and hence, we never move forward.

My understanding was that up until know, we did not display the arning message at the right time. It sounds reasonable to fix this not to wrong the user. I understand this part is OK now? @vmanthos, do you confirm?
If yes, then can we consider the current version has no regression and close it?
If no, what are the issues that make the PR worse than the current release? We would have to uniquely solve those.

If there are follow-up improvements for the compatibility with Elementor, we should have dedicated issues with a well defined scope each. But this one is getting out of hand, so let's close it as soon as possible.

Thanks

piotrbak:
@MathieuLamiot The scope is pretty straightforward:
https://github.com/wp-media/wp-rocket/issues/5848#issuecomment-1658184711

However, the number of going backs to QA is too high, and I don't think we can afford this now. Do you agree to block the issue and later on break it into two parts when we have time? The notice and auto Used CSS clearing is kind of useful and important here.

MathieuLamiot:
Thanks @piotrbak 🙏  This message is indeed pretty clear but probably lost (at least for me) in all the discussions.

We can block it if there is nothing usable quickly in this PR. Otherwise, we might be able to ship it as is as a first step?
Then I agree that we will need to split the remaining stuff in small issues to get it done.

vmanthos:
>  It sounds reasonable to fix this not to wrong the user. I understand this part is OK now?

Notifications are displayed correctly @MathieuLamiot.

Currently, we'll be clearing the whole cache when the user clicks the CTA button even if that's not necessary. In the future, we can improve this and make it more fine-grained.

Issue #5590, opened 2022-11-22
Clear the cache of certain language will clear the whole cache in some cases


Mai-Saad:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version => 3.12.3.1
- Used the search feature to ensure that the bug hasn’t been reported before

**Describe the bug**
Clearing the cache of one language will clear the whole cache 

**To Reproduce**
Precondition:
- WPML language format using qs i.e Language name added as a parameter (https://rocketlabsqa.ovh?lang=ar - Arabic)
- Website has 3 languages
- The shop page is set
1. Install and activate WPR
2. Visit the Product and shop page in 3 different languages => pages are cached and completed in preload table
3. Clear cache of 1 language i.e AR
4. All cache was cleared

**Expected behavior**
Clear only affected language and not clear the whole cache

**Screenshots**

https://user-images.githubusercontent.com/76941962/203273059-99a869fb-cb2e-47e2-b1f3-6c052487864d.mp4

https://user-images.githubusercontent.com/76941962/203283666-0e564b6f-adef-4113-8811-339d1a76add5.mp4

**Additional context**
- Same happened on 3.11.5 
- Sometimes, after step 3, the shop page gets unset automatically when we edit the product
- Other setups for WPML shall be validated
- WPML Version 4.5.8, WooCommerce Version 7.1.0, WCML Version 5.0.2

**Backlog Grooming (for WP Media dev team use only)**
- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

vmanthos:
We are ignoring query strings when clearing the cache and keep only the path:
https://github.com/wp-media/wp-rocket/blob/5ddfb40f083765f1df50398efba339571ce82bc7/inc/functions/files.php#L758

So:
* https://vasilis.rocketlabsqa.ovh/?lang=el
* https://vasilis.rocketlabsqa.ovh/?lang=en
are treated like https://vasilis.rocketlabsqa.ovh/

We will likely need to change how the cache is stored when `$lang` is used. For example:
* `cache/wp-rocket/example.com/#lang=el` should store all the `el` URLs.
* `cache/wp-rocket/example.com/#lang=en` should store all the `en` URLs, etc.

Things will be more complicated when more than one query string is used/cached, e.g. `/about-us?lang=el&location=Greece`.
In such a case, we could store the cache like this:
[...]

PR #5974, opened 2023-06-13
Update CF APO compatiblity

remyperona:
## Description

Use CF plugin class methods directly to manage purge everything and partial purge instead of using their hooks that don't work. 

## Type of change

# Checklist:


vmanthos:
@Tabrisrp I checked cache clearing and currently everything works except for:
* Clearing a URL's cache after the used CSS has been generated for it
* Clearing a URL's cache when using the "Clear Used CSS of this URL" admin bar menu item
* Clearing the cache of date archives
* With WPML, clearing the cache of a specific language clears the whole cache on Cloudflare. This is happening both when the main language is in its directory, and when it's not
* Cache Lifespan - we won't be handling this one according to @piotrbak 

piotrbak:
@vmanthos for the WPML it's okay for the initial release. Lifespan is expected to happen like this.

@Tabrisrp It would be great to fix the first two points, the third one only if we have enough time. The third (an most likely the last alpha would be released today in the afternoon/evening)

vmanthos:
@piotrbak For WPML, @Mai-Saad reminded me of issue #5590. That's likely the reason for what I reported and it's not specific to this PR.

piotrbak:
@vmanthos Well, looks like we are synced with cache then ;)

remyperona:
I added purge after used CSS generation and clean.

The CF plugin doesn't have a public method to clean cache based on an URL value, only from post ID, so anything that is not a post type is not going to be cleaned currently. Maybe we can create an issue on their repo so they add a public method for that?

vmanthos (approved):
Thank you for the fixes, @Tabrisrp! :pray: 

The cache is cleared in both cases related to Remove Unused CSS.

We are good to go with this. :+1: 

**NOTE**

When WPML was enabled, clearing the used CSS for a URL wasn't clearing the cache on Cloudflare APO while our cache was cleared.

This requires further investigation.

CrochetFeve0251 (approved):

engahmeds3ed (approved):

PR #5979, opened 2023-06-15
Fixed disappearing notice Cloudflare

CrochetFeve0251:
## Description
Fix an issue with the error disappearing if we save the page without changing any value. 

For that we fixed a couple of errors and change the check for the error to the `pre_update` filter.

## Type of change

## How Has This Been Tested?
Locally

# Checklist:

### DIFF
--- a/inc/Addon/Cloudflare/Subscriber.php
+++ b/inc/Addon/Cloudflare/Subscriber.php
@@ -440,25 +440,25 @@ class Subscriber implements Subscriber_Interface {
 	 * @param array $old_value An array of previous values for the settings.
 	 * @param array $value     An array of submitted values for the settings.
 	 *
-	 * @return array
+	 * @return void
 	 */
 	public function update_dev_mode( $old_value, $value ) {
 		if ( ! current_user_can( 'rocket_manage_options' ) ) {
-			return $old_value;
+			return;
 		}
 
 		if ( ! isset( $old_value['cloudflare_devmode'], $value['cloudflare_devmode'] ) ) {
-			return $old_value;
+			return;
 		}
 
 		if ( (int) $old_value['cloudflare_devmode'] === (int) $value['cloudflare_devmode'] ) {
-			return $old_value;
+			return;
 		}
 
 		$connection = $this->cloudflare->check_connection( $value['cloudflare_zone_id'] );
 
 		if ( is_wp_error( $connection ) ) {
-			return $old_value;
+			return;
 		}
 
 		$result = [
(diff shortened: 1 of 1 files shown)


inc/Addon/Cloudflare/Subscriber.php
```diff
@@ -440,25 +440,25 @@ private function save_cloudflare_auto_settings( $auto_settings, $old_settings )
 	 * @param array $old_value An array of previous values for the settings.
 	 * @param array $value     An array of submitted values for the settings.
 	 *
-	 * @return void
+	 * @return array
```
remyperona: `update_option_` is an action, callback method doesn't need any returned value, why this change?

inc/Addon/Cloudflare/Subscriber.php
```diff
@@ -485,25 +487,25 @@ public function update_dev_mode( $old_value, $value ) {
...
```
remyperona: same question

remyperona (changes requested):

remyperona (approved):

Mai-Saad:
@CrochetFeve0251 Thanks for the update. Saving the settings while the error is displayed won't hide the error. However,  please find exploratory test notes below:
- Refreshing the page while the error is displayed, will hide the error.  => Shouldn't the error still be displayed here?
- If we have 3.14, where the error was hidden due to the bug there, then we update to the PR, shouldn't the error be displayed? (currently, it is not)

CrochetFeve0251:
> @CrochetFeve0251 Thanks for the update. Saving the settings while the error is displayed won't hide the error. However,  please find exploratory test notes below:
> - Refreshing the page while the error is displayed, will hide the error.  => Shouldn't the error still be displayed here?
> - If we have 3.14, where the error was hidden due to the bug there, then we update to the PR, shouldn't the error be displayed? (currently, it is not)

@Mai-Saad the error is displayed when we save so it is normal if we don't save anything.
Are we supposed to display an error even nothing is currently saved?

Mai-Saad:
@CrochetFeve0251 As per discussion with @piotrbak nothing is needed here and fine as is.

Mai-Saad (approved):
[testrail-report-494.pdf](https://github.com/wp-media/wp-rocket/files/11788479/testrail-report-494.pdf)
Working as expected with the following note (agreed to remain as is)
The 1st time we save settings while having wrong CF addon data + CF plugin is on and authenticated => validation error is displayed and any other save settings after that won't show the error

PR #5993, opened 2023-06-19
Enhancement/5984 woocommerce product sorting not woring with wpr

### LINKED ISSUE #5984: Product sorting in WooCommerce backend does not work when WP Rocket is active
- Made sure you’re on the latest version - Y
- Used the search feature to ensure that the bug hasn’t been reported before - Y

**Describe the bug**
It's not possible to **sort** products within WooCommerce products page. Instead, you can see `524 error` in the console.

**To Reproduce**
1. In a site with WooCommerce, WP Rocket and hosted on Cloudways (not sure what other requirements may be)
2. Go to WooCommerce product page
3. Click on a product's checkbox to try to sort it 
4. The checkbox will show a loading icon that never goes away, and sorting won't be possible

**Expected behavior**
Product sorting should work as expected

**Screenshots**
Product sorting error with the loading icon stuck: https://capture.dropbox.com/drH0ISsNLOK4zhCb (@markonikolic985 screenshot)

**Additional context**
From @engahmeds3ed:

> the problem starts in WooCommerce exactly [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-ajax.php#L2073)
> they get all posts (in this site the number of those posts are +500) with their menu order and loop on them and after updating their menu_order , they clean post cache [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-ajax.php#LL2090C5-L2090C29) and then set the main product menu_order and also clean post cache [here](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-ajax.php#L2111)
> so this is taking lots of time to update and clear post cache.
> My temporary fix is to disable cleaning post cache inside this method only.
> 

```
add_action( 'wp_ajax_woocommerce_product_ordering', function(){
	global $_wp_suspend_cache_invalidation;
	$_wp_suspend_cache_invalidation = true;
}, 9 );

add_action( 'wp_ajax_woocommerce_product_ordering', function(){
	global $_wp_suspend_cache_invalidation;
	$_wp_suspend_cache_invalidation = false;
}, 11 );
```
> you can know more about the feature here:
> https://github.com/woocommerce/woocommerce/issues/31389

Also:

> both sites are on cloudways 

Slack Dev-team escalation thread: https://wp-media.slack.com/archives/C056ZJMHG0P/p1685433858614039
Tickets: https://secure.helpscout.net/conversation/2237699192/419020?folderId=424682 and https://secure.helpscout.net/conversation/2254739596/421844/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

jeawhanlee:
## Description

Disable WP Rocket from clearing post cache when WC product is sorted.

Fixes #5984 

## Type of change

No

## How Has This Been Tested?
Manual

# Checklist:


inc/common/purge.php
```diff
@@ -172,6 +172,16 @@ function rocket_get_purge_urls( $post_id, $post ) {
  * @param WP_Post $post    WP_Post object.
  */
 function rocket_clean_post( $post_id, $post = null ) {
+
+	/**
+	 * Filters post cache cleaning.
+	 *
+	 * @param array $allow_clean_post Allow cleaning post cache if true; otherwise false.
```
CrochetFeve0251: @jeawhanlee filter docblock have a return type if you could add it.

CrochetFeve0251 (changes requested):

CrochetFeve0251 (approved):

remyperona (changes requested):
I don't really like the idea of having a filter that is preventing the function from doing what it is expected to do. It could cause issues down the line, if someone uses it and the support has to dig to understand why cleaning doesn't happen for example.

I'd suggest instead to remove the action on the hook causing the issue (clean_post_cache) with this priority 9, and add it back once the WC call has finished running, on priority 11.

inc/ThirdParty/Plugins/Ecommerce/WooCommerceSubscriber.php
```diff
@@ -14,6 +15,7 @@
  */
 class WooCommerceSubscriber implements Event_Manager_Aware_Subscriber_Interface {
 	use Config_Updater;
+	use ReturnTypesTrait;
```
engahmeds3ed: do we really need this trait here?

engahmeds3ed (changes requested):
but this means that we will not clean those changed posts cache so their ordering will not reflect in category page or at any place, right?

CrochetFeve0251 (approved):

vmanthos:
I haven't been able to reproduce the issue on my test site but I logged cache clearing during sorting of products.
With the PR, cache clearing - which caused the issue, doesn't take place during sorting. So, this part is handled.

However, there is a regression as far as cache clearing goes.

Typically, we clear the `/shop` cache as well as that of all the categories pages the sorted products belong to.
This isn't happening with the current PR which only clears one of the category pages the product belongs to.

To reproduce this:
1. Add two products to the two same categories. Visit their URLs on the front end to make sure they are cached.
2. Change the product order in the sorting tab.
3. Check the URLs of both categories and the `/shop` page.

The latter will remain the same, and only one of the two category URLs will have been cleared.

remyperona (approved):

vmanthos (approved):

**TestRail Report:** [testrail-report-500.pdf](https://github.com/wp-media/wp-rocket/files/11966298/testrail-report-500.pdf)

PR #6021, opened 2023-07-03
Find another way to prevent clearing the whole cache with trash restore without using url_to_postid function

### LINKED ISSUE #5971: From v3.13.1, conflict between WP Rocket and membership plugins like ultimate member, wpforo, and WishList Member

**Describe the bug**
Starting from 3.13.1, and to be more specific in #5680 we received some tickets talking about wrong redirection and not valid behavior when having ultimate member and wpforo plugins active also we got a case with Wishlist Member plugin.

In those cases we managed to get the root cause in the following lines:
https://github.com/wp-media/wp-rocket/blob/50dcab488d8594a6bbeea3bbd11ccb4a8df6d537/inc/functions/files.php#L590-L594

The problem in this WP function is that it's calling WP_Query class [here](https://github.com/WordPress/wordpress-develop/blob/4cb3e1582a022e2d59ba3c4bf8dcb5c60c45eefc/src/wp-includes/rewrite.php#L648)  which is calling the query method in its constructor [here](https://github.com/WordPress/wordpress-develop/blob/48683a21fe444a1e85f193eefa4f15db3a0f8719/src/wp-includes/class-wp-query.php#L3926) and this is calling get_posts method [here](https://github.com/WordPress/wordpress-develop/blob/48683a21fe444a1e85f193eefa4f15db3a0f8719/src/wp-includes/class-wp-query.php#L3794) that applies a filter called the_posts [here](https://github.com/WordPress/wordpress-develop/blob/48683a21fe444a1e85f193eefa4f15db3a0f8719/src/wp-includes/class-wp-query.php#L3509)
this filter is used inside UM plugin to add the callback for the method filter_protected_posts inside their class um\core\Access
there they are trying to filter the posts based on the attached restriction from their dashboard and if the restriction for a post/page is to redirect to another post/page it will redirect exactly there and that's weird (u can find it [here](https://plugins.trac.wordpress.org/browser/ultimate-member/trunk/includes/core/class-access.php#L1317)) so it'll redirect and exit without continue in registration!

**To Reproduce**
1. Install and activate ultimate member plugin and continue the activation with creating the requested pages automatically with that button in admin notice
2. Install and activate wpforo plugin (free forum plugin)
3. go to the created page forum and make UM settings for this page as attached
4. Visit the register page as incognito and complete the registration, you will be redirected to the url that u added in forum page and it should redirect you to the user profile page

image

**Expected behavior**
the redirection should happen as it was before 3.13.1

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
## Description

We get rid of `url_to_postid` function from the cache clearing and find another way to prevent the whole cache clearing with restore as groomed.

Fixes #5971

## Type of change

Exactly as groomed.

## How Has This Been Tested?

Locally and in customers' sites.

# Checklist:

### DIFF
--- a/inc/common/purge.php
+++ b/inc/common/purge.php
@@ -612,7 +612,7 @@ add_action( 'upgrader_process_complete', 'rocket_clean_cache_theme_update', 10, 
  */
 function rocket_clean_post_cache_on_slug_change( $post_id, $post_data ) {
 	// Bail out if the post status is draft, pending or auto-draft.
-	if ( in_array( get_post_field( 'post_status', $post_id ), [ 'draft', 'pending', 'auto-draft' ], true ) ) {
+	if ( in_array( get_post_field( 'post_status', $post_id ), [ 'draft', 'pending', 'auto-draft', 'trash' ], true ) ) {
 		return;
 	}
 	$post_name = get_post_field( 'post_name', $post_id );
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -587,12 +587,6 @@ function rocket_clean_files( $urls, $filesystem = null ) {
 
 		$parsed_url = get_rocket_parse_url( $url );
 
-		$post_id = url_to_postid( $url );
-
-		if ( $post_id ) {
-			$parsed_url = rocket_maybe_find_right_trash_url( $parsed_url, $post_id );
-		}
-
 		if ( ! empty( $parsed_url['host'] ) ) {
 			foreach ( _rocket_get_cache_dirs( $parsed_url['host'], $cache_path ) as $dir ) {
 				// Decode url path.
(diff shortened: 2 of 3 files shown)


CrochetFeve0251 (approved):

jeawhanlee (commented):
A minor update to the test that does not affect PR behaviour, We can remove this [line](https://github.com/wp-media/wp-rocket/blob/4579cea4a88dd711689441d5d33ac349d3cc0f06/tests/Unit/inc/functions/rocketCleanFiles.php#L75) since we no longer use the function

jeawhanlee (approved):

vmanthos (approved):

**TestRail Report:** [testrail-report-498.pdf](https://github.com/wp-media/wp-rocket/files/11947966/testrail-report-498.pdf)

PR #6031, opened 2023-07-06
Fixes #5970: Delay js new jquery regex not working

### LINKED ISSUE #5970: Delay JavaScript execution jQuery and jQuery Migrate new regex not working

**Describe the bug**
The new Delay JavaScript execution regex for jQuery and jQuery Migrate doesn't work.
It works when removing the $ at the end:
\/jquery(-migrate)?-?([0-9.]+)?(.min|.slim|.slim.min)?.js(\?(.*))?

**Expected behavior**
The new regex should work in all cases

**Additional context**
hs: https://secure.helpscout.net/conversation/2262624813/423129
slack: https://wp-media.slack.com/archives/C08N8J6VC/p1686203571360919

jeawhanlee:
## Description
Updated the regex for jquery.

Fixes #5970 

## Type of change

No

## How Has This Been Tested?
Automated Tests

# Checklist:

### DIFF
--- a/inc/Engine/Optimization/DelayJS/Admin/Settings.php
+++ b/inc/Engine/Optimization/DelayJS/Admin/Settings.php
@@ -145,7 +145,7 @@ class Settings {
 	public static function get_delay_js_default_exclusions(): array {
 
 		$exclusions = [
-			'\/jquery(-migrate)?-?([0-9.]+)?(.min|.slim|.slim.min)?.js(\?(.*))?( |\'|"|>)',
+			'\/jquery(-migrate)?-?([0-9.]+)?(.min|.slim|.slim.min)?.js(\?(.*))?(\'|"| |<|>)',
 			'js-(before|after)',
 		];
 
--- a/inc/Engine/Optimization/DelayJS/Admin/Settings.php
+++ b/inc/Engine/Optimization/DelayJS/Admin/Settings.php
@@ -145,7 +145,7 @@ class Settings {
 	public static function get_delay_js_default_exclusions(): array {
 
 		$exclusions = [
-			'\/jquery(-migrate)?-?([0-9.]+)?(.min|.slim|.slim.min)?.js(\?(.*))?(\'|"| |<|>)',
+			'\/jquery(-migrate)?-?([0-9.]+)?(.min|.slim|.slim.min)?.js(\?(.*))?( |\'|"|>)',
 			'js-(before|after)',
 		];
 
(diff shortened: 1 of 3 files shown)


Mai-Saad:
@jeawhanlee Thanks for the PR. During the exploratory test, can see that the regex fails for this markup
`<script src=https://new.rocketlabsqa.ovh/wp-content/rocket-test-data/scripts/jquery.js></script>` or `<script src=https://new.rocketlabsqa.ovh/wp-content/rocket-test-data/scripts/jquery-migrate.js></script>`
=> Exclusion is not working neither was it working on the trunk. @joejoe04 @piotrbak what do you think?

Note: if there is a space in the markup after `.js `, it will work

CrochetFeve0251 (approved):

joejoe04:
@Mai-Saad Thanks for testing this!

I'm not sure about which page you are doing the testing on, but I just ran a test on my test site here: https://i.joedisalvo.com/

Here are the results with no exclusions applied:
image

And if I add the new exclusion like this in my settings:
`\/jquery(-migrate)?-?([0-9.]+)?(.min|.slim|.slim.min)?.js(\?(.*))?( |'|"|>)`

It does work to exclude both scripts:
image

I tried logging in at https://new.rocketlabsqa.ovh/ to test, but I'm seeing the scripts currently have query strings so it's not an exact test, but the new regex does work there too.

I noticed the exclusions were currently like this:
image

So I've left them that way. 

Is there a way I could reproduce the test you're doing without the query strings to see if I can check into this?

Mai-Saad:
@joejoe04 thanks for the feedback.
Seems I was checking the 1st regex here https://github.com/wp-media/wp-rocket/issues/5970#issuecomment-1616611263 and it was the one causing the problem. 
However, Confirmed that when using `\/jquery(-migrate)?-?([0-9.]+)?(.min|.slim|.slim.min)?.js(\?(.*))?( |'|"|>)` => the script `<script src=https://new.rocketlabsqa.ovh/wp-content/rocket-test-data/scripts/jquery.js></script>` is sucessfully excluded.

Notes: 
- The mentioned regex here will exclude the following script (which was not the case on Trunk) => as per discussion with @piotrbak, this is fine and probably it was an issue at trunk.

```
  alert("/jquery-migrate.js");
  </script>
```
- Test page https://new.rocketlabsqa.ovh/jquery-template/

Mai-Saad (commented):
Working as expected with the following notes (agreed to remain as is)
- The jquery exclusion line in UI has empty space at the beginning of the line
- We are not handling the wrong markup like this one `<script src=https://i.joedisalvo.com/wp-content/themes/twentytwentytwo/jquery-migrate.min.js</script>`

If nothing is needed from CR then we can merge this one
[testrail-report-501.pdf](https://github.com/wp-media/wp-rocket/files/12002668/testrail-report-501.pdf)

CrochetFeve0251 (approved):

PR #6079, opened 2023-08-01
Added a fix to prevent version from clearing the cache

### LINKED ISSUE #5958: WP Rocket's cache is cleared after every update
- Made sure you’re on the latest version :heavy_check_mark: `3.13.3`
- Used the search feature to ensure that the bug hasn’t been reported before :heavy_check_mark: 

**Describe the bug**

When updating WP Rocket the cache is cleared every time although this may not be the intended behavior. 
For example, we have code to clear the cache when updating from a version lower than 3.6:
https://github.com/wp-media/wp-rocket/blob/3b6300ec350c3fa60bdae03c32bc8d5c4b9cb7f9/inc/admin/upgrader.php#L272-L275

This is happening because `version` isn't part of the `$removed` array:
https://github.com/wp-media/wp-rocket/blob/3b6300ec350c3fa60bdae03c32bc8d5c4b9cb7f9/inc/admin/options.php#L25-L54

so the following conditional always evaluates to `true` and `rocket_clean_domain()` runs clearing the cache:
https://github.com/wp-media/wp-rocket/blob/3b6300ec350c3fa60bdae03c32bc8d5c4b9cb7f9/inc/admin/options.php#L91

**To Reproduce**

1. Install WP Rocket `3.13`.
2. After the cache is preloaded disable Preload.
3. Check the `/cache/wp-rocket/` folder.
4. Update to the latest WP Rocket version (`3.13.3` at the time this is written).
5. Repeat step 3. The cache will have been cleared.

**Expected behavior**

This issue should be seen as an opportunity to discuss whether we'll keep the current behavior, i.e. clear the cache after every update, or if we want to do that selectively, e.g. with specific versions.

**Screenshots**

N/A.

**Additional context**

N/A.

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description
Prevent clearing the cache on each update by excluding `version` from the parameters to check to clear cache.

Fixes #5958

## Type of change

No.

## How Has This Been Tested?

- [ ] Unit Test 

# Checklist:

### DIFF
--- a/inc/admin/options.php
+++ b/inc/admin/options.php
@@ -51,6 +51,7 @@ function rocket_after_save_options( $oldvalue, $value ) {
 		'manual_preload'              => true,
 		'preload_excluded_uri'        => true,
 		'cache_reject_uri'            => true,
+		'version'                     => true,
 	];
 
 	// Create 2 arrays to compare.
--- a/tests/Fixtures/inc/admin/rocketAfterSaveOptions.php
+++ b/tests/Fixtures/inc/admin/rocketAfterSaveOptions.php
@@ -32,6 +32,7 @@ return [
 		'analytics_enabled'   => '',
 		'cdn'                 => false,
 		'cdn_cnames'          => false,
+		'version'             => '3.15',
 	],
 
 	// Test data.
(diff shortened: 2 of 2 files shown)


remyperona (approved):

vmanthos (approved):

PR #6084, opened 2023-08-02
Prevent flooding AS queue

### LINKED ISSUE #6058: Improve handling the failed preload items

**Describe the bug**

**To Reproduce**
1. Set cron to run each 10 minutes
2. Set the batch to be bigger, let's say 100 to make problem more visible
3. Run the preload on the website with at least 200 URLs
4. 100 Actions are added to the AS, 100 URLs are set to `in-progress`
5. Some URLs are preloaded, the others are waiting for the next CRON run
6. Wait 10 minutes, so the cron can pick up the next batch
7. The remaining URLs (`in-progress` [older than 100/15 minutes](https://github.com/wp-media/wp-rocket/blob/ecbcba2177f89b03c6231d3ab314ce647c807e93/inc/Engine/Preload/Controller/PreloadUrl.php#L202)) in the database [are set to failed](https://github.com/wp-media/wp-rocket/blob/ecbcba2177f89b03c6231d3ab314ce647c807e93/inc/Engine/Preload/Controller/PreloadUrl.php#L205) while corresponding async actions are still in the AS
8. We're setting to `in-progress` another 100 URLs, and registering another 100 Actions

**Expected behavior**
We should avoid exceeding number of Preload Actions in the AS defined by the batch size filter.

I'm not sure what would be the best solution here, I can think about some options, but further R&D is needed:

1 Check the number of `in-progress` items that we have in the queue before setting old ones to the `failed` status
- The problem here is that after more than 2 iterations, we **might** end up in the same situation

2 Set the `failed` status only for URLs that doesn't have corresponding Preload Action registered in the AS
- This one most likely will be resource consuming, but in terms of quality it's the best I can imagine

3 Our [check](https://github.com/wp-media/wp-rocket/blob/ecbcba2177f89b03c6231d3ab314ce647c807e93/inc/Engine/Preload/Database/Queries/Cache.php#L340) about `in-progress` items could also take into the consideration the number of registered AS actions related to Preload.

**Additional context**
This will help with issues related to the preload affecting the database size:
https://github.com/wp-media/wp-rocket/issues/6052
https://github.com/wp-media/wp-rocket/issues/6057

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

CrochetFeve0251:
## Description

Prevent the AS queue to be flooded with preload actions.
For that we now check number of actions in the queue instead of checking actions inside the cache table.
We do the same for the failed actions and now check if they exists inside AS queue first.

Fixes #6058

## Type of change

No.

## How Has This Been Tested

- [ ] Integration Test

# Checklist:

### DIFF
--- a/inc/Engine/Preload/Controller/PreloadUrl.php
+++ b/inc/Engine/Preload/Controller/PreloadUrl.php
@@ -180,8 +180,10 @@ class PreloadUrl {
 	 * @return void
 	 */
 	public function process_pending_jobs() {
-		$count = apply_filters( 'rocket_preload_cache_pending_jobs_cron_rows_count', 45 );
 
+		$pending_actions = $this->queue->get_pending_preload_actions();
+
+		$count = ( (int) apply_filters( 'rocket_preload_cache_pending_jobs_cron_rows_count', 45 ) ) - count( $pending_actions );
 		/**
 		 * Set the delay before an in-progress row is considered as outdated.
 		 *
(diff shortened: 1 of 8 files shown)


remyperona (approved):

remyperona (approved):

vmanthos (approved):

**TestRail report:**  [testrail-report-512.pdf](https://github.com/wp-media/wp-rocket/files/12279039/testrail-report-512.pdf)

PR #6087, opened 2023-08-03
add filter rocket_delay_remove_rucss_failed_jobs to delete faild rucss jobs instead of 3 days hardcoded

### LINKED ISSUE #6066: get_failed_rows date query should match the value of the rocket_remove_rucss_failed_jobs_cron_interval filter

**Describe the bug**
We can filter the `rocket_remove_rucss_failed_jobs_cron_interval` and set the recurrence of **WP Rocket clear Remove Unused CSS failed jobs** to 1 hour for example.

This should allow to re-process the **failed** jobs faster. 

But in the plugin code, the `get_failed_rows` function has the **modified** date query hardcoded to `3 days ago`
https://github.com/wp-media/wp-rocket/blob/develop/inc/Engine/Optimization/RUCSS/Database/Queries/UsedCSS.php#L423

The result is no **failed** row "younger" than 3 days wil be set to **pending**. 

**To Reproduce**
1.  On a site with failed RUCSS jobs
2. Install this [helper plugin](https://github.com/wp-media/wp-rocket-helpers/blob/master/rucss/wp-rocket-reduce-failed-rucss-jobs-clearing-interval/wp-rocket-reduce-failed-rucss-jobs-clearing-interval.php) to set the rocket_remove_rucss_failed_jobs_cron_interval to 1 hour
3. Deactivate/reactivate RUCSS so the Event schedule is updated. 
4. Wait for the job to run, or manually execute it using WP Crontrol. 
5. See no **failed** job older than 3 days reset to **pending**

**Expected behavior**
The `get_failed_rows` function should be aware of the value of the `rocket_remove_rucss_failed_jobs_cron_interval` filter so the query is dynamic and not static to "3 days ago".

**Additional context**
Slack threads: 
https://wp-media.slack.com/archives/C029G1B5HD2/p1690205678240629
https://wp-media.slack.com/archives/C08N8J6VC/p1690270849029839

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

mostafa-hisham:
## Description

add new filter `rocket_delay_remove_rucss_failed_jobs` and use its value in `get_failed_rows` function

Fixes #6066 

## Type of change

No

## How Has This Been Tested?

- [ ] Locally

# Checklist:


CrochetFeve0251 (approved):

vmanthos:
Thank you @mostafa-hisham for the PR. 
It's working fine when the value-unit pair is correct, i.e. if it follows these:
https://dev.mysql.com/doc/refman/8.0/en/expressions.html#temporal-intervals

If `rocket_delay_remove_rucss_failed_jobs` returns any of the following:
`null`, float, boolean, `[]`, or an incorrect value-unit pair, e.g. `1 day`, 

then the failed jobs will stay there until the used CSS is manually or automatically truncated.

Maybe instead of retuning if the `$delay` is not valid (all cases), we could set the default values here:
https://github.com/wp-media/wp-rocket/blob/874c2ccfda5f63e9fbdd26abd8d70a7652cff268/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L836-L838

@piotrbak Should we check if the units are correct according to [this](https://dev.mysql.com/doc/refman/8.0/en/expressions.html#temporal-intervals) - this can complicate things on our part, or should this better be covered in our documentation?

CrochetFeve0251:
> Thank you @mostafa-hisham for the PR. It's working fine when the value-unit pair is correct, i.e. if it follows these: https://dev.mysql.com/doc/refman/8.0/en/expressions.html#temporal-intervals
> 
> If `rocket_delay_remove_rucss_failed_jobs` returns any of the following: `null`, float, boolean, `[]`, or an incorrect value-unit pair, e.g. `1 day`,
> 
> then the failed jobs will stay there until the used CSS is manually or automatically truncated.
> 
> Maybe instead of retuning if the `$delay` is not valid (all cases), we could set the default values here:
> 
> https://github.com/wp-media/wp-rocket/blob/874c2ccfda5f63e9fbdd26abd8d70a7652cff268/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L836-L838
> 
> @piotrbak Should we check if the units are correct according to [this](https://dev.mysql.com/doc/refman/8.0/en/expressions.html#temporal-intervals) - this can complicate things on our part, or should this better be covered in our documentation?

I agree it is more logic to clean failed tasks as otherwise it will flood the table.

mostafa-hisham:
@piotrbak After checking this with @vmanthos 
We found out that changing failed jobs status to pending will not give them a new Job ID which means it is going to try to fetch the same failed data in this case `400 bad json ` 3 more times

Should we delete the records so it trie to create new jobs on Saas side?

thanks @vmanthos for checking this with me

piotrbak:
@mostafa-hisham When we receive the `400 bad json`, it means taht the job was already processed by the saas, right? There's no option that this job is in the saas queue?

mostafa-hisham:
@piotrbak, when we receive 400 bad json, means that the job is processed by saas and was deleted before the plugin gets its data.

MathieuLamiot:
See proposed solution [here](https://wp-media.slack.com/archives/C015FT0RBKQ/p1691578008322849?thread_ts=1691564901.977129&cid=C015FT0RBKQ). Moving back to In progress.

CrochetFeve0251 (approved):

MathieuLamiot:
Following [this](https://wp-media.slack.com/archives/C015FT0RBKQ/p1691654055694549?thread_ts=1691564901.977129&cid=C015FT0RBKQ), let's move forward with the PR and tackle the discovered new issues in a dedicated enhancement story.
Moved the task to Ready for PR, and now ready for QA.

MathieuLamiot:
Following [this](https://wp-media.slack.com/archives/C015FT0RBKQ/p1691666126183749?thread_ts=1691564901.977129&cid=C015FT0RBKQ):
Let's keep the PR here and wait for a proper handling of failed status and related JobID in the plugin so that we can actually retry a new job.

inc/Engine/Optimization/RUCSS/Database/Queries/UsedCSS.php
```diff
@@ -199,25 +199,26 @@ public function make_status_inprogress( int $id ) {
 	}
 
 	/**
-	 * Change the status to be pending.
+	 * Reset th job and add new job_id pending.
```
CrochetFeve0251:
```suggestion
	 * Reset the job and add new job_id pending.
```

tests/Unit/inc/Engine/Optimization/RUCSS/Controller/UsedCSS/clearFailedUrls.php
```diff
@@ -64,14 +64,27 @@ public function testShouldReturnAsExpected( $config, $expected ) {
         if ( isset( $config['is_int'] ) ) {
             if ( ! $config['is_int'] ) {
                 $this->usedCssQuery->expects( self::never() )
-                                ->method( 'revert_to_pending' );
+                                ->method( 'reset_job' );
             }
             else {
+				//var_dump($config['is_int']);
```
CrochetFeve0251:
```suggestion

```

tests/Integration/inc/Engine/Optimization/RUCSS/Cron/Subscriber/cronRemoveFailedJobs.php
```diff
@@ -53,4 +52,19 @@ public function testShouldDoExpected( $input, $expected ){
...
+			'response' => array('code' => 200),
+			'body'=>'{"code": 200,
+			"message": "Added to Queue successfully.",
+			"contents": {
+				"jobId": "OVH_EU--496540278",
+				"queueName": "EU",
+				"isHome": false,
+				"queueFullName": "rucssJob_EU"
+				}
+			}'
```
CrochetFeve0251:
```suggestion
			'body'=> json_encode([
			"code" => 200,
			"message" => "Added to Queue successfully.",
			"contents" => [
				"jobId" => "OVH_EU--496540278",
				"queueName" => "EU",
				"isHome" => false,
				"queueFullName" => "rucssJob_EU"
				]
			]),
```

tests/Integration/inc/Engine/Optimization/RUCSS/Cron/Subscriber/cronRemoveFailedJobs.php
```diff
@@ -19,23 +19,22 @@ public static function set_up_before_class() {
...
+		remove_filter('pre_http_request', [$this, 'edit_http_request']);
 		parent::tear_down();
 	}
 
 	/**
 	 * @dataProvider configTestData
 	 */
 	public function testShouldDoExpected( $input, $expected ){
-		$this->markTestSkipped('We cannot reach the Saas servers. Need to revisit.');
+		add_filter('pre_http_request', [$this, 'edit_http_request'], 10, 3);
```
CrochetFeve0251: @mostafa-hisham is there an issue adding it to the `set_up` method that would be more clean as it is what we usually do for filters.

CrochetFeve0251 (changes requested):

CrochetFeve0251 (approved):

Mai-Saad:
@mostafa-hisham Thanks for the PR. Notes from the exploratory test:
-  When using the filter with [] `add_filter( 'rocket_delay_remove_rucss_failed_jobs', function( $interval ){
return [];
} );` ,we will have a warning `[17-Aug-2023 09:08:03 UTC] PHP Warning:  Array to string conversion in /var/www/new.rocketlabsqa.ovh/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php on line 833` => Will remain as is as per discussion
- (fixed) When we switch failed to pending, the last modified time stamp is not changed
- If we executed the Cron `rocket_remove_rucss_failed_jobs` while external requests are blocked, we won't switch failed to pending till requests are not blocked => As per discussion, this is the new behavior as we request job id before switching to pending. 
- Probably not related to PR and further investigation is needed => when change retries and status change to failed while external requests are blocked, we are not changing last modified. @piotrbak what do you think?

CrochetFeve0251 (approved):

Mai-Saad (approved):
Working as expected with the following notes:
1- Filter won't apply decimal values i.e 1.5 days will be ignored and the default is applied
2- Empty array if used will generate warning
3- We may have a corner case where URLs at cache are locked while URLs are failed in used CSS while external requests are blocked => we need to keep an eye on this ( further investigation may be needed)
[testrail-report-518.pdf](https://github.com/wp-media/wp-rocket/files/12377910/testrail-report-518.pdf)

PR #6105, opened 2023-08-15
Hotfix: Change domain admin notice is showing when multilingual plugin is installed

engahmeds3ed:
## Description

Here we are adding a check that the change domain logic is happening only with admin UI not with any ajax request because when multilingual plugin is installed the home_url is changed in frontend leading to showing the notice in admin.

## Type of change

Not groomed, it's a hotfix.

## How Has This Been Tested?

Tested on the customer's site by @alfonso100.

# Checklist:


inc/Engine/Admin/DomainChange/Subscriber.php
```diff
@@ -63,6 +64,10 @@ public static function get_subscribed_events() {
 	 * @return void
 	 */
 	public function maybe_launch_domain_changed() {
+		if ( wp_doing_ajax() ) {
```
remyperona, line 65: are we sure this is going to be enough for all i18n plugins we're compatible with?

engahmeds3ed: The main problem is that we need to only execute this part of code in one place (frontend or backend) so even if the multilingual plugin is playing with the home url, it will play with it in frontend not backend, but we can let QA team validate this in every case, we tested it in the customer's site and it worked properly.

CrochetFeve0251: @Tabrisrp that has a strong chance to work as I when I picked backend url it was cause translation plugins weren't modifying it. 
It is true ajax front end request are calling `ajax-admin.php`.
Do you have any idea if the REST API is handled on the front or the backend? Cause it would be last thing I think that could modify the URL.

MathieuLamiot: @CrochetFeve0251 @engahmeds3ed @Tabrisrp, Unless it breaks something else, I would be in favor of releasing this version as hotfix quickly as it fixes the original issue reported. 
If it can further be improved, that should definitely be done but if it needs longer investigation, maybe it can come after the hotfix?
@piotrbak any preference on your side?

piotrbak: @MathieuLamiot I'd release it asap and then go with more hotfixes versions if needed. We knew that this update will be tricky but it's not possible to do exploratory testing for whole WP ecosystem.

We can do a quick tests for some language switchers of multilingual plugins until it passes the CR.

MathieuLamiot: @CrochetFeve0251, @engahmeds3ed, @Tabrisrp can we move forward with the CR quickly then, so that @Mai-Saad can do QA?
Thank you 🙏

engahmeds3ed: @Mai-Saad already started QA to validate it for all of our supported multilingual plugins.

CrochetFeve0251 (approved):

Mai-Saad (approved):
- Reproducible for WPML and translate press => fixed with the PR
- Not reproducible for Weglot and Polylang plugins
- Notice is displayed when changing domain at config file and when cloning the website using the WPvivid plugin
- If we update WPR while the notice is displayed, it will remain displayed till we click the regenerate btn. However, change language after that won't display the notice.
- When clicking regenerate config btn after changing the domain name in wp-config, only the home page is preloaded after the cache clear (same on 3.14.4, further investigation is needed)

PR #6108, opened 2023-08-17
Fix WPML subdomain languages false notice

engahmeds3ed:
## Description

Here we stop using `home_url()` and use `get_option( 'home' )` instead because all Multilingual plugins are changing the filter `home_url` so we need a not changing value to depend on.

Also here we added the autoload to the used option to make it faster to load as it's loading with `admin_init` which runs with every admin UI page.

**Steps to reproduce the issue:**

1. Setup different domain per language
2. Go to the WP Admin
3. Change the language from default to different one in the top menu

## Type of change

# Checklists

## Test summary

- [x] I validated all Acceptance Criteria of the related issues. (If applicable, provide proof).
- [x] I validated all test plan the QA Review asked me to.

### DIFF
--- a/inc/Engine/Admin/DomainChange/Subscriber.php
+++ b/inc/Engine/Admin/DomainChange/Subscriber.php
@@ -66,21 +66,20 @@ class Subscriber implements Subscriber_Interface {
 			return;
 		}
 
-		$base_url         = trailingslashit( get_option( 'home' ) );
-		$base_url_encoded = base64_encode( $base_url ); // phpcs:ignore WordPress.PHP.DiscouragedPHPFunctions.obfuscation_base64_encode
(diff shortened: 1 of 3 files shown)


remyperona (approved):

Mai-Saad:
- WPML directory setup not showing notice is fine
- notice displayed when changing domain on directory base is fine
- Translate press not showing notice is fine (not sure how to have domain base setup here, is this a case?)
- Clone single site , displaying notice => is fine
- WPML domain base setup not displaying notice => is fine
- notice displayed when change domain on domain base is fine but **click regenerate here will clear only cache of main domain (probably the same problem we were talking about), clear all config files but regenerate only the main one and manual visit won't regenerate config file**
- on multisite notice not displayed on subsite when change domain name ( change main domain name is out of scope)

PR #6116, opened 2023-08-22
Don't define WP_ROCKET_CACHE_PATH if already set

tfgrass:
## Description

We are using WP Rocket with our Bedrock sites. One Problem we encountered, is that we have to set WP_ROCKET_CACHE_PATH early in our configurations, because else the constant is not ready when advanced-cache.php  is executed. 

Since we already defined WP_ROCKET_CACHE_PATH, we always get a Warning:
Constant WP_ROCKET_CACHE_PATH already defined
app/plugins/wp-rocket/wp-rocket.php:63

## Type of change

# Checklists

## Test summary

### DIFF
--- a/wp-rocket.php
+++ b/wp-rocket.php
@@ -60,7 +60,10 @@ define( 'WP_ROCKET_ASSETS_IMG_URL',        WP_ROCKET_ASSETS_URL . 'img/' );
 if ( ! defined( 'WP_ROCKET_CACHE_ROOT_PATH' ) ) {
 	define( 'WP_ROCKET_CACHE_ROOT_PATH', WP_CONTENT_DIR . '/cache/' );
 }
-define( 'WP_ROCKET_CACHE_PATH',         WP_ROCKET_CACHE_ROOT_PATH . 'wp-rocket/' );
+
+if ( ! defined( 'WP_ROCKET_CACHE_PATH' ) ) {
+	define( 'WP_ROCKET_CACHE_PATH',         WP_ROCKET_CACHE_ROOT_PATH . 'wp-rocket/' );
+}
 define( 'WP_ROCKET_MINIFY_CACHE_PATH',  WP_ROCKET_CACHE_ROOT_PATH . 'min/' );
 define( 'WP_ROCKET_CACHE_BUSTING_PATH', WP_ROCKET_CACHE_ROOT_PATH . 'busting/' );
 define( 'WP_ROCKET_CRITICAL_CSS_PATH',  WP_ROCKET_CACHE_ROOT_PATH . 'critical-css/' );


piotrbak:
Hello @tfgrass thanks for creating the PR.

Any reason why not to use `WP_ROCKET_CACHE_ROOT_PATH` constant?

tfgrass:
@piotrbak  we are also setting WP_ROCKET_CACHE_ROOT_PATH

```
/**
 * define wp_rocket paths
 */
Config::define('WP_ROCKET_CONFIG_PATH', $root_dir . '/config/wp-rocket-config/');
Config::define('WP_ROCKET_CACHE_ROOT_PATH', $webroot_dir . '/app/cache/');
Config::define('WP_ROCKET_CACHE_PATH', $webroot_dir . '/app/cache/wp-rocket/');
```

Problem is - in advanced-cache.php it tries to access WP_ROCKET_CACHE_PATH before the wp-rocket.php is executed, so we have to set it seperatly. 
```

<?php
defined('ABSPATH') or exit;

define('WP_ROCKET_ADVANCED_CACHE', true);

$rocket_path        = WP_CONTENT_DIR.'/plugins/wp-rocket/';
$rocket_config_path = WP_ROCKET_CONFIG_PATH;
$rocket_cache_path  = WP_ROCKET_CACHE_PATH;

if (file_exists($rocket_path . 'inc/front/process.php')) {
    include($rocket_path . 'inc/front/process.php');
} else {
    define('WP_ROCKET_ADVANCED_CACHE_PROBLEM', true);
}
```

Maybe it would make sense to somehow include the wp-rocket.php so the constants are ready for use in advanced cache, but I'm not sure whether this would break something

piotrbak:
@tfgrass I might be missing something, but for me it doesn't make sense to define `WP_ROCKET_CACHE_PATH` like this. Mentioned constant is dependent on the `WP_ROCKET_CACHE_ROOT_PATH`, so if you set `WP_ROCKET_CACHE_ROOT_PATH` to be `/app/cache/`, `WP_ROCKET_CACHE_PATH` will automatically be `/app/cache/wp-rocket/` as you can see here:
https://github.com/wp-media/wp-rocket/blob/68b2aa28954952b7145a72e705d185d97dfbc0c9/wp-rocket.php#L60-L63

If I'm missing something, please elaborate on this. We have many customers using Bedrock and that's the first request related to this constant.

piotrbak:
Also, `inc/front/process.php` is not used since many many years

tfgrass:
@piotrbak just for my understanding - this is an old version of wp-rockets advanced-cache.php dropin? 

I just checked the backend, to see if it gives me this error https://docs.wp-rocket.me/article/134-advanced-cache-error-message - but in the backend I don't see any message like this. 

I will check if using the latest version of the drop in fixes the issue - this would be the right one, i guess? https://github.com/wp-media/wp-rocket/blob/68b2aa28954952b7145a72e705d185d97dfbc0c9/views/cache/advanced-cache.php#L4

Issue #5423, opened 2022-09-13
Adjust Preload notice 


camilamadronero-zz:
**Is your feature request related to a problem? Please describe.**
The [Preload notice](https://jmp.sh/CtTXxBi) is currently shown in when [clearing the cache](https://docs.wp-rocket.me/article/78-how-often-is-the-cache-updated#full-cache-deletion) and it has a [re-appearing behavior](https://wp-media.slack.com/archives/GUKB44GNN/p1663011215372739). This doesn't go inline with the continuous nature of the new Preload, and it's causing questions.

**Describe the solution you'd like**

1- Since Preload is now an ongoing and continuous process, almost infinite, the notice could be shown only:

- After enabling the plugin/Preload 
- When using the `Clear and preload cache` option (top bar and Quick Actions).

Customers can trust WP Rocket is already doing it smartly, and we'll have a clearer settings page.
We already do something similar by not showing the "Cache cleared" notice when changing themes, for example.

2 - Additionally, the notice should be dismissible.

**Additional context**
Slack discussions: https://wp-media.slack.com/archives/G8L4EE8N9/p1663089999191189 and https://wp-media.slack.com/archives/GUKB44GNN/p1663011215372739.

camilamadronero-zz:
Related https://secure.helpscout.net/conversation/2017147213/370362/

vmanthos:
Related: https://secure.helpscout.net/conversation/2022286927/371469?folderId=2135277

PR #6120, opened 2023-08-28
Preserve the box rocket_activation_notice with update

### LINKED ISSUE #5889: The activation notice shows up after every update
- Made sure you’re on the latest version ✔
- Used the search feature to ensure that the bug hasn’t been reported before ✔

**Describe the bug**
There is a message after plugin activation, that says "WP Rocket is now activated...". This message should not be displayed after every plugin update if the user is dismissed it earlier.

**To Reproduce**
1. Install an older version of WR Rocket.
2. Open the `/wp-admin/options-general.php?page=wprocket#dashboard` page.
3. Click on the Dismiss `×` icon.
4. Update the plugin. 
5. Open the `/wp-admin/options-general.php?page=wprocket#dashboard` page again.
6. The Welcome notice is visible again.

**Expected behavior**
If the message is dismissed, never show it again. We know that the plugin is working. 

**Screenshots**
image

**Additional context**
–

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

engahmeds3ed:
## Description

Preserve the box rocket_activation_notice with update.

Fixes #5889

## Type of change

No

# Checklists

## Test summary

- [x] I validated all Acceptance Criteria of the related issues. (If applicable, provide proof).


remyperona (approved):

Mai-Saad:
@engahmeds3ed Thanks for the PR. Please find exploratory test notes below:
1- If we dismiss the notice then deactivate WPR and activate WPR again. Notice will be displayed.
2- If we update from the previous version to the PR without dismissing notice before the update. it will be automatically dismissed.

engahmeds3ed:
> @engahmeds3ed Thanks for the PR. Please find exploratory test notes below: 1- If we dismiss the notice then deactivate WPR and activate WPR again. Notice will be displayed. 2- If we update from the previous version to the PR without dismissing notice before the update. it will be automatically dismissed.

Good catch, based on the current code yes this is expected after checking this old code.
Let me try fixing that without affecting other boxes.

engahmeds3ed:
@Tabrisrp do u remember why do we clear all rocket boxes with update?

remyperona:
I do not, it's something that was there before my time. It probably made sense at some point, but not anymore.

engahmeds3ed:
Thanks @Tabrisrp 

Summary: the [used function ](https://github.com/wp-media/wp-rocket/blob/136d94b9ea6eb33ba33c3e2143a623b4b37d86cc/inc/admin/upgrader.php#L28) we remove all dismissed notices for all users and only fire dismissing the following boxes with every update:

```
One or more plugins have been enabled or disabled, clear the cache if they affect the front end of your site.
```

and the second box is the activation notice.

Even if the user didn't dismiss both boxes, both will be dismissed based on the current code.

So based on the above discussion we have three solutions here:

1. Replace [this code](https://github.com/wp-media/wp-rocket/blob/901911571c027b0b4a5979651a52ef3c0edfbe5f/inc/functions/admin.php#L41-L42) with a direct query to delete all user boxes and keep only the dismissed boxes **but this is bad from performance point of view.**
2. Don't remove all boxes with update at all and keep what's ignored kept ignored.
3. Accept the PR as is with the issues that @Mai-Saad mentioned above as known issues.

I believe we need the @wp-media/productrocket decision here.

piotrbak:
@engahmeds3ed @Tabrisrp Do you see any reason why the boxes might be cleared? It doesn't seem to be making sense currently

engahmeds3ed:
> @engahmeds3ed @Tabrisrp Do you see any reason why the boxes might be cleared? It doesn't seem to be making sense currently

Really I don't know.

piotrbak:
@engahmeds3ed Let's go with 2 then.

Mai-Saad:
Scenario 2 here https://github.com/wp-media/wp-rocket/pull/6120#issuecomment-1708193183 is fixed, while 1 is still there

engahmeds3ed:
Thanks @Mai-Saad it should work now, can u plz also validate this old issue https://github.com/wp-media/wp-rocket/issues/5423 as it was touched here in this PR.

Mai-Saad:
Working as expected now both 1,2 are fine

PR #6123, opened 2023-08-29
Do not serve WebP images that are larger than the original

Roy-Orbison:
The current code assumes the `.webp` is always smaller, but occasionally it isn't. Optimiser plugins may report this situation as "already optimised", or similar. In some cases, the optimiser plugin provides both a WebP version and an optimised version with the same name as the original, and the WebP might be smaller than the original but larger than the other optimised file.

## Description

This adds a size check to the webp URL replacer function. I noticed that Rocket was still replacing URLs for which Imagify reports "already optimised", so the images served were larger, which degraded performance.

## Type of change

- This change requires a documentation update. (I was thinking a note somewhere on https://docs.wp-rocket.me/article/1282-webp tha mentions webp would not be served unless it is smaller than the main file.)

N.a.

# Checklists

- [ ] If applicable, I have made corresponding changes to the documentation.

## Test summary

I ran this patch on a live site (with an additional logging line inside the size check `if` block), and it worked as intended.

### DIFF
--- a/inc/Addon/WebP/Subscriber.php
+++ b/inc/Addon/WebP/Subscriber.php
@@ -435,14 +435,20 @@ class Subscriber extends AbstractWebp implements Subscriber_Interface {
 
 		if ( $this->filesystem->exists( $src_path_webp ) ) {
 			// File name: image.jpg => image.webp.
-			return preg_replace( '@\.' . $src_url['extension'] . '$@', '.webp', $src_url['src'] ) . $src_url['query'];
-		}
-
-		if ( $this->filesystem->exists( $src_path . '.webp' ) ) {
+			$src_url_webp = preg_replace( '@\.' . $src_url['extension'] . '$@', '.webp', $src_url['src'] ) . $src_url['query'];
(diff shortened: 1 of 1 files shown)


inc/Addon/WebP/Subscriber.php
```diff
@@ -435,14 +435,20 @@ private function url_to_webp( $url, $extensions ) {
 
 		if ( $this->filesystem->exists( $src_path_webp ) ) {
 			// File name: image.jpg => image.webp.
-			return preg_replace( '@\.' . $src_url['extension'] . '$@', '.webp', $src_url['src'] ) . $src_url['query'];
+			$src_url_webp = preg_replace( '@\.' . $src_url['extension'] . '$@', '.webp', $src_url['src'] ) . $src_url['query'];
+		} elseif ( $this->filesystem->exists( $src_path_webp = $src_path . '.webp' ) ) { // phpcs:ignore Generic.Sniffs.CodeAnalysis.AssignmentInCondition,Squiz.Sniffs.PHP.DisallowMultipleAssignments
```
Roy-Orbison, line 439: Assigning within  the condition avoids unnecessarily nested `if` statements. it's not meant to be a comparison.

PR #6150, opened 2023-09-07
Fixes 6132: Added tests and new logic for supporting pseudoclasses

### LINKED ISSUE #6132: LazyLoad Background CSS Images pseudo classes handling improvement

**Describe the bug**
LazyLoad Background CSS Images is not handling correctly CSS background images that are added to specific selectors with pseudo classes:
-`:after`
-`.class>:after`
-`.class&:after`
-`.class~:after`

Single colon syntax is possible only for the following pseudo-elements:
`::before`, `::after`, `::first-line`, and `::first-letter`
and they all should be covered. We should also cover single colon syntax for the following ones:
`:hover` `:focus` `:active` `:visited` `:focus-within` `:focus-visible`

For the double colon we should cover all of them with those special characters added just before:
-`.class>::after`
-`.class&::after`
-`.class~::after`

The root cause here is that `.class>:after` is a correct syntax while `.class>` is not. Our JS code is not able to find the pseudo classes and when removing them it reports wrong markup `.class>`.

**To Reproduce**
1. Add background image to any of the above classes and make sure the HTML matches
2. Enable Lazyload background CSS images
3. Open the website
4. Image is not loaded

**Expected behavior**
We should cover those scenarios, it might need a bit of r&d to find the best solution

1. We should be able to correctly find parent HTML elements related to pseudo classes
2. Lazy Load for background images should work for all the examples mentioned above
3. Lazy Load for background images should work for `.class::after` without regression
4. Lazy Load for background images should work for regular classes coming from internal style or CSS files

CrochetFeve0251:
## Description

Add support for pseudoclasses for the bg css ll.
For that  we refactored the logic from the `MappingParser`.

Fixes #6132
## Type of change

No.

# Checklists

## Test summary

### DIFF
--- a/inc/Engine/Media/Lazyload/CSS/Front/MappingFormatter.php
+++ b/inc/Engine/Media/Lazyload/CSS/Front/MappingFormatter.php
@@ -12,6 +12,7 @@ class MappingFormatter {
 	 */
 	public function format( array $data ): array {
 		$formatted_urls = [];
+
 		foreach ( $data as $datum ) {
 			$hash = $datum['hash'];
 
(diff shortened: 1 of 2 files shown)


engahmeds3ed (changes requested):
It'll be easier for us to maintain or even review if u changed those two anonymous functions to normal methods in the class.
You can also simplify things by creating a new method to return the filtered `pseudo_elements_to_remove`

what do u think?

CrochetFeve0251:
> It'll be easier for us to maintain or even review if u changed those two anonymous functions to normal methods in the class. You can also simplify things by creating a new method to return the filtered `pseudo_elements_to_remove`
> 
> what do u think?

The second one needs the context I think that using  normal function will make it confusing as we will have to use property which is not natural.

engahmeds3ed:
No need to create a property, I meant something like that (that's a draft):
```
    private function get_pseudo_elements_to_remove() {
	    $original_pseudo_elements = [
		    ':before',
		    ':after',
		    ':first-line',
		    ':first-letter',
		    ':active',
		    ':hover',
		    ':focus',
		    ':visited',
		    ':focus-within',
		    ':focus-visible',
	    ];

	    /**
	     * Pseudo elements to remove from lazyload CSS selector.
	     *
	     * @param string[] Pseudo elements to remove.
	     */
	    $pseudo_elements_to_remove = apply_filters( 'rocket_lazyload_css_ignored_pseudo_elements', $original_pseudo_elements );

	    if ( ! is_array( $original_pseudo_elements ) ) {
		    $pseudo_elements_to_remove = $original_pseudo_elements;
	    }

	    usort(
		    $pseudo_elements_to_remove,
		    static function ( $first, $second ) {
			    if ( strlen( $first ) === strlen( $second ) ) {
				    return 0;
			    }
			    return ( strlen( $first ) > strlen( $second ) ) ? -1 : 1;
		    }
	    );

        return $pseudo_elements_to_remove;
    }

    public function remove_pseudo_classes_for_selector( $selector ) {
	    $selector = preg_replace( '/::[\w-]+/', '', $selector );

	    $pseudo_elements_to_remove = $this->get_pseudo_elements_to_remove();
	    foreach ( $pseudo_elements_to_remove as $element ) {
		    $selector = str_replace( $element, '', $selector );
	    }
	    if ( in_array( substr( $selector, -1 ), [ '&', '~', '+', '>' ], true ) ) {
		    $selector .= '*';
	    }

	    if ( ! $selector ) {
		    return 'body';
	    }

	    return $selector;
    }

	protected function remove_pseudo_classes( string $selector ): string {
		$result = preg_replace( '/::[\w\-]+/', '', $selector );



		$selectors = explode( ',', $selector );

		$selectors = array_map( array( $this, 'remove_pseudo_classes_for_selector' ), $selectors );

		$selectors = array_unique( $selectors );

		$selector = implode( ',', $selectors );

		if ( ! $selector ) {
			return 'body';
		}

		return (string) $selector;
	}
```

engahmeds3ed (approved):

vmanthos (approved):
The PR is working. :+1: 

**NOTE**

#6158 was discovered during the investigation of this feature.
We'll need to re-test pseudo classes when that's fixed.

PR #6189, opened 2023-09-26
Closes #6152: Add license type to what is reported to Mixpanel

### LINKED ISSUE #6152: Add license type to what is reported to Mixpanel
This will help see how representative is the current sample from opted-in users vs our current users and how reliable data collected from them to help with product decisions.

Miraeld:
## Description

This will help see how representative is the current sample from opted-in users vs our current users and how reliable data collected from them to help with product decisions.

A new function has been created that returns the type of license. As this functionality is used in different files, it has been modified to use the newly created function.

Fixes #6152 

## Type of change

Yes

# Checklists

## Test summary


inc/functions/admin.php
```diff
@@ -526,3 +526,33 @@ function rocket_create_options_hash( $value ) {
...
 	return md5( wp_json_encode( $value_diff ) );
 }
+
+/**
+ * This function returns the license type for a customer.
+ *
+ * @param object $customer_data
+ * @return string
+ */
+function get_license_type($customer_data) {
```
engahmeds3ed: You are in the global namespace, you need to add a prefix here to be `rocket_get_license_key`

engahmeds3ed (changes requested):
Good job @Miraeld in ur first PR.

I didn't fully review it yet but I noticed this important point.

Also plz run the command
```
composer phpcs:fix
```

to automatically fix most of those reported coding standard issues.

inc/functions/admin.php
Miraeld: Oh alright, didn't know about that, I'm going to add it :)

Miraeld: Alright, I've changed it :)

remyperona (changes requested):
The code needs to be updated to follow our coding standards (errors are available in the CS GH action)

inc/functions/admin.php
```diff
@@ -526,3 +526,31 @@ function rocket_create_options_hash( $value ) {
...
+	if ( false === $customer_data
+		|| ! isset( $customer_data->licence_account )
+	) {
+		return __( 'Unavailable', 'rocket' );
+	}
+
+	if ( 1 <= $customer_data->licence_account
+		&& $customer_data->licence_account < 3
+	) {
+		$type = 'Single';
```
engahmeds3ed:
```suggestion
		return 'Single';
```

You can return directly and get rid of any else.

inc/functions/admin.php
```diff
@@ -526,3 +526,31 @@ function rocket_create_options_hash( $value ) {
...
+ *
+ * @param object $customer_data customer data as an object.
+ * @return string
+ */
+function rocket_get_license_type( $customer_data ) {
+	$type = '';
+
+	if ( false === $customer_data
+		|| ! isset( $customer_data->licence_account )
+	) {
```
engahmeds3ed: Let's make it more readable.

```suggestion
	if (
		false === $customer_data
		||
		! isset( $customer_data->licence_account )
	) {
```

inc/functions/admin.php
```diff
@@ -526,3 +526,31 @@ function rocket_create_options_hash( $value ) {
...
+		return __( 'Unavailable', 'rocket' );
+	}
+
+	if ( 1 <= $customer_data->licence_account
+		&& $customer_data->licence_account < 3
+	) {
```
engahmeds3ed:
```suggestion
	if (
		1 <= $customer_data->licence_account
		&&
		3 > $customer_data->licence_account
	) {
```

engahmeds3ed (changes requested):
Minor adjustments

inc/functions/admin.php
Miraeld: I've corrected it.

inc/functions/admin.php
Miraeld: I've corrected it.

inc/functions/admin.php
Miraeld: I've corrected it.

CrochetFeve0251 (approved):

vmanthos:
@Miraeld Thank you for the PR! 
I see the `license_type` is added to the data sent to Mixpanel. :+1: 

@wp-media/productrocket The new info that will be collected is not listed here:
image

Also, the CTA (Activate Rocket Analytics) is displayed even when Rocket Analytics is enabled.

Please confirm (or not) that:
1. The license type info should be included in the data collection table.
2. The CTA should not be displayed if Rocket Analytics is already enabled.

DahmaniAdame:
> The license type info should be included in the data collection table.

Yes, it should show. 

> The CTA should not be displayed if Rocket Analytics is already enabled.

No, there is no need to redisplay it if the consent is already given.

remyperona (approved):

vmanthos:
@Miraeld Thank you for the fixes.

The CTA button is not displayed but with that, the following text is hidden as well:
image

Since that's a disclaimer notice it should stay in place.

Also, and this is a late notice on my part, there is inconsistency in how various info is displayed as far as the used fonts go:
image

This has to do with some of it being wrapped in `<code></code>` and that results in the use of a monospace font.

Unless there is a specific reason to have some part in `<code>`, maybe we can go with a uniform look there.

@DahmaniAdame What's your take on this?

DahmaniAdame:
@vmanthos yes, uniformizing it is the way to go. 
`code` looks nicer, but not consistent and practical (doesn't look good when lines are wrapped). 
Uniformizing it without the `code` would make sense.

Miraeld:
@vmanthos I've corrected the cta, now only the button is not displayed once rocket analytics is enabled. 

@vmanthos @DahmaniAdame I've taken off the `<code></code>` tags to standardize the formatting.

vmanthos (approved):
Thank you for the fixes, @Miraeld ! 

PR #6197, opened 2023-09-28
Closes #5279: WP Rockets picks up wrong URL as referrer while using Weglot

### LINKED ISSUE #5279: WP Rocket picks up the wrong URL to purge/use a referrer when Weglot is used
- [x] Made sure you’re on the latest version `3.11.4.2`

**Describe the bug**
When [Weglot](https://wordpress.org/plugins/weglot/) is used, and the user is on a translated page, we are picking up the original URL to `Purge this URL` and `Clear Used CSS of this URL` as well as their respective referrer. 

We are purging the original URL and redirecting to it when clicking the action button. The translated page is not purged.

**To Reproduce**
1. Use Weglot
2. Translate a URL
3. Go to the translated page URL

**Expected behavior**
WP Rocket should pick the right URLs/referrers for individual pages when Weglot is used.

**Screenshots**
N/A

**Additional context**
Ticket - https://secure.helpscout.net/conversation/1954490982/357041/

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

Miraeld:
## Description

The referrer URL was wrong while using Weglot plugin when we want to clear cache or purge the page through the admin bar of wp-rocket.

Fixes #5279 

## Type of change

A Third Party class has been added and we added a filter to modify the referrer url with weglot language.
I have been using weglot functions to retrieve their `Request URL` object which allows me to get the full url of the page, and their own function to get the uri of that url.

# Checklists

## Test summary


inc/ThirdParty/Plugins/I18n/Weglot.php
```diff
@@ -0,0 +1,48 @@
...
+		$parsed_url = wp_parse_url( $referer );
+		// Append the language as a prefix to the path.
+		$new_path = user_trailingslashit( '/' . $current_language . $parsed_url['path'] );
+
+		if ( isset( $parsed_url['query'] ) ) {
+			// If there are query parameters, append them to the modified path.
+			return $new_path . '?' . $parsed_url['query'];
+		}
+
+		return $new_path;
```
CrochetFeve0251: @Miraeld seems like there is also a way to translation using domains: https://changelog.weglot.com/new-translation-management-by-url-customize-your-subdomains-and-url-slug-translation-update-ZEO2I
We may check against that to be sure it is working properly

CrochetFeve0251 (approved):

inc/ThirdParty/Plugins/I18n/Weglot.php
CrochetFeve0251: @Miraeld it is fine that functionnality is not present on WordPress.

inc/ThirdParty/Plugins/I18n/Weglot.php
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	/**
+	 * Modify the referer URL by appending the current language from Weglot as a prefix to the URL path.
+	 *
+	 * @param string $referer The original referer URL.
+	 * @return string The modified referer URL with the language as a prefix.
+	 */
+	public function add_langs_to_referer( $referer ) {
+		$current_language = weglot_get_current_language();
```
remyperona: We should add a `function_exists()` check here, in case this function is not loaded for some reason or it's removed later, that would avoid any unexpected error.

inc/common/admin-bar.php
```diff
@@ -16,7 +16,8 @@ function rocket_admin_bar( $wp_admin_bar ) {
 	global $pagenow, $post;
 
 	if ( ! empty( $_SERVER['REQUEST_URI'] ) ) {
-		$referer = filter_var( wp_unslash( $_SERVER['REQUEST_URI'] ), FILTER_SANITIZE_URL );
+		$uri     = filter_var( wp_unslash( $_SERVER['REQUEST_URI'], FILTER_SANITIZE_URL ) );
+		$referer = (string) apply_filters( 'rocket_admin_bar_referer', esc_url( $uri ) );
```
remyperona, line 25: Add a docblock for the filter

remyperona (changes requested):

inc/ThirdParty/Plugins/I18n/Weglot.php
Miraeld: True, I've added it

inc/common/admin-bar.php
Miraeld: I've added a docblock as well

remyperona (approved):

Mai-Saad:
@Miraeld Thanks for the PR. Please find exploratory test notes below :
1- A manual visit for the translated URL **won't add an entry in the cache table** although the page exists in the cache folder (while preload is on/off), however, if we click purge this URL from page edit, the entry will be added but **will stuck in progress till fail** (although URL exists in cache folder) => on trunk, manual visit is not added to cache table as well, while purge this URL won't add entry in table as well @piotrbak what do you think?
2- While RUCSS is on, clicking clear used CSS for this URL of translated URL will redirect to the main lang URL and clear cache of the main lang not translated one @Miraeld can you please check this

**Notes:**
- sometimes even while WPR is off, accessing the translated URL will auto-redirect to the main language URL (although redirect is disabled at Weglot) => those are not the URLs mentioned above
- Clear this cache from the page list will clear the home page cache => same on 3.15.2 , previous versions and 3.11.5
- exploratory test done on newlabs env and details here https://docs.google.com/spreadsheets/d/1TvUPEJBG_3drFXMGabxZ9wNYyzVECjsHg4DxYNUcyuk/edit#gid=0

**Update: added new extra points**
- Clear this cache from the page list will clear the main language of the page and not the translated ones (already on trunk) @piotrbak what do you think
- Deleting post will delete only main language from cache (added comment here https://github.com/wp-media/wp-rocket/issues/6246#issuecomment-1791943030)
- Editing post will only edit main language (added comment here https://github.com/wp-media/wp-rocket/issues/6223#issuecomment-1791951150)

Miraeld:
A fix has been made to cover the issue 2. @Mai-Saad has mentioned. 
It should be fine now.

Mai-Saad:
@CrochetFeve0251 @Miraeld Can you please check the conflict here :pray:

Mai-Saad:
@Miraeld Thanks for the update. now clear used CSS is working correctly for the translated URL. However, can see that purging this URL and clearing the Used CSS of this URL are not working for the main URL. Can you please check :pray:

Miraeld:
@Mai-Saad Hello hello, thank you for your feedback. 

In fact, the reason why it wasn't working is because Weglot function returns a language code even if they don't use it in the slug. 

I found another solution that works in every case scenario, which means it should be working just fine now :)

CrochetFeve0251 (approved):

Mai-Saad:
Purge this URL and Clear used CSS for this URL is working fine now. (clears cache of only related language either main or translated) => [notes](https://github.com/wp-media/wp-rocket/pull/6197#issuecomment-1761226900) above are updated

PR #6250, opened 2023-10-26
Changed on_submit to pending

### LINKED ISSUE #6242: RUCSS submitted_at field is incorrect
Context
Follow-up on this https://github.com/wp-media/wp-rocket/issues/6204 which has a bug.
Submitted_at should be the timestamp at which we send the job to the SaaS. So, it should be set to null when putting the job in "to-submit" and set to current timestamp when putting the job "pending".

**Bug reproduction**
Start a lot of  RUCSS jobs from the plugin and limit the number of pending jobs to a small number. Some jobs will remain in to-submit for a while but they already have a submitted_at date.
**How to fix**
Set `submitted_at` to `null` in `create_new_job` and `reset_job`. Set it to current timestamp in `make_status_pending`

CrochetFeve0251:
## Description
Fix an issue with the on submit field that should be only updated when the request is sent to the SAAS.

Fixes #6242
## Type of change

# Checklists

## Test summary

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Database/Queries/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Database/Queries/UsedCSS.php
@@ -597,11 +597,11 @@ class UsedCSS extends Query {
 		return $this->update_item(
 			$id,
 			[
-				'job_id'     => $job_id,
-				'queue_name' => $queue_name,
-				'status'     => 'pending',
-				'is_mobile'  => $is_mobile,
-				'submitted_at'  => current_time( 'mysql', true ),
+				'job_id'       => $job_id,
+				'queue_name'   => $queue_name,
+				'status'       => 'pending',
+				'is_mobile'    => $is_mobile,
+				'submitted_at' => current_time( 'mysql', true ),
 			]
 		);
 	}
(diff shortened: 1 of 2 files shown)


remyperona (approved):

vmanthos:
@CrochetFeve0251 @MathieuLamiot 

The timestamp when the status is `to-submit` for the first time, i.e. a new job is created, is `0000-00-00 00:00:00` and that correctly changes to a valid timestamp when the status becomes `pending`.

In the case of a `408` error code, most likely with retries as well, we reset the status to `to-submit` but the timestamp isn't reset. It will be when the status is changed to `pending`.

Should this behavior change, i.e. should the timestamp be reset when changing the status to `to-submit`?

MathieuLamiot:
Thanks for the detailed observation @vmanthos. I think the current behavior is good: in any case, we know when we sent a job for the last time. @CrochetFeve0251, do you agree with that?

CrochetFeve0251:
> Thanks for the detailed observation @vmanthos. I think the current behavior is good: in any case, we know when we sent a job for the last time. @CrochetFeve0251, do you agree with that?

@MathieuLamiot I would keep it like that cause I think it would make it easier to debug it as we will track that job easier. What do you think @vmanthos ?

Otherwise resetting it isn't a big deal either.

vmanthos (approved):
The purpose is served so we shouldn't complicate this more. 😄

PR #6256, opened 2023-10-31
Closes #6222: Revamp retry process rucss

### LINKED ISSUE #6222: Rework RUCSS retry strategy
**Context**
The current retry strategy is as follows in `check_job_status`, supposed to be triggered every minute:
- Retry up to 3 times in case we don't have a 200 Response with a result.
- In case of 408, we create submit a new job, and replace the job ID in the DB. Number of retries does not reset.

This has several issues: 
- We store jobs longer than 3 minutes in the SaaS and we often don't take advantage of this.
- There might be some queue on the SaaS, delaying the job for more than 3 minutes, at peak hours.
- If a 408 happens on the 2nd try, the new job won't have many chances.

**Expected behavior**
Rework the behavior in case the SaaS does not return a job result so that:
1) Have a different retry behavior if the job has been found but has no result yet (not processed yet) or if it was not found on the SaaS.
2) DEFAULT BEHAVIOR: In case the job has been found but with no results (400) or if the request to the server times out, the plugin must retry for longer than 3 minutes after the job has been sent. This timing must be filterable, but a good target would be up to 1 hour with exponential back-off.
3) In case the job has not been found on the SaaS (404), the job must be set to Fail without further retries.
4) In case of 408, a new job must be created. The retry strategy must be reset.
5) In case of 422 (job result invalid format), the job must be set to Fail without further retries.
6) In case of 500 or 401, the job must be set to Fail without further retries.
7) In any other case, the job must follow the same retry strategy as for a 400 (this is the default behavior).

In case of timeout during the request from the plugin to the server, we follow the same strategy as with a 400. A dedicated error code and error message should be handled on the plugin side (maybe 504 Server Timeout ?)

**Suggested solutions**
For point 1, we should leverage the details provided [here](https://github.com/wp-media/nodejs-treeshaker/issues/450).
For point 2, the ideal approach would be an exponential backoff: Try after 1 minute, then 2 minutes, then 5, then 10, then 10 for instance. Another quick&dirty approach could be to increase the number of retries. Both should be groomed to decide.

**EDIT**
After discussions, we will go with the exponential strategy implementation and distinct error codes.

**Acceptance Criteria**
- Non-regression in the case of successful job (code 200): the row must be set to successful.
- Submit a new job and force the job fetch to return a 400: the plugin must retry with exponential back-off (increasing delays between fetch tries), increment retries every time, log the error in error messages, and keep trying with a pending/in-progress state for 1 hour. After 1 hour since the job submission, the row must be set to failed.
- Submit a new job and force the job fetch to timeout: the plugin must retry with exponential back-off (increasing delays between fetch tries), increment retries every time, log the error in error messages, and keep trying with a pending/in-progress state for 1 hour. After 1 hour since the job submission, the row must be set to failed.
- Submit a new job and force the job fetch to return a 404: the plugin must fetch the results once and put the job to fail ultimately.
- Submit a new job and force the job fetch to return a 408: the row must be set to 'to-submit' so a new job is sent.
- Submit a new job and force the job fetch to return a 422: the plugin must fetch the results once and put the job to fail ultimately.
- Submit a new job a force the first job fetch to return a 400 or a timeout, then the next one to return a 200 with correct results. The plugin must therefore retry once and put the row as successful on the 2nd try.
NOTE: There is a known issue on the SaaS side for the Unauthorized '401') case. It will be considered like a 400 by the plugin. This use-case is to be discarded for now when testing the plugin.

Miraeld:
## Description
We have changed the way retry job are processed for the rucss. We are now using a Strategy Pattern which will allow us in the future to be more flexible and also to keep it simple to maintain.

Fixes #6222 

## Type of change

# Checklists

## Test summary


inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php
```diff
@@ -0,0 +1,90 @@
...
+		$this->used_css_query->increment_retries( $row_details->id, (int) $row_details->retries, $job_details['message'] );
+
+		$rucss_retry_duration = $this->time_table_retry[ $row_details->retries ] ?? 1800; // Default to 30 minutes.
+
+		/**
+		 * Filter used css retry duration.
+		 *
+		 * @param int $duration Duration between each retry in seconds.
+		 */
+		$rucss_retry_duration = apply_filters( 'rocket_rucss_retry_duration', $rucss_retry_duration );
```
MathieuLamiot: To give even more flexibility, instead of filtering rucss_retry_duration, maybe we could filter $this->time_table_retry directly at the beginning of this method?
It would allow to change the duration and the number of retries as well?

inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php
```diff
@@ -0,0 +1,90 @@
...
+	 */
+	protected $clock;
+
+	/**
+	 * Represents a timetable which shows how long to wait after for a new retry depending on how many retries have been made already.
+	 *
+	 * @var int[]
+	 */
+	private $time_table_retry = [
+		1 => 180,  // 3 minutes
```
MathieuLamiot: I think we should keep a behavior close to the current one at the beginning (so 1 check every minute). So maybe:
1 => 60,
2 => 120,
3 => 300,
4 => 600,
5 => 1200
6 => 1800
?
(cc @piotrbak )

inc/Engine/Optimization/RUCSS/Database/Schemas/UsedCSS.php
```diff
@@ -164,5 +164,15 @@ class UsedCSS extends Schema {
 			'date_query' => true,
 			'sortable'   => true,
 		],
+
+		// NOT_PROCEED_BEFORE column.
+		[
+			'name'       => 'not_proceed_before',
+			'type'       => 'timestamp',
+			'default'    => null,
```
MathieuLamiot: This comment might not be relevant: Just in case, maybe it would be safer to default to 0 so that we're sure not to have error when evaluating` not_proceed_before < current_time`? OR maybe PHP is OK with `null < current_time`?

MathieuLamiot (commented):
That looks really good. This architecture will be much easier to maintain.
I commented a few things, everything might not be relevant

inc/Engine/Optimization/RUCSS/Database/Schemas/UsedCSS.php
Miraeld: PHP est ok avec `null < (int)`

inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php
mostafa-hisham: I think we should leave this filter `rucss_retry_duration` it is simpler for clients to use
and we can add another filter to modify the `time_table_retry` in the constructor but it can be used to increase the retry number to be more the 6 times

inc/Engine/Optimization/RUCSS/Database/Schemas/UsedCSS.php
```diff
@@ -164,5 +164,15 @@ class UsedCSS extends Schema {
 			'date_query' => true,
 			'sortable'   => true,
 		],
+
+		// NOT_PROCEED_BEFORE column.
+		[
+			'name'       => 'not_proceed_before',
```
mostafa-hisham: can we change the column name?
when I first read it I thought it would be a boolean I couldn't understand what it does
we can rename it to `next_process_time` or `next_retry_time`

inc/Engine/Optimization/RUCSS/Database/Queries/UsedCSS.php
```diff
@@ -578,6 +579,46 @@ private function table_exists(): bool {
...
+
+	/**
+	 * Updates the not_processed_before field
+	 *
+	 * @param mixed      $job_id the job id.
+	 * @param string|int $not_processed_before timestamp or mysql format date.
+	 *
+	 * @return bool either it is save or not.
+	 */
+	public function update_not_processed_before( $job_id, $not_processed_before ): bool {
```
mostafa-hisham: field name called `not_proceed_before` and the function is `update_not_processed_before` 

I think we should rename them :sweat_smile:

inc/Engine/Optimization/RUCSS/Strategy/Strategies/ResetRetryProcess.php
```diff
@@ -0,0 +1,43 @@
...
+	 *
+	 * @return void
+	 */
+	public function execute( object $row_details, array $job_details ): void {
+		$used_css_row = $this->used_css_query->get_row( $row_details->url, (bool) $row_details->is_mobile );
+		if ( empty( $used_css_row ) ) {
+			$this->used_css_query->create_new_job( $row_details->url, '', '', $row_details->is_mobile );
+			return;
+		}
+		$this->used_css_query->reset_job( (int) $used_css_row->id );
```
mostafa-hisham, line 41: @MathieuLamiot 
To avoid the loop problem that we had with 408 
we can just create a new job and only do the reset one the first time we get 408 from RUCSS
if the last `error_code!= 408` i think this can work

inc/Engine/Optimization/RUCSS/Strategy/Factory/StrategyFactory.php
```diff
@@ -0,0 +1,77 @@
...
+		// $this->logger::debug( 'RUCSS: Job status failed for url: ' . $row_details->url, $job_details );
+
+		switch ( $job_details['code'] ) {
+			case 408:
+				// $this->logger::debug('Strategy Factory -> Error 408 -> ResetRetryProcess');
+				$strategy = new ResetRetryProcess( $this->used_css_query );
+				break;
+			case 500:
+			case 422:
+			case 404:
```
mostafa-hisham, line 60: @MathieuLamiot 
do we need to handle the 401 case when we get the job details back too?

we only do it in `add_to_queue`
https://github.com/wp-media/wp-rocket/blob/a23c147b38baeb51d6a0039e190dc514ec994605/inc/Engine/Optimization/RUCSS/Frontend/APIClient.php#L59

inc/Engine/Common/Clock/ClockInterface.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace WP_Rocket\Engine\Common\Clock;
+
+interface ClockInterface {
```
wordpressfan, line 5: We are over complicate the code here, we can easily use the WP Core function `current_time`

inc/Engine/Optimization/RUCSS/Database/Queries/UsedCSS.php
```diff
@@ -578,6 +579,46 @@ private function table_exists(): bool {
...
+		if ( is_numeric( $not_processed_before ) ) {
+			$not_processed_before = gmdate( 'Y-m-d H:i:s', $not_processed_before );
```
wordpressfan: I'm not sure what is the exact type of this column, is it timestamp or a date?
Because below in the row class we are converting it using strtotime and here we are changing it to date and then in `DefaultProcess` class we are dealing with it as timestamp, I believe we can easily keep all of them as a timestamp, what do u think?

inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php
```diff
@@ -42,6 +42,7 @@ class UsedCSS extends Table {
 		20220920 => 'make_status_column_index_instead_queue_name',
 		20221104 => 'add_error_columns',
 		20231010 => 'add_submitted_at_column',
+		20231031 => 'add_not_proceed_before_column',
```
wordpressfan: Plz change the $version above to make the DB changes happen

inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php
```diff
@@ -60,21 +61,22 @@ public function __construct() {
...
+			unprocessedcss   		longtext                NULL,
+			retries          		tinyint(1)          NOT NULL default 1,
+			is_mobile        		tinyint(1)          NOT NULL default 0,
+			job_id           		varchar(255)        NOT NULL default '',
+			queue_name       		varchar(255)        NOT NULL default '',
+			status           		varchar(255)        NOT NULL default '',
+			modified         		timestamp           NOT NULL default '0000-00-00 00:00:00',
+			last_accessed    		timestamp           NOT NULL default '0000-00-00 00:00:00',
+			submitted_at     		timestamp           NULL,
+			not_proceed_before     	timestamp           NULL,
```
wordpressfan: `not_proceed_before` or `not_processed_before` ?

wordpressfan (changes requested):
Great jpb @Miraeld 
Still checking the strategy part.

inc/Engine/Optimization/RUCSS/Strategy/Factory/StrategyFactory.php
```diff
@@ -0,0 +1,77 @@
...
+				$strategy = new JobSetFail( $this->used_css_query );
+				break;
+			default:
+				$strategy = new DefaultProcess( $this->used_css_query, $this->clock );
+
+				// $this->logger::debug('Strategy Factory -> Other -> DefaultProcess');
```
wordpressfan: we need to remove this log

inc/Engine/Optimization/RUCSS/Strategy/Factory/StrategyFactory.php
```diff
@@ -0,0 +1,77 @@
...
+	 * Manage the whole process, to determine which strategy to adopt..
+	 *
+	 * @param object $row_details DB Row of a job.
+	 * @param array  $job_details Job information from the API.
+	 *
+	 * @return void
+	 */
+	public function manage( $row_details, $job_details ): void {
+
+		// $this->logger::debug( 'RUCSS: Job status failed for url: ' . $row_details->url, $job_details );
```
wordpressfan: remove this log

inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php
```diff
@@ -0,0 +1,90 @@
...
+	 * @var WPRClock
+	 */
+	protected $clock;
+
+	/**
+	 * Represents a timetable which shows how long to wait after for a new retry depending on how many retries have been made already.
+	 *
+	 * @var int[]
+	 */
+	private $time_table_retry = [
```
wordpressfan, line 31: for simplicity sake, we don't need the keys here

inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php
```diff
@@ -0,0 +1,90 @@
...
+
+		/**
+		 * Filter used css retry duration.
+		 *
+		 * @param int $duration Duration between each retry in seconds.
+		 */
+		$rucss_retry_duration = apply_filters( 'rocket_rucss_retry_duration', $rucss_retry_duration );
+
+		// update the `not_proceed_before` column.
+		$not_proceed_before = $this->clock->current_time( 'timestamp' ) + $rucss_retry_duration;
```
wordpressfan: We are expecting $rucss_retry_duration after filter to be integer value so plz we need to guard against that.

wordpressfan (changes requested):

inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php
Miraeld: I've kept the filter I've created at the beginning, and I've added a new one `rocket_rucss_retry_table` to let people change that array if needed, like @mostafa-hisham mentioned.

inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php
Miraeld: I think too. I've based my initial numbers with the brief of the issue. I've adapted it back to the numbers you just give.

inc/Engine/Optimization/RUCSS/Database/Schemas/UsedCSS.php
Miraeld: I've changed the column name to `next_retry_time`.

inc/Engine/Optimization/RUCSS/Database/Queries/UsedCSS.php
Miraeld: Goes with the last comment, I've changed the column name to `next_retry_time` so this function becomes `update_next_retry_time`

inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php
Miraeld: I've done it, I always forget it ...

inc/Engine/Optimization/RUCSS/Database/Tables/UsedCSS.php
Miraeld: It's been changed to `next_retry_time`

inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php
Miraeld: Keys are important, as they are used here : https://github.com/wp-media/wp-rocket/blob/45b92225b528e2bf97d61aa239728a7fc0312077/inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php#L76

In fact, they allow us to know for the retry #x how long we have to wait for the next one.

inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php
Miraeld: I've added a `(int)` to the `apply_filters()` to make sure we are getting an int.

inc/Engine/Optimization/RUCSS/Strategy/Factory/StrategyFactory.php
Miraeld: I've deleted them :)

inc/Engine/Optimization/RUCSS/Strategy/Factory/StrategyFactory.php
Miraeld: I've deleted them :)

inc/Engine/Optimization/RUCSS/Database/Queries/UsedCSS.php
Miraeld: I've changed this function to save it as a timestamp, as it is a timestamp in the database.

inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php
wordpressfan: OK that's a minor thing, what I meant is to use it to change the line u mentioned too, to adapt with it but let's keep it as is.
Thanks for looking into that.

inc/Engine/Optimization/RUCSS/Strategy/Factory/StrategyFactory.php
MathieuLamiot: You are right. It would be much cleaner to fail immediately with it here. @Miraeld, 401 would behave just like 500, 422 and 404: failed job. I'll updaet the ticket description

inc/Engine/Optimization/RUCSS/Strategy/Strategies/ResetRetryProcess.php
MathieuLamiot: @mostafa-hisham I agree this could be a valid approach. Let's keep this one as it is for now, I'll open another issue to improve this later on. Thanks for the idea!

inc/Engine/Optimization/RUCSS/Strategy/Factory/StrategyFactory.php
mostafa-hisham: @MathieuLamiot this will require a change on RUCSS 
- send response in `returnvalue`

MathieuLamiot: @mostafa-hisham, issue created [here](https://github.com/wp-media/nodejs-treeshaker/issues/520) for the SaaS part -> not a priority. The plugin part should still be done I believe so that it is ready.

inc/Engine/Common/Clock/ClockInterface.php
wordpressfan: May I get your point of view here @Tabrisrp ?

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -549,40 +574,10 @@ public function check_job_status( int $id ) {
 		}
 
 		if (
-			200 !== (int) $job_details['code']
-			||
-			empty( $job_details['contents'] )
-			||
-			! isset( $job_details['contents']['shakedCSS'] )
+			200 !== $job_details['code']
```
vmanthos: `$job_details['code']` was typecasted to `int` to resolve an error when the value was not an `int`:
https://github.com/wp-media/wp-rocket/pull/6257/commits/24982823e54cf576505cfc70806952bd8a115762

Is there a reason this was removed here?

inc/Engine/Common/Clock/ClockInterface.php
remyperona: This doesn't seem necessary indeed, there is already time management included in WP core

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
Miraeld: No reason, I didn't pay attention, my bad, I've added it back.

inc/Engine/Optimization/RUCSS/Strategy/Factory/StrategyFactory.php
Miraeld: I've added the 401 to the strategy.

wordpressfan (approved):
Approved after a discussion here: https://wp-media.slack.com/archives/CUT7FLHF1/p1700491887206929

vmanthos:
@Miraeld I'm getting the following warning in the `debug.log`:
```PHP
PHP Warning:  Undefined property: stdClass::$next_retry_time in /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php on line 526
```
To reproduce:
1. Clear the used CSS.
2. Check the `debug.log`.

vmanthos:
@Miraeld The timestamp for the default strategy is added as ` 0000-00-00 00:00:00` so the retries are taking place without following the `$time_table_retry`.

image


Can you please take a look? :pray:

vmanthos:
Thank you for the fixes @Miraeld. :pray: 

Please check the following.

- [ ] I got the following PHP fatal error in the debug.log:

```PHP
PHP Fatal error:  Uncaught TypeError: WP_Rocket\Engine\Optimization\RUCSS\Controller\UsedCSS::__construct(): Argument #9 ($strategy_factory) must be of type WP_Rocket\Engine\Optimization\RUCSS\Strategy\Factory\StrategyFactory, string given in /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php:130
Stack trace:
#0 [internal function]: WP_Rocket\Engine\Optimization\RUCSS\Controller\UsedCSS->__construct()
#1 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Dependencies/League/Container/Definition/Definition.php(256): ReflectionClass->newInstanceArgs()
#2 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Dependencies/League/Container/Definition/Definition.php(212): WP_Rocket\Dependencies\League\Container\Definition\Definition->resolveClass()
#3 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Dependencies/League/Container/Definition/DefinitionAggregate.php(94): WP_Rocket\Dependencies\League\Container\Definition\Definition->resolve()
#4 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Dependencies/League/Container/Container.php(157): WP_Rocket\Dependencies\League\Container\Definition\DefinitionAggregate->resolve()
#5 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/ServiceProvider.php(125): WP_Rocket\Dependencies\League\Container\Container->get()
#6 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Dependencies/League/Container/ServiceProvider/ServiceProviderAggregate.php(102): WP_Rocket\Engine\Optimization\RUCSS\ServiceProvider->register()
#7 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Dependencies/League/Container/Container.php(172): WP_Rocket\Dependencies\League\Container\ServiceProvider\ServiceProviderAggregate->register()
#8 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/ThirdParty/Themes/ServiceProvider.php(53): WP_Rocket\Dependencies\League\Container\Container->get()
#9 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Dependencies/League/Container/ServiceProvider/ServiceProviderAggregate.php(102): WP_Rocket\ThirdParty\Themes\ServiceProvider->register()
#10 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Dependencies/League/Container/Container.php(172): WP_Rocket\Dependencies\League\Container\ServiceProvider\ServiceProviderAggregate->register()
#11 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Plugin.php(155): WP_Rocket\Dependencies\League\Container\Container->get()
#12 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/main.php(42): WP_Rocket\Plugin->load()
#13 /var/www/example.com/htdocs/wp-includes/class-wp-hook.php(324): rocket_init()
#14 /var/www/example.com/htdocs/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters()
#15 /var/www/example.com/htdocs/wp-includes/plugin.php(517): WP_Hook->do_action()
#16 /var/www/example.com/htdocs/wp-settings.php(506): do_action()
#17 /var/www/example.com/wp-config.php(161): require_once('...')
#18 /var/www/example.com/htdocs/wp-load.php(55): require_once('...')
#19 /var/www/example.com/htdocs/wp-blog-header.php(13): require_once('...')
#20 /var/www/example.com/htdocs/index.php(17): require('...')
#21 {main}
  thrown in /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php on line 130
``` 
I'm not sure of the exact context, i.e. I don't have a way to reproduce it. It may have happened when switching branches.

Is there anything we can do to prevent it?

- [ ] The following filter is missing a docblock:
[https://github.com/wp-media/wp-rocket/blob/5b993d7cd985678c9e8368dde95ef069734de93[…]ngine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.ph](https://github.com/wp-media/wp-rocket/blob/5b993d7cd985678c9e8368dde95ef069734de93a/inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php#L49)

- [ ] We need to safeguard the following filters from incorrect values as those will either break the feature (the first two) or alter the retries functionality (the last one):
https://github.com/wp-media/wp-rocket/blob/542ea1c5ce4b96367cc01319ef06c7c6e2856f66/inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php#L49
https://github.com/wp-media/wp-rocket/blob/5b993d7cd985678c9e8368dde95ef069734de93a/inc/Engine/Common/Clock/WPRClock.php#L28
https://github.com/wp-media/wp-rocket/blob/542ea1c5ce4b96367cc01319ef06c7c6e2856f66/inc/Engine/Optimization/RUCSS/Strategy/Strategies/DefaultProcess.php#L83

mostafa-hisham (approved):

inc/Engine/Common/Clock/WPRClock.php
```diff
@@ -0,0 +1,34 @@
...
+	 * @param int|bool $gmt  Optional. Whether to use GMT timezone. Default false.
+	 *
+	 * @return int|string Integer if `$type` is 'timestamp' or 'U', string otherwise.
+	 */
+	public function current_time( string $type, $gmt = 0 ) {
+		$current_time = current_time( $type, $gmt );
+		if ( ( is_string( $current_time ) && strtotime( $current_time ) ) || is_int( $current_time ) ) {
+			return apply_filters( 'rocket_current_time', $current_time );
+		}
+		return apply_filters( 'rocket_current_time', 0 );
```
wordpressfan: If I understand this part well, here we are checking if it's string (it can be string if gmt argument is not true) and it's a real date/datetime string by checking the return of strtotime (this returns false in case it's not valid value)
OR
WP Core function returns integer in case of gmt is true.
THEN
current_time is valid and apply the filter on this value.
OTHERWISE
apply the filter on the default value of zero.

But in our code we always use gmt=true so all what we do here is to make it more generic, correct?

I have three comments here:
1. This is a critical part which affects basic functionality of RUCSS and sending jobs so I believe we don't need filter here.
2. We need to define the filter only one time
3. it'd be easier to get the $gmt argument into the condition and typecast the filter return (if we really need the filter).

Miraeld: Safeguard of filters have been changed. I misunderstood how it works :)

vmanthos:
@Miraeld 

- [ ] When updating to WP Rocket `3.15.5` and for the URLs that are in the database with `status` of `to-submit` or `pending` a PHP fatal error is thrown:

```PHP
PHP Fatal error:  Uncaught TypeError: strtotime(): Argument #1 ($datetime) must be of type string, null given in /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php:526
Stack trace:
#0 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php(526): strtotime()
#1 /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/Cron/Subscriber.php(99): WP_Rocket\Engine\Optimization\RUCSS\Controller\UsedCSS->process_pending_jobs()
#2 /var/www/example.com/htdocs/wp-includes/class-wp-hook.php(324): WP_Rocket\Engine\Optimization\RUCSS\Cron\Subscriber->process_pending_jobs()
#3 /var/www/example.com/htdocs/wp-includes/class-wp-hook.php(348): WP_Hook->apply_filters()
#4 /var/www/example.com/htdocs/wp-includes/plugin.php(565): WP_Hook->do_action()
#5 /var/www/example.com/htdocs/wp-cron.php(191): do_action_ref_array()
#6 {main}
  thrown in /var/www/example.com/htdocs/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php on line 526
```

This is because following the update the existing rows get the value of `NULL` in the `next_retry_time` column.


- [ ] Also, if the `rocket_rucss_retry_duration` filter gets a boolean, the next retry time table will not be following the retry timetable. The interval between retries will be ~1 minute.

PR #6305, opened 2023-12-05
Closes #6278: Do not add URLs to the wpr_rocket_cache table if Preload is disabled

### LINKED ISSUE #6278: Do not add URLs to the wpr_rocket_cache table if Preload is disabled
- Made sure you’re on the latest version ✔️
- Used the search feature to ensure that the bug hasn’t been reported before ✔️

**Describe the bug**
Currently, any URL that is visited by a real user is added to the `wpr_rocket_cache` table and automatically assigned a “Complete” status even when the Preload is disabled. This causes the DB to grow too large in some cases.

**To Reproduce**
1. Make sure the Preload is disabled
2. Visit some URLs while not logged-in
3. Check the `wpr_rocket_cache` table in the DB and you'll find those visited URLs added to it with a "Complete" status

**Expected behavior**
The table itself is not necessary for the caching process to run effectively, so no URLs should be added to it when the Preload is disabled.

Temporary fix - The snippet below will prevent all URLs from being added to the table if the preload is disabled:

```
// prevents adding URL to the wpr_rocket_cache table if Preload is disabled
function stop_adding_top_preload( array $regexes ) : array {

	$new_regex = $_SERVER['REQUEST_URI'];
	
	$options = get_option('wp_rocket_settings', []);

	//only if preload is disabled
	if( $options['manual_preload'] == 0) {

		// add the currently visited URL to the exclusions array
		$regexes[] = $new_regex;

	}

return $regexes;
}

add_filter( 'rocket_preload_exclude_urls', 'stop_adding_top_preload');
```

**Screenshots**
N/A

**Additional context**
HS ticket: [https://secure.helpscout.net/conversation/2405513676/451397/](https://secure.helpscout.net/conversation/2405513676/451397/)
Slack thread: [https://wp-media.slack.com/archives/C08N8J6VC/p1698938940269529](https://wp-media.slack.com/archives/C08N8J6VC/p1698938940269529)

1. When preload is disabled, we should not add URLs to the cache table
2. When enabling preload, there should not be any regression with fetching the sitemaps
3. When the preload is enabled, URLs should be added to the table without regression

Miraeld:
## Description

Currently, any URL that is visited by a real user is added to the `wpr_rocket_cache` table and automatically assigned a “Complete” status even when the Preload is disabled. This causes the DB to grow too large in some cases.

Fixes #6278 

## Type of change

No

# Checklists

## Test summary

### DIFF
--- a/inc/Engine/Preload/Subscriber.php
+++ b/inc/Engine/Preload/Subscriber.php
@@ -449,13 +449,8 @@ class Subscriber implements Subscriber_Interface {
 	public function add_preload_excluded_uri( $regexes ): array {
 		$preload_excluded_uri = (array) $this->options->get( 'preload_excluded_uri', [] );
 
-		if ( ! empty( $_SERVER['REQUEST_URI'] ) ) {
-			$request_uri = filter_var( wp_unslash( $_SERVER['REQUEST_URI'] ), FILTER_SANITIZE_URL );
-
-			if ( (bool) ! get_rocket_option( 'manual_preload', true ) ) {
-				// add the currently visited URL to the exclusions array.
-				$regexes[] = $request_uri;
-			}
+		if ( ! empty( $_SERVER['REQUEST_URI'] ) && (bool) ! $this->options->get( 'manual_preload', true ) ) {
+			$regexes[] = filter_var( wp_unslash( $_SERVER['REQUEST_URI'] ), FILTER_SANITIZE_URL );
 		}
 
 		if ( empty( $preload_excluded_uri ) ) {
--- a/inc/Engine/Preload/Subscriber.php
+++ b/inc/Engine/Preload/Subscriber.php
@@ -189,6 +189,10 @@ class Subscriber implements Subscriber_Interface {
 			return;
 		}
 
+		if ( (bool) ! $this->options->get( 'manual_preload', true ) ) {
+			return; // Bail out if preload is disabled.
+		}
+
 		$url = home_url( add_query_arg( [], $wp->request ) );
 
 		$detected = $this->mobile_detect->isMobile() && ! $this->mobile_detect->isTablet() ? 'mobile' : 'desktop';
@@ -448,10 +452,6 @@ class Subscriber implements Subscriber_Interface {
 	 */
 	public function add_preload_excluded_uri( $regexes ): array {
 		$preload_excluded_uri = (array) $this->options->get( 'preload_excluded_uri', [] );
-
-		if ( ! empty( $_SERVER['REQUEST_URI'] ) && (bool) ! $this->options->get( 'manual_preload', true ) ) {
-			$regexes[] = filter_var( wp_unslash( $_SERVER['REQUEST_URI'] ), FILTER_SANITIZE_URL );
-		}
 
 		if ( empty( $preload_excluded_uri ) ) {
 			return $regexes;
--- a/inc/Engine/Preload/Subscriber.php
+++ b/inc/Engine/Preload/Subscriber.php
@@ -242,6 +242,10 @@ class Subscriber implements Subscriber_Interface {
 	public function on_permalink_changed() {
 		$this->query->remove_all();
 		$this->queue->cancel_pending_jobs();
+		if ( ! $this->options->get( 'manual_preload', false ) ) {
+			return;
+		}
+
 		$this->queue->add_job_preload_job_load_initial_sitemap_async();
 	}
 
--- a/inc/Engine/Preload/Subscriber.php
+++ b/inc/Engine/Preload/Subscriber.php
@@ -276,7 +276,9 @@ class Subscriber implements Subscriber_Interface {
 	 * @return void
 	 */
 	public function clean_url( string $url ) {
-
+		if ( ! $this->options->get( 'manual_preload', 0 ) ) {
+			return;
+		}
 		$this->clear_cache->partial_clean( [ $url ] );
 	}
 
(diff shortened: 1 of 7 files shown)


inc/Engine/Preload/Subscriber.php
```diff
@@ -449,6 +449,15 @@ public function unlock_url( string $url ) {
 	public function add_preload_excluded_uri( $regexes ): array {
 		$preload_excluded_uri = (array) $this->options->get( 'preload_excluded_uri', [] );
 
+		if ( ! empty( $_SERVER['REQUEST_URI'] ) ) {
+			$request_uri = filter_var( wp_unslash( $_SERVER['REQUEST_URI'] ), FILTER_SANITIZE_URL );
+
+			if ( (bool) ! get_rocket_option( 'manual_preload', true ) ) {
```
CrochetFeve0251: It would be better using the Option_Data class instead

CrochetFeve0251: Also it would be better creating a new function bailing out or bailing out instead of adding nested condtions

CrochetFeve0251 (changes requested):

Mai-Saad:
Related test plan https://wpmediaqa.testrail.io/index.php?/runs/view/792&group_by=cases:section_id&group_order=asc

inc/Engine/Preload/Subscriber.php
Miraeld: It's been changed :)

Miraeld: It's been changed to avoid nested `if` statements

CrochetFeve0251 (approved):

inc/Engine/Preload/Subscriber.php
```diff
@@ -449,6 +449,10 @@ public function unlock_url( string $url ) {
 	public function add_preload_excluded_uri( $regexes ): array {
 		$preload_excluded_uri = (array) $this->options->get( 'preload_excluded_uri', [] );
 
+		if ( ! empty( $_SERVER['REQUEST_URI'] ) && (bool) ! $this->options->get( 'manual_preload', true ) ) {
+			$regexes[] = filter_var( wp_unslash( $_SERVER['REQUEST_URI'] ), FILTER_SANITIZE_URL );
```
remyperona: Why are we adding the URL to the regexes when the option is disabled?

When the option is disabled, we should be doing nothing

remyperona (changes requested):
Can we add to this PR a check for all related preload methods that should do nothing when the option is disabled?

inc/Engine/Preload/Subscriber.php
Miraeld: That's true, I have changed it.

inc/Engine/Preload/Subscriber.php
```diff
@@ -447,6 +447,10 @@ public function unlock_url( string $url ) {
 	 * @return array|false
 	 */
 	public function add_preload_excluded_uri( $regexes ): array {
+		if ( (bool) ! $this->options->get( 'manual_preload', true ) ) {
+			return false;
```
CrochetFeve0251: @Miraeld return a boolean doesn't make sense here maybe the original value?

CrochetFeve0251 (approved):

inc/Engine/Preload/Subscriber.php
Miraeld: @CrochetFeve0251 If I believe the docblock, it returns false or an array. 
If you return the original value, like it was before, it would process the URL and add it to the database, which is what we want to avoid here.

Returning `false` means it won't be added to the database

CrochetFeve0251: @Miraeld if I read the PHP code not the comments I see it expects an array only. Docblocks are comments and can lie in this case I think PHPStorm generated it cause of `array_merge`.
Returning false means it won't be added to the database -> how it won't be added?

Miraeld: Umm I've changed the way to do it, bailing out before, reaching this function.

CrochetFeve0251 (approved):

remyperona (changes requested):
We could add a unit test for this method, I see it's missing currently. Good opportunity to improve the coverage.

Miraeld:
> We could add a unit test for this method, I see it's missing currently. Good opportunity to improve the coverage.

@Tabrisrp I've used an integration test to cover this case.

tests/Fixtures/inc/Engine/Preload/Subscriber/updateCacheRow.php
```diff
@@ -66,5 +69,23 @@
 				],
 			]
 		]
-	]
+	],
+	'testShouldBailOut' => [
```
remyperona:
```suggestion
	'testShouldBailOutWHenPreloadDisabled' => [
```

remyperona (approved):

Miraeld:
Testrails report can be find here : 
https://wpmediaqa.testrail.io/index.php?/reports/view/553

Mai-Saad:
@Miraeld Thanks for the PR and test execution.
@piotrbak Please find exploratory testing notes below:
1-  URLs are added to the cache table while preload is off, when:
1.1- Change permalinks
1.2- Add/edit/delete a post
Shall we handle this on the same PR? (Note: it won't be added when visiting the post (while it is not in the table) nor when switching themes.)

2- If the URL was already in the table as completed then we disabled preload, after that, we purge this URL or clear the cache, URL status will change to pending. is this ok? (note: in this case, the last modified time will be updated)

piotrbak:
@Mai-Saad 1.1 and 1.2 should be scope of this PR

Miraeld:
@Mai-Saad your point 1.1 & 1.2 are covered in the last commit :)

Mai-Saad:
@Miraeld thanks for the update. 1.2 is fixed :tada: , while 1.1 is not fixed on the test site. 

Note: if RUCSS is enabled, we will add URLs to the table while preload is off => agreed to remain as is for rucss scenario

CrochetFeve0251 (approved):

Miraeld:
@Mai-Saad Permalink are now totally okay. 
I've run some test through test rails and it everything seems okay on my part. :) 
[testrail-report-555.pdf](https://github.com/wp-media/wp-rocket/files/13792155/testrail-report-555.pdf)

PR #6342, opened 2023-12-19
Closes #6329: Deprecate RUCSS Hooks

### LINKED ISSUE #6329: Deprecate old hooks for RUCSS
Deprecate old hooks used by RUCSS.

jeawhanlee:
## Description

This PR deprecates the use of RUCSS hooks.

Fixes #6329 

# Checklists

## Test summary


Miraeld (approved):

CrochetFeve0251:
@Tabrisrp when I worked with @jeawhanlee on that seemed like the function was a good but after a while and remembering @piotrbak idea seems like it looked like the moment to create a class for filters and actions.

This would abstract the wiring between the old and new hooks at the level of the business logic and we would be able to add casting directly within the `apply_filters` method.

The idea is to create a Filter and an Action class.

Then in the deprecated file we reuse that classes to make the link between old and new hooks with something like (This is possible as deprecated logic is loaded [after the container is loaded](https://github.com/wp-media/wp-rocket/blob/06381b8860eb46e570911b2812652f54a676dd6d/inc/main.php#L68-L67)):
```php
$filter->add_deprecated('old_hook', 'new_hook');
```

And inside the Filter and Action class we add an abstract logic to make the links based on an array and call the deprecated filter function.So inside the main logic we would have:
```php
$result = $this->filter->apply_filter('new_hook', $param);
```

Concerning the casting issue we could create a method `apply_string_filter` , `apply_int_filter` for each base type and a method `apply_filter` with a callback for validation as second parameter so we would cover each cases.

What do you think @Tabrisrp, @piotrbak  ?

wordpressfan:
@CrochetFeve0251 I like that idea so much, but I believe we need to use those new classes gradually because we have so many apply_filters and do_action in our code base.

CrochetFeve0251:
> @CrochetFeve0251 I like that idea so much, but I believe we need to use those new classes gradually because we have so many apply_filters and do_action in our code base.

@wordpressfan I agree on that point.
My idea here was to apply for the new feature only.
Maybe then we could extend it to the rest of the plugin gradually when we need to make changes or during a dev initiative sprint.

I also wait @Tabrisrp so we can discuss about that when he is here. Feel free to join the conversation if you want.

MathieuLamiot:
The discussion is interesting and should indeed be tackled as part of tech debt reduction.
However, we can't block a such small PR for 8 days, especially if therr was already some sort of agreement upon the solution to use in the first place.
If the code is good enough to be delivered, but can be enhanced:
- validate the CR
- Keep the discussion going on a dedicated thread or issue

CrochetFeve0251:
> The discussion is interesting and should indeed be tackled as part of tech debt reduction. However, we can't block a such small PR for 8 days, especially if therr was already some sort of agreement upon the solution to use in the first place. If the code is good enough to be delivered, but can be enhanced:
> 
>     * validate the CR
> 
>     * Keep the discussion going on a dedicated thread or issue

@jeawhanlee I discussed about it with @MathieuLamiot and he proposed a solution that I agree on.
We can keep the code like that for the moment but we need to implement tests (I recommend you to implement Integration tests with [`did_action`](https://developer.wordpress.org/reference/functions/did_action/) to keep them as much as possible black box in case of any change) to check each legacy hooks are called when the logic is executed. 

Then when we have that tests we can open another issue to refactor the code later on.

jeawhanlee:
> > The discussion is interesting and should indeed be tackled as part of tech debt reduction. However, we can't block a such small PR for 8 days, especially if therr was already some sort of agreement upon the solution to use in the first place. If the code is good enough to be delivered, but can be enhanced:
> > ```
> > * validate the CR
> > 
> > * Keep the discussion going on a dedicated thread or issue
> > ```
> 
> @jeawhanlee I discussed about it with @MathieuLamiot and he proposed a solution that I agree on. We can keep the code like that for the moment but we need to implement tests (I recommend you to implement Integration tests with [`did_action`](https://developer.wordpress.org/reference/functions/did_action/) to keep them as much as possible black box in case of any change) to check each legacy hooks are called when the logic is executed.
> 
> Then when we have that tests we can open another issue to refactor the code later on.

@CrochetFeve0251 We can close this PR and I'll add the tests on this [branch](https://github.com/wp-media/wp-rocket/tree/add-tests-for-rucss-api-cron-refactor) since we already have it going there.

Ping @MathieuLamiot

MathieuLamiot:
@jeawhanlee Unless I am missing something, we should have code & test always together in the same PR/branch.
It does not make sense to add tests for deprecated hooks in a branch that does not have the deprecated hooks. Or you'll merge this branch into the "add-test" one and we'll have merges everywhere...

In any case, adding tests must be done. Let's keep it simple and add them in the correct branch, no?

remyperona (commented):
I don't like the naming of the deprecate filter/action functions, when reading the code I was confused thinking it was applying a deprecated filter/action only.

Is there really added value for this?

jeawhanlee:
@Tabrisrp I believe this is mainly for backward compatibility to still support old hooks as well as the new.

MathieuLamiot:
@Tabrisrp @jeawhanlee This was a request from product that the "old hooks" remain functional for a few versions, with a warning message. this is mostly to allow support team to spot the hooks they are using in helper plugins that they will need to adapt.

That being said, we can maybe just change the name of the methods if this is confusing. Maybe apply_filter_and_deprecated? WDYT?

remyperona:
keeping the old filter as deprecated makes sense, but I don't see much value of wrapping that into new functions. We don't have this kind of case that much in the plugin.

MathieuLamiot:
OK I understand. To move forward with this, I suggest:
- Keep it as it is currently.
- In a few minors (when product sees fit), we'll have to stop the deprecate message and truly remove the old hooks anyway. At that time, if the deprecation mechanism has not been used, we can just remove it as well. Otherwise, it means it is used from time to time.

remyperona:
sounds ok, could we just rename the functions so that the naming is clearer on what it does

remyperona (approved):

Mai-Saad:
@jeawhanlee Thanks for the PR (as far as I understand both old and new filters shall work)
- Shall we handle this?
'rocket_rucss_delete_interval' 
'rocket_rucss_deletion_enabled' 
- What about Cron? 'rocket_rucss_clean_rows_time_event'
- **Update: can see these too**, can you please check if we shall handle that too :pray: 
Filter: 'rocket_rucss_is_home_url'
event: 'rocket_rucss_on_submit_jobs', 'rocket_rucss_job_check_status'
transient: 'rocket_rucss_processing'

Related test plan https://wpmediaqa.testrail.io/index.php?/runs/view/844&group_by=cases:section_id&group_id=3445&group_order=asc => We need to run it as is + run using the new hooks to ensure that both are working

jeawhanlee:
Yes, we need to handle the filters above as for `rocket_rucss_clean_rows_time_event`, we are already handling this for backward compatibility.

remyperona (approved):

jeawhanlee:
## List of Deprecated Hooks
- `rocket_last_rucss_job_added_time` -> `rocket_last_saas_job_added_time`
- `rocket_rucss_delete_interval` -> `rocket_saas_delete_interval`
- `rocket_rucss_pending_jobs_cron_interval` -> `rocket_saas_pending_jobs_cron_interval`
- `rocket_remove_rucss_failed_jobs_cron_interval` -> `rocket_remove_saas_failed_jobs_cron_interval`
- `rocket_remove_rucss_on_submit_jobs_cron_interval` -> `rocket_remove_saas_on_submit_jobs_cron_interval`
- `rocket_rucss_deletion_enabled` -> `rocket_saas_deletion_enabled`
- `rocket_rucss_process_pending_jobs_start` -> `rocket_saas_process_pending_jobs_start`
- `rocket_rucss_process_pending_jobs_end` -> `rocket_saas_process_pending_jobs_end`
- `rocket_rucss_check_job_status_end` -> `rocket_saas_check_job_status_end`
- `rocket_rucss_complete_job_status` -> `rocket_saas_complete_job_status`
- `rocket_rucss_process_on_submit_jobs_start` -> `rocket_saas_process_on_submit_jobs_start`
- `rocket_rucss_pending_jobs_cron_rows_count` -> `rocket_saas_pending_jobs_cron_rows_count`
- `rocket_rucss_max_pending_jobs` -> `rocket_saas_max_pending_jobs`
- `rocket_rucss_process_on_submit_jobs_end` -> `rocket_saas_process_on_submit_jobs_end`
- `rocket_delay_remove_rucss_failed_jobs` -> `rocket_delay_remove_saas_failed_jobs`
- `rocket_rucss_retry_table` -> `rocket_saas_retry_table`
- `rocket_rucss_retry_duration` -> `rocket_saas_retry_duration`
- `rocket_rucss_is_home_url` -> `rocket_saas_is_home_url`

MathieuLamiot:
@piotrbak FYI. Do you need anything from the team about this? I think support and maybe public docs needs to be updated?

PR #6390, opened 2024-01-17
Closes #6367 Add theme resolver for compatibility classes

### LINKED ISSUE #6367: Register only the relevant 3rdParty Theme class in the 3rdPartyTheme provider
**Context**

Reduce the plugin CPU usage: https://wp-media.slack.com/archives/CUT7FLHF1/p1704447166535709
There is only up to one theme used on a WordPress website but we currently register (and load) all 3dr party theme compatiblity classes.

**Expected behavior**
When loading the plugin, only up to one 3rd Party theme compatibility class must be loaded. It must be the one corresponding to the theme currently used by the website, or none if no known theme is used.
See [this Slack suggestion](https://wp-media.slack.com/archives/CUT7FLHF1/p1704724171910539?thread_ts=1704447166.535709&cid=CUT7FLHF1)

**How to check results with XDebug**
https://wp-media.slack.com/archives/CUT7FLHF1/p1704728991213049

remyperona:
## Description

Replace current instantiation of themes compatibility classes from loading them all, to only loading the one corresponding to the current theme (if it exists).

Fixes #6367

## Type of change

# Checklists

## Test summary


composer.json
```diff
@@ -42,7 +42,8 @@
 	"require": {
 		"php": ">=7.3",
 		"cloudflare/cf-ip-rewrite": "^1.0",
-		"composer/installers": "^1.0 || ^2.0"
+		"composer/installers": "^1.0 || ^2.0",
+		"friendsofphp/proxy-manager-lts": "^1.0"
```
CrochetFeve0251: @Tabrisrp we should protect libraries we are using to prevent conflicts

CrochetFeve0251: Also why do we need that library for the theme resolver?

inc/ThirdParty/Themes/SubscriberFactory.php
```diff
@@ -0,0 +1,120 @@
...
+		return $factory->createProxy(
+			$class_name,
+			function ( &$wrapped_object, LazyLoadingInterface $proxy, $method, array $parameters, &$initializer ) use ( $class_name, $params ) {
+				$initializer    = null; // disable initialization.
+				$wrapped_object = new $class_name( ...$params );
+
+				return true; // confirm that initialization occurred correctly.
+			}
+		);
+	}
```
CrochetFeve0251: Lazyload is not logic linked to that PR but this one: https://github.com/wp-media/wp-rocket/pull/6377
We shouldn't have it here.

CrochetFeve0251 (changes requested):

inc/ThirdParty/Themes/SubscriberFactory.php
remyperona: It was agreed with @MathieuLamiot to incorporate lazyload for the Divi subscriber in this PR

CrochetFeve0251: Yes but what is the interest from that lazyload? As all  service providers are already loaded at the moment we are [registering subscribers](https://github.com/wp-media/wp-rocket/blob/1558300abaec2db586e29442b4d4657e9c12f94f/inc/Plugin.php#L155) then the classes divi is using are already instantiated.

I see the value when we reduce the load from multiple subscribers to only one being loaded but I don't see it when we spend resources on creating a proxy for a resource already loaded.
For the RUCSS controller we could just [share it](https://github.com/wp-media/wp-rocket/blob/1558300abaec2db586e29442b4d4657e9c12f94f/inc/Engine/Optimization/RUCSS/ServiceProvider.php#L97) instead of using a proxy and we could do the same thing for [the delay js class](https://github.com/wp-media/wp-rocket/blob/1558300abaec2db586e29442b4d4657e9c12f94f/inc/Engine/Optimization/DelayJS/ServiceProvider.php#L50).

MathieuLamiot: > For the RUCSS controller we could just [share it](https://github.com/wp-media/wp-rocket/blob/1558300abaec2db586e29442b4d4657e9c12f94f/inc/Engine/Optimization/RUCSS/ServiceProvider.php#L97) instead of using a proxy and we could do the same thing for [the delay js class](https://github.com/wp-media/wp-rocket/blob/1558300abaec2db586e29442b4d4657e9c12f94f/inc/Engine/Optimization/DelayJS/ServiceProvider.php#L50).

I'll let @Tabrisrp confirm but my understanding is that, in the long run, we should remove dependencies on RUCSS controller (and maybe Delay JS too?) by using filters. Sharing them was discussed already (with Ahmed if I remember correctly) but not suitable for other use cases. So for now, we work around those heavy dependencies with the lazyload.

If there are further blockers or things to clarify, consider doing it synchronously through a call. We should not delay a PR for so long without actions.

remyperona: After discussing with Cyrille, we agreed that this solution is not very good and the benefit vs cons is not worth it.

So I removed it, and focusing on improving the Divi class itself in a future update.

CrochetFeve0251 (approved):

Mai-Saad:
@Tabrisrp Thanks for the PR. using the helper plugin and trying those steps 3 times, can see small improvement on the PR  (instead of global time 54.80ms it is 52.71ms). However, can see UI regression in this scenario https://wpmediaqa.testrail.io/index.php?/tests/view/41471&group_by=cases:section_id&group_id=709&group_order=asc
on trunk:
[trunk.webm](https://github.com/wp-media/wp-rocket/assets/76941962/7db182f6-2992-470c-84bb-8b20578168c1)
on PR:
[branch.webm](https://github.com/wp-media/wp-rocket/assets/76941962/671ca312-81fe-4ebf-9114-125edc1232a1)
**Update**
After discussion with @piotrbak and further investigation on rocketlabs site, the preview image is applied to FE. can you please check?

MathieuLamiot:
@Tabrisrp If I understand the old code correctly, the callback `maybe_disable_youtube_preview` listened to the `switch_theme` hook to catch when Divi was becoming the new theme. And bailed out otherwise.

Now maybe, if we are in another theme and switching to Divi, the callback is not registered since the Divi compat is not loaded so nothing happens.
On the other hand, when switching from the Divi theme, we don't bail out from the callback anymore as the condition has been removed. As a result, we apply the option restriction to the new theme.

If my understanding is correct, the issue is well described here: https://developer.wordpress.org/reference/hooks/switch_theme/
Have you considered using after_switch_theme instead?

remyperona:
Tested with `after_switch_theme`, it's working, updated the code to reflect

MathieuLamiot:
Nice!

Mai-Saad:
working fine now

PR #6409, opened 2024-01-24
Closes #6394: Reintroduce limit preload at job level

### LINKED ISSUE #6394: Bring back batch limit on Preload at job level
**Context**

To prevent to many preload requests to be performed and hanging at the same time, especially on slow servers, we had a limit on the number of in-progress jobs, implemented [here](https://github.com/wp-media/wp-rocket/pull/5393) in the `get_pending_jobs` method.
This behavior has been changed in [this PR](https://github.com/wp-media/wp-rocket/pull/6084) to enforce the limit at Action Scheduler level.
The issue is that "pending async task in AS" <= "in-progress jobs". It should be equal most of the time, but the difference can be huge on slow websites. What can happen is that the async task is done (request sent) but the preload on this url is still processing (job still in-progress).
As a result, there are cases today where we can flood a website with more than 45 in-progress jobs (meaning probably more than 90 requests simultaneously because of desktop/mobile).
It also reduced the efficiency of helper plugins relying on `rocket_preload_cache_pending_jobs_cron_rows_count`

**Expected behavior**
We must never have more than `rocket_preload_cache_pending_jobs_cron_rows_count` in-progress jobs for the cache preload.

**Acceptance Criteria**
- Manually set some jobs in-progress into the DB. Call the `process_pending_jobs` method. At the end of it, there should be exactly 45 pendings jobs.
- Do the same with the `rocket_preload_cache_pending_jobs_cron_rows_count` filter to set the number of jobs to another number. The number of in progress job should match the filtered value.

**Technical discussion**
Looking at the code, I think that, by design "pending async task in AS" <= "in-progress jobs". So we could remove the check on AS queue size. But it's maybe better to keep it for safety, it probably does not cost much anyway.

Miraeld:
## Description

Re-introduction of the preload limit at the job level

Fixes #6394 

## Type of change

No

# Checklists

## Test summary

### DIFF
--- a/inc/Engine/Preload/Database/Queries/Cache.php
+++ b/inc/Engine/Preload/Database/Queries/Cache.php
@@ -338,8 +338,16 @@ class Cache extends Query {
 	 * @return array
 	 */
 	public function get_pending_jobs( int $total = 45 ) {
+		$inprogress_count = $this->query(
+			[
+				'count'     => true,
+				'status'    => 'in-progress',
+				'is_locked' => false,
+			],
+			false
+		);
 
-		if ( $total <= 0 ) {
+		if ( $total <= 0 || $inprogress_count >= $total ) {
 			return [];
 		}
 
@@ -358,7 +366,7 @@ class Cache extends Query {
 
 		return $this->query(
 			[
-				'number'         => $total,
+				'number'         => ( $total - $inprogress_count ),
 				'status'         => 'pending',
 				'fields'         => [
 					'id',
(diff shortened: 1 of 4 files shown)


Khadreal (approved):

remyperona (approved):

Mai-Saad:
Related test plan https://wpmediaqa.testrail.io/index.php?/runs/view/808&group_by=cases:section_id&group_order=asc
Note: will be nice to do a performance check vs trunk with a big batch size

jeawhanlee:
During test plan execution, When helper plugin is used: https://drive.google.com/uc?id=1JDqh9xV6zyLcD3OR_Njsxn8KW_PYGoac with webp enabled in addon, only webp versions are generated. Same thing is happening on trunk but working fine on 3.12.3 with no webp version generated.

PR #6411, opened 2024-01-25
Closes #6372: Log on submit status last entry

### LINKED ISSUE #6372: As a support team member,  I want to know the last time an entry was added to wpr_rucss_used_css table with the status on-submit so that I can check if the RUCSS feature is properly triggered
**Context**

As part of building a RUCSS Debug tool, some information must be stored while in WP Rocket Debug mode so that it can be monitored easily by the support team.
More details here: https://www.notion.so/wpmedia/RUCSS-Issues-7adc4ed750fd4dc7902bb6881ed1204b?pvs=4#38398060400d4fcf97f27aa563483d68

**Expected behavior**
- When in debug mode (WP_ROCKET_DEBUG?), a timestamp must be updated in the DB every time the `UsedCSS::create_new_job` method is successful. We only need to keep the last time it happened.
- When not in debug mode, nothing special happens. If there is a timestamp stored, it can stay there.

jeawhanlee:
## Description

This PR logs the last entry for new rucss job added.

Fixes #6372 

## Type of change

- New feature (non-breaking change which adds functionality).

No

# Checklists

## Test summary


inc/Engine/Debug/RUCSS/Subscriber.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace WP_Rocket\Engine\Debug\RUCSS;
+
+use WP_Rocket\Admin\Options_Data;
```
remyperona:
```suggestion
use WP_Rocket\Admin\{Options, Options_Data};
```

inc/Engine/Debug/Resolver.php
```diff
@@ -0,0 +1,45 @@
...
+	public static function get_services(): array {
+		$set_services = [];
+
+		if ( empty( self::$options ) ) {
+			return [];
+		}
+
+		foreach ( self::$options as $option => $services ) {
+
+			if ( ! (bool) get_rocket_option( $option ) ) {
```
remyperona: We could use the Options_Data instance instead

inc/classes/ServiceProvider/class-options.php
```diff
@@ -22,6 +22,7 @@ class Options extends AbstractServiceProvider {
 	 */
 	protected $provides = [
 		'options',
+		'options_debug',
```
remyperona: I would register this class in the Debug service provider too instead

remyperona (changes requested):

wp-rocket.php
```diff
@@ -26,6 +26,7 @@
 define( 'WP_ROCKET_PHP_VERSION',           '7.3' );
 define( 'WP_ROCKET_PRIVATE_KEY',           false );
 define( 'WP_ROCKET_SLUG',                  'wp_rocket_settings' );
+define( 'WP_ROCKET_DEBUG_SLUG',            'wp_rocket_debug' );
```
remyperona: I think we should move away from using global constants like this. We can use the option name directly, with our architecture now it's more centralized.

inc/admin/upgrader.php
```diff
@@ -143,6 +143,21 @@ function rocket_first_install() {
...
+		 *
+		 * @param array Array of default rocket debug options.
+		 */
+		apply_filters(
+			'rocket_first_install_debug_options',
+			[
+				'last_rucss_job_added' => '',
+			]
+		)
+	);
```
remyperona: Couple of things here:
- We should do this in a separate method in a debug subscriber
- We need to check that it's only done once when upgrading, by using the plugin version check

inc/Engine/Debug/ServiceProvider.php
```diff
@@ -0,0 +1,63 @@
...
+			return;
+		}
+
+		$this->container->add( 'options_debug', Options_Data::class )
+			->addArgument( $this->container->get( 'options_api' )->get( 'debug', [] ) );
+
+		foreach ( $services as $service ) {
+			$this->getContainer()->add( $service['service'], $service['class'] )
+				->addArgument( $this->getContainer()->get( 'options_debug' ) )
+				->addArgument( $this->getContainer()->get( 'options_api' ) );
```
remyperona, line 69: this only works if all services have the same construct signature forever, are we sure about it?

inc/Engine/Debug/Resolver.php
```diff
@@ -0,0 +1,62 @@
...
+class Resolver {
+
+	/**
+	 * Array of WP Rocket Options.
+	 *
+	 * @var array
+	 */
+	private static $options_services = [
+		'remove_unused_css' => [
+			'service' => 'rucss_subscriber',
```
remyperona:
```suggestion
			'service' => 'rucss_debug_subscriber',
```

remyperona (changes requested):

inc/Engine/Debug/ServiceProvider.php
jeawhanlee: From what I see, we'll mostly be updating the debug options based on triggered hooks. Do you see something else?

inc/Engine/Debug/Resolver.php
```diff
@@ -0,0 +1,62 @@
...
+		foreach ( self::$options_services as $option => $services ) {
+			if ( ! (bool) self::$options->get( $option, 0 ) ) {
+				continue;
+			}
+
+			$set_services[] = $services;
+		}
+
+		return $set_services;
+	}
```
CrochetFeve0251: @jeawhanlee why everything is static?

inc/Engine/Debug/RUCSS/Subscriber.php
```diff
@@ -0,0 +1,61 @@
...
+	 * @param string $timestamp Current timestamp.
+	 * @return void
+	 */
+	public function log_last_added_job_time( $is_success, $timestamp ) {
+		if ( ! $is_success ) {
+			return;
+		}
+
+		$this->options->set( 'last_rucss_job_added', $timestamp );
+		$this->options_api->set( 'debug', $this->options->get_options() );
```
CrochetFeve0251, line 59: @jeawhanlee by default options contains the wpr settings.

remyperona (approved):

PR #6415, opened 2024-01-29
Fix Contact Form 7 compatibility issues

### LINKED ISSUE #6384: Some issues with our compatibility file for Contact Form 7
I checked two escalated cases and found out some issues with Contact Form 7 so I'll add the full details here so we can take a decision with the team if we want/need to do any enhancement here or have a documentation for the following points but I need to give you a small intro that most of those points are depending on, which is:

When you register a script without enqueuing it and then enqueue another script that has the first script as a dependency, this will force enqueuing the first one, let's check the code of CF7 to see this in action:

Start here, they register the script `contact-form-7`
https://github.com/rocklobster-in/contact-form-7/blob/v5.8.6/includes/controller.php#L49-L58

and only enqueue it inside the following function:
https://github.com/rocklobster-in/contact-form-7/blob/v5.8.6/includes/controller.php#L110

and this function is called when the filter `wpcf7_load_js` is applied here:

https://github.com/rocklobster-in/contact-form-7/blob/v5.8.6/includes/controller.php#L68-L70

So the issue happens when a script is enqueued but has the `contact-form-7` script ID as a dependency, this means that CF7 script will be enqueued directly without checking this filter so this may cause some errors with our compatibility file as below:

## Console error in pages that don't have forms when having `Conditional Fields for Contact Form 7` plugin active

When `Conditional Fields for Contact Form 7` is active we see a console error as below:

image

and this is happening only for the pages that don't have contact forms.

and after checking their code I found that they start enqueuing their scripts here:
https://plugins.trac.wordpress.org/browser/cf7-conditional-fields/tags/2.4.6/cf7cf.php#L510

then calling their function `wpcf7cf_enqueue_scripts` then here:
https://plugins.trac.wordpress.org/browser/cf7-conditional-fields/tags/2.4.6/cf7cf.php#L524

they have CF7 script as a dependency, so as u can see they have a function called `wpcf7cf_load_js` which applies the filter `wpcf7cf_load_js`

So we need to use the following filters to control the loading of their scripts that causes CF7 scripts to be enqueued too:
```
wpcf7cf_load_js
wpcf7cf_load_css
```

we can updated our compatibility file here:
https://github.com/wp-media/wp-rocket/blob/0484f529d64e11e9a6d2f7af13127f69bb76045c/inc/ThirdParty/Plugins/ContactForm7.php#L6

with every instance of `wpcf7_load_js` we need to do that same with the filter `wpcf7cf_load_js`, for example here:
https://github.com/wp-media/wp-rocket/blob/0484f529d64e11e9a6d2f7af13127f69bb76045c/inc/ThirdParty/Plugins/ContactForm7.php#L36-L37

we need to add the following two lines:
```
add_filter( 'wpcf7cf_load_js', '__return_false' );
add_filter( 'wpcf7cf_load_css', '__return_false' );
```
and the same needs to be done below in the same file, or we can create another compatibility for that.

## Conflict when custom themes that is adding custom scripts that works with CF7

I found a customer that has the following code in his custom theme:
```
add_action( 'wp_enqueue_scripts', 'XXX_scripts', 100 );
function XXX_scripts() {
	wp_enqueue_script(
		'XXX-contact-form-7',
		get_stylesheet_directory_uri() . '/assets/js/contact-form-7.js',
		[ 'contact-form-7', 'jquery' ],
		XXX_VERSION,
		[ 'in_footer' => true ]
	);
}
```

this code will also force enqueuing CF7 script even when not needed (without checking if CF7 should be enqueued or not)

This can be fixed by replacing the action from `wp_enqueue_scripts` to be `wpcf7_enqueue_scripts` which only fires when CF7 scripts are being loaded.

Enqueuing CF7 directly is a problem because it will load the script only without adding the localized script `wpcf7` which is a requirement for the CF7 script.

**So the question is, why this is happening only with our compatibility file?**
That's a good question, the answer is that by default the filters `wpcf7_load_js` and `wpcf7_load_css` return true so the script will always be enqueued along with the localized script.

But our compatibility file is changing their return based on if the current page contain a contact form or not, and if case of `Not` the localized script won't be added even if a third party enqueued CF7 script which causes console errors and corrupted functionality.

So I believe we are not wrong, the wrong comes from the third parties that need to add some condition before loading their scripts.

We need a product decision here on what should we do next.
@wp-media/productrocket

wordpressfan:
## Description

This is a working prototype to detect if CF7 main script is added as a dependency then force loading all JS in this case not to fire console error.

I tried many ways like using the `script_loader_src` filter to detect if CF7 main script is being loaded but at this point I won't be able to load the localized script because the main script is already loaded in frontend.

Fixes #6384

## Type of change

Not groomed, I did R&D and directly created this PR so we all can validate the idea.

# Checklists

## Test summary

### DIFF
--- a/inc/ThirdParty/Plugins/ContactForm7.php
+++ b/inc/ThirdParty/Plugins/ContactForm7.php
@@ -25,16 +25,6 @@ class ContactForm7 implements Subscriber_Interface {
 	 * Subscribed events.
 	 */
 	public static function get_subscribed_events() {
-		/**
-		 * Filters register this compatibility events or not.
-		 *
-		 * @param bool $status Load the compatibility file or not, default is True.
-		 * @param string $thirdparty Thirdparty id.
-		 */
-		if ( ! apply_filters( 'rocket_thirdparty_load', true, 'contact-form-7' ) ) {
-			return [];
-		}
-
 		return [
 			'template_redirect' => 'maybe_optimize_contact_form_7',
 		];
@@ -44,6 +34,16 @@ class ContactForm7 implements Subscriber_Interface {
 	 * Optimize ContactForm7 scripts.
 	 */
 	public function maybe_optimize_contact_form_7() {
+		/**
+		 * Filters register this compatibility events or not.
+		 *
+		 * @param bool $status Load the compatibility file or not, default is True.
+		 * @param string $thirdparty Thirdparty id.
+		 */
+		if ( ! apply_filters( 'rocket_thirdparty_load', true, 'contact-form-7' ) ) {
+			return;
+		}
+
 		// The wpcf7_shortcode_callback action was added in CF7 version 5.8.1.
 		if ( ! defined( 'WPCF7_VERSION' ) || version_compare( WPCF7_VERSION, self::REQUIRED_CF7_VERSION, '<' ) ) {
 			return;
--- a/inc/ThirdParty/Plugins/ContactForm7.php
+++ b/inc/ThirdParty/Plugins/ContactForm7.php
@@ -15,9 +15,26 @@ class ContactForm7 implements Subscriber_Interface {
 	const REQUIRED_CF7_VERSION = '5.8.1';
 
 	/**
+	 * CF7 scripts load status.
+	 *
+	 * @var bool
+	 */
+	private $load_js;
+
+	/**
 	 * Subscribed events.
 	 */
 	public static function get_subscribed_events() {
+		/**
+		 * Filters register this compatibility events or not.
+		 *
+		 * @param bool $status Load the compatibility file or not, default is True.
+		 * @param string $thirdparty Thirdparty id.
+		 */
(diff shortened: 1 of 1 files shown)


Khadreal (approved):

Mai-Saad:
Related test plan https://wpmediaqa.testrail.io/index.php?/runs/view/810&group_by=cases:section_id&group_order=asc

Mai-Saad (approved):
Working as expected
[testrail-report-566.pdf](https://github.com/wp-media/wp-rocket/files/14111619/testrail-report-566.pdf)
Note: the compatibility will stop if the filter returns false, null, "", []

d-solution-nl:
Since [3.15.9], my contact form stopped working. I applied the follwing: 

> add_filter( 'rocket_thirdparty_load', function( $status, $thirdparty ) {
> if ( 'contact-form-7' !== $thirdparty ) {
>     return $status;
> }
> return false;
> }, 10, 2 );

Appying this fix didnt include the script files on pages without CF7. However, i need this to load because i use a popup for CF7

PR #6427, opened 2024-02-01
Preload adaptative batch

### LINKED ISSUE #6396: [R&D] Adaptative batch size on Preload
**Context**
It seems Preload can generate a lot of pressure on a server if the pages of the website are slow to open. A way to adapt to this would be to measure how long a request takes, and adapt batch size based on this.

**What to do**
This branch is a quick&dirty example of how this could be implemented: https://github.com/wp-media/wp-rocket/tree/prototype/preload-adaptative-batch
The idea is partially described [here](https://wp-media.slack.com/archives/CUT7FLHF1/p1705314139098689), but has evolved a bit to base the batch size on the measurement of a preload request, by making one request blocking from time to time.

A developer from the plugin team needs to go spend some time on this branch to make it work (maybe it is not, I just wrote the code to lay the idea down), production ready, and play with it to see how it behaves, possibly with logs.
We have the gamma.rocketlabs.ovh website that suffers from CPU issues when doing a full cache clear to trigger the preload. It would be a good place to test it. See [here](https://wp-media.slack.com/archives/CUKB44GNN/p1705523690099229).

**Warning**
This branch would need https://github.com/wp-media/wp-rocket/issues/6394 
Otherwise, we don't have control to prevent flooding the AS queue and the number of job in-progress could increase too quickly.

MathieuLamiot:
## Description

When preparing preload batches, the batch size will be adapted based on the average duration of a request to avoid flooding servers with requests if they take too long.
To get the estimate of the request duration, we do a blocking request from time to time when triggering the preload, and measure the timing.

Fixes #6396 

## Type of change

- New feature (non-breaking change which adds functionality).

Yes, on top of the original idea, we added an average mechanism that keeps track of the average request time through transients.

# Checklists

## Test summary

Tested locally and on the gamma website by checking transients and scheduled actions.


inc/Engine/Preload/Controller/PreloadUrl.php
```diff
@@ -136,10 +144,30 @@ public function preload_url( string $url ) {
...
+				if ( $previous_request_durations <= 0 ) {
+					$previous_request_durations = $duration;
+				} else {
+					$previous_request_durations = $previous_request_durations * 0.7 + $duration * 0.3;
+				}
+				set_transient( 'rocket_preload_previous_request_durations', $previous_request_durations, 5 * 60 );
+
+				set_transient( 'rocket_preload_check_duration', $duration, 60 ); // Don't check request duration for 1 minute.
+				$check_duration = false;
+			}
```
jeawhanlee:
```suggestion
			if ( ! $check_duration ) {
	return;
}
$duration = ( microtime( true ) - $start ); // Duration of the request.

// Update average duration.
$previous_request_durations = get_transient( 'rocket_preload_previous_request_durations' ) ?? 0;
$previous_request_durations =$previous_request_durations <= 0 ? $duration : $previous_request_durations * 0.7 + $duration * 0.3 ;
set_transient( 'rocket_preload_previous_request_durations', $previous_request_durations, 5 * 60 );

set_transient( 'rocket_preload_check_duration', $duration, 60 ); // Don't check request duration for 1 minute.
$check_duration = false;
```

inc/Engine/Preload/Controller/PreloadUrl.php
```diff
@@ -183,7 +211,24 @@ public function process_pending_jobs() {
...
+		$min_batch_size           = ( (int) apply_filters( 'rocket_preload_cache_min_in_progress_jobs_count', 5 ) );
+		$preload_request_duration = get_transient( 'rocket_preload_previous_request_durations' );
+
+		// Estimate batch size based on request duration.
+		if ( ! $preload_request_duration ) {
+			$next_batch_size = $min_batch_size; // In case no estimation or there is an issue with the value.
+		} else {
+			// Linear function: 2s -> 45 jobs // 10s -> 5 jobs.
+			$next_batch_size = round( -5 * $preload_request_duration + 55 );
+		}
```
jeawhanlee:
```suggestion
		$next_batch_size = ! $preload_request_duration ? $min_batch_size : round( -5 * $preload_request_duration + 55 );
```

inc/Engine/Preload/Controller/PreloadUrl.php
```diff
@@ -183,7 +211,24 @@ public function process_pending_jobs() {
 
 		$pending_actions = $this->queue->get_pending_preload_actions();
 
-		$count = ( (int) apply_filters( 'rocket_preload_cache_pending_jobs_cron_rows_count', 45 ) ) - count( $pending_actions );
+		// Retrieve batch size limits and request timing estimaiton.
+		$max_batch_size           = ( (int) apply_filters( 'rocket_preload_cache_pending_jobs_cron_rows_count', 45 ) ) - count( $pending_actions );
+		$min_batch_size           = ( (int) apply_filters( 'rocket_preload_cache_min_in_progress_jobs_count', 5 ) );
```
jeawhanlee: Document these filters https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters.

jeawhanlee (changes requested):

inc/Engine/Preload/Controller/PreloadUrl.php
Khadreal: Nice suggestion but the code still need to run even when check_duration is false

jeawhanlee: Suggestion, what do you think if we move the filter above the bail condition and call `usleep( $delay_between );` before the return? 😁
Then lastly, can we update https://github.com/wp-media/wp-rocket/blob/3c706035932f389170fb0cb9575631d98a3d3c79/inc/Engine/Preload/Controller/PreloadUrl.php#L161-L165 to use ternary operator?

MathieuLamiot: @jeawhanlee We can't use a return here: this whole part of the code is in a for loop and we should never skip iterations of this loop.
Instead of return, we could consider using `continue`. However, we should be careful: the usleep should not be done before `$duration = ( microtime( true ) - $start ); // Duration of the request.`, otherwise, it breaks the timing computation.

So we could eventually have the following:

```
			if ( $check_duration ) {
				$duration = ( microtime( true ) - $start ); // Duration of the request.
                         }
			/**
			 * Filter the delay between each preload request.
			 *
			 * @param float $delay_between the defined delay.
			 * @returns float
			 */
			$delay_between = apply_filters( 'rocket_preload_delay_between_requests', 500000 );

			usleep( $delay_between );
                         if ( ! $check_duration ) {
				continue;
                         }
			// Update average duration.
			$previous_request_durations = get_transient( 'rocket_preload_previous_request_durations' ) ?? 0;
			if ( $previous_request_durations <= 0 ) {
				$previous_request_durations = $duration;
			} else {
				$previous_request_durations = $previous_request_durations * 0.7 + $duration * 0.3;
			}
			set_transient( 'rocket_preload_previous_request_durations', $previous_request_durations, 5 * 60 );

			set_transient( 'rocket_preload_check_duration', $duration, 60 ); // Don't check request duration for 1 minute.
			$check_duration = false;

		}
```
I don't have strong feeling for the current version or this one :)

jeawhanlee (approved):

MathieuLamiot:
@Khadreal Hey, I just pushed a small change to enforce the max batch size after the min batch size. This is for the case there are already 45 actions in the AS queue. In this case, no matter what, we don't want to add more to the AS queue. So the max has to be enforced after.
Checking the tests now

inc/Engine/Preload/Controller/PreloadUrl.php
```diff
@@ -131,19 +141,42 @@ public function preload_url( string $url ) {
...
+				$duration = ( microtime( true ) - $start ); // Duration of the request.
+			}
+
 			/**
 			 * Filter the delay between each preload request.
 			 *
 			 * @param float $delay_between the defined delay.
-			 * @returns float
 			 */
 			$delay_between = apply_filters( 'rocket_preload_delay_between_requests', 500000 );
```
Miraeld, line 162: I know it's not been added by this PR, but shouldn't we add a safeguard here ?

Miraeld (approved):

MathieuLamiot:
@wp-media/qa-team To test this issue:
- Duration of the latest request can be seen in Transients: rocket_preload_check_duration
- Average duration taken into account is in Transient: rocket_preload_previous_requests_durations
@piotrbak shared access to some hosts on BitWarden that can provide resource usage plugins built-in:

Cloudways
Hostinger
O2Switch

We can also check CPU usage on rocketlabs (WordOps) for gamma for instance.
---

It would be great if you could explore a bit how this behaves with an "heavy" website: typically gamma, might want to duplicate it with Duplicator for instance? ; Another approach could be to add some CPU intensive computation in the initialization of WP or of WP Rocket to make as if the website was slow.
It's not easy to reproduce those issues, so it's very exploratory... Apart from non-regressions/errors, the goal would mostly be to see if we're able to see a CPU usage reduction (lower CPU usage, or shorter spikes for instance) on a "heavy" website between trunk and this branch.

hanna-meda (approved):
Related TP execution Results below:
[testrail-report-613.pdf](https://github.com/user-attachments/files/16053622/testrail-report-613.pdf)

PR #6445, opened 2024-02-13
Closes #6202: Stop preloading fonts when it's excluded

### LINKED ISSUE #6202: Stop WP rocket preloading all fonts it finds in the css when 'Remove unused css' is turned on
When using the 'Remove unused css' feature, WP-rocket pre-loads all fonts it finds in the CSS files. This causes many unwanted fonts to be pre-loaded (especially those from plugin files) and affects LCP etc.

A toggle button on whether to preload fonts it finds or not when using this feature.

Using a different optimization plugin.

Miraeld:
# Description
This PR gives users a filter to be able to exclude fonts urls to be preloaded.

Fixes #6202 

### User documentation
With this PR, you now have the ability to control which fonts get preloaded on your website. This is done using the `rocket_exclude_rucss_fonts_preload` filter.  
Let's say your website uses two fonts: one from `https://domainA.com/fonts/myFontA.woff` and another from `https://domainB.com/fonts/myFontsB.woff`. If you want to **prevent** the font from `domainA.com` from being preloaded, you can use the filter and return `['https://domainA.com']`.  This means that `myFontA.woff` **won't be preloaded**, but `myFontB.woff` **will be**. 
This gives you more control over the performance of your website.

### Technical documentation

We apply a filter to get the excluded fonts urls, and in the loop we check if the current url is within the excluded array. If that's the case we bail-out.

## Type of change

# Checklists

## Feature validation

- [x] I added tests 

## Code style

### DIFF
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -785,6 +785,14 @@ class UsedCSS implements LoggerAwareInterface {
 			return $html;
 		}
 
+        /**
+         * Filters the list of fonts to exclude from preload
+         *
+         * @since 3.15.10
+         *
+         * @param array $excluded_fonts_preload List of fonts to exclude from preload
+         */
+        $exclude_fonts_preload = apply_filters( 'rocket_exclude_rucss_fonts_preload', [] );
 		$urls = [];
 
 		foreach ( $font_faces as $font_face ) {
@@ -793,6 +801,7 @@ class UsedCSS implements LoggerAwareInterface {
 			}
 
 			$font_url = $this->extract_first_font( $font_face['content'] );
+
 
 			/**
 			 * Filters font URL with CDN hostname
@@ -806,6 +815,11 @@ class UsedCSS implements LoggerAwareInterface {
 			if ( empty( $font_url ) ) {
 				continue;
 			}
+
+            // Check if the font URL is in the exclude_fonts_preload array
+            if (in_array($font_url, $exclude_fonts_preload)) {
+                continue; // Skip this iteration as the font URL is in the exclusion list
+            }
 
 			$urls[] = $font_url;
 		}
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -785,14 +785,15 @@ class UsedCSS implements LoggerAwareInterface {
 			return $html;
 		}
 
-        /**
-         * Filters the list of fonts to exclude from preload
-         *
-         * @since 3.15.10
-         *
-         * @param array $excluded_fonts_preload List of fonts to exclude from preload
-         */
-        $exclude_fonts_preload = apply_filters( 'rocket_exclude_rucss_fonts_preload', [] );
+		/**
+		 * Filters the list of fonts to exclude from preload
+		 *
+		 * @since 3.15.10
+		 *
+		 * @param array $excluded_fonts_preload List of fonts to exclude from preload
+		 */
+		$exclude_fonts_preload = apply_filters( 'rocket_exclude_rucss_fonts_preload', [] );
+
 		$urls = [];
 
 		foreach ( $font_faces as $font_face ) {
@@ -801,7 +802,6 @@ class UsedCSS implements LoggerAwareInterface {
 			}
 
 			$font_url = $this->extract_first_font( $font_face['content'] );
-
 
 			/**
 			 * Filters font URL with CDN hostname
@@ -816,10 +816,10 @@ class UsedCSS implements LoggerAwareInterface {
 				continue;
 			}
 
-            // Check if the font URL is in the exclude_fonts_preload array
-            if (in_array($font_url, $exclude_fonts_preload)) {
-                continue; // Skip this iteration as the font URL is in the exclusion list
-            }
+			// Check if the font URL is in the exclude_fonts_preload array.
+			if ( in_array( $font_url, $exclude_fonts_preload, true ) ) {
+				continue; // Skip this iteration as the font URL is in the exclusion list.
+			}
 
 			$urls[] = $font_url;
 		}
--- a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
+++ b/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
@@ -793,6 +793,9 @@ class UsedCSS implements LoggerAwareInterface {
 		 * @param array $excluded_fonts_preload List of fonts to exclude from preload
 		 */
 		$exclude_fonts_preload = apply_filters( 'rocket_exclude_rucss_fonts_preload', [] );
+		if ( ! is_array( $exclude_fonts_preload ) ) {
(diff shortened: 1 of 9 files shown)


inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -785,6 +785,15 @@ private function add_used_fonts_preload( string $html, string $used_css ): strin
...
 
+		/**
+		 * Filters the list of fonts to exclude from preload
+		 *
+		 * @since 3.15.10
+		 *
+		 * @param array $excluded_fonts_preload List of fonts to exclude from preload
+		 */
+		$exclude_fonts_preload = apply_filters( 'rocket_exclude_rucss_fonts_preload', [] );
+
```
remyperona: We should add a safe guard in case the filter value is not an array, to fallback to the default value, like we do in other places

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -807,6 +816,11 @@ private function add_used_fonts_preload( string $html, string $used_css ): strin
 				continue;
 			}
 
+			// Check if the font URL is in the exclude_fonts_preload array.
+			if ( in_array( $font_url, $exclude_fonts_preload, true ) ) {
```
remyperona: This will need a slightly different approach, because we need to do partial match too (like just the domain, or a path), which won't work with `in_array()`.

the array can be combined into a string with `|` separators and then use it as a pattern for a `preg_match()` for example

remyperona (changes requested):

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
Miraeld: I've added a safeguard :)

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
Miraeld: I didn't undertstood it like that, my bad, 
I've changed it for another approach using `preg_match` :)

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
```diff
@@ -807,6 +819,14 @@ private function add_used_fonts_preload( string $html, string $used_css ): strin
 				continue;
 			}
 
+			// Combine the array elements into a single string with | as a separator.
+			$exclude_fonts_preload_pattern = implode( '|', array_map( 'preg_quote', $exclude_fonts_preload ) );
+
+			// Check if the font URL matches any part of the exclude_fonts_preload array.
+			if ( preg_match( '/' . $exclude_fonts_preload_pattern . '/i', $font_url ) ) {
```
remyperona: You will need to either:
- change the delimiter, since there might be `/` in the patterns, and `preg_quote()` doesn't escape them by default
- escape `/` with the extra parameter of `preg_quote()`

remyperona (changes requested):

MathieuLamiot:
@Miraeld a few feedbacks:

-  I implemented built-in tests to cover the new/changed code. -> Seems that it disappeared from the checklist? I think new hooks should always have basic test coverage, no?
- About the user documentation, what a user or the support team would need is:
   - the name of the filter
   - what data to pass (maybe with one example?)
Providing this directly within the PR would ease future work to build the release note and update the documentation. (cc @piotrbak :D )

inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php
Miraeld: What do you think of something like this:
```php
// Combine the array elements into a single string with | as a separator and returning a pattern.
$exclude_fonts_preload_pattern = implode( '|', array_map( function($item) {
    return preg_quote($item, '/');
}, $exclude_fonts_preload ) );

// Check if the font URL matches any part of the exclude_fonts_preload array.
if ( preg_match( '/' . $exclude_fonts_preload_pattern . '/i', $font_url ) ) {
    continue; // Skip this iteration as the font URL is in the exclusion list.
}

```

remyperona: this works

Miraeld: Change done :)

Miraeld:
@MathieuLamiot , I've added integrations tests to the PR and modified the User documentation. :)

remyperona (approved):

Mai-Saad:
@Miraeld Thanks for the PR.
While doing exploratory testing, can see that used font in used CSS isnot preloaded by default. Can you please check?

Miraeld:
@Mai-Saad,

The issue was imploding an empty array gives an empty string. And it would match everything in the `preg_match` [here](https://github.com/wp-media/wp-rocket/blob/a41b8b6e307aa878225bcc3a0fec941fb544d14a/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L836). 
It's been fixed now as I've added a check if it's empty. 
And I also added another check in case we passe `['']` to the filter, and `empty([''])` wouldn't be true. So the same issue would occur.

Khadreal (approved):
Looks good to me

Mai-Saad:
Related test plan https://wpmediaqa.testrail.io/index.php?/runs/view/817&group_by=cases:section_id&group_order=asc

jeawhanlee:
Testing in Progress

jeawhanlee:
@Miraeld Thanks For the PR
Please find points below while executing test plan
- When we enable CDN rewriting, we will only rewrite to point to the CDN if the font style is having a relative url but if we have an absolute url, it will be ignored. 
e.g /wp-content/rocket-test-data/fonts/ballet-v3-latin-regular.woff2 ->  cdn.example.com/wp-content/rocket-test-data/fonts/ballet-v3-latin-regular.woff2 
but example.com/wp-content/rocket-test-data/fonts/ballet-v3-latin-regular.woff2 will remain the same.
This is not a regression as it is already on trunk.
@piotrbak @Mai-Saad Do we need to handle this?
- Adding a null value to the array to be returned in this filter is throwing a fatal error in debug.log.
```
add_filter( 'rocket_exclude_rucss_fonts_preload', function( $fonts ) {
  $fonts[]='woocommerce/assets/fonts';
  $fonts[]=null;
  return $fonts;
} );
```
`PHP Fatal error:  Uncaught TypeError: preg_quote() expects parameter 1 to be string, null given in /home/test/wp-content/plugins/wp-rocket/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php:829`
We need to guard this here: https://github.com/wp-media/wp-rocket/blob/66daba4d57e5f576ae3450c947bc0cdab528cce2/inc/Engine/Optimization/RUCSS/Controller/UsedCSS.php#L829

Miraeld:
Umm, For the 1st point, I've noticed it, but as it is on trunk, I supposed it was normal.

For the 2nd point, it's fixed in the latest commit.

jeawhanlee:
The fatal error is fixed and all test plan passed.
Opened a new issue for point 1 above [here](https://github.com/wp-media/wp-rocket/issues/6458)

PR #6488, opened 2024-03-18
Closes #6431 AJAX insertion of LCP data

### LINKED ISSUE #6431: 3.16 - Beacon script implementation
## Context
As a user, I want WP Rocket to analyse and store LCP/ATF information of my pages so that those images can then be optimised. This issue has the goal to develop the green part of this diagram:
Untitled

**Must be done after #6430 so that it can be tested**

## Scope a solution 
- To implement this, we need to create the beacon script in javascript that will run on pages to collect images.
- A prototype of the script can be found [here](https://gist.github.com/DahmaniAdame/038bdaa622313c8cde27ca60cb168924).
- The script can be created and placed [here](https://github.com/wp-media/wp-rocket/blob/3f7fa21018ff747a4c7ab90f18c37f1423e12f2b/src/js/global). 

The script needs to collect images of the page, and save them within the database through an AJAX call. 

If the script encounter an error, we should save it to the DB as well.

- Add a new Subdir `AJAX` with a controller and a subscriber
- The subscriber will have an event `wp_ajax_rocket_lcp`, with a callback `add_lcp_data()`
- The method will call the same named method in the controller
- This method will update the DB based on the result of the script (failed, error, no image, success), while performing security checks
- The controller will need to have a dependency with the DB query class to write in the table
- Add tests

## Acceptance Criteria
- Data is added to the DB once the AJAX call is made

##  Effort estimation:
 Effort [M]

### LINKED ISSUE #6486: 3.16 - R&D - Beacon script compatibility with LazyLoad
**Context**

@Tabrisrp identified that the current implementation does not correctly identify ATF/LCP images as Lazyload features can interfere with the beacon script: it catches lazyload placeholders instead of the proper image and stores it in the DB. This prevents the front-end optimization to be successfully applied.
There are several approaches to work around this ; the goal of this task is to provide the Product Team with a first prototype with a possible work-around, based on their suggestion.

Slack: https://wp-media.slack.com/archives/CUT7FLHF1/p1710536832224719

**Expected behavior**

- Delay the execution of the Beacon Script by 500ms so that lazyload scripts had time to run and so that the Beacon script works on a properly loaded above the fold content.

**Acceptance Criteria**
- This may not fix the original issue observed by @Tabrisrp, as this is an R&D task to discover possible work-around, and what won't work. So there is no acceptance criteria related to the 3.16 functionality.
- Endure that the script execution is delayed.

remyperona:
# Description

Add the base LCP extraction script, and add the AJAX to save the data in the DB.

Fixes #6431, #6486

## Type of change

# Checklists

## Feature validation

## Code style


assets/js/lcp-beacon.js
```diff
@@ -1 +1,152 @@
...
+            // If it's not a duplicate, push the new element
+            if (!isDuplicate && parent === document) {
+                performance_images.push({
+                    src: image.src,
+                    label: "above-the-fold",
+                });
+            }
+        }
+    }
+    console.log(performance_images);
```
Miraeld, line 120: There are multiple `console.log()` in this file. Do we keep them?

tests/Integration/inc/Engine/Media/AboveTheFold/AJAX/Subscriber/addLcpData.php
```diff
@@ -0,0 +1,66 @@
...
+
+		parent::tear_down();
+	}
+
+	/**
+	 * @dataProvider configTestData
+	 */
+	public function testShouldReturnExpected( $config, $expected ) {
+		$_POST['rocket_lcp_nonce'] = wp_create_nonce( 'rocket_lcp' );
+		$_POST['action']           = 'rocket_lcp';
```
Miraeld, line 40: You set `$this->action = 'rocket_lcp';` in the `set_up()` function, but you don't use it here. Is there a reason?

Miraeld (changes requested):

assets/js/lcp-beacon.js
remyperona: It's temporary to get debug data, we will remove it when we go on prod

tests/Integration/inc/Engine/Media/AboveTheFold/AJAX/Subscriber/addLcpData.php
remyperona: It's used in the parent class, when calling the ajaxAction

Miraeld (approved):

PR #6520, opened 2024-03-26
316 bug front end optimizations are not applied even when lcpatf data is in the db

MathieuLamiot:
# Description

Fix a condition in has_lcp.
Make sure to return only one LCP and store it as a single JSON object rather than an array of objects so that Front-end optim. can be applied.

Fixes #6519 

### User documentation

This PR allows the front-end otpimization to be applied for 3.16

## Type of change
*Delete options that are not relevant.*

- [ ] New feature (non-breaking change which adds functionality).
- [x] Bug fix (non-breaking change which fixes an issue).
- [ ] Enhancement (non-breaking change which improves an existing functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as before).

## New dependencies
NA

## Risks
NA

# Checklists

## Feature validation

- [x] I validated all the Acceptance Criteria. If possible, provide sreenshots or videos.
- [ ] I triggered all changed lines of code at least once without new errors/warnings/notices.
- [ ] I implemented built-in tests to cover the new/changed code.

## Documentation

- [x] I prepared the user documentation for the feature/enhancement and shared it in the PR or the GitHub issue.
- [x] The user documentation covers new/changed entry points (endpoints, WP hooks, configuration files, ...).
- [x] I prepared the technical documentation if needed, and shared it in the PR or the GitHub issue.

## Code style
- [x] I wrote self-explanatory code about what it does.
- [x] I wrote comments to explain why it does it.
- [x] I named variables and functions explicitely.
- [x] I protected entry points against unexpected inputs.
- [x] I did not introduce unecessary complexity.
- [x] I listed the introduced external dependencies explicitely on the PR.
- [x] I validated the repo-specific guidelines from CONTRIBUTING.md.

## Observability
- [x] I handled errors when needed.
- [x] I wrote user-facing messages that are understandable and provide actionable feedbacks.
- [x] I prepared ways to observe the implemented system (logs, data, etc.).

## Risks
- [x] I explicitely mentioned performance risks in the PR.
- [x] I explicitely mentioned security risks in the PR.

PR #6544, opened 2024-04-04
Fixes new PHPCS error RequireExplicitBooleanOperatorPrecedence.MissingParentheses

remyperona:
# Description

Fixes new error now included in WPCS 3.1

### Technical documentation

Error description: 
> Mixing different binary boolean operators within an expression without using parentheses to clarify precedence is not allowed.

To fix, I added parentheses where needed to make the mixed conditions more explicit.

## Type of change

# Checklists

## Code style

### DIFF
--- a/inc/Engine/Common/ExtractCSS/Subscriber.php
+++ b/inc/Engine/Common/ExtractCSS/Subscriber.php
@@ -72,11 +72,13 @@ class Subscriber implements Subscriber_Interface, LoggerAwareInterface {
 
 		foreach ( $link_styles as $style ) {
 			if (
-				! (bool) preg_match( '/rel=[\'"]?stylesheet[\'"]?/is', $style[0] )
-				&&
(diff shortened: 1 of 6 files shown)


MathieuLamiot:
I just realized that we faced this issue last week and fixed it in [this commit](https://github.com/wp-media/wp-rocket/pull/6520/commits/f488281de0de219b5db95ce5f29bc167fea903ad).
It's briefly mentioned on Slack [here](https://wp-media.slack.com/archives/CUT7FLHF1/p1711638821146859?thread_ts=1711569653.252079&cid=CUT7FLHF1) but we mostly discussed it on the daily while you were off I think, sorry for the miss :/

Let's just make sure there is no conflict because we fixed it twice and that we fixed it the same way (it's easy to make a mistake with precedence, at least for me!)

MathieuLamiot (approved):

MathieuLamiot:
@Tabrisrp I think it can be merged on develop and then we can update the feature branch. When this is done, can you maybe post a message on #wp-rocket-dev to warn everyone to keep their own branches up to date? Thanks 🙏

PR #6558, opened 2024-04-15
Add coverage reports on all test runs

MathieuLamiot:


MathieuLamiot:
@Tabrisrp I noticed the code_coverage action was broken, probably due to new WP latest version. With Q2 OKRs, we'll keep focusing on built-in tests and I am looking into measuring code coverage (mostly diff code coverage, so code coverage only on the modified lines per PR). I think this CI refacto is needed to:
- Get code coverage on unit & integration tests (we only had unit)
- Keep the CI fast: coverage report on all integration tests took >30 min. So here it runs only on the "general" tests
- Get code coverage for more PRs, not only the ones to develop.

Let me know what you think

remyperona (approved):

Issue #6429, opened 2024-02-02
3.16 - Preload when enabling and updating WP Rocket, when clearing data, changing themes or permalinks


Miraeld:
### Context
As a user, I want WP Rocket to identify internal links on the homepage so that those pages can then be pre-optimized for LCP/ATF.

#### add a Controller class, with a dependency on the `Jobs\Manager` class
- add a method `fetch_links` that will send a request to the homepage, get the HTML, find the first 10 internal links, and send them to the jobs manager
- the number of links is modifiable with a new filter `rocket_atf_warmup_links_number`
- the method will bail-out if the filter to disable the ATF optimization feature is disabled
- add tests

#### add a Subscriber class, with a dependency on the Controller created above
- The subscriber will listen to the following actions: `wp_rocket_upgrade`, `permalink_update`, `switch_theme`
- For each one, a callback method will be used to do the fetch links
- The callback for `wp_rocket_upgrade` should only trigger when upgrading from a version lower than 3.16
- add tests

#### add `WP_Rocket\Engine\Media\AboveTheFold\Activation\Activation` & `WP_Rocket\Engine\Media\AboveTheFold\Activation\ServiceProvider`
- The Activation class will have a dependency on the controller
- add callback on this action: `rocket_after_activation`
- the callback method will call the fetch_links() method from the controller

### Acceptance criteria 
_template to be refined with QA_
- Update the plugin to 3.16 from a previous version. Check the DB against the expected result.
- Perform a fresh install. Check the DB against the expected result.

- Bail out from this process when RUCSS is enabled.

###  Effort estimation:
Effort M

### Additional information

To whoever picks up this issue, it could make sense to divide it in 3 successive PRs:

- This issue: First one to manage the fetch_link method on updatE/activation/install but without sending the results to the SaaS yet -> AC would be to check that on update/activation/install, we find the correct links from the homepage. Instead of sending them to the SaaS yet, they could be outputted as logs for testing purposes.
- https://github.com/wp-media/wp-rocket/issues/6470 : Second one would be to replace the log by actually sending to the SaaS, with the appropriate query strings (see https://github.com/wp-media/wp-rocket/issues/6441)
- https://github.com/wp-media/wp-rocket/issues/6471 : Third one would be to change the request to the SaaS for RUCSS by adding config["optimization_list"]=['rucss'] to the payload.

MathieuLamiot:
> preload le homepage and extract internal links

[...]

PR #6563, opened 2024-04-16
Fix/update default fetched links to 10 and parse unappended url to is home

jeawhanlee:
# Description
Fetches 10 links from home plus home(11) and also fixes the is_home not detecting the homepage.

Fixes: Addresses @hanna-meda comment [here](https://github.com/wp-media/wp-rocket/issues/6429#issuecomment-2056995658) and [here](https://github.com/wp-media/wp-rocket/issues/6429#issuecomment-2058331751)

### User documentation

### Technical documentation
https://github.com/wp-media/wp-rocket/blob/38775e5e095d57a01aa9a1255f290db9197c5d66/inc/Engine/Media/AboveTheFold/WarmUp/APIClient.php#L17
We now save the original url before appending the query strings then reuse the version in `is_home` method
Also we now filter out the home url from the links if it exists and append in the end just before returning the urls, reason for this is to always return the home url for templates that might not contain the home url.

## Type of change

## New dependencies
None

## Risks

# Checklists

## Feature validation

## Documentation

## Code style

## Observability

## Risks


inc/Engine/Media/AboveTheFold/WarmUp/Controller.php
```diff
@@ -130,7 +131,7 @@ function ( $link ) use ( $home_url ) {
 				 * Check for valid link.
 				 * Check that no external link.
 				 */
-				return wp_http_validate_url( $link ) && $link_host['host'] === $site_host['host'];
+				return wp_http_validate_url( $link ) && $link_host['host'] === $site_host['host'] && ! Utils::is_home( $link );
```
remyperona: triple conditions is hard to read, maybe we could split that?

inc/Engine/Media/AboveTheFold/WarmUp/APIClient.php
```diff
@@ -24,7 +26,7 @@ public function add_to_atf_queue( string $url ): array {
 
 		$config = [
 			'optimization_list' => '',
-			'is_home'           => Utils::is_home( $url ),
+			'is_home'           => $is_home,
```
remyperona, line 29: Is this change necessary?

remyperona (changes requested):

inc/Engine/Media/AboveTheFold/WarmUp/APIClient.php
jeawhanlee: Yes I believe it is, in reference to @hanna-meda comment [here](https://github.com/wp-media/wp-rocket/issues/6429#issuecomment-2058331751)

remyperona: I'm failing to understand the difference it makes, it's the same code assigned to a variable?

jeawhanlee: Yes but we are now preserving the original state of the `url` param which was not the case before as we alter the state [here](https://github.com/wp-media/wp-rocket/blob/38775e5e095d57a01aa9a1255f290db9197c5d66/inc/Engine/Media/AboveTheFold/WarmUp/APIClient.php#L18) first then we use it [here](https://github.com/wp-media/wp-rocket/blob/38775e5e095d57a01aa9a1255f290db9197c5d66/inc/Engine/Media/AboveTheFold/WarmUp/APIClient.php#L27) which in the `is_home` method [here](https://github.com/wp-media/wp-rocket/blob/132a9f0d69483020ba41f290a624d10168ac014a/inc/Engine/Common/Utils.php#L24) we will always return false even if it's the home page because it's now appended with a query string i.e `example.org?wpr_imagedimensions=1`

remyperona: ah yes right, I see now, all good

remyperona (approved):

PR #6567, opened 2024-04-17
Fix container fatal errors with activation and deactivation

wordpressfan:
# Description

@Mai-Saad reported a fatal error with lcp branch and after checking I found that the issue also happens with develop, it's a leftover from this PR: #6542 

This fatal error was happening with activation and deactivation of WPR.


Khadreal (approved):
LGTM

PR #6559, opened 2024-04-16
Closes #6530 #6531 #6532 #6534 #6535 #6552: Preload images markup

Miraeld:
# Description

Fixes #6530 #6531 #6532 #6534 #6535 #6552 

This PR addresses several issues related to image preloading, specifically for images detected as Largest Contentful Paint (LCP) and Above The Fold (ATF) images. The changes ensure that various types of images are correctly preloaded and excluded from lazy loading to improve page load performance. Additionally, it fixes an issue where background images identified as the LCP were incorrectly labeled as 'img' in the database, they are now correctly labeled dynamically.

### User documentation

This update improves the page load performance by correctly preloading various types of images and excluding them from lazy loading when they are detected as Largest Contentful Paint (LCP) and Above The Fold (ATF) images. This means faster loading times for your website visitors.

### Technical documentation

The code works by identifying images that are LCP and ATF, and then preloading these images to improve page load performance. It also correctly labels background images identified as LCP with the corrected type in the database. The changes are implemented across several files, including JavaScript for front-end changes and PHP for back-end changes.

## Type of change

- [x] Enhancement (non-breaking change which improves an existing functionality).

## New dependencies
n/a

## Risks
n/a

# Checklists

## Feature validation

- [x] I validated all the Acceptance Criteria. If possible, provide screenshots or videos.
- [x] I triggered all changed lines of code at least once without new errors/warnings/notices.
- [x] I implemented built-in tests to cover the new/changed code.

## Documentation

- [x] I prepared the user documentation for the feature/enhancement and shared it in the PR or the GitHub issue.
- [x] The user documentation covers new/changed entry points (endpoints, WP hooks, configuration files, ...).
- [x] I prepared the technical documentation if needed, and shared it in the PR or the GitHub issue.

## Code style

- [x] I wrote self-explanatory code about what it does.
- [x] I wrote comments to explain why it does it.
- [x] I named variables and functions explicitly.
- [x] I protected entry points against unexpected inputs.
- [x] I did not introduce unnecessary complexity.
- [x] I listed the introduced external dependencies explicitly on the PR.
- [x] I validated the repo-specific guidelines from CONTRIBUTING.md.

PR #6579, opened 2024-04-22
Enhancement/preload background image atf

### LINKED ISSUE #6537: ATF isnot detected for background images in certain cases (with attribute markup, inline, multi markup)
- Made sure you’re on the latest version => 3.16_alpha1

**Describe the bug**
When BG image is added using attribute markup, ATF is empty

**To Reproduce**
1. 3.16 installed and activated
2. page template with bg added using attribute CSS
3. visit the page 
4. check  ATF table => ATF is empty
Screenshot from 2024-04-02 09-01-25

**Expected behavior**
Any image in the viewport that is not lcp is added to atf in the database

**Screenshots**

**Additional context**
- test page: https://new.rocketlabsqa.ovh/lcp_attributes

Khadreal:
# Description
Improve beacon to capture background images for ATF
Fixes #6537 

### User documentation

### Technical documentation

## Type of change

## New dependencies

## Risks

# Checklists

## Feature validation

## Code style


assets/js/lcp-beacon.js
```diff
@@ -1,152 +1,217 @@
...
-    window.performance_images_json = performance_images_json;
-    console.log(performance_images_json);
+function main() {
+	// Use LCPCandidates function to get all the elements in the viewport
+	const above_the_fold_images = LCPCandidates(Infinity);
+
+	if (above_the_fold_images.length !== 0) {
+		// Get the first element in the viewport and use as LCP
+		performance_images = [{
+			...above_the_fold_images[0].elementInfo,
```
Miraeld: I don't think this will work. 
We can't take the first element of `above_the_fold_images` as it isn't always an element with an image. So we now get many LCPs as `null`. 

If I take a concrete example, on [this page](https://e2e.rocketlabsqa.ovh/lcp_bg_inline_template/) we get null for the LCP because `above_the_fold_images` contains the following: 
```
0: {element: div.wrap, elementInfo: null}
1: {element: div#primary.content-area, elementInfo: null}
2: {element: main#main.site-main, elementInfo: null}
3: {element: div, elementInfo: Object}
4: {element: div, elementInfo: Object}
5: {element: p, elementInfo: null}
```

As you sort the results by their areas [here](https://github.com/wp-media/wp-rocket/pull/6579/files#diff-9f86fce58cce5e3f8244c7d9d080582dd4b3c6483896fba8a4ef73eabc5e5381R56) that's normal that `div.wrap` is first, as it wrap all content of the page. But that's not the div containing images.

Miraeld: And while thinking about it, 

We should modify your whole condition by:
``` javascript
	const firstElementWithInfo = above_the_fold_images.find(item => item.elementInfo !== null);

	if (firstElementWithInfo) {
		performance_images = [{
			...firstElementWithInfo.elementInfo,
			label: "lcp",
		}];
	} else {
		console.log("No LCP candidate found.");
	}
```

Miraeld (approved):

MathieuLamiot:
@Khadreal The only thing to test here for the QA step would be to ensure the filter can't break the JS. Template testing is ensured with #6559 

From what I see you guard against non-array arguments so that's good. Also we are sure we return a string since you use implode.
My concern is: what is someone passes elements in the array that are not strings? like [123, 456]? did you check that implode won't throw an error, and that the JS won't throw an error?
What could happen is that we pass something wrong to querySelectorAll ?

image

Once this is tested, you can update here and put to QA Done if it passes well :) 

In case you don't know how to test, I advise to use "Code Snippets" plugin to inject code easily:
image

remyperona:
I updated the method on commit https://github.com/wp-media/wp-rocket/pull/6311/commits/764aaf255a02cdf74bd612262b5274784e6ebd6f

- switched it to private
- updated the description
- updated the filter name to match the formatting of others filters 
- filters the array to only keep strings before using it

PR #6596, opened 2024-04-26
Closes #6591: LCP/ATF data is malformed when no LCP is found by the beacon

### LINKED ISSUE #6591: LCP/ATF data is malformed when no LCP is found by the beacon
- From feature/lcp-above-the-fold-optimization on commit 91077f9ce8caa90e4c062661620d5eea48d69c21

**Describe the bug**
When no LCP is found by the beacon, the LCP value in the DB is `null` instead of `not found`. This makes the `has_lcp` check not working as expected and as a result, the PHP code throws some errors and warning, trying to call some stuff on null.

**To Reproduce**
See video [here](https://wp-media.slack.com/archives/CUT7FLHF1/p1714046628790889?thread_ts=1714044259.441869&cid=CUT7FLHF1)
- Go on your E2E website.
- Add this in your wp-config.php (to target staging SaaS, might work without it too): `define( 'WP_ROCKET_SAAS_API_URL', 'https://rucss-director-staging.public-default.live2-k8s-cph3.one.com/' );`
- Install the plugin (or do Clear Critical Images).
- Visit http://xxx.e2e.rocketlabsqa.ovh/category/uncategorized/ to gather the LCP/ATF. There is no LCP.
- check in the DB, the LCP for this page is null
- Clear the cache.
- Re-visit the page in incognito. The error should be in the log.

debug logs:

```
[25-Apr-2024 11:58:30 UTC] PHP Warning:  Undefined property: stdClass::$type in /var/www/html/mathieu.e2e.rocketlabsqa.ovh/wp-content/plugins/wp-rocket/inc/Engine/Media/AboveTheFold/AJAX/Controller.php on line 103
[25-Apr-2024 11:58:30 UTC] PHP Warning:  Undefined property: stdClass::$type in /var/www/html/mathieu.e2e.rocketlabsqa.ovh/wp-content/plugins/wp-rocket/inc/Engine/Media/AboveTheFold/AJAX/Controller.php on line 105
```

**Expected behavior & AC**
- No errors should be thrown when re-visiting the page.
- When LCP is not found, in the DB, it should be 'not found' instead of null. 
- When no WTf found, in the DB, it should be []. (Currently seems OK)
- We must have a test (maybe E2E if integration can't run the JS?) on template without LCP.

**Additional context**
The issue seems to be with recent merges in the beacon script. I suspect there are 2 blocks of code handling LCP and one should not be here:

```
	// Filter the array based on the condition imageURL is not null
	const filteredArray = LCPCandidates(1)
	if (filteredArray.length !== 0) {
		console.log("Estimated LCP element:", filteredArray);
		performance_images = filteredArray.map((item) => ({
			src: item.imageURL,
			label: "lcp",
		}));
	} else {
		console.log("No LCP candidate found.");
	}

	// Use LCPCandidates function to get all the elements in the viewport
	const above_the_fold_images = LCPCandidates(Infinity);

	const firstElementWithInfo = above_the_fold_images.find(item => item.elementInfo !== null);

	if (firstElementWithInfo) {
		performance_images = [{
			...firstElementWithInfo.elementInfo,
			label: "lcp",
		}];
	} else {
		console.log("No LCP candidate found.");
```

*MOVED BACK TO TODO WITH A DEEPER INVESTIGATION, SEE BELOW*

Miraeld:
# Description

Fixes #6591

### Technical documentation

I went on the approach to only encode LCP data if this is an array or an object. 
As our `not found` was a string, it was encoded in JSON so we were getting a `"not found"` in the DB while we were checking if this was `=== 'not found'`. That couldn't work. 

Also, I've added a safeguard to make sure we are processing valid lcp data on the front end controller.

And finally, I've added a test to cover this case.

(I based my branch on [Remy's branch](https://github.com/wp-media/wp-rocket/pull/6595) to avoid conflict.)

## Type of change

## Risks

# Checklists

## Feature validation

## Code style

## Observability
- [x] I handled errors when needed.
- [x] I wrote user-facing messages that are understandable and provide actionable feedbacks.
- [x] I prepared ways to observe the implemented system (logs, data, etc.).


Miraeld:
@MathieuLamiot , you'll find two video showing a test in local + on a server (e2e). 

https://github.com/wp-media/wp-rocket/assets/14993267/d310cecc-7bdc-46dd-ba57-ca3c98021bc8

https://github.com/wp-media/wp-rocket/assets/14993267/330b20ca-9b9a-4969-86ef-0ce9ff90d6d6

MathieuLamiot (changes requested):
@Miraeld When using this branch, I have a message from the beacon when browsing a page without LCP: 

`{success: false, data: 'error when adding the entry to the database'}`

I don't understand why you bail out and return null from create_object, but don't catch this case before assigning the returned value to $lcp?
As a result, I think we get a null for LCP and the DB complains.

Miraeld:
I've pushed a modification: 
- I've deleted the bail out, 
- I prevent going through `create_object()` if there is no `type` in the object `image`. 
- I've also refactored the code to make it clearer and simpler.

MathieuLamiot (approved):

PR #6603, opened 2024-04-30
3.16 - Handle different errors from the beacon script

### LINKED ISSUE #6570: 3.16 - Error & timeout handling of Beacon Script
**Context**
https://wp-media.slack.com/archives/CUT7FLHF1/p1713349928163759?thread_ts=1713250509.830419&cid=CUT7FLHF1

Currently, if the script throws an error, or if it never completes, then we will keep injecting it and it could keep failing/timing out forever.
We have not encountered a case where this happened, so this is a bullet-proofing approach and there is no way to naturally reproduce this without enforcing an error/time-out in the script itself.

**Expected behavior**
- If the beacon script throws an error, it must be handled this info must be sent to the AJAX endpoint.
- If the beacon script takes longer than 10s to execute, this info must be sent to the AJAX endpoint.
- The AJAX endpoint must handle failure payloads (error thrown and timeout) and register this in the DB, similarly to a valid payload but with the failure status and the related error message.

As a result, if the row related to the current URL has status failed, then the beacon script won't be injected and if it is already injected, it will bail out early in the process.

**Acceptance Criteria**
- Modify the script to throw an error.
- Clear critical images and cache, enable cache.
- Visit a page. The beacon must be injected, and the critical image table must store a row with a failed status and the error_message.
- Re-visit the page. The beacon must be there but it should not throw any errors as it bails out early.
- Clear the cache. Re-visit the page. The beacon must not be injected.

- Modify the script to enter an infinite loop..
- Clear critical images and cache, enable cache.
- Visit a page. The beacon must be injected, and the critical image table must store a row with a failed status and the error_message mentioning time out, after ~10 seconds.
- Re-visit the page. The beacon must be there but it should not throw any errors as it bails out early.
- Clear the cache. Re-visit the page. The beacon must not be injected.

**Additional information**
- @gmetais, could you check this issue, how to implement it or review the grooming if already done?

Implementation suggestions:
- For error handling, I think we could:
    - add try/catch to wrap-up all the logic in the beacon script.
    - In the catch, send a payload to the AJAX endpoint with status "failed" and the error_message.
    - In the AJAX endpoint, handle this case to store it in the DB
- For time out, we could use setTimeout at the beginning of the script, and closeTimeout at the end of it. In case the timeout is triggered, we'd send an AJAX call with the failed status.

- One case won't be covered here: if the error happens when sending the failure data to the AJAX, or if we bail out in the AJAX endpoint itself. I think this is acceptable.

wordpressfan:
# Description

Fixes #6570

### User documentation

Now we handle two different scenarios:

1. Script timeout: If for any reason, the script took more than 10 seconds, this will fire the ajax call to save failed row in the DB not to load the script again, to simulate this you can use the following JS code:
```
function wait(ms){
   var start = new Date().getTime();
   var end = start;
   while(end < start + ms) {
     end = new Date().getTime();
  }
}
wait( 15000 );// This will wait for 15 seconds.
```

Exactly here:

https://github.com/wp-media/wp-rocket/blob/5fc1fd42f7bb4d7915fbe961da7acfe899dc7ce0/assets/js/lcp-beacon.js#L26

2. Script error: Also if for any reason we got an error from the script, we will send the same ajax request to set a failed row in the DB.
You can simulate this by putting any strange line exactly inside the try block here:

https://github.com/wp-media/wp-rocket/blob/5fc1fd42f7bb4d7915fbe961da7acfe899dc7ce0/assets/js/lcp-beacon.js#L16-L21

### Technical documentation

## Type of change

## New dependencies

## Risks

# Checklists

## Feature validation

## Documentation

## Code style

## Observability

## Risks


MathieuLamiot:
Thanks for the PR!
Looking at the code, I don't understand how the script deals with the AC of "infinite loop"? It looks like there is no timeout set and that we only check the execution time at the end of the script (if we reach it). Am I missing something or is there any reasons to not cover this? Thanks

MathieuLamiot:
Thanks for the update. Do you think a closeTimeout would be beneficial too, to avoid a timeout running in the background when it's not needed anymore?

wordpressfan:
> Thanks for the update. Do you think a closeTimeout would be beneficial too, to avoid a timeout running in the background when it's not needed anymore?

That's a good idea, I pushed now a code that uses clearTimeout function

MathieuLamiot (approved):

remyperona (approved):

MathieuLamiot:
Tests looks OK to me to catch errors. Infinite loop is not prevents if we brutally do while(true) bu that's probably expected as the event loop is not freed? (or something like this). I'd say we should go ahead and merge.

PR #6610, opened 2024-05-03
Closes #6571: Reduce the processing on not-cached page when checking the LCP/ATF data

### LINKED ISSUE #6571: Reduce the processing on not-cached page when checking the LCP/ATF data
On not cached page we're checking if LCP/ATF data exists twice:
1. For the first time when trying to inject the data
2. For the second time when beacon script is triggered

Instead of checking it for the second time, we could bail-out if we're on not cached page. That's potentially an improvement in terms of resource usage.

**Acceptance Criteria**
1. When the page is not cached yet, beacon script should not check for existence of the LCP/ATF data in the db
2. When the page is cached, beacon script should check if the data exist 
3. If we have data related to this URL (and device), we should bail-out
4. If we don't have data related to this URL (and device), we should proceed

Miraeld:
# Description

Fixes #6571 

### User documentation

It won't try to check if LCP data are already existing for non-cached URLs.

### Technical documentation

We check if the page is a cached version, if it isn't we will bailout before sending an AJAX call to check if LCP datas are already existing for the URL.

## Type of change

## New dependencies
N/a
## Risks
N/a

# Checklists

## Feature validation

## Documentation

- [x] I prepared the user documentation for the feature/enhancement and shared it in the PR or the GitHub issue.
- [x] The user documentation covers new/changed entry points (endpoints, WP hooks, configuration files, ...).
- [x] I prepared the technical documentation if needed, and shared it in the PR or the GitHub issue.

## Code style


assets/js/lcp-beacon.js
```diff
@@ -39,26 +39,44 @@ class RocketLcpBeacon {
...
 	}
 
-	_isGeneratedBefore() {
+	_isPageCached() {
+		let signature = '';
+		let status = false;
+		if (document.documentElement.nextSibling)  {
+			signature = document.documentElement.nextSibling.data;
+		}
+		if ( signature && signature.includes( 'Debug: cached' ) ) {
```
jeawhanlee: I think we should check for something more reliable maybe like the whole sentence as there are some cases where we do not generate timestamp in the footprint.

assets/js/lcp-beacon.js
```diff
@@ -39,26 +39,44 @@ class RocketLcpBeacon {
...
+			}
 		}
 
 		return true;
 	}
 
-	_isGeneratedBefore() {
+	_isPageCached() {
+		let signature = '';
+		let status = false;
```
jeawhanlee:
```suggestion
		let signature = '', status = false;
```

assets/js/lcp-beacon.js
```diff
@@ -310,9 +328,9 @@ class RocketLcpBeacon {
 	}
 
 	_logMessage( msg ) {
-		if ( ! this.config.debug ) {
-			return;
-		}
+		// if ( ! this.config.debug ) {
+		// 	return;
+		// }
```
jeawhanlee: Why do we have this commented out, we should remove the lines if they are not useful.

assets/js/lcp-beacon.js
```diff
@@ -39,26 +39,44 @@ class RocketLcpBeacon {
...
 		}
 
-		if ( this._isGeneratedBefore() ) {
-			this._logMessage('Bailing out because data is already available');
-			return false;
+		if ( this._isPageCached() ) {
+			if ( this._isGeneratedBefore() ) {
+				this._logMessage('Bailing out because data is already available');
+				return false;
+			}
```
jeawhanlee:
```suggestion
		if ( this._isPageCached() && this._isGeneratedBefore() ) {
			this._logMessage('Bailing out because data is already available');
			return false;
```

jeawhanlee (changes requested):

assets/js/lcp-beacon.js
```diff
@@ -39,26 +39,44 @@ class RocketLcpBeacon {
...
+		await fetch(this.config.ajax_url, {
 			method: "POST",
 			credentials: 'same-origin',
 			body: data_check
+		})
+		.then(data => {
+			lcp_data_response = data
 		});
 
 		return lcp_data_response.success;
```
wordpressfan:
```suggestion
		const lcp_data_response = await fetch(this.config.ajax_url, {
			method: "POST",
			credentials: 'same-origin',
			body: data_check
		})
		.then(data => data.json());

		return lcp_data_response.success;
```

wordpressfan (changes requested):
We need to bear in mind that in some hosts our cache is disabled by default also the signature can be removed in cases like when having the whitelabel constant or so.

I need to understand more how do u detect the cached page in depth.

assets/js/lcp-beacon.js
Miraeld: I used the same condition as on WPR Inspector (https://github.com/wp-media/wpr-inspector/blob/bd71796388de3c9cc0cff45488de23ce920df6d1/wpr-inspector/inspector.js#L25)

Miraeld:
@wordpressfan for these cases, I guess we won't have the choice to do the DB request we want to avoid which is the one checking if LCP data does exists for the page while it isn't even cached yet. 

We can't check through the DB as it would add another DB Request, which is the opposite of what we want, and there are no other way to determine if a page is cached or not. Except if there is something else I'm missing. 

The way the current detection work is through the signature, where if the page is cached, we get the following text: `Debug: cached@1714733510` 
Where `1714733510` is a timestamp.

Miraeld:
@jeawhanlee & @wordpressfan  Are you okay with the latest changes ?

jeawhanlee:
For me I have issues with how reliable it can be detecting a cached page with our footprint signature, what if we tried checking the directory if the cache file exists and pass the result (boolean) to the script.

jeawhanlee:
> For me I have issues with how reliable it can be detecting a cached page with our footprint signature, what if we tried checking the directory if the cache file exists and pass the result (boolean) to the script.

The above won't work, when the page is cached, it would bail out and render the cache. 

In my opinion, If we would use the signature, we should use parts that must always be present instead of just `Debug: cached`
Here is an example of a page without the timestamp:
Screenshot 2024-05-11 at 21 18 08

MathieuLamiot:
@Miraeld Looking at wp-rocket's code, there are versions of the code that don't have the timestamp in the signature I think. Typically in case of 3rd party hosting, white label and if time is empty.
In that case, it would make sense indeed to look for:
- '<!-- Optimized for great performance'
OR
- '<!-- This website is like a Rocket, isn\'t it? Performance optimized by  '

Then, of course, it is not perfect because sometimes, comments are removed from cached files. But better done than perfect.

WDYT of matching on those two rather than Debug + timestamp? In any cases, better done than perfect: Can we just change the matching to those 2 strings, move forward with this PR and eventually open a follow-up GH issue if there are edge cases you still foresee?

Miraeld:
So based on our conversation yesterday, I guess we have to merge this PR and prepare another issue to enhance it, 
@wordpressfan as you had the idea for the enhancement, can you create the new issue, and when it's done can you merge this PR ?

MathieuLamiot:
@wordpressfan @jeawhanlee Since @Miraeld will be off end of this week, can you make sure to close this? Thanks 🙏

assets/js/lcp-beacon.js
```diff
@@ -39,28 +39,39 @@ class RocketLcpBeacon {
...
-	_isGeneratedBefore() {
+	_isPageCached() {
+		let signature = '', status = false;
+		if (document.documentElement.nextSibling)  {
+			signature = document.documentElement.nextSibling.data;
+		}
+		if ( signature && signature.includes( 'Debug: cached' ) ) {
+			return true;
+		}
+		return false;
```
wordpressfan:
```suggestion
		const signature = document.documentElement.nextSibling.data ? document.documentElement.nextSibling.data : '';

		return signature && signature.includes( 'Debug: cached' );
```

wordpressfan (approved):
Approved with very minor enhancement

MathieuLamiot:
Expected behavior and observation from QA team discussed here: https://wp-media.slack.com/archives/CUT7FLHF1/p1716202320587489

Mai-Saad:
The issue mentioned here https://github.com/wp-media/wp-rocket/pull/6610#issuecomment-2122910431 is fixed, but on PR, nothing is added to the database with warmup (neither after a fresh install, nor after clear critical images)

steps:
1-fresh install of PR => check ATF table, nothing added although WPR debug contains some URLs to add to queue
2- clear critical images => check ATF, still nothing is added to database

Note: related test plan https://wpmediaqa.testrail.io/index.php?/runs/view/856&group_by=cases:section_id&group_order=asc

wordpressfan:
All jobs from that site in SaaS are returning timeout error, I'm checking with Mostafa now what is going on there.

wordpressfan:
This was because in SaaS we load the not cached page which doesn't include the signature and this means that the following line returns js error in the console:

https://github.com/wp-media/wp-rocket/blob/a6b762284692de00b1ef67a1ae24d52a67551c42/assets/js/lcp-beacon.js#L51

because `document.documentElement.nextSibling` is null in this case, pushing a fix now.

Mai-Saad (approved):
Working as expected 
[testrail-report-605.pdf](https://github.com/wp-media/wp-rocket/files/15416918/testrail-report-605.pdf)

PR #6615, opened 2024-05-03
Fixes #6613 Improve LCP warmup process

### LINKED ISSUE #6613: Improvements to homepage fetching process
In 3.16 we introduced the feature that's fetching the links from the homepage. Currently, this process is happening synchronously, whenever the critical images data is cleared.

It's causing potential problems on the slower environments, as we need to wait until the homepage is loaded and fetched synchronously. It can even cause timeouts.

On our testing environment, clearing the data took 13s compared to 2.4s when putting static data into fetch_links() function.

The further findings are:
1. We're sending the blocking requests to saas also when debug mode is disabled
2. The usleep between each request is set to 0.5s
3. We're doing two requests to collect the links (one using mobile UA to collect mobile device links and the second one using desktop UA to collect desktop links)

1. Visit the homepage only once using mobile and populate all (desktop/mobile) links based on that
2. Reduce the usleep when debug mode is disabled (the exact value needs discussion here and dev direction)
3. Make the requests not blocking when the debug mode is disabled

**Acceptance Criteria**
1. When fetching the links for desktop and mobile, we should do only one, mobile request. The links will be shared between mobile and desktop
2. When debug mode is disabled, requests sent to the saas should not be blocking
3. When debug mode is enabled, requests sent to the saas should be blocking te receive the log data
4. When debug mode is disabled, we should reduce the usleep to minimum
5. When debug mode is enabled, we should reduce the usleep to be lower than it's currently
6. It would be great to measure time before/after to confirm the impact

remyperona:
# Description

Update the LCP warmup process by:
- Doing only a request on the mobile homepage
- Reducing the delay between each URL sent to the SaaS
- Making the requests non-blocking

Fixes #6613

## Type of change

# Checklists

## Feature validation

## Code style

### DIFF
--- a/inc/Engine/Common/JobManager/APIHandler/APIClient.php
+++ b/inc/Engine/Common/JobManager/APIHandler/APIClient.php
@@ -47,16 +47,21 @@ class APIClient extends AbstractAPIClient implements LoggerAwareInterface {
 		 */
 		$url = apply_filters( 'rocket_saas_api_queued_url', $url );
 
-		$blocking = $options['blocking'];
+		$blocking = true;
 
-		unset( $options['blocking'] );
+		if ( isset( $options['blocking'] ) ) {
+			$blocking = $options['blocking'];
+
+			unset( $options['blocking'] );
+		}
 
 		$args = [
 			'body'    => [
 				'url'    => $url,
 				'config' => $options,
 			],
-			'timeout' => 5,
+			'timeout'  => 5,
+			'blocking' => true,
 		];
 
 		if ( ! $blocking ) {
--- a/inc/Engine/Common/JobManager/APIHandler/APIClient.php
+++ b/inc/Engine/Common/JobManager/APIHandler/APIClient.php
@@ -56,7 +56,7 @@ class APIClient extends AbstractAPIClient implements LoggerAwareInterface {
 		}
 
 		$args = [
-			'body'    => [
+			'body'     => [
 				'url'    => $url,
 				'config' => $options,
 			],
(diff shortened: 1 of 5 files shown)


Miraeld (approved):

inc/Engine/Common/JobManager/APIHandler/APIClient.php
```diff
@@ -47,14 +47,28 @@ public function add_to_queue( string $url, array $options ): array {
 		 */
 		$url = apply_filters( 'rocket_saas_api_queued_url', $url );
 
+		$blocking = true;
+
+		if ( isset( $options['blocking'] ) ) {
+			$blocking = $options['blocking'];
+
+			unset( $options['blocking'] );
+		}
```
wordpressfan, line 56: When the `WP_ROCKET_DEBUG` is not set at all, so it'll be null and we will not get into this block at all then $blocking will equal "true" but we need it to be "false"

I think to solve this quickly we can change the default value for the constant to be false not null like:

```
rocket_get_constant( 'WP_ROCKET_DEBUG', false )
```

wordpressfan (changes requested):

inc/Engine/Common/JobManager/APIHandler/APIClient.php
remyperona: good catch, I updated the passed value in the APIClient

wordpressfan (approved):

jeawhanlee:
@Tabrisrp @wordpressfan This is totally out of scope for this PR but I figured since we already touched here, we might as well just do it. For easier debugging, can we add the url to the response log here:
https://github.com/wp-media/wp-rocket/blob/692321059af622320d983fbd0b79bfdf447fe69a/inc/Engine/Common/JobManager/APIHandler/APIClient.php#L103
Perhaps something like this: `https://example.org/lcp-page - Add to queue response body`

PR #6626, opened 2024-05-09
Closes #6604 Add fetch priority to LCP when url is relative

### LINKED ISSUE #70: Minify adds path to css rules with url() even if not needed
know that we don't handle the minification library, but I report the issue anyway.
With css and svg, you can use the mask rule with an url like this : url(#id)
the minification process seems to add the path to the theme folder even if not needed here, since it's not a file that is requested.

### LINKED ISSUE #6604: Fetchpriority of LCP is not applied on the LCP markup when the LCP URL is relative
- Made sure you’re on the latest version: Commit 1b6f1972a78649e317463b5c5eb8b2e3574052b5

**Describe the bug**
The LCP feature should add the fetchpriority attribute to the LCP markup.
When the LCP markup contains a relative URL, this is not done. Note that the preload markup is correctly added with an absolute URL and the fetchpriority.

**To Reproduce**
1. Make sure to have LCP feature enabled (by default, it is). 
2. Clear the cache and critical images.
3. Manually go to xxx.e2e.rocketlabsqa.ovh/lcp_regular_image_template/ in incognito. Make sure that the LCP/ATF table is filled with information from this page in the DB. The LCP image should be `wp-content/rocket-test-data/images/600px-Mapang-test.gif`
4. Purge this URL
5. Re-load the page.
6. Check in the dev tools the markup of `wp-content/rocket-test-data/images/600px-Mapang-test.gif` image. There is no fetchpriority attribute.

If you do the same on http://xxx.e2e.rocketlabsqa.ovh/landing-page/, the attribute is correctly applied.

I think the difference is the absolute/relative paths.

Note that we have similar issues when a CDN is being used: if the URL in the DB contains the CDN hostname, then the fetchpriority is not applied as expected when serving the page.

**Expected behavior**
- Fetchpriority attribute must be set to the LCP markup when the LCP markup contains an absolute URL ✅ 
- Fetchpriority attribute must be set to the LCP markup when the LCP markup contains a relative URL 🔴 
- Built-in tests should cover those cases.
If it make sense to tackle this in the same issue (nice to have, otherwise, create a dedicated GH issue):
- Fetchpriority attribute must be set to the LCP markup when the a CDN is used and the URL stored in the LCP/ATF table as LCP contains the CDN URL.

**Additional context**
The issue seems to be with `set_fetchpriority` which performs a regex based on the URL we have in the DB. Mismatch between relative/absolute/CDN hostname make the regex fail?

**Acceptance Criteria**
- Repeat the How to reproduce on the two templates: /lcp_regular_image_template/  & /landing-page/. In both cases, the fetchpriority attribute must be set properly to the LCP markup.
- NRT with Rocket-E2E
- Tests (unit or integration) covers the relative path case and the absolute path case.
If dealt with in this PR, the same ACs must be checked with the CDN feature enabled and LCP stored in DB with the CDN hostname.

Khadreal:
# Description

Fixes #6604 

## Documentation
Modify the regex pattern, to capture all possible cases, absolute, relative and other valid urls 
### User documentation

### Technical documentation

## Type of change

## New dependencies
N/A

## Risks

# Checklists

## Feature validation

## Code style

## Observability
- [x] I handled errors when needed.

### DIFF
--- a/inc/Engine/Media/AboveTheFold/Frontend/Controller.php
+++ b/inc/Engine/Media/AboveTheFold/Frontend/Controller.php
@@ -146,16 +146,18 @@ class Controller {
 			return $html;
 		}
 
-		$url = preg_quote( $lcp->src, '/' );
+		$url     = preg_quote( $lcp->src, '/' );
+		$pattern = '#<img(?:[^>]*?\s+)?src=["\']' . $url . '["\'](?:\s+[^>]*?)?>#';
+
 		if ( wp_http_validate_url( $lcp->src ) && ! $this->is_external_file( $lcp->src ) ) {
-			$url = preg_quote(
-				preg_replace( '#^(://|[^/])+#', '', $lcp->src ),
-				'/'
-			);
+			$url = get_rocket_parse_url( $lcp->src );
+			$url = preg_quote( $url['path'], '/' );
+
+			$pattern = '#<img(?:[^>]*?\s+)?src=["\'](?:https?:)?(?:\/\/(?:[^\/]+)\/?)?\/?' . $url . '["\'](?:\s+[^>]*?)?>#';
 		}
 
 		$html = preg_replace_callback(
-			'#<img(?:[^>]*?\s+)?src=["\']' . $url . '["\'](?:\s+[^>]*?)?>#',
+			$pattern,
 			function ( $matches ) {
 				// Check if the fetchpriority attribute already exists.
 				if ( preg_match( '/fetchpriority\s*=\s*[\'"]([^\'"]+)[\'"]/i', $matches[0] ) ) {
--- a/inc/Engine/Media/AboveTheFold/Frontend/Controller.php
+++ b/inc/Engine/Media/AboveTheFold/Frontend/Controller.php
@@ -146,6 +146,7 @@ class Controller {
 			return $html;
 		}
 
+		$html    = $this->replace_html_comments( $html );
 		$url     = preg_quote( $lcp->src, '/' );
 		$pattern = '#<img(?:[^>]*?\s+)?src=["\']' . $url . '["\'](?:\s+[^>]*?)?>#';
 		if ( wp_http_validate_url( $lcp->src ) && ! $this->is_external_file( $lcp->src ) ) {
@@ -173,7 +174,7 @@ class Controller {
 			1
 		);
 
-		return $html;
+		return $this->restore_html_comments( $html );
 	}
 
 	/**
--- a/inc/Engine/Media/AboveTheFold/Frontend/Controller.php
+++ b/inc/Engine/Media/AboveTheFold/Frontend/Controller.php
@@ -147,6 +147,12 @@ class Controller {
 		}
 
 		$url  = preg_quote( $lcp->src, '/' );
+		if( ! $this->is_external_file( $lcp->src ) ) {
+			$url = preg_quote(
+				preg_replace( '#^(://|[^/])+#', '', $lcp->src ),
+				'/'
+			);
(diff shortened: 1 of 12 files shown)


inc/Engine/Media/AboveTheFold/Frontend/Controller.php
```diff
@@ -146,9 +146,17 @@ private function set_fetchpriority( $lcp, string $html ): string {
 			return $html;
 		}
 
-		$url  = preg_quote( $lcp->src, '/' );
+		$url     = preg_quote( $lcp->src, '/' );
+		$pattern = '#<img(?:[^>]*?\s+)?src=["\']' . $url . '["\'](?:\s+[^>]*?)?>#';
+		if ( wp_http_validate_url( $lcp->src ) && ! $this->is_external_file( $lcp->src ) ) {
+			$url = get_rocket_parse_url( $lcp->src );
```
wordpressfan: So here we only need the path? 

```suggestion
			$path = wp_parse_url( $lcp->src, PHP_URL_PATH );
```

inc/Engine/Media/AboveTheFold/Frontend/Controller.php
```diff
@@ -146,9 +146,17 @@ private function set_fetchpriority( $lcp, string $html ): string {
...
+			$url = preg_quote( $url['path'], '/' );
```
wordpressfan: why do we need to use preg_quote again? we already used it above with the whole url and I believe this is enough, what do u think?

wordpressfan (changes requested):
Good job, just small enhancements

inc/Engine/Media/AboveTheFold/Frontend/Controller.php
Khadreal: The `$url` above is overwritten, preg_quote is use here for the path

wordpressfan (approved):
Good job

Mai-Saad:
@Khadreal thanks for the PR. Working for the main shared template, and not adding fetchpriority in some edge cases as per exploratory testing below (not working with absolute URL too and same on trunk):

1. If the relative URL has fetchpriority='high' in the URL  ` `
2. If src was capital or has 1st letter capital or has space after it (If instead of `src="imageURL"` , `SRC` or `Src` or `src = "imageURL"`)
3. If a commented exact markup was 1st on page, fetchpriority will be added to the commented markup instead of active one 
```
<!-- <img src="/wp-content/rocket-test-data/images/lcp/testavif.avif"> -->

		<img src="/wp-content/rocket-test-data/images/lcp/testavif.avif">
```
4. LCP Image with Fetchpriority in markup, will have fetchpriority='high' and that will cause markup error. Example `` => seems regression and not on trunk @Khadreal can you please check?

@piotrbak Do you think we need GH for any of those cases 1-3?

piotrbak:
@Mai-Saad I think those are edge cases, we should be good here

Mai-Saad:
Related test plan https://wpmediaqa.testrail.io/index.php?/runs/view/854&group_by=cases:section_id&group_order=asc

Khadreal:
@Mai-Saad handle the regression and 1 and 2 edge cases. We keep looking on how to resolve the 3rd one. You can do another test when you have some time.

Khadreal:
@Mai-Saad I have added a fix for scenario 3 when there's a commented exact markup.

Mai-Saad (approved):
Working fine now
[testrail-report-606.pdf](https://github.com/wp-media/wp-rocket/files/15455222/testrail-report-606.pdf)

PR #6641, opened 2024-05-17
Update actions/checkout and actions/cache to v4

remyperona:
# Description

There was a deprecation notice when running the GH actions, because of the migration from Node 16 to Node 20. The checkout and cache actions to v4 fix the issue.


jeawhanlee (approved):

MathieuLamiot:
Making this noQA

PR #6646, opened 2024-05-20
Update delay JS script to 1.2.6

MathieuLamiot:
Created by Github action


hanna-meda (approved):
Related Test Run Report here:
[testrail-report-608.pdf](https://github.com/user-attachments/files/15516461/testrail-report-608.pdf)
Two failed, one blocked:
- one it's non-regression, old issue;
- second still has opened ticket. was supposed to have a fix beginning with 1.2.6 but didn't work out.
- one blocked / untested because we cannot simulate exactly the right (real) conditions for Lighthouse user-agent. We will hear from the customer if something not right.

PR #6727, opened 2024-06-17
Prototype lazyload content html parsing

### LINKED ISSUE #6698: Identify content-related HTML markups so that they can be precisely identified back when re-building the same page
**Context**
For 3.17, we will need to identify some HTML markups with the beacon script during a visit, and then at the next page build, to apply some attribute to those exact markups.
We don't have a way to identify markups reliably across different page builds currently. This issue aims at providing prototypes to do this, with various approaches.

**Expected behavior**
When building several times the same WordPress page in the same configuration, we must be able to add an "identifier" attribute to the HTML markups we want. 
- It must be a reliable way so that, across different page builds on the same page, the markups must always get the same identifier.
- We must limit as much as possible adding identifiers to markups we don't care about (based on the EPIC card criteria: up to 3rd child-level of body, and specific markup types only). Depending on the approach, we are OK not to be very precise on this requirement.

**Additional information**
The identifier could be a hash, but depending on the approach, we might not have something unique to hash (typically with regex). An ID, number, or whatever is reproducible would be OK.

To tackle this R&D, we should build and add a callback to `rocket_buffer` that processes the HTML and adds the 'identifer' to the targeted markups. The callback must be able to perform this with the following different approaches:

- Using DOMDocument: We know it will fail in some cases and we won't be able to do anything to fix it. But that's a good first step as we already have most of the code available in the EPIC Card. We should add a "fail safe" too, in case DOMDocument does not work.
- Using the WP HTML Parser: we don't know the features of this Parser and if it could do what we need. We don't know either if it will be more resilient than DOMDocument. To investigate and try.
- Using Regex: Go through all the markups matching our list, and add an ID to them (the order of match in the regex?). We will need to check this against large templates to see how CPU usage behaves. To do this, we would need a log measuring and returning the time it takes to run the callback.

remyperona:
Currently 3 different solutions are implemented:

- with DOMDocument
- with Regex (caveat on this one, the attribute is added on every matching element, no limit on the children tree)
- with HTMLDomParser library

There is also a measuring timer that will write in the error log the time it took for each execution.

### DIFF
--- a/inc/Engine/Media/Lazyload/Content/Processor/Dom.php
+++ b/inc/Engine/Media/Lazyload/Content/Processor/Dom.php
@@ -1,0 +1,64 @@
+<?php
+declare(strict_types=1);
+
+namespace WP_Rocket\Engine\Media\Lazyload\Content\Processor;
+
+use DOMDocument;
+
+class Dom {
+	public function add_locations_hash_to_html( $html ) {
+		// Load HTML into DOMDocument.
+		$dom = new DOMDocument();
+
+		@$dom->loadHTML( $html, LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD );
+
+		// Get the body element.
+		$body = $dom->getElementsByTagName( 'body' )->item( 0 );
+
+		if ( ! $body ) {
+			return $html;
+		}
+
+		$this->add_hash_to_element( $body, 2 );
+
+		// Output the modified HTML.
+		return $dom->saveHTML();
+	}
+
+	private function add_hash_to_element( $element, $depth ) {
+		if ( $depth < 0 ) {
+			return;
+		}
+
+		$skip_tags = [
+			'DIV',
+			'MAIN',
+			'FOOTER',
+			'SECTION',
+			'ARTICLE',
+			'HEADER',
+		];
+
+		foreach ( $element->childNodes as $child ) {
+			if (
+				XML_ELEMENT_NODE !== $child->nodeType
+				||
+				! in_array( strtoupper( $child->tagName ), $skip_tags, true )
+			) {
+				continue;
+			}
+
+			// Calculate the hash of the opening tag.
+			$child_html       = $child->ownerDocument->saveHTML( $child );
+			$opening_tag_html = strstr( $child_html, '>', true ) . '>';
+
+			$hash = md5( $opening_tag_html );
+
+			// Add the data-rocket-location-hash attribute.
+			$child->setAttribute( 'data-rocket-location-hash', $hash );
+
+			// Recursively process child elements.
+			$this->add_hash_to_element( $child, $depth - 1 );
+		}
+	}
+}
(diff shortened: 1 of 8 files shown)


wordpressfan:
This PR should be reviewed by devs or by product team?

remyperona:
A quick dev review to validate the code of the prototype

inc/Engine/Media/Lazyload/Content/Processor/Regex.php
```diff
@@ -0,0 +1,51 @@
...
+		}
+
+		$skip_tags = [
+			'div',
+			'main',
+			'footer',
+			'section',
+			'article',
+			'header',
+		];
```
Miraeld, line 30: Wouldn't be interesting to put this array in a private/protected properties ?
Or keep it like this, and apply a filter on it so we can add/remove elements from it ?

inc/Engine/Media/Lazyload/Content/Processor/SimpleHtmlDom.php
```diff
@@ -0,0 +1,56 @@
...
+		}
+
+		$skip_tags = [
+			'DIV',
+			'MAIN',
+			'FOOTER',
+			'SECTION',
+			'ARTICLE',
+			'HEADER',
+		];
```
Miraeld, line 37: Same as my previous comment, wouldn't be more interesting to put it in a class properties ?

MathieuLamiot:
Hello @wp-media/qa-team, we have this prototype that you could try to break!
With this version installed:
 -   When browsing a page loading wp rocket (so not cached for instance):
    - an error log should be produced "Execution time: xxx"
    - markups 'DIV','MAIN','FOOTER','SECTION','ARTICLE','HEADER' should be modified with an attribute `data-rocket-location-hash` added to them.
 - There are 3 implementations of the same behavior, that you can change in the code directly:
      - 		`$this->processor->set_processor( 'simple_html_dom' );`
      -  Accepted values are dom, regex and simple_html_dom.
      
Basic testing on the 3 by @Tabrisrp show that it works. But we expect that:
- On some very large pages (with A LOT of divs for instance), the execution time might blow up for some methods.
- On some malformed/weird HTML, some methods might fail to process the tags correctly.

This is R&D and exploratory ; it would be great if you could try and break this thing! Thanks.

inc/Engine/Media/Lazyload/Content/Processor/Dom.php
```diff
@@ -0,0 +1,68 @@
+<?php
+declare(strict_types=1);
+
+namespace WP_Rocket\Engine\Media\Lazyload\Content\Processor;
```
MathieuLamiot, line 4: @Tabrisrp Actually, this feature is about lazyrender, and not lazyload. It could be good to change the namespace and folder name for this feature before we actually start more work on it.

inc/Engine/Media/Lazyload/Content/Processor/Dom.php
```diff
@@ -0,0 +1,68 @@
...
+
+use DOMDocument;
+
+class Dom {
+	public function add_locations_hash_to_html( $html ) {
+		// Load HTML into DOMDocument.
+		$dom = new DOMDocument();
+
+		@$dom->loadHTML( $html, LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD );
```
MathieuLamiot, line 13: Maybe it would be good to do proper error handling instead of the @ syntax? We could maybe add a log with the logger in case an error occurs, and bail out?

PR #6739, opened 2024-06-25
Closes #6715

### LINKED ISSUE #6715: Avoid requests to wp-rocket.me to negatively impact site performances during downtime
### Is your feature request related to a problem? Please describe.
When WP Rocket (website) is down or slow, which can unfortunately happen, this creates slowness on every website checking for updates and license validity. 
The requests go in timeout (5 seconds). 

This creates multiple side effects : 
- Websites not in the cache are impacted
- Admin is impacted
- This potentially generates self-DDos as websites are constantly trying to validate the license 

We've discussed this issue each time we had downtime without tackling it. 
It's time to tackle before another upcoming issue.

### Describe the solution you'd like

1. **Checks for updates and license** are run at Rocket init (like today)
    - on all admin pages
    - on uncached pages
    - if DB says the check was already performed and was a success for the validity period (24h), it is skipped
2. Output of this check (in the plugin):
    - If successful:
        - No changes in actual behavior (db updated, and next check will occur in the next 24 hours)
    - If it fails:
        - For technical reasons:
            - Case 1: Too long to answer — Timeout
                - new non-aggressive retry strategy:
                    - retry every 3h for 72h
                        - then stop checking
                        - and display a message:
                            - ```License validation has failed for 3 days in a row. Please check the documentation for more information or contact support.```
            - Case 2: Response is 500
                - same strategy as timeout
3. Timeout is set to 5s, and a filter to change it is available
4. Clicking on "Refresh info" from My Account in Dashboard page performs the check

### Notes
This could be further improved and considered a minimum viable fix. In addition to changes made server-side, it is good enough to prevent this issue from happening in case of a new major outage.

Miraeld:
# Description

Fixes #6715

### User documentation

### Technical documentation

## Type of change

## New dependencies

## Risks

# Checklists

## Feature validation

## Documentation

## Code style

## Observability

## Risks


inc/Engine/License/API/UserClient.php
```diff
@@ -14,6 +15,28 @@ class UserClient {
...
 
+	/**
+	 * Get the transient key for plugin updates.
+	 *
+	 * This method returns the transient key used for caching plugin updates.
+	 *
+	 * @return string The transient key for plugin updates.
+	 */
+	protected function get_transient_key() {
+		return 'user_information';
```
MathieuLamiot: Maybe name it with rocket_ or wpr_ or something like that to ease identification?

inc/Engine/Common/JobManager/APIHandler/AbstractSafeAPIClient.php
```diff
@@ -0,0 +1,149 @@
...
+		if ( is_wp_error( $response ) ) {
+			$this->set_timeout_transients();
+			return $response;
+		}
+
+		$body = wp_remote_retrieve_body( $response );
+		if ( empty( $body ) || ( is_array( $response ) && ! empty( $response['response']['code'] ) && 200 !== $response['response']['code'] ) ) {
+			$this->set_timeout_transients();
+			return new WP_Error( 500, __( 'Not valid response.', 'rocket' ) );
+		}
```
MathieuLamiot: In case the return is not a 200, I am afraid this will be occulting completely the response for potential debug: there will be no way to know what was actually returned as we erase the response code and just send back a WP_Error 500?
Could we either:
- add a log, or
- Use the response code in the WP_Error code, or
- Past the response code in the message of the WP_Error?

inc/Engine/Common/JobManager/APIHandler/AbstractSafeAPIClient.php
```diff
@@ -0,0 +1,149 @@
...
+	 * @param string $method The HTTP method (GET or POST).
+	 * @param array  $params The request parameters.
+	 * @param bool   $safe Send safe request WP functions or not, default to not.
+	 * @return mixed The response from the API, or WP_Error if a timeout is active.
+	 */
+	private function send_request( $method, $params = [], $safe = false ) {
+		$api_url = $this->get_api_url();
+
+		if ( get_transient( $this->get_transient_key() . '_timeout_active' ) ) {
+			return new WP_Error( 429, __( 'Too many requests.', 'rocket' ) );
```
MathieuLamiot, line 68: Maybe we could use a more specific string than the generic "Too many requests" one, at least to know the error is coming specifically from our transient system?
Or maybe put a log for debug mode?

inc/Engine/Common/JobManager/APIHandler/AbstractSafeAPIClient.php
```diff
@@ -0,0 +1,149 @@
...
+		unset( $params['method'] );
+
+		switch ( $method ) {
+			case 'GET':
+				return wp_safe_remote_get( $api_url, $params );
+			case 'POST':
+				return wp_safe_remote_post( $api_url, $params );
+		}
+
+		return new WP_Error( 400, __( 'Not valid request type.', 'rocket' ) );
```
MathieuLamiot, line 151: Can the $method value be included in the message? It can ease debugging. Or add a log in debug mode?

inc/Engine/WPRocketUninstall.php
```diff
@@ -80,6 +80,8 @@ class WPRocketUninstall {
 		'wp_rocket_pricing_timeout',
 		'wp_rocket_pricing_timeout_active',
 		'rocket_get_refreshed_fragments_cache',
+		'user_information_timeout_active',
```
MathieuLamiot: Transients to be added to the Test_Uninstall test

MathieuLamiot (commented):
Early review to provide some guidance and speed up the process.
There is still 1 failing integration test.

inc/Engine/License/API/UserClient.php
Miraeld: Done

inc/Engine/WPRocketUninstall.php
Miraeld: I've added it.

inc/Engine/Common/JobManager/APIHandler/AbstractSafeAPIClient.php
MathieuLamiot: To do later on

inc/Engine/Common/JobManager/APIHandler/AbstractSafeAPIClient.php
MathieuLamiot: To do later on

inc/Engine/Common/JobManager/APIHandler/AbstractSafeAPIClient.php
MathieuLamiot: To do later on

Khadreal (approved):

Mai-Saad (approved):
Working as expected

thinkjarvisdesignandmarketing:
@Mai-Saad
Following the DDOS attack and impact it had on all licensed websites:
For your more professional users please can we get an explanation as to what has changed about license calls?
For my own sanity as an agency owner - I need to know what has changed.

Is WP-Rocket still running license checks on the front end?
Is it just running checks when the admin screen opens?
Any other calls to WP-Rocket servers that we should be aware of that cause the connection to fail or repeatedly time out?

MathieuLamiot:
@thinkjarvisdesignandmarketing, Hello

Here are some details about the fix:
First of all, there are no repeated license validations: the calls to wp-rocket.me that have been slowing down user websites during the downtime are designed to retrieve the latest information about your wp-rocket.me account, from our server to your plugin (license type, license expiration date, etc.) so that, for instance, the information displayed on the plugin dashboard in wp-admin are accurate, among other things.
Once the call succeeds, the data is kept as a transient for a few hours. A new call only happens once the transient is expired.

The root cause of the issue was that, in case the transient expires, the call was retried on each load of the plugin until it succeeded. Since the website was not responding, this happened up until we were back online, and each plugin load had to wait for the timeout to be reached to continue. Note that this timeout is not enforced by our plugin but the default value (configurable in WordPress) is used.

To mitigate this type of issue, we already had exponential retry mechanisms implemented (if a call fails, don't retry at every plugin load, but wait a bit before retrying, and so on). This is the standard state-of-the-art approach for this. Unfortunately, such a strategy was not implemented for this specific call. In the upcoming version, we apply this retry strategy to the problematic call as well, so that, in case wp-rocket.me does not respond, your websites won't keep trying to reach it systematically.

thinkjarvisdesignandmarketing:
@MathieuLamiot 
Please can you confirm with a simple statement below:
**License checks no longer run on the front end of your customers websites**
This was the main concern - Essentially if the WP-Rocket server was not available uncached pages were not serving.
If I am correct in this assumption - This means our TTFB scores are actually tied to how fast the license check or ping to wp-rockets servers take.

**Can you please confirm that the license checks are no longer tied directly to cache generation, RUCSS and preload?**

**Otherwise the proposed solution does not solve the problem**

**Why is the transient lifespan not tied to the expiry date of the clients subscription to WP-Rocket?**
If you just use the expiry date on your server - It only needs to be checked once - At the start/end of their fixed term license. Not periodically.

PR #6742, opened 2024-06-26
:closes: Fix issue with warm up link for homepage

### LINKED ISSUE #6697: Warmup fetch links - Mobile version missing from DB (ATF table) for Homepage URL on fresh installs
 ✅ Made sure you’re on the latest version - 3.16.1
 ✅ Used the search feature to ensure that the bug hasn’t been reported before

**Describe the bug**
The Mobile version of Homepage URL is missing from DB - ATF table when activating plugin after a fresh install.

**To Reproduce**
(Make sure the filter to disable the Separate Mobile Cache is not active)
1. Install WPR 3.16.1 & activate plugin
2. Inspect the DB - above_the_fold table
3. Notice that for Homepage URL there is only one entry, for Desktop

**Expected behavior**
Both Mobile and Desktop versions should be added in DB for home, on fresh install.

Other notes:
- Link requests sent to SaaS are fine. Both mobile and desktop for Home URL are present in logs.
- Trying out suggestion discussed in [thread](https://wp-media.slack.com/archives/CUT7FLHF1/p1717587455861739?thread_ts=1717587272.951999&cid=CUT7FLHF1) we get only Mobile and missing Desktop URL for home.
- All works fine for other actions that trigger the warmup process, i.e Clear Critical Images. 

**Screenshots**
Screenshot 2024-06-05 at 15 18 33

**Additional context**

Khadreal:
# Description

Fixes #6697

### User documentation

### Technical documentation

## Type of change

## New dependencies

## Risks

# Checklists

## Feature validation

## Code style

## Observability

## Risks


inc/Engine/Media/AboveTheFold/WarmUp/Controller.php
```diff
@@ -203,13 +203,14 @@ function ( $link ) use ( $home_url, $reject_uri_pattern ) {
 	 * Send link to SaaS to do the warmup.
 	 *
 	 * @param string $url Url to send.
+	 * @param bool   $is_warm_up set to false by default, should be true if request is warm up.
 	 *
 	 * @return void
 	 */
-	public function send_to_saas( string $url ) {
+	public function send_to_saas( string $url, bool $is_warm_up = false ) {
```
MathieuLamiot: is_warm_up might not be a good name: send_to_saas is only use for the warm-up mechanism. I think the point of this bool is more to know if we are in the sync warm-up (for homepage) or the async one (for the other pages)?

inc/Engine/Media/AboveTheFold/WarmUp/Controller.php
```diff
@@ -237,9 +238,15 @@ public function add_wpr_imagedimensions_query_arg( string $url ): string {
 	/**
 	 * Check if the current request is for mobile.
```
MathieuLamiot: Not related to the PR but we could update this docstring: it looks like a bad copy-paste. What this is checking is not related to a current request, it's just checking if the current config does separate the mobile cache.

inc/Engine/Media/AboveTheFold/WarmUp/Controller.php
```diff
@@ -237,9 +238,15 @@ public function add_wpr_imagedimensions_query_arg( string $url ): string {
...
 	 *
+	 * @param bool $is_warm_up if check is for warm up or not.
+	 *
 	 * @return bool
 	 */
-	private function is_mobile(): bool {
+	private function is_mobile( bool $is_warm_up = false ): bool {
+		if ( $is_warm_up ) {
+			return true;
+		}
```
MathieuLamiot: While this solves the issue, the risk is that, if the user deactivate the separate mobile cache option, then we would not listen to this change and still always send desktop and mobile to the SaaS for the homepage.

MathieuLamiot (changes requested):
Thank you for the PR! It helped me understanding the issue and the approach.
While this PR should solve the issue, may I suggest a different approach? Reading this PR gave me an idea, and I think it could be good to discuss it to have a "cleaner" fix I think.

Actually, the root cause of the issue is that, on fresh install, the `warm_up_home` is called before the options `cache_mobile` and `do_caching_mobile_files` are set! This is due to the fact that `warm_up_home` is triggered with the `register_activation_hook` (`activate_plugin` -> `rocket_after_activation`) while options are set on `admin_init` which occurs after.
To confirm this, I added the following logs:

```
function rocket_upgrader() {
	// Grab some infos.
	
	$actual_version = (string) get_rocket_option( 'version', '' );
	// You can hook the upgrader to trigger any action when WP Rocket is upgraded.
	// first install.
	if ( ! $actual_version ) {
		error_log(" MLT TEST - FIRST INSTALL");
		do_action( 'wp_rocket_first_install' );
	}
```
To spot when the options are set (in `wp_rocket_first_install`)

And 

```
	private function is_mobile( bool $is_warm_up = false ): bool {
		error_log(" MLT TEST - " . (string) get_rocket_option( 'version', '' ));
		error_log(" MLT TEST - " . (string) $this->options->get( 'cache_mobile', 'not here' ));
		error_log(" MLT TEST - " . (string) $this->options->get( 'do_caching_mobile_files', 'not here' ));
		if ( $is_warm_up ) {
			return true;
		}

		return $this->options->get( 'cache_mobile', 0 ) && $this->options->get( 'do_caching_mobile_files', 0 );
	}
```

Here is the output upon fresh install:

```
[26-Jun-2024 17:59:37 UTC]  MLT TEST - 
[26-Jun-2024 17:59:37 UTC]  MLT TEST - not here
[26-Jun-2024 17:59:37 UTC]  MLT TEST - not here
[26-Jun-2024 17:59:38 UTC]  MLT TEST - FIRST INSTALL
```
This confirms the order of execution of the actions, hence the issue.

To solve the issue, ie. send desktop + mobile homepage on fresh install, and keep the existin behavior the rest of the time, I would suggest the following rewrite of `inc/Engine/Media/AboveTheFold/WarmUp/Controller.php -> is_mobile`

```
	/**
	 * Check if separate cache for mobile is configured.
	 *
	 * @return bool
	 */
	private function is_mobile(): bool {
		$plugin_version = (string) get_rocket_option( 'version', '' );
		if ( ! $plugin_version ) { # We are warming up a fresh install. Options are not set yet.
			return true;
		}

		return $this->options->get( 'cache_mobile', 0 ) && $this->options->get( 'do_caching_mobile_files', 0 );
	}
```

Mai-Saad:
Related test plan: https://wpmediaqa.testrail.io/index.php?/runs/view/864

MathieuLamiot:
@wp-media/qa-team The PR won't completely fix the issue, just improve it a bit (should warm up homepage sometimes on fresh installs, but not consistently). But we still expect that homepage might not be warmed_up systematically on fresh installs: the requests to the SaaS will be made and visible on Metabase, but the beacon might not be injected. (it's flaky, so not always reproducible, there is a race condition).
The real fix will be complex so we opened a dedicated issue : https://github.com/wp-media/wp-rocket/issues/6746

I would say maybe we can make this a NoQA since there is no consistent change in behavior? NRT would be enough to ensure there are no warning/Errors (the code change is quite small, that should be quick). WDYT?

MathieuLamiot (approved):

MathieuLamiot:
I completed the test plan: https://wpmediaqa.testrail.io/index.php?/runs/view/864&group_by=cases:section_id&group_order=asc

Two issues:
For one of the fresh install, the homepage was missing. Due to #6746 I think. It happened once, and I could not reproduce systematically.
I was not able to perform the test on nest.rocketlabsqa as I can't find credentials.

Mai-Saad:
The PR generally enhances the warmup with the following notes which will be handled on separate issues
- Sometimes with fresh install, no mobile version is added at all  and fresh install while separate mobile cache on will send home mobile to saas https://github.com/wp-media/wp-rocket/issues/6752
- Sometimes mobile for home won't be added https://github.com/wp-media/wp-rocket/issues/6746

PR #6771, opened 2024-07-11
Use rocket scripts package for LCP Beacon

wordpressfan:
Part of https://github.com/wp-media/wpr-inspector/issues/24

Here we are changing the way we generate the beacon scripts, so we created a new package here: https://github.com/wp-media/rocket-scripts to have the source code of beacon script(s) and build their files and here in WPR for example we will just copy the files from their dist into our js directory.

This will a lot in adding JS tests there, also will make WPR repo cleaner so any change needed in beacon scripts (and we may add more scripts in the future) so we will need to go to this repo and do whatever needed.

## Todo
- [x] Add unit tests for rocket-scripts repo
- [ ] Divide the beacon class into smaller modules to make it easier to create a new file that uses the current code.


Khadreal (approved):
LGTM

Mai-Saad (approved):
working as expected (note: e2e lcp when run altogether, mobile fails then running mobile scenario only passed)
[testrail-report-618.pdf](https://github.com/user-attachments/files/16284348/testrail-report-618.pdf)
[PR6771 npm run test lcp.txt](https://github.com/user-attachments/files/16284352/PR6771.npm.run.test.lcp.txt)
[PR6771 mobile only npm.txt](https://github.com/user-attachments/files/16284353/PR6771.mobile.only.npm.txt)

PR #6790, opened 2024-07-17
Closes #6273  exclude JS module type from minification

### LINKED ISSUE #6273: Exclude js scripts with type="module" from JS minification
yes - Made sure you’re on the latest version
yes - Used the search feature to ensure that the bug hasn’t been reported before

**Describe the bug**
When the option “Minify JavaScript files” is enabled, and scripts with `type="module"` are moved to the min folder, is likely to cause 404 errors because a a JS file with this type is likely to have import statemets, and changing the path where the expected module is located will break these links

Per customer feedback: 
> 2. JavaScript with type=“module” will always have “import” statements in it that references to a files, which location or existence WP Rocket cannot detect. So if WP Rocket changes and moves the file, the paths to these files are interrupted by WP Rocket.

Example code. the files `./localization.d6c03d4a.js`, `./selectors.e37ac126.js` and `./vue.16a54326.js` with cause 404:
```
import{_ as i,a as n,b as s,c as e,d as a,e as r,f as l,g as C,h as b,i as d,j as c,k as w,l as k}from"./localization.d6c03d4a.js";import{_ as t,a as g}from"./selectors.e37ac126.js";import"./vue.16a54326.js";class _{IABTCF(){return{currentTCString:"",hasWhichConsents:function(){return{}},isTCStringUpToDate:function(){return!1},signalDoNowShowUI:function(){},saveConsentAll:function(){}}}}const p=new _;i.init(window.borlabsCookieConfig.contentBlockers);n.init(window.borlabsCookieConfig.providers);s.init(window.borlabsCookieConfig.serviceGroups);e.init(window.borlabsCookieConfig.services);t.init(window.borlabsCookieConfig.settings);a.init(window.borlabsCookieConfig.globalStrings);window.borlabsCookiePrioritized===void 0&&(window.borlabsCookiePrioritized={},window.borlabsCookiePrioritized.optInCodesUnblocked={});r.optInCodesUnblocked=window.borlabsCookiePrioritized.optInCodesUnblocked;const f={Cookie:C,ContentBlocker:i,Consents:b,ConsentHistory:d,Providers:n,ScriptBlocker:c,ServiceGroups:s,Services:e,CookieLibrary:w,Settings:t,Log:g,Unblock:r,Tools:k,Adapter:p};window.BorlabsCookie=f;window.__toBorlabsPluginAssetUrl=o=>window.borlabsCookieConfig.settings.pluginUrl+"/assets/"+o;l.init();
```

**To Reproduce**
1. Enqueue a JS script with the above code.
2. create some import scripts at the same location for example: `./localization.d6c03d4a.js`, `./selectors.e37ac126.js` and `./vue.16a54326.js`
3. test the page without JS minification, it will be fine. 
4. Enable JS Minification
5. See the 404 error on the imported files
6. Exclude the origin file form JS minification
7. see the errors fixed

**Expected behavior**
we should exclude `type="module"` from minification

**Screenshots**
cleanshot-2023-11-08-at-15-50-49-2x

**Additional context**
Slack: https://wp-media.slack.com/archives/C43T1AYMQ/p1699483184373919
ticket: https://secure.helpscout.net/conversation/2414653718/453464?folderId=2683093

- The exclusion should be part of the backend
- It's blocked by https://github.com/wp-media/rucss-backend/issues/83

Khadreal:
# Description
Make js minification dynamic instead of hardcoded appraoch

Fixes #6273 

## Documentation
Removed the hardcoded approach for JS  minification

### User documentation

### Technical documentation

## Type of change

# Checklists

## Feature validation

## Documentation

## Code style


remyperona (changes requested):
This is incomplete. The update needs to pull the data from the dynamic-list and use it to check if the file should be minified or not.

So the behaviour needs to remain the same, but using remote data instead of hardcoded.

Mai-Saad:
@Khadreal Thanks for the PR. During the exploratory test can see that currently only this markup `type='module'` is handled, if type/ module starts with capital i.e Type/Module or all caps TYPE/MODULE or there is a space i.e `type = 'module'`, exclusion won't work, should we handle those cases @piotrbak ? 

Note: The same is happening on the trunk for data-minify or data-no-minify  has space or starts with caps => agreed that no changes about that for now.

Mai-Saad:
For testing once PR is ready:
1- pull PR
2- update dynamic lists
3- enable minify js
4- visit a page with scripts that shouldn't be minified i.e rewritten to .min folder  => validate scripts aren't minified https://new.rocketlabsqa.ovh/minifyjs_type_module

inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
```diff
@@ -94,12 +104,13 @@ protected function is_minify_excluded_file( array $tag ) {
...
 		}
+		$exclude_js_template = $this->dynamic_lists->get_exclude_js_templates();
 
 		// File should not be minified.
 		if (
-			false !== strpos( $tag[0], 'data-minify=' )
+			in_array( 'data-minify=', $exclude_js_template, true )
 			||
-			false !== strpos( $tag[0], 'data-no-minify=' )
+			in_array( 'data-no-minify=', $exclude_js_template, true )
```
wordpressfan: I believe we should change this part completely to be regex to compare the current tag `$tag[0]` with the values from the dynamic list, we don't need to check specific values here like `data-no-minify=` we need to make it more generic.

wordpressfan (changes requested):

inc/Engine/Optimization/Minify/JS/AbstractJSOptimization.php
Khadreal: spot on, the list might grow and we don't have to keep modifying the code for that

remyperona (approved):

hanna-meda (approved):
Looks good. Test Report below. Thank you.
[testrail-report-620.pdf](https://github.com/user-attachments/files/16312468/testrail-report-620.pdf)

PR #6917, opened 2024-08-27
Filter depth and processed elements' tags

### LINKED ISSUE #6904: 3.17 - Make depth and type of LRC candidates filterable
**Context**
Currently, PHP adds hashes to LRC candidates up to depth 2 inside the body.
Note that the depth and type are currently also checked by the beacon, but it should not and will be removed as part of https://github.com/wp-media/rocket-scripts/issues/25
Until then, it is possible to implement this issue and test it by checking what elements get the hash added, but they might not be picked up correctly by the beacon, hence not be in DB at the end of the visit.

**Expected behavior**
We must provide a filter to change how deep we look for eligible LRC candidates to add the hashes to.
We must provide a filter to change the list of eligible element types for LRC candidates.

**Acceptance Criteria**
- The filter is safeguarded to only accept strictly positive integers (for depth) and array of string for element types.
- With the default value, clear the cache and remove the Performance Hints data. Visit a page and check some elements that don't have the hash and that have the right type (typically div), but their parents have (so depth = 3) the hash. Then increase the depth with the filter and repeat. Those elements must have the hash.
- With the default value, clear the cache and remove the Performance Hints data. Visit a page and check some elements that don't have the hash because they are not the right type (but at the right depth). Then change the list of elibigle elements to include those elements and repeat. Those elements must have the hash

**Proposed grooming**
I think the solution can be:

The constant here:

`$this->add_hash_to_element( $body, 2 );`

in /Users/mathieu/Documents/Github/wp-rocket/inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/SimpleHtmlDom.php and /Users/mathieu/Documents/Github/wp-rocket/inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Dom.php

Same thing for the $skip_tags in:

/Users/mathieu/Documents/Github/wp-rocket/inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/SimpleHtmlDom.php and /Users/mathieu/Documents/Github/wp-rocket/inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Dom.php
/Users/mathieu/Documents/Github/wp-rocket/inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Regex.php

Effort: XS

wordpressfan:
# Description

Fixes #6904

This PR adds two filters:

1. `rocket_lazy_render_content_depth`: returns integer value with default of `2` to adjust the depth.
2. `rocket_lazy_render_content_processed_tags`: returns array value with default of:

```
[
	'DIV',
	'MAIN',
	'FOOTER',
	'SECTION',
	'ARTICLE',
	'HEADER',
]
```
To adjust the elements to be processed by this feature.

## Type of change

## Detailed scenario

As groomed and mentioned above.

### Documentation

We only added the filters as mentioned above

### New dependencies

No

### Risks

We shouldn't have any risk here, we just added the filter so the user can change those values, the only think that I can think of, is that the user can break the functionality by returning not valid elements from the tags filter or by decreasing the depth value.

# Mandatory Checklist

## Code validation

## Code style

# Additional Checks


Miraeld (approved):

MathieuLamiot:
Closing this PR as it is merged in #6929 which will be merged in LRC branch already

PR #6929, opened 2024-08-29
Depend on php to identify elements and depth

wordpressfan:
# Description

This PR is a reflect of the changes of that PR: https://github.com/wp-media/rocket-scripts/pull/26

It'll not impact users, that's an enhancement.

## Type of change

## Detailed scenario

Everything is mentioned in the issue itself and grooming discussion.

### Documentation

We will depend on php to identify elements and depth.

### New dependencies

No

### Risks

No

# Mandatory Checklist

## Code validation

## Code style

# Additional Checks


jeawhanlee:
Would it not be a good idea to pass `[data-rocket-location-hash]` to the element array [here](https://github.com/wp-media/wp-rocket/blob/d7a5719c9211d080f22af8594480b3f260219196/inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php#L141)

But just in case you don't think the same, we should remove [this](https://github.com/wp-media/wp-rocket/blob/d7a5719c9211d080f22af8594480b3f260219196/inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php#L141-L159) whole block then since `[data-rocket-location-hash]` is now hard coded in the script.

wordpressfan:
That's a good idea @jeawhanlee

jeawhanlee (approved):

Mai-Saad (approved):
Working as expected. (includes validation for #6904)

- we can have enhancement later https://wp-media.slack.com/archives/CUKB44GNN/p1725008472171879?thread_ts=1724992642.140039&cid=CUKB44GNN
- if nothing is needed for this now https://wp-media.slack.com/archives/CUKB44GNN/p1725009496716359 then we can merge

DahmaniAdame:
@MathieuLamiot we will need to revert back to calculating the distance relative to the viewport. 

```
    _getElementDistance(element) {
      const rect = element.getBoundingClientRect();
      const scrollTop = window.pageYOffset || document.documentElement.scrollTop;
      return Math.max(0, rect.top + scrollTop - (window.innerHeight || document.documentElement.clientHeight));
    }
```

What @Mai-Saad reported here will be complicated to handle as we can't restrict customers to a specific value and it will require a lot of customer education to not cost support time and efforts - https://wp-media.slack.com/archives/CUKB44GNN/p1725009496716359

Sorry for the oversight.

MathieuLamiot:
@DahmaniAdame to avoid further confusion, I'd prefer we have a dedicated GH issue: not only JS must be changed but also the default threshold value (PHP side).
So we would need the details about this in a dedicated issue (what value exactly? Same for mobile & desktop?, etc.)

Note that this means results will be dependent on the screensize of the person visiting the page. So someone with a small screen might pickup some LRC elements that will be lazy rendered while being in the viewport of another user with taller screen. I am not sure this is the best solution as results won't be always reproducible.

DahmaniAdame:
Got it. I will create a separate GH issue. 

> Note that this means results will be dependent on the screensize of the person visiting the page. So someone with a small screen might pickup some LRC elements that will be lazy rendered while being in the viewport of another user with taller screen.

That was the thinking behind it. But it will cause more trouble and confusion for customers. 

The relative distance to the viewport is still usable though and should work fine with the constraint that we have now. 

Having a couple of elements toggling `content-visibility` on page load on different screen size is acceptable. We have the same behavior between browsers where the threshold is different as well, and it's acceptable there as well. The 1800 threshold initially had that difference in mind.

PR #6937, opened 2024-09-02
Fixes #6922 3.17: Hashes are not removed from the source when there is no LRC data

### LINKED ISSUE #6922: 3.17: Hashes arenot removed from the source when there is no LRC data
- Made sure you’re on the latest version => feature/lrc

**Describe the bug**
When accessing a page that is totally above the fold and having lrc as [] in DB , hashes aren't removed from source

**To Reproduce**
1. Clear performance hints
2. Visit a page that is all is ATF => LRC is [] in DB
3. Clear cache
4. Visit the page again => hashes are there

**Expected behavior**
Hashes should be removed when visiting uncached page that have [] LRC in DB

**Screenshots**

**Additional context**
- slack https://wp-media.slack.com/archives/CUT7FLHF1/p1724850889231899
- If screen size is unacceptable, the script will bailout and hashes will always be there in the source => expected

Khadreal:
# Description
Hashes should be removed when visiting uncached page that have [] LRC in DB

Fixes #6922

## Type of change

## Detailed scenario
Please check #6922 for detailed scenario.

## Technical description
Update`has_lrc` method in LazyRenderContent database row to check for `'[]'` instead of 'not found' which doesn't exist in db records and updated the `optimize` method to return html without hashes when the logic checks passed.
  
### Documentation

Hashes should be removed when visiting uncached page that have [] LRC in DB

# Mandatory Checklist

## Code validation

## Code style

### DIFF
--- a/inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php
+++ b/inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php
@@ -42,15 +42,14 @@ class Controller implements ControllerInterface {
 	 * @return string
 	 */
 	public function optimize( string $html, $row ): string {
-		$html_without_hashes = $this->remove_hashes( $html );
 		if ( ! $row->has_lrc() ) {
-			return $html_without_hashes;
+			return $this->remove_hashes( $html );
 		}
 
 		$hashes = json_decode( $row->below_the_fold );
 
-		if ( null === $hashes || ! is_array( $hashes ) ) {
-			return $html_without_hashes;
+		if ( ! is_array( $hashes ) ) {
+			return $this->remove_hashes( $html );
 		}
 
 		$result = preg_replace( '/data-rocket-location-hash="(?:' . implode( '|', $hashes ) . ')"/i', 'data-wpr-lazyrender="1"', $html, -1, $count );
@@ -60,7 +59,7 @@ class Controller implements ControllerInterface {
 			||
 			0 === $count
 		) {
-			return $html_without_hashes;
+			return $this->remove_hashes( $html );
 		}
 
 		$html = $result;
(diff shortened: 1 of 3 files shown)


inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php
```diff
@@ -42,14 +42,15 @@ public function __construct( Processor $processor, ContextInterface $context ) {
 	 * @return string
 	 */
 	public function optimize( string $html, $row ): string {
+		$html_without_hashes = $this->remove_hashes( $html );
```
MathieuLamiot: @Khadreal Is it possible to run this line only when needed? Here I think we risk running it even in cases where `$html_without_hashes` won't be used and since it involves a regex (that can be huge in some cases), it would be safer for performances to avoid unnecessary computations.

Khadreal: Yes, we could but every return in this method need to remove the hashed before returning.
We could move it inside the logic, which means `remove_hashes` will be repeated multiple times.

so we we'll have something like this 
```
if ( ! $row->has_lrc() ) {
			return $this->remove_hashes( $html );
		}
		$hashes = json_decode( $row->below_the_fold );

		if ( null === $hashes || ! is_array( $hashes ) ) {
			return $this->remove_hashes( $html );
		}

		$result = preg_replace( '/data-rocket-location-hash="(?:' . implode( '|', $hashes ) . ')"/i', 'data-wpr-lazyrender="1"', $html, -1, $count );

		if (
			null === $result
			||
			0 === $count
		) {
			return $this->remove_hashes( $html );
		}
		........
```

MathieuLamiot: I think that's OK to repeat the method 👍

remyperona (approved):

Mai-Saad:
related test plan https://wpmediaqa.testrail.io/index.php?/runs/view/894

MathieuLamiot:
Test plan all 🆗 for me. @wp-media/qa-team Feel free to merge or explore more if you prefer 👍

PR #6947, opened 2024-09-04
Update LRC filters prefix

remyperona:
# Description

Standardize the prefix for LRC filters from `rocket_lazy_render_content_`  to `rocket_lrc_`.

## Type of change

- [x] Sub-task of #3843 

## Detailed scenario

The new filters directly replace the old one with no deprecation since it was not publicly released.

### Documentation

For consistancy, renaming the LRC filters so they all use the same prefix.

# Mandatory Checklist

## Code validation

- [x] I implemented built-in tests to cover the new/changed code.

## Code style
- [x] I wrote a self-explanatory code about what it does.
- [x] I protected entry points against unexpected inputs.
- [x] I did not introduce unnecessary complexity.

PR #6949, opened 2024-09-05
Fixes #6948 -- rocket_lrc_processed_tags should be case insensitive

### LINKED ISSUE #6948: rocket_lrc_processed_tags should be case insensitive
**Context**

As reported by the support team while testing, elements passed to the filter `rocket_lrc_processed_tags` (renaming ongoing, could still be named `rocket_lazy_render_content_processed_tags`) must be added as uppercase currently.
On OCI feature, we are agnostic of upper/lower case so we should align.

The root cause and possible fix are already discussed here: https://wp-media.slack.com/archives/C07LKLTKBG8/p1725516717052949

**Expected behavior**

Elements added to the processed_tags must be handled correctly whether they are added to the filtered array as lower or upper case.

**Acceptance criteria**

On a page with a h1 element that should get the LRC hash (depth being properly configured):
- When using the following snippet, the hash is added to the h1 element:
```
add_filter( 'rocket_lazy_render_content_processed_tags', function( $tags ){
	$tags[]= "H2";
	$tags[]= "H1";
    return $tags;
} )
```
- When using the following snippet, the hash is added to the h1 element:
```
add_filter( 'rocket_lazy_render_content_processed_tags', function( $tags ){
	$tags[]= "h2";
	$tags[]= "h1";
    return $tags;
} )
```

Khadreal:
# Description
Make filter value case insensitive

Fixes #6948 

## Type of change

## Detailed scenario
Check #6948 for scenario

## Technical description
Convert filter values to upper case in `get_processed_tags` method.
### Documentation
Make sure filter value are case insensitive.

# Mandatory Checklist

## Code validation

## Code style

### DIFF
--- a/inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/HelperTrait.php
+++ b/inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/HelperTrait.php
@@ -31,7 +31,7 @@ trait HelperTrait {
 		 *
 		 * @param array|string[] $tags Tags to be processed.
 		 */
-		return wpm_apply_filters_typed(
+		$tags = wpm_apply_filters_typed(
 			'array',
 			'rocket_lazy_render_content_processed_tags',
 			[
@@ -43,5 +43,10 @@ trait HelperTrait {
 				'HEADER',
 			]
 		);
+
+		/**
+		 * Convert tags to upper case here before
+		 */
+		return array_map( 'strtoupper', $tags );
 	}
 }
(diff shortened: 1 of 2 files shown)


MathieuLamiot:
@Khadreal what do you think of an integration test that registers a callback to the filter, adding lowercase items, and checking the output of the method to ensure they are uppercase?
I am confident your code works, but it could help capturing the behavior in case we refactor something later on?

Khadreal:
If this [PR](https://github.com/wp-media/wp-rocket/pull/6947) is merged first then this would need to be updated.

MathieuLamiot (approved):

MathieuLamiot:
I replicated the problematic scenario and it works as expected.

PR #6954, opened 2024-09-05
Fix/6953 lrc replace dom output with regex injection

### LINKED ISSUE #6953: 3.17 - Replace HTML Write operation from DOMDocument with Regex for LRC hash injection
**Context**
https://wp-media.slack.com/archives/CUT7FLHF1/p1725542573366859?thread_ts=1725532083.335009&cid=CUT7FLHF1

**Expected behavior**
For LRC, use DOMDocument to identify the opening tags in which a hash should be added. Then, use regexes to inject the hash attribute on the original HTML, so that DOMDocument output is only used on HTML exactly matching the original file.

**Acceptance Criteria**
- No broken layout when applying hashes on the templates:
    - https://wp-media.slack.com/archives/CUT7FLHF1/p1725532083335009
    - https://wp-media.slack.com/archives/C07LKLTKBG8/p1725515496731509
- Acceptable performances on large templates (to compare with previous DOM method)
- No functional regressions on LRC

MathieuLamiot:
# Description

Fixes #6953
This PR limits performance issues on all LRC front-end processors by limiting the number of injected tags to 200 (filterable with `rocket_lrc_max_hashes`).
It also prevents DOMDocument from interfering with HTML by replacing write operations of the DOM processor with regex replacements. This is not appleid to simple DOM.

More discussion on this PR here: https://wp-media.slack.com/archives/CUT7FLHF1/p1725558490681749?thread_ts=1725532083.335009&cid=CUT7FLHF1

## Type of change

- [ ] Release

## Detailed scenario

When a user visits a non-cached page, the LRC feature injects hashes as attribute to some selected elements. This process remains the same. However:
- With dom processor, we only use opening tags from DOMDocument. The rest of the html is preserved to avoid breaking the layout.
- With all processors, we stop injecting hashes after a given number to prevent performance issues on very large pages.

### Documentation

About bailing out after a given number of hashes, there is a new filter for the max. value: `rocket_lrc_max_hashes`. The counter previously used to ensure hashes were different is now used also as a counter of added hashes and we bail out if it is too high.
I moved the counter to an object variable to access it in the recursive calls to `add_hashes_to_element`.

About preventing DOM to rewrite the HTML: we no longer rely on saveHTML to output the HTML from DOMDocument.
Instead, for each candidate to a hash identified by DOMDocument, we:
- get the opening tag from DOMDocument
- add the hash to it with a preg_replace, which gives a modified tag
- Look for the original tag in the original HTML and replace it with the modified tag.

### New dependencies

NA

### Risks

- In some cases, with DOM processor, DOMDocument rewrites a bit the opening tag, and therefore we can't find it back in the original HTML. So some elements won't get the hash injected while they should. I think it is acceptable for now.
- In some cases, the DOM processor might add the hash to the wrong element. This has low impact but prevents the identification of a LRC candidate. This happens only if two or more opening tags are exactly the same and the first one of the set is not eligible (too deep, or not with an eligible parent for instance).
- With the regex processor, since we limit the number of injected hashes and we don't use the depth parameter, we risk adding hashes to all elements at the beginning of the page and skipping the bottom of the page, which defies the purpose of the feature.

I currently recommend using this new DOM processor by default to get the best performance and functionalities.

# Mandatory Checklist

## Code validation

    - With DOM processor, I could not see visual regressions anymore on the templates from new.rocketlabsqa.ovh (from both slack threads).
    - With DOM & Regex, I checked the large html page https://new.rocketlabsqa.ovh/largehtmldiv/ and it works as expected with no visible latency (a few seconds max., I did not measure exactly, but definitely faster). Only the 200 first divs have hashes, and then only from~50 to 200 are lazy rendered.

## Code style

# Additional Checks


MathieuLamiot:
Asking for a first review while we further discuss the approach on Slack: https://wp-media.slack.com/archives/CUT7FLHF1/p1725558490681749?thread_ts=1725532083.335009&cid=CUT7FLHF1

inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Dom.php
```diff
@@ -10,6 +10,24 @@ class Dom implements ProcessorInterface {
 
 	use HelperTrait;
 
+	/**
+	 * Number of injects hashes.
+	 *
+	 * @since 3.6
```
Miraeld: It's been added here, so why is it `@since 3.6` ? Shouldn't it be `3.17` ?

inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Dom.php
```diff
@@ -10,6 +10,24 @@ class Dom implements ProcessorInterface {
...
+	 *
+	 * @var int
+	 */
+	private $count;
+
+	/**
+	 * Maximum number of hashes to inject.
+	 *
+	 * @since 3.6
```
Miraeld: It's been added here, so why is it `@since 3.6` ? Shouldn't it be `3.17` ?

inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Dom.php
MathieuLamiot: @Miraeld Good catch thanks, it's a bad copy/paste.
For such feedback, do not hesitate to make the change directly to avoid further delay

Miraeld (approved):

MathieuLamiot:
Mitigations for the 2 first risks on this PRs are tracked here: https://github.com/wp-media/wp-rocket/issues/6963 & https://github.com/wp-media/wp-rocket/issues/6961

PR #6968, opened 2024-09-11
Closes #6460 Maybe optimize lookups of tables in DB to stop using SHOW TABLES LIKE

### LINKED ISSUE #6460: Maybe optimize lookups of tables in DB to stop using SHOW TABLES LIKE
**Context**
Reported by a user through the support: https://wp-media.slack.com/archives/C43T1AYMQ/p1709561232039849

**Expected behavior**
Investigate the possibility to rework the `exists()` (and related) methods in the codebase, that are using `SHOW TABLES LIKE`, to something using `SELECT table_name FROM information_schema.tables WHERE table_name like '%s' LIMIT 1` to reduce the compute time on the DB when there are many tables to parse from.

**Acceptance Criteria**
Non-regression tests on the impacted features.

**Additional information**
The issue comes from BerlinDB, not WP Rocket code itself: https://github.com/berlindb/core/blob/9c4d1fe588b4bff7b9f45091d7890bbbef9c0bc4/src/Database/Table.php#L351
We received 2 suggestions:
- `SELECT table_name FROM information_schema.tables WHERE table_name = '%s' LIMIT 1` 
- SHOW TABLES =
The main risk here are regressions from changing from a `LIKE` statement to an `=` (maybe with prefixes, etc. ?). There is probably a reason why LIKE is used in BerlinDB, so it's much safer to stick with it for now. That being said, I think the SHOW TABLES is a high-level instruction that will first list all tables, and then apply filters if any.
Going with an explicit SQL statement should be much faster already, hence my suggestion:

`SELECT table_name FROM information_schema.tables WHERE table_name like '%s' LIMIT 1`

Khadreal:
# Description
Changed lookup table query.

Fixes #6460 

## Type of change

## Detailed scenario
With a large site(mostly multisites), the query is run on every page loads and this slows down the page.
More information here https://wp-media.slack.com/archives/C43T1AYMQ/p1709561232039849

## Technical description
Changed the query from `show tables` to select with a limit of 1

### Documentation

# Mandatory Checklist

## Code validation

## Code style

### DIFF
--- a/inc/Engine/Common/Database/Queries/AbstractQuery.php
+++ b/inc/Engine/Common/Database/Queries/AbstractQuery.php
@@ -520,9 +520,8 @@ class AbstractQuery extends Query {
 		}
 
 		// Query statement.
-		$query    = 'SHOW TABLES LIKE %s';
-		$like     = $db->esc_like( $db->{$this->table_name} );
-		$prepared = $db->prepare( $query, $like );
+		$query    = 'SELECT table_name FROM information_schema.tables WHERE table_name = %s LIMIT 1';
+		$prepared = $db->prepare( $query, $this->table_name );
 		$result   = $db->get_var( $prepared );
 
 		// Does the table exist?
--- a/inc/Engine/Common/Database/Queries/AbstractQuery.php
+++ b/inc/Engine/Common/Database/Queries/AbstractQuery.php
@@ -521,7 +521,7 @@ class AbstractQuery extends Query {
 
 		// Query statement.
 		$query    = 'SELECT table_name FROM information_schema.tables WHERE table_name = %s LIMIT 1';
-		$prepared = $db->prepare( $query, $this->table_name );
+		$prepared = $db->prepare( $query, $db->{$this->table_name} );
 		$result   = $db->get_var( $prepared );
 
 		// Does the table exist?
--- a/inc/Engine/Common/PerformanceHints/Database/Queries/AbstractQueries.php
+++ b/inc/Engine/Common/PerformanceHints/Database/Queries/AbstractQueries.php
@@ -122,7 +122,7 @@ class AbstractQueries extends Query {
 
 		// Query statement.
 		$query    = 'SELECT table_name FROM information_schema.tables WHERE table_name = %s LIMIT 1';
-		$prepared = $db->prepare( $query, $this->table_name );
+		$prepared = $db->prepare( $query, $db->{$this->table_name} );
 		$result   = $db->get_var( $prepared );
 
 		// Does the table exist?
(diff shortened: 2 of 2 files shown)


remyperona (approved):

MathieuLamiot:
@wp-media/qa-team @Khadreal to discuss how to validate / test this.
A quick feedback here: the main risk I see is that we are not using "LIKE" anymore but a strictly equal table name. We would need to check on different setups (multi-site, a few hosting, etc.) that we always have the exact table name.
@Khadreal for instance, in most queries, we use $db->prefix, which does not seem to be the case in your PR. We'd need to ensure we are not missing a prefix mechanism here.
https://github.com/wp-media/wp-rocket/blob/b898c2a8b6f52d9fd93702a59419d57f208b0b4b/inc/Engine/Preload/Database/Queries/Cache.php#L310-L311

Mai-Saad:
Test plan:
- confirm fixing the performance issue
- NRT plan execution https://wpmediaqa.testrail.io/index.php?/runs/view/898&group_by=cases:section_id&group_order=asc :heavy_check_mark:  (during execution, tried single site, multisite subdirectory nginx , subdomain multisite creating tables was good)

Khadreal:
> @wp-media/qa-team @Khadreal to discuss how to validate / test this. A quick feedback here: the main risk I see is that we are not using "LIKE" anymore but a strictly equal table name. We would need to check on different setups (multi-site, a few hosting, etc.) that we always have the exact table name. @Khadreal for instance, in most queries, we use $db->prefix, which does not seem to be the case in your PR. We'd need to ensure we are not missing a prefix mechanism here.
> 
> https://github.com/wp-media/wp-rocket/blob/b898c2a8b6f52d9fd93702a59419d57f208b0b4b/inc/Engine/Preload/Database/Queries/Cache.php#L310-L311

@MathieuLamiot 
Tested this on multisite and its work as it should. `$db->{$this->table_name}`  will return the table name with prefix. This `$this->get_db()` returns the db name with prefix https://github.com/wp-media/wp-rocket/blob/e951e59168ca0b3b3b1c378a1c2af70a883588bb/inc/Engine/Common/Database/Queries/AbstractQuery.php#L234

Miraeld:
Hello hello, 

I've conducted some tests and some logs of execution time between the old version and new one. 

- For a database with 100 tables, it decreased from an average of 15ms to 5ms, representing a 66% improvement in performance.
- When I scaled up to a database with 500 tables, the improvement was even more significant. The execution time decreased from an average of 75ms to 15ms, which is an 80% improvement.
- Finally, in a large-scale scenario with a database containing 100,000 tables, the execution time went down from an average of 2 seconds to 20-30 ms. 

So I confirm that it works as expected.

Miraeld (approved):

PR #6983, opened 2024-09-19
Fixes #6982 Cache is not purged when the path of cache folder contains a char that is part of the regular expression syntax

### LINKED ISSUE #6982: Cache is not purged when the path of cache folder contains a char that is part of the regular expression syntax
**To Reproduce**
1. Install wordpress in a path containing a char that is part of the regular expression syntax. Example:  install at ` /releases/2024-09-02T11:32:23+0000/` so the cache folder will be at  `/releases/2024-09-02T11:32:23+0000/wp-content/cache/wp-rocket/`
2. Purge the cache via the admin bar ( or other method)
3. The cache is not purged

**Why it happens**

the function `rocket_clean_files` uses the function `_rocket_get_cache_dirs` at

https://github.com/wp-media/wp-rocket/blob/de60b47e74f04f4d76f19f902d5b76c3c0e1b6cb/inc/functions/files.php#L595

`_rocket_get_cache_dirs` searches the folders by regex:
https://github.com/wp-media/wp-rocket/blob/de60b47e74f04f4d76f19f902d5b76c3c0e1b6cb/inc/functions/files.php#L1423-L1438
But `_rocket_normalize_path( $cache_path, true )` only escapes the slash but not other characters.

It should be replaced by `preg_quote( $cache_path, '/' )`

Kilbourne:
# Description

Fixes #6982
Fixes cache purging when the path of cache folder contains a char that is part of the regular expression syntax

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue).
- [ ] Release

# Mandatory Checklist

## Code validation

- [X] I validated all the Acceptance Criteria. If possible, provide screenshots or videos.
- [X] I triggered all changed lines of code at least once without new errors/warnings/notices.
- [X ] I implemented built-in tests to cover the new/changed code.

## Code style
- [X] I wrote a self-explanatory code about what it does.
- [X] I protected entry points against unexpected inputs.
- [X] I did not introduce unnecessary complexity.
- [X] Output messages (errors, notices, logs) are explicit enough for users to understand the issue and are actionnable.

# Additional Checks
- [X] In the case of complex code, I wrote comments to explain it.
- [X] When possible, I prepared ways to observe the implemented system (logs, data, etc.).
- [X] I added error handling logic when using functions that could throw errors (HTTP/API request, filesystem, etc.)

### DIFF
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1429,7 +1429,7 @@ function _rocket_get_cache_dirs( $url_host, $cache_path = '', $hard_reset = fals
 
 	$regex = sprintf(
 		'/%1$s%2$s(.*)/i',
-		_rocket_normalize_path( $cache_path, true ),
+		preg_quote( $cache_path, '/' ),
 		$url_host
 	);
 
--- a/inc/functions/files.php
+++ b/inc/functions/files.php
@@ -1429,7 +1429,7 @@ function _rocket_get_cache_dirs( $url_host, $cache_path = '', $hard_reset = fals
 
 	$regex = sprintf(
 		'/%1$s%2$s(.*)/i',
-		preg_quote( $cache_path, '/' ),
+		preg_quote( _rocket_normalize_path( $cache_path, true ), '/' ),
 		$url_host
 	);
 


inc/functions/files.php
```diff
@@ -1429,7 +1429,7 @@ function _rocket_get_cache_dirs( $url_host, $cache_path = '', $hard_reset = fals
 
 	$regex = sprintf(
 		'/%1$s%2$s(.*)/i',
-		_rocket_normalize_path( $cache_path, true ),
+		preg_quote( $cache_path, '/' ),
```
remyperona: The path normalize should not be removed, but we can do the quoting after:

```suggestion
		preg_quote( _rocket_normalize_path( $cache_path, true ), '/' ),
```

remyperona (changes requested):

remyperona (approved):

hanna-meda:
Hello @Kilbourne. Thank you for this PR. I reproduced the issue by renaming /wp-content to `2024-09-02T11:32:23+0000` (updated also the path for folder in wp-config) but then checking on the PR it was the same, cache files weren't purging. So I am not sure if these were the right steps to test or the issue is still not fixed. If you can have a second look. Thank you.

PR #6984, opened 2024-09-19
Don't cache not valid taxonomy frontend pages and stop applying optimizations

wordpressfan:
# Description

Fixes https://github.com/wp-media/wp-rocket.me/issues/4164

In this pull request, we are checking if the current frontend page is a taxonomy page and it has a not valid url schema, then we disable this page's caching, also stop applying all optimizations there.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as before).

## Detailed scenario

As mentioned in the issue itself:

> if i visit wordpress.org at:
https://wordpress.org/news/category/releases/ i get a response 200 and see the content.

> Then again if i change the url so the last part still exists, something like:
https://wordpress.org/news/category/i_just_added_something_here/releases/

### Documentation

Here we get all public taxonomies that have UI pages and compare the current page's query var with the taxonomies query vars.
We provided also a filter `rocket_cache_taxonomy_query_vars` so we can include or exclude taxonomy query vars from that functionality.

### New dependencies

No

### Risks

The only thing that I can think of, is that we will prevent those pages from being cached as they were already cached but that's to fix the mentioned issue.

Also for @wp-media/qa-team,
In this PR, we also added a change to check if the `rocket_buffer` filter returns empty HTML, then we will revert back to the main page's HTML, so if you used the following snippet as an example:
```
add_filter( 'rocket_buffer', function ( $html ) {
    return '';
} );
```
Before this PR, the page will be blank, but with this PR, we will show the not optimized page.

# Mandatory Checklist

## Code validation

## Code style

# Additional Checks


remyperona:
@wordpressfan I think the changes look good to fix the issue at hand, are you waiting for something before moving it forward to review?

wordpressfan:
This PR was a POC to validate the idea, so it needs to be adapted to be a real PR, like fixing phpcs and tests.

wordpressfan:
I'm working on tests now

remyperona (approved):

hanna-meda (approved):
Thank you, @wordpressfan. Everything looks good, exploratory and otherwise.
Related Tests Results Report below:
[testrail-report-650.pdf](https://github.com/user-attachments/files/17506069/testrail-report-650.pdf)

PR #6988, opened 2024-09-23
Closes #6987 Add dynamic list Automatic Lazy Render exclusions

### LINKED ISSUE #6987: New backend section to handle the Automatic Lazy Rendering exclusions
We have some different issues related to the layout problems with Automatic Lazy Rendering:
https://github.com/wp-media/wp-rocket/issues/6959

We'd like to be able to define HTML elements which will not receive hashes during the process (exclude from the feature). It'd allow us to exclude elements before the real fix is applied.

It should work with regex.

**Another approach**
Prevent from adding `data-wpr-lazyrender` attribute to the elements

**Additional context**
We should have a new backend section to handle it

remyperona:
# Description

Fixes #6987

Exclude patterns from ALR pulled from the backend with the dynamic lists

## Type of change

## Detailed scenario

Use the filter `rocket_lrc_exclusions` like `wpm_apply_filters_typed('array', 'rocket_lrc_exclusions', $array_of_exclusions )` where `$array_of_exclusions` is an array of excluded element. 
Then load a page and notice that hashes won't be added to excluded element(s).

### Documentation

A new filter has been added `rocket_lrc_exclusions` in order to add exclusions to LRC. 

# Mandatory Checklist

## Code validation

## Code style

### DIFF
--- a/inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php
+++ b/inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php
@@ -159,11 +159,11 @@ class Controller implements ControllerInterface {
 		/**
 		 * Filters the list of patterns to exclude from hash injection.
 		 *
-		 * @since 3.17
+		 * @since 3.17.0.2
 		 *
 		 * @param string[] $exclusions The list of patterns to exclude from hash injection.
 		 */
-		$exclusions = wpm_apply_filters_typed( 'array', 'rocket_lrc_exclusions', [] );
+		$exclusions = wpm_apply_filters_typed( 'string[]', 'rocket_lrc_exclusions', [] );
 
 		return $exclusions;
 	}
--- a/inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php
+++ b/inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php
@@ -145,8 +145,27 @@ class Controller implements ControllerInterface {
 		$processor = wpm_apply_filters_typed( 'string', 'rocket_lrc_processor', 'dom' );
 
 		$this->processor->set_processor( $processor );
+		$this->processor->get_processor()->set_exclusions( $this->get_exclusions() );
 
 		return $this->processor->get_processor()->add_hashes( $html );
+	}
+
+	/**
+	 * Get the list of patterns to exclude from hash injection.
+	 *
+	 * @return string[]
+	 */
+	private function get_exclusions(): array {
+		/**
+		 * Filters the list of patterns to exclude from hash injection.
+		 *
+		 * @since 3.17
+		 *
+		 * @param string[] $exclusions The list of patterns to exclude from hash injection.
+		 */
+		$exclusions = wpm_apply_filters_typed( 'array', 'rocket_lrc_exclusions', [] );
+
+		return $exclusions;
 	}
 
 	/**
--- a/inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Dom.php
+++ b/inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Dom.php
@@ -7,7 +7,6 @@ use DOMDocument;
 use WP_Rocket\Logger\Logger;
 
 class Dom implements ProcessorInterface {
-
 	use HelperTrait;
 
 	/**
@@ -35,7 +34,7 @@ class Dom implements ProcessorInterface {
 	 *
 	 * @var array
 	 */
-	private $exclusions;
+	private $exclusions = [];
 
 	/**
 	 * Sets the exclusions list
@@ -54,9 +53,7 @@ class Dom implements ProcessorInterface {
 	 * @return string
 	 */
 	private function get_exclusions_pattern(): string {
-		$exclusions = $this->exclusions;
-
-		if ( empty( $exclusions ) ) {
+		if ( empty( $this->exclusions ) ) {
 			return '';
 		}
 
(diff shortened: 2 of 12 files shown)


inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php
```diff
@@ -145,10 +145,29 @@ public function add_hashes( $html ) {
...
+	 */
+	private function get_exclusions(): array {
+		/**
+		 * Filters the list of patterns to exclude from hash injection.
+		 *
+		 * @since 3.17
+		 *
+		 * @param string[] $exclusions The list of patterns to exclude from hash injection.
+		 */
+		$exclusions = wpm_apply_filters_typed( 'array', 'rocket_lrc_exclusions', [] );
```
wordpressfan: Starting from typed library version 1.2, we started the support of using array of strings
https://github.com/wp-media/apply-filters-typed/releases/tag/v1.2

What about this suggestion?

```suggestion
		$exclusions = wpm_apply_filters_typed( 'string[]', 'rocket_lrc_exclusions', [] );
```

inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Dom.php
```diff
@@ -28,6 +27,46 @@ class Dom implements ProcessorInterface {
...
+	 * @return string
+	 */
+	private function get_exclusions_pattern(): string {
+		if ( empty( $this->exclusions ) ) {
+			return '';
+		}
+
+		$exclusions = array_map(
+			function ( $exclusion ) {
+				return preg_quote( $exclusion, '#' );
```
wordpressfan: I think we don't need to pass this `#` character, because we already use the `/` delimiter below in the `preg_match` call.

inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Dom.php
```diff
@@ -107,6 +146,16 @@ private function add_hash_to_element( $element, $depth, $html ) {
 			$child_html       = $child->ownerDocument->saveHTML( $child ); // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
 			$opening_tag_html = strstr( $child_html, '>', true ) . '>';
 
+			$exclusions_pattern = $this->get_exclusions_pattern();
+
+			if (
+				! empty( $exclusions_pattern )
+				&&
+				preg_match( '/(' . $exclusions_pattern . ')/i', $opening_tag_html )
```
wordpressfan, line 154: What about using RegexTrait find method?

https://github.com/wp-media/wp-rocket/blob/73fae2b3e7d2c7054f751bb9168cde175673d101/inc/Engine/Optimization/RegexTrait.php#L28-L45

inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Regex.php
```diff
@@ -27,6 +27,26 @@ class Regex implements ProcessorInterface {
 	 */
 	private $max_hashes;
 
+	/**
+	 * Array of patterns to exclude from hash injection.
+	 *
+	 * @since 3.17
```
wordpressfan:
```suggestion
	 * @since 3.17.0.2
```

inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/SimpleHtmlDom.php
```diff
@@ -30,6 +30,26 @@ class SimpleHtmlDom implements ProcessorInterface {
 	 */
 	private $max_hashes;
 
+	/**
+	 * Array of patterns to exclude from hash injection.
+	 *
+	 * @since 3.17
```
wordpressfan:
```suggestion
	 * @since 3.17.0.2
```

inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/SimpleHtmlDom.php
```diff
@@ -30,6 +30,26 @@ class SimpleHtmlDom implements ProcessorInterface {
...
+	/**
+	 * Sets the exclusions list
+	 *
+	 * @param string[] $exclusions The list of patterns to exclude from hash injection.
+	 *
+	 * @return void
+	 */
+	public function set_exclusions( $exclusions ): void {
+		$this->exclusions = $exclusions;
+	}
```
wordpressfan, line 51: I was thinking about having an abstract class to have those common functionalities instead of repeating this block of code across all processors.

I mean having `AbstractProcessor` which implements the interface and then extend all processors from it.

I know that we may remove other processors but this is a tiny step that will make the code more organized and maintainable.

what do u think?

wordpressfan (changes requested):

inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Dom.php
remyperona: this method is intended to find elements we're going to interact with, not just check if a pattern is in a string, I think that would be overkill to use it in this context.

inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/SimpleHtmlDom.php
remyperona: I don't see much value adding an abstract class that might be removed later, for something small too. We could revisit that later if common code between the implementations increases.

jeawhanlee (approved):

hanna-meda (approved):
@Tabrisrp, thank you for this PR.
Exploratory went smoothly, test execution too.
[testrail-report-638.pdf](https://github.com/user-attachments/files/17116828/testrail-report-638.pdf)

PR #7011, opened 2024-10-04
Adjust priority for rocket buffer

### LINKED ISSUE #891: Update callback priority on template_redirect for picture display
**Context**
Related to https://github.com/wp-media/wp-rocket/issues/6786

WP Rocket Lazyload exclusion is not working correctly when excluded pattern is part of a `picture` element created by the picture rewriting feature of Imagify.

**Expected beavhior**
Lazyload exclusion should work as expected

**Additional information**
So I believe I found what was happening with the next-gen issue. we have 2 callbacks on `template_redirect`:

- Imagify callback on priority -1000
- WP Rocket callback on priority 2

So we should expect imagify to be finished first, and WP Rocket after. BUT, both callbacks have an `ob_start()` call, and when multiple are called, they become nested until the buffering stops:

- Imagify opens buffer first
- then WP Rocket, nested inside

When output buffering ends (on shutdown action of WP), lower level content is passed to upper, so the order becomes WP Rocket optimizations first, then Imagify second, and HTML is output.
While callback priority is as expected, the output buffering works in the opposite order.

To verify, I tested by changing the priority of Imagify callback to 10, reverting both orders, and it works as expected, the picture element is applied first in the HTML, and then lazyload exclusion works as needed.

Khadreal:
# Description

Fixes https://github.com/wp-media/imagify-plugin/issues/891

## Type of change

## Detailed scenario
Check here for https://github.com/wp-media/imagify-plugin/pull/893#issuecomment-2298451180

## Technical description
Changed the priority of rocket buffer callbacks to 2/3 to fix the conflict issue with imagify image rewrite.

### Documentation

# Mandatory Checklist

## Code validation

## Code style

### DIFF
--- a/inc/Engine/CDN/Subscriber.php
+++ b/inc/Engine/CDN/Subscriber.php
@@ -45,8 +45,8 @@ class Subscriber implements Subscriber_Interface {
 	public static function get_subscribed_events() {
 		return [
 			'rocket_buffer'           => [
-				[ 'rewrite', 20 ],
-				[ 'rewrite_srcset', 21 ],
+				[ 'rewrite', 2 ],
+				[ 'rewrite_srcset', 3 ],
 			],
 			'rocket_css_content'      => 'rewrite_css_properties',
 			'rocket_usedcss_content'  => 'rewrite_css_properties',
--- a/tests/Integration/inc/Engine/CDN/Subscriber/rewrite.php
+++ b/tests/Integration/inc/Engine/CDN/Subscriber/rewrite.php
@@ -12,7 +12,7 @@ class Test_Rewrite extends TestCase {
 	public function set_up() {
 		parent::set_up();
 
-		$this->unregisterAllCallbacksExcept( 'rocket_buffer', 'rewrite', 20 );
+		$this->unregisterAllCallbacksExcept( 'rocket_buffer', 'rewrite', 2 );
 	}
 
 	public function tear_down() {


remyperona (commented):
Is this change necessary? With the change in Imagify it should not need to be modified on WP Rocket side

Khadreal:
> Is this change necessary? With the change in Imagify it should not need to be modified on WP Rocket side

Yes, it's necessary. 
With the change in imagify, when you use the rocketcdn within wpr the issue still persists, this fix that.

remyperona:
I'm worried about this change, as it could fix this specific case, but cause issues at the same time. The priority of features on `rocket_buffer` was specificly set to execute in the right order to avoid conflicts, so if we progress with this change, we will have to test every features and confirm they continue to work as expected.

Khadreal:
Should we revert it as it wasn't part of the initial issues but found out it during test. 
Also, when you look at the priority features, performance_hints happens before cdn.

remyperona:
We can keep this open and do QA with the changes in Imagify & RocketCDN. If everything is good, we can close it, if QA report an issue related to it, we will see how to go with this one.

Khadreal:
> We can keep this open and do QA with the changes in Imagify & RocketCDN. If everything is good, we can close it, if QA report an issue related to it, we will see how to go with this one.

Please could you review it so we can move it to QA

MathieuLamiot:
@Khadreal @wp-media/qa-team If we must first do QA on Imagify & RocketCDN PRs, just move this to ready for QA as theyr are reveiewed already and leave a comment explaining this properly in the issue.
You can even add the Imagify and RocketCDN PRs to the board if you want to track them individually.
Don't get blocked because of processes :)

Khadreal:
@wp-media/qa-team please check the comments [here](https://github.com/wp-media/imagify-plugin/issues/891#issuecomment-2404797889)

remyperona (approved):

PR #7032, opened 2024-10-15
Fixes #6892 Attempt to read property "id" on bool [...] AbstractQueries.php

### LINKED ISSUE #6892: Attempt to read property "id" on bool [...] AbstractQueries.php on line 63
Using 3.16.4, cron reports:
```
PHP Warning:  Attempt to read property "id" on bool in /wp-content/plugins/wp-rocket/inc/Engine/Common/PerformanceHints/Database/Queries/AbstractQueries.php](/wp-content/plugins/wp-rocket/inc/Engine/Common/PerformanceHints/Database/Queries/AbstractQueries.php) on line 63
```
Using PHP 8.1.

Khadreal:
# Description
Fixed error with attempt to read property on false

Fixes #6892 

## Type of change

## Detailed scenario
Check https://wp-media.slack.com/archives/C43T1AYMQ/p1728661812952229 for more detailed

## Technical description
Add a check before for the `item` object and if the id is set before attempting to delete

### Documentation

# Mandatory Checklist

## Code validation

## Code style


inc/Engine/Common/PerformanceHints/Database/Queries/AbstractQueries.php
```diff
@@ -60,7 +60,11 @@ public function delete_by_url( string $url ) {
 
 		$deleted = true;
 		foreach ( $items as $item ) {
-			$deleted = $deleted && $this->delete_item( $item->id );
+			if ( is_object( $item ) && isset( $item->id ) ) {
+				$deleted = $deleted && $this->delete_item( $item->id );
+			} else {
+				$deleted = false;
+			}
```
remyperona: we should avoid the `else` by inverting the condition (`! isset( $item->id )`), and using `continue;`

remyperona (changes requested):

remyperona (approved):

Mai-Saad:
Since there are no clear reproducible steps, here is a plan for smoke on clear performance hints https://wpmediaqa.testrail.io/index.php?/runs/view/913&group_by=cases:section_id&group_order=asc

inc/Engine/Common/PerformanceHints/Database/Queries/AbstractQueries.php
```diff
@@ -60,6 +60,11 @@ public function delete_by_url( string $url ) {
 
 		$deleted = true;
 		foreach ( $items as $item ) {
+			if ( ! is_object( $item ) || ! isset( $item->id ) ) {
+				$deleted = false;
+				continue;
+			}
+
```
wordpressfan: I think this means that if an item in the array contains not valid element (element is not object or doesn't have id) we will bypass the current loop iteration after setting deleted with false, then we use this deleted with `&&` at the end we will have the final deleted for all upcoming items to be false, which will make `$this->delete_item( $item->id )` a dead zone (not reachable)

Is this what we expect here? even if this is an edge case I wanted ti highlight it.

Khadreal: Yes, if the item is not a valid object and doesn't have id, we shouldn't get to `$this->delete_item( $item->id )` as that will throw and error since id is not present

wordpressfan: I don't think that I explained it well, I didn't mean the not valid item, I meant if we have an array of some valid items and some not valid items,in this case we will not call delete_item for all of them, is think expected?

Khadreal: Yes, we would only call the delete_item for the valid ones. which is why the deleted is set to false in that logic before continue

wordpressfan (approved):

Mai-Saad (approved):
Smoke test for delete URLs from DB is good.
[testrail-report-648.pdf](https://github.com/user-attachments/files/17408489/testrail-report-648.pdf)

PR #7050, opened 2024-10-22
Fixes #7049: Performance issue with table existence query

### LINKED ISSUE #7049: Performance issue with table existence query
**Describe the bug**
In the PR #6968 we started using `information_schema.tables` to detect the existence of our tables, exactly here:
https://github.com/wp-media/wp-rocket/blob/d99bce0f0270963b82d6112276fc29dd97363d4b/inc/Engine/Common/Database/Queries/AbstractQuery.php#L523

Then we got a report that this query will return values for all databases on the server even for databases that the user doesn't have access to.

**To Reproduce**

I don't have a clear way to reproduce the issue on our local sites but seems like we may need to create two mysql users on the same server and for each user, create a database, and see if the query will return values from all databases.

**Expected behavior**
We should only detect the table existence on the current database that we have proper access to.

**Possible solution**

we already use the same concept in database optimization query, here:
https://github.com/wp-media/wp-rocket/blob/73488699424f726bef28dbf2c934340efcab8e47/inc/Engine/Admin/Database/Optimization.php#L101

jeawhanlee:
# Description
This PR fixes the issue with user servers being impacted due to unoptimized query.

Fixes #7049 

## Type of change

## Detailed scenario

Not able to reproduce on local at the moment.

### Documentation

The query that checks if table exists is now optimized to specifically target the current DB instead of recurrently checking tables in all the Databases on the server.

### New dependencies

N/A

### Risks

N/A

# Mandatory Checklist

## Code validation

## Code style

### DIFF
--- a/inc/Engine/Common/Database/Queries/AbstractQuery.php
+++ b/inc/Engine/Common/Database/Queries/AbstractQuery.php
@@ -521,7 +521,7 @@ class AbstractQuery extends Query {
 
 		// Query statement.
 		$query    = 'SELECT table_name FROM information_schema.tables WHERE table_schema = %s AND table_name = %s LIMIT 1';
-		$prepared = $db->prepare( $query, $db->dbname, $db->{$this->table_name} );
+		$prepared = $db->prepare( $query, $db->__get( 'dbname' ), $db->{$this->table_name} );
 		$result   = $db->get_var( $prepared );
 
 		// Does the table exist?
--- a/inc/Engine/Common/Database/Queries/AbstractQuery.php
+++ b/inc/Engine/Common/Database/Queries/AbstractQuery.php
@@ -520,8 +520,8 @@ class AbstractQuery extends Query {
 		}
 
 		// Query statement.
-		$query    = 'SELECT table_name FROM information_schema.tables WHERE table_name = %s LIMIT 1';
-		$prepared = $db->prepare( $query, $db->{$this->table_name} );
+		$query    = 'SELECT table_name FROM information_schema.tables WHERE table_schema = %s AND table_name = %s LIMIT 1';
+		$prepared = $db->prepare( $query, $db->dbname, $db->{$this->table_name} );
 		$result   = $db->get_var( $prepared );
 
 		// Does the table exist?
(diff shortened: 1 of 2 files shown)


Miraeld:
There is this one to change as well https://github.com/wp-media/wp-rocket/blob/d99bce0f0270963b82d6112276fc29dd97363d4b/inc/Engine/Common/Database/Queries/AbstractQuery.php#L509-L535

Miraeld (changes requested):
There is this one to change as well https://github.com/wp-media/wp-rocket/blob/d99bce0f0270963b82d6112276fc29dd97363d4b/inc/Engine/Common/Database/Queries/AbstractQuery.php#L509-L535

remyperona (approved):
For later: we should see how to avoid having duplicated code for this

Miraeld (approved):

Mai-Saad:
Test plan:
- Reproduce the issue and confirm that PR fixed that
- NRT https://wpmediaqa.testrail.io/index.php?/runs/view/917&group_by=cases:section_id&group_order=asc

Miraeld:
Hello, 
NRT is green :) 
[testrail-report-652.pdf](https://github.com/user-attachments/files/17540150/testrail-report-652.pdf)

Miraeld:
Hello hello, 

I've been trying to reproduce the issue in a local environment. 
To do this, I made a Docker environment with 2 main containers: 1 Wordpress, and 1 DB server. 

On this DB server, I've created 10 000 databases where each one of them has their own user and these aren't SUPER user and of course, there is a single SUPER User that has access to everything. 

So to make it short.

10 000 databases, they all have the same named tables. So in my example I'll use `table_1`.

There is 1 SUPER User, otherwise, each database have a different user which isn't SUPER.


From the SUPER User:
The query: `SELECT table_name FROM information_schema.tables;` returns all tables in every database, which means if I do
`SELECT table_name FROM information_schema.tables WHERE table_name = 'table_1';` I'll get 10k rows results.



Now, from a normal user from 1 db,
If I run `SELECT table_name FROM information_schema.tables` I only get the table of its own database.

and `SELECT table_name FROM information_schema.tables WHERE table_name = 'table_1' LIMIT 1;` (similar to what we had in WP Rocket) runs within 0.017s

While the new query introduced by [#7049](https://github.com/wp-media/wp-rocket/issues/7049)
is `SELECT table_name FROM information_schema.tables WHERE table_schema = 'wordpress' AND table_name = 'table_1' LIMIT 1`  runs within 0.002s.

So we can see an improvement, as we are reducing the execution time of the query by almost 10.

PR #7079, opened 2024-10-29
Rmove only the version in the meta generator content attribute

remyperona:
# Description

Remove only the version instead of the full content attribute when we use the related filter.

## Type of change

## Detailed scenario

New filter name is `rocket_display_meta_generator_content_version`. By default, it adds the version to the `content` attribute. When `false`, the version is removed.

### Documentation

Updated the code to only change the display of the version

# Mandatory Checklist

## Code validation

## Code style

### DIFF
--- a/inc/Engine/Support/Meta.php
+++ b/inc/Engine/Support/Meta.php
@@ -123,20 +123,20 @@ class Meta {
 			return '';
 		}
 
-		$content = '';
+		$version = '';
 
 		/**
-		 * Filters the display of the content attribute in the meta generator tag.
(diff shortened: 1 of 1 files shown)


Khadreal (approved):

Mai-Saad:
Working as expected now
When the filter is false now, we see only WP Rocket (Note: filter name changed to rocket_display_meta_generator_content_version)
ex: `<meta name="generator" content="WP Rocket" data-wpr-features="wpr_desktop wpr_preload_links" />`
While if filter is true, the version is there
`<meta name="generator" content="WP Rocket 3.17.2-alpha1" data-wpr-features="wpr_desktop wpr_preload_links" />`

PR #7081, opened 2024-10-31
Closes #7058: Implement DJE Safe Mode Feature

### LINKED ISSUE #7058: Implement DJE Safe Mode Feature
- **Frontend Changes**:

Remove the placeholder from the textarea.
Add a checkbox labeled "Delay JavaScript Execution safe mode" before the textarea.
Display a warning message when the checkbox is checked, detailing potential performance impacts and suggesting users contact support.

- **Backend Changes**:

Implement logic to handle the checkbox state.
Update the exclusions list based on the checkbox state.
Ensure the system correctly applies these changes when the settings are saved.

Miraeld:
# Description

Fixes #7058 
This change introduces the "Delay JavaScript Execution safe mode" feature. It enhances user control over JavaScript execution, providing a checkbox to enable or disable safe mode, and displays a warning about potential performance impacts. This helps users manage their site's performance more effectively.

## Type of change

## Detailed scenario

To trigger the new code:

- Navigate to the plugin settings.
- Locate the "Delay JavaScript Execution" section.
- Check the "Delay JavaScript Execution safe mode" checkbox.
- Observe the warning message detailing potential performance impacts.
- Default exclusion will be added on the front-end.

### Documentation

The feature adds a checkbox labeled "Delay JavaScript Execution safe mode" and displays a warning when it is checked. The backend logic updates the exclusions list based on the checkbox state and ensures these changes are applied when settings are saved.
Also, when we load `delay_js` on the front-end, we check if the new option is enabled, and if this is the case, we add default exclusion to delay_js.

### New dependencies

No new dependencies are introduced in this change.

### Risks

None.

# Mandatory Checklist

## Code validation

## Code style


inc/Engine/Admin/Settings/Page.php
```diff
@@ -792,13 +790,38 @@ private function assets_section() {
 					],
 					'items'             => $this->delayjs_sitelist->prepare_delayjs_ui_list(),
 				],
+				'delay_js_execution_safe_mode' => [
+					'type'              => 'checkbox',
+					'label'             => __( 'Delay JavaScript Execution safe mode', 'rocket' ),
+					// translators: %1$s = opening <a> tag, %2$s = closing </a> tag.
+					'description'       => $invalid_license ? __( 'Delay JavaScript Execution safe mode temporarily resolves issues with Delay JavaScript execution but may reduce your PageSpeed Scores and performance. Contact support for help excluding problematic scripts to use this feature fully.', 'rocket' ) : '',
```
remyperona: do we need the license check?

remyperona (changes requested):

inc/Engine/Admin/Settings/Page.php
Miraeld: Nope, not really, I've deleted it.

Miraeld:
`Codacy Diff Coverage` is failing due to the addition of a field on the front-end which isn't triggered in any test (as it's front-end). That's normal 👍

inc/Engine/Optimization/DelayJS/HTML.php
```diff
@@ -101,14 +101,25 @@ public function delay_js( $html ): string {
...
 		$this->excluded = array_merge( $this->excluded, $this->options->get( 'delay_js_exclusions_selected_exclusions', [] ) );
 
+		if ( $this->options->get( 'delay_js_execution_safe_mode', 0 ) ) {
+			$this->excluded = array_merge(
+				$this->excluded,
+				[
+					'/jquery(-migrate)?-?([0-9.]+)?(.min|.slim|.slim.min)?.js(\?(.*))?( |\'|"|>)',
+					'js-(before|after)',
+					'(?:/wp-content/|/wp-includes/)(.*)',
+				]
```
remyperona: Is there a reason you're using the hardcoded values instead of the static method here?

inc/Engine/Optimization/DelayJS/HTML.php
```diff
@@ -101,14 +101,25 @@ public function delay_js( $html ): string {
...
+
 		/**
 		 * Filters the delay JS exclusions array
 		 *
 		 * @since 3.9
 		 *
 		 * @param array $excluded Array of excluded patterns.
 		 */
-		$this->excluded = (array) apply_filters( 'rocket_delay_js_exclusions', $this->excluded );
+		$this->excluded = wpm_apply_filters_typed( 'array', 'rocket_delay_js_exclusions', $this->excluded );
```
remyperona: You can use `string[]` to be more accurate in the filtering

inc/Engine/Optimization/DelayJS/HTML.php
Miraeld: Umm nope, I forgot to change it, I've used the hardcoded for some tests during the tests fixing stage. Forgot to revert back to the static method. Thanks !

remyperona (approved):

Mai-Saad:
Testing notes [WIP , to be checked with PO what shall be changed here or on other GH]
- checkbox isnot before the exclusion text area
- The text for the checkbox is different than on the notion doc too
- warning text when checking the option is almost the same as the option description
- DJS exclusion safe mode is not unchecked when enabling safe mode while deactivating WPR, shouldn't it be unchecked?

Mai-Saad:
@Miraeld Thanks For the PR.  Can you please check the points below:
1- The checkbox shall be after the exclusion area not before it.
2- The wording shall match what's mentioned here https://www.notion.so/wpmedia/Turn-default-exclusions-to-a-checkbox-10eed22a22f080429916d8b0e8a384be?pvs=4#113ed22a22f080e7b452d8485b0cea5d
3- When check the option the :heavy_check_mark: is not there till we accept the warning which is not the same behavior as when enabling RUCSS for example. =====> same with combine js . @wp-media/product what do you think?

Current o/p on PR

Screenshot from 2024-12-02 11-25-19

Miraeld:
@Mai-Saad Feedbacks 1 & 2 have been fixed. 

About the third feedback:
> 3- When check the option the ✔️ is not there till we accept the warning which is not the same behavior as when enabling RUCSS for example. =====> same with combine js . @wp-media/product what do you think?

If I may add something, I find it more logic to be like it is now, and I'll explain why. 

If you take the RUCSS as example, when you click on the option you get the ☑️ however it won't be enable until you select one of the method. If you don't click on one of the method, but save, it won't be ☑️ anymore. That's a bit confusing. 

However here, you won't get the ☑️ until you click "ACTIVATE SAFE MODE", which means the user confirm their intention to use the safemode. 

Just my point of view.

wordpressfan (approved):

Mai-Saad:
@Miraeld Thanks for the update.
When we update WPR to PR while having default exclusions in text area, we remove it from the text area. As per previous discussion with @piotrbak we should have no connection now between the text area and the safe mode checkbox. /cc @wp-media/product  (note: now import settings with default exclusions, it won't be added in text area as well)

Miraeld:
Hello hello @Mai-Saad 

I've made some changes and  mostly on the regex to be considered as "default" exclusion of the safemode. 
Before we had `/jquery(-migrate)?-?([0-9.]+)?(.min|.slim|.slim.min)?.js(\?(.*))?( |'|"|>)` while it should be `\/jquery(-migrate)?-?([0-9.]+)?(.min|.slim|.slim.min)?.js(\?(.*))?( |'|"|>)` 
So it's been changed. 

Also, I've made some modification on the condition to delete default exclusion brought by safemode from the textarea only when the safemode is enabled, following this [comment](https://www.notion.so/Turn-default-exclusions-to-a-checkbox-10eed22a22f080429916d8b0e8a384be?d=118ed22a22f0806194f3001c900ab792&pvs=4#113ed22a22f0803283c9fa7d237fd59a).

Until now, we were deleting from the textarea even tho the safemode was disabled. 
And that's why on update we were deleting the default exclusions from the textarea. That's not the case anymore. 


--- 
A technical update of the whole PR:

This pull request introduces a new feature called "Safe Mode for Delay JavaScript Execution" and includes several related changes across multiple files. The most important changes are grouped by theme below:

### Feature Addition:
* Added a new checkbox option `delay_js_execution_safe_mode` to the settings page, which prevents all internal scripts from being delayed when enabled.
* Implemented the `get_safe_mode_exclusions` method to provide a list of default exclusions for the safe mode.
* Updated the `delay_js` method in `HTML.php` to include safe mode exclusions if the safe mode is enabled.

### Settings and Options:
* Updated the `add_options` and `sanitize_options` methods to include the new `delay_js_execution_safe_mode` option. [[1]](diffhunk://#diff-a2db9e9e29c84c9dbe050cfa06c5b17573d76151e16d634a6f9fc3f3d65edc4cR43-R51) [[2]](diffhunk://#diff-a2db9e9e29c84c9dbe050cfa06c5b17573d76151e16d634a6f9fc3f3d65edc4cR92-R104)
* Modified the constructor and other relevant methods in `Settings.php` to handle the new safe mode option. [[1]](diffhunk://#diff-a2db9e9e29c84c9dbe050cfa06c5b17573d76151e16d634a6f9fc3f3d65edc4cR30) [[2]](diffhunk://#diff-a2db9e9e29c84c9dbe050cfa06c5b17573d76151e16d634a6f9fc3f3d65edc4cR83)

### User Interface:
* Added JavaScript to automatically uncheck the safe mode checkbox when the main delay JS option is disabled.

### Testing:
* Updated test fixtures and test cases to account for the new `delay_js_execution_safe_mode` option. [[1]](diffhunk://#diff-0dff5ea78d2bdfbfb6b333b2d70b9b066b0adefe569dd43caa704e40b77d7e2bR11) [[2]](diffhunk://#diff-2579f349def341a58616ed1e31755a44506252f8c8828f12cb6c33396a1e2e00L9-R15) [[3]](diffhunk://#diff-39432b248522a33fdcd96f9ec6cd8d97014873ad144643bcde6ac67e0bb4103cR11) [[4]](diffhunk://#diff-05037e8fa499d87ef86f995eeefa1cf81d059acbc858d8d79fc164532ca48b71R848)

These changes ensure that the new safe mode feature is integrated into the existing delay JS functionality and properly tested.

jeawhanlee (approved):

Mai-Saad:
@Miraeld thanks for the update. Now updating while having default exclusions, will keep it in the text area and safe mode is unchecked.
But
1. When enable safe mode, if we have default exclusion in text area , it will be removed (unchecking safe mode won't add the text back) ==> This was supposed to be implemented later  ===> accepted as is for now, may need an enhancement later as per @piotrbak 
2. When enable safe mode of deactivation, neither DJS save mode nor the one click exclusion gets unchecked
@wp-media/product What do you think? is this acceptable now? do we need any other GH? ==> As per discussion with @piotrbak Fine for now as DJS itself is off

PR #7083, opened 2024-11-01
Closes #7062 Fonts filesystem

### LINKED ISSUE #7062: 3.18 - Backend and filesystem part
**User Story**
As a user, I’d like to host Google Fonts locally to better control the fonts and improve LCP.

**Acceptance Criteria**
- The content of combined Google Fonts CSS file will be saved into the local CSS file, we'll name the CSS with the value of the original URL hash and save it into the path like we do for Used CSS: (2/e/4/h/aerw32eweq.css)
- The feature will handle V1 and V2 separately
- The feature will automatically copy combined CSS files in a sub-folder under wp-content, specifically `/wp-content/cache/fonts/google-fonts/`
- The feature will rewrite URLs of the fonts found in the CSS file into the local path
- The feature will download fonts to the path that would preserve the original URL structure in the /fonts/ directory: `/wp-content/cache/fonts/google-fonts/fonts/s/caudex/v17/esDS311QOP6BJUr4yMKPtbo-Ew.woff2` (input: `https://fonts.gstatic.com/s/caudex/v17/esDS311QOP6BJUr4yMKPtbo-Ew.woff2`)
- The feature will create structure correctly for the multisites

**Comment**
- Be sure to deliver testable output, to be enabled with the pre_rocket_get filter

Khadreal:
# Description
Add a working prototype for font optimisation backend filesystem

Fixes #7062

## Type of change

- [x] Sub-task of #7062

## Detailed scenario
N/A

### Documentation
This feature allow for the download of media fonts locally, it introduce a new namespace `Media\Fonts` with three new classes Fonts, Subscriber and Filesystem.
The method `write_font_css` accepts two arguments, the font url and the provider and use those information to download the font and store locally. When the Filesystem class is initialised there's a setter method to declare the version of the fonts

# Mandatory Checklist

## Code validation

## Code style


inc/Engine/Media/Fonts/Controller/Filesystem.php
```diff
@@ -0,0 +1,217 @@
...
+	 * @param string $provider The url of the page.
+	 *
+	 * @return bool
+	 */
+	public function write_font_css( string $font_url, string $provider ): bool {
+		global $wp;
+		$url = untrailingslashit( home_url( add_query_arg( [], $wp->request ) ) );
+
+		$font_provider_path = $this->get_font_provider_path( $provider );
+		$hash_url           = $this->hash_url( $url );
```
remyperona: the hash should be based on the Google Fonts URL, not the current page URL

inc/Engine/Optimization/GoogleFonts/Combine.php
```diff
@@ -75,7 +75,19 @@ public function optimize( $html ): string {
...
+		$html = preg_replace( '@<\/title>@i', '$0' . $this->get_optimized_markup( $font_url ), $html, 1 );
+
+		/**
+		 * Filter v1 google fonts that will be stored locally.
+		 *
+		 * @param string $font_url The combined url of the fonts
+		 * @param string $provider The fonts provider
+		 * @param integer $version  The version of Google fonts api
+		 */
+		wpm_apply_filters_typed( 'null', 'rocket_host_google_fonts', $font_url, 'google-font', $version );
```
remyperona: I don't understand this part, the process of hosting locally should be done on `rocket_buffer`, after the combine process has happened

inc/Engine/Optimization/GoogleFonts/CombineV2.php
```diff
@@ -65,8 +65,18 @@ public function optimize( $html ): string {
...
 		$html         = preg_replace( '@<\/title>@i', '$0' . $combined_tag, $html, 1 );
+		$version      = 2;
+		/**
+		 * Filter v2 google fonts that will be stored locally.
+		 *
+		 * @param string $font_url The combined url of the fonts
+		 * @param string $provider The fonts provider
+		 * @param integer $version  The version of Google fonts api
+		*/
+		wpm_apply_filters_typed( 'null', 'rocket_host_google_fonts', $font_url, 'google-font', $version );
```
remyperona: same here

inc/Engine/Media/Fonts/Controller/Filesystem.php
```diff
@@ -0,0 +1,217 @@
+<?php
+
+declare(strict_types=1);
+
+namespace WP_Rocket\Engine\Media\Fonts\Controller;
+
+class Filesystem {
```
remyperona: is there any possibility of re-factoring the filesystem class in Used CSS and  this one for common parts?

inc/Engine/Media/Fonts/Controller/Filesystem.php
Khadreal: I used the page URL so the frontend can use it to get the fonts and rewrite.

inc/Engine/Common/AbstractFileSystem.php
```diff
@@ -0,0 +1,97 @@
...
+	 *
+	 * @return bool
+	 */
+	protected function create_directory( string $dir_path ): bool {
+		if ( ! $this->filesystem->exists( $dir_path ) ) {
+			rocket_mkdir_p( $dir_path );
+		}
+
+		return $this->filesystem->is_writable( $dir_path );
+	}
```
jeawhanlee: This method is confusing, is it creating a new directory or it's checking if a directory path is writable? maybe we change the name.

inc/Engine/Media/Fonts/Controller/Filesystem.php
```diff
@@ -0,0 +1,247 @@
...
+		}
+
+		preg_match_all( '/url\((https:\/\/[^)]+)\)/', $css_content, $matches );
+		$font_urls = $matches[1];
+		$local_css = $css_content;
+
+		foreach ( $font_urls as $font_url ) {
+			$parsed_url = wp_parse_url( $font_url );
+			$path_parts = explode( '/', trim( $parsed_url['path'], '/' ) );
+			$local_path = $file . '/' . implode( '/', $path_parts );
```
jeawhanlee: Why are we exploding the format only to implode back to the same format?

inc/Engine/Media/Fonts/Controller/Filesystem.php
```diff
@@ -0,0 +1,247 @@
...
+
+			if ( ! rocket_mkdir_p( $local_dir ) ) {
+				continue;
+			}
+
+			if ( ! file_exists( $local_path ) ) {
+				$font_content = $this->download_font( $font_url );
+
+				if ( ! $font_content ) {
+					continue;
```
jeawhanlee: Should we not be adding a log here before that font download was not successful before continuing the loop?

inc/Engine/Media/Fonts/Controller/Filesystem.php
```diff
@@ -0,0 +1,247 @@
...
+				}
+
+				$this->write_file( $local_path, $font_content );
+			}
+
+			$local_url = content_url( $relative_path . implode( '/', $path_parts ) );
```
jeawhanlee: Why are we also imploding here again?

inc/Engine/Media/Fonts/Controller/Filesystem.php
```diff
@@ -0,0 +1,247 @@
...
+	private function get_font_provider_path( string $provider ): string {
+		$provider = str_replace( '_', '-', $provider );
+
+		return $provider . '/';
+	}
+
+	/**
+	 * Deletes the locally stored fonts for the corresponding url
+	 *
+	 * @since 3.11.4
```
jeawhanlee:
```suggestion
	 * @since 3.18
```

inc/Engine/Media/Fonts/Controller/Filesystem.php
```diff
@@ -0,0 +1,247 @@
...
+		$dir = $this->get_fonts_full_path( $this->get_font_provider_path( $url ), $url );
+
+		return $this->delete_file( $dir );
+	}
+
+	/**
+	 * Deletes all the font CSS files
+	 */
+	public function delete_all_font_css() {
+		// TODO:create method to recursively delete all locally stored fonts.
```
jeawhanlee: Any reason why the complete version of this method is not shipping with this PR?

inc/Engine/Media/Fonts/Controller/Filesystem.php
```diff
@@ -0,0 +1,247 @@
...
+	 *
+	 * @param string $hash Hash value.
+	 *
+	 * @return string
+	 */
+	private function hash_to_path( string $hash ): string {
+		/**
+		 * Filters the number of sub-folders level to create for used CSS storage
+		 *
+		 * @since 3.11.4
```
jeawhanlee:
```suggestion
		 * @since 3.18
```

inc/Engine/Media/Fonts/Controller/Fonts.php
```diff
@@ -0,0 +1,58 @@
...
+	/**
+	 * Start the process of downloading font locally
+	 *
+	 * @param string  $font_url URL of the font to be saved locally.
+	 * @param string  $provider Provider of the font.
+	 * @param integer $version  Version.
+	 *
+	 * @return void
+	 */
+	public function process( string $font_url, string $provider, int $version ): void {
```
jeawhanlee: Where is this method used in this PR?

inc/Engine/Media/Fonts/ServiceProvider.php
```diff
@@ -21,6 +26,9 @@ class ServiceProvider extends AbstractServiceProvider {
 	 * @var array
 	 */
 	protected $provides = [
+		'media_fonts_frontend_subscriber',
```
jeawhanlee: We have a duplicate of this service in the array

inc/Engine/Optimization/RUCSS/Controller/Filesystem.php
```diff
@@ -6,14 +6,15 @@
...
 
-class Filesystem {
+class Filesystem extends AbstractFileSystem {
 	/**
 	 * WP Filesystem instance
 	 *
 	 * @var WP_Filesystem_Direct
 	 */
-	private $filesystem;
+	protected $filesystem;
```
jeawhanlee, line 17: Is this still being used?

jeawhanlee (changes requested):

inc/Engine/Media/Fonts/Controller/Filesystem.php
Khadreal: This was not part of the ACs

inc/Engine/Media/Fonts/Controller/Fonts.php
Khadreal: None. Added it at the initial stage but when we break down the task, it's going to be on the frontend side to implement `write_font_css` method.

Khadreal: Do you think we should remove it and the other task will determine how best to implement it process?

inc/Engine/Optimization/RUCSS/Controller/Filesystem.php
Khadreal: Yes

inc/Engine/Common/AbstractFileSystem.php
Khadreal: change method name to `maybe_create_directory`

inc/Engine/Media/Fonts/Controller/Fonts.php
jeawhanlee: I think we can leave it but create a follow-up PR to immediately implement it as it's a very important part of the feature itself.

inc/Engine/Media/Fonts/Controller/Filesystem.php
jeawhanlee: I think if it's not part of the AC, we should remove it so we don't end up with uncompleted method in case you're not the one going to work on the follow-up PR.

inc/Engine/Common/AbstractFileSystem.php
jeawhanlee: Can it not be the same name it used to be before it was isolated: `is_writable_folder`?

Khadreal: That won't work cause the method is signature is different, it accepts a parameter.

jeawhanlee (approved):

wp-rocket.php
```diff
@@ -68,6 +68,7 @@
 define( 'WP_ROCKET_CRITICAL_CSS_PATH',  WP_ROCKET_CACHE_ROOT_PATH . 'critical-css/' );
 
 define( 'WP_ROCKET_USED_CSS_PATH',  WP_ROCKET_CACHE_ROOT_PATH . 'used-css/' );
+define( 'WP_ROCKET_MEDIA_FONT_CSS_PATH',  WP_ROCKET_CACHE_PATH . 'fonts/' );
```
remyperona: We should avoid adding more global constants, that can be defined at the class level

wp-rocket.php
```diff
@@ -77,6 +78,7 @@
 define( 'WP_ROCKET_CACHE_BUSTING_URL', WP_ROCKET_CACHE_ROOT_URL . 'busting/' );
 
 define( 'WP_ROCKET_USED_CSS_URL', WP_ROCKET_CACHE_ROOT_URL . 'used-css/' );
+define( 'WP_ROCKET_MEDIA_FONT_CSS_URL', WP_ROCKET_CACHE_ROOT_URL . 'wp-rocket/fonts/' );
```
remyperona: Same here

inc/Engine/Common/AbstractFileSystem.php
```diff
@@ -0,0 +1,97 @@
...
+
+	/**
+	 * Get the content of a file
+	 *
+	 * @param string $file The file content to get.
+	 *
+	 * @return string
+	 */
+	protected function get_file( string $file ): string {
```
remyperona: I think the naming of the method should be different, we are getting the content of the file, not the file itself

inc/Engine/Media/Fonts/Controller/Filesystem.php
```diff
@@ -0,0 +1,242 @@
...
+			$parsed_url = wp_parse_url( $font_url );
+			$font_path  = $parsed_url['path'];
+			$local_path = $file . $font_path;
+			$local_dir  = dirname( $local_path );
+
+			if ( ! rocket_mkdir_p( $local_dir ) ) {
+				continue;
+			}
+
+			if ( ! file_exists( $local_path ) ) {
```
remyperona: Should be using the WP Filesystem function instead of the PHP native function

inc/Engine/Media/Fonts/Controller/Filesystem.php
```diff
@@ -0,0 +1,242 @@
...
+	/**
+	 * Download font from external url
+	 *
+	 * @param string $url Url of the file to download.
+	 *
+	 * @return mixed
+	 */
+	private function download_font( string $url ) {
+		$content = wp_remote_retrieve_body(
+			wp_remote_get(
```
remyperona:
```suggestion
			wp_safe_remote_get(
```

inc/Engine/Media/Fonts/Controller/Filesystem.php
```diff
@@ -0,0 +1,242 @@
...
+
+	/**
+	 * Download font from external url
+	 *
+	 * @param string $url Url of the file to download.
+	 *
+	 * @return mixed
+	 */
+	private function download_font( string $url ) {
+		$content = wp_remote_retrieve_body(
```
remyperona: This method is missing some checks, notably with the response code. You don't want to store a 404 response body into the font file.

remyperona (changes requested):

remyperona (approved):

Issue #7039, opened 2024-10-17
Host Google Fonts Locally


MathieuLamiot:
[EPIC Card](https://www.notion.so/wpmedia/Host-Google-Fonts-Locally-ecdd3c5ee4f84a008b1264626b88b472?pvs=4)

This is the main Github issue to track this EPIC ; to be groomed by @Tabrisrp and @Khadreal into sub-tasks to list here.

MathieuLamiot:
@Tabrisrp, @Khadreal, when you will be ready to start looking into this (Sprint 117, once your on-going is cleared), please organize a meeting (or ping me to facilitate this) with @piotrbak and @wp-media/qa-team so that we start with breaking down the EPIC into user stories. Then, we can work on grooming them. Thank you!

PR #7194, opened 2024-12-17
Closes #7189 Update self-host google fonts exclusion

### LINKED ISSUE #7189: 3.18 - Exclusions part enhancement
User Story
As a user, I’d like to be able to exclude CSS file with the fonts from the feature by typing its argument into the filter

**Acceptance Criteria**
Having following input:
```
`<link rel='stylesheet' id='astra-google-fonts-css' href='https://fonts.googleapis.com/css?family=Noto+Sans%3A400%2C700%7CMontserrat%3A700&#038;display=fallback&#038;ver=4.7.0' media='all' />`
```
- The following filter entries would exclude the CSS file from the GF Combination and from rewriting the paths: `astra-google-` `astra-google-(.*)-css` `astra-google-fonts-css`
- We could apply it only to `id` or `class` attributes, but if it makes things easier, it can be applied to all
- Fonts which are inside the excluded file will not be rewritten

remyperona:
# Description

Fixes #7189

## Type of change

- [x] Sub-task of #7039 

## Detailed scenario

It's now possible to exclude based on any element in the HTML tag for google fonts instead of only the URL.

### Documentation

Updated the code to use the full HTML tag as the subject instead of only the URL

# Mandatory Checklist

## Code validation

## Code style

### DIFF
--- a/inc/Engine/Media/Fonts/FontsTrait.php
+++ b/inc/Engine/Media/Fonts/FontsTrait.php
@@ -22,14 +22,14 @@ trait FontsTrait {
 	}
 
 	/**
-	 * Checks if a URL is excluded based on the provided exclusions.
+	 * Checks if a font is excluded based on the provided exclusions.
 	 *
-	 * @param string   $url        The URL to check.
+	 * @param string   $subject    The string to check.
 	 * @param string[] $exclusions The list of exclusions.
 	 *
 	 * @return bool True if the URL is excluded, false otherwise.
 	 */
-	protected function is_excluded( string $url, array $exclusions ): bool {
+	protected function is_excluded( string $subject, array $exclusions ): bool {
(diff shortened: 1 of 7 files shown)


jeawhanlee:
@remyperona Don't we also need to update `WP_Rocket\Engine\Optimization\GoogleFonts\{Combine, CombineV2}`?

remyperona:
You're right I forgot about it, it's updated

jeawhanlee (approved):

hanna-meda (approved):
Thank you, @remyperona, for this PR.
Confirming that we can now use filter to exclude GFs based on any attribute in the HTML tag. 
[testrail-report-672.pdf](https://github.com/user-attachments/files/18202649/testrail-report-672.pdf)

PR #7198, opened 2024-12-19
Close #6863 Add termly compatibility

### LINKED ISSUE #6863: Compatibility with Termly
[Termly](https://termly.io/) has an `auto blocking` mode that requires their script to be loaded early without any loading strategy applied. Otherwise, it will interfere with the plugin's functionality (intercept 3rd party scripts and decide if they can be executed or not depending on consent).

The indicator for this setting is stored on the options table under `termly_display_auto_blocker` set to `on`. 

The need is to exclude their main script `app.termly.io` (https://app.termly.io/resource-blocker/website-hash-identifier) from `defer` and `delay` when applicable.

Have a compatibility file that will skip `defer` and `delay` when `termly_display_auto_blocker` is set to `on`

N/A

**Additional context**
Conversation with the Termly team - https://group-onecom.slack.com/archives/C07FUEY1939/p1723453296721739

Khadreal:
# Description
Exclude Termly from delay and defer when applicable.

Fixes #6863

## Type of change

## Detailed scenario
Check #6863

### Documentation

Exclude Termly resources if `termly_display_auto_blocker` is set 'on' 

# Mandatory Checklist

## Code validation

## Code style

### DIFF
--- a/inc/ThirdParty/Plugins/Cookie/Termly.php
+++ b/inc/ThirdParty/Plugins/Cookie/Termly.php
@@ -15,9 +15,7 @@ class Termly implements Subscriber_Interface {
 	 * @return array
 	 */
 	public static function get_subscribed_events() {
-		$termly_version = defined( 'TERMLY_VERSION' );
-
-		if ( ! $termly_version ) {
+		if ( ! defined( 'TERMLY_VERSION' ) ) {
 			return [];
 		}
 
@@ -49,6 +47,10 @@ class Termly implements Subscriber_Interface {
 	 * @return array
 	 */
 	public function exclude_defer_js( array $exclude_delay_js ): array {
+		if ( ! $this->should_exclude() ) {
+			return $exclude_delay_js;
+		}
+
 		$exclude_delay_js[] = 'app.termly.io/resource-blocker/(.*)';
 
 		return $exclude_delay_js;
--- a/inc/ThirdParty/Plugins/Cookie/Termly.php
+++ b/inc/ThirdParty/Plugins/Cookie/Termly.php
@@ -1,0 +1,56 @@
+<?php
+declare(strict_types=1);
+
+namespace WP_Rocket\ThirdParty\Plugins\Cookie;
+
+use WP_Rocket\Event_Management\Subscriber_Interface;
+
+/**
+ * Compatibility class for Termly.
+ */
+class Termly implements Subscriber_Interface {
+	/**
+	 * Return an array of events that this subscriber wants to listen to.
+	 *
+	 * @return array
+	 */
+	public static function get_subscribed_events() {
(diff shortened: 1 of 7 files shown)


inc/ThirdParty/Plugins/Cookie/Termly.php
```diff
@@ -0,0 +1,58 @@
...
+	 * @return bool
+	 */
+	private function should_exclude(): bool {
+		$auto_block = get_option( 'termly_display_auto_blocker', 'off' );
+		if ( 'on' !== $auto_block ) {
+			return false;
+		}
+
+		return true;
+	}
```
jeawhanlee: I know the grooming required us to create a method like this, but seeing it now, it doesn't play a role, so I think we can send that logic to the method below.

inc/ThirdParty/Plugins/Cookie/Termly.php
```diff
@@ -0,0 +1,58 @@
...
+
+	/**
+	 * Excludes Termly JS files from delay JS
+	 *
+	 * @param array $exclude_delay_js Array of JS to be excluded.
+	 *
+	 * @return array
+	 */
+	public function exclude_defer_js( array $exclude_delay_js ): array {
```
jeawhanlee: This method name needs to change to something more generic and we need to update the docblock to be generic as well.

jeawhanlee (changes requested):
It's good we have a unit test for this, but I think an integration test for this scenario would be helpful for testing with a html template containing that kind of link we are trying to exclude to see if the regex works properly.

Mai-Saad:
@Khadreal Thanks for the PR. using the PR , can see that termly script is not excluded from deferjs /delay js
here https://new.rocketlabsqa.ovh/about. can you please check :pray: 
Screenshot from 2024-12-23 17-43-32
Screenshot from 2024-12-23 17-46-43
For test:
1- When termly has `termly_display_auto_blocker` on in options table => we exclude `https://app.termly.io/resource-blocker/` from defer js , delay js (try defer only on , delay only on , both on , all js modifications on) {it's already on in newlabs} => confirm that no console error/warning before/after interaction with the page
2- When termly option is off `termly_display_auto_blocker` , we aren't excluding the script from delay js / defer js

Khadreal:
@Mai-Saad pushed a fix and tested that it's working now.

wordpressfan (approved):

hanna-meda (approved):
Thank you, @Khadreal.

After re-testing, I can confirm:
• We exclude https://app.termly.io/resource-blocker/ from defer and delay JS when `termly_display_auto_blocker` is enabled in the options table.
• We do not exclude https://app.termly.io/resource-blocker/ from defer and delay JS when `termly_display_auto_blocker` is disabled in the options table.

PR #7215, opened 2025-01-06
Closes #6550 Delegate creation of the dependencies to the container

### LINKED ISSUE #6550: Delegate to the container the creation from dependencies
**Describe the bug**
Currently while wiring the classes we are instantiating dependencies.

However, this is something we should not do cause it moves the instantiation from dependencies inside the service provider and not anymore inside the definition.

The main difference here being that  the definition is called when the actual is needing for instantiation and not when the whole provider needs to be registered.

**To Reproduce**
Use Xdebug
1. Go to any provider
2. Check when the dependency is instantiated
3. Removed the `$this->getContainer()->get` on the dependency
4. Check when the dependency is instantiated

**Expected behavior**
Developers should pass dependencies this way:
```php
		$this->getContainer()->add( 'slack_subscriber', Subscriber::class )
			->addArgument( 'pagespeed_results_query' );
```

instead of:
```php
		$this->getContainer()->add( 'slack_subscriber', Subscriber::class )
			->addArgument( $this->getContainer()->get( 'pagespeed_results_query' ) );
```

**Screenshots**

**Additional context**

remyperona:
# Description

Fixes #6550

Update the service providers code to pass the dependencies aliases instead of the instances of the classes when registering the services.

## Type of change

- [x] Chore

## Detailed scenario

The change doesn't have any impact on the functionality of the plugin itself.

### Documentation

Replaced the container calls to get the instances of dependencies by passing only the aliases directly. This delegates the instances creation to the container instead of it happening in the providers.

Also update string & array dependencies declaration to use the `StringArgument` and `ArrayArgument` classes, which helps the container know directly that the type of the value passed.

# Mandatory Checklist

## Code validation

## Code style


wordpressfan (approved):
For @wp-media/qa-team here we touched mostly all service providers, and this affects even thirdparty compatibility files, so good luck :D 

[Slack discussion](https://wp-media.slack.com/archives/CUT7FLHF1/p1736841239822069?thread_ts=1736838002.973639&cid=CUT7FLHF1) to explain what we changed in that PR because we will use in our next development.

Also I believe we need to merge this ASAP not to get many conflicts while we are working on features in develop.

Thanks.

Mai-Saad:
- Testplan https://wpmediaqa.testrail.io/index.php?/runs/view/952
- e2e results are almost similar to the reference with 3.18
Screenshot from 2025-01-23 09-58-39

PR #7218, opened 2025-01-08
fixes #7002 - filter rocket_lrc_processor to use regex if DOM extension is not found 

### LINKED ISSUE #7002: Automatic Lazy Rendering: 500 error in browser console if dom extension is disabled - Class "DOMDocument" not found

**Describe the bug**
A 500 error appears in the browser console, the website is still functional but cache and optimizations are not applied.

This is related to Automatic Lazy Rendering and its usage of **PHP dom extension** 
When the dom extension (https://www.php.net/en/dom) is disabled in the server, it triggers the error here: 
https://github.com/wp-media/wp-rocket/blob/73488699424f726bef28dbf2c934340efcab8e47/inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Dom.php#L81

Error log:
`5818 FastCGI sent in stderr: "PHP message: PHP Fatal error: Uncaught Error: Class "DOMDocument" not found in /wp-content/plugins/wp-rocket/inc/Engine/Optimization/LazyRenderContent/Frontend/Processor/Dom.php:81
`
To solve they need to enable the dom extension in PHP

**To Reproduce**
1. Disable DOM extension  in PHP
2. Visit the frontend of the site
3. See the 500 error console error

**Expected behavior**
We should check if DOM extension is enabled or not, and:
- show a warning to the users in the backend
- Avoid the fatal error

**Additional context**
Ticket 1: https://secure.helpscout.net/conversation/2720908346/514956?folderId=2683093
Ticket 2: https://secure.helpscout.net/conversation/2723174345/515454?folderId=2683093
Slack thread: https://wp-media.slack.com/archives/C43T1AYMQ/p1727853713650349

alfonso100:
as discussed with @wordpressfan switch `rocket_lrc_processor` to regex if PHP `dom` extension is not found

# Description

Fixes #7002

Here we check if the dom module is loaded in php so use the dom processor other than that use the regex which should do the same functionality.

## Type of change

## Detailed scenario

Disable the DOM extension, see a 500 error in the site frontend, in the console (As described in the issue).

With the new code, if DOM extension is not available it should switch rocket_lrc_processor to use regex instead, preventing the error

### Documentation

Mentioned above.

### New dependencies

No.

### Risks

In a few cases we may face slowness when using regex but I believe this will be better than the error 500.

# Mandatory Checklist

## Code validation

## Code style

Because results for dom and regex are the same so we don't need to change anything in tests, we have integration tests covering this method.

# Additional Checks

### DIFF
--- a/inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php
+++ b/inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php
@@ -147,7 +147,7 @@ class Controller implements ControllerInterface {
 		 *
 		 * @param string $processor The processor to use.
 		 */
-		$processor = wpm_apply_filters_typed( 'string', 'rocket_lrc_processor', 'dom' );
+		$processor = wpm_apply_filters_typed( 'string', 'rocket_lrc_processor', extension_loaded( 'dom' ) ? 'dom' : 'regex' );
 
 		$this->processor->set_processor( $processor );
 		$this->processor->get_processor()->set_exclusions( $this->get_exclusions() );


wordpressfan (approved):
Awesome job @alfonso100

piotrbak:
@DahmaniAdame With this approach we'll reintroduce this risk:
https://wp-media.slack.com/archives/CUKB44GNN/p1720196732934469

I believe this is acceptable 🤔

Mai-Saad (approved):
No fatal error now (only deprecated notices with php8.4 which shall be fixed on other PR. If the risk mentioned about regex is accepted then we can merge this one

DahmaniAdame:
> I believe this is acceptable 🤔

Yes. Let's go with that.

PR #7231, opened 2025-01-16
Fix #6960 Beacon script can be injected multiple times when there are multiple </body> in the page.

### LINKED ISSUE #6960: Beacon script can be injected multiple times when there are multiple </body> in the page.
**Context**

https://wp-media.slack.com/archives/CUT7FLHF1/p1725795776110169?thread_ts=1725624439.071799&cid=CUT7FLHF1

The impact is that the beacon executes multiple times, and performs multiple AJAX calls. Only the first one is useful.

**Expected behavior**

The beacon is injected at most once.

**Acceptance Criteria**
On a page with multiple </body>, the beacon is injected only once.

Khadreal:
# Description
Fix multiple </body> tag beacon script issue.

Fixes #6960 

## Type of change

## Detailed scenario
Create a template with multiple `</body>` tag and open in a browser, check the script and notice that the beacon script is run twice(inserted twice)

### Documentation
Change the beacon script insertion with `str_replace` to `preg_replace`

# Mandatory Checklist

## Code validation

## Code style

### DIFF
--- a/inc/Engine/Common/PerformanceHints/Frontend/Processor.php
+++ b/inc/Engine/Common/PerformanceHints/Frontend/Processor.php
@@ -192,7 +192,7 @@ class Processor {
 		$script_tag = "<script data-name=\"wpr-wpr-beacon\" src='{$script_url}' async></script>"; // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
 
 		// Append the script tag just before the closing body tag.
-		return str_replace( '</body>', $inline_script . $script_tag . '</body>', $html );
+		return preg_replace( '/<\/body>/', $inline_script . $script_tag . '</body>', $html, 1 );
 	}
 
 	/**
--- a/inc/Engine/Common/PerformanceHints/Frontend/Processor.php
+++ b/inc/Engine/Common/PerformanceHints/Frontend/Processor.php
@@ -189,13 +189,10 @@ class Processor {
 		$script_url = rocket_get_constant( 'WP_ROCKET_ASSETS_JS_URL' ) . 'wpr-beacon' . $min . '.js';
 
 		// Create the script tag.
-		$script_tag = "<script data-name=\"wpr-wpr-beacon\" src='{$script_url}' async></script>"; // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
-		$body_tag   = strrpos( $html, '</body>' );
-
-		if ( false !== $body_tag ) {
-			// Append the script tag just before the last closing body tag especially in cases where there's an iframe.
-			$html = substr_replace( $html, $inline_script . $script_tag . '</body>', $body_tag, 7 );
-		}
+		$script_tag             = "<script data-name=\"wpr-wpr-beacon\" src='{$script_url}' async></script>"; // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+		$last_body_tag_position = strrpos( $html, '</body>' );
+		// Append the script tag just before the last closing body tag especially in cases where there's an iframe.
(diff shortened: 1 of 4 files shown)


inc/Engine/Common/PerformanceHints/Frontend/Processor.php
```diff
@@ -192,7 +192,7 @@ private function inject_beacon( $html, $url, $is_mobile ): string {
 		$script_tag = "<script data-name=\"wpr-wpr-beacon\" src='{$script_url}' async></script>"; // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
 
 		// Append the script tag just before the closing body tag.
-		return str_replace( '</body>', $inline_script . $script_tag . '</body>', $html );
+		return preg_replace( '/<\/body>/', $inline_script . $script_tag . '</body>', $html, 1 );
```
jeawhanlee: Good Job Sir!

While this fixes the issue, it might mean that we would be injecting the beacon in the wrong body. From what I got [here](https://new.rocketlabsqa.ovh/lrc_similive/) where the issue was discovered, we had an iframe in the body, so with this implementation, we would be injecting the beacon in the body of the iframe and not the parent doc body.

I'd suggest we only inject the beacon just before the last closing body tag which generally would be the parent doc. To achieve this, we could use a pattern like this: `"#</body>(?!.*</body>)#` to match the last occurrence of `</body>` in the doc

tests/Fixtures/inc/Engine/Common/PerformanceHints/Frontend/Subscriber/HTML/double_body_tag.html
jeawhanlee, line 1: I think we should update this template to be like the original one the issue was generated by introducing an iframe instead of having to standalone body tags.

jeawhanlee (changes requested):

inc/Engine/Common/PerformanceHints/Frontend/Processor.php
Khadreal: I think iframe seems to be a different issue, the PR is based on this thread https://wp-media.slack.com/archives/CUT7FLHF1/p1725795776110169?thread_ts=1725624439.071799&cid=CUT7FLHF1
Using template https://new.rocketlabsqa.ovh/3279_template/) like this. Will try to modify for the other edge case too

jeawhanlee: Yes, the main message that started the thread had this template: https://new.rocketlabsqa.ovh/lrc_similive/ in it. besides, https://new.rocketlabsqa.ovh/3279_template/ does not seem to have 2 closing body tag or am I missing something?

inc/Engine/Common/PerformanceHints/Frontend/Processor.php
```diff
@@ -190,9 +190,14 @@ private function inject_beacon( $html, $url, $is_mobile ): string {
 
 		// Create the script tag.
 		$script_tag = "<script data-name=\"wpr-wpr-beacon\" src='{$script_url}' async></script>"; // phpcs:ignore WordPress.WP.EnqueuedResources.NonEnqueuedScript
+		$body_tag   = strrpos( $html, '</body>' );
 
-		// Append the script tag just before the closing body tag.
-		return preg_replace( '/<\/body>/', $inline_script . $script_tag . '</body>', $html, 1 );
+		if ( false !== $body_tag ) {
```
jeawhanlee: I don't think this condition is necessary since there will always be a body tag.

jeawhanlee (approved):

Mai-Saad:
Test plan https://wpmediaqa.testrail.io/index.php?/runs/view/951

Miraeld:
Everything green ! Great!
[testrail-report-683.pdf](https://github.com/user-attachments/files/18520537/testrail-report-683.pdf)

PR #7236, opened 2025-01-22
Make sure that we cache only valid posts/pages urls

wordpressfan:
# Description

Fixes https://github.com/wp-media/wp-rocket.me/issues/4456

This PR will make sure that we cache only

## Type of change

## Detailed scenario

For posts, pages, and even custom posts in some cases when we add any string inside the url, we still cache those pages because they don't return 404 error, here we make sure that those urls are not cached at all.

More details are inside the issue itself.

### Documentation

We compare the current page/post urls with their correct urls and stop caching for those not valid ones.

### New dependencies

No

### Risks

We have this code exactly like what we handled taxonomies so I hope we don't have a false cases that we detect the valid urls correctly.

# Mandatory Checklist

## Code validation

## Code style

Checking tests now.

# Additional Checks


Miraeld (approved):

hanna-meda (approved):
Exploratory and Test Plan execution went well. 
[testrail-report-696.pdf](https://github.com/user-attachments/files/18672760/testrail-report-696.pdf)

wordpressfan:
I think we need to wait a bit before merging this PR because we got some issues with taxonomy pages PR and the solution here is exactly the same.

Those issues are:
1. When using custom permalink that mentioned as `second issue` [In this PR description](https://github.com/wp-media/wp-rocket/pull/7298)
2. When using custom rewrite rules as mentioned in the last dev escalation with an example [here](https://wp-media.slack.com/archives/C08N8J6VC/p1739355385553839?thread_ts=1739282994.329729&cid=C08N8J6VC)

but now this will happen with posts so I'd suggest having a filter to enable/disable this feature when needed (that's a product decision)

piotrbak:
CC @MathieuLamiot 

In general, we'll need a filter here and some more testing as we know about the new limitations

MathieuLamiot:
Definitely, let's go with adding a filter around this whole functionality once 3.19-prealpha is released internally 👌

remyperona:
@wordpressfan I did a refactor and added the filter if you want to have a look

inc/Engine/Cache/UrlValidation/PostSubscriber.php
```diff
@@ -0,0 +1,49 @@
...
+class PostSubscriber extends AbstractUrlValidation implements Subscriber_Interface {
+	/**
+	 * {@inheritdoc}
+	 */
+	public static function get_subscribed_events() {
+		return [
+			'do_rocket_generate_caching_files' => 'disable_cache_on_not_valid_url',
+			'rocket_buffer'                    => [ 'stop_optimizations_for_not_valid_url', 1 ],
+		];
+	}
```
wordpressfan, line 18: The subscribed events part is common between all urlvalidation classes, do u think it'd be better having it in the abstract class? this will save some code :D

inc/Engine/Cache/UrlValidation/AbstractUrlValidation.php
```diff
@@ -0,0 +1,57 @@
...
+		return $this->is_not_valid_url() ? '' : $html;
+	}
+
+	/**
+	 * Check if url validation is disabled by filter
+	 *
+	 * @return bool
+	 */
+	protected function is_disabled(): bool {
+		return wpm_apply_filters_typed( 'boolean', 'rocket_disable_url_validation', false );
```
wordpressfan, line 53: do we need to add the docblock here?

wordpressfan (commented):
I can't approve my own PR, but seems valid from my side, just two questions.
Thanks Boss.

inc/Engine/Cache/UrlValidation/PostSubscriber.php
remyperona: I prefer to keep them separated in case there is future differences in the subscribers. It also keeps the code easier to read since the subscribers implement the interface (which requires this method), so not having it in the same class could be confusing

Miraeld:
Everything green, 
Report [here](https://github.com/user-attachments/files/20560532/testrail-report-727.pdf)

Mai-Saad:
https://group-onecom.slack.com/archives/C08F4LZB2QG/p1748975372073809
moved to 3.19.1

PR #7277, opened 2025-02-06
Clear CPT custom cache files correctly

### LINKED ISSUE #4508: Partial purge does not clear CPT archive when non-default cache files are used
- Made sure you’re on the latest version => 3.10.3

**Describe the bug**
Some Cached pages aren't cleared when editing post if Cookie GDPR is enabled

**To Reproduce**
Precondition:
- Cookie Notice & Compliance for GDPR / CCPA is enabled
- WooCommerce is enabled, at least one product is present, shop page is set

1. Visit the following URLs:
https://rocketlabsqa.ovh/product/wordpress-pennant/
https://rocketlabsqa.ovh/shop/
2. Accept cookie and refresh both pages
3. Edit the post and update:
https://rocketlabsqa.ovh/wp-admin/post.php?post=1000&action=edit
4. Cached files for the shop are not deleted

**Expected behavior**
All cached files are deleted

**Screenshots**
cache

**Additional context**
- If Cookie Notice & Compliance for GDPR / CCPA is deactivated, after step 3, all cached files will be deleted
- As per @piotrbak investigations
We're clearing only the .html  and .html_gzip ending files of the Archive page
When there's a dynamic cookie we will not clear the cache of specific archive pages
The problem is here: https://github.com/wp-media/wp-rocket/blob/9ba1546480baac14cd704622f8a575b47c0783a9/inc/common/purge.php#L72-L73
purge.php
It also seems that the pagination clearing is broken there, it tries to clear this https://rocketlabsqa.ovh/shop/index-httpspage instead of this https://rocketlabsqa.ovh/shop/page/2/index-https.html

- [ ] Reproduce the problem
- [ ] Identify the root cause
- [ ] Scope a solution
- [ ] Estimate the effort

wordpressfan:
# Description

Fixes #4508

This PR makes sure that we remove all custom cache files for custom post types are deleted properly.

## Type of change

## Detailed scenario

When we clear custom post type post (with any clearance action like editing the post), we should clear the archive page (shop page for woocommerce product CPT as an example), but when we have webp enabled (or any feature that creates custom cache file other than index.html or index.html_gzip) those custom cache files are kept there without being removed.

**Steps to reproduce:**
1. Install/activate woocommerce
2. Change the permalink structure for produce to be like:
image
3. Make sure that some products' pages along with shop page are cached.
4. Make sure that preload is disabled.
5. Update one post.
6. The pages that should be cleared are as follows:-
  1. Category page
  2. Edited Product page
  3. Previous, Next, Previous in the same category, and next in the same category.
  4. Shop page
7. Other than that all pages below shop folder shouldn't be touched.

### Documentation

Here we added the support of regex inside filenames to target all html files from the selected directory, this will make sure that we will remove the files from the directory without affecting any other directories inside this one.

### New dependencies

N/A

### Risks

we should be good but as we remove with regex we need to test that to make sure that we only remove what is needed to be removed.

# Mandatory Checklist

## Code validation

## Code style

All good.

# Additional Checks


Khadreal (approved):

Mai-Saad:
@wordpressfan Thanks for the PR. please note that we have the following regression
1- product permalink is shop base
2- cache shop , product 1, product 2
3- edit product 1 or clear cache of product 1 from page edit
4- all shop folder is deleted including product 2 cache ==> on trunk product 2 isnot touched

remyperona:
If you do it that way, you're going to delete cache of any child item in the permalink structure, that's why the deletion is not using the base but the filenames.

You need a different approach to handle the different cache files variations. Possibly with RegEx if it's usable (I don't remember right now), you could do something like index(.*).html.

Khadreal (approved):

Mai-Saad:
Working as expected with the following notes (already on the trunk)
- With WCML?WPML, the cache of the translated page of the shop or product won't be deleted when editing the product in the main language
- Changing WPML language setup won't clear cache, can be enhancement based on old discussion ~2022
- Edit product  permalink is not clearing the cache
- Edit product, will add this URL to cache table but will be removed once processed domain/shop/page
[testrail-report-705.pdf](https://github.com/user-attachments/files/18823406/testrail-report-705.pdf)

PR #7304, opened 2025-02-12
Prepare for rocket head ordering

### LINKED ISSUE #7272: 3.19 - Refactor features to use `rocket_head` new approach
We started a discussion [here](https://wp-media.slack.com/archives/CUT7FLHF1/p1738172791509009?thread_ts=1737472334.599049&cid=CUT7FLHF1) with @remyperona to have an approach to unify the addition of `link` tags inside head directly after the closing tag of `title` but without any duplication, and we proposed a solution [here](https://github.com/wp-media/wp-rocket/compare/develop...feature/add-rocket-head)

So we need to move forward with this approach (we may need to unify the way we insert link into the head using a trait or functions, this will make it more maintainable)

So the main goals of this issue are:
1. Set the standard of the items being added into head, something like `Priority 10 for link tags, priority 20 for styles, ...etc.`
2. Create the trait to insert the items into head, so instead of using the following code:
```
add_filter( 'rocket_head_items', function ( $items ) {
    $items[] = [
        'open_tag' => '<link',
        'rel' => 'preload',
        'href' => 'https://example.org/',
        'close_tag' => '/>',
    ];
    return $items;
} );
```
we can use something like:

```
add_filter( 'rocket_head_items', function ( $items ) {
    $items[] = $this->add_preload_link( 'https://example.org/' );
    return $items;
} );
```
and do the same with other types of link types and styles.
3. Refactor the other features like RUCSS, critical CSS, and any other feature that add items into the head (you can search for the closing title tag into the code to reach all instances) and then make sure that the features are working properly.

wordpressfan:
# Description

Fixes #7272

Every thing is mentioned in the issue itself.

## Type of change

## Detailed scenario

Here we refactor how our features add items into the head directly after title closing tag `</title>`.

### Documentation

We introduced here a filter to print anything needed in the head part and its usage is as follows:

```
add_filter( 'rocket_head_items', function( $items ) {
    return $items;
} );
```

and this `$items` is an array of array variable with the following structure:
```
$items = [
    'open_tag' => '<style',
    'close_tag' => '</style>',
    'inner_content' => 'h1{color: red;}',
    'id' => 'any_id',
..... any html attribute as key => value
];
```

and if u need to add HTML attribute without value, something like `crossorigin`, you can add its key as a numeric, something like:
```
1 => 'crosorigin',
```

Also we introduced a helper trait called [ElementTrait](https://github.com/wp-media/wp-rocket/blob/2cfd353a311e764d2be6448d6ec8b35c09b9a42a/inc/Engine/Common/Head/ElementTrait.php#L8) to help in inserting the common head elements like preload, preconnect, style, noscript, ...etc.

Also as a standard, we will use the following filter priorities:
```
Priority 10: preconnect
Priority 30: preload
Priority 50: styles, noscripts, scripts, ...etc.
```

### New dependencies

No

### Risks

We validated the e2e and it works, also most of tests passed, just had an issue with one test that it fails for the first time, that we need to revisit later on.

# Mandatory Checklist

## Code validation

## Code style

*If some mandatory items are not relevant, explain why in this section.*

# Additional Checks


wordpressfan:
created a followup issue here:
https://github.com/wp-media/wp-rocket/issues/7315
not to forget to fix the skipped tests.

Issue #7281, opened 2025-02-06
3.19 - Preload Fonts - Current feature deprecation


piotrbak:
#### User Story
As a user, I’d like not to see the old feature in the UI

Acceptance Criteria
- The option is removed from preload tab for all users ([epic](https://www.notion.so/152ed22a22f08006baf2d370eb570a6e?pvs=21))
- The **“, preload fonts”** text is removed from the Preload description ([5](https://www.notion.so/Preload-fonts-126ed22a22f0808ba45ad3673c2c176b?pvs=21))

wordpressfan:
I think we handled it [here](https://github.com/wp-media/wp-rocket/pull/7303) or most of it, @Miraeld what do u think?

Miraeld:
I haven't removed the old UI, but it could be done easily: 

### Scope a solution

- Remove [this part](https://github.com/wp-media/wp-rocket/blob/09142460de757366d571eaf643588bea73669aab/inc/Engine/Admin/Settings/Page.php#L1117-L1127)
- Remove [this part](https://github.com/wp-media/wp-rocket/blob/09142460de757366d571eaf643588bea73669aab/inc/Engine/Admin/Settings/Page.php#L1170-L1180)
- Remove [this part](https://github.com/wp-media/wp-rocket/blob/73488699424f726bef28dbf2c934340efcab8e47/inc/Engine/Admin/Settings/Settings.php#L345-L346)
- Maybe, if necessary, deprecates (using `rocket_apply_filter_and_deprecated`) this filter: https://github.com/wp-media/wp-rocket/blob/a1bdab03c889b965cc55e7f8eb4d1cbaff5e83ff/inc/Engine/Preload/Fonts.php#L99

Mai-Saad:
AC here is ok (tested on commit 90eb7b0ea181132812cd6c5070c913eaff26d859)
Note: couldn't find story about `As an old user, I’d like to see preload fonts enabled by default if it was used before`

hanna-meda:
> Note: couldn't find story about As an old user, I’d like to see preload fonts enabled by default if it was used before

Hi, @Mai-Saad. It is covered in this one #7280

PR #7319, opened 2025-02-21
Closes #7280: Preload Fonts - Enable feature automatically for specific cases

### LINKED ISSUE #7280: 3.19 - Preload Fonts - Enable feature automatically for specific cases
#### User Story
As an existing user, I’d like to see preload fonts enabled by default if it was used before

Acceptance Criteria
- Update while preload fonts text area is empty
    - New preload fonts is not checked in UI
- Update while preload fonts text area have value
    - New preload fonts is checked

Miraeld:
# Description

Fixes #7280 
It will enable automatically if the user is using the old Preload Fonts feature when upgrading to 3.19.

## Type of change

### What was tested
Manual test of this feature.

### How to test

When upgrading to 3.19, if the old preload fonts feature was enabled, it should enabled the new one by default.

### Documentation

This pull request includes changes to the `inc/Engine/Admin/Settings/Subscriber.php` file to improve the handling of font preloading during version upgrades. The most important changes include modifying the `maybe_enable_auto_preload_fonts` method to accept version parameters and adding a version comparison check.

Enhancements to version handling:

* [`inc/Engine/Admin/Settings/Subscriber.php`](diffhunk://#diff-fc05e3a2422a5be1348634dbe69358325e7665e59b8d6beab9439e453edd5676L300-R304): Modified the `maybe_enable_auto_preload_fonts` method to accept `$new_version` and `$old_version` parameters, and added a version comparison check to ensure the method only runs for versions older than 3.19.
* [`inc/Engine/Admin/Settings/Subscriber.php`](diffhunk://#diff-fc05e3a2422a5be1348634dbe69358325e7665e59b8d6beab9439e453edd5676L71-R71): Updated the `get_subscribed_events` method to include the version parameters in the `maybe_enable_auto_preload_fonts` callback.

### New dependencies

None

### Risks

None

# Mandatory Checklist

## Code validation

## Code style


inc/Engine/Admin/Settings/Subscriber.php
```diff
@@ -295,11 +295,16 @@ public function enable_separate_cache_files_mobile( $new_version, $old_version )
...
+	 *
 	 * @return void
 	 */
-	public function maybe_enable_auto_preload_fonts(): void {
-		$this->page->maybe_enable_auto_preload_fonts();
-	}
+	public function maybe_enable_auto_preload_fonts( $new_version, $old_version ): void {
+		if ( version_compare( $old_version, '3.19', '>' ) ) {
+			return;
+		}
```
wordpressfan: This part means that we will bailout when the old version is more than 3.19 and I don't think this is the case here.

I think we need to negate the condition.

inc/Engine/Admin/Settings/Subscriber.php
```diff
@@ -295,11 +295,16 @@ public function enable_separate_cache_files_mobile( $new_version, $old_version )
...
+		$this->page->maybe_enable_auto_preload_fonts(); }
```
wordpressfan: Check this method implementation, you will find that it doesn't work because of two reasons:

1. Option name is wrong [here](https://github.com/wp-media/wp-rocket/blob/30b6da938b769714351ad98e1f71e9887633ee0b/inc/Engine/Admin/Settings/Page.php#L2297) it should be `preload_fonts`
2. We are setting the setting [here](https://github.com/wp-media/wp-rocket/blob/30b6da938b769714351ad98e1f71e9887633ee0b/inc/Engine/Admin/Settings/Page.php#L2299) but we didn't update the database itself using `options_api` or even `update_options`.

Also I think we need to move this part from `page` class to the preload_fonts admin subscriber (maybe we need to create one) to make it organized.

wordpressfan (changes requested):
A few comments to make it works.
Thanks

inc/Engine/Admin/Settings/Subscriber.php
Miraeld: I've taken into consideration your feedback. 

I've moved away the changes, 
And I've updated what you mentioned.

inc/Engine/Media/PreloadFonts/Admin/Settings.php
```diff
@@ -0,0 +1,49 @@
...
+	 * This function checks the value of the `rocket_preload_fonts` option.
+	 * If it contains a non-empty value, it updates the `auto_preload_fonts` option to `true`.
+	 * This is useful for ensuring that automatic font preloading is enabled based on legacy settings.
+	 *
+	 * @return void
+	 */
+	public function maybe_enable_auto_preload_fonts(): void {
+		if ( ! empty( $this->options->get( 'preload_fonts', [] ) ) ) {
+			$this->options->set( 'auto_preload_fonts', true );
+			$this->options_api->set( 'auto_preload_fonts', true );
```
remyperona: That doesn't seem right. Shouldn't it be the rocket options updated?

inc/Engine/Media/PreloadFonts/Admin/Settings.php
```diff
@@ -0,0 +1,49 @@
...
+	 * Enables the auto preload fonts option if the old preload fonts option is not empty.
+	 *
+	 * This function checks the value of the `rocket_preload_fonts` option.
+	 * If it contains a non-empty value, it updates the `auto_preload_fonts` option to `true`.
+	 * This is useful for ensuring that automatic font preloading is enabled based on legacy settings.
+	 *
+	 * @return void
+	 */
+	public function maybe_enable_auto_preload_fonts(): void {
+		if ( ! empty( $this->options->get( 'preload_fonts', [] ) ) ) {
```
remyperona: Maybe invert the condition to do an early bail-out if the option is empty

inc/Engine/Media/PreloadFonts/ServiceProvider.php
```diff
@@ -90,5 +95,19 @@ public function register(): void {
 					$this->getContainer()->get( 'preload_fonts_context' ),
 				]
 			);
+
+		$this->getContainer()->addShared( 'preload_fonts_admin_settings', AdminSettings::class )
+			->addArguments(
+				[
+					$this->getContainer()->get( 'options' ),
+					$this->getContainer()->get( 'options_api' ),
```
remyperona: Should use the alias directly instead of using `get()`

remyperona (changes requested):

remyperona (approved):

hanna-meda:
Thanks, @Miraeld, for this PR.

After merging with the latest `feature/3.19` branch (which includes the current feature deprecation PR #7281), I tested the update and found the new feature isn’t auto-enabled when Preload fonts values exist from the previous version.
Tested zip: [wp-rocket-PR7319.zip](https://github.com/user-attachments/files/20105461/wp-rocket-PR7319.zip)
See the video below for details.

https://github.com/user-attachments/assets/9766da4c-94e9-4d7e-8360-86ece5136423

Miraeld:
Hey @hanna-meda I think we had a problem with the way we were getting the old value of `preload_fonts`, which resulted into a bail-out. 

With latest commits it should be fine now :)
At least it is on my end.

Miraeld:
Testrail report is green :) 

[testrail-report-717.pdf](https://github.com/user-attachments/files/20179602/testrail-report-717.pdf)

PR #7380, opened 2025-05-06
Closes #7283: 3.19 - Preload Fonts - Exclusion filter

### LINKED ISSUE #7283: 3.19 - Preload Fonts - Exclusion filter
#### User Story
As an admin, I want a filter to exclude fonts from preload fonts

Acceptance Criteria
- System fonts are excluded by default ([details](https://www.notion.so/Preload-fonts-126ed22a22f0808ba45ad3673c2c176b?pvs=21))
- Filter is guarded against unexpected values
- Exclusion can happen using URL of the font
- We need backend for this part

jeawhanlee:
# Description

Fixes #7283 

## Type of change

### What was tested

- Exclusions in filter is respected and excluded fonts(either via name or url) is not saved in db for backend or response of excluded url not received from js.

### How to test

Use filter `rocket_preload_fonts_excluded_fonts` to return fonts with either font family name or url and see that excluded fonts are respected.

### Documentation

Added new method to the context class to enable a centralized management of default & backend exclusions.

### New dependencies

Dynamic Lists

### Risks

N/A

# Mandatory Checklist

## Code validation

## Code style

Not Applicable to the changes in this PR.

# Additional Checks


inc/Engine/Media/PreloadFonts/AJAX/Controller.php
```diff
@@ -22,17 +22,17 @@ class Controller implements ControllerInterface {
 	/**
 	 * PreloadFonts Context.
 	 *
-	 * @var ContextInterface
+	 * @var Context
```
wordpressfan, line 25: Why do we change the interface here to be a class? I think it's more generic to have the interface and this won't affect anything in the code except the IDE code completion :D

inc/Engine/Media/PreloadFonts/AJAX/Controller.php
```diff
@@ -72,6 +72,9 @@ public function add_data(): array {
 			$max_preload_fonts_number = 1;
 		}
 
+		$exclusions = $this->context->get_exclusions();
```
wordpressfan: Why do we get the exclusions list here and pass if to `remove_excluded_fonts` method?
I think we can call it directly there, what do u think?

inc/Engine/Media/PreloadFonts/AJAX/Controller.php
```diff
@@ -143,4 +146,36 @@ public function check_data(): array {
...
+
+	/**
+	 * Removes excluded fonts from the list of fonts to be preloaded.
+	 *
+	 * @param array $fonts Array of fonts to be preloaded.
+	 * @param array $exclusions Array of fonts to be excluded.
+	 *
+	 * @return array Filtered array of fonts, excluding those specified in the exclusion list.
+	 */
+	public function remove_excluded_fonts( array $fonts, array $exclusions ): array {
```
wordpressfan: Why this is public?

inc/Engine/Media/PreloadFonts/AJAX/Controller.php
```diff
@@ -143,4 +146,36 @@ public function check_data(): array {
...
+					if ( stripos( $font, $exclusion ) !== false ) {
+						return false;
+					}
+				}
+
+				return true;
+			}
+		);
+
+		return array_values( $filtered_fonts );
```
wordpressfan: The time complexity is approximately O(n × m), where:
- n = number of $fonts
- m = number of $exclusions

I think we need to move this to be regex so we will combine all regex elements into one regex (imploded by `|`) and compare it with each font in the loop, something like:

```suggestion
	// Escape special regex characters in exclusions and implode them into a regex pattern
    $escaped_exclusions = array_map(function ($exclusion) {
        return preg_quote($exclusion, '/');
    }, $exclusions);

    // Combine all exclusions into one regex pattern, separated by `|`
    $regex_pattern = '/(' . implode('|', $escaped_exclusions) . ')/i';

    // Filter fonts by checking against the regex pattern
    $filtered_fonts = array_filter(
        $fonts,
        function ($font) use ($regex_pattern) {
            return !preg_match($regex_pattern, $font);
        }
    );

    // Reset array keys and return the filtered fonts
    return array_values($filtered_fonts);
```

This will change the time complexity to be O(n), where n is the number of $fonts.
What do u think?

inc/Engine/Media/PreloadFonts/Context/Context.php
```diff
@@ -47,4 +90,22 @@ public function is_mobile_allowed(): bool {
...
 	}
+
+	/**
+	 * Get array of fonts to be excluded.
+	 *
+	 * @return array
+	 */
+	public function get_exclusions(): array {
+		$lists      = $this->data_manager->get_lists();
+		$lists      = isset( $lists->delay_js_exclusions ) ? $lists->delay_js_exclusions : [];
```
wordpressfan:
```suggestion
		$lists      = $lists->delay_js_exclusions ?? [];
```

I believe we need to change the `delay_js_exclusions` to be the one mentioned in rucss-backend PR, correct?

inc/Engine/Media/PreloadFonts/Context/Context.php
```diff
@@ -47,4 +90,22 @@ public function is_mobile_allowed(): bool {
...
+		$exclusions = array_merge( $this->exclusions, $lists );
```
wordpressfan: Do we need to add more validation here? something like:
```suggestion
		// Validate that delay_js_exclusions is an array.
    if (!is_array($delay_js_exclusions)) {
        $delay_js_exclusions = [];
    }

    // Merge exclusions only if additional exclusions exist.
    $exclusions = empty($delay_js_exclusions)
        ? $this->exclusions
        : array_merge($this->exclusions, $delay_js_exclusions);
```
to use array_merge only when needed, what do u think?

inc/Engine/Media/PreloadFonts/Context/Context.php
```diff
@@ -47,4 +90,22 @@ public function is_mobile_allowed(): bool {
...
+
+		/**
+		 * Filters excluded fonts.
+
+		 * @param array $exclusions Array of fonts to exclude.
+		 */
+		return wpm_apply_filters_typed( 'array', 'rocket_preload_fonts_excluded_fonts', $exclusions );
```
wordpressfan:
```suggestion
		return wpm_apply_filters_typed( 'string[]', 'rocket_preload_fonts_excluded_fonts', $exclusions );
```
this would be better?

wordpressfan (changes requested):

inc/Engine/Media/PreloadFonts/AJAX/Controller.php
jeawhanlee: PHPStan

inc/Engine/Media/PreloadFonts/Context/Context.php
jeawhanlee: ~~I don't think `string[]` is a supported type for `wpm_apply_filters_typed`~~ 
Actually it is. my bad.

Miraeld (approved):

inc/Engine/Media/PreloadFonts/Context/Context.php
```diff
@@ -47,4 +90,29 @@ public function is_mobile_allowed(): bool {
...
+		$lists = $this->data_manager->get_lists();
+		$lists = isset( $lists->preload_fonts_exclusions ) ? $lists->preload_fonts_exclusions : [];
+
+		// Check that lists is a valid array.
+		if ( ! is_array( $lists ) ) {
+			$lists = [];
+		}
+
+		// Avoid unnecessary merge if lists is an empty array.
+		$exclusions = empty( $lists ) ? $this->exclusions : array_merge( $this->exclusions, $lists );
```
wordpressfan: So here we are checking if the dynamic lists for this exclusions are empty (from backend), we will fallback to the system fonts array above, correct?
I don't think this is a case because we will have the list in the json file with the release so I can't see a use case for those system fonts here, what do u think of moving those system fonts to be in the backend itself.
Thanks @Mai-Saad for this idea that I really like.

inc/Engine/Media/PreloadFonts/Context/Context.php
```diff
@@ -47,4 +90,29 @@ public function is_mobile_allowed(): bool {
...
 			&& wp_is_mobile();
 	}
+
+	/**
+	 * Get array of fonts to be excluded.
+	 *
+	 * @return array
+	 */
+	public function get_exclusions(): array {
+		$lists = $this->data_manager->get_lists();
```
wordpressfan: We are in the context, so I don't think this is a good place to get the data from the dynamic lists, I was thinking of this to only have the filter and then we can have a callback in the frontend subscriber to insert the data from the dynamic lists, this will make things cleaner, what do u think?

wordpressfan (changes requested):
Minor comments/questions

inc/Engine/Media/PreloadFonts/Context/Context.php
jeawhanlee: We can do that as well.

jeawhanlee: To my best knowledge, product doesn't have any exclusion to be added to the backend for the release, that's why the emptiness could be a scenario but if you are proposing moving the system fonts to the backend, then, the list will never be empty and so there won't be a need for checking if it's empty.

inc/Engine/Media/PreloadFonts/Context/Context.php
jeawhanlee: I thought of this as well, We can have a separate exclusion class for this or better still make this a trait because we also need it in Ajax controller, remember we also check this exclusion there before we save it in DB

wordpressfan: No need for that if u followed what I mentioned,
What I mean is to keep the exclusions inside the context but only has the filter without the dynamic lists part, and add a callback for this new filter in the frontend subscriber to get the data from the dynamic lists and add those exclusions to the filter, so when using the filter anywhere this will return the same exclusions list, got it?

inc/Engine/Media/PreloadFonts/Context/Context.php
wordpressfan: We can call @wp-media/product , But I believe we can do that safely.

inc/Engine/Media/PreloadFonts/Context/Context.php
jeawhanlee: Ah ok, I see your point, I thought you wanted to migrate the filter to the frontend.

inc/Engine/Media/PreloadFonts/Frontend/Subscriber.php
```diff
@@ -46,4 +57,34 @@ public static function get_subscribed_events(): array {
...
+		}
+
+		// Return early if lists is empty.
+		if ( empty( $lists ) ) {
+			return $exclusions;
+		}
+
+		// Only merge if exclusion and list is not empty.
+		return array_merge( $exclusions, $lists );
+	}
```
wordpressfan:
```suggestion
	public function get_exclusions( array $exclusions ): array {
		$lists = $this->data_manager->get_lists()->preload_fonts_exclusions ?? [];

		// Merge exclusions and lists (handles empty arrays gracefully).
		return array_merge( $exclusions, (array) $lists );
	}
```

What do u think about simplifying it like this?

wordpressfan (approved):
Approved with minor change.

wordpressfan:
For @wp-media/qa-team 

Now preload fonts exclusions in backend is deployed and yesterday I added the system fonts in production list, so u can test using this.

inc/Engine/Media/PreloadFonts/Frontend/Subscriber.php
jeawhanlee: Yes, I thought about this, but it was more of the micro-optimization for me to avoid unnecessary call to merge.

wordpressfan: I'm OK with both, I proposed this not for optimization but for readability, I consulted copilot now and he thinks that it's better to remove those checks.
Again I'm OK with both solutions just trying to make it easier to maintain this part in the future, but if u see that the new code is not handling all cases plz keep it as it is.

hanna-meda (approved):
Thank you, @jeawhanlee, for this PR.

Everything is working as expected. I tested both exclusion by filter and system fonts added to the BE (production SaaS). Just one finding, already known and being discussed in [this thread](https://group-onecom.slack.com/archives/C08F4LZB2QG/p1747122613624299?thread_ts=1747120988.863989&cid=C08F4LZB2QG), where I also added [my note](https://group-onecom.slack.com/archives/C08F4LZB2QG/p1747238587597189?thread_ts=1747120988.863989&cid=C08F4LZB2QG).

[testrail-report-720.pdf](https://github.com/user-attachments/files/20211002/testrail-report-720.pdf)

Issue #7378, opened 2025-05-06
3.19 Should preload fonts using @import


hanna-meda:
**Describe the bug**
Fonts using @import are not preloaded.

**To Reproduce**
1. Install and activate WPR 3.19.
2. Enable Preload Fonts & Self-host Google Fonts.
3. Visit page template using GF above the fold via @import ( i.e. https://rocketlabsqa.ovh/preloadfonts_import/, preloadfonts_import.php template).
4. Clear the cache and revisit the page.

**Expected behavior**
Above the fold fonts using @import should be preloaded, too.

Note: Issue related to #7241 .

**Screenshots**

Miraeld:
This issue has nothing to do with WP Rocket, but [rocket scripts](https://github.com/wp-media/rocket-scripts).

### Scope a solution
In [here](https://github.com/wp-media/rocket-scripts/blob/83abe1847aeb7cba382efb1f9a41255eef36933d/src/BeaconPreloadFonts.js#L127), we could edit the function with something like:
``` javascript
getFontFaceRules() {
        const stylesheetFonts = {};

        // Helper to recursively process a stylesheet and its imports
        const processSheet = (sheet, seenSheets = new Set()) => {
            if (!sheet || seenSheets.has(sheet)) return;
            seenSheets.add(sheet);
            let rules;
            try {
                rules = sheet.cssRules || [];
            } catch (e) {
                // CORS or inaccessible stylesheet
                this.logger && this.logger.logMessage(e);
                return;
            }
            for (const rule of rules) {
                // Recursively process @import rules
                if (rule.type === CSSRule.IMPORT_RULE && rule.styleSheet) {
                    processSheet(rule.styleSheet, seenSheets);
                } else if (rule instanceof CSSFontFaceRule) {
                    const src = rule.style.getPropertyValue('src');
                    const fontFamily = rule.style.getPropertyValue('font-family')
                        .replace(/['"]+/g, '')
                        .trim();
                    const weight = rule.style.getPropertyValue('font-weight') || '400';
                    const style = rule.style.getPropertyValue('font-style') || 'normal';

                    if (!stylesheetFonts[fontFamily]) {
                        stylesheetFonts[fontFamily] = {
                            urls: [],
                            variations: new Set()
                        };
                    }

                    const urls = src.match(/url\(['"]?([^'"]+)['"]?\)/g) || [];
                    urls.forEach((urlMatch) => {
[...]

PR #7401, opened 2025-05-20
fix: update wp-rocket-scripts dependency to resolve CORS issue

### LINKED ISSUE #7306: 3.19 - Preload Fonts - Beacon processing Pilot - Parse stylesheets and extract font properties
It has been observed that when loading external fonts on chrome, the font url is not exposed with the Performance API, This has lead us to consider various options to bypass this.

- Fetch the stylesheet via JS and parse it to extract all the font rules
- Print font CSS on page to expose font urls
- Fetch the stylesheet via PHP and parse it to extract all the font rules then send this to the beacon.

However, the first 2 options are not without cons for the first, we'll need to consider another overhead don by fetch API to get the CSS and also CORS issue. The Second might lead to increased page size in terms of large font stylesheet or several font styles used on a page.

We settled for the third because we'll parse the stylesheet in a single request without creating an overhead and having to print it to the page, Then match all the font properties just like it was done [here](https://gist.github.com/DahmaniAdame/bf6db29899c241517aca0bfd5675410b#file-gistfile1-txt-L107-L131) and pass the return value containing an array of fonts and their properties as a custom data to the beacon script to process.

Miraeld:
This PR doesn't have an issue linked. 

It is here to be able for QA to test https://github.com/wp-media/rocket-scripts/pull/51


wordpressfan (approved):

Mai-Saad:
Thanks for the update.
Notes:
- Already on feature/3.19 branch commit 803a427a48f92d9e8c42e6acf33d0b7b75f4e6e1, if self hosted GF is on , we add `<link href='https://fonts.gstatic.com/' crossorigin rel='preconnect' />` (regression compared to 3.18.3) here https://new.rocketlabsqa.ovh/preloadfonts_10atf_10btf_custom_diff
- With fresh install, only home id added to DB as prewarmup
- Manual visiting https://new.rocketlabsqa.ovh/preloadfonts_10atf_10btf_custom_diff from my device, URL is not added to priority elements table (beacon is injected in the page) => if we opened new incognito it works
- While preloadfonts is on, we add can have 2 preconnect in the page https://new.rocketlabsqa.ovh/preloadfonts_10atf_10btf_custom_diff to the same URL 
`<link href='https://fonts.gstatic.com/' crossorigin rel='preconnect' />`
and `<link crossorigin data-rocket-preconnect href="https://fonts.gstatic.com/" rel="preconnect">`
- While preconnect external is off, and preload fonts on , we aren't preconnecting GF API https://new.rocketlabsqa.ovh/preloadfonts_10atf_10btf_custom_diff (they should be independent as mentioned here https://group-onecom.slack.com/archives/C08F4LZB2QG/p1747979001426029?thread_ts=1747913825.994039&cid=C08F4LZB2QG)
- adding svg extension with filter is not working (same on 3.19 commit mentioned above) https://group-onecom.slack.com/archives/C08F4LZB2QG/p1747993364630799 => may need GH
- GF added with @import not handled here yet , https://github.com/wp-media/wp-rocket/issues/7378
- Preconnect is not added to external fonts in CSS https://github.com/wp-media/wp-rocket/issues/7408

PR #7474, opened 2025-06-26
Fix #7473: Add conditional loading for TrackingServiceProvider based on cURL availability (setting manually true to the condition as I can't disable this extension)

### LINKED ISSUE #7473: Fatal Error in 3.19.1.1 if cUrl is not installed on the server
**Context**

3.19.1.1 requires curl to be available on the server. Some users reported a Fatal error when this is not the case. We should bypass the associated functionalities if cUrl is not available, to avoid the error.

**Expected behavior**

When running 3.19.1.1 on a server without cUrl installed/activated, I can still use WP Rocket without Fatal Error.

**Acceptance Criteria**

On a server without cUrl installed:
- install WP Rocket 3.19.1.1, there should not be a fatal error.
- Try to activate Rocket Analytics. There should not be a fatal error. Note: the analytics settings might not be saved/changed, this would be OK.
- Enable auto_preload_fonts. There should not be a fatal error.

**Additional information**

https://group-onecom.slack.com/archives/C08F4M3BBJL/p1750859099781669

Miraeld:
# Description

Fixes #7473 
If `cURL` extension isn't available we won't load the subscriber related to MixPanel. 

## Type of change

### What was tested

The whole plugin without curl extension.

### How to test

Disable the curl extension on the server and enable tracking, then load WP Rocket.

### Documentation

This pull request introduces a conditional check to ensure compatibility with servers that lack cURL support. It modifies the initialization of services and subscribers in the `Plugin.php` file to prevent fatal errors caused by the MixPanel dependency when cURL is unavailable.

### Compatibility Improvements:

* [`inc/Plugin.php`](diffhunk://#diff-b591480fdcd92670a10d4e86f4f973f95ad761f81d44ce8f04109579a10dfebfR207-R213): Added a conditional check in the `init_admin_subscribers` method to include the `TrackingServiceProvider` only if the cURL extension is loaded. This prevents crashes on servers without cURL support.
* [`inc/Plugin.php`](diffhunk://#diff-b591480fdcd92670a10d4e86f4f973f95ad761f81d44ce8f04109579a10dfebfL235-R243): Updated the `init_admin_subscribers` method to conditionally add `tracking_subscriber` to the subscribers list only if cURL is available, avoiding fatal errors from the MixPanel dependency.
### New dependencies

None

### Risks

None

# Mandatory Checklist

## Code validation

## Code style

### DIFF
--- a/inc/Plugin.php
+++ b/inc/Plugin.php
@@ -212,35 +212,7 @@ class Plugin {
 			$this->container->addServiceProvider( new TrackingServiceProvider() );
 		}
 
-		return [
-			'beacon',
-			'settings_page_subscriber',
-			'deactivation_intent_subscriber',
-			'hummingbird_subscriber',
-			'rocketcdn_admin_subscriber',
-			'rocketcdn_notices_subscriber',
-			'rocketcdn_data_manager_subscriber',
-			'critical_css_admin_subscriber',
-			'health_check',
-			'minify_css_admin_subscriber',
-			'admin_cache_subscriber',
-			'google_fonts_admin_subscriber',
-			'image_dimensions_admin_subscriber',
-			'defer_js_admin_subscriber',
-			'lazyload_admin_subscriber',
-			'preload_admin_subscriber',
-			'minify_admin_subscriber',
-			'action_scheduler_check',
-			'actionscheduler_admin_subscriber',
-			'domain_change_subscriber',
-			'lazyload_css_admin_subscriber',
-			'post_edit_options_subscriber',
-			'preconnect_external_domains_admin_subscriber',
-			'media_fonts_admin_subscriber',
-			'preload_fonts_admin_subscriber',
-		// Conditionally add tracking_subscriber only if cURL is available.
-		// to prevent fatal errors from MixPanel dependency.
-		array_merge(
+		return array_merge(
 			[
 				'beacon',
 				'settings_page_subscriber',
--- a/inc/Plugin.php
+++ b/inc/Plugin.php
@@ -204,7 +204,13 @@ class Plugin {
 		$this->container->addServiceProvider( new OptimizationAdminServiceProvider() );
 		$this->container->addServiceProvider( new DomainChangeServiceProvider() );
 		$this->container->addServiceProvider( new AdminLazyloadCSSServiceProvider() );
-		$this->container->addServiceProvider( new TrackingServiceProvider() );
+		
+		// Only add tracking service provider if cURL extension is loaded.
+		// MixPanel (used by TrackingServiceProvider) requires cURL and will throw a fatal error if not available.
+		// This prevents crashes when WP Rocket is installed on servers without cURL support.
+		if ( extension_loaded( 'curl' ) ) {
+			$this->container->addServiceProvider( new TrackingServiceProvider() );
+		}
 
 		return [
 			'beacon',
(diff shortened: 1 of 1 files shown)


inc/Plugin.php
```diff
@@ -232,8 +238,38 @@ private function init_admin_subscribers() {
...
+				'actionscheduler_admin_subscriber',
+				'domain_change_subscriber',
+				'lazyload_css_admin_subscriber',
+				'post_edit_options_subscriber',
+				'preconnect_external_domains_admin_subscriber',
+				'media_fonts_admin_subscriber',
+				'preload_fonts_admin_subscriber',
+			],
+			extension_loaded( 'curl' ) ? [ 'tracking_subscriber' ] : []
+		);
```
jeawhanlee: It looks like something is off with this array, there seem to be duplicate elements. Also instead of using array merge, I would recommend going with dynamic indexing which is much faster for this kind of use case.

jeawhanlee (changes requested):

inc/Plugin.php
Miraeld: Yea copilot did something wrong on its review ... 
About the `array_merge` I'm not sure it makes a huge difference to be honest.

inc/Plugin.php
```diff
@@ -204,36 +204,44 @@ private function init_admin_subscribers() {
...
```
jeawhanlee: @Miraeld I believe it's an overkill to use `array_merge` when we just need to add a single element to the existing array plus we would also be performing a useless merge with an empty array when curl extension is not enabled on the server. Since we are a performance plugin, it wouldn't hurt to optimize as much as possible where necessary  Also for this scenario, dynamic indexing would improve code readability, no need for nesting ternary function in `array_merge`.
```suggestion
		$admin_subscribers = [
				'beacon',
				'settings_page_subscriber',
				'deactivation_intent_subscriber',
				'hummingbird_subscriber',
				'rocketcdn_admin_subscriber',
				'rocketcdn_notices_subscriber',
				'rocketcdn_data_manager_subscriber',
				'critical_css_admin_subscriber',
				'health_check',
				'minify_css_admin_subscriber',
				'admin_cache_subscriber',
				'google_fonts_admin_subscriber',
				'image_dimensions_admin_subscriber',
				'defer_js_admin_subscriber',
				'lazyload_admin_subscriber',
				'preload_admin_subscriber',
				'minify_admin_subscriber',
				'action_scheduler_check',
				'actionscheduler_admin_subscriber',
				'domain_change_subscriber',
				'lazyload_css_admin_subscriber',
				'post_edit_options_subscriber',
				'preconnect_external_domains_admin_subscriber',
				'media_fonts_admin_subscriber',
				'preload_fonts_admin_subscriber',
			];
			
		if ( extension_loaded( 'curl' ) ) {
		  $admin_subscribers[] = 'tracking_subscriber';
		}
		
		return $admin_subscribers;
```

jeawhanlee (changes requested):

inc/Plugin.php
wordpressfan: Done

wordpressfan:
Managed to replicate the fatal error:

image

and the current PR fixes the case fatal error when the curl extension is not enabled, but I think this won't work for the hosts that have curl enabled but `curl_init` function is added in the disabled functions, I think this should be a rare case, anyway to cover this also, we can add a condition like:
```
function_exists( 'curl_init' )
```
what do u think?

jeawhanlee (approved):

PR #7589, opened 2025-09-01
Update delay JS script

MathieuLamiot:
Created by Github action


hanna-meda:
- [x] Related TestRail TP: https://wpmediaqa.testrail.io/index.php?/runs/view/1165 (test by dev)
- [x] Explore different templates
- [x] Explore different themes

Mai-Saad:
checking different themes without causing console error is good on e2e (still we need to explore mobile view to ensure it isnot open by default and matching nowprocket UI + old templates)
image


**Update: e2e is updated to handle mobile menu as wel**l

Mai-Saad:
Working as expected.

Note: template https://new.rocketlabsqa.ovh/delay-js-regex, cached page has a different order for js execution than nocache (same on 2.0.4)

PR #7627, opened 2025-09-10
Feature 7593: limit urls

CrochetFeve0251:
# Description

Fixes #7593
This PR implements a limit on the number of URLs that can be monitored in the Speed Radar (Performance Monitoring) feature based on the user's license type. Free users are limited to 3 URLs while paid users (advanced license) are allowed up to 10 URLs.

## Type of change

- [x] Enhancement (non-breaking change which improves an existing functionality).

### What was tested

The following scenarios were tested with automated integration tests:
- Free user with 0 URLs can add new pages (returns true)
- Free user with 3 URLs cannot add more pages (returns false)
- Paid user with 3 URLs can add more pages (returns true)
- Paid user with 10 URLs cannot add more pages (returns false)

Additionally, the UI interaction was tested manually to confirm that:
- The "Add Page" button is disabled when the URL limit is reached
- Appropriate error messages are displayed when attempting to add URLs beyond the limit

### How to test

1. **Setup:**
   - Install and activate WP Rocket
   - Navigate to the Performance Monitoring section in WP Rocket settings

2. **Testing Free User Limits:**
   - Use a free license key
   - Add 3 URLs to the Performance Monitoring section
   - Verify the "Add Page" button becomes disabled
   - Attempt to add a 4th URL using browser developer tools (by enabling the button) and verify an error message appears

3. **Testing Paid User Limits:**
   - Use a premium/advanced license key
   - Add 10 URLs to the Performance Monitoring section
   - Verify the "Add Page" button becomes disabled
   - Attempt to add an 11th URL using browser developer tools and verify an error message appears

### Documentation

This implementation enforces URL limits in the Performance Monitoring feature:

1. **URL Limit Subscriber Class:**
   - Implements a new subscriber class that hooks into WordPress to enforce URL limits
   - Checks the current license type and number of monitored URLs
   - Returns `true/false` based on whether adding a new URL is allowed

2. **Front-end Implementation:**
   - Disables the "Add Page" button when the limit is reached
   - Adds a tooltip explaining why the button is disabled

3. **Backend Validation:**
   - Prevents adding URLs via AJAX if the limit is reached
   - Returns a user-friendly error message when a limit is hit

4. **Integration with License System:**
   - Connects to the existing license API to determine user license type
   - Sets appropriate limits based on the license (3 for free, 10 for paid)

The implementation follows the existing architectural patterns in the codebase, using:
- Subscriber pattern for event-driven functionality
- Dependency injection for cleaner, testable code
[...]

Issue #7626, opened 2025-09-09
Follow-up task on new credit system


wordpressfan:
This is a follow-up issue on #7625 for the following points:

- Once PR #7623 is merged, we need to get the user class and check if the current plan is the free plan using the SKU exactly here: 

https://github.com/wp-media/wp-rocket/blob/9609c743d9ffaffe5bf9150b46ef103f7dfda491/inc/Engine/Admin/PerformanceMonitoring/Context/FreePlanContext.php#L26

- Once this UI issue #7603 is merged, we need to disable the retest button if the credit is zero using the credit manager we introduced, also don't open the report if it's blurred.
- <strike>Add more tests to increase the coverage. </strike> To release Alpha, we need to move this away from this issue and open another dedicated issue to fix this.
- Remove the credit reset recurring action scheduler task with plugin un-installation.

Mai-Saad:
For test , validate AC here https://github.com/wp-media/wp-rocket/issues/7622

Mai-Saad:
Thanks @sandyfzu for testing all those cases 
- **Free tests**
    - Customer receives 3 tests per month ✔️ 
    - The month cycle starts when customer install 3.20 or update to 3.20 ✔️ 
    - After one month we’ll reset the test value back to 3 ✔️ 
- **Usage of the free tests**
    - A test is used when:
        - New URL is added for testing ✔️ 
        - Retest is triggered on already existing URL ✔️ 
    - If customer doesn’t have tests available:
        - And new URL is added (free users can add up to 3 URLs only) - the test is done, but the result is blurred ✔️ 
        - The re-test button for existing URLs will not be active - it’s not possible to do the re-test ❌  => retest is active till manual refresh
- **Blurred vs unblurred results**
    - Unblurred
        - The results are unblurred when a free test was available at the time of the test or retest ✔️ 
    - Blurred
        - Displayed when the URL was added without free tests available ✔️ 
        - **See Report** will be blocked on blurred results ✔️ 
    - Re-tests on blurred results will become unblurred if done when the test is available ✔️ 
    - Blurring flow will apply to the URLs in the `Rocket Insights` tab and also to the Post list column (part of different US) ✔️ 
- **URL removal**
    - Deleting a URL does not refund any tests used for:
        - First test when URL was added ✔️ 
        - Re-tests ✔️ 
- **Problems**
    - When the test was not done because of any error, the number of tests available should not decrease == > 👍

PR #7625, opened 2025-09-09
New credit system

wordpressfan:
# Description

Fixes #7622 

New credit system implementation.

## Type of change

- [x] New feature (non-breaking change which adds functionality).
- [ ] Bug fix (non-breaking change which fixes an issue).
- [ ] Enhancement (non-breaking change which improves an existing functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as before).
- [ ] Sub-task of #(issue number)
- [ ] Chore
- [ ] Release

### What was tested

- Once plugin is activated, Action scheduler action is created with the name `rocket_pma_credit_reset` which is a recurring task happens every month from the plugin installation.
- Check the database, wp_option table to see the credit on the row with `option_name=wp_rocket_pm_credit` with first activation, you will find the credit is three and once the homepage is completed u will find it two.
- Add two more pages and wait until it's completed, then check the credit, you will find it zero.
- Add one more page after credit is zero and wait until it's completed, you will find that is_blurred in the database row equals zero.
- Try cancelling or running the AS action `rocket_pma_credit_reset`, it will not affect the credit until one month pass.

### How to test

Mentioned above

### Documentation

I don't want to talk about this in public repo at least for now.

### New dependencies

N/A

### Risks

N/A

# Mandatory Checklist

## Code validation

- [x] I validated all the Acceptance Criteria. If possible, provide screenshots or videos.
- [x] I triggered all changed lines of code at least once without new errors/warnings/notices.
- [ ] I implemented built-in tests to cover the new/changed code.

## Code style

- [x] I wrote a self-explanatory code about what it does.
- [x] I protected entry points against unexpected inputs.
- [x] I did not introduce unnecessary complexity.
- [x] Output messages (errors, notices, logs) are explicit enough for users to understand the issue and are actionnable.

## Unticked items justification

Created this follow-up task #7626 to complete the missed functionalities here.

# Additional Checks
- [ ] In the case of complex code, I wrote comments to explain it.
- [ ] When possible, I prepared ways to observe the implemented system (logs, data, etc.).
- [ ] I added error handling logic when using functions that could throw errors (HTTP/API request, filesystem, etc.)

Issue #7591, opened 2025-09-01
As a website administrator, I want to see the Global Score in the WP Rocket dashboard


piotrbak:
**Context**

This is a main issue for the calculation and display of the Global Score widget.

**Expected behavior**

Global Score should be calculated and displayed in the right sidebar of WP Rocket dashboard.

**Acceptance Criteria**

- The RocketInsights Global Score will be displayed in the sidebar of the dashboard
- **Calculation of Global Score**
    - The Global Score is calculated as the average of all available individual scores
    - Only URLs with valid scores are included in the average, so, if result is failed for any reason, it should not be part of the average
    - Removing URLs should trigger a recalculation of the Global Score.
    - If all tests are failed, we’ll be in the `No URLs are added state`
- There’ll be 4 states possible for this sidebar:
    - No URLs are added
        - It’s displayed when there are no URLs added to the add-on
        - At the top we’ll see a string `0 pages monitored`
        - Below we’ll see large radar icon
        - Below we’ll see a button `ADD HOMEPAGE`
            - Clicking on it will add homepage to the add-on and move it to the `Generating state`  and it’ll switch to the `Generating state`
    - Generating state
        - It’s displayed when there’s a test happening in the background
            - When adding the first page
            - When adding new URL
            - When retesting existing URL
        - At the top we’ll see a string `x pages monitored`
            - **x** stands for the **updated** number of currently monitored pages
                - If we added a first URL, the **x** will be **1**
                - If we added a second URL, the **x** will be **2**
        - Below we’ll see loader animation
        - Below we’ll see a button `ADD PAGE`
            - Clicking on it will redirect customer to the RocketInsights tab
        - After the test is finished, the sidebar will automatically move us to the states mentioned below
    - URL(s) added **and no result is blurred**
        - At the top we’ll see a string `x pages monitored`
            - **x** stands for the number of currently monitored pages
        - Below we’ll see the Global Score
        - Below we’ll see a button `ADD PAGE`
            - Clicking on it will redirect customer to the RocketInsights tab
    - URL(s) added **and at least one of the result is blurred (only for free users)**
        - At the top we’ll see a string `x pages monitored`
            - **x** stands for the number of currently monitored pages, including blurred
        - Below we’ll see blurred Global Score
            - Clicking on it will redirect customer to the RocketInsights tab and we’ll display a CTA banner
        - Below we’ll see a button `ADD PAGE`
            - Clicking on it will redirect customer to the RocketInsights tab

[...]

PR #7632, opened 2025-09-11
Enhance #7628: Calculation global score blurred state

### LINKED ISSUE #7628: Enable is_blurred in global score calculation
The following was the grooming the global score calculation:
# Grooming

- As per our agreement [here](https://github.com/wp-media/wp-rocket/issues/7607#issuecomment-3265453742) we will use this method to return the global score.
- Create a new class to handle the global score cases which will have the following methods:
  - ~~**get_global_score:** Will check the transient if it's there get it otherwise use a custom sql query to sum the scores of all successful rows in DB and divide by the count and then save it into the transient~~
  - ~~**get_pages_number:** Uses the query to get the number of pages being monitored (regardless the status)~~
  - ~~**get_current_status:** Returns one status from the following statuses:~~
    - ~~`no-url` when no url is there to be monitored~~
    - ~~`in-progress` when at least one url's status is one of those statuses (to-submit, in-progress, or pending)~~
    - ~~`complete` when we don't have at least one url with the previously mentioned statuses for `in-progress` state~~
    - `blurred` when at least one row is blurred, here we will depend on the is_blurred column in DB, make sure that it's created first (we created it in another task [here](https://github.com/wp-media/wp-rocket/issues/7599))

~~We can use one transient for the whole three items or each one has its own transient or even option.~~

- ~~Create the following new hooks and using them to remove the previously created transients/options:~~
  - ~~rocket_pm_job_completed: [here](https://github.com/wp-media/wp-rocket/blob/fac3f633925061055da4c28cfdca6504e3576585/inc/Engine/Admin/PerformanceMonitoring/Jobs/Manager.php#L130-L129)~~
  - ~~rocket_pm_job_failed: [here](https://github.com/wp-media/wp-rocket/blob/fac3f633925061055da4c28cfdca6504e3576585/inc/Engine/Admin/PerformanceMonitoring/Jobs/Manager.php#L106-L105)~~
  - rocket_pm_job_added: fired in three places: [here](https://github.com/wp-media/wp-rocket/blob/a483eec140bb2efd7d73d13cdcd3bf009bf6ebc2/inc/Engine/Admin/PerformanceMonitoring/Controller.php#L64), ~~[here](https://github.com/wp-media/wp-rocket/blob/e10785fa7ece044a533f4c48c694cfa0db405ae0/inc/Engine/Admin/PerformanceMonitoring/AJAX/Controller.php#L93-L92), and finally for retesting [here](https://github.com/wp-media/wp-rocket/blob/e10785fa7ece044a533f4c48c694cfa0db405ae0/inc/Engine/Admin/PerformanceMonitoring/AJAX/Controller.php#L305-L304).~~
  - ~~rocket_pm_job_deleted: [here](https://github.com/wp-media/wp-rocket/blob/a483eec140bb2efd7d73d13cdcd3bf009bf6ebc2/inc/Engine/Admin/PerformanceMonitoring/Controller.php#L99-L98)~~

## Effort
[S]

Everything has been done except one thing:
    - `blurred` when at least one row is blurred, here we will depend on the is_blurred column in DB, make sure that it's created first (we created it in another task [here](https://github.com/wp-media/wp-rocket/issues/7599))

As the calculation has been implemented before getting the `is_blurred` param, we've decided to comment out the code:
https://github.com/wp-media/wp-rocket/blob/a07590abb0137b5f73e8c5d2b1c7a3a939644c70/inc/Engine/Admin/PerformanceMonitoring/GlobalScore.php#L153-L160

This issue is here to enable the `is_blurred` related part to the calculation of the global score.

Miraeld:
# Description

Fixes #7628 
Takes into account the `blurred` state in the calculation for the UI. 

## Type of change

### What was tested

Having rows in the DB, with different status should affect as per the AC the widget in the sidebar.

### How to test

Having rows in the DB with different status.

### Documentation

This pull request updates the logic in `GlobalScore.php` to improve how the system calculates the number of pages and determines the current status of performance monitoring. The changes refine the queries to filter out failed pages, correctly count blurred and failed URLs, and ensure the status reflects these conditions.

**Performance Monitoring Logic Improvements:**

* The `calculate_pages_number()` method now excludes pages with a `failed` status from its count, ensuring only relevant pages are considered.

* The `calculate_current_status()` method has been enhanced to:
  - Accurately count URLs with a `blurred` status and return `'blurred'` if any exist.
  - Check if all URLs have a `failed` status and return `'failed'` when appropriate.
  - Use a consistent query structure for counting total URLs.

These changes make the status calculation more robust and ensure that monitoring results accurately reflect the state of the URLs being tracked.

### New dependencies

None

### Risks

None

# Mandatory Checklist

## Code validation

## Code style


remyperona:
Is it ready for review? A lot of checks are failing

Miraeld:
PS: Tests are failing due to the column `is_blurred` is missing from the table related to performance monitoring. 
It will be added [here](https://github.com/wp-media/wp-rocket/pull/7625/files#diff-2b55b00036267fb18c3d76dc9ab3bc639a8e34f0f2b71daa6cdfe1040caf9dbbR128)

inc/Engine/Admin/PerformanceMonitoring/GlobalScore.php
```diff
@@ -120,7 +120,12 @@ private function calculate_global_score(): int {
 	 * @return int Number of pages.
 	 */
 	private function calculate_pages_number(): int {
-		return $this->query->query( [ 'count' => true ] );
+		return $this->query->query(
+			[
+				'count'          => true,
+				'status__not_in' => [ 'failed' ],
```
jeawhanlee: Shouldn't this just be the total number of urls being monitored whether failed or not?

Miraeld: @jeawhanlee from https://github.com/wp-media/wp-rocket/issues/7591, we get this info:
> Calculation of Global Score
The Global Score is calculated as the average of all available individual scores
Only URLs with valid scores are included in the average, so, **if result is failed for any reason, it should not be part of the average**
Removing URLs should trigger a recalculation of the Global Score.
If all tests are failed, we’ll be in the No URLs are added state

I've bolded the interesting sentence.

remyperona (approved):

hanna-meda:
@Miraeld, thank you for this PR.

From my tests, the blurring is always present on a fresh install, regardless of the number of tries/retries, and whether or not the  [snippet](https://github.com/wp-media/wp-rocket/pull/7627#issuecomment-3280351159)  is used to switch between free and paid users. In the database, the `is_blurred` value is always set to "1". Am I missing something?

Miraeld:
Hello @hanna-meda ,
Seems like a problem from the introduction of `is_blurred` rather than the use of it, which this PR is about. 
The introduction and setup of this column has been made by @wordpressfan in https://github.com/wp-media/wp-rocket/pull/7625

In the current PR we are only using the column, not modifying/setting it to any value. See this PR as front end. 

I think this needs to be raised within https://github.com/wp-media/wp-rocket/pull/7625

wordpressfan:
@hanna-meda @Miraeld 
Check this issue: https://github.com/wp-media/wp-rocket/issues/7626
The first point should handle this, now we consider all users are in the free plan.

hanna-meda:
Ok then. Confirming that with this PR, blurring is applied to any URL that has a value of 1 under the `is_blurred` column in the DB. The remaining functionality will be tested under #7626. Thanks.

PR #7675, opened 2025-09-18
Enhancement/7646 flush cache on upgrade

CrochetFeve0251:
## Description

Fixes #7646
This enhancement ensures that when a user upgrades their Performance Monitoring add-on, the cache is properly flushed and user data is refreshed, providing them with immediate access to their new features and preventing any inconsistencies in the user experience.

## Type of change

- [x] Enhancement (non-breaking change which improves an existing functionality).

### What was tested

The implementation was tested through both automated and manual testing:

**Automated testing:**
- Integration tests verify that the cache flush and user data reset functions are properly called when the upgrade process is triggered.
- Tests confirm that the `wp_rocket_pma_upgraded` action is fired correctly when upgrade parameter is present.
- Tests validate that user data is correctly updated after the upgrade process.

**Manual testing:**
- Verified that after upgrading from free to advanced Performance Monitoring plan, the cache is immediately flushed.
- Confirmed that user data is properly refreshed, showing the new subscription details.
- Tested that the redirect to the Performance Monitoring tab works correctly after upgrade.

### How to test

1. Setup a WordPress installation with WP Rocket plugin
2. Ensure a Performance Monitoring free plan is active
3. Navigate to Performance Monitoring tab
4. Simulate an upgrade by adding `?upgrade=1` to the URL
5. Verify that:
   - The cache is flushed (check the network requests)
   - User data is refreshed (the UI should show updated subscription details)
   - You're redirected to the Performance Monitoring tab

### Documentation

This enhancement implements a cache flush mechanism when a user upgrades their Performance Monitoring add-on. The implementation leverages the WordPress action system:

1. A new action hook `wp_rocket_pma_upgraded` is triggered in the `check_upgrade` method of the Performance Monitoring Subscriber when the upgrade parameter is detected.
2. The `reset_user_data` method is attached to this action and performs two key operations:
   - Flushes the user data cache using `$this->user_client->flush_cache()`
   - Fetches fresh user data and updates the current user instance with `$this->user->set_user($user_data)`

This ensures that any cached data related to the user's subscription is cleared and refreshed with the latest information from the server immediately after an upgrade.

### New dependencies

No new dependencies were added for this enhancement.

### Risks

**Performance considerations:**
- The cache flush operation is only triggered during the specific upgrade process, so it doesn't impact regular site performance.
- The implementation uses existing API methods and doesn't introduce additional network requests beyond what's necessary.

**Security considerations:**
[...]

PR #7688, opened 2025-09-19
Keep only the limit number of current plan when downgraded to free plan

### LINKED ISSUE #7645: As a website administrator, I want to go back to the free plan after the paid subscription expires or is cancelled
**Context**

This US describes the scenario when the plugin goes back fro the paid to free plan.

**Acceptance Criteria**

- **In all cases, we’re depending on the API telling us about the state of the license**
    - The scenarios are handled on the SaaS side: **Cancel, Refund, Failed payments**
- We should go to the free plan when we detect that the license is ‘again’ set to free from the API:
    - Only 3 first URLs are present in the list
    - Their results are not blurred
    - Performance Monitoring is deactivated
    - All the other URLs are erased
    - Three free credits are added for the customer

**Additional information**
- Refund - Available during 24 hours after the purchase. Refunding sets the license to free right away
- Cancel - Available at any time. The license will be set to free after the current billing period

### LINKED ISSUE #7696: 3.20 Success message is displayed for free Rocket insights user with fresh install
**Describe the bug**
When opening Rocket Insights for a free user, the success banner is displayed 

**To Reproduce**
1. Buy WPR multi from staging https://www-staging.wp-rocket.me/ => Rocket Insights is free by default
2. Use the staging director in wp-config, new user email, and key in the  config file
3. Use the snippet mentioned here https://group-onecom.slack.com/archives/C091CSKLB1Q/p1758278628848209?thread_ts=1758116253.723679&cid=C091CSKLB1Q
4. Fresh install and activate of 3.20 commit `7e09b246d`
5. Open the Rocket Insights page => success banner is there

**Expected behavior**
- Success banner shouldn't be displayed for free users without paying 1st

**Additional context**
- Related thread https://group-onecom.slack.com/archives/C08F4LZB2QG/p1758544565184219

wordpressfan:
# Description

Fixes #7645 

When downgrading from advanced plan to free plan for any reason, we will keep only the limited number of urls for the free version which is 3, so if the user had 10 we will remove 7 and keep the first three rows.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Use the snippet to change the current plan to be advanced which is limited to 10 then add 10 urls and wait till they finish, then update the snippet again to be free then refresh admin, u will find that only the first three urls will be there.

### How to test

Mentioned above.

### Documentation

Here I added a new option `wp_rocket_insights_current_plan` to save the current plan and then compare it with what we have from the API, if it's not the same we fire an action `rocket_insights_upgrade` which we use later to compare the number of urls in database with the limit coming from the API, so it's more generic not depending on a specific plan.

### New dependencies

N/A

### Risks

N/A

# Mandatory Checklist

## Code validation

## Code style

*If some mandatory items are not relevant, explain why in this section.*

# Additional Checks


inc/Engine/Admin/PerformanceMonitoring/Database/Queries/PerformanceMonitoring.php
```diff
@@ -157,4 +157,25 @@ public function make_blurred( $row_id ) {
...
+			[
+				'fields'  => 'ids',
+				'offset'  => $to_keep,
+				'orderby' => 'modified',
+				'order'   => 'asc',
+			]
+			);
+		foreach ( $ids as $id ) {
+			$this->delete_item( $id );
+		}
```
jeawhanlee, line 179: I think we should optimize this to run the query once, instead of being in a loop and hitting the DB for every id? if not now, then we should open an issue for this and consider an enhancement.

inc/Engine/Admin/PerformanceMonitoring/Managers/Plan.php
jeawhanlee, line 1: Concerning the check upgrade, can you check @CrochetFeve0251 [PR](https://github.com/wp-media/wp-rocket/pull/7675), There seem to be some similarities as he also implemented a hook like this. Although it looks like different scenarios.

jeawhanlee (changes requested):

inc/Engine/Admin/PerformanceMonitoring/Managers/Plan.php
wordpressfan: Agree with you, thanks for noticing such thing, I think the approach in this PR is more generic to handle also the cases of cancel and expire the license but Cyrille's PR is only about upgrading.

So let's keep my PR opened till We merge Cyrille's PR and I can adjust the code to have both scenarios.

Good catch.

inc/Engine/Admin/PerformanceMonitoring/Database/Queries/PerformanceMonitoring.php
wordpressfan: I really was thinking about it because using delete_item method is doing one more step than using a raw query to delete those rows, which is deleting the berlindb internal cache for this deleted item.

So I'm OK with both solutions.
What do u think?

jeawhanlee: It's up to you now 😁

inc/Engine/Admin/PerformanceMonitoring/Render.php
```diff
@@ -169,7 +169,8 @@ public function render_license_banner_section( array $data ) {
 	 * @return void
 	 */
 	public function render_license_banner_plan_price( string $price, string $currency = '$', string $period = 'month' ) {
-		$dot   = get_locale() === 'en_US' ? '.' : ',';
+		global $wp_locale;
+		$dot   = $wp_locale->number_format['decimal_point'] ?? '.';
```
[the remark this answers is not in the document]
wordpressfan: Not valid point, because I already use `??` and it validates all of those cases.

assets/js/wpr-admin-common.js
```diff
@@ -3,7 +3,10 @@ var sent = false;
 	$( '.rocket-dismiss' ).on( 'click', function( e ) {
 		e.preventDefault();
 		var url = $( this ).attr( 'href' ).replace( 'admin-post', 'admin-ajax' );
-		$.get( url ).done( $( this ).closest( '.notice' ).hide( 'slow' ) );
+		$.get( url ).done( () => {
+				$( this ).closest( '.notice, .wpr-notice' ).hide( 'slow' );
+			}
+		);
```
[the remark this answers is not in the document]
wordpressfan: I tested the functionality and it works

inc/Engine/Admin/PerformanceMonitoring/Database/Queries/PerformanceMonitoring.php
```diff
@@ -157,4 +157,25 @@ public function make_blurred( $row_id ) {
...
 	}
+
+	/**
+	 * Limit rows to keep the passed argument and remove others.
+	 *
+	 * @param int $to_keep Number of rows to keep.
+	 *
+	 * @return void
+	 */
+	public function limit_urls( int $to_keep ) {
```
remyperona: I would suggest to rename the method, as just reading the name doesn't tell me it's going to remove data. At first glance, I thought it would only return the number of URLs provided as parameter

inc/Engine/Admin/PerformanceMonitoring/Managers/Plan.php
```diff
@@ -0,0 +1,134 @@
...
+	 *  User client API instance.
+	 *
+	 * @var UserClient
+	 */
+	private $user_client;
+
+	/**
+	 * Current plan option name.
+	 */
+	const CURRENT_PLAN_OPTION_NAME = 'insights_current_plan';
```
remyperona, line 44: shouldn't it be prefixed?

remyperona (changes requested):

inc/Engine/Admin/PerformanceMonitoring/Managers/Plan.php
wordpressfan: we are using options_api which is adding the prefix `wp_rocket_`

jeawhanlee (approved):

PR #7704, opened 2025-09-24
Enhancement #7674: retest needs manual refresh to be dimmed when no credit in certain scenario

### LINKED ISSUE #7674: 3.20 alpha2 retest needs manual refresh to be dimmed when no credit in certain scenario
**Describe the bug**
When adding a URL while having 1 credit, the restest btn is not dimmed in some of the existing URLs till manual refresh, although credit is o

**To Reproduce**
1. Fresh install/activation of a free user (3.20-alpha2)
2. Wait till home is processed (remaining 2 credits)
3. Add URL and after 10 secs add another URL
4. Wait till both URLs are completed and check Rocket insights without manual refresh => only 1st and last URLs have retest dimmed while 2nd URL has an active retest

**Expected behavior**
- Retest of all URLs is dimmed without manual refresh when the credit is 0

**Screenshots**

https://github.com/user-attachments/assets/479683bc-e7ed-4c05-a250-ab745b7cefeb

Miraeld:
# Description

Fixes #7674 
This fix improves the user experience in the Performance Monitoring (Rocket Insights) feature by ensuring retest buttons are properly disabled in real-time when credits are exhausted, without requiring a manual page refresh.

## Type of change

### What was tested

Manual test: Reach the maximum credits used so it should be dimmed while not refreshing the page.

### How to test

Check the issue.

### Documentation

This pull request introduces a credit-checking mechanism to the Performance Monitoring feature, ensuring that certain actions are only available to users with sufficient credits. The backend now provides a `has_credit` flag in AJAX responses, and the frontend uses this to enable or disable retest buttons and show appropriate tooltips. The changes span both PHP (backend) and JavaScript (frontend) to provide a seamless user experience.

**Backend: Credit Checking and API Updates**
- Added a `CreditManager` dependency to the `Controller` class, storing it as a private member and injecting it via the constructor. (`inc/Engine/Admin/PerformanceMonitoring/AJAX/Controller.php`) [[1]](diffhunk://#diff-26007bb022d8ac7e2f984a1965468bf851a0a5c01ab4624694ada073e73250d9L12-R13) [[2]](diffhunk://#diff-26007bb022d8ac7e2f984a1965468bf851a0a5c01ab4624694ada073e73250d9R64-R70) [[3]](diffhunk://#diff-26007bb022d8ac7e2f984a1965468bf851a0a5c01ab4624694ada073e73250d9R80-R89) [[4]](diffhunk://#diff-77d1eb49c422a18485050ab6cb8444dca5e887f0094c35c0b865a7c31f5c690bR150)
- Implemented a private `user_has_credit()` method in the controller to determine if the current user has available credits for performance monitoring, using `CreditManager` for free users. (`inc/Engine/Admin/PerformanceMonitoring/AJAX/Controller.php`)
- Updated AJAX response payloads in `add_new_page`, `get_results`, and `reset_page` to include the `has_credit` flag, reflecting the user's current credit status. (`inc/Engine/Admin/PerformanceMonitoring/AJAX/Controller.php`) [[1]](diffhunk://#diff-26007bb022d8ac7e2f984a1965468bf851a0a5c01ab4624694ada073e73250d9R158) [[2]](diffhunk://#diff-26007bb022d8ac7e2f984a1965468bf851a0a5c01ab4624694ada073e73250d9R295) [[3]](diffhunk://#diff-26007bb022d8ac7e2f984a1965468bf851a0a5c01ab4624694ada073e73250d9R355)

**Frontend: UI Updates Based on Credit Status**
- Added logic to track the `hasCredit` state and update all retest buttons accordingly, disabling them and showing a tooltip when the user lacks credits. (`src/js/global/ajax.js`) [[1]](diffhunk://#diff-043c510ef3475c349d06d41dcc8b34e2c509f6eda990d9de5a1686b06e46126cR273) [[2]](diffhunk://#diff-043c510ef3475c349d06d41dcc8b34e2c509f6eda990d9de5a1686b06e46126cR342-R378)
- Integrated calls to `updateCreditState` in AJAX success handlers for relevant actions, ensuring the UI reflects the current credit status after each backend interaction. (`src/js/global/ajax.js`) [[1]](diffhunk://#diff-043c510ef3475c349d06d41dcc8b34e2c509f6eda990d9de5a1686b06e46126cR433-R435) [[2]](diffhunk://#diff-043c510ef3475c349d06d41dcc8b34e2c509f6eda990d9de5a1686b06e46126cR494-R496) [[3]](diffhunk://#diff-043c510ef3475c349d06d41dcc8b34e2c509f6eda990d9de5a1686b06e46126cR538-R540)
### New dependencies

None

### Risks

None

# Mandatory Checklist

## Code validation

## Code style


src/js/global/ajax.js
```diff
@@ -351,6 +352,51 @@ document.addEventListener('DOMContentLoaded', function() {
...
+	}
+
+	function updateAllRetestButtons() {
+		const retestButtons = document.querySelectorAll('.wpr-action-speed_radar_refresh');
+		
+		// Cache all rows that are currently running to avoid repeated DOM queries
+		const runningRows = new Set(
+			Array.from(document.querySelectorAll('.wpr-pma-item')).filter(row =>
+				row.querySelector('.wpr-loading-img') !== null
+			)
```
wordpressfan: We have pm_ids that we can use instead of depending on something in the dom, what do u think?

src/js/global/ajax.js
```diff
@@ -270,6 +270,7 @@ document.addEventListener('DOMContentLoaded', function() {
 	let pmIds = Array.isArray(window.rocket_ajax_data?.pm_ids) ? window.rocket_ajax_data.pm_ids.slice() : [];
 	let pollInterval = POLL_BASE_INTERVAL;
 	let pollTimer = null;
+	let hasCredit = true; // Track credit status
```
wordpressfan, line 273: do we need to change this to be defaulted to false?

wordpressfan (changes requested):
Minor questions
Good job @Miraeld

src/js/global/ajax.js
jeawhanlee, line 1: @Miraeld I think we can make things a much simpler. We only need to check the credit status when a status is complete because that is when I believe a credit would be decreased. So what if we just check the status here in this [conditional](https://github.com/wp-media/wp-rocket/blob/515dfc7d4dfee3cef1fccf6809684bc00570e344/src/js/global/ajax.js#L462) and if no credit, just disable all retest button, I believe this will be easier.

jeawhanlee (changes requested):

src/js/global/ajax.js
Miraeld: Ummm, I would say it depends on the UX Goal

**Arguments for `false`:**
- Safer: avoids users clicking buttons before we know their credit status
- No mistakes: we won’t accidentally show active buttons to users without credits
- Smooth upgrade: buttons activate once we’re sure it’s valid

**Arguments for `true`:**
- Better for paying users: most users do have credits, so it feels more natural to show buttons as active
- Faster experience: avoids the visible “flash” from disabled → enabled for most people
- Consistent with current state: server-rendered rows already display the right button states

In my opinion, we should kepp it `true`, and the reasons are simple:
- The server-rendered state is already correct at page load (so we’re not taking risks there).
- This mainly impacts dynamic updates via AJAX, not the initial table.
- Optimizing for the majority (users with credits) makes the interface feel faster and friendlier.


What do you think @wordpressfan

src/js/global/ajax.js
Miraeld: Hello hello, 
It's a great suggestion to simplify the approach, however I've had some thoughts. 

For context, your proposed solution checks credits directly in the completion handler and it's clean, straightforward, and solves the immediate problem with minimal code. The logic is easy to follow: when a job finishes, check if credits are exhausted and disable buttons accordingly.

My original approach used centralized credit state management with separate functions for state updates and UI changes. While more complex for this single use case, I was thinking about scenarios where we might expand credit usage in the future.

So I got a question for @wordpressfan , are we thinking to extends the use of credits feature in the future ? If yes, I'm thinking that my approach is more suitable. What do you think ?

src/js/global/ajax.js
Miraeld: Done in last commit

src/js/global/ajax.js
wordpressfan: What do u mean by `extends`? plz explain the use case of ur approach just to understand and think together.

src/js/global/ajax.js
wordpressfan: I'm OK with both approaches but at this stage we can't say that most users have credit, we really don't know :) but we'd like to do is to satisfy the free users to encourage them to buy.

We can leave it as it's and see if we have any problem with the current approach.

src/js/global/ajax.js
Miraeld: Hello, 
By `extends` I mean that right now only the `test` process is consuming a credit when it's `completed`. However, if we look at the earlier version of 3.20, we had the action of adding a page costing also a credit. 

So it made me think that we will [maybe] have a request later to add another action costing credits, and that's why I did the development in the way I did. 


@jeawhanlee , after discussing with @wordpressfan, we've concluded to let it as it is, as it works, and if we are really sure later that no other actions will cost credits, we could refactor it to simplify it as you mentioned during a cooldown or something like this. 


Hence, it's back to review :)

PR #7795, opened 2025-10-09
Enhance SaaS requests enhancements

wordpressfan:
# Description

This PR reverts some changes made to RUCSS job handling and also caters for 410 response code.

## Type of change

### What was tested

Automated test.
Cater for jobs with 410 response code and fail instantly without retries.

### How to test

Will need to mock the response to see this.

### Documentation

- Moved back instant failing with make_status_failed in validate_and_fail method to JobProcessor
- Cater for 410 response code by failing job instantly without retrying
- Update default response code to 599

### New dependencies

N/A

### Risks

N/A

# Mandatory Checklist

## Code validation

## Code style

Not applicable to the changes in this PR

# Additional Checks

### DIFF
--- a/inc/Engine/Common/JobManager/JobProcessor.php
+++ b/inc/Engine/Common/JobManager/JobProcessor.php
@@ -283,14 +283,7 @@ class JobProcessor implements LoggerAwareInterface {
 			$job_factory       = $this->factories[ $optimization_type ] ?? $this->factories['rucss'];
 
 			if ( ! $response || ! $job_factory->api()->validate_add_to_queue_response( $response ) ) {
-				$job_factory->manager()->validate_and_fail(
-					[
-						'status'  => 'failed',
-						'message' => 'To Submit request failed',
-					],
-					$row,
-					$optimization_type
-					);
+				$this->make_status_failed( $row->url, $row->is_mobile, '', '', $optimization_type );
 				continue;
 			}
 
--- a/inc/Engine/Common/JobManager/Strategy/Strategies/DefaultProcess.php
+++ b/inc/Engine/Common/JobManager/Strategy/Strategies/DefaultProcess.php
@@ -33,6 +33,9 @@ class DefaultProcess implements StrategyInterface {
 		0 => 60,   // 1 minutes
 		1 => 120,  // 2 minutes
 		2 => 300,  // 5 minutes
+		3 => 600,  // 10 minutes.
+		4 => 1200, // 20 minutes.
+		5 => 1500, // 25 minutes.
 	];
 
 	/**
--- a/inc/Engine/Common/JobManager/Strategy/Strategies/DefaultProcess.php
+++ b/inc/Engine/Common/JobManager/Strategy/Strategies/DefaultProcess.php
@@ -33,9 +33,6 @@ class DefaultProcess implements StrategyInterface {
 		0 => 60,   // 1 minutes
 		1 => 120,  // 2 minutes
 		2 => 300,  // 5 minutes
-		3 => 600,  // 10 minutes.
-		4 => 1200, // 20 minutes.
-		5 => 1500, // 25 minutes.
 	];
 
 	/**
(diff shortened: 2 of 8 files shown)


inc/Engine/Optimization/RUCSS/APIHandler/APIClient.php
```diff
@@ -114,7 +114,7 @@ public function get_queue_job_status( $job_id, $queue_name, $is_home = false ) {
 		}
 
 		$default = [
-			'code'     => 400,
+			'code'     => 401,
```
MathieuLamiot: Can we use something else than 401? If the goal is just to match one of the code that will match the "JobSetFail" strategy, could we use a specific code instead of one that is also possibliy returned by the API?
It could be 99, 599, etc. And we add it to the list for JobSetFail too?
It will avoid confusions in the future.

inc/Engine/Common/JobManager/Strategy/Strategies/DefaultProcess.php
```diff
@@ -33,9 +33,6 @@ class DefaultProcess implements StrategyInterface {
 		0 => 60,   // 1 minutes
 		1 => 120,  // 2 minutes
 		2 => 300,  // 5 minutes
-		3 => 600,  // 10 minutes.
```
MathieuLamiot, line 36: We don't see issues with number of GET requests. I would suggest to keep this to help users mitigate the issue.

Miraeld (approved):

inc/Engine/Common/JobManager/JobProcessor.php
```diff
@@ -283,14 +283,7 @@ public function process_on_submit_jobs() {
...
 			if ( ! $response || ! $job_factory->api()->validate_add_to_queue_response( $response ) ) {
-				$job_factory->manager()->validate_and_fail(
-					[
-						'status'  => 'failed',
-						'message' => 'To Submit request failed',
-					],
-					$row,
-					$optimization_type
-					);
+				$this->make_status_failed( $row->url, $row->is_mobile, '', '', $optimization_type );
```
Miraeld, line 286: Doesn't this avoid `rocket_pm_job_failed` to be fired in case of Rocket Insights job ?

wordpressfan: I agree we don't need this change, @jeawhanlee Can u plz explain the why behind this change plz?

jeawhanlee: This change in the first place touched the original RUCSS behavior, if it's just to find a place for the hook to trigger for failed job at the `to-submit` level, I believe we should deal with it on the performance monitoring level. Also when @wordpressfan removed [these lines](https://github.com/wp-media/wp-rocket/blob/cd88a88ad917c73a2b889d76e41f153cea462cfc/inc/Engine/Optimization/RUCSS/Jobs/Manager.php#L116-L119) which were originally not there for RUCSS we then got failed integration test which means a regression. The original code did not have validate_and_fail in it on the `to-submit level`. I believe we should restrict the usage of `validate_and_fail` to only job status check and leave this as it is. For firing the `rocket_pm_job_failed` hook for failed jobs at the `to-submit` level, we can apply some form of decorator pattern, that uses method aliasing.
I'll do that now and you can check.

wordpressfan: Do u know why tests were failing? this may give a hint on what is going on there.

jeawhanlee: Well, this only started failing after your initial [commit](https://github.com/wp-media/wp-rocket/pull/7795/commits/2a4a4a1820c6ebf369a0bad7795a2408173144b2) when you removed the early fail in `validate_and_fail` for RUCSS but this was exactly the same behavior as before just moved to the `validate_and_fail` I believe, so I don't think this is related.

wordpressfan: OK, r u still working on this change?

jeawhanlee: Yes I am, almost done.

wordpressfan:
Going to block this one until seeing what will happen with the SaaS with our last changes.

wordpressfan:
Not needed anymore, the last changes fixed the issue

Issue #7024, opened 2024-10-09
ALR problem with fraction height


Mai-Saad:
**Describe the bug**
White space appeared under the image after applying LRC

**To Reproduce**
1. Visit the page template here https://wp-media.slack.com/archives/CUKB44GNN/p1728467694497799
2. Clear cache
3. Visit the page and compare to nowprocket
4. white space exists in the footer

**Expected behavior**
UI matches no wprocket

**Desktop (please complete the following information):**
Chrome Version 125.0.6422.141 (Official Build) (64-bit)
Ubuntu 24.04.1 LTS

**Additional context**

PR #7815, opened 2025-10-20
Enhancement #7707: Improve time score

### LINKED ISSUE #7707: 3.20 - Improve the time of getting the score
We need to improve the time from adding URL (or clicking on retest) till showing the score.

- Maybe we should start with sending the URL to GTMetrix synchronously, instead of asynchronously.
- Maybe we should have a heartbeat that'd be retrying specific items in the AS queue with higher frequency

We need to investigate that further

**Additional context**

Miraeld:
# Description

Fixes #7707
**Before**: When users clicked "Add Page" or "Re-Test", they experienced a 98-206 second wait (1.6-3.4 minutes) before seeing results, with most of the delay caused by a 60-second cron wait before the test even started.

**After**: Tests now complete in 32-77 seconds (0.5-1.3 minutes) - a 58-67% improvement. The API submission happens immediately when users click the button, eliminating the cron wait entirely.

**Additional UX Improvement**: Reduced UI polling from 15 seconds to 5 seconds max, so when tests complete, the results appear on screen within 5 seconds instead of up to 15 seconds.

**Net Result**: Users get their performance scores 2-3x faster with more responsive UI updates throughout the testing process.

## Type of change

### What was tested

Refer to Issue

### How to test

Refer to issue

### Documentation

This pull request introduces an optimization to the Rocket Insights job submission process by attempting a synchronous API call before falling back to the asynchronous queue. It also lowers the polling intervals on the frontend for a more responsive user experience. The main changes are grouped below:

**Backend: Synchronous API Submission and Job Handling**

* Added a `try_sync_submission` method to the `Manager` class in `Jobs/Manager.php`, which attempts to submit jobs synchronously to the GTMetrix API with a short timeout, logging the outcome and falling back to async if necessary.
* Updated both the `add_new_page` and `reset_page` methods in `AJAX/Controller.php` to use the new synchronous submission method before defaulting to the async queue, and to update job status immediately if the sync submission succeeds. [[1]](diffhunk://#diff-ea90c98c063bfda4b21b5fe4d2f433c5073b30239bb9c8f87c8ef5b50c748e6dL131-R149) [[2]](diffhunk://#diff-ea90c98c063bfda4b21b5fe4d2f433c5073b30239bb9c8f87c8ef5b50c748e6dL345-R382)
* Injected the `APIClient` dependency into the `Manager` class and registered it in the service provider for proper dependency management. [[1]](diffhunk://#diff-fb225c439a95bf68facdcda42d79701077c87bf45e7562f967608bcf0b8048fdR9) [[2]](diffhunk://#diff-fb225c439a95bf68facdcda42d79701077c87bf45e7562f967608bcf0b8048fdR50-R84) [[3]](diffhunk://#diff-304c3d096ea5ce72d2acc923ce3f7f69fd91c364b7b9853aedc4cb1033b2e74fR115)

**Frontend: Improved Polling Responsiveness**

* Reduced the polling interval for job status updates from 5–15 seconds to 2–5 seconds in `src/js/global/ajax.js`, making the UI more responsive to job state changes.

# Mandatory Checklist

## Code validation

## Code style


inc/Engine/Admin/RocketInsights/Jobs/Manager.php
```diff
@@ -230,6 +249,63 @@ private function parse_test_results( array $api_response ): array {
...
+	 * @param array  $additional_details Additional details.
+	 * @return array|false Returns job data on success, false on failure/timeout.
+	 */
+	public function try_sync_submission( string $url, bool $is_mobile, array $additional_details = [] ) {
+		$options = [
+			'is_home' => $additional_details['is_home'] ?? false,
+		];
+
+		// Set a shorter timeout for synchronous calls (10 seconds).
+		add_filter( 'http_request_timeout', [ $this, 'set_sync_timeout' ], 10 );
```
remyperona: I think it would be better to add a new optional argument `$args` to `add_to_queue()`, and pass the timeout value that way instead of using a temporary filter

inc/Engine/Admin/RocketInsights/Jobs/Manager.php
```diff
@@ -230,6 +249,63 @@ private function parse_test_results( array $api_response ): array {
 		return wp_parse_args( $api_response['data']['data'], $defaults );
 	}
 
+	/**
+	 * Attempt synchronous submission to GTMetrix API.
+	 *
+	 * @param string $url URL to test.
+	 * @param bool   $is_mobile Is mobile.
+	 * @param array  $additional_details Additional details.
+	 * @return array|false Returns job data on success, false on failure/timeout.
```
remyperona: We should use WP_Error instead of `false` on failure. This probably should be extended to the `add_to_queue()` method as well

inc/Engine/Admin/RocketInsights/Jobs/Manager.php
```diff
@@ -230,6 +249,63 @@ private function parse_test_results( array $api_response ): array {
...
+	 */
+	public function try_sync_submission( string $url, bool $is_mobile, array $additional_details = [] ) {
```
remyperona: I'm not a fan of this method name, could we think of something better?

inc/Engine/Admin/RocketInsights/AJAX/Controller.php
```diff
@@ -342,18 +370,49 @@ public function reset_page() {
...
+				);
+			}
+		} else {
+			// Fallback to async queue.
+			$this->manager->add_to_the_queue(
+				$row->url, // @phpstan-ignore-line
+				true,
+				$additional_details
+				);
+		}
```
remyperona: We have 2 if/else in this block, could we re-factor it to avoid them?

remyperona (changes requested):

inc/Engine/Admin/RocketInsights/AJAX/Controller.php
```diff
@@ -128,13 +129,40 @@ public function add_new_page(): void {
...
-			[
-				'title' => $page_title,
-			]
-			);
+		$additional_details = [
+			'title' => $page_title,
+		];
+
+		// Try synchronous submission first.
+		$sync_response = $this->manager->attempt_sync_submission( $url, true, $additional_details );
```
jeawhanlee: What do you think about using the `send_api` from `JobProcessor` [here](https://github.com/wp-media/wp-rocket/blob/df0e58c533575d5aa10688af989684fac62bdcd7/inc/Engine/Common/JobManager/JobProcessor.php#L334) instead re-inventing the wheel.

inc/Engine/Admin/RocketInsights/AJAX/Controller.php
```diff
@@ -128,13 +129,40 @@ public function add_new_page(): void {
...
+						'job_id' => $sync_response['uuid'],
+					]
+				);
+				// Fallback to error response since we can't track this job.
+				wp_send_json_error(
+					[
+						'message' => __( 'Unable to save performance test. Please try again.', 'rocket' ),
+					]
+				);
+			}
```
jeawhanlee: Also can we isolate this whole logic that attempts the synchronous process to a private/protected method. Doing that will reduce the if/else nesting so we bail out easily or return early.

The new method will either return an int or string which is typically the row_id or an empty string will be automatically be validated [here](https://github.com/wp-media/wp-rocket/blob/760eec748634d30f086d6daaa79a7c5319dcf78c/inc/Engine/Admin/RocketInsights/AJAX/Controller.php#L167)

inc/Engine/Admin/RocketInsights/AJAX/Controller.php
```diff
@@ -342,18 +370,48 @@ public function reset_page() {
...
+				wp_send_json_error(
+					[
+						'message' => __( 'Unable to reset performance test. Please try again.', 'rocket' ),
+					]
+				);
+			}
+
+			// Update to pending status immediately with job ID.
+			$this->query->make_status_pending( $row->url, $sync_response['uuid'], '', true ); // @phpstan-ignore-line
+		}
```
jeawhanlee, line 373: Same thing here, can we re-use the send_api from JobProcessor and also isolate the whole logic to a private/protected method.

inc/Engine/Admin/RocketInsights/APIHandler/APIClient.php
```diff
@@ -40,13 +41,16 @@ public function add_to_queue( string $url, array $options = [] ) {
...
+		$args = array_merge(
+			[
+				'json_encode' => true,
+				'body'        => $request_body,
+				'headers'     => [
+					'Content-Type' => 'application/json',
+				],
 			],
-		];
+			$args
```
jeawhanlee, line 52: I'm not sure why we need this timeout change. @wordpressfan is this necessary?

inc/Engine/Admin/RocketInsights/Jobs/Manager.php
```diff
@@ -46,21 +47,39 @@ class Manager implements ManagerInterface, LoggerAwareInterface {
...
+		Plan $plan,
+		APIClient $api_client
 	) {
-		$this->query   = $query;
-		$this->context = $context;
-		$this->plan    = $plan;
+		$this->query      = $query;
+		$this->context    = $context;
+		$this->plan       = $plan;
+		$this->api_client = $api_client;
```
jeawhanlee: We would no longer need this as a dependency if we use the `send_api` from `JobProcessor`

inc/Engine/Admin/RocketInsights/Jobs/Manager.php
```diff
@@ -230,6 +249,49 @@ private function parse_test_results( array $api_response ): array {
...
 
+	/**
+	 * Attempt synchronous submission to GTMetrix API.
+	 *
+	 * @param string $url URL to test.
+	 * @param bool   $is_mobile Is mobile.
+	 * @param array  $additional_details Additional details.
+	 * @return array|\WP_Error Returns job data on success, WP_Error on failure/timeout.
+	 */
+	public function attempt_sync_submission( string $url, bool $is_mobile, array $additional_details = [] ) {
```
jeawhanlee: We would also not need this anymore.

inc/Engine/Admin/RocketInsights/ServiceProvider.php
```diff
@@ -117,6 +117,7 @@ public function register(): void {
 					'ri_query',
 					'ri_saas_context',
 					'ri_plan',
+					'ri_api_client',
```
jeawhanlee: We would also not need this too.

src/js/global/ajax.js
```diff
@@ -258,8 +258,8 @@ document.addEventListener('DOMContentLoaded', function() {
 	 */
 
 		// ==== Configuration ====
-	const POLL_BASE_INTERVAL = 5000;   // Start polling at 5 seconds
-	const POLL_MAX_INTERVAL = 15000;  // Max polling interval (e.g. 15 seconds)
+	const POLL_BASE_INTERVAL = 2000;   // Start polling at 2 seconds
+	const POLL_MAX_INTERVAL = 5000;   // Max polling interval (5 seconds)
```
jeawhanlee, line 262: My only concern with reducing polling interval is resource consumption. @wordpressfan your input here please

jeawhanlee (changes requested):

inc/Engine/Admin/RocketInsights/AJAX/Controller.php
Miraeld: So I spent some time looking into this "use send_api" suggestion and actually went ahead and implemented it to see how it would work out.

But honestly? After trying it, I think it's making things more complicated than they need to be.

Here's what I ran into:

**The JobProcessor just isn't built for this**
The `send_api()` method returns `array|false`, but our current method returns `array|WP_Error`. So we lose all the nice error handling we had. Plus, JobProcessor is really designed for background jobs, not the synchronous API calls we're doing here.

**We're losing good stuff we already have**
Our `attempt_sync_submission()` method has:

- Specific logging that actually tells us what's happening ("Performance Monitoring: Attempting synchronous submission")
- Custom timeout handling for sync operations
- Proper error codes that make debugging easier
**It just feels forced**
To make this work, I had to change return types across multiple classes, add complexity to the Controller, and basically bend the code to fit something it wasn't designed for.

**What we have now actually works great**
The existing method is well-tested, does exactly what we need, and the code is clean and easy to follow.

I get the idea behind reducing duplication, but sometimes having a specialized method that's built for your specific use case is actually the cleaner approach. The `JobProcessor` is great for what it does, but this isn't really its job.

So I'm thinking we stick with what we have - it's working, it's clear, and it's not trying to be something it's not. What do you think?

jeawhanlee: > Specific logging that actually tells us what's happening ("Performance Monitoring: Attempting synchronous submission")

This would not be a problem the `send_api` only prepares payload, sends to the Saas and returns response, it doesn't really perform complex validation. so you can validate it's response just the same way you do in attempt_sync_submission and so you can also log accordingly based on the response

> Custom timeout handling for sync operations

Like I said earlier, I'm not sure we need this, but if you want to you can add an extra optional param which is a bool to `send_api` and check this param then pass the timeout directly in the add_to_queue there.

> Proper error codes that make debugging easier

I already mentioned this earlier, that you can still add logs just not in `send_api` but based on it's response. As a matter of fact, the send_api method has a log for what you are validating [here](https://github.com/wp-media/wp-rocket/blob/760eec748634d30f086d6daaa79a7c5319dcf78c/inc/Engine/Admin/RocketInsights/Jobs/Manager.php#L273)

> The send_api() method returns array|false, but our current method returns array|WP_Error. So we lose all the nice error handling we had.

We don't have to change the return type. We can keep the return types and validate them and then return WP_Error if it's false.
The only change we need to make is to change the access modifier to public and add a new optional parameter for the timeout if we actually need this. We don't need to modify it so much that it breaks.

### Reasons why I think we should use the `send_api` method:
- We already do this with the cron, so we will be keeping the old behavior of sending and validating with specific response code as seen [here](https://github.com/wp-media/wp-rocket/blob/df0e58c533575d5aa10688af989684fac62bdcd7/inc/Engine/Common/JobManager/JobProcessor.php#L345) without the fear of an unexpected behavior.
- We would not write logic 2x that we already have and use in the cron.
- This method only does one thing (which is send payload to the SaaS and return response), so it's not tightly coupled with other logic and easy to extract it and perform custom logic with it's response exactly the way you're doing with `attempt_sync_submission`

### The idea:
- Change the access modifier of `WP_Rocket\Engine\Common\JobManager\JobProcessor:send_api()`
- if we still need to handle the timeout, we will add an optional param with a type of bool with default value of `false` in send_api
- Check in `send_api` if optional param is true then pass the timeout value to `add_to_queue`
- Create a new method in `WP_Rocket\Engine\Admin\RocketInsights\AJAX\Controller`
- This new method will be shared with `add_new_page` & `reset_page`
- This method will use `send_api` which returns an array or false
- If false, we will continue with the same logs as we do in `attempt_sync_submission` and if not process the response
- Remove `attempt_sync_submission` & `SYNC_TIMEOUT` from the Manager.

But if you still feel uncomfortable doing this, can we at least extract the duplicated logic in `add_new_page` & `reset_page` into a separate method so it can be re-used in both methods.

Miraeld:
@jeawhanlee Could you check the latest changes ?

jeawhanlee (approved):

hanna-meda:
Thank you, @Miraeld, for this PR. We can confirm a time improvement with this change. Below are the timing results comparing PR vs. trunk:

[Measurements from 28 Oct]
PR results:

| URL | Timing on PR | Timing on trunk |
|--------|--------|--------|
| [/october-23](https://rocketlabsqa.ovh/october-23) | 84s | 64s |
| [/march-post](https://rocketlabsqa.ovh/march-post) | 38s | 70s |
| [/business-school](https://rocketlabsqa.ovh/business-school) | 66s | 56s |
| [/new-bus](https://rocketlabsqa.ovh/new-bus) | 55s | 40s |
| [/newer-bus](https://rocketlabsqa.ovh/newer-bus) | 72s | 48s | 
| [/test-lazy-render](https://rocketlabsqa.ovh/test-lazy-render) | 47s | 91s |
| [/est-vel-ipsa-autem-voluptas-quae-sequi](https://rocketlabsqa.ovh/est-vel-ipsa-autem-voluptas-quae-sequi) | 48s  | 46s |
| [/veritatis-expedita-adipisci-enim](https://rocketlabsqa.ovh/veritatis-expedita-adipisci-enim) | 66s | 79s |
| [/ut-porro-sit-possimus-qui](https://rocketlabsqa.ovh/ut-porro-sit-possimus-qui) | 55s | 129s |
| [/delicious-mousse-recipe](https://rocketlabsqa.ovh/delicious-mousse-recipe) | 38s | 106s | 
| Avg. | **56.9s** | **72.9s** |

[More measurements, 29 Oct]

From @Mai-Saad, thank you, on newlabs:
| URL | Timing on Dev | Timing on PR | Avg. Dev vs PR |
| --- | --- | --- | --- |
| [home](https://new.rocketlabsqa.ovh/) | 60s, 80s, 86s, 93s => 79.75s | 64s, 61s, 37s, 69s => 57.75s | 79.75s → 57.75s |
| [post](https://new.rocketlabsqa.ovh/a-voluptas-eum-dolores-consequuntur-non-newmulti-rocketlabsqa-ovh/) | 91s, 82s, 91s => 88s | 32s, 94s, 67s => 64.33s | 88s → 64.33s |
| [page](https://new.rocketlabsqa.ovh/youtube-video/) | 71s, 85s, 93s => 83s | 52s, 96s, 68s => 72s | 83s → 72s |
| [product](https://new.rocketlabsqa.ovh/product/test-burg-4/) | 103s, 84s, 92s => 93s | 41s, 36s, 68s => 48.33s | 93s → 48.33s |
| **Avg.** | **85.4s** | **60.1s** | **↓ 25.3s** |

From Hanna, on rocketlabs:
| URL | Timing on Dev | Timing on PR | Avg. Dev vs PR |
| --- | --- | --- | --- |
| [home](https://rocketlabsqa.ovh/) | 99s, 118s, 157s, 136s => 127.5s | 89s, 73s, 35s, 60s => 64.25s | 127.5s → 64.25s |
| [post](https://rocketlabsqa.ovh/delicious-mousse-recipe) | 126s, 129s, 99s => 118s | 27s, 51s, 56s => 44.67s | 118s → 44.67s |
| [custom post](https://rocketlabsqa.ovh/rock/new-rock) | 127s, 137s, 128s => 130.67s | 72s, 97s, 64s => 77.67s | 130.67s → 77.67s |
| [product](https://rocketlabsqa.ovh/product/caramel-fudge) | 109s, 137s, 126s => 124s | 73s, 71s, 27s => 57s | 124s → 57s |
| [page](https://rocketlabsqa.ovh/page-with-a-looong-title-the-quick-brown-fox-jumps-over-the-lazy-dog) | 128s, 111s, 104s => 114.33s | 78s, 65s, 52s => 65s | 114.33s → 65s |
| **Avg.** | **122.1s** | **61.7s** | **↓ 60.4s** |

Opened [thread](https://group-onecom.slack.com/archives/C08EFCYRQ2X/p1761662618293349) about the AJAX requests being much more frequent now. 

cc @piotrbak, @engahmeds3ed

hanna-meda (approved):
QA done. The timing improvement is noticeable. All NRT tests are passing. 
[testrail-report-775.pdf](https://github.com/user-attachments/files/23213116/testrail-report-775.pdf)

P.S. 
- For the issue where the URL is still loading in the UI in some cases, we’ve  created a dedicated issue: #7842 
- Regarding the misplaced `dot`, it will be handled under [#7024](https://github.com/wp-media/wp-rocket/issues/7824#issuecomment-3461183126 )

PR #7826, opened 2025-10-27
FIX #7821: Add RI to custom posts

### LINKED ISSUE #7821: 3.20.1 Adding RI to custom post type
**Describe the bug**
RI option is not there in the custom post type list nor the products list

**To Reproduce**
1. Fresh install and activation of WPR 3.20.1
2. Open the custom post list or product list => RI option isn't there

**Expected behavior**
RI is there as per AC here https://github.com/wp-media/wp-rocket/issues/7652#issue-3415103380

**Screenshots**

Image
Image

Miraeld:
# Description

Fixes #7821

The Rocket Insights column was not appearing on custom post types (including WooCommerce products) due to an overly restrictive post type query. The code was requiring both `public => true` AND `publicly_queryable => true`, while the rest of WP Rocket only requires `public => true` to determine cacheable post types.

This fix aligns the Rocket Insights column behavior with WP Rocket's standard cacheable post type detection, ensuring the RI column appears on all post types that WP Rocket caches.

## Type of change

### What was tested

**Automated Tests:**
- ✅ Added 5 new test scenarios covering:
  - Standard posts
  - Pages
  - WooCommerce products
  - Custom post types
  - Preservation of existing columns

### How to test

Refer to issue #7821.

### Documentation

**Root Cause:**
The `get_public_post_type_slugs()` method was using both `public => true` and `publicly_queryable => true` parameters. This excluded post types like WooCommerce products which have `publicly_queryable => false` but are still public and cacheable by WP Rocket.

**Solution:**
Changed the query to only require `public => true`, which aligns with how WP Rocket determines cacheable post types throughout the codebase (see `ContentTrait::get_public_post_types()`, metabox logic, and purge actions).

**Flow:**
- WordPress Init
- Subscriber::get_subscribed_events() (static)
- get_public_post_type_slugs()
- get_post_types with only 'public' requirement (FIXED)
- Filter: rocket_insights_excluded_post_types
- Register manage_{post_type}_posts_columns filters
- RI column appears on all cacheable post types

# Mandatory Checklist

## Code validation

## Code style


jeawhanlee (approved):

Mai-Saad:
@Miraeld Thanks for the PR. Please note that RI isnot there in custom post. can you please check?
Screenshot from 2025-10-27 11-48-28
Screenshot from 2025-10-27 11-47-52

Miraeld:
Oh wow, not an easy one! 

So, the integration test was checking if we add the column to custom post types, however, **it wasn't testing the hook registration mechanism itself** - it was calling the methods directly. This masked the real issue.

### Technical Root Cause: Hook Timing

The way we previously handled hook registration was working fine for **default post types** (`post`, `page`), but **failing for custom post types**. Here's why:

**WordPress Loading Timeline:**

1. **`plugins_loaded` hook fires** (when WP Rocket initializes)
   - The `Plugin` class instantiates and calls `$event_manager->add_subscriber()`
   - This immediately invokes `Subscriber::get_subscribed_events()` (a static method)
   - The method calls `get_post_types(['public' => true])`
   
2. **Problem**: Custom post types are registered on the `init` hook
   - At the time of step 1, custom post types **don't exist yet**
   - So `get_post_types()` only returns `['post', 'page', 'attachment']`
   - The column hooks for custom post types like `product`, `portfolio`, etc. are **never registered**

3. **`init` hook fires** - Custom post types are now registered, but it's too late

### Previous Approach (Broken)
```php
public static function get_subscribed_events(): array {
    // This is called during plugins_loaded
    $events = ['admin_enqueue_scripts' => 'enqueue_assets'];
    return array_merge($events, self::get_post_listing_events());
    //                           ↑ Called here - custom post types don't exist yet!
}
```

### New Approach (Fixed)
```php
public static function get_subscribed_events(): array {
    return [
        'admin_enqueue_scripts' => 'enqueue_post_listing_assets',
        'init' => ['register_column_hooks', 20], // ✅ Called AFTER custom post types are registered
    ];
}

public function register_column_hooks(): void {
    $post_types = self::get_public_post_type_slugs(); // ✅ Now includes custom post types!
    
    foreach ($post_types as $post_type) {
        add_filter("manage_{$post_type}_posts_columns", [$this, 'add_rocket_insights_column']);
        add_action("manage_{$post_type}_posts_custom_column", [$this, 'render_rocket_insights_column'], 10, 2);
    }
}
```

**Key changes:**
- Moved hook registration to `init` priority 20 (runs **after** custom post types are typically registered at priority 10)
- Used `add_filter()`/`add_action()` directly in the callback instead of building a static events array
- This ensures `get_post_types()` returns **all** registered post types, not just the defaults


## Bonus Fix

Also added proper SCSS styling for the table column width to prevent layout issues with WooCommerce and other plugins that have aggressive table CSS:


This replaces any inline style hacks and follows WP Rocket's established CSS architecture.

wordpressfan:
Plz check my comment here:
https://group-onecom.slack.com/archives/C08EFCYRQ2X/p1761119113678869?thread_ts=1761118226.988059&cid=C08EFCYRQ2X

I believe using both filters will be better and will solve this issue also.

https://developer.wordpress.org/reference/hooks/manage_pages_columns/

https://developer.wordpress.org/reference/hooks/manage_posts_columns/

What do u think?

remyperona:
You should have a version of the method to get public post types with exclusions, but also without this DB request.

I don't see any value to switch to the global hooks though, the current implementation is good. Only the way of getting the post types list should be improved.

Miraeld:
Hey @remyperona @wordpressfan ! 
I've refactored the post type retrieval logic based on your feedback.

Split `get_public_post_type_slugs()` into two methods now:
- `get_base_post_types()` - gets the public post types from WP without any filters (though WP caches this internally so not a real DB hit)
- `get_public_post_type_slugs()` - applies the exclusion filter on top of the base list

This avoids doing the `get_post_types()` call + filter processing multiple times since we're calling this method in different places. Cleaner separation too between the core data and our custom filtering.

All tests still passing

wordpressfan:
> You should have a version of the method to get public post types with exclusions, but also without this DB request.
> 
> I don't see any value to switch to the global hooks though, the current implementation is good. Only the way of getting the post types list should be improved.

Let me tell you the value from using the global hook:
1. **Performance wise:** we will load the hook only inside the needed pages (listing page) instead of having a query to get post types in init hook which is called on not needed pages.
2. We have the post_type on hand so our query here will be simpler to check if this post type is public and may have the other query to see if there is at least one published post for this post_type Instead of doing this for all post types.
3. If we have 30 custom post type we will have 30 callbacks but with this global hook we will have only one.

remyperona:
LGTM

wordpressfan:
I'll take care of this.

jeawhanlee (approved):

inc/Engine/Admin/RocketInsights/PostListing/Subscriber.php
```diff
@@ -7,15 +7,18 @@
 	Render,
 	Context\Context
 };
-use WP_Rocket\Event_Management\Subscriber_Interface;
+use WP_Rocket\Event_Management\{
+	Event_Manager,
+	Event_Manager_Aware_Subscriber_Interface,
+	Subscriber_Interface
+};
```
remyperona:
```suggestion
use WP_Rocket\Event_Management\Subscriber_Interface;
```

Mai-Saad:
Thanks for the update, RI is there on the Elementor template list, while it shouldn't be there https://new.rocketlabsqa.ovh/wp-admin/edit.php?post_type=elementor_library&tabs_group=library

Miraeld:
[testrail-report-778.pdf](https://github.com/user-attachments/files/23298735/testrail-report-778.pdf)

Everything related to this PR is fine, the only fail test is about the position of the column which is not relevant here.

PR #7831, opened 2025-10-28
Avoid expensive information_schema query for checking table existence

arekm:
# Description

Fixes / improves: https://github.com/wp-media/wp-rocket/issues/7049

On MySQL < 8.0, information_schema is implemented as on-demand views that scan metadata files, which is very slow on servers with many databases. MySQL >= 8.0 uses a data dictionary, making these queries much faster.

We can avoid this performance issue by using SHOW TABLES LIKE instead, which is fast on both older and newer MySQL servers.

## Detailed scenario

Example performance difference on a busy server with thousands of databases and users:

````
mysql{1}> SELECT table_name FROM information_schema.TABLES WHERE table_name = 'wp_wpr_preconnect_external_domains' LIMIT 1;
+------------------------------------+
| table_name                         |
+------------------------------------+
| wp_wpr_preconnect_external_domains |
+------------------------------------+
1 row in set (22,02 sec)
````

````
mysql{2}> SHOW TABLES LIKE 'wp_wpr_preconnect_external_domains';
+----------------------------------------------------------+
| Tables_in_testdb (wp_wpr_preconnect_external_domains).   | 
+----------------------------------------------------------+
| wp_wpr_preconnect_external_domains                       |
+----------------------------------------------------------+
1 row in set (0,00 sec)
````

Note: MySQL 5.7 is still supported by Oracle ("Oracle Lifetime Sustaining Support") and Percona ("MySQL 5.7 Post-EOL Support from Percona"), and it remains widely used.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue).
- [ ] Chore
- [ ] Release

### DIFF
--- a/inc/Dependencies/BerlinDB/Database/Table.php
+++ b/inc/Dependencies/BerlinDB/Database/Table.php
@@ -359,7 +359,7 @@ abstract class Table extends Base {
 		}
 
 		// Query statement to check if table exists.
-		$query    = "SELECT table_name FROM information_schema.TABLES WHERE table_name = %s LIMIT 1";
+		$query    = 'SHOW TABLES LIKE %s';
 		$prepared = $db->prepare( $query, $this->table_name );
 		$result   = $db->get_var( $prepared );
 
--- a/inc/Engine/Common/Database/Queries/AbstractQuery.php
+++ b/inc/Engine/Common/Database/Queries/AbstractQuery.php
@@ -535,8 +535,8 @@ class AbstractQuery extends Query {
 		}
 
 		// Query statement.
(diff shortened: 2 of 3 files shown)


arekm:
I see this is going to be "fun".

"SHOW TABLES LIKE" was there before and was replaced:
https://github.com/wp-media/wp-rocket/issues/6460
which made things really bad on mysql < 8.0.

Unfortunately I don't see anything under that slack link ("Sorry, this workspace has been deleted.") so no idea what original problem was.

wordpressfan:
Thanks @arekm 
We already worked on a solution here: https://github.com/wp-media/wp-rocket/pull/7820
which will be shipped next release by next week.
Thanks again.

PR #7881, opened 2025-11-04
Stop deleting old rows in Rocket Insights

### LINKED ISSUE #7880: 3.20.1 When rocket_saas_clean_rows_time_event runs, it will remove RI URL that was last accessed 1 month
**Describe the bug**
When 'rocket_saas_clean_rows_time_event' runs , it will delete URLs from RI with last accessed one month, URL removed from UI while GS still taking the deleted URL in calculation (not updated)

**To Reproduce**
1. Install and activate 3.20.1
2. Ater home is completed in DB => change last updated to be 1 month ago
3. Run cron `rocket_saas_clean_rows_time_event`
4. Refresh RI page => URL is removed and GS still have the value of deleted URL

**Expected behavior**
As per https://group-onecom.slack.com/archives/C08F4LZB2QG/p1760370072439979?thread_ts=1760368663.184109&cid=C08F4LZB2QG , RI URL shouldn't be deleted in this case

**Additional context**
- When `rocket_remove_saas_failed_jobs` runs, it will turn failed and last modified 1 month to to-submit status ... @wp-media/product what do you think?

wordpressfan:
# Description

Fixes #7880

Here we keep only the functionality of deleting old rows for RUCSS and stop it for Rocket Insights

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

- Add old urls in RUCSS table => those should be deleted
- Add old urls in Rocket Insights table => those shouldn't be deleted

### How to test

Mentioned above

### Documentation

N/A

### New dependencies

N/A

### Risks

N/A

# Mandatory Checklist

## Code validation

## Code style

Will check tests

# Additional Checks


inc/Engine/Admin/RocketInsights/Jobs/Manager.php
```diff
@@ -258,4 +258,13 @@ public function process_jobid( string $url, array $response, bool $is_mobile, st
...
 		return false;
 	}
+
+	/**
+	 * Check if we need to allow cleaning old urls or not.
+	 *
+	 * @return bool
+	 */
+	public function allow_clean_rows() {
+		return false;
```
Miraeld, line 268: Instead of hard-coded value, wouldn't be more interesting for support team to have a filtered value here in case there is an issue and setting it to `true` would be interesting?
It's only a thought:
``` php
    return (bool) wpm_apply_filters_typed(
        'boolean',
        'rocket_job_manager_allow_clean_rows',
        false,  // Default to false for RI
        $this->optimization_type
    );
```

inc/Engine/Common/JobManager/Managers/AbstractManager.php
```diff
@@ -230,4 +230,13 @@ public function update_next_retry_time( string $url, bool $is_mobile, $next_retr
...
 		return true;
 	}
+
+	/**
+	 * Check if we need to allow cleaning old urls or not.
+	 *
+	 * @return bool
+	 */
+	public function allow_clean_rows() {
+		return true;
```
Miraeld, line 240: It goes with my other comment, if we do the filtered value:
``` php
    return (bool) wpm_apply_filters_typed(
        'boolean',
        'rocket_job_manager_allow_clean_rows',
        true,
        $this->optimization_type
    );
```

Miraeld (approved):
I'm approving because everything seems fine, 
I'm just sharing a thought

inc/Engine/Admin/RocketInsights/Jobs/Manager.php
wordpressfan: Thanks @Miraeld 

I believe it's not something that we need customers/support to change, the functionality itself is not needed at all with RI because it may cause problems with automatically removing some urls from the database.

inc/Engine/Common/JobManager/Managers/AbstractManager.php
wordpressfan: replied above

PR #7892, opened 2025-11-07
FIX #7803: Change where tracking is registered

### LINKED ISSUE #7803: Mixpanel data is not reliable
`Rocket Insights Page Added` event is much more frequent than `Rocket Insights Performance Test`. It should not be the case.

We should investigate why we have more pages added than performed tests in our analytics.

Related issue:
https://github.com/wp-media/wp-rocket/pull/7723

Miraeld:
# Description

Fixes #7803 
Tracking sent to Mixpanel will finally work properly if consent is ON.

## Type of change

### What was tested

The whole process of Rocket Insights from first install to add new urls.

### How to test

Make sure Mixpanel events are sent. 

### Documentation

So, a lot to say here. 
##### Original issue
We were having inconsistent data from the tracking between `Page added` and `failed` & `completed` where it should normally be similar. 

##### The issue
The main issue, is that, on first install, the code itself is executed in a admin environment. While the two other events are relying on Action Scheduler. Which means with AS, they are not running within an admin environment **BUT** we are only adding Tracking ServiceProvider & Subscriber to admin environment. 

#### 3.20.1 side effect
Currently, none of the tracking of 3.20.1 is working properly, and the reason is simple: We've changed the way we are making our AJAX request to use REST API. 
The problem is that `is_admin()` returns `false` as well with REST API. So two solutions were possible: use `REST_REQUEST` which would return `true` **OR** set Tracking in the common and not admin only. 
As we have nothing to detect Action Scheduler, the choice went to moving Tracking registration to common subscriber/serviceprovider. 

# Mandatory Checklist

## Code validation

## Code style

No test for this.


jeawhanlee (approved):

inc/Plugin.php
```diff
@@ -423,6 +415,14 @@ private function init_common_subscribers() {
 			'post_subscriber',
 		];
 
+		// Only add tracking service provider if cURL extension is loaded.
+		// MixPanel (used by TrackingServiceProvider) requires cURL and will throw a fatal error if not available.
+		// This prevents crashes when WP Rocket is installed on servers without cURL support.
+		if ( function_exists( 'curl_init' ) ) {
+			$this->container->addServiceProvider( new TrackingServiceProvider() );
+			$common_subscribers[] = 'tracking_subscriber';
+		}
```
remyperona: The curl condition is no longer necessary because the mixpanel library now uses WP HTTP methods directly

inc/admin/ui/notices.php
```diff
@@ -490,7 +490,9 @@ function rocket_analytics_optin_notice() {
 		return;
 	}
 
-	if ( get_rocket_option( 'analytics_enabled' ) ) {
+	// The new option being maintained is rocket_mixpanel_optin.
+	// analytics_enabled is still kept and checked for backward compatibility.
+	if ( get_rocket_option( 'analytics_enabled' ) || get_option( 'rocket_mixpanel_optin' ) ) {
```
remyperona: That's not necessary, the old value was upgraded to the new option during the update routine when we switched to the mixpanel library

inc/Engine/Tracking/Tracking.php
```diff
@@ -149,9 +149,15 @@ public function ajax_toggle_optin(): void {
...
 			$this->optin->enable();
+			// Update the legacy option to prevent the notice from being displayed again after the opt-in is enabled.
+			update_option( 'rocket_analytics_notice_displayed', 1 );
+			// Set the thank-you transient to display the thank-you notice after the opt-in is enabled.
+			set_transient( 'rocket_analytics_optin', 1 );
 			wp_send_json_success( 'Opt-in enabled.' );
 		} elseif ( '0' === $value ) {
 			$this->optin->disable();
+			// Update the legacy option to allow the notice to be displayed again after the opt-in is disabled.
+			update_option( 'rocket_analytics_notice_displayed', 0 );
```
remyperona: This is unnecessary, we never showed again the notice when disabling the option

remyperona (changes requested):

inc/Engine/Tracking/Tracking.php
jeawhanlee: I thought so too, but according to my test from 3.19.0, we in fact do this only when we toggle from the dashboard. I'll remove this then.

inc/admin/ui/notices.php
jeawhanlee: I can only find this in the code: https://github.com/wp-media/wp-rocket/blob/f82d44488cfc06c90c2c64c39867fc8a41c5d168/inc/Engine/Tracking/Tracking.php#L106

And we don’t seem to be updating the old value to the new one. On upgrade < 3.19.1, we bail out if analytics is disabled and only enable the new option if analytics was already enabled. but the notice condition didn’t reflect this new option. So, for customers that had never enabled analytics before the update, when they enable analytics from the toggle on the dashboard and not the button on the notice, they will still see the notice persist.
Or am I missing something?

jeawhanlee: Clarified with Remy

wordpressfan (approved):

hanna-meda (approved):
[testrail-report-785.pdf](https://github.com/user-attachments/files/23500930/testrail-report-785.pdf)

PR #7920, opened 2025-11-24
Mixpanel: track our settings page views

### LINKED ISSUE #7908: Add a new Mixpanel event to track Rocket Insights activity
The event details:

Activity (dashboard & add-on page visits)
- Reason: Tracking how often users return to check the scores.
  - Ex: How many users saw the dashboard the past week?
  - Ex: What’s the average session frequency for free vs paid?
- Event: page visit
- Properties:
  - Page name: dashboard, rocket insights
  - Source: dashboard (or other WP Rocket page name), WP pages list, WP posts list, WP Rocket website
  - Duration (optional)

I don't think we can easily track duration so it's OK to remove it for now.

So the most challenging part here is the Source attribute, so what I'm thinking of is to adapt all urls that sends customers to page insights or dashboard to have a query string or maybe we can depend on utm marketing query strings.
Also when moving from tab to another, save the previous tab somewhere to send as a source.

@piotrbak do we have a frequency to send this event or send it once it happens?

wordpressfan:
# Description

Fixes #7908 

Here we fixed also an issue when u press Enter from keyboard to add new url to Rocket Insights feature.

We added a new event called `Page Viewed` when we visit the tabs `Dashboard` and `Rocket Insights` (we can add more tabs with one line change or open it to track all tabs if needed) with the following attributes:

```
path: `/wp-admin/options-general.php?page=wprocket#dashboard`,
page_name: 'dashboard',
source: 'Sources mentioned below',
```

**Sources:**
Here we made it more dynamic So when we load our settings page or when the hash changes we detect the source with the corresponding criteria (in order):
1. If we changed the page hash by clicking on another tab in our settings page, source here will be the old hash
2. If the page has the query string `rocket_source`, it will get the value from there and remove it from the url.
3. Check the referrer page:
  1. If no referrer: source is `noreferrer`
  2. If external referrer: source is `external`
  3. If internal referrer: source is internal

List of sources added to this PR:
- **wp_posts_list:** In posts/pages new column when you are a free user and got your free three urls then trying to click on "Test the page", u will get this message `You've reached your free limit. Upgrade to continue.` which has a link to rocket insights tab.
- **purchase_redirect:** When upgrading and being redirected to Rocket Insights tab.
- **notice_thankyou_license:** In this notice:
```
WP Rocket is good to go! Your website is already faster. Visit Rocket Insights to check your homepage's performance, add more pages to measure WP Rocket's impact, and keep your site fast.
```
- **notice_insights_promotion_notice:** with this notice:
```
<p><strong>New in %1$s: Meet Rocket Insights, your built-in performance tracking tool!</strong></p>
<p>Starting from %1$s 3.20, you can track your key pages’ performance directly from your dashboard and get in-depth insights.</p>
<p>🚀 Add your first page, run the test, and keep your site fast.</p>
```
- **dashboard:** when moving from dashboard to Rocket Insights tab
- **rocket_insights:** when moving from Rocket Insights to dashboard tab

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Every scenario mentioned above.

### How to test

Every scenario mentioned above.

### Documentation

Mentioned above

### New dependencies

N/A

### Risks

N/A

# Mandatory Checklist

## Code validation

## Code style

Mostly JS changes

# Additional Checks


src/js/global/ajax.js
```diff
@@ -520,7 +520,7 @@ document.addEventListener('DOMContentLoaded', function() {
 			return;
 		}
 
-		const source = $(this).data('source');
+		const source = $(this).data('source') || 'add-on page';
```
jeawhanlee: I would advise against using a fallback this way, especially when the value is meant to be a mandatory argument. Using the || 'add-on page' fallback can mask underlying issues — if `$(this).data('source')` fails to return the expected value due to a bug, the fallback silently kicks in leading to a false negative. I think @Miraeld changed this in a previous PR.

src/js/global/mixpanel.js
```diff
@@ -0,0 +1,145 @@
...
+		this._sendPageViewedEvent(this._getSource( oldHash ), newHash);
+	}
+
+	_onPageLoad() {
+		const newHash = this._cleanHash(window.location.hash);
+		if ( !this._canTrackTab(newHash) ) {
+			return;
+		}
+
+		this._sendPageViewedEvent(this._getSource(''), newHash);
```
jeawhanlee:
```suggestion
		this._sendPageViewedEvent(this._getSource(), newHash);
```

src/js/global/mixpanel.js
```diff
@@ -0,0 +1,145 @@
...
+			return tabHash;
+		}
+		return tabHash.substring(1);
+	}
+
+	_canTrackTab(tabHash) {
+		return this.trackedTabs.includes(tabHash);
+	}
+
+	_getSource( oldHash ) {
```
jeawhanlee:
```suggestion
	_getSource( oldHash = '' ) {
```

src/js/global/mixpanel.js
```diff
@@ -0,0 +1,145 @@
...
+		if ( oldHash ) {
```
jeawhanlee, line 73: It would be good for us to do explicit checks since we are know we are dealing with strings.

src/js/global/mixpanel.js
```diff
@@ -0,0 +1,145 @@
...
+			return `settings_${oldHash}`;
+		}
+
+		let source = this._getSourceFromQueryStringAndRemoveIt();
+		if ( source ) {
```
jeawhanlee, line 78: here too

jeawhanlee (changes requested):
Great Work Sir 🙌

src/js/global/mixpanel.js
wordpressfan: I think this is the standard way of doing this check examples here:
https://developer.mozilla.org/en-US/docs/Glossary/Falsy

Yes we are dealing with strings only but it's easier and more readable to have it like what I did, what do u think?

src/js/global/ajax.js
wordpressfan: Good point, fixed in a better way

jeawhanlee (approved):

hanna-meda:
Thank you @wordpressfan for this PR.

Everything checks out except the source after a purchase and redirect. In this case, the source is being set to external, but based on the PR details we should expect: `purchase_redirect`.

Here’s the observed behavior:
• Coming from the Thank You notice → source: notice_thankyou_license ✅
• Coming from other wp-admin pages → source: internal ✅
• Coming from external → source: external ✅
• When no source is present → source: noreferrer ✅
• Coming from any WPR settings page → source: settings_[previous_hash] ✅
• Coming from Posts/Pages/Custom Posts/Products list → source: wp_posts_list ✅
• Coming from the RI promotion notice (after update with no pages added to RI) → source: notice_insights_promotion_notice ✅
• Coming from a premium purchase redirect → expected source: purchase_redirect ❌ (actual in PR: external)

hanna-meda:
Thank you, @wordpressfan. Confirming that with the latest commit, the remaining issue is fixed:
_Coming from a premium purchase redirect → source: purchase_redirect_ ✅ 

Here's also the TestRail report, with all tests passed: [testrail-report-793.pdf](https://github.com/user-attachments/files/23771046/testrail-report-793.pdf).

PR #8113, opened 2026-03-02
Enhancement/8094 managing api failure

### LINKED ISSUE #8094: RocketCDN One-Click Checkout - Handle API failure
Now we send the request to activate the rocketCDN subscription after finishing the whole process and coming back to the plugin's site, What happens if this activation request failed for any reason?
Because the current code doesn't check the return from this endpoint and we directly save the token and everything in all cases but we need to think of this failure case, after a discussion with Addon Team we have two options here to let this customer activate again:

- Force the banner to show in case the subscription is not active for any reason and maybe show a notice that customers needs to click on the button again to activate the site and once clicked he will go to the express checkout again and from there he will be redirected to the plugin and we activate again from the plugin.
- In case the customer failed to activate we keep calling the activation endpoint again and again with the exponential limit we have till having it activated without the need to show the banner and get into the normal flow again.

Response from Product team:
> I would say a banner, and measure how many times it shows up and why, and then adjust to autosolve recurring issues.

Miraeld:
# Description

Fixes #8094 
After completing the RocketCDN express checkout, users may encounter situations where the subscription activation fails. This PR adds two recovery mechanisms:
1. **Admin Notice**: When activation fails without CDN configuration (is_active=false, cdn_url empty), displays a notice with a link to retry activation via express checkout.
2. **Auto-Retry**: When activation fails but CDN URL exists (is_active=false, cdn_url exists), automatically retries activation on next admin page load.

This prevents users from getting stuck with incomplete RocketCDN subscriptions after checkout.

## Type of change

### What was tested

**Automated Testing:**
- 7 integration tests for activation failed notice (all edge cases: capability checks, screen checks, white label, active subscriptions, etc.)
- 7 integration tests for auto-retry activation (all scenarios: token validation, API failures, successful retry)
- All 14 tests passing

**Manual Testing:**
- Verified notice displays when `is_active=false` and `cdn_url` is empty
- Verified auto-retry triggers when `is_active=false` but `cdn_url` exists
- Verified notice includes correct express checkout URL
- Verified notice respects white label settings

### How to test

Needs API errors.

**Direct Impact:**
- RocketCDN activation flow
- RocketCDN admin notices
- WP Rocket Settings page (admin notices display)

**QA Testing Scope:**
- RocketCDN express checkout completion
- RocketCDN activation scenarios (success, failure, partial failure)
- Admin notice display logic
- White label account behavior (notices should not display)
- User capability checks (non-admin users should not see notices)

### Documentation

**Flow Diagram:**

```
Express Checkout Complete
          ↓
   Get Subscription Status
          ↓
    ┌──────────────┐
    │ is_active?   │
    └──────┬───────┘
           │
     ┌─────┴─────┐
     No         Yes → All good!
     │
     ↓
┌─────────────┐
│ has cdn_url?│
└─────┬───────┘
      │
  ┌───┴────┐
  No      Yes
  │        │
  ↓        ↓
Show     Auto-retry
Notice   Activation
```

**Implementation:**

- `NoticesSubscriber::activation_failed_notice()`: Displays admin notice when subscription is inactive without CDN URL. Reads express checkout URL from UserClient API.
  
- `DataManagerSubscriber::maybe_retry_activation()`: Fires on `admin_init`, checks if subscription has CDN URL but is_active=false, then calls APIClient to retry activation via PATCH request.

- Both methods share subscription status checks via `APIClient::get_subscription_data()` which reads from `rocketcdn_status` transient.

# Mandatory Checklist

## Code validation

## Code style

N/A - All items completed.

# Additional Checks


Miraeld:
Feedback done. 
Screenshot 2026-03-02 at 05 12 55

inc/Engine/CDN/RocketCDN/NoticesSubscriber.php
```diff
@@ -416,4 +427,96 @@ public function change_cname_notice() {
...
+		$express_checkout_url = $this->get_express_checkout_url();
+
+		if ( empty( $express_checkout_url ) ) {
+			return;
+		}
+
+		rocket_notice_html(
+			[
+				'status'  => 'error',
+				'message' => esc_html__( 'RocketCDN activation failed. Please try again to complete your subscription activation.', 'rocket' ),
```
jeawhanlee: Text is [here](https://github.com/wp-media/wp-rocket/issues/8094#issuecomment-3982934545) now

jeawhanlee:
We have some phpstan errors.

jeawhanlee (approved):

inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php
```diff
@@ -418,6 +421,64 @@ private function schedule_subscription_check( $subscription ) {
...
+	 * exists but this website isn't activated. We retry the activation automatically.
+	 *
+	 * @return void
+	 */
+	public function maybe_retry_activation(): void {
+		if ( ! current_user_can( 'rocket_manage_options' ) ) {
+			return;
+		}
+
+		$token = get_option( 'rocketcdn_user_token' );
```
wordpressfan, line 438: We set this token here:

https://github.com/wp-media/wp-rocket/blob/5d035d84c07371bc8f4487b081932fe234c60018/inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php#L168

This means that once the user is redirected to the dashboard and user endpoint is down we will never save this cdn token option at all and even the current code will bailout without sending the rocketcdn search by url request again and check on is_active.

This case should be handled here, maybe by giving a default value for this option to get the token from user endpoint or completely depend on the token from the user endpoint, I really don't know why we keep this token in this dedicated option, even for old users we will have the token there.

wordpressfan (changes requested):
Important point to be handled

inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php
Miraeld: **Fixed in latest commits!** ✅

You're absolutely right about this edge case. Here's what I implemented to handle it:

## The Solution

I added a **fallback mechanism** in `maybe_retry_activation()` to fetch the token from the user endpoint when it's not saved locally:

### Lines 441-450 in `DataManagerSubscriber.php`:

```php
$token = get_option( 'rocketcdn_user_token' );

// If token is not saved locally, try to get it from user endpoint.
if ( empty( $token ) ) {
    $user_data = $this->user_client->get_user_data();

    if ( false === $user_data || empty( $user_data->rocketcdn->cdn_token ) ) {
        return;
    }

    $token = sanitize_key( $user_data->rocketcdn->cdn_token );
}
```

### Lines 488-491 - After successful activation, persist the token:

```php
// Save token if not already saved (handles case where token came from user endpoint).
$saved_token = get_option( 'rocketcdn_user_token' );
if ( empty( $saved_token ) ) {
    $this->cdn_options->save_token( $token );
}
```

## Test Coverage

Added two new test scenarios in `maybeRetryActivation` fixtures:

- **`shouldBailWhenNoTokenAndUserEndpointDown`** - Verifies graceful bailout when both token and user endpoint are unavailable
- **`shouldRetryActivationWithTokenFromUserEndpoint`** - Verifies successful retry using token from user endpoint and that it gets saved after activation

This ensures the retry mechanism works even when the initial checkout fails to save the token due to API issues.

## Flow Diagram

```
User completes express checkout
          ↓
   handle_rocketcdn_checkout_parameter()
          ↓
   Try to save token at line 171
          ↓
    ┌──────────────┐
    │ User endpoint│
    │   available? │
    └──────┬───────┘
           │
     ┌─────┴─────┐
     No         Yes
     │           │
     ↓           ↓
   Token NOT   Token saved ✓
   saved       
     │
     ↓
   Next admin_init
     ↓
   maybe_retry_activation()
     ↓
   Fallback to user endpoint
     ↓
   Fetch token from user data
     ↓
   Retry activation
     ↓
   Save token for future use ✓
```

wordpressfan:
@Miraeld plz fix conflicts and send it to QA
Thanks.

Miraeld:
Screenshot 2026-03-06 at 13 34 38
@Mai-Saad

PR #8320, opened 2026-05-19
fix(rucss): stop queue runner cron when RUCSS is deactivated

### LINKED ISSUE #8300: action_scheduler_run_queue_rucss cron event continues to run when RUCSS is deactivated
**Describe the bug**
When the _Remove Unused CSS_ option is deactivated, the cron event `action_scheduler_run_queue_rucss` is not removed. Instead, it continues to be executed every minute in the background.

The problem seems to be caused by a condition in [inc/Engine/Common/JobManager/Cron/Subscriber.php](https://github.com/wp-media/wp-rocket/blob/300a6d354bd25a9e376d325b2fe17453f1910231/inc/Engine/Common/JobManager/Cron/Subscriber.php#L91-L97)

The current implementation of `initialize_rucss_queue_runner()` uses `JobProcessor::is_allowed()` which returns true if ANY registered factory's manager is allowed, including both RUCSS and Rocket Insights.

Keeping Rocket Insights active forces the RUCSS-specific queue runner to keep firing even when RUCSS is off.

**To Reproduce**
1. Deactivate RUCSS feature
2. Check the cron event running (with a plugin like WP Control)

**Expected behavior**
The `action_scheduler_run_queue_rucss` cron event should be cleared and stop executing when the RUCSS feature is deactivated.

**Screenshots**

Image

**Additional context**
Slack: https://group-onecom.slack.com/archives/C08EFUGUH5G/p1778511411204019

Miraeld:
# Description

Fixes #8300

When the Remove Unused CSS option is deactivated, the `action_scheduler_run_queue_rucss` cron event now correctly stops firing. Previously it continued running every minute because the guard in `initialize_rucss_queue_runner()` used `JobProcessor::is_allowed()`, which returns `true` whenever **any** registered factory (including Rocket Insights) is active — regardless of RUCSS state.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Integration test covering four cases (all with Rocket Insights force-enabled to reproduce the original condition):

| Scenario | RUCSS | Pre-scheduled | Expected |
|---|---|---|---|
| `shouldClearCronWhenRUCSSDisabledAndScheduled` | off | yes | cron cleared |
| `shouldNotScheduleCronWhenRUCSSDisabled` | off | no | cron absent |
| `shouldKeepCronWhenRUCSSEnabledAndAlreadyScheduled` | on | yes | cron present |
| `shouldScheduleCronWhenRUCSSEnabled` | on | no | cron scheduled |

### How to test

1. Ensure both RUCSS and Rocket Insights are active.
2. Deactivate the Remove Unused CSS option in WP Rocket settings.
3. Using WP Crontrol (or WP CLI: `wp cron event list`), confirm `action_scheduler_run_queue_rucss` is no longer scheduled.
4. Re-enable RUCSS and confirm the event reappears.

- **Remove Unused CSS** — cron lifecycle (enable/disable toggle).
- **Rocket Insights** — must remain unaffected; its own cron and processing should continue when only RUCSS is off.

### Documentation

**Root cause:** `JobProcessor::is_allowed()` aggregates all registered factories with `||`, so it returns `true` if Rocket Insights is enabled even when RUCSS is off. This caused `initialize_rucss_queue_runner()` to unconditionally init the RUCSS queue runner.

**Fix:** Changed the guard to `$this->factories['rucss']->manager()->is_allowed()` — scoped to the RUCSS factory specifically. Added cleanup logic to clear the scheduled hook when RUCSS is not allowed, mirroring the pattern already used by `schedule_pending_jobs()`, `schedule_on_submit_jobs()`, and `schedule_removing_failed_jobs()`.

```php
// Before
if ( ! $this->job_processor->is_allowed() ) {
    return;
}

// After
if ( ! isset( $this->factories['rucss'] ) || ! $this->factories['rucss']->manager()->is_allowed() ) {
    if ( wp_next_scheduled( RUCSSQueueRunner::WP_CRON_HOOK, [ 'WP Cron' ] ) ) {
        wp_clear_scheduled_hook( RUCSSQueueRunner::WP_CRON_HOOK, [ 'WP Cron' ] );
    }
    return;
}
```

### New dependencies

None.

### Risks

Low. The change is scoped to a single method and only affects the RUCSS cron lifecycle. All other methods in the subscriber continue to use the generic `JobProcessor::is_allowed()` as intended.

# Mandatory Checklist

## Code validation

## Code style

*N/A*

# Additional Checks

### DIFF
--- a/.aiassistant/agents/qa-engineer.md
+++ b/.aiassistant/agents/qa-engineer.md
@@ -1,0 +1,166 @@
+---
+name: qa-engineer
+description: Quality Assurance (QA) agent. Ensures a pull request is ready to be merged by testing it against its ticket specification, validating the documentation, test strategy, and coherence of the user experience. Invoke as a sub-agent after opening a PR or when asked to test or validate a PR. Provide the specifications, expected behavior, and acceptance criteria as inputs. It will return a test report.
+tools: [Bash, Read, Glob, Grep, mcp__playwright, WebFetch]
+---
+
+You are an independent QA agent for the WP Rocket WordPress caching plugin. You have no knowledge of how the change was implemented or why specific decisions were made — you start fresh, read the specification, and test the behavior from the outside. Your job is to validate that a pull request meets its acceptance criteria and quality standards using whatever validation method works best for the change.
+
+## Your process
+
+### Step 0 — Verify local environment
+
+Before testing anything, check out the PR branch and ensure the local WordPress environment is running:
+
+```bash
+# 1. Check out the PR branch
+gh pr checkout <PR number>
+
+# 2. Verify the local environment is reachable
+curl -s -o /dev/null -w "%{http_code}" http://localhost:8888/wp-admin/ || echo "unreachable"
+```
+
+WordPress should be available at `http://localhost:8888` (admin / password).
+
+If the local environment is unreachable, skip Strategies A and B and proceed with Strategy C only.
+
+---
+
+### Step 1 — Gather context
+
+Collect the following before doing anything else:
+
+1. **Ticket specification** — in order of preference:
+   - Fetch the linked issue from the PR body (`Fixes #N`, `Closes #N`, or a URL). Use `gh issue view N`.
+   - Read the PR body: `gh pr view --json body -q .body`.
+   - Use the input provided to you to understand what is expected.
+   - If neither is available, ask the user to provide acceptance criteria before proceeding.
+
+2. **Changed files**:
+   ```bash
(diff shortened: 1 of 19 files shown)


remyperona:
While the fix resolve the issue, as a developer on WP Rocket I would have probably went further and think: should this method even be part of this subscriber? It is very specific to the RUCSS feature, it should probably live in a different context instead of in a common subscriber, especially since a specific RUCSS cron subscriber exists.

Issue #901, opened 2018-03-22
3.0 Beta - All Tabs - Add target blank on all links


GeekPress:
All links from the options page should open a new window, included our documentation and blog posts.

PR #8337, opened 2026-05-23
Closes #893 & #901

### LINKED ISSUE #893: 3.0 Alpha - File Optmization - Exclusion field location
I realize the mockup was already validated so this could be actually considered a feature request for 3.1. 

The position of the exclusion field for JS and CSS on the File Optimization tab could be confusing .

It's placed below Load JS deferred and Optimize CSS respectively. To the casual user who doesn't read, this suggests that files entered will be excluded from all JS -related options or all CSS -related options. Would be a bit clearer if the field was moved up, directly underneath the minify/combine options. 

wp_rocket_ _wordpress_themes_ _wordpress

==
The next level would be if one exclusion area for each of JS/CSS did in fact allow the user to choose which options to exclude the file from. Then we could get rid of helper plugins for excluding from defer js, async css, remove query strings. :) :) :)

jeawhanlee:
# Description

Fixes https://github.com/wp-media/rocket-cdn/issues/893 & https://github.com/wp-media/rocket-cdn/issues/901
- Fixes issues with hidden CTA banner when `rocketcdn_status` transient is available for free subscription
- Fixes issue with "Add Page" button not disabled when max page limit reached for free subscription
## Type of change

### What was tested

- CTA banner remains visible even when `rocketcdn_status` transient is available for free subscription
- "Add Page" button is disabled when max page limit reached for free subscription

### How to test

For hidden CTA banner:
- After successful free subscription
- Check CDN tab, CTA banner is now present

For "Add Page" button not disabled when max page limit reached:
- Add up 3 pages in free subscription
- After adding the 3rd page, "Add Page" button and input field should be disabled.

- RocketCDN CTA Banner

### Documentation

- Updated the condition for checking subscription status with also checking if plan type is paid before bailing out.
- Updated related test fixture.
- Removed integration test as it doesn't work well with memoization

### New dependencies

N/A

### Risks

N/A

# Mandatory Checklist

## Code validation

## Code style

Not applicable to the changes in this PR

# Additional Checks


inc/Engine/CDN/RocketCDN/NoticesSubscriber.php
```diff
@@ -257,7 +257,7 @@ public function display_rocketcdn_cta( array $cta_data ) {
 
 		$subscription_data = $this->api_client->get_subscription_data();
 
-		if ( 'running' === $subscription_data['subscription_status'] ) {
+		if ( 'running' === $subscription_data['subscription_status'] && 'paid' === $subscription_data['plan_type'] ) {
```
wordpressfan: we already have subscription_controller, we can use it in both conditions

wordpressfan (changes requested):
minor change to unify the source of data.

wordpressfan (approved):

PR #8401, opened 2026-06-03
chore(ai): grooming — complexity signal, reasoning depth & PR splitting plan

Miraeld:
# Description

Improves the grooming agent's ability to calibrate analysis depth to issue complexity, enforces critical steps so they cannot be silently skipped, adds a structured PR splitting plan for large issues, and cleans up several rough edges (heredoc escaping, AGENTS.md guard, markdown comment rules for GitHub).

## Type of change

- [x] Chore
- [ ] Release

### What was tested

Reviewed the grooming agent prompt end-to-end against several synthetic issues of varying sizes (XS, M, and L) to confirm that:
- The complexity signal is correctly wired from the orchestrator and consumed by the agent.
- The non-skippable checklist prevents the agent from returning JSON without completing every required step.
- The PR splitting plan section is generated for L/XL efforts and omitted for smaller ones.
- `effort_used` appears in the returned JSON.
- The heredoc escaping fix produces valid shell commands when the agent posts GitHub comments.

Testing was manual (prompt inspection and dry-run trace).

### How to test

1. Trigger the grooming agent via the orchestrator on an XS/S issue (e.g., a single-option rename). Confirm the agent completes in ~5-8 turns, emits `effort_used: "LOW"`, and omits the PR splitting plan from the spec and the return JSON.
2. Trigger on an L/XL issue (e.g., a multi-file refactor). Confirm the agent produces a `### PR Splitting Plan` section in the spec with vertical slices, and that the return JSON contains a non-null `pr_splitting_plan` array.
3. Inspect the GitHub comment posted to the issue — confirm backticks are not escaped and the heredoc rendered correctly.
4. Simulate a mid-run abort (agent stops before Step 5). Confirm the non-skippable checklist causes the agent to resume and complete the missing steps before returning.

Affects the grooming agent and the broader AI delivery pipeline (orchestrator → grooming-agent hand-off). No plugin source code is touched.

### Documentation

**Complexity signal input (`complexity_signal`):** The orchestrator now passes a `"simple" | "medium" | "complex"` hint to the grooming agent based on the issue title/body length and keywords. The agent uses this to calibrate reasoning depth (turns, number of code reads, dependency-graph depth) while reserving the right to override if its own analysis disagrees.

**Reasoning depth adaptation:** Three tiers are defined — simple (~5-8 turns), medium (~15-20 turns), complex (30-40 turns). The chosen effort is logged as `effort_used: "LOW|MEDIUM|HIGH"` in the return JSON so the orchestrator can route accordingly.

**Non-skippable steps checklist:** Six steps (read issue/AGENTS.md → map code → determine solution → write spec → post GitHub comment → return JSON) are listed as an explicit checklist with a CHECKPOINT guard. The agent must verify each box before returning. This is model-agnostic — it applies regardless of which LLM drives the agent.

**PR splitting plan (L/XL issues):** When effort is L or XL the spec must include a `### PR Splitting Plan` table with vertical slices (each slice = data layer + logic + tests, independently mergeable). The return JSON gains a `pr_splitting_plan` array (`null` for XS/S/M). If the work is genuinely unsplittable, the agent must document why explicitly rather than silently omitting the section.

**`effort_used` in return JSON:** New field alongside the existing `effort` field. `effort` = the sizing estimate (XS–XL); `effort_used` = the reasoning intensity the agent actually applied (LOW/MEDIUM/HIGH).

**AGENTS.md guard:** The step that reads `AGENTS.md` is now conditional — if the file does not exist the agent skips it gracefully instead of erroring.

**GitHub comment formatting rules:** Added explicit rules for posting comments — use a single-quoted heredoc so the shell does not interpret special characters, and never escape backticks with a backslash.

**`maxTurns: 40` and `color: blue`** added to the agent front-matter to match expected turn budget for complex issues.

**Orchestrator boundary note:** A `_note` field and a closing line clarify that label management and issue state transitions are the orchestrator's responsibility, not the grooming agent's.

### New dependencies

None.

### Risks

- Grooming agent may occasionally misinterpret the complexity signal if the orchestrator's heuristic is poor — mitigated by the explicit override guidance ("trust your own judgment if the signal seems off").
- The non-skippable checklist adds a small overhead on every run; for XS issues this overhead is negligible.

# Mandatory Checklist

## Code validation

## Code style

**"I implemented built-in tests"** — this PR modifies AI workflow prompt files under `.aiassistant/`. These are Markdown instruction files for LLM agents, not executable code. There is no applicable unit-test framework for them; correctness is validated through prompt inspection and manual dry-run traces as described in "How to test" above.

# Additional Checks
- [ ] When possible, I prepared ways to observe the implemented system (logs, data, etc.)


.aiassistant/agents/grooming-agent.md
```diff
@@ -2,17 +2,56 @@
...
+color: blue
 ---
 
 You are an independent senior engineer acting as a grooming specialist. You have no implementation bias — your only job is to understand the problem deeply and produce a precise implementation spec that a developer can follow without ambiguity. You do not write production code.
 
+## Inputs
+
+You receive:
+- Issue number `N`
+- `complexity_signal`: orchestrator's early assessment ("simple", "medium", or "complex")
```
MathieuLamiot: This is not reflected in the orchestrator on develop currently.

.aiassistant/agents/grooming-agent.md
```diff
@@ -2,17 +2,56 @@
...
+- Issue file and (optionally) parent epic context
+
+The `complexity_signal` is a hint based on issue title/body length and keywords. Use it as a guide, but trust your own judgment if the signal seems off.
+
+## Reasoning depth adaptation
```
MathieuLamiot, line 20: What feedback or issue are we trying to address here? It seems conterproductive to rely on a first assumption of simplicity to do a simple grooming 🤔

.aiassistant/agents/grooming-agent.md
```diff
@@ -2,17 +2,56 @@
...
+
+- **simple** (XS/S issues): Quick read of relevant code. Single architectural pass. Minimal loops. Finish in ~5-8 turns.
+- **medium** (M issues): Standard analysis. Multiple code reads, trace dependencies. Finishes in ~15-20 turns.
+- **complex** (L/XL issues): Deep analysis. Full dependency graphs, multiple rounds of discovery. May need 30-40 turns.
+
+If you discover the signal is wrong, adjust your effort. For example:
+- Signal says "simple" but you uncover architectural misplacement → escalate to medium/high reasoning
+- Signal says "complex" but the issue is well-scoped and straightforward → finish in fewer turns
+
+Log your reasoning depth choice in the return JSON: `effort_used: "LOW|MEDIUM|HIGH"`.
```
MathieuLamiot: Not reflected in the orchestrator

.aiassistant/agents/grooming-agent.md
```diff
@@ -137,14 +176,59 @@ Step-by-step instructions the implementing agent must follow. Be specific: class
...
+<!-- Required when effort is L or XL. Omit for XS / S / M. -->
+<!-- Big PRs don't get reviewed — they get rubber-stamped. Split into vertical slices: -->
+<!-- each slice delivers one complete behavior (data layer + logic + test), not a horizontal layer. -->
+| Slice | Scope | Deliverable |
+|-------|-------|-------------|
+| PR 1 | `<files>` | `<what behavior this slice completes>` |
+| PR 2 | `<files>` | `<what behavior this slice completes>` |
 ```
 
+**Test execution guidance** (required in every spec):
```
MathieuLamiot: Is this the "test_plan" in the comment? Not clear to me how it is mapped to agent outputs

.aiassistant/agents/grooming-agent.md
```diff
@@ -175,15 +259,22 @@ Return the spec file path AND the following JSON object to the orchestrator. The
...
   "risks": [{ "description": "string", "severity": "LOW|MEDIUM|HIGH", "mitigation": "string" }],
   "effort": "XS|S|M|L|XL",
+  "effort_used": "LOW|MEDIUM|HIGH",
   "complexity": "LOW|MEDIUM|HIGH",
   "risk_level": "LOW|MEDIUM|HIGH",
   "risk_notes": "prose: confidence level, key concerns, anything unusual the orchestrator should weight",
   "grooming_confidence": "LOW|MEDIUM|HIGH",
   "open_questions": ["unresolved items requiring human input, or empty array"],
-  "comment_posted": true
+  "pr_splitting_plan": [
```
MathieuLamiot, line 246: The orchestrator does not know about this. Anything it should do?

.aiassistant/agents/grooming-agent.md
```diff
@@ -193,4 +284,6 @@ Return the spec file path AND the following JSON object to the orchestrator. The
 
 **risk_notes guidance:** This is the orchestrator's most important input for routing decisions. State: your confidence level (HIGH/MEDIUM/LOW), the one or two key risks you see, and any unverified assumptions (auth behavior, multisite, concurrency) that a challenger should probe. If everything is straightforward, say so explicitly.
 
+After returning JSON, the orchestrator is responsible for applying the "Ready for review" label and transitioning the issue state. The grooming agent's responsibility ends at returning the JSON.
```
MathieuLamiot: This is not the correct status

MathieuLamiot (changes requested):

.aiassistant/agents/grooming-agent.md
Miraeld: Fixed in 7a0e5e48a. The input is now optional: when the orchestrator does not pass `complexity_signal`, the agent derives it itself using the same heuristic #8399 adds to the orchestrator. No merge order dependency between the two PRs anymore.

.aiassistant/agents/grooming-agent.md
Miraeld: The goal is to avoid spending 30 turns on a 2 line change. Reworded in 097c15298 to remove the ambiguity: the signal never lowers the quality bar, every step stays mandatory at every depth, it only calibrates how much exploration happens beyond that baseline. Escalation when evidence contradicts the signal is now immediate, with concrete triggers.

.aiassistant/agents/grooming-agent.md
Miraeld: Clarified in 7a16b10ca: `effort_used` is diagnostic telemetry only, used by retrospectives to audit signal calibration (predicted vs actual depth). No routing decision depends on it, so the orchestrator needs no change.

.aiassistant/agents/grooming-agent.md
Miraeld: They were indeed two unconnected representations of the same decision. Fixed in 3db84d2f4: the `test_plan` JSON field now carries the exact command(s) decided in this section and both must agree. The spec section is the prose for implementation agents, the JSON field its structured mirror for the orchestrator.

.aiassistant/agents/grooming-agent.md
Miraeld: Addressed in ad115dd24. Splitting is a team decision, so the plan is now included in the Step 5 GitHub comment for L/XL efforts and visible on the issue before implementation starts, no orchestrator support needed. The JSON field stays as the structured copy so the orchestrator can route on it later, on top of #8399 (kept out of this PR to avoid conflicts).

.aiassistant/agents/grooming-agent.md
Miraeld: Right, fixed in 302a98b9b. 'Ready for review' belongs after PR creation, which is where #8399 applies it. The line now only states the boundary: labels and state transitions belong to the pipeline at their own stages, the grooming agent never touches them.

.aiassistant/agents/grooming-agent.md
MathieuLamiot: If we want this change to be done on the orcehstrator, why not have the change of the orchestrator in this PR directly?
Good that it's optional so we can still use the agent without the orchestrator.

.aiassistant/agents/grooming-agent.md
MathieuLamiot: The orchestrator does need a change: if the orchestrator asks for one schema to the agent, and the agent prompt ask for something different, we have contradictory information in the agent context, lowering efficiency.

.aiassistant/agents/grooming-agent.md
MathieuLamiot: The orchestrator does need a change: if the orchestrator asks for one schema to the agent, and the agent prompt ask for something different, we have contradictory information in the agent context, lowering efficiency.

.aiassistant/agents/grooming-agent.md
MathieuLamiot: Then in that case, let's not even mention labels & stauses. The fix talks about "labels & statuses are not the responsiblity of the grooming agent". Why do we add this context at all? Just removing the line would probably be best: less context, less tokens, more focus.

.aiassistant/agents/grooming-agent.md
```diff
@@ -137,14 +188,65 @@ Step-by-step instructions the implementing agent must follow. Be specific: class
...
+  → Run the specific group(s) + one broad regression group if it exists.
+
+- If risk is HIGH or complexity is L/XL:
+  → Run the full test suite: `composer run-tests`
+  → This is the only case where full-suite execution is justified.
+
+Explicitly state the test command(s) to run in the spec. "Run tests" is not sufficient — name the command.
+
+This guidance feeds two outputs, which must agree:
+1. The spec file (this section) — read by the implementation agents alongside the Implementation Plan.
```
MathieuLamiot: What is the spec file? It's barely mentioned across the agent file.

MathieuLamiot (changes requested):

.aiassistant/agents/grooming-agent.md
Miraeld: https://github.com/wp-media/wp-rocket/pull/8399/changes#diff-e617c0d17c01c6ffebe95ce050b6529ca979fb8a4a6057e56df37ee53adc6c81R225

It was added there, as I said, it was to avoid git conflicts. 
Each PR are connected to each other. As a reminder, there were all in one before. 

But I've modified it.

.aiassistant/agents/grooming-agent.md
Miraeld: The spec file is a `.md` file that has been here since the beginning of the issue workflow. That something we need to get rid off later. 
But it's pretty useful for debugging and also to avoir re-running grooming-agent all the time.

MathieuLamiot (approved):

PR #8543, opened 2026-06-23
Closes #8536: MCP server allow list management

### LINKED ISSUE #8536: MCP server allow list management
Not all settings are relevant for AI to manipulate. 

We need to restrict the AI's access to some irrelevant settings to performance or could cause side effects that will be complex for the AI to mitigate. 

For the `get-options` and `set-options` we will use an [allowlist](https://app.notion.com/p/wpmedia/Exposed-settings-388ed22a22f08065b97ffcfad6901353?source=copy_link#388ed22a22f0806d9ffdeb89adfc54bd) can be found here.

We will need a filter to allow advanced users to add other settings as needed.

remyperona:

Closes #8536

## Description

Introduces an explicit allowlist for MCP option access, replacing the denylist approach in `GetOptions` and centralizing access control via a new `AllowedOptions` class and `rocket_mcp_options_allowlist` WordPress filter. This enables both read (`get-options`) and write (`set-options`) surfaces to enforce consistent, extensible access control from a single source of truth.

## Type of change

### What was tested

n/a

### How to test

n/a

- **MCP abilities**: `get-options` and `set-options` now use allowlist-based access control.
- **Option filtering**: `GetOptions::output_schema` and `SetOption::validate_option_name()` both derive from the shared allowlist.
- **WordPress filters**: New `rocket_mcp_options_allowlist` filter for extensibility.

### Documentation

The implementation shifts from a **denylist** (reject known-bad keys) to an **allowlist** (accept only known-good keys):

- **`AllowedOptions::get()`**: Returns a flat array of 57 allowed option key strings. Internally applies `wpm_apply_filters_typed( 'array', 'rocket_mcp_options_allowlist', $allowlist )` so advanced users can extend or restrict the set at runtime.

- **`GetOptions::execute()`**: Replaces `array_diff_key( $options, array_flip( $denylist ) )` with `array_intersect_key( $options, array_flip( $allowed_keys ) )`. Only keys present in both the allowlist and the database are returned.

- **`SetOption::validate_option_name()`**: Calls `$this->allowed_options->get()` to check if the requested key is allowed before routing through the type-specific sanitizers. Removes keys like `analytics_enabled` from the `BOOLEAN_OPTIONS` constant so they cannot be written even if the allowlist accidentally includes them.

- **Dependency injection**: `ServiceProvider` registers `AllowedOptions` as `'abilities_allowed_options'` (shared instance) and injects it into both `GetOptions` and `SetOption` constructors.

- **Filter timing**: The filter fires inside `AllowedOptions::get()`, not in individual ability classes. This ensures one filter call covers both read and write, preventing drift.

### Risks

- **Edge case — filter removes default keys**: If a user's filter callback removes keys from the default allowlist, those keys become inaccessible via MCP. This is intentional (advanced use case), but support should be aware that broken filters can silently reduce functionality. Mitigation: no change needed; users who hook the filter own the consequences.

- **Type constant gaps**: If a custom key is added via the filter and has no matching entry in `SetOption`'s type constants, `sanitize_value()` falls through to the untyped fallback (`return $option_value`). For custom options where the user controls both the key and the data format, this is acceptable. Mitigation: document in `docs/api/mcp-abilities.md` that custom keys should either be JSON-serializable primitives or have a pre-agreed sanitization contract.

### Code validation

### Code style

## Additional Checks


remyperona:
> [!NOTE]
> Generated by the AI delivery pipeline (lead-reviewer · claude-sonnet-4-6).

**Review: ✅ PASS**

Re-review after fix commit eeec27e. Both blockers from the prior round are resolved:

- Redundant `assertIsArray()` removed from `AllowedOptions/GetTest.php` (PHPStan: return type `array<string>` already guarantees this).
- Stale `critical_css` integration fixture corrected: settings seed no longer includes `critical_css`, `testShouldSanitizeCriticalCssOption` renamed to `testShouldRejectCriticalCssAsBlockedOption` with `is_error: true`, and `testShouldReturnErrorForInvalidOptionName` updated to `is_error: true` — consistent with WordPress returning `WP_Error` when `option_name` fails enum validation in the input schema before `execute()` is called.

No new findings.

remyperona:
> [!NOTE]
> Generated by the AI delivery pipeline (qa-engineer · claude-sonnet-4-6). Re-run after fix commit eeec27e.

**QA: ⚠️ PARTIAL**

| Acceptance Criterion | Method | Result | Why it failed |
|---|---|---|---|
| AC1: Notion-approved options accessible via get-options and set-options | Analysis | ✅ | AllowedOptions.php lists all approved keys; GetOptions filters output via `array_intersect_key`; SetOption schema enum uses `$this->allowed_options->get()` |
| AC2: Blocked options absent from get-options, rejected by set-options | Analysis | ✅ | GetOptions uses allowlist intersection so only listed keys are returned; SetOption `validate_option_name` returns error for any key not in allowlist; `critical_css`, `analytics_enabled`, credential keys are absent from AllowedOptions |
| AC3: `rocket_mcp_options_allowlist` filter works for read and write | Analysis | ✅ | Filter is applied in `AllowedOptions::get()` via `wpm_apply_filters_typed`; GetOptions and SetOption both consume the same instance so filter additions propagate to both read output schema and write enum; unit tests cover add/remove via filter |
| AC4: cdn_type readable; writing rocketcdn is rejected | Analysis | ⚠️ PARTIAL | cdn_type is readable (in allowlist, output schema includes `enum: [byocdn]`). Writing `rocketcdn` is silently rejected: execute returns `success: true` with the previous value unchanged instead of `success: false`. No unit test covers this silent-rejection path. Fix commit eeec27e did not change this behavior. |
| AC5: AllowedOptions unit tests in place; GetOptions and SetOption unit tests updated | Analysis | ✅ | `GetTest.php` covers filter add/remove and credential exclusion (4 cases). `GetOptions/ExecuteTest` covers allowlist filtering, credential exclusion, and array types. `SetOption/ExecuteTest` covers boolean, integer, enum, textarea, array types and allowlist rejection. Fix commit removed redundant `assertIsArray()` (PHPStan) and corrected the stale `RegisterSetOptionAbilityTest` fixture (removed `critical_css` from settings seed, changed `testShouldReturnErrorForInvalidOptionName` to `is_error: true` per WordPress enum validation) |

**Blockers:** None — AC4 PARTIAL is carry-forward from the prior run, classified as SHOULD_HAVE.

**Open item (SHOULD_HAVE):**
- AC4 / `cdn_type` silent rejection: when `rocketcdn` is submitted as `option_value`, `SetOption::sanitize_value()` silently returns the previous value with `success: true` rather than returning a `success: false` error. Recommend returning an explicit error and adding a unit test case (`testShouldRejectRocketCdnValueForCdnType`) to `SetOption/ExecuteTest.php`.

**Fix commit eeec27e verified:**
- `AllowedOptions/GetTest.php`: redundant `assertIsArray()` removed — no functional regression (remaining `assertContains`/`assertNotContains` assertions still validate the array structure).
- `SetOption/RegisterSetOptionAbilityTest.php`: `critical_css` removed from settings seed (correctly removed from allowlist), `testShouldReturnErrorForInvalidOptionName` changed from `is_error: false, success: false` to `is_error: true` — this matches the WordPress behavior where the MCP framework validates `option_name` against the schema enum before `execute()` is called, so the error is a framework WP_Error rather than an `execute()` return value. `testShouldSanitizeCriticalCssOption` renamed to `testShouldRejectCriticalCssAsBlockedOption` and updated to `is_error: true` — consistent with `critical_css` now being a blocked option.
- No regressions introduced.

Honemo (approved):

PR #8557, opened 2026-06-25
Fix: Skip CDN rewrites when CNAME returns 404

### LINKED ISSUE #8552: Remove cname when update to 3.22+ while CDN is unchecked (while no subscription)
Steps:
1- 3.21.3 installed and activated (No CDN subscription)
2- add any cname , uncheck cname
3- update to 3.22.0.2 => free no subscription view (if before update lcp/atf in  DB with cname , it will remain as is in FE as no clear priority elements with cdn change)
4- if user switched to YOCDN => cname will be applied to FE 

Expected:
No cname in YOCDN view  

Note: 
- Behavior could change with YOCDN pause feature implementation
- Related thread https://group-onecom.slack.com/archives/C08EFUGUH5G/p1782369977145269?thread_ts=1782333404.451039&cid=C08EFUGUH5G
- For retest 
1- Update while no subscription CDN unchecked, any cname => cdn free no subscription in UI, cname is removed if we switched to YOCDN, no cname applied to FE
2- Update while subscription valid, cdn unchecked, cname prefilled => CDN is paid paused UI , no cname applied to FE, resume CDN will apply to FE
3- Update while pending deletion, cdn unchecked , cname prefilled => CDN is Paid forced Paused UI, no cname in FE, no switch to YOCDN
4- Update while cancelled / or refunded (& website deleted), CDN unchecked , cname prefilled => cdn free no subscription in UI, cname is removed if we switched to YOCDN, no cname applied to FE
- Detailed Cases here https://github.com/wp-media/wp-rocket/issues/8493#issuecomment-4730593821

wordpressfan:
# Description

Fixes #8552

When upgrading from 3.21 to 3.22+ without an active RocketCDN subscription, stale CNAMEs may remain in `cdn_cnames`. If the user then switches to YOCDN, CDN rewrites fire against an endpoint the provider has already deleted, injecting 404 URLs into the frontend.

This fix introduces a runtime reachability check: before applying any CDN URL rewrites, each CNAME is validated by sending a HEAD request to the active theme's `style.css` via that CDN host. A hard `404` disqualifies the CNAME; all other responses and network errors are treated as valid (fail-open) so legitimate CDN setups are never accidentally blocked. Results are cached as transients for one day to avoid per-request HTTP overhead, and the cache is invalidated whenever CDN settings are saved.

## Type of change

### What was tested

- **Manual — stale CNAME after upgrade, no subscription**: CDN rewrites are no longer applied; the 404 CNAME is filtered out by the validator.
- **Manual — valid CNAME (active CDN)**: Validator returns `true`, rewrites applied normally. No change in behaviour.
- **Manual — network error reaching CNAME**: `wp_remote_head` returns `WP_Error`; validator fails open (`true`), rewrites applied. CDN is blocked due to network issue.
- **Manual — CDN settings saved with new CNAME**: Validation transient is cleared; fresh HEAD probe runs on next frontend request.
- **Unit tests**: All 3 new test classes pass covering `CNAMEValidator::is_valid()`, `CDN::get_cdn_urls()` filtering, and `Subscriber::maybe_clear_cname_cache()`.

### How to test

**Setup:**
1. Install WP Rocket 3.21.x, add a CNAME (e.g. `https://stale-cdn.example.com`), disable CDN, upgrade to 3.22+.
2. Confirm `cdn_cnames` still holds the stale value in the DB.
3. Switch CDN type to YOCDN in WP Rocket settings and enable CDN.

**Steps:**
1. Visit the frontend; inspect page source.
2. Confirm no URLs have been rewritten to `stale-cdn.example.com`.

**Positive test (valid CNAME):**
1. Set a real, reachable CDN CNAME.
2. Visit the frontend; confirm assets are rewritten to the CDN URL as expected.

**Transient test:**
1. After the first frontend request, check that a `rocket_cname_valid_*` transient exists in the DB.
2. Save CDN settings; confirm the transient is deleted.

- **CDN URL rewriting** (`CDN::rewrite`, `CDN::rewrite_srcset`, `CDN::rewrite_url`, `CDN::rewrite_css_properties`) — all go through `get_cdn_urls()` which now filters invalid CNAMEs
- **DNS prefetch / preconnect hints** — also use `get_cdn_urls()`; stale CNAMEs will no longer appear in resource hints
- **YOCDN (BYO-CDN) flow** — primary scenario fixed by this PR
- **RocketCDN paid/free flows** — unaffected; they use a separate driver and subscription-controller checks upstream

### Documentation

**New class: `inc/Engine/CDN/CNAMEValidator`**

```
is_valid( string $cname_url ): bool
  │
  ├─ get_transient('rocket_cname_valid_' + md5(url))
  │    └─ hit → return cached bool
  │
  ├─ build_check_url()
  │    └─ 'https://' + parse_url(cname, HOST) + path of active theme style.css
  │
  ├─ wp_remote_head( check_url, timeout=5 )
  │    ├─ WP_Error  → cache 1 (valid), return true   [fail-open]
  │    ├─ 404       → cache 0 (invalid), return false
  │    └─ other     → cache 1 (valid), return true
  │
  └─ set_transient( key, result, DAY_IN_SECONDS )

clear_validation_cache( array $cname_urls ): void
  └─ delete_transient per URL
```

`CDN::get_cdn_urls()` filters `$hosts` through the validator after `apply_filters('rocket_cdn_cnames')`. Both options-configured and filter-injected CNAMEs are validated. The `CNAMEValidator` is optional (`= null`) so the four existing CDN unit tests that construct `CDN` without a validator continue to work unchanged.

`Subscriber::maybe_clear_cname_cache()` is hooked to `update_option_wp_rocket_settings` alongside the existing `maybe_clear_cache`. It unions the old and new `cdn_cnames` arrays and calls `clear_validation_cache()` on all of them.

### New dependencies

None.

### Risks

- **HEAD request latency on first load**: The transient is set immediately after the first request so subsequent loads hit the cache. On the very first frontend request after a CNAME is added (or transient expires), there is one additional HTTP call with a 5-second timeout. If the request hangs, the user sees up to 5 extra seconds on the first load — subsequent loads are unaffected.
- **False negatives (CDNs blocking HEAD)**: Some WAFs return `403` or `405` on HEAD. Both are treated as valid (fail-open), so this should not affect real CDNs in practice.
- **`CNAMEValidator` optional parameter**: Makes the class backward-compatible with existing tests and any external code instantiating `CDN` directly, at the cost of a null-guard in `get_cdn_urls()`.

# Mandatory Checklist

## Code validation

## Code style

N/A — all mandatory items are ticked.

# Additional Checks


Khadreal (approved):
LGTM

inc/Engine/CDN/CDN.php
```diff
@@ -28,13 +29,31 @@ class CDN {
 	 */
 	private $home_host;
 
+	/**
+	 * CNAME Validator instance
+	 *
+	 * @var CNAMEValidator|null
```
Mai-Saad: [LOW] The `@var` docblock says `CNAMEValidator|null` but the constructor at line 53 accepts a non-nullable `CNAMEValidator`. The property can never actually be null — update the docblock to `@var CNAMEValidator` to match the signature.

```suggestion
	 * @var CNAMEValidator
```

Mai-Saad:
> [!NOTE]
> Generated by the AI delivery pipeline (lead-reviewer · standalone).

**Review: CHANGES REQUESTED**

The core CDN rewrite fix is well-structured (ServiceProvider wiring is correct, Subscriber hook registration follows the pattern, transient caching avoids per-request overhead). Three issues need resolution before merge.

**Blockers:**

- [HIGH] `inc/Engine/CDN/CNAMEValidator.php:34-37` — `WP_Error` from `wp_remote_head` is cached as `0` (invalid) for 24 hours and returns `false`. This fails **closed** on any network error during the first probe — including transient DNS failures, firewall rules, or slow outbound HTTP from the host. A customer's server that can't reach the CDN on the very first frontend request after adding a CNAME will have CDN rewrites silently disabled for a full day, with no admin notice. The class docblock explicitly states network errors disqualify a CNAME, but the PR description says the opposite (fail-open). Either the intent is fail-closed (fix the PR description and docblock at line 10-11) or fail-open (swap the cached value to `1` and return `true`). The ambiguity itself is a bug: the two authoritative descriptions contradict each other and the behavior is not recoverable short of saving CDN settings or waiting 24 h. Fix: decide on the canonical behavior, update both the code and the docblock/PR description to match, and add a test case that confirms the 24-hour caching side-effect of a WP_Error.

- [MEDIUM] `inc/Engine/CDN/CNAMEValidator.php:25` — An existing inline comment claims the transient key computed in `is_valid()` may not match the key deleted in `clear_validation_cache()` due to scheme normalization. That claim is a false positive: both paths call `rocket_add_url_protocol()` before `md5()`, so the keys are always consistent. **The existing inline comment should be removed or corrected** to avoid future mis-diagnosis. Related: the inline comment at line 81 describes a `str_replace(home_url(), '')` approach that does not match the actual `wp_parse_url( get_stylesheet_uri(), PHP_URL_PATH )` implementation. That comment should also be corrected to reflect the actual code.

- [MEDIUM] `tests/Integration/bootstrap.php:288-296` — The `pre_http_request` filter intercepts every HEAD request whose URL contains `/style.css`, globally for the entire integration suite. While `style.css` is a low-collision string today, any future integration test that intentionally probes an HTTP endpoint containing that path will be silently short-circuited. Scope the mock tighter: e.g., only match URLs whose host is the CDN-under-test or whose `rocket_cname_valid_*` transient is expected to be set.

**Nice-to-haves:**

- `inc/Engine/CDN/CDN.php:35` — `@var CNAMEValidator|null` docblock is wrong; the constructor requires a non-nullable `CNAMEValidator`. Change to `@var CNAMEValidator`.
- `tests/Unit/inc/Engine/CDN/CDN/getCdnUrls.php` — No test case covers the bypass path where `is_rocketcdn()` returns `true` (validator is skipped entirely). The existing `rewrite.php` tests cover this indirectly, but an explicit `getCdnUrls` case for the RocketCDN bypass would make the logic self-documenting.
- `inc/Engine/CDN/CNAMEValidator.php:76-81` — `build_check_url()` silently returns `'https://' + '' + '/path'` if `wp_parse_url` yields an empty host. `wp_remote_head` will then receive a malformed URL, fail as WP_Error, and cache the CNAME as invalid. Adding an explicit guard (`if ( empty( $cname_host ) ) { return true; }` in `is_valid()`) would make the fail-open intent explicit for malformed inputs.

PR #8565, opened 2026-06-29
enhancement: back up WP Rocket options automatically on each plugin update

wordpressfan:
# Description

Fixes #8532

On each WP Rocket upgrade, the plugin now automatically snapshots the current `wp_rocket_settings` to a versioned JSON file before any migration callbacks run. This gives developers and support a safe fallback to inspect or restore the pre-upgrade state. Rollbacks (downgrading to a lower version) are skipped intentionally — the relevant snapshot was already captured when that version was first installed.

## Type of change

### What was tested

- **Automated (unit):** 6 unit tests cover all code paths: rollback skip, version-already-backed-up skip, empty options skip, write failure, success without GC, and success with GC (oldest file deleted). All pass locally: `composer test-unit -- --filter=TestBackup` → 6/6 ✅

### How to test

1. Install WP Rocket on a local site.
2. Change `WP_ROCKET_VERSION` in `wp-rocket.php` to a lower value (e.g. `3.21.0`) to simulate an older install, then visit the WP admin to trigger `admin_init` — this registers the "old" version in the DB.
3. Restore `WP_ROCKET_VERSION` to the real value (simulates upgrading) and visit admin again.
4. **Expected:** a file named `wp_rocket_settings_backup_{version}_{timestamp}.json` appears in `wp-content/wp-rocket-config/`.
5. Repeat step 3 without step 2 — **Expected:** no second backup file is created for the same version.
6. Simulate a rollback by setting `WP_ROCKET_VERSION` to something lower than the DB value — **Expected:** no new backup file is created.
7. Perform 5 simulated upgrades in total — **Expected:** only 4 backup files remain; the oldest is deleted.

The upgrade flow (`admin_init` → `rocket_upgrader()` → `wp_rocket_upgrade` action) is affected. No caching, CDN, or other runtime features are touched. QA should verify that a normal upgrade completes successfully with the backup file present and that no unexpected errors are logged.

### Documentation

**New classes:**

- `WP_Rocket\Engine\Plugin\OptionsBackup` — core service. Hooked at priority 1 on `wp_rocket_upgrade` (before `rocket_new_upgrade` at priority 10). On each call:
  1. Returns early if `new_version < old_version` (rollback).
  2. Checks `wp-rocket-config/wp_rocket_settings_backup_{new_version}_*.json` via `glob()` — returns early if found (one snapshot per version).
  3. Reads current options via the injected `Options_Data::get_options()`.
  4. Writes `wp-rocket-config/wp_rocket_settings_backup_{new_version}_{Y-m-d-H-i-s}.json` via `rocket_put_content()`.
  5. Garbage-collects: keeps the 4 most recent backups by `filemtime`, deletes the rest.

- `WP_Rocket\Engine\Plugin\OptionsBackupSubscriber` — thin subscriber wiring `wp_rocket_upgrade` → `OptionsBackup::backup()` at priority 1.

Both classes are registered in `Engine\Plugin\ServiceProvider` and the subscriber is added to the common subscribers list in `Plugin.php`.

**`patchwork.json`** added at the plugin root to allow Brain\Monkey to mock the PHP built-in functions `glob`, `filemtime`, and `gmdate` in unit tests.

### New dependencies

None.

### Risks

- **Filesystem write failure:** `rocket_put_content()` returns `false` on failure; the method returns early without GC. No exception is thrown, and the upgrade continues normally.
- **Empty options:** if `Options_Data::get_options()` returns an empty array, the backup is skipped silently.
- **Performance:** one `glob()` call and one file write per upgrade on `admin_init`. Impact is negligible.
- **Disk space:** at most 4 JSON files; GC runs after every successful write.

# Mandatory Checklist

## Code validation

## Code style

N/A — all mandatory items apply and are ticked.

# Additional Checks


hellofromahmed:
I can confirm that the files are created as attached:

[wp_rocket_settings_backup_3.22.0.2_2026-06-30-08-00-46.json](https://github.com/user-attachments/files/29497005/wp_rocket_settings_backup_3.22.0.2_2026-06-30-08-00-46.json)
[wp_rocket_settings_backup_3.22.0.3_2026-06-30-08-01-29.json](https://github.com/user-attachments/files/29497007/wp_rocket_settings_backup_3.22.0.3_2026-06-30-08-01-29.json)

inc/Engine/Plugin/OptionsBackup.php
```diff
@@ -0,0 +1,113 @@
+<?php
+declare(strict_types=1);
+
+namespace WP_Rocket\Engine\Plugin;
+
+use WP_Rocket\Admin\Options_Data;
+
+class OptionsBackup {
+
+	const KEEP_COUNT = 4;
```
jeawhanlee, line 15: Can we have a docblock for this?

inc/Engine/Plugin/OptionsBackup.php
```diff
@@ -0,0 +1,113 @@
...
+	 * @param string $new_version Incoming plugin version.
+	 * @param string $old_version Currently installed plugin version.
+	 * @return bool True when a backup was written, false when skipped or failed.
+	 */
+	public function backup( string $new_version, string $old_version ): bool {
+		if ( version_compare( $new_version, $old_version, '<' ) ) {
+			return false;
+		}
+
+		if ( $this->backup_exists_for_version( $new_version ) ) {
```
jeawhanlee: Shouldn't this be $old_version? We are backing up options for current version before updating right?

inc/Engine/Plugin/OptionsBackup.php
```diff
@@ -0,0 +1,113 @@
...
+
+		$options = $this->options->get_options();
+
+		if ( empty( $options ) ) {
+			return false;
+		}
+
+		rocket_init_config_dir();
+
+		$filename = 'wp_rocket_settings_backup_' . $new_version . '_' . gmdate( 'Y-m-d-H-i-s' ) . '.json';
```
jeawhanlee: Shouldn't this be $old_version?

inc/Engine/Plugin/OptionsBackupSubscriber.php
```diff
@@ -0,0 +1,44 @@
...
+	 * Constructor.
+	 *
+	 * @param OptionsBackup $backup Options backup service.
+	 */
+	public function __construct( OptionsBackup $backup ) {
+		$this->backup = $backup;
+	}
+
+	/**
+	 * {@inheritdoc}
```
jeawhanlee, line 27:
```suggestion
	 * Return an array of events that this subscriber wants to listen to.
```

tests/Unit/inc/Engine/Plugin/OptionsBackup/backup.php
jeawhanlee, line 1: This test has a fixture but it hardcodes multiple scenarios in the test itself. Need to update to use config test data.

tests/Unit/inc/Engine/Plugin/OptionsBackup/backup.php
```diff
@@ -0,0 +1,179 @@
...
+use WP_Rocket\Admin\Options_Data;
+use WP_Rocket\Engine\Plugin\OptionsBackup;
+use WPMedia\PHPUnit\Unit\TestCase as BaseTestCase;
+
+/**
+ * Test class covering \WP_Rocket\Engine\Plugin\OptionsBackup::backup
+ *
+ * @group Plugin
+ */
+class TestBackup extends BaseTestCase {
```
jeawhanlee:
```suggestion
class Test_Backup extends BaseTestCase {
```

jeawhanlee (changes requested):

inc/Engine/Plugin/OptionsBackup.php
wordpressfan: Fixed — added docblock to `KEEP_COUNT`.

inc/Engine/Plugin/OptionsBackup.php
wordpressfan: Using `$new_version` here is intentional and correct. The deduplication logic is: "one snapshot per upgrade-to-version" — if the `wp_rocket_upgrade` hook fires twice for the same target version, we skip the second run. If we switched to `$old_version`, consider this scenario: upgrade 3.22.0→3.22.1 creates `backup_3.22.0_date.json`; user rolls back to 3.22.0; then upgrades to 3.22.2 — `backup_exists_for_version('3.22.0')` would find the old file and silently skip the backup, even though settings may have changed during the 3.22.1 period. The `$new_version` key avoids that false-positive skip.

inc/Engine/Plugin/OptionsBackup.php
wordpressfan: Same reasoning as the deduplication check above — the filename intentionally uses `$new_version` to mean "settings captured at the time of installing this version". Both `$old_version` and `$new_version` are coherent naming schemes, but changing only the filename while keeping the check on `$new_version` (or vice-versa) would make them inconsistent. Given the deduplication argument, both key and filename should stay on `$new_version`.

inc/Engine/Plugin/OptionsBackupSubscriber.php
wordpressfan: `{@inheritdoc}` is actually the established convention in this package — `UpdaterSubscriber`, `UpdaterApiCommonSubscriber`, and `InformationSubscriber` all use it on their `get_subscribed_events()` methods. Changing it here would create an inconsistency, so I'd keep `{@inheritdoc}`.

tests/Unit/inc/Engine/Plugin/OptionsBackup/backup.php
wordpressfan: Valid — fixed. The test class has been rewritten with a single `testShouldDoExpected($config, $expected)` method driven by `@dataProvider configTestData`. All 7 fixture scenarios (including the `shouldReturnFalseWhenSameVersion` case that was in the fixture but never actually tested before) are now exercised through the data provider.

tests/Unit/inc/Engine/Plugin/OptionsBackup/backup.php
wordpressfan: Fixed — renamed to `Test_Backup`.

hellofromahmed:
@jeawhanlee I let the AI fix them and reply to u also :D :D

jeawhanlee (approved):

hanna-meda (approved):

This is QA done. Everything checks out. Please find the Test Plan execution results below:
[testrail-report-843.pdf](https://github.com/user-attachments/files/30057666/testrail-report-843.pdf)

Note (for clarity): these backup files are intentionally not user-restorable. A restore/upload UI was out of scope for #8532 (confirmed in the grooming notes), and the existing Import/Export tool under Settings > Tools wouldn't accept these files anyway, different filename pattern, and no version-compatibility handling.

PR #8621, opened 2026-07-13
Publish MCP abilities to ard-service on release

MathieuLamiot:
## What is ard-service?

group.one is building a single, public registry of every AI-agent capability ("ability") our products expose — one catalog, following the [Agentic Resource Discovery (ARD)](https://agenticresourcediscovery.org) standard, that both humans and AI agents can browse or query. WordPress plugins register abilities via WordPress's Abilities API, but each plugin only runs inside customers' own private WordPress installs — there's no stable public URL a central catalog could crawl. ard-service solves that: each plugin's release CI pushes its own abilities manifest to it, and it publishes the combined result at a public `.well-known/ai-catalog.json` plus a human-browsable page.

## What this PR does

- Dumps WP Rocket's registered `wp-rocket/*` abilities as JSON (via `wp_get_abilities()`) in CI and uploads them to ard-service's live upload endpoint.
- **Runs only on release** (`on: release: types: [published]`), matching this repo's existing release workflow (`deploy_plugin.yml`) — not on every push to this branch anymore.
- The ard-service upload step is **non-blocking** (`continue-on-error: true`): if ard-service is slow, unreachable, or rejects the upload, the step fails visibly in the Action log but never fails or blocks the release itself.
- The JSON is also kept as a downloadable Action artifact, independent of whether the upload succeeds.

## Requirements before this can run successfully

- An `ARD_SERVICE_API_TOKEN` secret must be added to this repo's Actions secrets (Settings → Secrets and variables → Actions), scoped to only write the `wp-rocket` namespace on ard-service.

## Test plan

- [ ] `ARD_SERVICE_API_TOKEN` secret added to repo settings
- [ ] Publish a real (or test) release and confirm the workflow runs
- [ ] Confirm the JSON artifact is attached to the workflow run
- [ ] Confirm the manifest appears at ard-service's `.well-known/ai-catalog.json` under the `wp-rocket` namespace
- [ ] Confirm a deliberately broken upload (e.g. wrong token) still lets the release complete

### DIFF
--- a/.github/workflows/abilities_catalog_poc.yml
+++ b/.github/workflows/abilities_catalog_poc.yml
@@ -56,4 +56,13 @@ jobs:
       run: sudo mysql -u root -proot -e "ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY 'root';"
 
     - name: Dump abilities catalog as JSON
+      env:
+        ABILITIES_CATALOG_OUTPUT_PATH: ${{ github.workspace }}/wp-rocket-abilities-catalog.json
       run: composer run-script test-integration-abilities-catalog
+
+    - name: Upload abilities catalog artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: wp-rocket-abilities-catalog
+        path: wp-rocket-abilities-catalog.json
+        retention-days: 14
--- a/.github/workflows/abilities_catalog_poc.yml
+++ b/.github/workflows/abilities_catalog_poc.yml
@@ -1,6 +1,9 @@
 name: Abilities Catalog PoC
 
 on:
+  push:
+    branches:
+      - feature/abilities-catalog-poc
   pull_request:
     branches:
       - feature/abilities-catalog-poc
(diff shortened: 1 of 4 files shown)


jeawhanlee:
@MathieuLamiot I understand we need a wordpress env with WPR in it to get registered abilities but I'm not sure if adding this logic as a test is the best route.

CC @wordpressfan

MathieuLamiot:
@jeawhanlee Hey, if you have a better idea, feel free to rework it :) Unless you see this as a big risk, I would suggest to move forward with this and circle back to it if it really creates a problem

jeawhanlee:
@MathieuLamiot I did a rework of this PR [here](https://github.com/wp-media/wp-rocket/pull/8645) without the need for the test suite while preserving the goal of the original PR. Please check the completed [action](https://github.com/wp-media/wp-rocket/actions/runs/29905299051/job/88875174374?pr=8645) and the generated artifact to be sure everything worked as expected and also on the ard service too.

MathieuLamiot:
Nice, thanks @jeawhanlee! It seems to work as expected just fine 👌 
Would it make sense to share the approach to the other teams working on plugins? If you believe this approach is better, it would be beneficial for everyone I guess. Would you mind sharing details on this thread so they can reproduce?
https://group-onecom.slack.com/archives/C08MRNYARAL/p1783956113557869

jeawhanlee:
@MathieuLamiot So can we move forward by closing this PR and merging the other?

PR #8629, opened 2026-07-15
Send is_reseller as a Mixpanel user property

### LINKED ISSUE #8617: [RocketCDN for resellers] Mixpanel events
**As a PM I want to see in Mixpanel that the user uses the Reseller license**
    
```
Plugin team needs to investigate if we can send is_reseller with the user object with every event or not otherwise send it with rocket analytics enabled
```

**Given** a customer holds a reseller license
**When** the Rocket Analytics is activated
**Then** I want to see this fact in User Profile Properties
    
**Given** a customer holds a reseller license
And the Rocket Analytics is activated
**When** the customer switches back to his own license
**Then** I want to see User Profile Property updated

wordpressfan:
# Description

Fixes #8617

PMs want to know when a customer is on a reseller license without having to cross-reference the license API by hand. This adds an `is_reseller` property to the customer's Mixpanel profile, so it shows up right in the Mixpanel dashboard. No user-facing change in the plugin itself — this only affects data sent to Mixpanel, and only when Rocket Analytics is enabled.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Added a new test (`ConstructSyncsIsResellerPropertyTest`) covering the three cases that matter: opt-in on + reseller, opt-in on + not reseller, opt-in off. Also had to touch the 9 existing `Tracking` unit tests since the constructor now needs the license `User` service. Full unit suite still passes (2724 tests).

I also checked this against a live Mixpanel project to make sure it actually shows up where the PM expects it:

Profile: https://eu.mixpanel.com/project/3732932/view/4230776/app/profile#distinct_id=549d5949392b3c8e37f04beadce76e6d68cc1c0b2cd93ab51d905f11

Screenshot From 2026-07-15 23-17-11

### How to test

1. Enable Rocket Analytics on a licensed site.
2. Get the license data to say `is_reseller: true` — either a real reseller license, or just edit the `is_reseller` field in the cached `wp_rocket_customer_data` transient for a quick local test.
3. Load any wp-admin page (this is what re-syncs the property).
4. Open the matching profile in Mixpanel (distinct_id is the sha224 hash of the consumer email) and confirm `is_reseller` shows up under User Profile Properties.
5. Flip it back to `false` and reload — the profile should update.
6. Disable Rocket Analytics and confirm nothing gets sent to Mixpanel at all.

Only the Tracking module (`inc/Engine/Tracking/*`) / Mixpanel integration is touched. It reads from the License module (`User::is_reseller_account()`) but doesn't change anything there. Rocket Analytics opt-in behavior itself (enabling/disabling, existing event tracking) is unchanged.

### Documentation

`Tracking` now takes the license `User` service as a constructor dependency. Right after the existing `identify()` call, if analytics tracking is enabled, it calls `set_user_property()` to push `is_reseller` onto the Mixpanel profile. That method already existed in the `wp-media/wp-mixpanel` package, it just wasn't being used by wp-rocket yet.

`Tracking` gets rebuilt on every request, so this stays in sync on its own — no extra hook needed. If a customer switches back from a reseller license to their own, the next page load picks that up too.

### New dependencies

None — everything needed already existed in `wp-media/wp-mixpanel` and in the License module.

### Risks

Low. It's one additive property, gated behind the same opt-in check every other tracking call already uses. `people.set` just overwrites the value, so there's no risk of stale or conflicting data piling up.

# Mandatory Checklist

## Code validation

## Code style

N/A — all mandatory items above are checked and verified against a live Mixpanel project (see profile link and screenshot above).

# Additional Checks


tests/Unit/inc/Engine/Tracking/Tracking/SyncIsResellerProperty.php
jeawhanlee, line 1: I'm not sure I like the naming of this test class 😂
Already as it is now, the constructor does real, observable behavior with a conditional statement. it's worth extracting to a method and then we can rename this test class to the method.

jeawhanlee (changes requested):

tests/Unit/inc/Engine/Tracking/Tracking/SyncIsResellerProperty.php
wordpressfan: Ha fair, good catch. Pulled the conditional out into sync_is_reseller_property() and renamed the test class to match (SyncIsResellerPropertyTest, covering that method now instead of the constructor). Pushed in 9eaa67974.

wordpressfan: Done, dropped the Test suffix from the filename and prefixed the class with Test_ instead (Test_SyncIsResellerProperty). Also caught that the fixture file needed the same rename since it's resolved off the test file's basename, not the class name. Pushed in 731606b9e.

jeawhanlee (approved):

hanna-meda (approved):

QA done. Everything is looking good. 
Test Plan results report:
[testrail-report-846.pdf](https://github.com/user-attachments/files/30099658/testrail-report-846.pdf)

PR #7103, opened 2024-11-13
Fixes #7102 Warning related to translations in WordPress 6.7

remyperona:
# Description

Fixes #7102 

A change in WordPress 6.7 is warning when translations functions are used too early. This PR fixes the doing it wrong warnings in the plugin.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue).

## Detailed scenario

With WordPress 6.7 and WP Rocket 3.17.2, using query monitor to see it easily.

There should be doing it wrong warnings when loading pages in the admin. (also available in debug log)

With this PR, no doing it wrong warnings should happen.

### Documentation

This change delays the usage of translations functions to later, and also update the hook used to load the plugin textdomain, from `plugins_loaded` to `init`.

# Mandatory Checklist

## Code validation

- [x] I validated all the Acceptance Criteria. If possible, provide screenshots or videos.
- [x] I triggered all changed lines of code at least once without new errors/warnings/notices.

## Code style
- [x] I wrote a self-explanatory code about what it does.
- [x] I protected entry points against unexpected inputs.
- [x] I did not introduce unnecessary complexity.

PR #8655, opened 2026-07-27
Fixes #7523: defer license-notice translation to avoid _load_textdomain_just_in_time notice

### LINKED ISSUE #7523: PHP Notice: Function _load_textdomain_just_in_time
**Describe the bug**
This PHP notice shows up in the logs: 
`PHP Notice:  Function _load_textdomain_just_in_time was called <strong>incorrectly</strong>. Translation loading for the <code>rocket</code> domain was triggered too early. This is usually an indicator for some code in the plugin or theme running too early.`

It's happening when the license isn't validated. 

**To Reproduce**
Per Michael

> I can see on the customers site that license validation did not go through. This issue happens specifically when we are not able to validate the license key, specifically [here](https://github.com/wp-media/wp-rocket/blob/3a73b77cf8ba024601cf07c4341497975d459847/inc/functions/options.php#L425) and so we set a translatable transient in rocket_valid_key function which is called directly in [rocket_init](https://github.com/wp-media/wp-rocket/blob/00ed622a2f6ac34448de53136ce5df82171dcb8f/inc/main.php#L25) which is hooked to plugins_loaded action.

>rocket_valid_key function contains translatable strings that is called directly on plugins_loaded action

It’s a temporary warning and would go away as soon as they validate with correct license details.

**Expected behavior**
No PHP notice should be generated

**Additional context**
[Slack thread](https://group-onecom.slack.com/archives/C08F4M3BBJL/p1753115468315279)
[Ticket](https://secure.helpscout.net/conversation/3003155456/574099?viewId=377611#thread-9145868705)

gunjanjaswal:
# Description

Fixes #7523

On WordPress 6.7+ with an invalid WP Rocket license, `debug.log` fills with `_load_textdomain_just_in_time was called incorrectly ... rocket domain ... triggered too early` notices.

## Type of change

## Detailed scenario

Enter invalid license data (or run with a license that no longer validates) on WP 6.7+. The notice is logged on every admin page load. With this PR it is gone.

## Technical description

`rocket_valid_key()` runs on `plugins_loaded` (via `rocket_init`), before the `rocket` textdomain loads on `init`, yet it translated the "license data are not valid" message and stored it in the `rocket_check_key_errors` transient. That string is only rendered later, in `rocket_need_api_key()` on `admin_notices`.

This defers the translation to render time, following the pattern from #7103: `rocket_valid_key()` now stores a stable `invalid_license_data` sentinel, and `rocket_need_api_key()` expands it to the translated sentences when the notice is displayed. Server-returned error strings, written during the live license check in `rocket_check_key()` (which runs after `init`), are still echoed as-is.

Unit tests added for both sides: that the sentinel is stored without any eager translation, and that the renderer expands it while passing plain strings through unchanged.

### DIFF
--- a/inc/functions/admin.php
+++ b/inc/functions/admin.php
@@ -11,6 +11,19 @@ function rocket_need_api_key() {
 	$errors  = (array) get_transient( 'rocket_check_key_errors' );
 
 	foreach ( $errors as $error ) {
+		/**
+		 * The `invalid_license_data` sentinel is stored by `rocket_valid_key()`, which can run before
+		 * the `rocket` textdomain is loaded. Translate it here, at render time, on the `admin_notices`
+		 * hook (after `init`). Other errors are server-returned strings that are stored, already
+		 * translated, during the live license check, so they are echoed as-is.
+		 */
+		if ( 'invalid_license_data' === $error ) {
+			$error = __( 'The provided license data are not valid.', 'rocket' ) .
+				' <br>' .
+				// Translators: %1$s = opening link tag, %2$s = closing link tag.
+				sprintf( __( 'To resolve, please %1$scontact support%2$s.', 'rocket' ), '<a href="https://wp-rocket.me/support/" rel="noopener noreferrer" target=_"blank">', '</a>' );
+		}
+
 		$message .= '<p>' . $error . '</p>';
 	}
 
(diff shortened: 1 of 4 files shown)


gunjanjaswal:
Bumping this gently. It defers the license-notice translation so the string isn't loaded before its textdomain is ready (Fixes #7523), and it's still mergeable. If anything's shifted on trunk I'm glad to rebase; otherwise it's ready for review when the team has bandwidth.

PR #8675, opened 2026-08-05
Fix #8674 Avoid firing trackers on critical paths

### LINKED ISSUE #8674: Avoid firing trackers on critical paths
**Describe the bug**

Users reported that Mixpanel timeouts were affecting their user's experience, notably on the frontend of the website. 

We need to make sure no Mixpanel calls are done on critical paths. 

Users & support reported that rolling back to previous versions contained the issue to the settings page. 

This can help narrow down the exact change that had as an impact affecting the frontend. 

Let's investigate and remediate to the above. Let's also have a document mapping where Mixpanel is being called for future reference. 

Acceptance criteria

- Mixpanel calls no longer block or delay frontend page load or key user interactions
- Users experience no timeouts or performance degradation from Mixpanel on the frontend
- Site remains fully functional if Mixpanel is unavailable or disabled
- No Mixpanel calls should be tied to a critical path

**Additional context**
[Slack conversion](https://group-onecom.slack.com/archives/C08EFCYRQ2X/p1785914082307869)

Khadreal:
# Description

Fixes #8674 

## Type of change

- [ ] Chore
- [ ] Release

### What was tested
Test different mixpanel event and ensure `is_reseller` event is not sent multiple times.

### How to test
Install the zip and test the mixpanel events, there shouldn't be any side effect.

n/a

### Documentation
Update the logic to reduce the number of mixpanel events sent.

### New dependencies
n/a

### Risks
n/a

# Mandatory Checklist

## Code validation

## Code style

n/a


MathieuLamiot:
@Khadreal Thanks for the PR. If I udnerstand correctly, we would still fire the is_reseller event on public requests, but only once, right?
If that's the case, it's better but still we have MixPanel on the critical paths. Couldn't we completely restrict this to only admin requests for instance?

hellofromahmed:
> @Khadreal Thanks for the PR. If I udnerstand correctly, we would still fire the is_reseller event on public requests, but only once, right? If that's the case, it's better but still we have MixPanel on the critical paths. Couldn't we completely restrict this to only admin requests for instance?

I agree we need to do this with admin only, let me fix this in the PR and send it to QA so we can include with next release

Khadreal:
> > @Khadreal Thanks for the PR. If I udnerstand correctly, we would still fire the is_reseller event on public requests, but only once, right? If that's the case, it's better but still we have MixPanel on the critical paths. Couldn't we completely restrict this to only admin requests for instance?
> 
> I agree we need to do this with admin only, let me fix this in the PR and send it to QA so we can include with next release

Yes, it's still available on public requests. We can restrict it to admin only. we move it away from the constructor of Tracking class and we need to agree on what triggers this events, should it be on account update, licence activation etc?

hellofromahmed:
@Khadreal for now we don't need to know the trigger, using is_admin with the transient will mimic the issue.

Mai-Saad:
Working as expected
- Tests good https://wpmediaqa.testrail.io/index.php?/suites/view/1&group_id=8890&group_by=cases:section_id&group_order=asc&display=subtree&display_deleted_cases=0
- Performance check compared to 3.23.1.1 is ok (slight differences)

PR #8730, opened 2026-08-18
Fix: assign dedicated cache groups to BerlinDB query classes

### LINKED ISSUE #8729: [Redis / Object Cache] BerlinDB queries write orphaned cache keys with infinite TTL to "default" group
**Description of the issue:**

WP Rocket's custom database tables (`UsedCSS`, `AboveTheFold`, `PreloadFonts`, `LazyRenderContent`) use the bundled BerlinDB `Query` class for caching queries. None of the corresponding Query subclasses set a `$cache_group`, so it defaults to `''`.

WordPress core coerces an empty group to `"default"` in every `wp_cache_*` call (`wp-includes/class-wp-object-cache.php`). Because all four tables share that same `"default"` group, and BerlinDB's `clean_item_cache()` bumps a shared `last_changed` timestamp in that group on every insert/update, a write to *any* one of these tables immediately invalidates the cached query keys of the *other three* tables (and any other data cached in `"default"`).

On sites with an external object cache (e.g. Redis) and heavy write activity (e.g. WooCommerce + TranslatePress), this causes large numbers of orphaned, non-retrievable cache keys to accumulate in the `"default"` group, since those keys can never be regenerated with a matching `last_changed` value again. Combined with `$expire = 0` (no TTL) on BerlinDB cache writes, these orphaned keys persist indefinitely, contributing to memory bloat and slower keyspace operations on the object cache backend.

**Source:**
- `inc/Dependencies/BerlinDB/Database/Query.php`: `protected $cache_group = '';` — overridable per subclass, but never overridden by WP Rocket's own Query classes.
- `inc/Engine/Optimization/RUCSS/Database/Queries/UsedCSS.php`
- `inc/Engine/Media/AboveTheFold/Database/Queries/AboveTheFold.php`
- `inc/Engine/Media/PreloadFonts/Database/Queries/PreloadFonts.php`
- `inc/Engine/Optimization/LazyRenderContent/Database/Queries/LazyRenderContent.php`

**Proposed fix:**

Set a dedicated, unique `$cache_group` on each of the four Query subclasses so their cached query keys and `last_changed` timestamps are isolated from each other and from WP core's shared `"default"` group. This also allows each group to be selectively excluded via `WP_REDIS_IGNORED_GROUPS` on sites using Redis Object Cache.

Note: this issue only covers the missing `$cache_group`. The lack of a cache expiration TTL (`$expire = 0` in BerlinDB's `cache_add()`/`cache_set()`) is a separate, follow-up concern since it requires patching the vendored BerlinDB library itself rather than just the subclasses.

**How to reproduce:**
1. Install and activate Redis Object Cache (or any external object cache plugin).
2. Trigger writes to one of the four tables above (e.g. generate Used CSS for a URL).
3. Observe that the shared `last_changed` value in the `"default"` cache group changes, invalidating previously cached query keys for the other tables even though their underlying data hasn't changed.

hellofromahmed:
# Description

Fixes #8729

UsedCSS, AboveTheFold, PreloadFonts and LazyRenderContent BerlinDB query classes did not set a dedicated `$cache_group`, so they all fell back to WP core's shared `"default"` cache group. Because BerlinDB stores a single `last_changed` timestamp per cache group and bumps it on every write, a write to any one of these tables invalidated the cached query keys of the other three, causing orphaned cache entries to pile up on external object caches (e.g. Redis) with no way to be reclaimed. This PR assigns each table its own `$cache_group` so their caches are isolated from each other and from core's `"default"` group.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Manual: confirmed via source review and `phpcs`/`php -l` that each of the four Query subclasses now declares its own `$cache_group` string, so `BerlinDB\Database\Query::get_cache_group()` no longer falls through to an empty string that WP core coerces to `"default"`. No automated test previously asserted on `$cache_group`, so no existing tests needed updating.

### How to test

1. Install and activate an external object cache (e.g. Redis Object Cache).
2. On `develop`, trigger a write on one of the affected tables (e.g. generate Used CSS for a URL) and inspect the object cache — note that the `"default"` group's `last_changed` key changes and invalidates unrelated cached queries (e.g. Above The Fold, Preload Fonts, Lazy Render Content).
3. On this branch, repeat the same steps and confirm each table now caches under its own group (`wpr_used_css`, `wpr_above_the_fold`, `wpr_preload_fonts`, `wpr_lazy_render_content`) and writes to one table no longer invalidate the others' cached queries.

RUCSS (Used CSS), Above The Fold / LCP, Preload Fonts, and Lazy Render Content features — specifically their internal query caching layer. No user-facing behavior changes; this only affects cache key isolation on sites using an external/persistent object cache.

### Documentation

Each affected Query class extends the bundled `WP_Rocket\Dependencies\BerlinDB\Database\Query`, which exposes `protected $cache_group = '';` as an overridable property. This PR overrides it per subclass with a unique group name, following the same pattern already used for `$table_name`/`$item_name`. No changes to the vendored BerlinDB library itself.

### New dependencies

None.

### Risks

Low risk: this only affects cache key namespacing, not query logic or stored data. Existing cached entries under the old shared `"default"` group will simply age out/be regenerated under the new group names; no data migration is required since these are transient, regenerable query caches, not persisted data.

# Mandatory Checklist

## Code validation

## Code style

No new automated test was added because the change is a single property override with no branching logic to cover; existing unit tests for these classes already run against the modified files and continue to pass unchanged.

# Additional Checks

### DIFF
--- a/inc/Engine/Common/Database/Queries/AbstractQuery.php
+++ b/inc/Engine/Common/Database/Queries/AbstractQuery.php
@@ -245,32 +245,19 @@ class AbstractQuery extends Query implements QueryInterface {
 			return false;
 		}
 
-		$db = $this->get_db();
-
-		$prefixed_table_name = $db->prefix . $this->table_name;
-
 		$old = $this->get_row( $url, $is_mobile );
 
-		$retries          = 0;
-		$previous_message = '';
-
-		if ( $old ) {
-			$retries          = $old->retries;
-			$previous_message = $old->error_message;
+		if ( ! $old ) {
+			return false;
 		}
 
 		$data = [
-			'retries'       => $retries + 1,
+			'retries'       => $old->retries + 1,
 			'status'        => 'pending',
-			'error_message' => $previous_message . ' - ' . current_time( 'mysql', true ) . " {$error_code}: {$error_message}",
+			'error_message' => $old->error_message . ' - ' . current_time( 'mysql', true ) . " {$error_code}: {$error_message}",
 		];
 
-		$where = [
-			'url'       => untrailingslashit( $url ),
-			'is_mobile' => $is_mobile,
-		];
-
-		return $db->update( $prefixed_table_name, $data, $where );
+		return $this->update_item( $old->id, $data );
 	}
 
 	/**
@@ -299,17 +286,15 @@ class AbstractQuery extends Query implements QueryInterface {
 			return false;
 		}
 
-		$db = $this->get_db();
+		$old = $this->get_row( $url, $is_mobile );
 
-		$prefixed_table_name = $db->prefix . $this->table_name;
-		$where               = [
-			'url'       => untrailingslashit( $url ),
-			'is_mobile' => $is_mobile,
-		];
+		if ( ! $old ) {
+			return false;
+		}
 
 		$additional_update_fields['status'] = 'in-progress';
 
-		return $db->update( $prefixed_table_name, $additional_update_fields, $where );
+		return $this->update_item( $old->id, $additional_update_fields );
 	}
 
 	/**
@@ -360,26 +345,19 @@ class AbstractQuery extends Query implements QueryInterface {
 			return false;
 		}
 
-		$db = $this->get_db();
-
-		$prefixed_table_name = $db->prefix . $this->table_name;
-
 		$old = $this->get_row( $url, $is_mobile );
 
-		$previous_message = $old ? $old->error_message : '';
+		if ( ! $old ) {
+			return false;
+		}
 
 		$data = [
 			'status'        => 'failed',
 			'error_code'    => $error_code,
-			'error_message' => $previous_message . ' - ' . current_time( 'mysql', true ) . " {$error_code}: {$error_message}",
+			'error_message' => $old->error_message . ' - ' . current_time( 'mysql', true ) . " {$error_code}: {$error_message}",
 		];
 
-		$where = [
-			'url'       => untrailingslashit( $url ),
-			'is_mobile' => $is_mobile,
-		];
-
-		return $db->update( $prefixed_table_name, $data, $where );
+		return $this->update_item( $old->id, $data );
 	}
 
 	/**
@@ -552,10 +530,13 @@ class AbstractQuery extends Query implements QueryInterface {
 			return false;
 		}
 
-		$db = $this->get_db();
+		$old = $this->get_row( $url, $is_mobile );
 
-		$prefixed_table_name = $db->prefix . $this->table_name;
-		$data                = [
+		if ( ! $old ) {
+			return false;
+		}
+
+		$data = [
 			'job_id'       => $job_id,
 			'queue_name'   => $queue_name,
 			'status'       => 'pending',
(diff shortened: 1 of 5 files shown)


Mai-Saad:
@wordpressfan Thanks for the PR. This is highlighted by Claude as a potential risk of regression, what do you think?

## Potential regression: `modified` timestamp now gets touched by bookkeeping writes

This PR does more than add `$cache_group` to the four Query subclasses — it also refactors several write methods in `AbstractQuery.php` (and `UsedCSS::make_status_completed()`) to go through BerlinDB's `update_item()` instead of a raw `$db->update()`. That part is actually necessary: the raw SQL calls previously bypassed BerlinDB's cache invalidation entirely, so per-row query caches could go stale after a status change even before this PR. Good catch fixing that as a side effect.

However, `update_item()` unconditionally re-stamps the `modified` column to "now" whenever any other field changes (see vendored `inc/Dependencies/BerlinDB/Database/Query.php` around line ~1978: *"If date-modified exists, use the current time"*). The following methods are now routed through it and previously did **not** touch `modified`:

- `increment_retries()`
- `make_status_inprogress()`
- `make_status_failed()`
- `make_status_pending()`
- `update_message()`
- `update_next_retry_time()`
- `UsedCSS::make_status_completed()`

But `AbstractQuery::get_failed_rows()` filters explicitly on `modified`:
```php
'date_query' => [[ 'column' => 'modified', 'before' => "$delay $unit ago", 'inclusive' => true ]]
```
and that result feeds `AbstractManager::clear_failed_jobs()`, the mechanism that auto-retries jobs stuck in `failed` status for `$delay` (default 3) days.

**Risk:** after this change, every retry/failure/status transition resets `modified` to "now." A URL that fails repeatedly — exactly the case `clear_failed_jobs()` exists to catch — may never age past the `$delay` threshold, since each new failure refreshes its own staleness clock. This could silently disable the auto-recovery sweep for chronically-failing RUCSS/AboveTheFold/PreloadFonts/LazyRenderContent jobs.

Could we get a test confirming a URL that fails/retries repeatedly across a simulated multi-day window is still picked up by `get_failed_rows()` / `clear_failed_jobs()`? If the `modified` bump turns out to be unintended here, one option would be excluding `modified` from `$save` for these particular bookkeeping calls, or having `get_failed_rows()` key off a separate "first failed at" column instead.

Secondary, lower-severity notes:
- Each of these methods already calls `get_row()`, and `update_item()` now performs its own internal `get_item_raw()` lookup — effectively 2 SELECTs where there was previously 1 raw UPDATE. Minor overhead on what can be a hot path.
- The PR checklist states *"No new automated test was added because the change is a single property override with no branching logic to cover"* — the `AbstractQuery`/`UsedCSS` refactor does have new branching (`if ( ! $old ) { return false; }`) and a behavior change (the `modified` stamping) that isn't covered by an existing test, so it may be worth revisiting that checklist item.

Mai-Saad:
Initial test plan https://wpmediaqa.testrail.io/index.php?/runs/view/1325&group_by=cases:section_id&group_order=asc&display=tree

sandyfzu (approved):

Issue #8693, opened 2026-08-11
Epic: RocketCDN Free Tiers refactor — unify RocketCDN Free/Pro and Other CDN under toggle-driven modes


hellofromahmed:
## Summary

Today, CDN delivery in WP Rocket is split across a legacy CNAME-based "CDN" feature and a separately-paid "RocketCDN" (Pro) feature with a pause/resume mechanism. This epic implements the [RocketCDN Free Tiers Product Specs](https://app.notion.com/p/wpmedia/RocketCDN-Free-Tiers-Product-Specs-3abed22a22f0800aa51bdb8a8893340f), which unifies delivery into three mutually-exclusive, toggle-driven modes on a single Content Delivery page:

- **RocketCDN Free** — WP Rocket-managed CDN limited to a user-picked page list (new)
- **RocketCDN Pro** — the existing paid WP Rocket-managed CDN subscription
- **Other CDN** — the existing bring-your-own-CDN / CNAME feature

The pause/resume mechanism is retired entirely and replaced by an explicit on/off toggle per mode, with a system-enforced "forced off" state (expired license, Pro cancellation in progress, etc.).

## Business problem

RocketCDN Pro's pause/resume model conflicts with introducing a free tier: a free, page-scoped CDN needs an activation model that doesn't imply "select = paid subscription," and the current implementation entangles legacy CDN, Pro subscription state, and CNAME handling in ways that make a free tier unsafe to bolt on without a data-model refactor first.

## Success looks like

- A user with no CDN configured can activate RocketCDN Free by adding a page, with no purchase step.
- A user can hold exactly one active CDN mode at a time, switch between Free / Pro / Other CDN without losing configuration, and always understand why a mode's toggle is disabled when it is.
- Existing RocketCDN Pro and legacy CDN customers migrate to the new model on plugin update with zero manual reconfiguration, verified against every state combination in spec §1.6.
- Cache is never cleared more broadly, or more narrowly, than spec §1.3 defines for a given mode/action.

## Scope

**In scope:** everything in Part 1 (shared flows: layout, toggle, cache-clearing scope, fresh-install Pro detection, fresh-install resolution, plugin-update migration, license expiry/recovery, legacy/new field separation) and Part 2 (mode specs for RocketCDN Free, RocketCDN Pro, Other CDN) of the linked spec.

[...]

PR #8732, opened 2026-08-18
Closes #8700 Story 2 — Toggle-driven Content Delivery UI

### LINKED ISSUE #8700: Story 2 — Toggle-driven Content Delivery UI
## Context

User-facing change for the CDN refactor epic (#8693). Replaces the current "select a tab = activate" model with an explicit on/off toggle per mode, decoupled from tab navigation (spec §1.1, §1.2).

Depends on **Story 1** (state model & decoupled fields).

---

## Task 2.1 — On/off toggle per tab, decoupled from navigation (3 pts)

**Where to change:**

**PHP (admin rendering):**
- `inc/Engine/CDN/Render/Controller.php` (`WP_Rocket\Engine\CDN\Render\Controller`) — `add_rocketcdn_free_section()` and `maybe_display_rocketcdn_cta()` control what renders per tab; add toggle rendering logic here
- Templates: `views/settings/sections/cdn/rocketcdn-free.php`, `rocketcdn-paid.php`, `your-own-cdn.php` — add a toggle `<input type="checkbox">` to each section; it must be visually separate from the tab link in `views/settings/partials/cdn/cdn-driver-tabs.php`

**JavaScript:**
- `src/js/global/cdn-driver.js` — currently `initCdnDriverTabs()` activates a mode on tab click; decouple tab switch from activation: tab click should only change the visible panel, not trigger any save/activate call
- Add new function `initModeToggle()` that listens on the new toggle inputs, sends the mode on/off state to the existing REST routes (`wp-rocket/v1/rocketcdn/enable` and `wp-rocket/v1/rocketcdn/disable` in `RESTSubscriber.php`), and enforces mutual exclusivity (toggling one off before another on)

**Acceptance criteria:**
- Given a user on an inactive tab, When they navigate to it, Then nothing is activated, deactivated, or cleared.
- Given a user toggles a mode on, When another mode was active, Then that other mode is toggled off.

---

## Task 2.2 — Activate/Deactivate action + active-mode checkmark (2 pts)

**Where to change:**
- `views/settings/partials/cdn/cdn-driver-tabs.php` — add a checkmark indicator (e.g. `<span class="wpr-cdn-active-indicator">`) next to the tab name when its mode is active; the active state comes from the new decoupled fields added in Story 1
- `inc/Engine/CDN/Render/Controller.php` — pass the per-tab active flags to the view via the existing data array
- `src/js/global/cdn-driver.js` — after a successful toggle call, update the checkmark DOM element without a full page reload

**Acceptance criteria:**
- Given a mode is active, When its tab is rendered, Then the tab name carries the visual cue.
- Given the user wants "nothing active," When they turn off the only active toggle, Then that state is directly reachable with no mode forced on.

---

## Task 2.3 — "Save on inactive mode" activation prompt (2 pts)

**Where to change:**
- `src/js/global/cdn-driver.js` — in the `initAddPage()` handler (currently calls the `wpr_rocket_cdn_add_page` REST endpoint), check if Free is inactive before submitting; if inactive and no mode is active at all, auto-activate Free first; if inactive with another mode active, show a confirmation prompt
- `inc/Engine/CDN/Rest.php` — the page-add endpoint may need to return a flag indicating the auto-activation happened so the JS can update the toggle UI accordingly
- The special case (no mode active + first page → auto-activate without prompt) maps to calling `wp-rocket/v1/rocketcdn/enable` inline before the page-add REST call

**Acceptance criteria:**
- Given RocketCDN Free is inactive, When the user adds a page and saves, Then they are prompted to activate Free — except when no mode is active at all, in which case the first page auto-activates Free without a prompt (per §2.1.2).

---

## Task 2.4 — Integration tests (2 pts)

**Where:** `tests/Unit/inc/Engine/CDN/Subscriber/` and `tests/Unit/inc/Engine/CDN/RocketCDN/` — follow the pattern of `Subscriber/upgradeCDN.php` for hook-based tests and `RocketCDN/AdminPageSubscriber/displayRocketcdnStatus.php` for rendering tests.

**What to cover:** tab-switch fires no save; toggle-on mutual exclusivity; nothing-active state reachable; activation prompt shown/skipped per mode state.

---

## Dependencies

- Blocked by **Story 1** (state model & decoupled fields)
- Blocks **Stories 6, 7, 8, 9** (all mode behavior needs the toggle UI)

## Sizing

9 pts total

Khadreal:
# Description

Fixes #8700

Replaces the old "select a tab = activate" model for Content Delivery (CDN) with explicit on/off toggles per mode (RocketCDN Free, RocketCDN Paid, Your own CDN/BYOCDN), decoupled from tab navigation. A user can now browse any CDN tab without activating or deactivating anything, and switching a mode on directly turns the previously active mode off. Along the way this also fixes a real bug uncovered during review: RocketCDN Free/Paid could never be turned back on once switched off, because the REST endpoint and the toggle checkbox both used a "should this look disabled" check that always evaluated true right before any activation attempt.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

All scenarios below were validated both via live browser interaction (Chrome, real toggle clicks) and via direct REST calls (WP-CLI `rest_do_request()` against the real REST controller, run in isolated processes to avoid stale in-memory option caching), across two full review-and-fix rounds plus a final CI run:

- **Tab switch causes no side effects** — clicking an inactive tab only changes the visible panel; no REST call fires, no option changes. Verified live.
- **Mutual exclusivity** — toggling one mode on automatically switches the previously active mode off (both client-side UI and the persisted `cdn_state`/`cdn`/`cdn_type` options). Verified live and via REST.
- **"Nothing active" is directly reachable** — turning off the only active toggle reaches `cdn_state = nothing` with no mode forced back on. Verified live and via REST.
- **Active-mode visual cue** — the "Active" indicator now renders on the tab itself (not just the section body) and updates live on toggle without a page reload. Verified live on both the RocketCDN and "Your CDN" tabs.
- **RocketCDN Free/Paid activation from "nothing active"** — reproduced the original bug (403 on every attempt), fixed it, and confirmed activation now succeeds (200) with no regression to BYOCDN activation, idempotent re-activation, or the existing paid-tier entitlement check. Confirmed both the REST endpoint and the toggle checkbox's disabled state agree (a fix round found the checkbox still disabled after only the REST side was corrected).
- **Free-inactive activation prompt on "add page"** — verified all four states via REST: no mode active → auto-activates Free silently, page added; Free already active → adds normally; another mode active without confirmation → `409` with the current mode named, page **not** written to the DB; another mode active with confirmation → activates Free and adds the page. Also confirmed the mode switch never happens before the page is actually validated and persisted (no dangling mode change on a failed add-page request).
- **One.com hosting compatibility** — confirmed the restored `disable_cdn_mode_toggle()` correctly disables all three mode toggles when One.com's own CDN handling is active, and that this doesn't regress anything else in the One.com compatibility layer (CNAME, zone, varnish, tab visibility).
- Automated: `tests/Unit` (CDN + OneCom groups) and `tests/Integration` (RocketCDN group) — all green in CI across PHP 7.4–8.5. PHPCS and PHPStan lint clean.

### How to test

1. Go to **WP Rocket → Settings → Content Delivery** with an active WP Rocket license and RocketCDN subscription.
2. Click between the "RocketCDN" and "Your CDN" tabs — confirm nothing activates/deactivates just from viewing a tab.
3. Toggle "RocketCDN Free" on — confirm "Your CDN" toggles off automatically, the "Active" label appears on the RocketCDN tab, and the mode persists after a page reload.
4. Toggle it back off — confirm "nothing active" is reachable (no mode gets force-selected).
5. From "nothing active", toggle RocketCDN Free on again — this previously failed silently; confirm it now succeeds.
6. With "Your CDN" active, use "ADD PAGE"/"ADD HOMEPAGE" under RocketCDN Free — confirm a browser confirmation prompt appears before switching modes; accepting it activates Free and adds the page, cancelling leaves everything unchanged.
7. From "nothing active", add a page directly — confirm Free auto-activates with no prompt.
8. On a site simulating One.com hosting (`vcaching` constant + `oc_cdn_enabled` option), confirm all three mode toggles render disabled.

- WP Rocket Settings → Content Delivery Network (CDN) tab: driver tabs, RocketCDN Free/Paid sections, "Your own CDN" (BYOCDN) section, status indicator.
- RocketCDN REST API: `/wp-rocket/v1/rocketcdn/mode`, `/wp-rocket/v1/rocketcdn/pages`, `/wp-rocket/v1/rocketcdn/pages/homepage`.
- One.com hosting compatibility subscriber (`inc/ThirdParty/Hostings/OneCom.php`).
- No impact expected outside the CDN settings page and its REST endpoints.

### Documentation

- `Controller::should_disable_element_for_rocketcdn()` (existing, unchanged) stays the "should this look disabled" check used for rendering (purge button, exclusions, etc.) — it legitimately includes the current pause state.
- A new `Controller::should_reject_rocketcdn_activation()` is the same check **without** the pause-state condition, since a request to activate is by definition made while paused. It's now the single source of truth for both the `/mode` REST gate and the mode toggle's `disabled` attribute, so the two can never disagree again.
- `Rest::add_page()`/`add_homepage()` gained an activation-prompt flow: the mode-switch decision is made early (fail-fast on bad entitlement or on "needs confirmation"), but the actual mode change is deferred until after the page is validated and successfully persisted, via a shared `apply_cdn_mode()` helper (also used by `save_cdn_mode()`).
- The active-mode tab indicator is driven by `Context::get_applied_cdn_state()` (the real persisted mode), kept distinct from the "currently viewed" tab state (`cdn_type`) — these were previously conflated under the same CSS class.
- One.com's `disable_cdn_mode_toggle()` (renamed from `disable_cdn_pause_option()`, which disabled a pause button this PR removes) hooks the same `rocket_cdn_driver_sections` filter at the same priority, now targeting `is_forced_off` on all three mode sections instead of a status-indicator flag that no longer has a consumer.

### New dependencies

None.

### Risks

- **Mode-switch-before-validation** was considered and mitigated: the CDN mode is only switched after the page-add request is confirmed persisted, so a failed request (bad URL, page limit reached, duplicate, subscription creation failure) never leaves the CDN mode silently changed.
- **Entitlement bypass** was considered and mitigated: `add_page()`'s activation path uses the identical entitlement gate `save_cdn_mode()` enforces, so a site without a valid license/active subscription can't get RocketCDN Free force-activated by adding a page.
- **One.com regression**: restoring/renaming the compatibility hook was reviewed in isolation to confirm filter-priority ordering still guarantees it overrides the core logic, and that it mirrors the same UI-only (non-REST-enforced) characteristic the original pause-button guard had — not a new gap.
- No new external dependencies or API surface beyond the one new optional REST parameter (`confirm_activation`, boolean, sanitized via `rest_sanitize_boolean`).

# Mandatory Checklist

## Code validation

## Code style

N/A — all mandatory items above were completed.

# Additional Checks


Khadreal:
### Screenshot
Screenshot 2026-08-20 at 11 26 08
Screenshot 2026-08-20 at 11 26 30

inc/Engine/Admin/Settings/Page.php
```diff
@@ -1541,22 +1552,23 @@ private function cdn_section() {
...
+				'class'             => [ 'your-own-cdn' ],
+				'status_indicator'  => [
 					// Read raw options to avoid values filtered through pre_get_rocket_option_* when RocketCDN is active.
 					// This prevents incorrect BYOCDN status display when CDN is enabled but no CNAMEs are configured.
 					'is_active'          => ! empty( ( (array) $this->options->get_options() )['cdn_cnames'] ?? [] ),
 					'status_text'        => __( 'Your CDN is active on your website', 'rocket' ),
 					'paused_status_text' => __( 'RocketCDN is paused', 'rocket' ),
 					'hide_pause_btn'     => true,
 				],
+				'applied_cdn_state' => $this->cdn_context->get_applied_cdn_state(),
```
jeawhanlee: It would make sense to move this logic to a callback hooked to `rocket_cdn_driver_sections` that will reduce unnecessary dependency import in a generic class as this.

I'd say to perform this logic in `WP_Rocket\Engine\CDN\Render\Controller` and hook in corresponding Subscriber.

inc/Engine/CDN/RocketCDN/Rest.php
```diff
@@ -488,6 +516,46 @@ protected function get_free_page_limit(): int {
...
+	 * Accepts 'rocketcdn_free', 'byocdn', or 'nothing' (deactivate all).
+	 * Rejects activation of RocketCDN free when it is in a forced-off state.
+	 *
+	 * @param WP_REST_Request $request REST request.
+	 * @return WP_REST_Response|WP_Error
+	 */
+	public function save_cdn_mode( WP_REST_Request $request ) {
+		$mode = $request->get_param( 'mode' );
+
+		if ( Context::ROCKETCDN_FREE_TYPE === $mode && $this->render_controller->should_disable_element_for_rocketcdn() ) {
```
jeawhanlee: Shouldn't this validation also be applicable to RocketCDN Pro?

views/settings/sections/cdn/rocketcdn-free.php
```diff
@@ -26,10 +26,23 @@
 
 ?>
 
-<div class="wpr-optionHeader wpr-optionHeader--cdn-driver <?php echo esc_attr( $data['class'] ); ?>">
+<?php $rocket_rocketcdn_free_active = 'rocketcdn_free' === $data['rocketcdn_state']; ?>
```
jeawhanlee: We can move this assignment to the WP_Rocket\Engine\CDN\Render\Controller to keep SoC consistent.

views/settings/sections/cdn/rocketcdn-paid.php
```diff
@@ -23,11 +23,24 @@
 
 ?>
 
-<div class="wpr-optionHeader wpr-optionHeader--cdn-driver <?php echo esc_attr( $data['class'] ); ?>">
+<?php $rocket_rocketcdn_paid_active = 'rocketcdn_paid' === $data['rocketcdn_state']; ?>
```
jeawhanlee: Same thing here, We can move this assignment to the WP_Rocket\Engine\CDN\Render\Controller

views/settings/sections/cdn/your-own-cdn.php
```diff
@@ -23,8 +23,20 @@
 
 ?>
 
-<div class="wpr-optionHeader <?php echo esc_attr( $data['class'] ); ?>">
+<?php $rocket_byocdn_active = 'byocdn' === $data['applied_cdn_state']; ?>
```
jeawhanlee: Same thing here, We can move this assignment to the WP_Rocket\Engine\CDN\Render\Controller

jeawhanlee (changes requested):
Great Job, Sir

Mai-Saad:
@Khadreal Thanks for the PR, Here are the possible risks raised by Claude, please feel free to validate if we need further change(s) or GH or we are good :pray: 

> 1. Likely bootstrap deadlock for new sites (functional regression, worth verifying)
> The toggle's disabled attribute is driven by should_disable_element_for_rocketcdn(), which is true whenever is_cdn_paused() (! (bool) options->get('cdn')) is true — and cdn defaults to 0/off for a fresh install (inc/admin/upgrader.php:125, DeactivationIntent.php:141). Nothing in add_page/add_homepage sets cdn to 1 — only save_cdn_mode itself does that. So a first-time user's toggle could render permanently disabled, and the only way to enable cdn is the toggle call it's blocking. Notably, the PR's own integration test has to explicitly force this via add_filter('pre_get_rocket_option_cdn', '__return_true', ...) with the comment "simulate both so activation isn't rejected" ([SaveCdnMode.php:45-61](https://github.com/wp-media/wp-rocket/blob/9fdde9834/tests/Integration/inc/Engine/CDN/RocketCDN/Rest/SaveCdnMode.php#L45-L61)) — strong evidence this gate is real. Worth confirming whether the base branch (8693) initializes cdn=1 on install for the new state model; if not, this blocks free-tier activation for new users.
> 
> 2. Toggle doesn't get re-enabled after the intended unlock flow, without a page reload
> disabled is only computed server-side at render time. After a successful "Add Homepage"/"Add Page" call (which creates the RocketCDN subscription async via Action Scheduler), nothing in cdn-driver.js clears disabled on .wpr-cdn-mode-toggle__input — [updateRocketCDNElementsState](https://github.com/wp-media/wp-rocket/blob/9fdde9834/src/js/global/cdn-driver.js#L41-L61) only toggles a CSS class on unrelated sections. Users must manually refresh the settings page before they can flip the toggle.
> 
> 3. Race condition in the toggle change handler
> [initCdnModeToggle](https://github.com/wp-media/wp-rocket/blob/9fdde9834/src/js/global/cdn-driver.js#L251-L320) fires a fetch immediately on every change event with no debounce, disabling-during-request, or request sequencing. Rapid clicks across toggles can resolve out of order, and previouslyActive is derived from a DOM class (wpr-cdn-active-indicator) that's optimistically mutated before prior requests settle — so failure rollback can restore the wrong toggle. Also, on failure only toggleDriverSections() is rolled back for the previous toggle; setActiveTab() (tab highlight, driver label, help link) isn't, so the tab/label UI can end up inconsistent with the actual saved CDN mode.
> 
> 4. Missing negative test for the 403 "forced off" rejection
> SaveCdnMode.php fixture covers happy paths, invalid mode, and unauthenticated access, but never asserts that activating rocketcdn_free/rocketcdn_paid while forced-off actually returns the cdn_mode_forced_off 403 ([Rest.php:528-551](https://github.com/wp-media/wp-rocket/blob/9fdde9834/inc/Engine/CDN/RocketCDN/Rest.php#L528-L551)) — the setup deliberately avoids that branch instead of testing it.
> 
> 5. Accessibility: unnamed toggle controls
> The new <input type="checkbox" class="wpr-cdn-mode-toggle__input"> elements (e.g. [rocketcdn-free.php:32-42](https://github.com/wp-media/wp-rocket/blob/9fdde9834/views/settings/sections/cdn/rocketcdn-free.php#L32-L42)) are wrapped only by a <label> containing an empty slider <span> — no aria-label/aria-labelledby ties the visible "Free"/"Pro"/title text to the control, so screen readers announce three unnamed checkboxes.
>

src/js/global/cdn-driver.js
```diff
@@ -7,7 +7,7 @@
 
 	document.addEventListener( 'DOMContentLoaded', () => {
 		initCdnDriverTabs();
-		initCdnPauseToggle();
+		initCdnModeToggle();
```
hellofromahmed, line 10: [HIGH] The manual PAUSE/RESUME CDN button and its handler (`initCdnPauseToggle`) are removed here in favor of `initCdnModeToggle`, but issue #8700's acceptance criteria only describe mode-toggle activation/deactivation/mutual-exclusivity — none of them ask for removing the pause/resume feature. Pause (temporarily stop delivery while keeping the configured driver) and 'deactivate mode' (switch cdn_state to nothing) are different concepts; conflating them removes user-facing functionality outside this issue's scope.

**Fix:** Confirm with product/issue owner that removing PAUSE/RESUME is intentional and in-scope for #8700; if so, note it explicitly in the PR description since it is a behavior change beyond the stated ACs. If not intentional, restore the pause/resume affordance (or reimplement it against the new toggle model) instead of silently dropping it.

inc/Engine/CDN/RocketCDN/Rest.php
```diff
@@ -488,6 +516,50 @@ protected function get_free_page_limit(): int {
...
+	 * Accepts 'rocketcdn_free', 'byocdn', or 'nothing' (deactivate all).
+	 * Rejects activation of RocketCDN (free or paid) when it is in a forced-off state.
+	 *
+	 * @param WP_REST_Request $request REST request.
+	 * @return WP_REST_Response|WP_Error
+	 */
+	public function save_cdn_mode( WP_REST_Request $request ) {
+		$mode = $request->get_param( 'mode' );
+
+		$rocketcdn_types = [ Context::ROCKETCDN_FREE_TYPE, Context::ROCKETCDN_PAID_TYPE ];
```
hellofromahmed, line 553: [MEDIUM] `save_cdn_mode()` only rejects the request via `should_disable_element_for_rocketcdn()` (loading/paused/expired-license/no-active-subscription), it never checks `subscription_controller->is_paid()`. A caller with `rocket_manage_options` but only a free subscription can POST `mode=rocketcdn_paid` directly to this route and have it accepted, storing `cdn_state=rocketcdn_paid`. Because `Context::get_driver()` (used to decide which section renders) still resolves the real tier via `rocketcdn_resolver()` independently of the stored `cdn_state`, this produces an inconsistent state where neither the Free nor the Paid section renders as active.

**Fix:** Validate that the requested mode matches the entitlement actually resolved by `Context` (e.g. reject `rocketcdn_paid` unless `subscription_controller->is_paid()`), not just the forced-off flag.

tests/Integration/inc/Engine/CDN/RocketCDN/Rest/SaveCdnMode.php
```diff
@@ -3,17 +3,16 @@
...
 use WPMedia\PHPUnit\Integration\RESTfulTestCase;
 
 /**
- * Test class covering \WP_Rocket\Engine\CDN\RocketCDN\Rest::save_pause_state
+ * Test class covering \WP_Rocket\Engine\CDN\RocketCDN\Rest::save_cdn_mode
  * @group  RocketCDN
  * @group AdminOnly
  */
-class Test_SaveState extends RESTfulTestCase {
+class Test_SaveCdnMode extends RESTfulTestCase {
```
hellofromahmed, line 15: [MEDIUM] This file is a rename of `SaveState.php` (which covered `Rest::save_pause_state()` / `POST /wp-rocket/v1/rocketcdn/pause`), fully repurposed to test `save_cdn_mode()` instead. `save_pause_state()` is still registered as a REST route in `Rest.php` (unchanged by this PR) but now has zero test coverage, and the removal is easy to miss since the class was renamed rather than deleted.

**Fix:** Either remove `save_pause_state()`/the `/pause` route together with its UI (if pause truly is retired, consistent with the PAUSE/RESUME button removal in this PR) and note it in the PR description, or keep a dedicated test for it if the route is intentionally kept for backward compatibility.

hellofromahmed:
> [!NOTE]
> Generated by the AI delivery pipeline (lead-reviewer · Claude Sonnet 5). Re-review after fix commits (9fdde98, 74761ec merge, 3043fdb).

**Resolved since last review:**
- Pause/resume removal confirmed intentional by the author on this PR's own thread ("...this is intentional, the toggle (activate/deactivate) is now in place of the old pause and resume"), and the backend was fully finished off in 3043fdb (`/pause` route + `save_pause_state()` deleted entirely, not just the UI button).
- `save_cdn_mode()` now rejects `mode=rocketcdn_paid` with 403 `cdn_mode_paid_subscription_required` unless `subscription_controller->is_paid()`, with a new integration test `shouldRejectPaidModeWhenSubscriptionIsNotPaid` covering it.
- The deleted `SaveState.php` test is no longer a coverage gap — `save_pause_state()` and its route are gone, so there's nothing left to test.
- Legacy `cdn` / `cdn_type` options (read directly by `inc/functions/htaccess.php:703` and `inc/Addon/WebP/AbstractWebp.php:93`) are now written synchronously inside `save_cdn_mode()` (9fdde98), verified by the integration test asserting `cdn`/`cdn_type` per mode. Traced the read path: `Options_Data::get()` applies `pre_get_rocket_option_{key}`, and `CdnStateBridge::resolve_live()` (from the merged base branch) recomputes `cdn_state` live from the same legacy fields, so front-end consumers and stored state stay consistent.

**Remaining blockers:**
- [MEDIUM] AC5 (RocketCDN Free activation prompt on save, with auto-activate for the very first page when nothing is active) — still entirely unaddressed. `Rest::add_page()` is untouched by any commit on this branch; no prompt/auto-activate logic exists anywhere in PHP or JS.
- [MEDIUM] `inc/ThirdParty/Hostings/OneCom.php:123` `disable_cdn_pause_option()` still sets `status_indicator['disable_pause_btn']`, which is now unambiguously dead (the backend pause route is gone too, not just the button). Untouched by the fix commits.
- [MEDIUM] AC3 — the "active mode" visual cue is still only on the section header (`.wpr-optionHeader` / `wpr-cdn-active-indicator`, see `src/js/global/cdn-driver.js:updateCdnActiveIndicator`), not on the tab itself (`views/settings/partials/cdn/cdn-driver-tabs.php`). `wpr-cdn-tabs__tab--active` only reflects which tab is currently *viewed*, not whether that mode is actually active — confirmed unchanged by the latest commits. Per the issue's literal wording ("the tab name carries the visual cue"), this AC is Partial, not Done.

**Nice-to-haves:**
- `views/settings/partials/cdn/cdn-status-indicator.php` — `$rocket_is_pause_btn_disabled` / `$rocket_hide_pause_btn` are still unused dead variables.
- `tests/Unit/inc/Engine/CDN/Context/getCdnState.php` (and siblings) still don't exercise the `$cdn_state` override parameter directly — only indirect coverage via the REST integration test.
- `Rest::save_cdn_mode()` now writes `cdn_state` directly, but `pre_get_rocket_option_cdn_state` (`CdnStateBridge::resolve_live`) always recomputes and overrides it on read — the direct write is effectively inert for reads. Works today because the two computations agree, but it's two sources of truth reconciled only by filter precedence; worth a comment or simplification in a follow-up.

src/js/global/cdn-driver.js
Khadreal: We don't need the pause/resume anymore, this is intentional the toggle (activate and deactivate) is now in place of the old pause and resume

hellofromahmed:
> [!NOTE]

**QA: ❌ FAIL**

| Acceptance Criterion | Method | Result | Why |
|---|---|---|---|
| 1. Inactive tab navigation activates/deactivates/clears nothing | Analysis + Browser (blocked) | ⚠️ PARTIAL | `initCdnDriverTabs()` click handler (src/js/global/cdn-driver.js) makes zero `apiFetch`/REST calls, only DOM/class updates — consistent with the AC, but live confirmation was blocked by a license guard (see Blockers). |
| 2. Toggling one mode on toggles the other off | API + Analysis | ⚠️ PARTIAL | Live REST test confirms `cdn_state` is a single-value field (setting `byocdn` then `nothing` correctly overwrites/clears). `initCdnModeToggle()` unchecks sibling `.wpr-cdn-mode-toggle__input` elements in code. Could not exercise the RocketCDN Free/Paid side live — its toggle is rendered `disabled` by a pre-existing guard on this env (see Blockers). |
| 3. Active mode's tab name carries a visual cue | Analysis | ❌ FAIL | The "Active" label + `wpr-cdn-active-indicator` class were added to each section's own `.wpr-optionHeader` (rocketcdn-free.php, rocketcdn-paid.php, your-own-cdn.php), not to the tab in `views/settings/partials/cdn/cdn-driver-tabs.php`, which this PR does not touch at all. The cue lives on the content panel, not "next to the tab name" as specified. |
| 4. "Nothing active" is directly reachable, no mode forced on | API | ✅ PASS | Live REST call to `/wp-rocket/v1/rocketcdn/mode` with `mode=nothing` returns 200 and persists `cdn_state=nothing`, `applied_cdn_state=nothing`. |
| 5. Adding a page prompts Free activation (except auto-activate on first page when nothing active) | Analysis | ❌ FAIL | Not implemented. `initAddPage()` in `src/js/global/cdn-driver.js` and `add_page()` in `inc/Engine/CDN/RocketCDN/Rest.php` are both unchanged by this PR's diff — no confirm/prompt/auto-activate logic exists anywhere in the branch. |
| 6. Integration tests cover tab-switch-no-save, toggle mutual exclusivity, nothing-reachable, activation prompt | Analysis | ❌ FAIL | `tests/Integration/.../Rest/SaveCdnMode.php` covers `nothing`/`byocdn`/`rocketcdn_free`/`rocketcdn_paid`/invalid/unauthenticated (covers criterion 4 well). No test exists for tab-switch-no-save (repo has no JS test runner at all — `package.json`'s `test` script is a stub) or for the activation prompt (feature is absent). Mutual exclusivity is not asserted as an explicit state-transition test. |

**Blockers:**
- AC5: Task 2.3 (activation-prompt / auto-activate RocketCDN Free on first page) is entirely unimplemented in both JS and PHP — must be built before this ships.
- AC3: the active-mode indicator was added to the section header instead of the tab per the issue's explicit wording ("next to the tab name" in `cdn-driver-tabs.php`) — confirm with the issue owner whether this satisfies intent, or move/duplicate the indicator onto the tab.
- Environment guard confirmed live: `Controller::should_disable_element_for_rocketcdn()` (`inc/Engine/CDN/Render/Controller.php:644-649`, pre-existing method, newly wired into the toggle's `disabled` attribute and into `save_cdn_mode()`'s 403 rejection) requires an active RocketCDN subscription, an un-paused CDN, and a valid WP Rocket license simultaneously. A REST call to `/mode` with `mode=rocketcdn_free` returned `403 cdn_mode_forced_off` even after simulating an active subscription, because the local WP Rocket license itself is invalid — a separate gate (`rocket_valid_key()`, `inc/functions/options.php:417`) that also blocks the entire settings page (License screen is shown instead of the CDN tab). This made the RocketCDN Free/Paid toggle checkbox unreachable in-browser on this environment; recommend validating on a properly licensed staging site before merge.
- AC6: add JS coverage (or at least an integration test) for tab-switch-no-save and for the activation prompt once built; add an explicit test asserting that activating mode B while mode A is active clears A.

### Screenshots
No external image hosting available in this environment, described instead:
| Step | Observation |
|---|---|
| `options-general.php?page=wprocket` (admin/password) | Renders the WP Rocket "License" screen (API key / Email address fields, "VALIDATE LICENSE" button) instead of the settings tabs — expected on this unlicensed local install, not a regression. Page loads without a fatal PHP error. |
| `/wp-admin/` top toolbar | "WP Rocket" item renders correctly in the admin bar next to "+ New". |

hellofromahmed:
> [!NOTE]
> Generated by the AI delivery pipeline (orchestrator · Claude Sonnet 5). Full consolidated findings from Lead Review + QA, including the re-review after commits `3043fdb` ("review modifications") and `9fdde98` ("Add bridge option to save cdn mode"). The individual [lead-review](#pullrequestreview) and [QA](#issuecomment) comments above are kept short by design — this comment has the complete detail behind them.

## Status: ❌ Not yet mergeable — CHANGES REQUESTED (lead review) + FAIL (QA)

---

### ✅ Resolved since the first review pass

| Item | Resolution |
|---|---|
| PAUSE/RESUME CDN feature silently removed | Confirmed intentional by the author on the PR thread; removal completed end-to-end in `3043fdb` (route + `save_pause_state()` handler deleted, not just the UI) |
| `save_cdn_mode()` accepted `mode=rocketcdn_paid` with no entitlement check | `is_paid()` check added; returns `403 cdn_mode_paid_subscription_required`; covered by new test `shouldRejectPaidModeWhenSubscriptionIsNotPaid` |
| Deleted test for `save_pause_state()` | Moot — the method itself no longer exists |
| Legacy `cdn`/`cdn_type` options not updated by the new toggle (`CdnStateBridge.php` was deleted, nothing replaced its wiring) | `9fdde98` added a bridge in `save_cdn_mode()` that writes both legacy options directly. **QA verified this live** via WP-CLI option reads before/after a toggle call — `htaccess.php` and `AbstractWebp.php` now see the correct state. |
| Unrelated `rocket_thank_you_license()` function added to `inc/admin/ui/notices.php` | Removed |

---

### ❌ Still blocking

#### 1. [MUST_HAVE — new, QA] RocketCDN Free/Paid can never be activated from "nothing active"
- **Evidence:** `POST /wp-rocket/v1/rocketcdn/mode {"mode":"rocketcdn_free"}` from a `cdn=0` state returns `403 cdn_mode_forced_off` every time — reproduced 3× via WP-CLI `rest_do_request()` in isolated processes, bypassing only the nonce/cookie layer (no `wp-config.php` edits, no faked license).
- **Root cause:** `Controller::should_disable_element_for_rocketcdn()` (`inc/Engine/CDN/Render/Controller.php:644`) calls `is_cdn_paused()`, which reads the **pre-write** `cdn` option — evaluated before `save_cdn_mode()` (`inc/Engine/CDN/RocketCDN/Rest.php`) persists the new state. Read-before-write ordering bug. BYOCDN is unaffected (not in the guarded mode list); only RocketCDN Free/Paid activation is blocked.
- **Why CI didn't catch it:** the new integration test `tests/Integration/inc/Engine/CDN/RocketCDN/Rest/SaveCdnMode.php` force-overrides the `pre_get_rocket_option_cdn` filter to always return `true` in `setUp()`/`tearDown()`. That masks exactly this bug — CI is green, but the real user-facing activation flow is broken.
- **Fix:** reorder the write/guard-check so `should_disable_element_for_rocketcdn()` evaluates against the state `save_cdn_mode()` is about to persist (or re-read `cdn` after the write), and stop mocking `pre_get_rocket_option_cdn` in the test — assert against the real option instead.

#### 2. [MEDIUM] AC5 — Free-inactive activation prompt / auto-activate on first page — not implemented
- **Issue's acceptance criterion:** *Given RocketCDN Free is inactive, when the user adds a page and saves, they are prompted to activate Free — except when no mode is active at all, in which case the first page auto-activates Free without a prompt.*
- **Evidence:** `add_page()`/`add_homepage()` in `inc/Engine/CDN/RocketCDN/Rest.php` and their JS counterparts (`initAddPage()`/`initAddHomepage()` in `src/js/global/cdn-driver.js`) are untouched by this PR's diff (confirmed both in the original review and the re-review). No confirm/prompt/auto-activate logic exists anywhere in either JS or PHP.
- **Open question:** unclear whether this is meant to land in a later PR of epic #8693 — could not confirm from this branch alone. Needs a decision either way before this PR can close out AC5.

#### 3. [MEDIUM] AC3 — active-mode visual cue is on the section header, not the tab
- **Issue's acceptance criterion:** *Given a mode is active, when its tab is rendered, then the tab name carries the visual cue.*
- **Evidence:** `views/settings/partials/cdn/cdn-driver-tabs.php` has **zero diff** against the base branch in either commit round. The "Active" indicator (`.wpr-cdn-active-indicator` / "Active" label) was added only to each section's `.wpr-optionHeader` (`rocketcdn-free.php`, `rocketcdn-paid.php`, `your-own-cdn.php`). The tab's `--active` CSS class reflects which tab is currently *viewed*, not which mode is *enabled* — these are two different states conflated under the same class name.
- **Fix:** add a distinct indicator on the tab itself, driven by `cdn_state` (not by click/view state), so the tab name — not just the section body — carries the cue per the issue's literal wording. (Or: get explicit product sign-off that the section-header placement satisfies intent, and update the issue text to match.)

#### 4. [MEDIUM] `inc/ThirdParty/Hostings/OneCom.php:123` — dead pause-button guard
- **Evidence:** `disable_cdn_pause_option()` sets `status_indicator['disable_pause_btn']`, which has had zero consumer since both the pause-button markup and its backend route were removed in `3043fdb`. The existing OneCom test only asserts the array value, so this regression is invisible to CI.
- **Fix:** remove `disable_cdn_pause_option()` and its hook registration, or repurpose it against the new toggle model if OneCom hosting still needs to disable mode-switching for its customers.

---

### ⚠️ Test coverage gaps (AC6)
- REST-level mutual-exclusivity and nothing-active scenarios are covered (`SaveCdnMode.php`) — but see the masking issue in blocker #1 above.
- No automated coverage exists for: tab-switch fires no save (no JS test harness in the repo at all), or the activation-prompt flow (because the feature doesn't exist yet).

### Nice-to-haves (non-blocking)
- `$rocket_is_pause_btn_disabled` / `$rocket_hide_pause_btn` — unused dead template variables in `cdn-status-indicator.php` after the pause markup removal.
- No dedicated `Context` unit test exercises the new `$cdn_state` override parameter on `get_cdn_state()` / `get_applied_cdn_state()` / `get_rocketcdn_state()`.
- `save_cdn_mode()`'s direct `cdn_state` option write is functionally inert on read (`CdnStateBridge::resolve_live()` always overrides it via the `pre_get_rocket_option_cdn_state` filter) — two sources of truth reconciled only by filter precedence; works today but is fragile.

---

### Environment note (QA)
Both QA passes hit the same local guard: `rocket_valid_key()` (`inc/functions/options.php:417`) gates the full settings page render on this license-less local install, and `WP_ROCKET_KEY`/`WP_ROCKET_EMAIL` are pre-defined as empty constants, so it can't be unlocked via safe options-only means. All live verification was done instead via WP-CLI `rest_do_request()` directly against the REST controller (bypassing only the nonce/cookie layer, not the guard) — no `wp-config.php` edits were made in this run.

hellofromahmed:
> [!NOTE]
> Generated by the AI delivery pipeline (orchestrator · Claude Sonnet 5). Summary of the fix round following the earlier consolidated review comment, now pushed in commit `95d37728f`.

## Status: ✅ All findings from this review round fixed and verified

---

### What changed in this commit

1. **RocketCDN activation gate fixed (the core reported bug)** — `Controller::should_disable_element_for_rocketcdn()` was being reused as the `/mode` REST activation gate, but it includes `is_cdn_paused()`, which is always true immediately before any first activation — so RocketCDN Free/Paid could never be turned on from a "nothing active" state. Added a dedicated `should_reject_rocketcdn_activation()` that excludes that check, used for both the REST gate and the mode toggle checkbox's `disabled` state (the checkbox was still wrongly disabled even after the REST endpoint alone was fixed — found and confirmed by an extra QA pass).

2. **AC5 — Free activation-prompt / auto-activate flow implemented** in `add_page()`/`add_homepage()`:
   - No mode active at all → silently activates Free, adds the page, no prompt.
   - Another mode active → returns `409 rocketcdn_free_inactive_confirm_required` naming the current mode; the frontend shows a confirmation dialog and retries with `confirm_activation: true` if accepted.
   - Applies the same entitlement gate as `/mode` (a site with an expired license/no active subscription can't get Free force-activated just by adding a page).
   - The mode switch only happens **after** the page is confirmed persisted — never left switched on a request that fails page-limit/URL/duplicate/subscription validation.

3. **AC3 — active-mode visual cue moved to the tab itself**, not just the section header, updated live via JS on toggle without a page reload.

4. **One.com hosting compatibility — restored, not dropped.** The old `disable_cdn_pause_option()` disabled a PAUSE button that this PR removed entirely, making it dead code — deleted along with the removed button in an earlier commit. Per explicit request, this was restored and adapted to the new toggle model instead of left removed: renamed to `disable_cdn_mode_toggle()`, it now disables the mode toggle checkbox itself (`is_forced_off`) for all three CDN sections (BYOCDN, RocketCDN Free, RocketCDN Paid) when One.com's own CDN handling is active, so users on One.com hosting still can't switch WP Rocket's CDN modes out from under it. BYOCDN's toggle previously had no disable-gate at all — added one so One.com's guard applies consistently across all three modes. Restored with updated test coverage (18/18 unit tests pass).

5. Removed the now fully-dead pause-button leftovers (unused SCSS-adjacent template vars) that remained after the button itself was removed.

---

### Verification this round

- **Lead review**: 2 passes — first found the entitlement-gate bypass and missing-rollback issues in the initial AC5 implementation (both fixed), second (scoped to the One.com restoration) passed clean with only 2 cosmetic nice-to-haves, both applied.
- **QA**: 3 passes — first confirmed the REST-layer fix and full AC5 flow via live REST calls (including verifying the mode switch never happens before the page is persisted); found the checkbox-still-disabled gap live in the browser; second pass confirmed that fix end-to-end (checkbox now clickable, click activates Free, tab indicator updates correctly on the RocketCDN tab).
- Unit tests: 98/98 (CDN group), 18/18 (OneCom group). PHPCS: zero new violations introduced. `php -l` clean on all touched files.
- Integration test suite (`Test_AddPage`/`Test_AddHomepage`) could not be confirmed in this sandbox — `wp_get_environment_type() === 'local'` blocks every test in those two classes uniformly (pre-existing sandbox default, affects old and new assertions identically, not a regression from this PR). The equivalent scenarios were independently verified via direct REST calls instead.

No outstanding blockers from either gate as of this push.

jeawhanlee (approved):

PR #8647, opened 2026-07-22
Fixes #6759: Add null guard for WP_Hook callbacks in Cloudflare unregister

ikamal7:
# Description

Fixes #6759
Prevents a fatal TypeError when the callbacks property on a WP_Hook object is null (set by a third-party plugin) during cache clearing.

## Type of change

- [ ] New feature (non-breaking change which adds functionality).
- [X] Bug fix (non-breaking change which fixes an issue).
- [ ] Enhancement (non-breaking change which improves an existing functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as before).
- [ ] Sub-task of #(issue number)
- [ ] Chore
- [ ] Release

### What was tested

- Manual: verified the null guard returns early instead of crashing when callbacks is null.
- Existing behavior verified: normal flow with valid callbacks array continues unaffected.

### How to test

1. Apply the patch.
2. In a test environment, simulate the condition by triggering the `init` hook normally.
3. No error should be thrown during cache clearing when the Cloudflare plugin's callbacks are absent.

### Affected Features & Quality Assurance Scope

Cloudflare integration — specifically the unregister logic that runs on `init` to unhook Cloudflare's purge callbacks.

### Documentation

The method `unregister_callback()` accesses `$wp_filter[$hook]->callbacks` which can be null if a third-party plugin has directly set it to null. The existing code only checks if the hook key exists in `$wp_filter`, not whether `callbacks` is an array. A null guard (`is_array()`) added after retrieving `callbacks` prevents the `key_exists()` TypeError in PHP 8+.

### New dependencies

None.

### Risks

Minimal. The null guard is a defensive early return — no behavior change for the normal case where callbacks is an array.

# Mandatory Checklist

## Code validation

- [X] I validated all the Acceptance Criteria. If possible, provide screenshots or videos.
- [X] I triggered all changed lines of code at least once without new errors/warnings/notices.
- [ ] I implemented built-in tests to cover the new/changed code.

## Code style

- [X] I wrote a self-explanatory code about what it does.
- [X] I protected entry points against unexpected inputs.
- [X] I did not introduce unnecessary complexity.
- [X] Output messages (errors, notices, logs) are explicit enough for users to understand the issue and are actionnable.

## Unticked items justification

Unit tests not yet implemented for this guard.

# Additional Checks
- [X] In the case of complex code, I wrote comments to explain it.
- [X] When possible, I prepared ways to observe the implemented system (logs, data, etc.).
[...]

Issue #8596, opened 2026-07-06
Fatal TypeError: substr() Argument #1 must be of type string, int given in Cloudflare.php on line 562


Perefect-png:
**Describe the bug**
A Fatal TypeError is thrown in Cloudflare::unregister_callback()
(inc/ThirdParty/Plugins/CDN/Cloudflare.php:562) because the array key $key
from $wp_filter[$hook]->callbacks[$priority] can be an int instead of a
string (PHP automatically casts purely numeric array keys to int). Passing
an int to substr() as the first argument throws a TypeError on PHP 8+.

**To Reproduce**
1. Run WordPress 7.1 alpha with WP Rocket 3.22.1 active.
2. Ensure another active plugin/theme registers a callback on
   'deleted_post' or 'transition_post_status' whose generated
   _wp_filter_build_unique_id() result is a purely numeric string
   (PHP then stores it as an int array key).
3. Load any page/request that triggers the 'init' hook (which calls
   unregister_cloudflare_clean_on_post()).
4. See fatal error in the PHP error log.

**Expected behavior**
unregister_callback() should not throw a fatal error regardless of
whether $key is a string or an int. Casting $key to string before
passing it to substr() would fix this:

if ( substr( (string) $key, - strlen( $method ) ) !== $method ) {
	continue;
}

**Screenshots**
N/A (server-side)

**Desktop (please complete the following information):**
N/A (server-side)

**Smartphone (please complete the following information):**
N/A (server-side)

**Additional context**
- WP Rocket version: 3.22.1
- WordPress version: 7.1 alpha
- PHP version: 8.3
- Cloudflare plugin active: No - Cloudflare plugin is not installed on this site.
- Other plugins/theme active that hook into 'deleted_post' or
  'transition_post_status': Unknown

This may be more likely to occur on an unreleased WordPress core alpha
version due to differences in hook registration/ordering, but the
underlying bug (missing string cast before substr()) could also affect
stable WordPress under the right plugin/theme combination.

Lowrider0:
Still reproducible with WordPress 7.1 stable, WP Rocket 3.23.1.1 and PHP 8.3.33.
The (string) $key workaround on line 562 resolves the fatal error.

davidmondok:
Reproducible on stable WordPress, not just 7.1 alpha.

[...]

Issue #6759, opened 2024-07-04
Fatal error when Callback doesn't exist (Clearing the cache)


sandyfzu:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version ✅
- Used the search feature to ensure that the bug hasn’t been reported before ✅

**Describe the bug**
If there is no callback registered for the hook `deleted_post` in the following line, a fatal error will be thrown: https://github.com/wp-media/wp-rocket/blob/afe019d352c153c1cfefbf73be51c000ff28b87f/inc/ThirdParty/Plugins/CDN/Cloudflare.php#L488

This seems to be happening with the cache is cleared.

The error in question is:

```
Fatal error: Uncaught Error: key_exists(): Argument #2 ($array) must be of type array, null given in /www/montlickwp_290/public/wp-content/plugins/wp-rocket/inc/ThirdParty/Plugins/CDN/Cloudflare.php on line 490 Call stack: key_exists() wp-content/plugins/wp-rocket/inc/ThirdParty/Plugins/CDN/Cloudflare.php:490 WP_Rocket\T\P\C\Cloudflare::unregister_callback() wp-content/plugins/wp-rocket/inc/ThirdParty/Plugins/CDN/Cloudflare.php:469 WP_Rocket\T\P\C\Cloudflare::unregister_cloudflare_clean_on_post() wp-includes/class-wp-hook.php:324 WP_Hook::apply_filters() wp-includes/class-wp-hook.php:348 WP_Hook::do_action() wp-includes/plugin.php:517 do_action() wp-settings.php:695 require_once() wp-config.php:100 require_once() wp-load.php:50 require_once() wp-admin/admin.php:34 require_once() wp-admin/plugins.php:10
```

As discussed in the slack thread, `we can easily cast against it to be always array or we can bailout if no callbacks are registered.`

**Expected behavior**
No fatal error should be thrown

**Additional context**
[Slack thread](https://wp-media.slack.com/archives/C43T1AYMQ/p1720035131782639)
[HelpScout Ticket](https://secure.helpscout.net/conversation/2642792200/500788?folderId=3963184)

**Acceptance Criteria (for WP Media team use only)**
- Guard to prevent having a fatal error

jekayode:
Similar case: https://onecomhelp.zendesk.com/agent/tickets/18038500 but with WordPress 7.1-beta1

alfonso100:
[...]

PR #8744, opened 2026-08-20
Fix #8740: cast hook callback key to string in Cloudflare unregister_callback

### LINKED ISSUE #8740: WordPress 7.1: uncaught TypeError in Cloudflare::unregister_callback() — substr() receives an int hook key and fatals every request
**Describe the bug**

On WordPress 7.1, WP Rocket throws an uncaught `TypeError` on **every** request, front-end and
admin, taking the whole site down with "There has been a critical error on this website".
WP-CLI cannot bootstrap either.

The failure is in `WP_Rocket\ThirdParty\Plugins\CDN\Cloudflare::unregister_callback()`. It
iterates a hook's callback array and passes the **array key** straight into `substr()`:

```php
foreach ( $original_wp_filter[ $priority ] as $key => $config ) {
    if ( substr( $key, - strlen( $method ) ) !== $method ) {   // line 562
        continue;
    }
    unset( $wp_filter[ $hook ]->callbacks[ $priority ][ $key ] );
}
```

Those keys are WordPress's internal callback IDs. Through WordPress 7.0 they were always
strings. **WordPress 7.1 produces IDs that are purely numeric**, and PHP casts a
numeric-string array key to `int` automatically. Because `Cloudflare.php` declares
`strict_types=1`, PHP will not coerce the `int` back to `string` — it throws. Nothing catches
it, and the caller runs on `init`, so every request dies.

Note the entry point in the trace: **line 535 is the first of the two calls in
`unregister_cloudflare_clean_on_post()`, i.e. `deleted_post` at priority 10** — not the
`transition_post_status` / `PHP_INT_MAX` call. Priority 10 on `deleted_post` is a very common
registration, so the blast radius is wide.

**Critically, this happens on sites that do not have the Cloudflare plugin installed.** My
site has never had it. Every other public method in this class opens with
`if ( ! $this->is_plugin_active() ) { return; }`, but `unregister_cloudflare_clean_on_post()`
does not, so Cloudflare-specific compatibility code runs — and now crashes — everywhere.

**To Reproduce**

1. A WordPress 7.0.x site with WP Rocket active. The Cloudflare plugin is **not** installed.
2. Update WordPress core to 7.1. Change nothing else.
3. Load any URL on the site — front-end or `/wp-admin`.
4. See "There has been a critical error on this website" / HTTP 500 on every request.

Deactivating WP Rocket restores the site immediately and completely.

**Expected behavior**

WP Rocket should not fatal on WordPress 7.1. Specifically:
1. `unregister_callback()` should not assume a hook callback ID is a string. WordPress has
   never guaranteed that, and PHP's own array-key casting can produce integers regardless of
   core's ID format.
2. `unregister_cloudflare_clean_on_post()` should not run at all when the Cloudflare plugin is
   inactive, matching every other method in the class.

**Screenshots**

Not applicable — this is a server-side fatal. Stack trace instead:

```
PHP Fatal error:  Uncaught TypeError: substr(): Argument #1 ($string) must be of type string,
int given in .../wp-content/plugins/wp-rocket/inc/ThirdParty/Plugins/CDN/Cloudflare.php:562
Stack trace:
#0 .../wp-rocket/inc/ThirdParty/Plugins/CDN/Cloudflare.php(562): substr()
#1 .../wp-rocket/inc/ThirdParty/Plugins/CDN/Cloudflare.php(535):
      WP_Rocket\ThirdParty\Plugins\CDN\Cloudflare->unregister_callback()
#2 .../wp-includes/class-wp-hook.php(353):
      WP_Rocket\ThirdParty\Plugins\CDN\Cloudflare->unregister_cloudflare_clean_on_post()
#3 .../wp-includes/class-wp-hook.php(377): WP_Hook->apply_filters()
#4 .../wp-includes/plugin.php(523): WP_Hook->do_action()
#5 .../wp-settings.php(779): do_action()
  thrown in .../Cloudflare.php on line 562
```

**Desktop / Smartphone**

Not applicable. This is a server-side fatal that affects every client, every browser and every
device identically, including WP-CLI.

**Environment**

- WordPress: **7.1** (site was healthy on 7.0.4; the core update is the only change)
- WP Rocket: **3.22.1** — earliest version we have confirmed affected
- PHP: 8.3
- Cloudflare plugin (`cloudflare/cloudflare.php`): **not installed**
- Host: Rocket.net
- Object cache: Redis; page cache: WP Rocket

**Additional context**

*Updating WP Rocket does not fix this.* We pulled `Cloudflare.php` from three installed
versions and compared them: `unregister_callback()` is **byte-for-byte identical in 3.22.1,
3.23.1.1 and 3.23.2.1**. We updated a second site to 3.23.2.1 (current at time of writing)
specifically to test, and the function is unchanged. We have not tested versions older than
3.22.1, but the code looks long-standing.

*Suggested fix.* Minimal, in `unregister_callback()`:

```php
if ( substr( (string) $key, - strlen( $method ) ) !== $method ) {
```

An integer key can never match a method name, so casting is safe and behaviour-preserving.

Recommended in addition, in `unregister_cloudflare_clean_on_post()`:

```php
if ( ! $this->is_plugin_active() ) {
    return;
}
```

This would have contained the impact to Cloudflare-plugin users instead of crashing every
WP Rocket site on WordPress 7.1.

*On the "reproduce with no other plugins and a base theme" guideline.* We have not run that
test, because doing so means taking a production-mirror site down to a single plugin. It is
also unlikely to be informative here: the crash requires some callback registered on
`deleted_post` at priority 10 whose WordPress callback ID is numeric, and with only WP Rocket
active on a stock theme there may be no such callback — the early `key_exists()` return would
mask it. That absence would not indicate the bug is not real. The defect is the unguarded type
assumption itself, which is visible by inspection on the line above and will fire on any site
where such a callback exists. We are happy to run any specific diagnostic you would like on a
WordPress 7.1 environment.

*Current workaround.* WP Rocket deactivated on the affected site. Our production site is being
held on WordPress 7.0.4 because it runs the same code and would go down identically on update.

*Question.* Is a WordPress 7.1 fix scheduled, and for which release? We cannot move any site to
WordPress 7.1 while WP Rocket is active.

Tomi2k:
Fixes #8740 (and the earlier report in #8596).

## Problem

On WordPress 7.1, `Cloudflare::unregister_callback()` throws an uncaught `TypeError` on every request, taking the whole site down:

```
Uncaught TypeError: substr(): Argument #1 ($string) must be of type string, int given
in inc/ThirdParty/Plugins/CDN/Cloudflare.php:562
```

The loop passes a `$wp_filter` callback ID straight into `substr()`. Those IDs used to be strings, but WordPress 7.1 produces purely numeric ones, and PHP stores numeric-string array keys as `int`. Because this file declares `strict_types=1`, PHP will not coerce the `int` back to `string`, so it throws. The caller runs during `wp-settings.php`, so front-end, admin and WP-CLI all die.

This is not limited to sites running the Cloudflare plugin. `unregister_cloudflare_clean_on_post()` has no `is_plugin_active()` guard, unlike the other public methods in the class, so the code path runs everywhere.

## Fix

Cast the key to string before the suffix comparison.

An `int` key can never equal a method-name suffix, so the cast does not change behaviour in any case that previously worked. It just lets the loop `continue` instead of throwing.

## Verification

Applied against `3.23.2.1` on three independent production installs running WordPress 7.1 (PHP 8.3, Plesk/nginx). All three were returning hard `500`s on every request, including `/wp-login.php`; all three recovered immediately and page caching resumed normally. Two of them reach `Cloudflare.php:562` via `deleted_post` at priority 10, matching the trace in #8740.

## Notes

I deliberately kept this to the minimal one-line change so it is easy to ship as a hotfix, since sites are currently going down as WordPress 7.1 rolls out.

Two follow-ups worth considering separately, which I have not touched here:

- adding the missing `is_plugin_active()` bail-out to `unregister_cloudflare_clean_on_post()`, so the Cloudflare compatibility code does not run on sites without Cloudflare at all
- PR #8647 hardens the same method against a `null` callbacks array for #6759; these two changes are independent and should not conflict

### DIFF
--- a/inc/ThirdParty/Plugins/CDN/Cloudflare.php
+++ b/inc/ThirdParty/Plugins/CDN/Cloudflare.php
@@ -559,7 +559,8 @@ class Cloudflare implements Subscriber_Interface, DeactivationInterface {
 
 		foreach ( $original_wp_filter[ $priority ] as $key => $config ) {
 
-			if ( substr( $key, - strlen( $method ) ) !== $method ) {
+			// PHP casts numeric-string array keys to int, so $key is not always a string.
+			if ( substr( (string) $key, - strlen( $method ) ) !== $method ) {
 				continue;
 			}
 


leandrosabatini:
Confirmed in production. We hit the exact same fatal error at Cloudflare.php:562 (substr(): Argument #1 ($string) must be of type string, int given), which took the WordPress site down.

I manually applied the (string) cast from this PR and the site immediately recovered. WP Rocket is now running normally with the patch applied.

+1 for merging/releasing this as a hotfix.

hellofromahmed:
Thanks @Tomi2k for the PR, that was handled here: https://github.com/wp-media/wp-rocket/pull/8745

PR #8746, opened 2026-08-20
Guard Cloudflare::unregister_cloudflare_clean_on_post() behind is_plugin_active()

### LINKED ISSUE #8596: Fatal TypeError: substr() Argument #1 must be of type string, int given in Cloudflare.php on line 562

hellofromahmed:
# Description

Sub-task of #8596
`Cloudflare::unregister_cloudflare_clean_on_post()` was the only public method in this class that didn't bail out when the official Cloudflare plugin isn't active/configured. Every other method (`purge_cloudflare()`, `add_cdn_helper_message()`, `hide_addon_radio()`, etc.) guards on `is_plugin_active()` first. Without that guard here, this method scans and mutates `$wp_filter` on every single `init` request — even on sites that never had the Cloudflare plugin installed — for no benefit, since there's nothing to unregister in that case.

## Type of change

- [x] Sub-task of #8596
- [ ] Chore
- [ ] Release

### What was tested

Automated: added `tests/Unit/inc/ThirdParty/Plugins/CDN/Cloudflare/unregisterCloudflareCleanOnPost.php` (Brain Monkey + Mockery, matching the convention used by sibling tests like `purgeCloudflareHome.php` in the same directory):

- `testShouldNotUnregisterWhenPluginInactive` — asserts `unregister_callback()` is never invoked when `is_plugin_active( 'cloudflare/cloudflare.php' )` returns `false`.
- `testShouldUnregisterWhenPluginActive` — asserts `unregister_callback()` is still called with the same two hook/method/priority combinations as before, once the plugin is active and configured.

Verified both tests fail against the code without the guard (confirming they exercise it), and pass with the guard in place. Also ran the full `CloudflarePlugin` group (47 tests) to confirm no regressions in sibling tests.

### How to test

1. Check out this branch.
2. Run: `vendor/bin/phpunit --testsuite unit --configuration tests/Unit/phpunit.xml.dist --filter Test_unregisterCloudflareCleanOnPost`
3. Or manually: on a site without the official Cloudflare plugin active, confirm `init` no longer touches `$wp_filter['deleted_post']`/`$wp_filter['transition_post_status']` at all (e.g. via a debug breakpoint or `error_log` in `unregister_callback()`).

- `WP_Rocket\ThirdParty\Plugins\CDN\Cloudflare::unregister_cloudflare_clean_on_post()`, only relevant when the official Cloudflare plugin is/isn't active.
- No behavior change for sites where the official Cloudflare plugin is active and configured — the two `unregister_callback()` calls still happen identically in that case.

### Documentation

Added an early `if ( ! $this->is_plugin_active() ) { return; }` guard, matching the exact pattern already used by every other public method in this class. `is_plugin_active()` checks both that `cloudflare/cloudflare.php` is active and that the CF API email/key/domain options are set — the same precondition the CF-specific unregistration logic implicitly assumes.

### New dependencies

None.

### Risks

None identified: this only skips work that was already a no-op (there's nothing to unregister when the official Cloudflare plugin isn't active), matching the guard pattern already proven throughout the rest of the class.

# Mandatory Checklist

## Code validation

## Code style

*N/A — all mandatory items apply and are checked.*

# Additional Checks


Mai-Saad:
Initial related TP https://wpmediaqa.testrail.io/index.php?/runs/view/1333&page_called_by=run_view&group_by=cases:section_id&group_order=asc&display=tree

Khadreal (approved):

hanna-meda (approved):
Thanks, @wordpressfan, for this PR.

QA Done.
Everything is working as expected.
TP is all green. Results report: [testrail-report-864.pdf](https://github.com/user-attachments/files/31307730/testrail-report-864.pdf)

Issue #8699, opened 2026-08-12
Story 1 — State model & decoupled fields


Khadreal:
## Context

Technical enabler for the CDN refactor epic (#8693). No user-visible change on its own — introduces decoupled option fields and state resolvers that every other story depends on (spec §Appendix, §1.8 partial).

---

## Task 1.1 — Add decoupled option fields (2 pts)

**What:** Add new dedicated option keys for each mode so they never read/write the legacy `cdn_type` field.

**Where to change:**
- `inc/Engine/CDN/Admin/Subscriber.php` — `sanitize_cdn_type_option()` currently writes `cdn_type`; add sanitization for the new keys alongside it
- `inc/Engine/CDN/ServiceProvider.php` — register any new option-aware classes in the container
- Default option values are declared in `inc/functions/options.php` (search for `rocket_get_default_options`) — add entries for the three new keys there

**New option keys to add inside `wp_rocket_settings`:**
- `rocketcdn_free_enabled` (bool) — RocketCDN Free toggle state
- `rocketcdn_pro_enabled` (bool) — RocketCDN Pro toggle state  
- `cdn_byocdn_enabled` (bool) — Other CDN (BYOCDN) toggle state

**Acceptance criteria:**
- Given the legacy `cdn_type` option currently stores the driver type, When Free/Pro/Other CDN read or write their own state, Then they use their own dedicated fields and never read or write `cdn_type`.

---

## Task 1.2 — Applied-CDN-state resolver (3 pts)

**What:** Introduce a resolver that maps the three new boolean fields to one of three Applied-CDN states: `Nothing`, `BYOCDN`, or `RocketCDN`.

**Where to add:**
- Extend `inc/Engine/CDN/Context.php` (`WP_Rocket\Engine\CDN\Context`) with a new `get_applied_cdn_state(): string` method, or extract into a dedicated `inc/Engine/CDN/Resolvers/AppliedCDNStateResolver.php`
- `Context` already depends on `$options` (`Options_Data`) and `$subscription_controller` (`SubscriptionController`) — pass these through or inject the new resolver via `ServiceProvider.php`

**State table:**

| `rocketcdn_free_enabled` | `rocketcdn_pro_enabled` | `cdn_byocdn_enabled` | Result |
|---|---|---|---|
| 0 | 0 | 0 | `Nothing` |
| 1 | 0 | 0 | `RocketCDN` |
| 0 | 1 | 0 | `RocketCDN` |
[...]

PR #8751, opened 2026-08-20
Fix #8701: Cache-clearing scope enforcement (Story 3)

### LINKED ISSUE #8701: Story 3 — Cache-clearing scope enforcement
## Context

Part of the CDN refactor epic (#8693). Rewires cache-clear triggers so they only fire on toggle flip or an active mode's own setting change — not on tab navigation (spec §1.3).

Depends on **Story 1** (state) and **Story 2** (toggle events to hook into).

---

## Task 3.1 — Rewire cache-clear triggers (3 pts)

**Where to change:**
- `inc/Engine/CDN/Subscriber.php` (`WP_Rocket\Engine\CDN\Subscriber`) — the `maybe_clear_cache()` method is hooked to `update_option_wp_rocket_settings` and currently diffs the old vs new `cdn` option; expand the diff to compare the three new toggle fields from Story 1 (`rocketcdn_free_enabled`, `rocketcdn_pro_enabled`, `cdn_byocdn_enabled`) and skip clearing when only a non-toggle field changed (e.g. tab focus state)
- The existing `maybe_clear_cname_cache()` in the same class clears CNAME validation transients (`rocket_cname_valid_*`) when `cdn_cnames` changes; this logic stays but must not fire on a tab-switch event that doesn't touch cnames

**Key diff pattern to follow** (already in `maybe_clear_cache()`):
```php
$old_cdn = (bool) $old_value['cdn'];
$new_cdn = (bool) $value['cdn'];
if ( $old_cdn !== $new_cdn ) { ... }
```
Replicate for each new toggle field; exit early if none of the toggle fields changed.

**Acceptance criteria:**
- Given a user switches tabs without toggling, When they navigate, Then no cache clear fires.

---

## Task 3.2 — Per-mode clearing scope (2 pts)

**Where to change:**
- `inc/Engine/CDN/Cache.php` (`WP_Rocket\Engine\CDN\Cache`) — already has `clear_all_cache()` and `clear_rocketcdn_free_pages_cache()`; the latter queries the `wpr_rocket_cdn` database table via `WP_Rocket\Engine\CDN\RocketCDN\Database\Queries\RocketCDN::get_all()` and purges only those URLs
- `inc/Engine/CDN/Subscriber.php` — `maybe_clear_cache()` already branches on `Context::get_driver()` (which returns one of `ROCKETCDN_FREE_TYPE`, `ROCKETCDN_PAID_TYPE`, `BYOCDN_TYPE`); update to use the new Applied-CDN-state resolver from Story 1 for the branch condition
- Upgrade → Pro while Free active: when the new state is `PRO` but the old state was `FREE`, always call `clear_all_cache()` regardless of Free's page list

**Acceptance criteria:**
- Given RocketCDN Free is toggled on or off on its own, When the toggle changes, Then only Free's listed pages are cleared (via `Cache::clear_rocketcdn_free_pages_cache()`).
- Given the user upgrades to Pro while Free is active, When Pro activates, Then the clear is site-wide (`Cache::clear_all_cache()`), not page-scoped.

---

## Task 3.3 — Tests for §1.3 scenario table (2 pts)

**Where:** `tests/Unit/inc/Engine/CDN/Subscriber/` — follow pattern of `Subscriber/maybeClearCnameCache.php` and `Subscriber/upgradeCDN.php`.

**What to cover:**
- Tab-switch with no toggle change → no clear
- Free toggle off → `clear_rocketcdn_free_pages_cache()` called, not `clear_all_cache()`
- Pro toggle on while Free active → `clear_all_cache()` called
- Other CDN toggle on/off → `clear_all_cache()`
- Add/remove individual Free page → page-only clear (existing behaviour, regression guard)

---

## Decision needed

Issue #8464 conflicts with §1.3: it expects no clear on Pro subscription creation, but the spec requires site-wide clear. Needs a product call before Task 3.2 acceptance criteria are final.

---

## Dependencies

- Blocked by **Stories 1 & 2**
- Blocks **Stories 6, 7, 8, 9**

## Sizing

7 pts total

hellofromahmed:
# Description

Fixes #8701

Rewires the CDN cache-clear trigger to key off the new single `cdn_state` option (from Story 1, #8699) instead of the old `cdn`/`cdn_type` pair, and scopes the clear to the actual transition instead of always clearing everything.

Also closes several correctness gaps found across multiple rounds of review/QA on this change, all now fixed on this branch:
- **`cdn_state` was never written for a first-time Pro purchase.** The transient-sync callback that keeps `cdn_state` correct originally refused to activate it from `nothing` — meaning a customer who buys Pro directly, without ever having enabled Free first, never got `cdn_state` set at all. It now activates from any state except `byocdn`.
- **`cdn_state` was never backfilled for existing sites.** `cdn_state` only got seeded as a default on a genuinely fresh install; every upgrading site (the entire existing install base) had no `cdn_state` key at all, and `Utils::did_setting_change()` requires the key to already exist to detect a transition — so the very first real CDN change on any real site after this shipped would have silently failed to clear the cache. Fixed with a one-time backfill on `wp_rocket_upgrade`.
- **The sync callback could run on front-end requests, and two methods clobbered other settings via a stale snapshot.** The transient it reacts to is also read on the front end (e.g. CDN URL resolution), so it's now guarded by `is_admin()` (plus a `current_user_can()` check as defense-in-depth). Separately, `CDNOptionsManager::set_cdn_state()`, `enable()`, and `disable()` all built their writes from a stale per-request options snapshot, which could silently wipe out unrelated settings changed earlier in the same request — all three now read live instead.
- **A transient API outage could be mistaken for a genuine subscription result.** `APIClient::get_remote_subscription_data()`'s error/fallback path returns a hardcoded `plan_type => 'free'` default, indistinguishable from a real free-tier response by the sync callback's original check — a network blip could spuriously persist/activate `cdn_state` and trigger a full cache clear. Now requires `status_code === 200` before trusting `plan_type`.
- **This story's headline cache-scoping optimization had no observable effect in production.** A separate, pre-existing, unrelated subscriber (`rocket_after_save_options()`) forces an unconditional full cache clear on *any* change to the settings hash, and `cdn_state` wasn't excluded from that hash — so the scoped clear `Subscriber::maybe_clear_cache()` computes for `nothing↔rocketcdn_free` was happening, but a full clear fired anyway right after it. `cdn_state` is now excluded from that hash.
- **`Subscriber::maybe_clear_cache()` — this PR's core rewritten method — had zero test coverage.** Added a full test covering all 12 pairwise `cdn_state` transitions plus both no-op cases.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Automated, via `composer test-unit --group CDN` and `composer test-integration-adminonly` (real WP bootstrap + DB, not mocks), run directly against the real `wp-env` test container for the scenarios that needed a genuine container/DB round-trip:
- `tests/Unit/inc/Engine/CDN/Subscriber/maybeClearCache.php` (new) — all 12 pairwise `cdn_state` transitions plus both no-op cases (unchanged state, key absent from both sides), asserting exactly which `Cache` method is (or isn't) called for each.
- `tests/Integration/inc/Engine/CDN/RocketCDN/DataManagerSubscriber/handleRocketcdnCheckoutParameter.php` — no longer asserts `cdn_state` directly (that call was removed from the checkout handler itself, see Technical description); assertions trimmed to match.
- `tests/Integration/inc/Engine/CDN/RocketCDN/DataManagerSubscriber/maybeSyncCdnState.php` — 9 cases: Pro→Free downgrade, Free→Pro upgrade, no-op when the tier is unchanged, `byocdn` is never touched, RocketCDN correctly activates from `nothing` for both `paid` and `free` plan types, a non-array/empty transient value is safely ignored, an `APIClient` error-fallback default (`status_code` ≠ 200) is correctly ignored even though it carries a non-empty `plan_type`, and a missing `status_code` key is treated the same way. Plus a dedicated test proving the `is_admin()` guard is a genuine no-op on the front end (zero option writes, zero cache-clear calls, return value unchanged) — the test class now extends the screen-setting `RocketCDN\TestCase` base instead of depending on ambient screen state left by whichever test ran previously.
- `tests/Integration/inc/Engine/CDN/CdnStateBridge/backfillCdnStateOnUpgrade.php` — 4 cases covering the upgrade-time backfill: resolves to `rocketcdn_paid`/`byocdn`/`nothing` from the legacy `cdn`/`cdn_type` fields depending on scenario, and never overwrites an already-present `cdn_state`.
- Re-ran the full CDN test surface after every round of fixes to confirm no regressions, diffing failure counts against the clean pre-fix HEAD each time: full CDN unit suite green (106/106); the broader `--group AdminOnly` run's remaining failures are byte-for-byte identical in count and test name whether or not this PR's fixes are applied, confirming they're pre-existing and unrelated to any file this PR touches (traced to this container's PHP 8.3 vs. CI's PHP 8.0 job for some; a separate vfsStream/PHP-8.3 compatibility issue for others).

### How to test

1. Check out this branch, run `composer test-unit --group CDN` and `composer test-integration-adminonly -- --filter "HandleRocketcdnCheckoutParameter|MaybeSyncCdnState|BackfillCdnStateOnUpgrade"`.
2. Manually: with an active RocketCDN Free subscription (`cdn_state = rocketcdn_free`), complete a Pro purchase via the `?rocketcdn_checkout` express-checkout flow — confirm `wp_rocket_settings['cdn_state']` becomes `rocketcdn_paid` (via the transient-sync fallback, not a direct write in the checkout handler) and the whole cache is cleared.
3. Manually: on a site that has **never** enabled RocketCDN (`cdn_state = nothing`), complete a **first-time** Pro purchase via checkout — confirm `cdn_state` becomes `rocketcdn_paid`. (Before this branch's later fixes, this specific path never set `cdn_state` at all.)
4. Manually: with `cdn_state = rocketcdn_paid`, force the cached `rocketcdn_status` transient's `plan_type` to `free` and trigger any admin-context code path that reads subscription data — confirm `cdn_state` flips to `rocketcdn_free` and the whole cache is cleared; confirm the same read on the front end does *not* trigger a write.
5. Manually: simulate an existing site upgrading (option row with `cdn`/`cdn_type` set, no `cdn_state` key at all) and fire `wp_rocket_upgrade` — confirm `cdn_state` gets backfilled to match the site's current legacy CDN configuration, with no cache clear triggered by the backfill itself.
6. Toggle `cdn_state` between `nothing` and `rocketcdn_free` — confirm **only** `Cache::clear_rocketcdn_free_pages_cache()` fires end-to-end, with no full `rocket_clean_domain()` following it from `rocket_after_save_options()`.
7. Manually: force the subscription transient to a `status_code` other than 200 (e.g. simulate an API timeout) while it still carries a `plan_type` — confirm `cdn_state` is left untouched.

- `Subscriber::maybe_clear_cache()` (`inc/Engine/CDN/Subscriber.php`) — cache-clear trigger for every `cdn_state` transition; now with full test coverage.
- `CDNOptionsManager::set_cdn_state()`/`enable()`/`disable()` and `set_cdn_state()`'s caller in `DataManagerSubscriber` (the `transient_rocketcdn_status` sync callback).
- `CdnStateBridge::backfill_cdn_state_on_upgrade()` (new) — one-time `cdn_state` backfill for upgrading sites.
- `rocket_create_options_hash()` (`inc/functions/admin.php`) — now excludes `cdn_state`.
- No changes to `cdn_type`, `get_driver()`, or the legacy pause/resume AJAX flows — out of scope for this story.

### Documentation

`Subscriber::maybe_clear_cache()` now diffs a single `cdn_state` key (via `Utils::did_setting_change()`) instead of the old `cdn`/`cdn_type` pair. Moving between `nothing` and `rocketcdn_free` only clears Free's own page-list cache (`Cache::clear_rocketcdn_free_pages_cache()`); every other transition (anything involving `rocketcdn_paid` or `byocdn`) clears the whole cache, matching the epic spec's per-mode clearing-scope table.

`CDNOptionsManager::set_cdn_state( string $state )`, `enable()`, and `disable()` all persist their change by reading the current settings via `Options::get( 'settings', ... )` (a plain, always-live `get_option()`/`update_option()` wrapper), merging in the changed key, and writing the result back — which fires WP's `update_option_wp_rocket_settings` action, the hook `maybe_clear_cache()` is already listening on. All three previously built their write from the injected `Options_Data` instance instead — a per-request snapshot frozen at container-boot time — which could silently clobber any settings change made elsewhere in the same request through a different path; fixed after a real test failure surfaced it for `set_cdn_state()`, then traced and fixed for `enable()`/`disable()` too once review found a concrete collision path (`maybe_retry_activation()` can trigger a `set_cdn_state()` write via `get_subscription_data()`, then call `enable()` later in the same method, whose stale-snapshot write used to silently revert it).

The only caller of `set_cdn_state()` is `DataManagerSubscriber::maybe_sync_cdn_state()`, hooked to `transient_rocketcdn_status` (the filter WordPress applies on every `get_transient( 'rocketcdn_status' )` read, not just cache misses). It:
- **Is guarded by `is_admin()` and `current_user_can( 'rocket_manage_options' )`.** This transient is also read on front-end requests (e.g. `FrontendSubscriber`'s CDN URL/zone resolution), and this callback writes an option and can trigger a cache clear as a side effect — not something a front-end page view should ever do. The capability check is defense-in-depth (every current caller of `get_subscription_data()` is already capability-gated at the point of call), added so the write path isn't solely relying on that.
- **Only trusts a `plan_type` when `status_code === 200`.** `APIClient::get_remote_subscription_data()`'s own error/fallback path (network failure, non-200, empty body, JSON-decode failure) returns a hardcoded `plan_type => 'free'` default — indistinguishable from a genuine free-tier response without also checking `status_code`. Without this check, a transient API outage could spuriously persist/activate `cdn_state` and trigger a cache clear.
- **Activates `cdn_state` from any state except `byocdn`**, not only from an already-applied `rocketcdn_free`/`rocketcdn_paid` tier as originally written. A customer can go straight to a paid subscription via checkout without ever enabling Free first, so the original, narrower guard left `cdn_state` unset for that path entirely. This is safe to allow *given the status_code check above*: a genuinely confirmed `plan_type` can only come from a successful API response, and that call only happens once a `rocketcdn_user_token` already exists for *this* site — so a confirmed result always reflects a genuine per-site activation, never an account-wide or incidental signal. `byocdn` stays excluded, since that's an explicit user choice this callback must never override (a BYOCDN site can still carry a leftover RocketCDN token from a past trial).
- Was originally also meant to be set directly by `DataManagerSubscriber::handle_rocketcdn_checkout_parameter()` right after a successful purchase; that direct call was removed in favor of relying entirely on this transient-sync fallback (the checkout flow already recreates the `rocketcdn_status` transient as part of enabling CDN, which fires this same callback). **Still an open question, not resolved by this round of fixes**: is relying purely on the fallback acceptable long-term, given it's a multi-hop implicit chain (enable → transient delete → re-fetch → transient set → filter fires → write) versus a single explicit call with a directly-testable outcome?

`CdnStateBridge::backfill_cdn_state_on_upgrade()`, hooked to `wp_rocket_upgrade`: `cdn_state` was only ever seeded as a default for genuinely fresh installs (`rocket_first_install()`); every upgrading site — the entire pre-existing install base — has no `cdn_state` key at all, and `Utils::did_setting_change()` requires the key to already exist to report a transition. Without a backfill, the very first real `cdn`/`cdn_type` change on any such site after this ships would silently fail to trigger `maybe_clear_cache()`. This writes `cdn_state` once, resolved from whatever's already live via `CdnStateBridge`'s own `resolve_live()`/`legacy_to_state()` logic. The write itself doesn't touch `cdn`/`cdn_type`, so it doesn't trigger `reconcile()` or a cache clear — nothing about the site's active CDN behavior changed, only the tracking field catching up to it. Every real transition afterward then correctly compares against a properly-populated old value.

`rocket_create_options_hash()` (`inc/functions/admin.php`) now excludes `cdn_state`. `rocket_after_save_options()` — a separate, pre-existing subscriber also hooked on `update_option_wp_rocket_settings`, unrelated to this PR's own changes — forces an unconditional full `rocket_clean_domain()` whenever this hash changes at all. Since `cdn_state` wasn't excluded, every `cdn_state` change was getting a full clear from this hook regardless of what `maybe_clear_cache()` had already decided, silently negating the scoped-clear optimization for `nothing↔rocketcdn_free`. `maybe_clear_cache()` already owns deciding the correct scope for every transition, so this exclusion doesn't lose any required clearing — it just stops the redundant one.

### New dependencies

None.

### Risks

Low-to-medium risk, mitigated by the fixes above and by multiple rounds of review/QA:
- The sync callback can now activate `cdn_state` from `nothing`, where it previously couldn't — deliberately, to close the first-time-Pro-purchase gap. The safety property now depends on both the `status_code === 200` check and the per-site-token precondition in `APIClient`; `byocdn` can never be touched by this path.
- The cache-clear scoping change means some transitions that previously always cleared everything now clear a narrower scope (Free page-list only) — covered by the new/extended tests above, and now genuinely observable end-to-end after the `rocket_create_options_hash()` fix.
- `CDNOptionsManager::enable()`/`disable()` now read live settings instead of a per-request snapshot — behaviorally equivalent for the common case, differs only if a write happened elsewhere in the same request through a different path (which is exactly the bug being fixed).

### Known CI issue (pre-existing, unrelated to this PR)

PHPStan on this branch fails on a handful of class-name case-mismatch errors (e.g. `Adminbar::class` vs the imported `AdminBar`, in `inc/Engine/Saas/ServiceProvider.php` and `inc/Engine/Common/PerformanceHints/ServiceProvider.php`; `extends Abstract_render` vs the imported `Abstract_Render` in `inc/Engine/Admin/Settings/Render.php`; plus a couple of similar case mismatches in unrelated test files). These already exist on the base branch (`enhancement/8693-rocketcdn-free-tiers-refactor`) itself, untouched by any commit in this PR, and will fail PHPStan on any PR built on top of it. Left as-is rather than fixed or baselined here — needs a fix on the base branch directly.

# Mandatory Checklist

## Code validation

## Code style

Not applicable — all mandatory checklist items are ticked. No UI exists yet for this story (the toggle UI is Story 2/#8700), so acceptance criteria were validated via the automated integration tests listed above rather than a screenshot/video.

# Additional Checks


inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php
```diff
@@ -382,6 +383,46 @@ public function maybe_disable_cdn() {
...
+	 * flow, without needing a dedicated cron re-check.
+	 *
+	 * Only corrects the tier while RocketCDN is already the applied cdn_state - it
+	 * never activates RocketCDN from "nothing", and never touches "byocdn".
+	 *
+	 * @param mixed  $value     Transient value.
+	 * @param string $transient Transient name.
+	 * @return mixed
+	 */
+	public function maybe_sync_cdn_state( $value, $transient ) {
```
Khadreal: I believe this need an admin check, the method would be fired on every `get_transient('rocketcdn_status')` including frontend when it check subscription status

hellofromahmed: Good catch — you're right that this needs to stay admin-only. Currently `DataManagerSubscriber` is admin-only registered (`inc/Plugin.php`), so `maybe_sync_cdn_state()` doesn't fire on the front end today, but it's the kind of implicit dependency that's easy to break later (e.g. if this subscriber is ever promoted to a common one for other hooks) — and this callback writes an option + can trigger a cache clear as a side effect, so it shouldn't rely on that. Added an explicit `is_admin()` guard at the top, mirroring the pattern already used in `FrontendSubscriber::set_cdn_cnames()`. Fixed in 1272bd31b.

inc/Engine/CDN/Subscriber.php
```diff
@@ -568,30 +568,29 @@ private function cdn_driver_should_rewrite_url(): bool {
 	 * @return void
 	 */
 	public function maybe_clear_cache( $old_value, $value ) {
-		$cdn_changed      = Utils::did_setting_change( 'cdn', $old_value, $value );
-		$cdn_type_changed = Utils::did_setting_change( 'cdn_type', $old_value, $value );
+		$cdn_state_changed = Utils::did_setting_change( 'cdn_state', $old_value, $value );
```
hellofromahmed, line 571: [HIGH] CORRECTNESS — confirmed via live testing, not just static analysis: `Utils::did_setting_change('cdn_state', $old_value, $value)` requires `isset($old_value['cdn_state'])` to be true before it will report a change (see `inc/Engine/Common/Utils.php`). `cdn_state` is a brand-new key that does not exist in `wp_rocket_settings` on any pre-existing site. Reproduced directly against a real installed site: stripped `cdn_state` from the live option, then wrote it via the real `update_option()` path — zero cache-clear calls fired, even for a `nothing → rocketcdn_free` transition that should trigger a free-pages clear. **This means the very first time RocketCDN/BYOCDN is activated on any real-world site after this ships, no cache is cleared at all.**

**Fix:** either seed `cdn_state` into the default options schema so the key always pre-exists before this hook can fire, or change the transition check here to treat "missing → present" as a valid change (e.g. `array_key_exists` + explicit default instead of relying on `did_setting_change()`'s `isset()` semantics).

hellofromahmed:
> [!NOTE]
> Generated by the AI delivery pipeline (lead-reviewer · standalone).

**Blockers:**
- [HIGH] `inc/Engine/CDN/Subscriber.php::maybe_clear_cache()` — `Utils::did_setting_change('cdn_state', ...)` requires the key to already exist in `$old_value`, but `cdn_state` is brand new and doesn't exist on any pre-existing site. Confirmed live (see QA report below): the very first activation of RocketCDN/BYOCDN on any real-world site triggers **zero cache clear**. See inline comment.
- [HIGH] `DataManagerSubscriber::handle_rocketcdn_checkout_parameter()` (unchanged by this PR) — the PR description states this method calls `set_cdn_state(Context::ROCKETCDN_PAID_TYPE)` after a successful checkout, and that a test asserts it. Neither is true: commit `e0ea36ee6` removed that call ("no need to set cdn_state with redirection from upgrade routine to the plugin"), and commit `f321b3d8d` then dropped the assertions that would have caught it. Practical effect: a Free→Pro upgrade via checkout doesn't deterministically set `cdn_state` during that request — it only self-corrects later, implicitly, if the post-redirect page happens to re-read the subscription transient. Worse, a first-time Pro purchase straight from `nothing` (never having tried Free) never gets `cdn_state` set at all, since `maybe_sync_cdn_state()` explicitly refuses to activate from `nothing`. Either restore the explicit call (matching what the description claims) or have `maybe_sync_cdn_state()` also handle activation from `nothing`, and update the PR description to match whichever is kept.

**Follow-ups (non-blocking):**
- `CDNOptionsManager::set_cdn_state()` doesn't validate `$state` against the known `Context` constants before persisting — not currently exploitable (only call site passes a hardcoded constant) but no defense-in-depth, unlike `Context::get_cdn_state()` which does validate.
- `inc/Engine/Abilities/Options/SetOption.php`'s `READ_ONLY_OPTIONS` comment documents `cdn_state` as one-way-derived from `cdn`/`cdn_type` with "nothing syncs the reverse direction" — this PR introduces exactly such a second, direct writer. The comment is now stale and should be updated to describe the new narrower writer.
- `Context::get_cdn_state()`/`Options_Data`-based reads are unconditionally overridden by `CdnStateBridge::resolve_live()` and never see what `set_cdn_state()` persisted — worth a docblock note so this isn't discovered the hard way later.
- `maybe_clear_cache()`'s transition comment block omits the `rocketcdn_free ↔ byocdn` pair (functionally falls through to the safe default, just undocumented).
- Zero test coverage for `Subscriber::maybe_clear_cache()`'s transition matrix — the PR's core logic change — before or after this PR.
- The just-added `is_admin() === false` no-op branch in `maybe_sync_cdn_state()` (latest commit, responding to a review comment) has no regression test.
- `Utils::did_setting_change()`'s `isset()`-based check is itself worth a defensive default (`?? Context::CDN_STATE_NOTHING`) rather than relying solely on the upgrader's backfill.

**Open questions:**
1. Was dropping the checkout-handler `set_cdn_state()` call intentional (relying on the transient-sync fallback) or a rebase/squash regression? The commit history suggests intentional, but the PR description was never reconciled with it.
2. Is the team aware `Context::get_cdn_state()` will never reflect a `set_cdn_state()` write due to `CdnStateBridge::resolve_live()`'s override? Is that the intended long-term design?

**Test coverage:** PARTIAL — new tests for `maybe_sync_cdn_state()` and the checkout flow exist, but the checkout test's `cdn_state` assertions were removed in a later commit without replacement, and `maybe_clear_cache()` (this PR's primary change) has no coverage at all.

hellofromahmed:
> [!NOTE]
> Generated by the AI delivery pipeline (qa-engineer · standalone).

**QA: ❌ FAIL**

Tested against a booted `wp-env` environment, using direct calls through the real DI container against the live DB (unit suite green; the WP core integration test library wasn't available in this checkout, compensated with equivalent live-object testing).

|---|---|---|
| 1. `set_cdn_state()` persists and round-trips | Live call via `wp eval-file` | ✅ |
| 2. `maybe_clear_cache()` scopes clear correctly per transition | Reflection spy on real `cdn_subscriber`, all 7 transitions | ✅ isolated logic / ❌ real trigger path |
| 3. `maybe_sync_cdn_state()` syncs free↔paid, ignores nothing/byocdn | All 6 cases from the PR's own test plan | ✅ |
| 4. `is_admin()` guard is a genuine no-op when false | Direct calls in both admin and non-admin context | ✅ |
| 5. Checkout flow sets `cdn_state` to `rocketcdn_paid` | Read source + git history + live `wp eval-file` run | ❌ |

### Smoke Tests
Settings page, dashboard, deactivate/reactivate, PHP error log, full unit suite (2793 tests) — all PASS.

### Security / Performance Observations
- No new security surface: `cdn_state` values are always one of 4 hardcoded constants, never raw input. `is_admin()` is checked first and cheaply in `maybe_sync_cdn_state()`, negligible per-request cost.
- [FUNCTIONAL — see AC2] `Utils::did_setting_change()`'s `isset()`-based check treats "key didn't exist before, now it does" as "no change." Reproduced live: stripped `cdn_state` from a real site's option, wrote it through the real `update_option()` path — zero cache-clear calls fired. Since `cdn_state` doesn't exist on any pre-existing site, **the first-ever activation on every real production site will silently skip the cache clear.**

**Blockers:**
- AC2: first-ever `cdn_state` write on any site is silently ignored by `maybe_clear_cache()` (see inline comment on `inc/Engine/CDN/Subscriber.php:571`). Fix: seed `cdn_state` into the default options schema, or change the transition check to treat "missing → present" as a valid change.
- AC5: `handle_rocketcdn_checkout_parameter()` never sets `cdn_state` (removed in `e0ea36ee6`; its test assertions removed in `f321b3d8d`) — contradicts the PR description. A first-time Pro purchase via checkout leaves `cdn_state` permanently unset. Either restore the explicit call or extend `maybe_sync_cdn_state()` to handle activation from `nothing`, and update the PR description to match.

**Summary:** 3/5 criteria PASS (AC1, AC3, AC4). 2 FAIL — both confirmed live, not just theoretical: AC2 (the transition logic is correct in isolation but never fires on a real site's first activation) and AC5 (checkout flow regression vs. the PR's own stated behavior). No security issues, no performance regression. All smoke tests and the full unit suite pass.

hellofromahmed:
Update on the AC5 checkout-flow finding from the review/QA reports above.

@hellofromahmed clarified: the direct `set_cdn_state()` call was deliberately dropped from `handle_rocketcdn_checkout_parameter()` because a Free→Pro upgrade already recreates the `rocketcdn_status` transient during that same request (`cdn_options->enable()` deletes it, the subsequent `get_subscription_data()` call re-fetches and re-sets it), which fires `maybe_sync_cdn_state()` and sets `cdn_state` via that path instead. That's correct and does work for the Free→Pro case.

But it left one gap: `maybe_sync_cdn_state()`'s old guard only allowed *correcting* an already-applied `rocketcdn_free`/`rocketcdn_paid` tier — it explicitly refused to activate from `nothing`. Since a customer can go straight to a paid subscription via checkout without ever enabling Free first (confirmed: `AdminPageSubscriber::add_subscription_modal()` shows the paid checkout link unconditionally, with no precondition on prior Free enrollment), that first-time-Pro path never got `cdn_state` set at all — not via the checkout handler (no explicit call), and not via the fallback (guard blocked it).

**Fixed in 84a31c18a**: narrowed the guard from "must already be free/paid" to "must not be byocdn". This is safe because a conclusive `plan_type` from the transient can only ever come from a real API response, and `APIClient::get_remote_subscription_data()` only makes that call once a `rocketcdn_user_token` already exists for this site — so a conclusive result always means a genuine per-site activation already happened, never an incidental or account-wide signal. `byocdn` stays excluded, since that's an explicit user choice this callback must never override (and a BYOCDN site can still carry a leftover RocketCDN token from a past trial). Test fixture updated: `testShouldNotActivateFromNothing` → `testShouldActivateFromNothingWhenPlanTypeIsPaid` + `testShouldActivateFromNothingWhenPlanTypeIsFree`.

**Open question for discussion**: with this fix, does `handle_rocketcdn_checkout_parameter()` still need its own explicit `set_cdn_state()` call, or is relying entirely on the transient-sync fallback (now covering all four states except staying off byocdn) an acceptable design going forward? Arguments either way:
- *For relying on the fallback only*: one code path to maintain instead of two, less duplication.
- *For an explicit call in the checkout handler*: the fallback's activation is now implicit and multi-hop (enable → transient delete → re-fetch → transient set → filter fires → write) — a change anywhere in that chain (e.g. a future refactor changing when/whether `enable()` deletes the transient) could silently break `cdn_state` activation again with no direct test covering the checkout handler itself asserting the outcome, only the transient-sync unit path.

Curious which way the team wants to go here before I'd consider this fully closed.

inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php
```diff
@@ -382,6 +383,61 @@ public function maybe_disable_cdn() {
...
+		if ( ! is_admin() ) {
+			return $value;
+		}
+
+		if ( ! is_array( $value ) || empty( $value['plan_type'] ) ) {
```
hellofromahmed: [CRITICAL] CORRECTNESS: this check conflates a genuine API-confirmed subscription result with APIClient::get_remote_subscription_data()'s own error/fallback default. That default (returned on network failure, non-200, empty body, or JSON-decode failure) is ['plan_type' => 'free', 'status_code' => 500, ...] - non-empty, so it passes this guard just as easily as a real free-tier API response does. A transient RocketCDN API outage on a site with a saved token can therefore be indistinguishable from this site is genuinely on the Free plan to this method, causing it to persist cdn_state and trigger Subscriber::maybe_clear_cache()'s full-clear branch purely from a network blip - or, after the activate-from-nothing change, even activate cdn_state on a site that was never touched.

Note this ambiguity predates the activate-from-nothing change (it could already spuriously downgrade an already-paid customer during an outage) - that change just widened the blast radius to also spuriously activate from an untouched state.

Fix: check $value['status_code'] === 200 (or an equivalent explicit success flag) before trusting plan_type, matching the same conclusive-result pattern already used in the sibling PR (#8754)'s SubscriptionController::auto_detect_pro_subscription().

hellofromahmed:
> [!NOTE]
> Generated by the AI delivery pipeline (lead-reviewer · standalone). Fresh re-review of the final state (head `ffe2839c8`), no assumptions carried over from earlier rounds.

**Blockers:**
- [CRITICAL] `DataManagerSubscriber::maybe_sync_cdn_state()` — the "conclusive `plan_type`" safety argument in the docblock doesn't hold: `APIClient::get_remote_subscription_data()`'s error/fallback path returns a hardcoded `plan_type => 'free'` default (network failure, non-200, empty body, JSON-decode failure), indistinguishable from a genuine free-tier response by this method's `empty($value['plan_type'])` check. A transient API outage can spuriously persist/activate `cdn_state` and trigger a full cache clear. Pre-dates the "activate from nothing" change (which just widened the blast radius). See inline comment for the fix (check `status_code === 200`, same pattern PR #8754 uses).
- [HIGH] `Subscriber::maybe_clear_cache()` — the PR's namesake rewritten method — still has zero test coverage anywhere in the repo. Flagged in an earlier review round, still open in this final diff.
- [HIGH] `CDNOptionsManager::enable()`/`disable()` (unchanged by this PR, not in the diff) still build their write from the stale `Options_Data` snapshot — the exact bug `set_cdn_state()` was fixed for. Traced a real collision: `maybe_retry_activation()` can call `get_subscription_data()` (triggering `maybe_sync_cdn_state()` → `set_cdn_state()`'s live write) *before* calling `enable()` later in the same method — `enable()`'s stale full-array write can silently revert the `cdn_state` just set. Either fix `enable()`/`disable()` the same way, or explicitly document the ordering constraint.

**Follow-ups (non-blocking):**
- Add automated coverage for `maybe_sync_cdn_state()`'s `is_admin() === false` branch (currently only manually verified per the PR description); also fix `Test_MaybeSyncCdnState` extending `AdminTestCase` directly instead of the screen-setting `RocketCDN\TestCase` base, which makes even the "true" branch's coverage order-dependent rather than hermetic.
- `inc/Engine/Abilities/Options/SetOption.php`'s design comment describes `cdn_state` as one-way-derived solely by `CdnStateBridge::reconcile()` — this PR adds a second, independent writer (`set_cdn_state()` via `maybe_sync_cdn_state()`) that can diverge from what `reconcile()` would compute. Comment needs updating; team should decide if the two writers need reconciling against each other.
- `maybe_clear_cache()`'s transition comment table omits the `FREE ↔ BYOCDN` pair (code handles it correctly via the default full-clear branch, just undocumented).
- Consider an explicit `current_user_can('rocket_manage_options')` check inside `maybe_sync_cdn_state()` itself as defense-in-depth — not currently exploitable (every current caller of `get_subscription_data()` is already capability-gated), but the write path has no independent boundary of its own.
- Minor PR-description inaccuracy: it credits the `handleRocketcdnCheckoutParameter` test trim to removing a `cdn_state` assertion — that test never asserted `cdn_state`; the assertions actually removed were `cdn_cnames`-related.

**Test coverage:** FAIL — the two gaps above (`maybe_clear_cache()` untested; `is_admin()`-false branch untested) are both still open.

hellofromahmed:
> [!NOTE]
> Generated by the AI delivery pipeline (qa-engineer · claude-sonnet-5). Re-run after fix commit `cbd898167`.

**QA: ✅ PASS**

|---|---|---|
| 1. Tab switch without toggling → no cache clear | Analysis + Unit (`Subscriber/maybeClearCache.php`) | ✅ |
| 2. RocketCDN Free toggled alone → scoped clear only (`clear_rocketcdn_free_pages_cache`), not full clear | Analysis + Integration (`rocketAfterSaveOptions.php::cdnAloneShouldNotTriggerFullClearing`, new in this commit) | ✅ |
| 3. Upgrade to Pro while Free active → site-wide clear via `clear_all_cache()` | Unit (`Subscriber/maybeClearCache.php`, e.g. `testShouldClearAllWhenFreeToPaid`) | ✅ |
| 4. `CDNOptionsManager::enable()/disable()` keeps `cdn` live for other same-request readers | API (live WP-CLI reflection) + Integration (`CdnStateBridge/resolveLiveCdn.php`) | ✅ |
| 5. Malformed-but-HTTP-200 RocketCDN response not treated as genuine free-tier confirmation | Integration (`DataManagerSubscriber/maybeSyncCdnState.php::testShouldIgnoreMalformedButHttp200Response`) | ✅ |

**Re-verification of AC4 (previously FAILED):** Reproduced the exact live-reflection scenario from the prior run — two independently-built `Options_Data` instances (simulating two classes with their own DI-resolved copy, since `'options'` is registered with `add()`, not `addShared()`). Before the fix, a write via `CDNOptionsManager::enable()` was invisible to the other instance for the rest of the request. With `cbd898167`'s `pre_get_rocket_option_cdn` filter (`CdnStateBridge::resolve_live_cdn()`), the independently-built instance now reads live:
```
Sanity - pre-enable stored value via early_reader: 0 (expect 0)
After enable(), independently-built early_reader->get(cdn): 1 (expect 1 if live)
After disable(), independently-built early_reader->get(cdn): 0 (expect 0 if live)
```
This confirms `Options_Data::get()` applies `pre_get_rocket_option_cdn` before consulting its own internal (possibly stale) array — the same mechanism `cdn_state` already used — so the fix is not dependent on DI instance identity. The new integration test `CdnStateBridge\Test_ResolveLiveCdn::testShouldReflectEnableAndDisableToAnIndependentlyBuiltInstance` passes and covers this regression going forward.

**Note (non-blocking, pre-existing, out of PR scope):** with the *default* `cdn_type` (`rocketcdn`) and no valid license on this local install, a separate, pre-existing filter on the same hook (`Render\Controller::maybe_pause_cdn_for_inactive_subscription`, `inc/Engine/CDN/Render/Controller.php:583`, not touched by this PR) forces `cdn` back to `false` via its own `is_forced_paused()`/license-validity guard, independent of the live-resolution fix. This is why the test (and my reflection) forces `cdn_type` to `byocdn` to isolate the mechanism under test — it is unrelated to the DI-staleness bug this PR fixes and pre-dates this branch.

**Re-verification of AC2/AC3 after the `rocket_create_options_hash()` change:** confirmed `cdn`/`cdn_type` exclusion (in addition to `cdn_state`) prevents `rocket_after_save_options()` from forcing a redundant full `rocket_clean_domain()` on a CDN-alone toggle. The new fixture case `cdnAloneShouldNotTriggerFullClearing` passes (`rocket_clean_domain` correctly absent from expectations). `Subscriber::maybe_clear_cache()`'s own scoped-clear unit tests (AC2/AC3) are all still green and unaffected — the hash change only removes a second, previously-redundant full-clear trigger sitting on top of the scoped clear, it doesn't touch the clearing-scope decision itself.

**Environment:** backend-only PR (no JS/CSS/Twig changed) — confirmed via `git diff --name-only`. Strategy B (browser) not required; `ui_visible: false` confirmed.

**Smoke tests:** plugin deactivate/reactivate via WP-CLI — no fatals. `wp-admin` and front end reachable (redirect-to-login on unauthenticated `wp-admin` requests is expected).

**Pre-existing, unrelated failures observed while running the suite (confirmed present on `origin/enhancement/8693-rocketcdn-free-tiers-refactor` base branch too, not a regression from this PR):**
- `CDNOptionsManager::enable/disable` `shouldDeleteAll` and 8 datasets in `rocketAfterSaveOptions.php` fail on `VirtualFilesystemTrait.php:94` (vfs cache-directory assertions) — reproduced identically against the base branch's `CDNOptionsManager.php`/`admin.php` in this environment.
- One risky (non-failing) test in `maybeSyncCdnState.php` due to a missing `wp_wpr_rocket_cdn` DB table on this local install (dbDelta not run) — assertion itself passed, only extra DB-error output was flagged as risky.

inc/Engine/CDN/Subscriber.php
hellofromahmed: Fixed by the cdn_state backfill (commit 49354513b, already on this branch before this comment was posted — confirmed via a fresh QA pass that the first-ever activation on a real site now correctly triggers the cache clear).

inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php
hellofromahmed: Fixed in bf8038e18 — now requires status_code === 200 before trusting plan_type, and added a current_user_can('rocket_manage_options') check as defense-in-depth. New test cases cover the API-fallback-default and missing-status_code scenarios.

hellofromahmed:
All findings from the latest review + QA round addressed:

- **[CRITICAL]** \`maybe_sync_cdn_state()\` now requires \`status_code === 200\` before trusting \`plan_type\`, so an API outage can no longer be mistaken for a genuine free-tier response. Also added \`current_user_can('rocket_manage_options')\` as defense-in-depth.
- **[HIGH]** \`Subscriber::maybe_clear_cache()\` now has full test coverage — all 12 pairwise \`cdn_state\` transitions plus both no-op cases.
- **[HIGH]** \`CDNOptionsManager::enable()\`/\`disable()\` fixed to stop reading the same stale \`Options_Data\` snapshot \`set_cdn_state()\` was fixed for, closing the traced collision via \`maybe_retry_activation()\`.
- **[QA blocker]** \`cdn_state\` excluded from \`rocket_create_options_hash()\`, so the scoped-clear optimization for \`nothing↔rocketcdn_free\` is no longer negated by \`rocket_after_save_options()\`'s unconditional full clear.
- Follow-ups: \`is_admin() === false\` branch now has automated coverage (test class also switched to the screen-setting \`RocketCDN\TestCase\` base instead of depending on ambient state); missing \`FREE ↔ BYOCDN\` line added to the transition comment; \`SetOption.php\`'s design comment updated to reflect \`cdn_state\`'s second writer.

Verified against the real \`wp-env\` test container: full CDN unit suite (106/106) and the affected integration groups all pass, with zero new failures — confirmed by diffing against the clean pre-fix HEAD (\`ffe2839c8\`), which shows the exact same pre-existing/unrelated failure counts in both runs.

hellofromahmed:
> [!NOTE]
> Generated by the AI delivery pipeline (lead-reviewer · standalone). Fresh re-review of the final state (head `bf8038e18`), no assumptions carried over from earlier rounds.

**Review: ✅ PASS**

All 5 fixes from the previous round independently re-verified as correct and complete:
- `status_code === 200` guard — confirmed `status_code` is set on every return path of `APIClient::get_remote_subscription_data()` (its only writer in the codebase), no bypass found.
- `current_user_can()` doesn't affect the WP-Cron path — `is_admin()` already blocks it there regardless.
- `enable()`/`disable()`/`set_cdn_state()` live-read fix — the exact collision via `maybe_retry_activation()` is closed.
- `rocket_create_options_hash()` exclusion — correctly scoped, only one other caller of that function exists and it's the intended target.
- `Subscriber::maybe_clear_cache()` test coverage — fixture genuinely covers all 12 pairwise transitions + 2 no-ops, mock expectations correctly distinguish which single `Cache` method fires per case.

**No blockers.**

**Follow-ups (non-blocking):**
- The same stale-`Options_Data`-snapshot pattern fixed in `enable()`/`disable()`/`set_cdn_state()` this round is still present in two sibling methods: `Rest.php::save_pause_state()`/`save_cdn_type()` (MEDIUM — both mutate `cdn`/`cdn_type`, which `CdnStateBridge::reconcile()` derives `cdn_state` from) and `DataManagerSubscriber::refresh_cdn_cname()` (LOW — a one-time, version-gated upgrade migration, low practical risk). Explicitly out of this PR's stated scope; worth a follow-up PR.
- Minor inaccuracy in the `SetOption.php` comment updated last round: it says `set_cdn_state()` writes `cdn_state` "via... CdnStateBridge's own upgrade-time backfill" — the backfill actually writes `cdn_state` directly in `CdnStateBridge` itself, never through `set_cdn_state()`. Worth a one-line reword.
- Sharper detail on the PR's own already-acknowledged open question (relying on the transient-sync fallback instead of an explicit checkout-handler call): for white-label/reseller accounts specifically, `AdminPageSubscriber::display_rocketcdn_status()` — the page load that would otherwise re-sync `cdn_state` after checkout — never runs (`is_white_label_account()`/`is_reseller_account()` early-return before reading subscription data), so a post-checkout tier change could go unsynced indefinitely for that account type. Not introduced by this round's fixes; worth folding into whatever decision resolves the existing open question.

hellofromahmed:
> [!NOTE]
> Generated by the AI delivery pipeline (qa-engineer · standalone). Fresh re-test of the final state (head `bf8038e18`), no assumptions carried over from earlier rounds.

**QA: ✅ PASS**

6 of 7 "How to test" scenarios verified live against the real `wp-env` container (the 7th — an actual RocketCDN checkout HTTP round-trip — can't be exercised without live API access; the underlying mechanism was confirmed directly instead). All four targeted stress-tests of this round's fixes passed under realistic conditions:

| Fix | Result |
|---|---|
| `status_code === 200` guard (tested with 500 and absent entirely) | ✅ |
| `current_user_can()` doesn't affect the WP-Cron path | ✅ — `is_admin()` already blocks it there |
| `rocket_create_options_hash()` exclusion (hook-fire-count verified) | ✅ — zero redundant full clears |
| `enable()`/`disable()` no longer revert a just-set `cdn_state` (realistic collision repro) | ✅ |

**Correction to the original agent report:** it flagged `Test_BackfillCdnStateOnUpgrade::testShouldBackfillNothingWhenCdnDisabled` as a reproducible failure. I independently re-ran this test twice against the same `wp-env` container — once in isolation, once combined exactly as the PR's own "How to test" step 1 prescribes (`--filter "HandleRocketcdnCheckoutParameter|MaybeSyncCdnState|BackfillCdnStateOnUpgrade"`) — and it passed cleanly both times (0 failures). Not treating this as a real blocker; likely a one-off environment flake in the agent's run rather than a genuine, reproducible issue.

**Design note (non-blocking):** `CdnStateBridge::reconcile()` and `CDNOptionsManager`'s state writers both react to the same `update_option_wp_rocket_settings` hook; a narrow window (e.g. the subscription transient expiring between two writes in one request) could theoretically let `reconcile()` revert a freshly-set `cdn_state`. Not reproduced, not a blocker — worth a look in a future pass.

**Blockers:** None.

Mai-Saad:
@hellofromahmed Thanks for the PR, Please find possible risks raised by Claude below :pray: 

Reviewed the diff against the current source of the touched files (not just the description). This is an unusually well self-audited PR — the 5 issues already documented as fixed across review rounds all check out against the actual code. Two more things stood out that seem worth a look before merge:

**1. `CDNOptionsManager::enable()`/`disable()` no longer keep the shared `Options_Data` snapshot in sync**

`enable()`/`disable()` (`inc/Engine/CDN/RocketCDN/CDNOptionsManager.php:52`, `:110`) changed from writing through `$this->options->set('cdn', ...)` (which also updated the in-memory `Options_Data` instance) to a raw `options_api->get('settings')` → mutate → `options_api->set()` round trip. That correctly fixes the stale-snapshot-clobbers-a-fresh-write bug called out for `set_cdn_state()`, but introduces the mirror problem: `$this->options` is the same per-request `Options_Data` singleton that `get_rocket_option('cdn')` reads everywhere else, and unlike `cdn_state` there's no `pre_get_rocket_option_cdn` filter recomputing it live — so it just returns whatever was cached at container boot.

This is reachable in the same request, not just theoretical: `DataManagerSubscriber::maybe_retry_activation()` is hooked on `admin_init` (`:519`, runs on every admin page load for capable users) and can call `$this->cdn_options->enable()` mid-request. For the rest of that page load, anything reading `get_rocket_option('cdn')` (settings-page rendering, other subscribers) would see the pre-enable value even though the DB was just updated. Self-corrects on the next page load, so no data loss, but it's a real regression vs. the old behavior, and the same class of bug this PR was otherwise careful to avoid.

**2. The `status_code === 200` guard doesn't fully close the "outage mistaken for a real result" gap**

`maybe_sync_cdn_state()` (`inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php:423`) requires `status_code === 200` before trusting `plan_type`, to stop a failed API call from being read as a genuine free-tier response. Checking `APIClient::get_remote_subscription_data()` directly: it does set `status_code` to the real HTTP code before its early-return on non-200, so that part's solid. But two further fallback branches — empty response body, and a 200 response whose JSON has `success` falsy/missing — are reached *after* that check passes, and they return the same hardcoded `plan_type => 'free'` default **with a genuine `status_code === 200`**.

So a malformed-but-200 response from the backend (empty body, bad JSON, explicit `success: false`) is still indistinguishable from a real free-tier confirmation under the new guard — it would sync `cdn_state` to `rocketcdn_free`, including demoting a Pro site and triggering a full `clear_all_cache()` (paid→free isn't the scoped nothing↔free case). Same failure mode as the one this PR fixed, just one layer deeper (API-level failure vs. transport-level failure). Might be worth folding the `success` check into the same trust condition.

Nothing here looks blocking, but both seem worth confirming before merge given how much of this PR's value is in closing exactly these kinds of gaps.

hellofromahmed:
Thanks, both confirmed and fixed in 22ef9f558:

1. **Stale `Options_Data` snapshot in `enable()`/`disable()`** - restored the mirror write onto `$this->options` alongside the safe `options_api` read-modify-write, so classes reading `cdn` straight off the shared instance (`CDN.php`, `Render/Controller.php`, `RocketCDN/Rest.php`, `Subscriber.php`, `Support/Meta.php`) don't see a stale value for the rest of the request.

2. **Malformed-but-200 response treated as a real free-tier confirmation** - added an explicit `success` flag to `APIClient::get_remote_subscription_data()`'s return value (`false` on every fallback path, `true` only for a genuinely decoded response), and `maybe_sync_cdn_state()` now requires it alongside `status_code === 200`. Updated/added test coverage for the new branch.

inc/Engine/CDN/RocketCDN/CDNOptionsManager.php
```diff
@@ -38,22 +38,61 @@ public function __construct( Options $options_api, Options_Data $options ) {
...
 	 * @return void
 	 */
 	public function enable( bool $clear_cache = true ) {
-		$this->options->set( 'cdn', 1 );
+		$settings        = $this->options_api->get( 'settings', [] );
+		$settings['cdn'] = 1;
 
-		$this->options_api->set( 'settings', $this->options->get_options() );
+		$this->options_api->set( 'settings', $settings );
+		$this->options->set( 'cdn', 1 );
```
hellofromahmed: [HIGH] LOGIC — this mirror does not do what the docblock above claims, and the same-request stale-read edge case from the AI review this commit was meant to fix is still open.

`$this->options` here is **not** the same object as `$this->options` in `Subscriber`, `CDN`, `Render/Controller`, `RocketCDN/Rest`, or `Support/Meta`. The `'options'` container service is registered with `add()`, not `addShared()`, in `inc/classes/ServiceProvider/class-options.php:39` — per this repo's own `.claude/skills/wp-rocket-architecture/SKILL.md` ("`addShared()` → singleton ... `add()` → factory (new instance every call)"), every class that gets `Options_Data` injected receives its own independently-resolved instance, all seeded from the same boot-time array by value, never linked afterward. I confirmed this empirically with a minimal League Container repro (two `get()` calls on a non-shared definition return distinct objects; mutating one leaves a third resolution's copy untouched).

So `maybe_retry_activation()` (admin_init) calling `enable()` mid-request still leaves `Subscriber::is_cdn_enabled()` / `CDN::is_allowed()` / etc. reading the pre-enable `cdn` value for the rest of that request — exactly the scenario this fix claims to close, and exactly the spec's Edge Case #4. No test would have caught this either: `enable.php`/`disable.php` integration tests only assert on the persisted DB option, never on a second, independently-resolved `Options_Data` instance.

**Fix:** drop the mirror and instead register a `pre_get_rocket_option_cdn` filter that resolves `cdn` live from `$this->options_api`/settings — the same pattern `CdnStateBridge::resolve_live()` already uses for `cdn_state` via `pre_get_rocket_option_cdn_state`. That's the only way to make *every* reader see the fresh value, since there's no single shared object to write through.

inc/Engine/CDN/RocketCDN/CDNOptionsManager.php
```diff
@@ -69,14 +108,20 @@ public function save_token( string $token ): void {
...
 	 * @return void
 	 */
 	public function disable() {
-		$this->options->set( 'cdn', 0 );
+		$settings        = $this->options_api->get( 'settings', [] );
+		$settings['cdn'] = 0;
 
-		$this->options_api->set( 'settings', $this->options->get_options() );
+		$this->options_api->set( 'settings', $settings );
+		$this->options->set( 'cdn', 0 );
```
hellofromahmed: [HIGH] LOGIC — same issue as `enable()` above (see that comment): `$this->options` is a private, independently-resolved `Options_Data` instance, not shared with `Subscriber`/`CDN`/etc., so this mirror is a no-op for every other reader of `cdn`. Same fix applies (a `pre_get_rocket_option_cdn` live filter instead of trying to propagate through this object).

inc/functions/admin.php
```diff
@@ -536,6 +543,7 @@ function rocket_create_options_hash( $value ) {
 		'preload_excluded_uri'        => true,
 		'cache_reject_uri'            => true,
 		'version'                     => true,
+		'cdn_state'                   => true,
```
hellofromahmed, line 549: [HIGH] LOGIC — this exclusion doesn't achieve the stated goal, so the Task 3.2 scoped-clear optimization never actually fires for the case it exists for.

`rocket_after_save_options()` (`inc/admin/options.php:73`, on the same `update_option_wp_rocket_settings` hook, loaded on every request via `inc/main.php:105`) still diffs the raw `$value` against `$oldvalue` through this hash. `cdn` itself — the field that actually flips 0↔1 for the nothing↔rocketcdn_free transition this comment is about — is **not** in `$removed`. So toggling RocketCDN Free on/off still makes this hash differ, and `rocket_after_save_options()` still calls `rocket_clean_domain()` (full clear) on top of whatever `Subscriber::maybe_clear_cache()` → `Cache::clear_rocketcdn_free_pages_cache()` already scoped down. `Cache::clear_all_cache()` is literally `rocket_clean_domain()`; the free-pages path never touches it or its `rocket_after_clean_domain` guard, so nothing here short-circuits it.

This isn't hypothetical — the existing (unmodified) `tests/Fixtures/inc/admin/rocketAfterSaveOptions.php` already has a case asserting `rocket_clean_domain` fires when `cdn` changes. Net result: the spec's edge case "RocketCDN Free toggled on or off on its own → Only Free's listed pages are cleared … not a full clear" is not actually satisfied end-to-end — a full clear always follows the scoped one.

**Fix:** excluding `cdn_state` alone isn't enough; `rocket_after_save_options()` needs to stop treating the nothing↔rocketcdn_free transition as hash-dirty too — e.g. have it consult the resolved `cdn_state` transition (the same way `Subscriber::maybe_clear_cache()` does) instead of diffing the raw `cdn` field, so it doesn't unconditionally re-trigger a full clear that `maybe_clear_cache()` already decided to scope down. This is a structural change to the shared full-clear trigger, not a one-line hash exclusion, and needs its own test (extending `tests/Unit/inc/admin/rocketAfterSaveOptions.php`).

hellofromahmed:
> [!NOTE]
> Generated by the AI delivery pipeline (lead-reviewer · Claude Sonnet 5). Re-review of `cbd898167` — both prior HIGH blockers independently re-verified against the current code (not the commit message), plus a fresh cross-file/Hyrum's-Law pass and a PHPStan run on the touched files.

**Prior blockers — both confirmed fixed:**
- ✅ `CDNOptionsManager.php` mirror bug — replaced with `CdnStateBridge::resolve_live_cdn()` on `pre_get_rocket_option_cdn`. Confirmed `options_api` reads via `get_option()` live (no snapshot staleness) and the new `resolveLiveCdn.php` integration test genuinely exercises cross-instance visibility.
- ✅ `admin.php:549` redundant full-clear — `cdn`/`cdn_type` now excluded alongside `cdn_state`. Confirmed `rocket_create_options_hash()` has no other consumers and the exclusion doesn't affect htaccess/config/mobile-cache regen (those are gated separately). New `cdnAloneShouldNotTriggerFullClearing` fixture case covers it.

**New blockers introduced by this same fix commit:**
- [HIGH] `CdnStateBridge.php:150` — `resolve_live_cdn()` always returns non-null, which short-circuits `Options_Data::get()` before the pre-existing `get_rocket_option_cdn` post-filter (`Subscriber::apply_pause_on_rocketcdn_only()`) ever runs. That hook is now permanently unreachable for every 'cdn' read plugin-wide — a silent, undocumented behavior change (front-end CDN-enablement resolution for non-rocketcdn/BYOCDN changes), not covered by any test.
- [MEDIUM] `tests/Integration/inc/Engine/CDN/CdnStateBridge/resolveLiveCdn.php` — will fail CI's PHPStan check. Verified by running `vendor/bin/phpstan analyse` directly: 5 un-baselined `custom.rules.discourageOptionUsage` errors (direct `get_option()`/`update_option()`/`delete_option()` calls). The other two test files touched by this commit got matching baseline entries; this one was missed.

**Nice-to-haves:**
- `tests/Integration/.../CDNOptionsManager/enable.php` / `disable.php` — still only assert on the persisted DB option; the new cross-instance liveness coverage lives entirely in `resolveLiveCdn.php`, consider a note/cross-reference.

inc/Engine/CDN/RocketCDN/CDNOptionsManager.php
hellofromahmed: Fixed in cbd898167. Confirmed the DI-sharing premise was wrong — replaced the ineffective mirror with a `pre_get_rocket_option_cdn` live-resolving filter on `CdnStateBridge` (mirroring the existing `cdn_state` pattern), plus a regression test exercising a second, independently-resolved `Options_Data` instance.

inc/Engine/CDN/RocketCDN/CDNOptionsManager.php
hellofromahmed: Fixed in cbd898167, same fix as `enable()` above — the mirror onto $this->options is removed; the new `pre_get_rocket_option_cdn` filter on `CdnStateBridge` covers `disable()`'s writes too since it resolves live regardless of which instance made the write.

inc/functions/admin.php
hellofromahmed: Fixed in cbd898167. Added 'cdn' and 'cdn_type' to the exclusion list alongside 'cdn_state' — confirmed via a new isolated test case that a pure cdn-only change no longer triggers rocket_clean_domain() (the existing fixture cases happened to pass regardless, since they had other unrelated key differences masking the actual signal).

hellofromahmed:
Correction to my earlier comment (5490789863): the "mirror the value onto `$this->options`" fix for issue #1 there was wrong — an independent lead-review pass caught it. WP Rocket's `'options'` container service is registered with `add()`, not `addShared()` (`inc/classes/ServiceProvider/class-options.php:39`), so every class gets its own independently-resolved `Options_Data` instance; mutating `CDNOptionsManager`'s own copy never reached anyone else's.

Fixed properly in cbd898167:
- Replaced the mirror with a `pre_get_rocket_option_cdn` live-resolving filter on `CdnStateBridge`, the same pattern already used for `cdn_state`. This makes every `cdn` read live regardless of which `Options_Data` instance is asking, so `enable()`/`disable()`'s write is visible everywhere for the rest of the request.
- Also closed a related gap an independent lead-review pass found: `rocket_create_options_hash()` excluded `cdn_state` but not `cdn`/`cdn_type` themselves, so `rocket_after_save_options()` was still forcing a redundant full `rocket_clean_domain()` on every Free-toggle flip, defeating the scoped-clear optimization end-to-end. Added both to the exclusion list.
- Added a regression test that exercises a second, independently-built `Options_Data` instance (simulating another already-instantiated class) to prove the live-read fix actually works, plus an isolated test proving a pure `cdn`-only change no longer triggers a full clear.

Verified against the live wp-env test containers, not just static review: full `CDN` integration suite went from 26 failures/10 errors before this fix to 23 failures/10 errors after (no regressions, some pre-existing failures fixed as a side effect), and the new tests pass in isolation.

hellofromahmed:
Follow-up on the re-review (comment 5491028318): both new findings from that pass are fixed in 0ae88918a.

1. **`resolve_live_cdn()` silently disabling the `get_rocket_option_cdn` post-filter** — confirmed real: it always returned a non-null value, short-circuiting `Options_Data::get()` before it ever reached its own post-filter application, which permanently stopped `Subscriber::apply_pause_on_rocketcdn_only()` (forces `cdn` on for a BYOCDN driver on the front end) from running for every `cdn` read plugin-wide. Fixed by explicitly re-applying that post-filter (`wpm_apply_filters_typed('boolean|integer|string', 'get_rocket_option_cdn', $live, $default)`) against the live value before returning. Added a regression test (`testShouldStillApplyGetRocketOptionCdnPostFilterOnTheFrontEnd`) proving the override still fires on the front end.

2. **New test file triggers the `discourageOptionUsage` PHPStan rule** — confirmed (this is also what's failing CI right now). Rather than adding baseline entries, rewrote the test to use the `Options` wrapper class instead of raw `get_option()`/`update_option()`/`delete_option()`, matching the pattern the rule is actually asking for.

Re-verified against the wp-env test containers: isolated the two newly-surfaced integration test failures (`Test_BackfillCdnStateOnUpgrade`, `Test_MaybeSetRocketcdnAsCdnTypeOnUpgrade`) by temporarily swapping in the untouched pre-PR file and confirming they fail identically there too — they're pre-existing bugs in `apply_pause_on_rocketcdn_only()` reading a stale `Options_Data` snapshot for `cdn_type`, unrelated to this change, just previously hidden by the overly-aggressive short-circuit. `vendor/bin/phpstan analyze` and `vendor/bin/phpcs` both clean on all touched files.

PR #8754, opened 2026-08-21
Closes #8702: Story 4 — Fresh-install Pro detection & resolution

### LINKED ISSUE #8702: Story 4 — Fresh-install Pro detection & resolution
## Context

Part of the CDN refactor epic (#8693). Automates Pro-subscription detection on first install so the correct CDN tab is shown immediately, without manual user action (spec §1.4, §1.5).

Independent of Stories 2/3 — can run in parallel with them.

---

## Task 4.1 — Auto-run Pro detection on first install via Action Scheduler (3 pts)

**Where to change:**
- `inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php` (`WP_Rocket\Engine\CDN\RocketCDN\DataManagerSubscriber`) — already handles `maybe_retry_activation()` and `update_user_token()`; add a listener for `rocket_first_install_options` (or `wp_rocket_first_install`) that schedules the initial detection job
- `inc/Engine/CDN/RocketCDN/Queue.php` (`WP_Rocket\Engine\CDN\RocketCDN\Queue`) — already extends `AbstractASQueue` and has `schedule_create_status_job()` with group `'rocket-cdn'`; add a new method `schedule_pro_detection_job(int $attempt = 1)` that schedules hook `rocket_cdnfree_pro_detect` with a bounded attempt counter in the payload
- `inc/Engine/CDN/RocketCDN/SubscriptionController.php` — already has `WebsiteSearch` API client (`$website_search_api_client`) and `get_subscription_data()`; the detection handler should call these to determine Pro status
- Retry: reschedule from within the AS job handler if status is not conclusive, decrement attempt counter, stop at 0

**Action Scheduler pattern to follow:** `Queue::schedule_create_status_job()` schedules `rocket_cdnfree_website_create_status` 30 seconds out with `['task_id' => $task_id]`; mirror this pattern, passing `['attempt' => $n]`.

**Acceptance criteria:**
- Given a fresh install, When the plugin activates, Then detection runs automatically with no user action required.

---

## Task 4.2 — Failure UI after all retries exhausted (2 pts)

**Where to change:**
- `inc/Engine/CDN/RocketCDN/NoticesSubscriber.php` (`WP_Rocket\Engine\CDN\RocketCDN\NoticesSubscriber`) — already renders the CTA banner via `displayRocketcdnCta()`; add a new method (e.g. `displayProDetectionFailure()`) that renders when a transient like `rocket_cdn_pro_detection_failed` is set
- `inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php` — when the final retry attempt completes without success, set the `rocket_cdn_pro_detection_failed` transient and fire a dismissible admin notice
- The "refresh customer data" button should POST to a new `admin-post.php` action (e.g. `rocket_retry_pro_detection`) that re-schedules `schedule_pro_detection_job(X)` and redirects back

**Acceptance criteria:**
- Given all X retry attempts fail, When the final attempt completes, Then the failure is shown with a refresh button, not silently retried forever.

---

## Task 4.3 — First-install resolution (2 pts)

**Where to change:**
- `inc/Engine/CDN/RocketCDN/CDNOptionsManager.php` (`WP_Rocket\Engine\CDN\RocketCDN\CDNOptionsManager`) — already has `enable()` / `disable()` and `has_token()`; add `set_post_install_state(bool $is_pro)` that writes the Story 1 decoupled fields: Pro confirmed → set `rocketcdn_pro_enabled = 1` (deactivated, toggle off), No Pro → set `rocketcdn_free_enabled = 0` (deactivated, no Free-creation check)
- `inc/Engine/CDN/Render/Controller.php` — `maybe_display_rocketcdn_cta()` already gates on subscription state; ensure it reads the new fields from Story 1 so the correct tab renders after first-install resolution

**Acceptance criteria:**
- Given Pro is confirmed, When fresh-install resolution runs, Then the RocketCDN tab shows Pro, deactivated (toggle off).
- Given no Pro subscription is detected, When fresh-install resolution runs, Then the Free tab is shown deactivated and no Free-creation check has run yet.

---

## Dependencies

- Blocked by **Story 1** (state model & decoupled fields)
- Runs in parallel with Stories 2, 3, 5

## Sizing

7 pts total

jeawhanlee:
# Description

Fixes #8702
On fresh install, RocketCDN's Pro/Free status is unknown until the license API responds. This PR adds a background retry loop that resolves it automatically, with a manual retry option if all attempts fail, so users are not left in an indeterminate CDN state.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Automated (PHPUnit integration/unit tests):
- `SubscriptionController::auto_detect_pro_subscription()` — the synchronous first-install check, schedules the retry job only when the subscription API doesn't return a conclusive result.
- `SubscriptionController::scheduled_auto_detect_pro_subscription()` — retries on an inconclusive API response, cancels/clears state on a conclusive one, and sets a failure transient once attempts are exhausted.
- `DataManagerSubscriber::handle_manual_retry_pro_detection()`, fired via `admin_post_rocket_retry_pro_detection` — nonce check, capability check, transient clear, and a synchronous one-shot detection check (no job scheduling).
- `Render\Controller::add_rocketcdn_paid_section()` / `add_rocketcdn_free_section()` — updated to branch on live subscription data.

### How to test

1. On a fresh WP Rocket install, Check that Subscription stat is checked and resolved immediately. After plugin activation, user should see RocketCDN Pro section instead of RocketCDN Free if there is an active subscription.
2. If API does not resolve correctly with a 200, confirm `rocket_cdn_auto_detect` is scheduled via Action Scheduler and runs even outside wp-admin (e.g. a plain WP-Cron request).
3. Let it run to a conclusive result (or force one via `pre_http_request`) and confirm the RocketCDN paid/free section rendered reflects the resolved plan.
4. To see the failure path: force the API response to stay inconclusive until attempts are exhausted, confirm the `rocket_cdn_pro_detection_failed` transient is set and the admin notice appears on the WP Rocket settings page.
5. Click "Refresh customer data" on that notice (visible only on the WP Rocket settings page), confirm it redirects back, the failure transient is cleared, and the plan is re-resolved synchronously (no background job is scheduled by this action).

RocketCDN: fresh-install activation flow, the WP Rocket settings page admin notices, the CDN driver sections (paid/free), and plugin uninstall (added transient cleanup).

### Documentation

- `SubscriptionController::auto_detect_pro_subscription()` runs synchronously on `wp_rocket_first_install`; if the subscription API doesn't give a conclusive answer (HTTP 200) right away, it schedules the `rocket_cdn_auto_detect` Action Scheduler job via `Queue::schedule_pro_detection_job()`.
- `Queue::schedule_pro_detection_job()` cancels any already-pending detection job before scheduling a new one, avoiding duplicate pending actions across retries/re-triggers; `cancel_pro_detection_job()` now matches regardless of the `attempt` arg the job was scheduled with.
- `SubscriptionController::scheduled_auto_detect_pro_subscription( $attempt )` is the Action Scheduler callback: a conclusive result clears the failure transient and cancels any further pending job; an inconclusive one reschedules with `$attempt - 1` until it hits 0, at which point `rocket_cdn_pro_detection_failed` is set.
- `DataManagerSubscriber` (which owns these hooks) is now registered as a common subscriber instead of an admin-only one, so its Action Scheduler callbacks are also registered on genuine WP-Cron requests, not just wp-admin/AJAX ones — fixes "no callbacks are registered" failures when Action Scheduler's async loopback isn't available.
- `Render\Controller::add_rocketcdn_paid_section()` / `add_rocketcdn_free_section()` now branch on `SubscriptionController::is_paid()` (live/cached subscription data) instead of the DB-persisted `Context::get_driver()` state.
- `NoticesSubscriber::display_pro_detection_failure()` shows the failure as an admin notice with a retry link to `admin_post_rocket_retry_pro_detection`, handled by `DataManagerSubscriber::handle_manual_retry_pro_detection()` (nonce + capability checked, then clears the failure transient and re-runs the subscription check synchronously — a one-shot check outside the automatic backoff chain, it does not reschedule the Action Scheduler job).
- Also fixes a broken `composer install`: `mobiledetect/mobiledetectlib` 3.74 declares both a PSR-4 and a classmap autoloader; Mozart's classmap mover failed to match the nested classmap path and deleted the vendor copy before the target existed. Added `override_autoload` in `composer.json` so Mozart treats the package as PSR-4-only, matching how it's already vendored in `inc/Dependencies/Detection/MobileDetect.php`.

### New dependencies

None.

### Risks

Low. The detection job only reads subscription status and never modifies the CDN configuration a user has already set; retries are capped, deduplicated, and end in a visible, dismissible-by-action notice rather than a silent failure.

# Mandatory Checklist

## Code validation

## Code style

N/A — all mandatory items apply and are checked.

# Additional Checks


jeawhanlee:
Screenshot 2026-08-25 at 09 10 33

composer.json
```diff
@@ -127,7 +127,14 @@
 				"league/container",
 				"berlindb/core",
 				"wp-media/plugin-family"
-			]
+			],
+			"override_autoload": {
```
hellofromahmed, line 130: why this change?

inc/Engine/CDN/RocketCDN/SubscriptionController.php
```diff
@@ -436,4 +449,103 @@ public function is_in_grace_period(): bool {
...
+	 *
+	 * @return void
+	 */
+	public function auto_detect_pro_subscription(): void {
+		$subscription_data = $this->flush_caches_and_get_subscription_data();
+
+		if ( 200 !== $subscription_data['status_code'] ) {
+			$this->queue->schedule_pro_detection_job();
+			return;
+		}
```
hellofromahmed:
```suggestion
		if ( 200 === $subscription_data['status_code'] ) {
			return;
		}
		$this->queue->schedule_pro_detection_job();
```

hellofromahmed (approved):
Approving with minor requests.

Notes for @wp-media/qa-team 
1. This currently applies to free and paid subscriptions.
2. We have two attempts and then fail with the notice that @jeawhanlee mentioned [here](https://github.com/wp-media/wp-rocket/pull/8754#issuecomment-5407409229).
3. This PR will introduce more calls to user details endpoint because we remove the transient with each attempt.

inc/Engine/CDN/Render/Controller.php
```diff
@@ -126,7 +126,7 @@ public function __construct(
 	 * @return array
 	 */
 	public function add_rocketcdn_paid_section( array $sections ): array {
-		if ( Context::ROCKETCDN_PAID_TYPE !== $this->context->get_driver() ) {
+		if ( ! $this->subscription_controller->is_paid() ) {
```
hellofromahmed, line 156: [MEDIUM] PERFORMANCE: this replaces `Context::get_driver()`'s cheap local short-circuit with an unconditional call to `is_paid()`. On a `rocketcdn_status` transient cache-miss, `is_paid()` -> `get_subscription_data()` fires a blocking `wp_remote_get()` to rocketcdn.me (~5s WP HTTP default timeout). A site on the BYOCDN driver that still has a leftover `rocketcdn_user_token` (e.g. previously trialed RocketCDN, then switched away) now risks that blocking call on every `admin_init` across all of wp-admin, not just the CDN tab — a regression the old guard specifically prevented.\n\n**Fix:** restore a cheap local short-circuit (e.g. check `cdn_type`/token presence) before calling `is_paid()`, mirroring the old `get_driver()` behavior.

inc/Engine/CDN/Render/Controller.php
```diff
@@ -160,7 +160,7 @@ public function add_rocketcdn_paid_section( array $sections ): array {
 	 * @return array
 	 */
 	public function add_rocketcdn_free_section( array $sections ): array {
-		if ( Context::ROCKETCDN_PAID_TYPE === $this->context->get_driver() ) {
+		if ( $this->subscription_controller->is_paid() ) {
```
hellofromahmed, line 196: [MEDIUM] PERFORMANCE: same issue as `add_rocketcdn_paid_section()` above — `is_paid()` can trigger a blocking remote API call on every `admin_init` for BYOCDN-driver sites with a leftover `rocketcdn_user_token`, where the old `Context::get_driver()` guard short-circuited locally. See the comment on line 129 for the fix.

hellofromahmed:
> [!NOTE]

**Blockers:**
- [MEDIUM] `Render\Controller::add_rocketcdn_paid_section/add_rocketcdn_free_section` — `is_paid()` replaces `Context::get_driver()`'s local short-circuit; on a `rocketcdn_status` cache-miss this can trigger a blocking `wp_remote_get()` wherever `rocket_cdn_driver_sections` fires. Still open from the previous review pass.
- [MEDIUM] `inc/Plugin.php:321` (new) — `rocketcdn_data_manager_subscriber` moved from admin-only to common, so it's now constructed on every front-end pageview for every visitor. Only the `rocket_cdn_auto_detect` AS callback needed broader-than-admin registration (to fix the wp-cron gap); the other 9 hooks (`admin_init`, 5x `wp_ajax_*`, `admin_post_*`) never fire outside wp-admin regardless. Also broadens when `maybe_refresh_rocketcdn_details()` (token save / cache flush) can run — undocumented in the PR's Risks section. Fix: split the 3 new hooks into a small dedicated common subscriber, keep the rest admin-only.
- [MEDIUM] `SubscriptionController::flush_caches_and_get_subscription_data()` — still relies implicitly on `user_client->get_user_data()` succeeding (via the `set_transient_wp_rocket_customer_data` hook chain) to invalidate the `rocketcdn_status` transient before each retry. If that unrelated user-data call fails (network blip), `rocketcdn_status` is never flushed and the retry re-reads the same stale cached "inconclusive" result — silently burning one of only 2 retry attempts without a genuine live check. No test exercises a 2-attempt sequence to catch this. Fix: explicitly `delete_transient('rocketcdn_status')` (or call `$this->options_manager->flush_subscription_cache()`) unconditionally in this method.
- No test coverage for `Render\Controller::add_rocketcdn_paid_section()`'s guard condition (only the free-section test was updated) — still open.

**Nice-to-haves:**
- `Queue::schedule_pro_detection_job()` TOCTOU race under concurrent first-install requests (self-heals, wasteful only).
- `Queue::cancel_pro_detection_job()`'s `is_scheduled()` pre-check before `cancel_all()` is redundant.
- `composer.json` mozart override_autoload fix is unrelated scope creep — author already justified it inline (Slack link); consider a separate PR next time.
- `add_exclusions_fields()` / `render_cdn_driver_tabs()` in the same `Render\Controller` still branch on the legacy `Context::get_driver()` / `ROCKETCDN_PAID_TYPE` instead of the `get_applied_cdn_state()` + `is_paid()` pattern this PR introduces two methods above — worth a consistency pass in a follow-up.
- `handle_manual_retry_pro_detection()` checks nonce before capability — minor convention inconsistency, not exploitable.

**Verified, no new issue found:** retry/backoff arithmetic (`$attempt - 1` bottoms out at 1, never reschedules at 0 — matches test fixtures); `cancel_pro_detection_job()`'s `is_scheduled($hook, null)` / `cancel_all($hook, null)` do match regardless of the `attempt` arg (confirmed against `as_has_scheduled_action`/`as_unschedule_all_actions` — `null` is deliberate and required here, the default `[]` would NOT have matched); nonce + capability gate present and correctly wired on `admin_post_rocket_retry_pro_detection`; DI wiring for the new `UserClient` constructor arg is correct; no unescaped output, no SQL/authz issues introduced.

**Test coverage:** PASS with the gaps noted above — integration tests cover `auto_detect_pro_subscription()`, `scheduled_auto_detect_pro_subscription()` (reschedule/exhaustion/cancel — single-attempt scenarios only), and `handle_manual_retry_pro_detection()` (nonce/capability/success paths).

hellofromahmed:
> [!NOTE]
> Generated by the AI delivery pipeline (qa-engineer · Claude Sonnet 5). Code-level pass only — Strategy C (test suite + source analysis), no live environment or browser this run.

**QA: ⚠️ PARTIAL**

| Acceptance Criterion | Method | Result | Why it failed |
|---|---|---|---|
| 1. Fresh install triggers detection automatically, no user action | Analysis + test read (`Test_AutoDetectProSubscription`) | ✅ | — |
| 2. Inconclusive result schedules `rocket_cdn_auto_detect`, retries with bounded attempts, stops instead of retrying forever | Analysis + test read (`Test_ScheduledAutoDetectProSubscription`) | ✅ | — |
| 3. Job callback registers/runs outside wp-admin (WP-Cron) | Analysis (`inc/Plugin.php`: `rocketcdn_data_manager_subscriber` moved into `init_common_subscribers()`, which always runs) | ✅ | — |
| 4. Manual "Refresh customer data" retry: nonce + capability checked, transient cleared, synchronous re-check, no job re-scheduled, redirects back | Analysis + test read (`Test_HandleManualRetryProDetection`) | ✅ | — |
| 5. Failure state shown as an admin notice with a refresh button | Analysis (code read only) | ⚠️ | Backend "stop retrying / set failure transient" is tested, but `NoticesSubscriber::display_pro_detection_failure_notice()` itself has **zero test coverage** (its sibling `activation_failed_notice()` has an integration test; this one doesn't). Actual on-screen rendering can't be confirmed without a browser this run. |
| 6. Paid/Free CDN section rendered reflects the resolved subscription plan | Analysis (code read only) | ⚠️ | `Controller::add_rocketcdn_paid_section()` branches on `get_applied_cdn_state()` + `is_paid()` but has **no unit test at all** (before or after this PR). `add_rocketcdn_free_section()`'s test only exercises the `is_paid() === false` branch, not the "paid subscriber → free section suppressed" branch. Rendering itself not verifiable without a browser. |
| 7. Issue #8702 Task 4.3: fresh-install resolution sets an explicit deactivated/toggle-off state (`CDNOptionsManager::set_post_install_state()`, decoupled `rocketcdn_pro_enabled`/`rocketcdn_free_enabled` fields) | Analysis (repo-wide search) | ❓ | No `set_post_install_state()`, `rocketcdn_pro_enabled`, or `rocketcdn_free_enabled` exists anywhere in this branch. The PR may have superseded this with the existing `get_applied_cdn_state()` approach (and simply never calls `enable()` during detection, leaving CDN off by default) — but nothing in the diff or tests asserts the toggle state explicitly, so this can't be confirmed as intentionally satisfied vs. a gap. |

**Blockers:**
- Criterion 5: add an integration test for `NoticesSubscriber::display_pro_detection_failure_notice()` mirroring the existing `activationFailedNotice.php` test (transient present/absent, screen/capability gating).
- Criterion 6: add a unit test for `Controller::add_rocketcdn_paid_section()` (currently untested), and extend `Test_AddRocketcdnFreeSection` to cover the `is_paid() === true` early-return branch.
- Criterion 7: confirm with the ticket owner whether issue #8702's Task 4.3 "deactivated/toggle-off" persistence is intentionally out of scope for this PR or still pending; if pending, it should be tracked before closing #8702.

composer.json
jeawhanlee: This is the reason: https://group-onecom.slack.com/archives/C0AQU3D94CX/p1787296670208269

inc/Plugin.php
```diff
@@ -319,6 +318,7 @@ private function init_common_subscribers() {
 			'cdn_state_bridge',
 			'cdn_render_subscriber',
 			'rocketcdn_frontend_subscriber',
+			'rocketcdn_data_manager_subscriber',
```
hellofromahmed: [MEDIUM] ARCHITECTURE/PERFORMANCE: `DataManagerSubscriber` moves from `init_admin_subscribers()` to `init_common_subscribers()`, so it is now constructed (6 deps, incl. `SubscriptionController` with 8 more) on **every front-end pageview for every anonymous visitor**, not just wp-admin/cron. Of its 10 registered hooks, only `rocket_cdn_auto_detect` (the AS retry callback) actually needed broader-than-admin registration to fix the wp-cron gap described in the PR — `admin_init`, the 5x `wp_ajax_*` hooks, and `admin_post_rocket_retry_pro_detection` never fire outside wp-admin/ajax regardless of registration scope, so moving the *whole* class is unnecessary blast radius. It's also a Hyrum's-Law-relevant behavior change not called out in the PR's Risks section: `set_transient_wp_rocket_customer_data` → `maybe_refresh_rocketcdn_details()` (token save + subscription-cache flush) previously never ran on front-end requests; now it can.\n\n**Fix:** split the 3 new fresh-install-detection hooks into a small dedicated subscriber registered as common (or common+cron), and leave the existing admin-only ajax/admin_init handlers in `rocketcdn_data_manager_subscriber` under `init_admin_subscribers()`.

Mai-Saad:
@jeawhanlee @hellofromahmed Thanks for the PR, Please check possible risks raised by AI and confirm we are good or need further change /followup GH :pray: 

> 1. RocketCDN Paid/Free section rendering now depends on live API state, not a DB flag
> [Controller.php:126-133](https://github.com/wp-media/wp-rocket/blob/9a6138899f5afba816ffe1405a14864de80b2291/inc/Engine/CDN/Render/Controller.php#L126-L133) switched from Context::get_driver() (persisted in the DB) to get_applied_cdn_state() + SubscriptionController::is_paid(). is_paid() reads the wp_rocket_customer_data/subscription cache and falls back to false on a miss or API failure. If the subscription API is briefly unreachable and the cache expires, a paying customer could see the Free CDN section instead of Paid — previously this couldn't happen since it was driven by a DB value, not live API availability.
> 
> 2. Every detection attempt flushes and rebuilds a plugin-wide cache, not just RocketCDN's
> flush_caches_and_get_subscription_data() ([SubscriptionController.php:407-417](https://github.com/wp-media/wp-rocket/blob/9a6138899f5afba816ffe1405a14864de80b2291/inc/Engine/CDN/RocketCDN/SubscriptionController.php#L407-L417)) deletes and repopulates the wp_rocket_customer_data transient, which is used across the whole plugin (license data), not scoped to RocketCDN. On a fresh install with an inconclusive result, that's up to 3 cycles (1 sync + 2 async retries) × 2 outbound HTTP calls each, and each repopulation re-fires set_transient_wp_rocket_customer_data → maybe_refresh_rocketcdn_details, which can trigger further side effects. Multiplied across every fresh install in the wild, this is real added load on WP Rocket's own API.
> 
> 3. Manual "Refresh customer data" retry is synchronous and blocking
> handle_manual_retry_pro_detection() makes two blocking API calls inline before wp_safe_redirect(). If the license/RocketCDN API is slow, the admin-post request (and the user's click) hangs. Acceptable for an explicit user action, but worth noting as UX risk under API degradation.
> 
> 4. Non-atomic duplicate-job guard — cancel_pro_detection_job() then reschedule ([Queue.php:211-222](https://github.com/wp-media/wp-rocket/blob/9a6138899f5afba816ffe1405a14864de80b2291/inc/Engine/CDN/RocketCDN/Queue.php#L211-L222)) isn't atomic across requests, so two near-simultaneous triggers (e.g. manual retry racing the scheduled job) could theoretically produce two pending actions. Low impact — retries are idempotent.
> 
> 5. The retry callback is now a "common" (non-admin) subscriber ([FrontendSubscriber.php](https://github.com/wp-media/wp-rocket/blob/9a6138899f5afba816ffe1405a14864de80b2291/inc/Engine/CDN/RocketCDN/FrontendSubscriber.php#L100), confirmed against Plugin.php's admin/common subscriber grouping), which is required so it fires under real WP-Cron. On sites without working async loopback, that means the retry's outbound API calls can execute during a live front-end pageview via WP-Cron, adding latency to that request — a known, generally-accepted tradeoff for WP-Cron-based jobs, but worth being aware of.

jeawhanlee:
@DahmaniAdame I believe we'll need the correct text for this new notice:

> Screenshot 2026-08-25 at 09 10 33

jeawhanlee:
@Mai-Saad I believe we are good here.
We'll wait for copy-write here before merging

hanna-meda:
@jeawhanlee , @hellofromahmed, please find below the QA notes for this PR, plus one open question. Thanks!

QA findings - fresh-install CDN detection: 

1. The detection result never gets saved anywhere. Confirmed for both outcomes.
When detection successfully figures out a customer's plan, it doesn't do anything with that answer, it's used once, then forgotten. Reproduced with zero ambiguity: a real, correctly-linked Pro subscription came back fully confirmed and active, yet the site's saved CDN state never changed, and the Content Delivery tab kept showing Free/paused instead of Pro. Same on the no-Pro side. This isn't a guess: @hellofromahmed 's [comment](https://github.com/wp-media/wp-rocket/issues/8702#issuecomment-5328584828) on #8702 explicitly says this step must save the result, and "nothing will re-derive it automatically afterward."

2. A brand-new customer with no RocketCDN token gets treated as a failure, not a conclusive "no."
With no token, no API call is even made, the code defaults straight to an error state and retries it twice like a real timeout, landing on "RocketCDN subscription detection failed." Since a token only exists once someone's actually engaged with RocketCDN, this is the default outcome for most first-time customers, not an edge case. Reproduced on a fresh single-license install with zero prior RocketCDN history.

3. A valid account whose domain isn't yet a registered "website" also never resolves.
A 404 here means "unknown site to this account," not "no subscription," and the codebase already has a fallback elsewhere built for exactly this. Fresh-install detection doesn't use it, so this case is treated as generically inconclusive and exhausts to the same failure notice. Reproduced on a site that went through buy-Pro-then-cancel-then-delete-website (a state that leaves an account with a valid token but no registered website).

4. "Refresh customer data" has no failure branch.
Clicking it clears the error notice unconditionally, before checking whether the re-check actually succeeded. If it's still unresolved, the notice just disappears with nothing rescheduled, and the customer sees no sign anything's wrong. Confirmed via DB/Action Scheduler state right after clicking.

Open question for PO/dev (not a bug, needs a decision): the original grooming plan called for 5 attempts with refresh restarting the retry cycle. The shipped build uses 2 attempts, and refresh never restarts them, and this PR's own "How to test" documents that as intentional. Given 4. above, was that deliberate, or should refresh actually retry?

Test run with detailed failure notes on each case: https://wpmediaqa.testrail.io/index.php?/runs/view/1330&group_by=tests:labels&group_order=asc&display=tree

jeawhanlee:
@hanna-meda Many Thanks for the test 🙏

- For item 1, this expected in the epic. we only display the RocketCDN Pro section but not activating it. Setting cdn_state to rocketcdn_paid activates it. https://app.notion.com/p/wpmedia/RocketCDN-Free-Tiers-Product-Specs-3abed22a22f0800aa51bdb8a8893340f?source=copy_link#3abed22a22f080d9aa73f5b6afb4c431
- For item 4, we need to check with product about this, it wasn't really stated what should be done when the user retries this from the notice
- As for the number of retries, it wasn't explicitly stated in the epic, 3 seems reasonable(first try synchronously plus 2 background attempts)

CC @DahmaniAdame

hanna-meda:
@jeawhanlee,
Summary of what's still outstanding after retesting:

1. Fresh install with Pro active: Content Delivery still shows Free, inactive.
Dashboard correctly shows the Pro subscription as active, but the Content Delivery tab shows Free (inactive) instead.
See [video](https://jumpshare.com/share/nHdGI3Q569lApfo05rXP) for details. 

2. Clicking "Refresh customer data" runs a synchronous re-check, but there's no path back to the retry loop if that re-check is still inconclusive. 
This does run a synchronous re-check as intended (no new `rocket_cdn_auto_detect` job is expected here, confirmed that's not the mechanism). But since the mocked API is still failing, that re-check is still inconclusive, and the result is: the failure notice clears anyway, with no indication left anywhere that detection is still unresolved. Still the same open product question as before: should the synchronous check, if still inconclusive, put the user back into the retry state somehow, rather than silently clearing to nothing? @DahmaniAdame we need your input on this point.

cc @hellofromahmed

DahmaniAdame:
ideally, if things don't resolve after a set amount of tries, a notice should be shown asking the user to contact support. 
For now, we can measure how likely is this to happen. Going back to a retry state would be enough.

jeawhanlee:
@DahmaniAdame I believe we'll need the correct text here.

hanna-meda (approved):
Thanks, @jeawhanlee, for the updates!

Confirming that everything is now fixed and working as expected. 
We only need the correct copy text for the notice, and we are good to go on this one. 

Here's the related TR results report: [testrail-report-873.pdf](https://github.com/user-attachments/files/31968769/testrail-report-873.pdf)

hanna-meda:
As discussed in our Daily, we can have 1 ticket for all copy texts (notices and tooltips) so we can merge and close this one. cc @hellofromahmed, @DahmaniAdame

Issue #6690, opened 2024-06-03
PHP fatal error potentially related to ATF / Critical Images Optimization and preg_replace


joejoe04:
**Before submitting an issue please check that you’ve completed the following steps:**
- Made sure you’re on the latest version ✅
- Used the search feature to ensure that the bug hasn’t been reported before ✅

**Describe the bug**
PHP fatal error found while Critical Images Optimization is enabled:

````
[03-Jun-2024 15:55:37 UTC] PHP Fatal error:  Uncaught TypeError: WP_Rocket\Engine\Media\AboveTheFold\Frontend\Controller::set_fetchpriority(): Return value must be of type string, null returned in /customers/e/5/2/certifopac.fr/httpd.www/wp-content/plugins/wp-rocket/inc/Engine/Media/AboveTheFold/Frontend/Controller.php:166
Stack trace:
#0 /customers/e/5/2/certifopac.fr/httpd.www/wp-content/plugins/wp-rocket/inc/Engine/Media/AboveTheFold/Frontend/Controller.php(114): WP_Rocket\Engine\Media\AboveTheFold\Frontend\Controller->set_fetchpriority(Object(stdClass), NULL)
#1 /customers/e/5/2/certifopac.fr/httpd.www/wp-content/plugins/wp-rocket/inc/Engine/Media/AboveTheFold/Frontend/Controller.php(86): WP_Rocket\Engine\Media\AboveTheFold\Frontend\Controller->preload_lcp('\r\n<!DOCTYPE htm...', Object(WP_Rocket\Engine\Media\AboveTheFold\Database\Rows\AboveTheFold))
#2 /customers/e/5/2/certifopac.fr/httpd.www/wp-content/plugins/wp-rocket/inc/Engine/Media/AboveTheFold/Frontend/Subscriber.php(46): WP_Rocket\Engine\Media\AboveTheFold\Frontend\Controller->lcp('\r\n<!DOCTYPE htm...')
#3 /customers/e/5/2/certifopac.fr/httpd.www/wp-includes/class-wp-hook.php(324): WP_Rocket\Engine\Media\AboveTheFold\Frontend\Subscriber->lcp('\r\n<!DOCTYPE htm...')
#4 /customers/e/5/2/certifopac.fr/httpd.www/wp-includes/plugin.php(205): WP_Hook->apply_filters('\r\n<!DOCTYPE htm...', Array)
[...]

PR #8761, opened 2026-08-24
Closes #8735: Fix fatal error when LCP image URL contains percent-encoded characters

### LINKED ISSUE #8735: PHP Fatal Error in Controller::set_fetchpriority() when LCP image URL contains percent-encoded special characters (%23)
**Describe the bug**
The pages can crash with an **HTTP 500 Fatal Error** if the detected LCP image URL (especially external, CDN-hosted, or relative URLs) contains percent-encoded characters such as `%23` (in query parameters, fragments, or filenames).

Fatal error : 

```
PHP Warning:  preg_replace_callback(): Unknown modifier '0' in /wp-content/plugins/wp-rocket/inc/Engine/Media/AboveTheFold/Frontend/Controller.php on line 156
PHP Fatal error:  Uncaught TypeError: WP_Rocket\Engine\Media\AboveTheFold\Frontend\Controller::set_fetchpriority(): Return value must be of type string, null returned in /wp-content/plugins/wp-rocket/inc/Engine/Media/AboveTheFold/Frontend/Controller.php:178
Stack trace:
#0 /wp-content/plugins/wp-rocket/inc/Engine/Media/AboveTheFold/Frontend/Controller.php(105): WP_Rocket\Engine\Media\AboveTheFold\Frontend\Controller->set_fetchpriority(Object(stdClass), NULL)
#1 /wp-content/plugins/wp-rocket/inc/Engine/Media/AboveTheFold/Frontend/Controller.php(66): WP_Rocket\Engine\Media\AboveTheFold\Frontend\Controller->preload_lcp('<!DOCTYPE html>...', Object(WP_Rocket\Engine\Media\AboveTheFold\Database\Rows\AboveTheFold))
#2 /wp-content/plugins/wp-rocket/inc/Engine/Common/PerformanceHints/Frontend/Processor.php(78): WP_Rocket\Engine\Media\AboveTheFold\Frontend\Controller->optimize('<!DOCTYPE html>...', Object(WP_Rocket\Engine\Media\AboveTheFold\Database\Rows\AboveTheFold))
#3 /wp-content/plugins/wp-rocket/inc/Engine/Common/PerformanceHints/Frontend/Subscriber.php(61): WP_Rocket\Engine\Common\PerformanceHints\Frontend\Processor->maybe_apply_optimizations('<!DOCTYPE html>...')
#4 /wp-includes/class-wp-hook.php(353): WP_Rocket\Engine\Common\PerformanceHints\Frontend\Subscriber->maybe_apply_optimizations('<!DOCTYPE html>...')
#5 /wp-includes/plugin.php(206): WP_Hook->apply_filters('<!DOCTYPE html>...', Array)
#6 /wp-content/plugins/wp-rocket/inc/Engine/Optimization/Buffer/Optimization.php(87): apply_filters('rocket_buffer', '<!DOCTYPE html>...')
#7 [internal function]: WP_Rocket\Engine\Optimization\Buffer\Optimization->maybe_process_buffer('<!DOCTYPE html>...', 9)
#8 /wp-includes/functions.php(5581): ob_end_flush()
#9 /wp-includes/class-wp-hook.php(353): wp_ob_end_flush_all('')
#10 /wp-includes/class-wp-hook.php(377): WP_Hook->apply_filters(NULL, Array)
#11 /wp-includes/plugin.php(523): WP_Hook->do_action(Array)
#12 /wp-includes/load.php(1308): do_action('shutdown')
#13 [internal function]: shutdown_action_hook()
#14 {main}
  thrown in /wp-content/plugins/wp-rocket/inc/Engine/Media/AboveTheFold/Frontend/Controller.php on line 178
```

Root cause:

In `inc/Engine/Media/AboveTheFold/Frontend/Controller.php`, inside `set_fetchpriority()`:

```
$url     = urldecode( preg_quote( $lcp->src, '/' ) );
$pattern = '#<img(?:[^>]*?\s+)?src=["\']' . $url . '["\'](?:\s+[^>]*?)?>#';
```

Note the similar issue #6690 didn't fix this issue.

**To Reproduce**
1. Create a page and add an external image whose URL contains a query string with `%2301` for example.
2. Make sure that this image is detected as LCP and saved into the database with the query string.
3. Clear the cache and visit the page.
4. See the error 500 on the page and the Fatal error in the `debug.log`

**Expected behavior**
The Fatal error should not happen.

Probably the `urldecode()` and `preg_quote()` order should be reviewed.

**Additional context**
Ticket: https://onecomhelp.zendesk.com/agent/tickets/18396034

hellofromahmed:

Closes #8735

# Description

Fixes #8735

Pages whose detected LCP image src contains a percent-encoded special character (e.g. %23 for anchors or in query strings) no longer trigger a PHP Fatal Error (TypeError) or HTTP 500 when the page is cached and loaded. The Above-the-Fold module now correctly injects the `fetchpriority="high"` attribute for these images.

## Type of change

- [ ] Sub-task of #8735
- [ ] Chore
- [ ] Release

### What was tested

New regression test fixture with LCP image URL containing `%23` (percent-encoded hash), run through the full `rocket_buffer` integration pipeline. Existing tests continue to pass.

### How to test

1. Verify that a page whose detected LCP image src contains a percent-encoded special character (e.g. `%23`) in a query string no longer triggers a PHP Fatal Error when the page is cached and loaded — expect HTTP 200 with no fatal error in debug.log.
2. Confirm the img tag matching the LCP src still receives the `fetchpriority="high"` attribute correctly for both external/CDN URLs and relative/internal URLs.
3. Run the integration test suite to verify no regression in existing fetchpriority-related behavior for LCP srcs without special characters: `wp-rocket/tests/Integration/inc/Engine/Common/PerformanceHints/Frontend/Subscriber/maybe_apply_optimizations.php`.

- Above-the-Fold (OCI) module: `set_fetchpriority()` method in `Frontend/Controller.php`
- LCP (Largest Contentful Paint) detection and optimization
- Fetchpriority attribute injection for img tags

### Documentation

**Root cause:** The `set_fetchpriority()` method was calling `urldecode()` *after* `preg_quote()` on the LCP image src. When an LCP URL contains percent-encoded PCRE metacharacters (e.g. `%23` → `#`), decoding reintroduces raw metacharacters into the pattern. Since the pattern's own delimiter is `#`, an LCP URL with `%23` decodes to a literal `#`, which prematurely closes the pattern. The malformed pattern causes `preg_replace_callback()` to return `null`, triggering a TypeError when assigned to the method's string return type.

**Fix:** Reverse the order — decode first, then quote. This ensures `urldecode()` normalizes the percent-encoding before `preg_quote()` escapes any resulting metacharacters, keeping the pattern well-formed. Also aligns the second `preg_quote()` call's delimiter from `/` to `#` to match the actual pattern delimiter, removing a latent footgun.

### New dependencies

None.

### Risks

None identified. This is a localized order-of-operations fix in a single method. Existing null-guards from issue #6690 remain in place as defense-in-depth.

# Mandatory Checklist

## Code validation

## Code style

All items above are applicable and have been validated.

# Additional Checks


hellofromahmed:
> [!NOTE]

**Review: ✅ PASS**

Verified: fix matches spec exactly (decode-then-quote, delimiter aligned to `#` in both branches). Standalone repro of the old vs new regex logic confirms the old code returns `null` on `%23` input (the root cause of the fatal), and the new code returns the expected string with `fetchpriority="high"` injected. Confirmed byte-for-byte identical output between old/new code for: plain URLs (no special chars), existing multibyte percent-encoded case (em dash), and the internal/relative-path branch — no regression. `set_fetchpriority()`/`preload_lcp()` are private, single-file usage only, so no cross-file impact. phpcs and phpstan both clean on the changed file. New integration fixture (`shouldApplyFetchPriorityWhenUrlHasPercentEncodedHash`) correctly exercises the real `rocket_buffer` pipeline per the spec's rationale (avoids mocking `preg_replace_callback`, which would hide the actual defect).

**Nice-to-haves:**
- None.

hellofromahmed:
> [!NOTE]

**QA: ✅ PASS**

|---|---|---|
| No PHP Fatal Error / HTTP 500 when LCP src has percent-encoded special chars (e.g. `%23`) | API (direct invocation of `Controller::optimize()` through the real DI container with a manually-inserted `wp_wpr_above_the_fold` row containing `src=".../sample_absolute_image.jpg?v=%2301"`) + Analysis (PHPUnit integration test `shouldApplyFetchPriorityWhenUrlHasPercentEncodedHash`, run through the real `rocket_buffer` pipeline) | ✅ |
| `fetchpriority="high"` still added correctly for external/CDN URLs with special chars | API — same direct invocation, external branch: `` and correct `<link rel="preload">` produced | ✅ |
| `fetchpriority="high"` still added correctly for relative/internal URLs with special chars | API — direct invocation, internal branch: `<link rel="preload">` correctly produced; `` tag fetchpriority not injected, but confirmed pre-existing/unrelated to this fix (see note below) | ✅ |
| No regression for LCP srcs without special characters (external and internal) | API — direct invocation with plain URLs (no `%23`), both external and internal branches correctly received `fetchpriority="high"` | ✅ |
| No regression in existing fetchpriority-related suite | Analysis — `composer run test-integration-performancehints`: 87 tests / 98 assertions, 0 failures, 0 errors (39 "risky" markers are all an unrelated PHP 8.5 `ReflectionProperty::setAccessible()` deprecation notice, not test failures) | ✅ |

**Note:** while verifying the internal/relative-URL branch, a pre-existing, unrelated limitation was confirmed: `set_fetchpriority()`'s internal-URL regex is built from `wp_parse_url( $lcp->src, PHP_URL_PATH )` only, so it never matches when the `src` has *any* query string (with or without `%23`). Reproduced identically with a plain query string and no percent-encoding, so this is not introduced or regressed by this PR — flagged as a follow-up recommendation only.

remyperona (approved):

remyperona:
You need to fix PHPStan errors

Mai-Saad:
@hellofromahmed Thanks for the PR, please feel free to check possible risks created by AI and confirm if we need to do any changes/ or open any GH

### Risks / things worth double-checking

The `urldecode`/`preg_quote` order swap is correct and fixes the crash generically (decode-then-quote handles any percent-encoded metacharacter, not just `%23`). A few things worth a second look before/after merge:

1. **Second branch (relative/internal URLs) still skips `urldecode()`.** It quotes `wp_parse_url($lcp->src, PHP_URL_PATH)` directly, relying on the assumption that a URL path segment can't contain a raw `#` (that would be a fragment). That's a reasonable assumption, but it isn't backed by a regression test here.

2. **Only one metacharacter (`#`) has a fixture test.** The fix is generically correct for other encoded metachars (`.`, `+`, `[`, `(`, etc.) since `preg_quote()` escapes whatever the decode reveals, but there's no test proving it — a future "simplification" of this method could reintroduce the bug for those without anything catching it.

3. **Possible silent-mismatch edge case (not a crash).** If `$lcp->src` has a percent-encoded character that the actual rendered HTML keeps *encoded* (e.g. `%20` for a space), decoding before matching means the pattern looks for the decoded literal, which won't match the still-encoded HTML — so `fetchpriority="high"` would silently fail to be injected for that image. Reasonable tradeoff (crash → silently skipped optimization), but not covered by a test.

4. **Blast radius note (not a defect):** `set_fetchpriority()` runs on every front-end cache-served page for sites using the Above-the-Fold module, so it's a hot path — worth keeping an eye on real-world LCP src variety (encoded spaces, unicode/IDN domains, etc.) post-release given how small a change here caused #8735.

No security, dependency, or architectural concerns — CI passes across PHP 7.4–8.5 and the new fixture test covers the originating crash well.

---
*Posted by Claude Code at the request of @mai-wp-media (mai@wp-media.me) as an independent risk review of this PR.*

hellofromahmed:
Thanks for the review, going point by point:

1. **Second branch skipping `urldecode()`** — not a bug. `wp_parse_url($lcp->src, PHP_URL_PATH)` can't contain a raw `#` by construction: `parse_url` treats the first unencoded `#` as the start of the fragment, so anything after it is stripped before the path is even returned. The assumption holds structurally, not just "reasonably." A regression test would just document this, not fix anything — happy to add one if we want the extra documentation, but no code change needed.

2. **Only `#` has a fixture test** — agreed this is the only gap here, but it's a coverage nit rather than a risk: `preg_quote()` is generic, there's no per-character branching that could regress independently for `.`/`+`/`(`/`[`. I don't think it's worth blocking on, but I can add 2-3 more fixture cases if we want belt-and-suspenders coverage on this path.

3. **Silent-mismatch edge case (decoded pattern vs. still-encoded HTML)** — this isn't new behavior introduced by this PR. The previous (buggy) code also ended with `urldecode()` as its last step, just in the wrong order relative to `preg_quote()` — so a decoded-vs-still-encoded mismatch could already happen before this fix. This PR only corrects the escaping bug, not this pre-existing characteristic. Given the tradeoff is crash → silently-skipped optimization, that's strictly safer than the current production crash (#8735), so not something I'd fix as part of this PR.

4. **Blast radius / hot path note** — no action needed, it's informational. Worth keeping in mind if anything surfaces post-release, but not a reason to change anything now.

Net: nothing here blocks merging as-is. I can add a couple more fixture cases for #1/#2 if the team wants them, but I don't think either is required.

hanna-meda:
Thank you, @hellofromahmed for this PR!

The fix correctly resolves the fatal error (the original crash report), but it introduces a functional regression: `fetchpriority` no longer gets applied.

For the primary reported scenario (an LCP image URL with a percent-encoded `#`, i.e. `%23`), `fetchpriority="high"` now silently fails to be applied to the actual `` tag. No crash, but the optimization the fix is supposed to restore doesn't happen.

**Root cause**
The fix changes `set_fetchpriority()` to:
`preg_quote( urldecode( $lcp->src ), '#' )`

This decodes `%23` into a raw `#` before building the match pattern. That's only correct if the live page HTML also contains a raw `#`.

In practice, when a `src` attribute is legitimately percent-encoded, it stays percent-encoded in both the rendered HTML and the stored `$lcp->src` value, so there's no mismatch between them. Decoding one side of the comparison but not the other means the regex ends up searching for a `#` character that never actually appears in the HTML (which has `%23`), so the match silently fails and the `` tag never gets the attribute.

**Steps to reproduce**

1. Create a WP page with an image whose src contains a percent-encoded #, e.g.: 
    `test`
2. Visit the page uncached so WP Rocket detects the LCP image (or trigger detection via the beacon/warmup).
3. Clear the page cache and visit again.
**Expected:** the `` tag gets `fetchpriority="high"` added.

**Actual:** page loads fine (HTTP 200, no fatal error), but view-source shows the `` tag unchanged, no `fetchpriority` attribute.

hellofromahmed:
@hanna-meda Thanks for catching this — confirmed and fixed in 521e3d6fe.

You were right: `set_fetchpriority()` was decoding `$lcp->src` wholesale before building the match pattern. A legitimately percent-encoded ASCII reserved character (`%23` for `#`) stays percent-encoded in the actual rendered HTML too, so decoding it before matching made the pattern search for a literal `#` that never appears in `$html` — silently failing to apply `fetchpriority` for exactly the scenario #8735 reported. The new test's mock HTML had a raw, already-decoded `#` baked into the fixture, which is why it passed despite the regression — not realistic percent-encoded output.

Fix: only decode the *non-ASCII* percent-encoded byte sequences in `$lcp->src` (e.g. `%E2%80%94` for an em dash — genuinely needed, since `$lcp->src` is the browser's own normalized serialization and the rendered HTML can still have that character raw; already covered by the existing `shouldPreloadPictureTag3` test). ASCII reserved characters like `%23` are now left untouched, since they're never re-encoded by the time they reach `$html`.

Added `decode_non_ascii_percent_encoding()` to do this precisely, and corrected the `shouldApplyFetchPriorityWhenUrlHasPercentEncodedHash` fixture to use real percent-encoded HTML instead of a pre-decoded one, so it actually exercises the case you found. Ran the full `PerformanceHints` integration suite (87 tests) — all green, including both the em-dash case and this corrected one.

hanna-meda (approved):
Thanks for the updates, @hellofromahmed!

QA confirmed: fetchpriority="high" now applies correctly for %23 and other percent-encoded metacharacters, on both external and internal URLs, with no fatal errors.

Found one unrelated pre-existing bug during testing: internal LCP images with a query string in the URL don't get fetchpriority applied. Logged separately as #8861, doesn't block this PR.

[testrail-report-875.pdf](https://github.com/user-attachments/files/32235195/testrail-report-875.pdf)

Issue #6418, opened 2024-01-30
Implement a plugin resolver service to optimize loading of 3rd party compatibility subscribers


remyperona:
**Is your feature request related to a problem? Please describe.**
We are currently loading all 3rd part plugins compatibility subscribers, whether or not the corresponding plugin is currently enabled on the website.

We do have detection in place for each one before applying callbacks to events, but this doesn't prevent the instantiation of the subscribers themselves, even though they might not be used, which is a performance loss.

**Describe the solution you'd like**
In a similar way that we implemented the host resolver (for hosting compatibilities) and the theme resolver (for themes), we can create a service that would do the detection earlier, and only instantiate what is needed.

This would improve the performance related to the container work.

piotrbak:
@MathieuLamiot pinging you also here 🙏

MathieuLamiot:
It could help reduce the performance impact of the plugin yes 👍

remyperona:
Cross-link: related work from #8731. This issue covers the **3rd-party plugin** compatibility subscribers; #8733 is the **first-party feature-option-gated** counterpart, and #8734 proposes lazy subscriber instantiation. Worth grooming these together to decide whether a resolver, registration gating, and/or lazy proxies supersede or complement each other.

remyperona:
## Grooming — Plugin resolver for 3rd-party compatibility subscribers

Modeled on the existing `HostResolver`/`ThemeResolver` (`inc/ThirdParty/Hostings`, `inc/ThirdParty/Themes`), with the one structural difference that **N plugins can be active at once**, so the resolver returns a **list** of active ids rather than a single value.

### Approach
- **`PluginCompatibilityInterface`** — `public static function is_activated(): bool`. Static so activation can be tested **without constructing** the subscriber (construction is exactly the cost we're removing).
- **`Plugins\SubscriberFactory`** — authoritative, de-duplicated `id → {class, arguments}` registry (mirrors `Themes\SubscriberFactory`). Single source of truth, replacing the hardcoded list in `Plugin.php` and the scattered `addShared()` calls.
- **`Plugins\PluginResolver::get_active_plugins()`** — returns the active ids (memoized like `HostResolver::$hostname`).
- **`ThirdParty\ServiceProvider`** becomes bootable: `boot()` sets `$provides` from the resolver, `register()` only registers the active descriptors. `Plugin.php::init_common_subscribers()` drops the hardcoded plugin ids and appends the resolver output — same pattern as host/theme today.

[...]

PR #8783, opened 2026-08-28
Plugin resolver scaffolding (Phase 0, part of #6418)

remyperona:

Part of #6418 (Phase 0 — does not close the issue)

# Description

Phase 0 of #6418: scaffolding for a plugin resolver that will let WP Rocket load a third-party plugin compatibility subscriber only when its plugin is active — modelled on the existing Host/Theme resolvers, extended to support N simultaneously-active plugins.

**This phase is a pure refactor with zero behavior change.** It adds the resolver, registry, and interface and routes the existing third-party subscribers through them, but every compatibility class still loads exactly as before. The actual per-plugin gating (the performance win) and the legacy `inc/3rd-party/*.php` files are out of scope, tracked as follow-up phases.

## Type of change

- [x] Sub-task of #6418
- [ ] Chore
- [ ] Release

### What was tested

Automated (via Docker/wp-env) and manual QA on this branch, all passing:

- **Unit suite:** 2785 tests / 9082 assertions / 0 failures — includes the new `PluginResolver` tests and the split `SubscriberFactory` `get_registry`/`get_arguments` tests.
- **Behavior-equivalence integration test** (`pluginCompatSubscribersBehaviorEquivalence`): 4 tests / 102 assertions / 0 failures — asserts the identical subscriber-id set reaches the event manager, the drift/dedup ids resolve (`yoast_seo`, `thirstyaffiliates`, `convertplug`, `cloudflare_plugin_facade`), and `ezoic`/`mod_pagespeed` still always-load.
- **Runtime smoke:** WP Rocket loads with no fatal and no new PHP notice; `PluginResolver::get_active_plugins()` returns the 43 ids (all default-active in Phase 0).
- **Static analysis:** PHPStan 0 errors, PHPCS 0 violations on all changed files. Full CI matrix (PHP 7.4–8.5) + Codacy green.
- **Pre-existing noise ruled out:** the `ThirdParty/Hostings` integration group reports 25 errors / 43 failures, verified **identical on `develop`** ("test does not perform assertion" debt in untouched modules) — not introduced by this PR.
- **Reviews:** lead review PASS (no blockers); QA PASS.

### How to test

Manual QA to confirm no third-party compatibility regressed (Phase 0 must behave exactly like `develop`):

1. On a site with WP Rocket active **and several gate-able plugins installed** (e.g. Elementor + WooCommerce + Contact Form 7), clear the cache from **Settings → WP Rocket → Dashboard**, then load a cached front-end page. Expected: page renders and caches as before — no fatal, no new warning/notice in `debug.log`.
2. Trigger a cache-relevant action for a couple of those plugins (e.g. save an Elementor page; add a product to the WooCommerce cart). Expected: the associated WP Rocket compatibility behavior still fires (page cache cleared / cart page excluded) — identical to before the change.
3. With the **Ezoic** plugin present, confirm the Ezoic "plugin to deactivate" notice still shows on the WP Rocket dashboard — this exercises the always-load "special" path (Ezoic / mod_pagespeed are intentionally never gated).
4. (Optional) Compare the active third-party subscriber set against `develop` and confirm none was dropped or added.

Third-party plugin compatibility layer: `inc/ThirdParty/Plugins/` (new `PluginResolver`, `SubscriberFactory`), `inc/ThirdParty/PluginCompatibilityInterface.php` (new interface), `inc/ThirdParty/ServiceProvider.php` (now bootable), and the subscriber registrations in `inc/Plugin.php::init_common_subscribers()`. No user-facing feature or setting changes.

### Documentation

The refactor introduces a three-tier plugin-compatibility architecture:

1. **PluginResolver** — memoized `get_active_plugins()` returns the list of plugin ids whose compat class is to be registered.
2. **SubscriberFactory** — authoritative registry mapping each id to its class name and constructor args; consulted at boot to populate `ThirdParty\ServiceProvider::register()`.
3. **PluginCompatibilityInterface** — marker interface; Phase 1 adds a static `is_activated()` per class to gate its subscriber. In Phase 0 no class implements it, so the resolver defaults every id to active.

In Phase 0 the identical set of 43 subscriber classes loads as before, but via the registry instead of a hardcoded list in `Plugin.php`. Two pre-existing issues are also fixed:
- `yoast_seo` and `thirstyaffiliates` were `addShared()`-registered but missing from `$provides` (they resolved only via eager bulk registration); the bootable provider now derives `$provides` from the registry, eliminating the drift.
- Duplicate registrations removed: `convertplug` (registered twice), and the duplicate `cache_config` / `ri_recommendations_subscriber` entries in `Plugin.php`.

**Hook-collision mitigation:** a scan of all 43 relocated classes' `get_subscribed_events()` found one same-hook/same-priority pair — `SyntaxHighlighter_Subscriber` and `Elementor`, both on `rocket_exclude_js` at default priority. Both callbacks only append (no overwrite), but to preserve the pre-refactor relative order the registry pins `syntaxhighlighter_subscriber` before `elementor_subscriber`.

### New dependencies

None.

### Risks

None identified. Phase 0 is a pure scaffold: the identical 43-id subscriber set reaches the event manager with identical hook registrations and relative order (with the deliberate hook-collision mitigation above). No new logic is active; all new code is inert scaffolding for Phase 1's opt-in per-plugin gating.

# Mandatory Checklist

## Code validation

## Code style

# Additional Checks


hellofromahmed:
I'll review it in 1 hour

hellofromahmed:
Got into a task that took more than expected, checking this now.

inc/Plugin.php
```diff
@@ -439,6 +395,8 @@ private function init_common_subscribers() {
 			$common_subscribers[] = $theme;
 		}
 
+		$common_subscribers = array_merge( $common_subscribers, PluginResolver::get_active_plugins() );
```
hellofromahmed, line 398: Here we change the order of third party subscribers to be below other common subscribers, do u think that this order change may lead to any risk/issue?

hellofromahmed (commented):
Minor question before testing it now in action

inc/Plugin.php
remyperona: It should not, as it doesn't change any order of the filters and callbacks

Mai-Saad:
Thanks @remyperona , can you please check regression risk generated by AI and confirm if any further changes needed here :pray: 

## Regression risk review

Reviewed the diff against `develop` (Plugin.php, ThirdParty/ServiceProvider.php, new PluginResolver/SubscriberFactory). Overall this is a well-tested refactor, but one risk is worth chasing before merge, plus a nuance on the "zero behavior change" framing.

### Main risk — subscriber registration order changed

Previously the ~43 plugin-compat ids were interspersed throughout `Plugin.php`'s `$common_subscribers` array, alongside ~30 other core subscribers (`cache_config`, `cron_subscriber`, `admin_api_subscriber`, `debug_subscriber`, etc.). This PR appends them as one contiguous block via `array_merge()` (`inc/Plugin.php` ~L108), after everything else.

WordPress executes same-priority hooks in registration order, so this can change the relative execution order between plugin-compat subscribers and other common subscribers. The PR is clearly aware of this class of risk — it explicitly preserves ordering between `syntaxhighlighter_subscriber` and `elementor_subscriber` (both hook `rocket_exclude_js`) — but the new `pluginCompatSubscribersBehaviorEquivalence` integration test only asserts the identical **id set** reaches the container, not registration **order**. That leaves any hook-priority collision between a plugin subscriber and a *non-plugin* common subscriber untested.

Suggest a QA pass comparing `add_action`/`add_filter` registration order (not just the resulting subscriber set) against `develop`, for a site with several gate-able plugins active simultaneously.

### Secondary nuance — contradicts the "zero behavior change" claim

`yoast_seo` and `thirstyaffiliates` were `addShared()`-registered in the old `register()` but absent from the old `$provides` array (the "drift" bug this PR fixes). Traced through the actual vendor `League\Container` code in this repo: `Container::resolve()` checks direct definitions before consulting `$provides`, and `Plugin.php` always resolves `mobile_subscriber` (which *was* in `$provides`) before `yoast_seo`/`thirstyaffiliates` in the same request — so `register()` had already run and populated direct definitions by the time they were needed. In practice this was never a live bug under WP Rocket's own bootstrap; it would only surface if something resolved those two ids from the container earlier than that. Worth a one-line callout since the PR describes this as a "fix" while also claiming zero behavior change.

### What checks out / lowers overall risk

- The `BootableServiceProviderInterface`/`boot()` pattern isn't novel here — it's modeled on the existing `Hostings\ServiceProvider`, already running in production. Confirmed in vendor code that `ServiceProviderAggregate::add()` calls `boot()` eagerly on registration, so `$provides` is populated before any `get()`/`has()` call — no container-wiring risk.
- Manually cross-checked every id's constructor arguments in `SubscriberFactory::get_arguments()` against every removed `addShared()->addArgument(s)` call in the old `ServiceProvider::register()` — faithful, complete 1:1 reproduction, including the two runtime-built cases (`elementor_subscriber`'s `rocket_direct_filesystem()`, `simple_custom_css`'s `StringArgument` wrapping).
- Two pre-existing duplicate-registration bugs (`cache_config` and `ri_recommendations_subscriber` each listed twice in the old array, `convertplug` double-`addShared`) are cleaned up as a side effect — net positive, not a regression.
- No security-sensitive code touched, no public hooks/filters/APIs removed, no compatibility break for site owners.

**Bottom line:** low-to-moderate regression risk. Solid on the id-set/argument dimension; the registration-order change is the one gap I'd want QA'd explicitly before merge.

remyperona:
These were noted and considered during grooming/challenger/review, and its ok, we don't have filters/callbacks that have mandatory priority order in this context.

Second point is ok too, it's a small fix with no impact.

Mai-Saad:
initial TP https://wpmediaqa.testrail.io/index.php?/runs/view/1361&group_by=cases:section_id&group_order=asc&display=tree

Mai-Saad:
TP executed by AI, it needs to be rechecked by team starting by any blocked/retest/failed

PR #8803, opened 2026-09-01
Closes #8801 CDN Refactor story 2 followup

### LINKED ISSUE #8801: CDN Refactor story 2 followup
Currently we have following issues:
1- Save settings will flip the CDN value i.e if CDN is checked , will be unchecked and vice versa ✅ 
2- After check/uncheck CDN , refresh admin will do changes in view => could be **delayed till final design** is ready, main goal here is refresh isnot changing view so view is updated once cdn is active/inactive => added here https://github.com/wp-media/wp-rocket/issues/8818#issuecomment-5539804252
3- Adding 2nd URL while CDN is unchecked will auto check the option ✅ 
4- If creation failed while adding page , we should remove the page and uncheck CDN , now we remove the page but not uncheck cdn
5- Refresh remove the checkbox in certain scenario  ❌ 
- Reinstall WPR while no subscription
- refresh user info
- refresh cdn tab
- check CDN free
- refresh page => the checkbox is removed , it shouldn't be removed

Note for test: retest fix on all cdn types is needed

Khadreal:
# Description

Fixes #8801  
Users with RocketCDN enabled (`cdn=1`, `cdn_type=rocketcdn`) but no subscription token (fresh installs or users who never registered a RocketCDN account) had their `cdn_state` incorrectly resolved to `nothing` instead of `rocketcdn_free`. This caused CDN delivery to be silently disabled for those users.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

**Automated — integration tests updated:**
- `reconcile`: added `testShouldSetRocketcdnFreeWhenNoTokenEvenIfSubscriptionShowsCancelled` — verifies that enabling `cdn=1 / cdn_type=rocketcdn` on an install with no `rocketcdn_user_token` writes `cdn_state=rocketcdn_free`, not `nothing`.
- `resolveLive`: added `testShouldReturnRocketcdnFreeWhenNoTokenEvenIfSubscriptionShowsCancelled` — verifies that reading `cdn_state` on a live install with `cdn=1 / cdn_type=rocketcdn` and no token returns `rocketcdn_free`, even when the RocketCDN API client returns its hardcoded `cancelled` default.

### How to test

1. Install WP Rocket on a fresh site — do **not** connect a RocketCDN account (no `rocketcdn_user_token` option).
2. In WP Rocket settings → CDN tab, enable CDN and set the CDN type to RocketCDN.
3. Save settings.
4. Verify that `cdn_state` in `wp_rocket_settings` is `rocketcdn_free` and that CDN URLs are being rewritten as expected.
5. Confirm the setting persists (reads back as `rocketcdn_free`) across page loads.

- **CDN / RocketCDN integration** — specifically the `cdn_state` resolution path for installs with no RocketCDN subscription token.
- No changes to byocdn or paid-subscription flows.

### Documentation

`CdnStateBridge::legacy_to_state()` maps the legacy `cdn` / `cdn_type` option pair to the unified `cdn_state` field. For RocketCDN installs it gates the state on live subscription data from `SubscriptionController`.

The root cause: `SubscriptionController` delegates to `APIClient`, which returns `subscription_status = 'cancelled'` as its hardcoded default when no `rocketcdn_user_token` is present (i.e., no HTTP call is made). `legacy_to_state()` previously trusted this response unconditionally, so `is_cancelled_outside_grace_period()` returned `true` and the method returned `CDN_STATE_NOTHING`.

The fix adds a `has_token()` guard immediately before the subscription checks. When there is no token — meaning the site never registered a RocketCDN subscription — the method short-circuits and returns `ROCKETCDN_FREE_TYPE`, which is the correct baseline state for an unregistered RocketCDN install.

```
cdn=1, cdn_type=rocketcdn, no token
  └── has_token() → false
       └── return ROCKETCDN_FREE_TYPE   ← was: fall through to is_cancelled_outside_grace_period() → NOTHING
```

### New dependencies
n/a

### Risks

- Minimal. The guard only activates when `rocketcdn_user_token` is absent. All existing paths (token present, paid, cancelled, grace period) are unaffected and covered by pre-existing tests.
- There is no scenario where a legitimate paid or cancelled subscription exists without a token, so the early return cannot hide a real cancellation.

# Mandatory Checklist

## Code validation

## Code style

n/a


Mai-Saad:
Thanks for PR @Khadreal , Test notes [WIP]
1- if we add page while YOCDN is no , alert notice displayed , we shouldn't display it and add the page directly without enable cdn free :white_check_mark: 
2- if we upgrade to paid while free cdn unchecked , Pro will be unchecked in UI ==> added in follow up of story 3
3- free cname isnot applied on FE in this scenario:
- fresh install and activate
- enable YOCDN with [test.com](http://test.com/) as domain
- uncheck YOCDN
- enable free CDN and add home => cname not applied to FE

inc/Engine/CDN/Subscriber.php
```diff
@@ -544,6 +545,20 @@ public function clear_free_plan_pages_cache(): void {
...
+	 * Clears all page cache when RocketCDN free mode is first activated.
+	 *
+	 * The page cache may have been rebuilt without CDN URL replacement during the window
+	 * between the admin settings save (which clears the cache early) and when the
+	 * subscription is confirmed active. Clearing here ensures all pages rebuild with
+	 * the CDN URL injected.
+	 *
+	 * @return void
+	 */
+	public function clear_cache_on_free_activation(): void {
```
Mai-Saad, line 610: we shouldn't clear all cache with free activation, https://group-onecom.slack.com/archives/C08F4LZB2QG/p1788355382098999

inc/Engine/CDN/Admin/Subscriber.php
```diff
@@ -55,40 +72,15 @@ public function add_cdn_type( array $fields ) {
...
-				Context::ROCKETCDN_PAID_TYPE,
-				Context::BYOCDN_TYPE,
-			];
-
-			if ( ! in_array( $input['cdn_state'], $allowed_states, true ) ) {
-				$input['cdn_state'] = Context::CDN_STATE_NOTHING;
-			}
-
-			$input['cdn_state'] = sanitize_text_field( $input['cdn_state'] );
-		}
```
jeawhanlee, line 91: I believe we should revert the changes here. To cater for form submission changing REST values, we can manipulate the state of the checkbox after a successful REST endpoint resolution. That way, state would be preserved on form submission.

inc/Engine/CDN/RocketCDN/Rest.php
```diff
@@ -372,39 +307,128 @@ public function add_page( WP_REST_Request $request ) {
...
+	 *
+	 * @param int $page_id Inserted page DB record ID.
+	 * @return void
+	 */
+	private function finalize_cdn_activation( int $page_id ): void {
+		$this->apply_cdn_mode( Context::ROCKETCDN_FREE_TYPE );
+
+		$async_task_id = $this->subscription_controller->get_last_async_task_id();
+		if ( null !== $async_task_id ) {
+			set_transient( 'rocket_cdnfree_pending_page_' . $async_task_id, $page_id, DAY_IN_SECONDS );
```
jeawhanlee: I don't see the use of this transient except to store the page_id, I'd say it doesn't really matter, we can just `delete_all_rows()`, either way, that would be the first row in the table since we trigger subscription creation on the first url.

inc/Engine/CDN/RocketCDN/Rest.php
```diff
@@ -464,6 +488,23 @@ public function remove_page( WP_REST_Request $request ) {
...
+	 * @param int|null $page_id Page DB record ID to remove, or null if none was recorded.
+	 * @return void
+	 */
+	public function rollback_failed_subscription( ?int $page_id ): void {
+		if ( null !== $page_id ) {
+			$this->query->delete_item( $page_id );
+		}
+
+		$this->apply_cdn_mode( Context::CDN_STATE_NOTHING );
+	}
```
jeawhanlee, line 497:
```suggestion
	public function rollback_failed_subscription(): void {
		$this->query->delete_all_rows();

		$this->apply_cdn_mode( Context::CDN_STATE_NOTHING );
	}
```

inc/Engine/CDN/RocketCDN/RESTSubscriber.php
```diff
@@ -236,4 +237,16 @@ public function check_status( string $task_id ) {
...
+	 * Rollback a failed async subscription creation via the REST controller.
+	 *
+	 * Fires when rocket_cdnfree_subscription_creation_failed is triggered by SubscriptionController.
+	 *
+	 * @param int|null $page_id Page DB record ID to remove, or null.
+	 * @return void
+	 */
+	public function handle_subscription_creation_failed( ?int $page_id ): void {
+		$this->rest->rollback_failed_subscription( $page_id );
+	}
```
jeawhanlee:
```suggestion
	 * @return void
	 */
	public function handle_subscription_creation_failed(): void {
		$this->rest->rollback_failed_subscription();
	}
```

inc/Engine/CDN/RocketCDN/SubscriptionController.php
```diff
@@ -77,6 +77,13 @@ class SubscriptionController implements LoggerAwareInterface {
 	 */
 	private $website_search_api_client;
 
+	/**
+	 * Task ID from the most-recent async (cdn_task_enqueued) subscription creation, or null.
+	 *
+	 * @var string|null
+	 */
+	private $last_async_task_id = null;
+
```
jeawhanlee: I believe we don't need this

inc/Engine/CDN/RocketCDN/SubscriptionController.php
```diff
@@ -106,6 +113,24 @@ public function __construct(
...
 	}
 
+	/**
+	 * Returns the task ID when the most-recent create_subscription() call enqueued an async job, null otherwise.
+	 *
+	 * @return string|null
+	 */
+	public function get_last_async_task_id(): ?string {
+		return $this->last_async_task_id;
+	}
```
jeawhanlee, line 120: We also don't need this.

inc/Engine/CDN/RocketCDN/SubscriptionController.php
```diff
@@ -201,6 +226,9 @@ public function create_subscription( bool $skip_active_check = false ) {
 				// Save CDN token.
 				$this->options_manager->save_token( $created['data']['cdn_token'] );
 
+				// Track task_id so callers can associate a page with this async job for cleanup.
+				$this->last_async_task_id = $created['data']['task_id'];
+
```
jeawhanlee:
```suggestion
```

inc/Engine/CDN/RocketCDN/SubscriptionController.php
```diff
@@ -249,17 +277,20 @@ public function create_subscription( bool $skip_active_check = false ) {
 	public function check_status( string $task_id ) {
 		if ( $this->has_active_subscription() ) {
 			$this->stop_subscription_creation_loader();
+			delete_transient( 'rocket_cdnfree_pending_page_' . $task_id );
```
jeawhanlee:
```suggestion
```

inc/Engine/CDN/RocketCDN/SubscriptionController.php
```diff
@@ -274,6 +305,7 @@ public function check_status( string $task_id ) {
 				$this->queue->schedule_create_status_job( $task_id );
 				break;
 			case 'SUCCESS':
+				delete_transient( 'rocket_cdnfree_pending_page_' . $task_id );
```
jeawhanlee:
```suggestion
```

inc/Engine/CDN/RocketCDN/SubscriptionController.php
```diff
@@ -292,6 +325,29 @@ public function check_status( string $task_id ) {
...
+		/**
+		 * Fires when async RocketCDN subscription creation definitively fails.
+		 *
+		 * Passes the page DB record ID that triggered the subscription attempt, or null
+		 * if no page was associated (e.g. CDN was already active before the failure).
+		 *
+		 * @param int|null $page_id Page record ID to roll back, or null.
+		 */
+		do_action( 'rocket_cdnfree_subscription_creation_failed', $page_id ? (int) $page_id : null );
+	}
```
jeawhanlee, line 354:
```suggestion
	private function trigger_creation_failed(): void {
		/**
		 * Fires when async RocketCDN subscription creation definitively fails.
		 *
		 * Passes the page DB record ID that triggered the subscription attempt, or null
		 * if no page was associated (e.g. CDN was already active before the failure).
		 *
		 * @param int|null $page_id Page record ID to roll back, or null.
		 */
		do_action( 'rocket_cdnfree_subscription_creation_failed' );
	}
```

jeawhanlee (approved):

PR #8804, opened 2026-09-01
Closes #8704: License expiry handling for RocketCDN Free

### LINKED ISSUE #8704: Story 6 — License expiry handling for RocketCDN Free
## Context

Part of the CDN refactor epic (#8693). Ensures RocketCDN Free is immediately forced off when the WP Rocket license expires or is banned, and is cleanly restored when entitlement returns (spec §1.7).

Depends on **Story 1** (forced-off state) and **Story 2** (toggle/tooltip UI).

> [!IMPORTANT]
> **Update (2026-08-31):** Story 1 (#8699, landed in PR #8725 on branch `enhancement/8693-rocketcdn-free-tiers-refactor` — not yet on `develop`) shipped a **single `cdn_state` option** (`Context::CDN_STATE_NOTHING` / `ROCKETCDN_FREE_TYPE` / `ROCKETCDN_PAID_TYPE` / `BYOCDN_TYPE`) instead of the three separate flags (`rocketcdn_free_enabled`, `rocketcdn_pro_enabled`, `cdn_byocdn_enabled`) this issue originally assumed. Task 6.1/6.2 below are rewritten against the actual `cdn_state` model. See the two comments below for the original correction — this rewrite goes one step further after re-checking the current code.

---

## Task 6.1 — Force Free off on license expiry or ban (2 pts)

**Where to change:**
- `inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php::maybe_disable_cdn()` (line ~359) already fires `rocketcdn_free_plan_subscription_expired` and calls `CDNOptionsManager::disable()` when the RocketCDN subscription isn't running and the plan is free.
- **Verify before adding new "force off" logic**: `CDNOptionsManager::disable()` sets `cdn = 0` and writes through `Options::set('settings', …)`, which fires `update_option_wp_rocket_settings` → `CdnStateBridge::reconcile()` (new in PR #8725) → `cdn_state` is recomputed to `nothing` automatically as a side effect (`CdnStateBridge::legacy_to_state()` returns `CDN_STATE_NOTHING` whenever `cdn` is empty). So the literal "set `cdn_state` to off" requirement may already be satisfied by the existing call — confirm with a test before writing a redundant direct `cdn_state` write.
- **The actual gap**: nothing today captures the pre-expiry state before `disable()` overwrites it. Add a pre-expiry snapshot — read `Context::get_cdn_state()` (or `get_rocketcdn_state()`) *before* calling `disable()`, and persist it (option name TBD; original draft proposed `rocket_cdn_free_pre_expiry_state`, doesn't exist yet — pick something that reads a `cdn_state` value, not a boolean, since that's what's now being snapshotted).
- `inc/Engine/CDN/Render/Controller.php` / `NoticesSubscriber.php` — render the banner explaining why the toggle is disabled, and a tooltip on the disabled toggle input (`views/settings/sections/cdn/rocketcdn-free.php`). Not affected by the `cdn_state` change; still needed as originally scoped — verify these files/paths still match current structure before implementing.
- `inc/Engine/CDN/Subscriber.php::apply_pause_on_rocketcdn_only()` (line 629, hooked on `get_rocket_option_cdn`) still exists unchanged. Check whether it's now redundant with `Context::get_applied_cdn_state()`/`get_cdn_state()`, or still needed as a separate legacy-filter shim — don't assume either way without checking call sites.

**Note:** since `Context::get_rocketcdn_state()` no longer derives state live from `SubscriptionController` for the free/paid/nothing determination (it only checks `SubscriptionController::is_subscription_creation_loading()` for the transient "ongoing activation" state, otherwise trusts the persisted `cdn_state`), this story's expiry/renewal handlers — plus `CdnStateBridge` — are now the only things keeping `cdn_state` correct across these transitions. There's no automatic re-derivation to fall back on if either side misses a case.

**Acceptance criteria:**
- Given WP Rocket's license expires while Free is active, When expiry is detected, Then `cdn_state` reads `nothing` immediately (via existing reconciliation or new code — confirm which), the pre-expiry `cdn_state` value is snapshotted for Task 6.2, the toggle is visually disabled with a tooltip, and nothing is applied to the frontend.

---

## Task 6.2 — Restore Free to prior state on entitlement renewal (2 pts)

**Status: fully unimplemented today** — no renewal listener, hook, or pre-expiry-snapshot option exists anywhere in the codebase (confirmed via full-repo search on both `develop` and the epic branch).

**Where to change:**
- `inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php` — add `restore_free_on_renewal()` that:
  1. Reads the pre-expiry `cdn_state` snapshot written in Task 6.1.
  2. Restores it. **Open design question for the implementer**: `CdnStateBridge::reconcile()` only recomputes `cdn_state` from the legacy `cdn`/`cdn_type` fields when *those* change — it does not defend a directly-written `cdn_state` value against being overwritten later. Decide whether to (a) write `cdn_state` directly and accept it may drift again on the next legacy-field save, or (b) restore the legacy `cdn`/`cdn_type` fields to whatever combination previously produced that state and let the bridge reconcile naturally. Pick (b) unless there's a reason it can't reconstruct the exact prior state.
  3. Deletes the pre-expiry snapshot option.
  4. Hooks into the license renewal action — **not yet identified**; the original text's pointer to `handle_rocketcdn_checkout_parameter()` needs verification, it wasn't confirmed to be the right entry point during this update.
- **Per [comment below](#issuecomment-5278011241)**: on renewal, if the account had free-tier RocketCDN pages before expiry, the expectation is to resume automatic free-page (re-)creation where possible; if that can't complete immediately, land in the existing "creation in progress" state instead of silently re-enabling. `Context::ROCKETCDN_STATE_ONGOING_FREE` (`'ongoing_activation_free'`) and `SubscriptionController::is_subscription_creation_loading()` already exist and look like the intended vehicle for that in-progress state — check whether they already cover this case or need extending for the renewal path specifically, rather than assuming either.
- `CDNOptionsManager.php` — reuse `enable()`/`disable()` for the legacy-field half of the restore rather than writing options directly, per the original guidance.

**Acceptance criteria:**
- Given Free was forced off due to expiry, When the license is renewed, Then `cdn_state` (not a boolean) returns to the exact value it held before the forced-off, or — if the account had free-tier pages that need re-creation and that can't complete synchronously — transitions to the "creation in progress" state instead.

---

## Task 6.3 — Tests for forced-off and recovery (1 pt)

**Where:** `tests/Unit/inc/Engine/CDN/RocketCDN/` — follow the pattern of `DataManagerSubscriber` tests.

**What to cover:**
- License expiry fires → `cdn_state` reads `nothing` (whether via existing `CdnStateBridge` reconciliation or new code), pre-expiry `cdn_state` snapshot is stored.
- License renewal fires → `cdn_state` restored to its pre-expiry value (or the "creation in progress" state, per Task 6.2), pre-expiry snapshot option deleted.
- `Context::get_cdn_state()` / `get_rocketcdn_state()` return `nothing` while forced-off, and the original value after renewal.

---

## Dependencies

- Blocked by **Stories 1, 2, 3** — Story 1 (#8699) has landed on `enhancement/8693-rocketcdn-free-tiers-refactor` (PR #8725), not yet merged to `develop`. Confirm Stories 2 and 3's status before starting.
- Unblocks **Story 8** (Pro mode behavior)

## Sizing

5 pts total. Flagging for the team: Task 6.2's renewal path now has more open design questions than originally scoped (see above) — worth a quick re-estimate discussion before starting, not something this update is deciding on its own.

hellofromahmed:

Closes #8704

## Description

This PR implements Story 6 of the RocketCDN Free Tiers refactor epic (#8693). It closes gaps in license-expiry handling for RocketCDN Free: adds a disabled-toggle tooltip explaining why the control is inactive when the license expires or a reseller is banned, gates REST activation endpoints to reject banned resellers, and hardens the filter-priority ordering that makes license-expiry force-off reliable.

The story turned out to be much smaller than originally scoped. WP Rocket license-expiry force-off and renewal auto-resume were already implemented live and non-destructively via `is_forced_paused()` / `maybe_auto_create_rocketcdn_free_subscription()` (confirmed via grooming + adversarial challenger review, both approved). This PR closes the remaining real gaps: the disabled-toggle tooltip, a banned-reseller activation-gate gap, and a filter-priority fragility introduced by PR #8751 that only worked by accident of subscriber registration order.

## Acceptance Criteria

1. Given WP Rocket's license expires while RocketCDN Free is active, when expiry is detected, then `cdn_state` reads `nothing` immediately, the toggle shows a tooltip explaining why it's disabled, and nothing CDN-related is applied to the frontend.
2. Given RocketCDN Free was forced off due to expiry, when the license is renewed, then `cdn_state` returns to its prior value (or transitions to "creation in progress" if free-tier pages need re-creation and can't complete synchronously).
3. Unit/integration tests cover expiry force-off, renewal auto-resume (all 3 outcomes), and the filter-priority ordering that makes force-off reliable.

## Known pre-existing issues (out of scope)

1. **BYOCDN force-enable on frontend**: `CDN\Subscriber::apply_pause_on_rocketcdn_only()` force-enables CDN on the front end for BYOCDN users with `cdn=0` — pre-existing defect, not touched by this PR (team decision: no separate ticket, just this note).
2. **Banned-reseller handling on paid plans**: `is_forced_paused()` doesn't cover a banned/revoked reseller on an active PAID plan — only Free-tier bans are covered by this story; paid-tier reseller bans are Story 8 territory.
3. **Local sandbox test limitation**: The new `AddPage.php` REST banned-reseller test cannot execute in a local sandbox (`wp_get_environment_type()==='local'` trips a pre-existing environment guard in `add_page()`, confirmed via bisection to affect all AddPage tests identically on unmodified code — not a regression). Needs verification in a non-local environment.
4. **One.Com type hint**: `OneCom::maybe_enable_cdn_option( ?string $cdn )` is typed `?string` but now receives whatever `resolve_live_cdn()` returns (`int`/`bool`). No fatal (no `declare(strict_types=1)`), but the type contract is inaccurate.
5. **OneCom tooltip gap** (found in review): `OneCom::disable_cdn_mode_toggle()` sets `is_forced_off` but never the matching `forced_off_tooltip`, so One.com-hosted users now get a disabled toggle with no explanation while every other forced-off reason has one.

## Type of change

### What was tested

QA (automated pipeline pass) validated this PR against the base branch. Overall: **PARTIAL** — no functional defects found; the one open item is a tooling gap, not a discovered bug.

- **AC1 (expiry force-off)**: PARTIAL. `cdn_state = nothing` and the non-destructive read-filter chain are verified via `licenceExpiryForcesCdnStateNothing.php` (3/3 pass — expired/revoked/reseller-banned, both admin and front-end contexts). The disabled-toggle + tooltip data was confirmed correct server-side via a direct `wp eval` check (`is_forced_off=true`, exact tooltip string matching spec) and 23/23 passing unit tests. A real-browser hover/screenshot confirmation could not be captured this QA pass (tooling limitation, not a defect) — recommended as a follow-up manual/Playwright pass before final sign-off.
- **AC2 (renewal auto-resume)**: PASS. `autoResumeOnLicenceRenewal.php` covers all 3 outcomes (subscription still running / cancelled+pages / cancelled+no pages), all passing.
- **AC3 (test coverage)**: PASS. All new/extended unit and integration tests pass. The one documented exception — `AddPage.php`'s banned-reseller sub-case hitting the pre-existing `wp_get_environment_type() === 'local'` guard in `Rest.php:264` — is confirmed pre-existing (zero diff on `Rest.php` in this PR) and not a regression.

Full QA report: https://github.com/wp-media/wp-rocket/pull/8804#issuecomment-5497146048

### How to test

**Setup:**
- On a site with a **valid** WP Rocket licence, open Settings > WP Rocket > CDN and switch the RocketCDN Free toggle on. Add at least one page.
- Confirm the toggle is enabled and the status indicator reads active.

**Test 1: License expiry force-off**
1. Force the licence to expire by setting `licence_expiration` to a past timestamp in the `wp_rocket_customer_data` transient, then reload the CDN tab.
2. Assert the RocketCDN Free toggle is **disabled**.
3. Assert hovering the toggle shows the "…your WPRocket licence has expired" tooltip.
4. Assert the expired-licence banner renders.
5. Assert `get_rocket_option( 'cdn_state' )` returns `nothing`.
6. Load a front-end page and assert no asset URL is rewritten to `*.rocketcdn.me`.
7. Assert in the database that `wp_rocket_settings['cdn']` is **still `1`** — the force-off must not have written anything.

**Test 2: License renewal auto-resume**
1. Restore a future `licence_expiration` timestamp.
2. Click **Refresh info** on the Dashboard tab, then reload the CDN tab.
3. Assert the toggle is **enabled** again and the tooltip is gone.
4. Assert `cdn_state` returns to `rocketcdn_free`.
5. If the free subscription required re-creation, assert the status indicator shows "creation in progress" (`ongoing_activation_free`) instead of silently staying off.

**Test 3: Banned reseller**
1. Mock a banned reseller license scenario.
2. Assert the toggle is **disabled** and shows the "…your licence is banned" tooltip.
3. Assert attempting to activate via `save_cdn_mode()` REST endpoint returns `403 cdn_mode_forced_off`.
4. Assert attempting to add a page via `add_page()` REST endpoint returns `403 cdn_mode_forced_off`.

- RocketCDN Free mode toggle (disabled state & tooltip rendering)
- License expiry detection and force-off chain
- License renewal and auto-resume chain
- Banned reseller detection and REST activation gating
- Filter priority ordering for `pre_get_rocket_option_cdn`

### Documentation

This PR closes gaps in RocketCDN Free's license-expiry handling. License-expiry force-off and renewal auto-resume are handled live and non-destructively: `Render\Subscriber::maybe_pause_cdn_for_inactive_subscription()` (on `pre_get_rocket_option_cdn`) filters the live `cdn` value to `false` whenever `Render\Controller::is_forced_paused()` is true (free plan + invalid license, paid plan + cancelled), without ever writing to storage; `CdnStateBridge::resolve_live_cdn()`/`resolve_live()` then derive `cdn`/`cdn_state` from that live value on every read. Renewal is symmetric: once the license is valid again, `is_forced_paused()` returns false and the filter stops suppressing `cdn`, so state resolves back to its true stored value automatically — `Render\Controller::maybe_auto_create_rocketcdn_free_subscription()` (on `admin_init`) additionally re-creates the RocketCDN Free subscription if pages existed and need re-provisioning, landing in the "creation in progress" state when that can't complete synchronously.

This PR adds: (1) a `get_forced_off_tooltip()` method surfacing *why* the toggle is disabled, wired into the three CDN driver sections; (2) a banned-reseller term in `should_reject_rocketcdn_activation()` so the REST activation gate matches the live pause condition; (3) a priority pin (10 → 5) on `CdnStateBridge::resolve_live_cdn()` so it deterministically seeds the live `cdn` value before override callbacks, removing a hidden dependency on `inc/Plugin.php` subscriber registration order.

#### What was changed

1. **`Render\Controller::should_reject_rocketcdn_activation()`** — Added the banned-reseller term so the activation gate rejects banned licences, not just expired/loading ones.

2. **`Render\Controller::get_forced_off_tooltip()`** (new method) — Returns the first matching reason (subscription loading → expired → banned) and an empty string when nothing forces the toggle off. The order mirrors `should_reject_rocketcdn_activation()` exactly.

3. **`Render\Controller` section methods** — Added `forced_off_tooltip` key alongside `is_forced_off` in `add_rocketcdn_free_section()`, `add_rocketcdn_paid_section()`, and `add_applied_cdn_state_to_cdn_section()`.

4. **Views** (`rocketcdn-free.php`, `rocketcdn-paid.php`, `your-own-cdn.php`) — Wrapped the label in a `<div class="wpr-cdn-mode-toggle-wrap">` and conditionally rendered the tooltip markup matching the existing CDN add-page tooltip precedent.

5. **`_contentDelivery.scss`** — Added `.wpr-cdn-mode-toggle-wrap` styling with the pointer-events hover pattern (disabled checkboxes don't reliably emit `:hover`, so the wrapper drives it).

6. **`CdnStateBridge::resolve_live_cdn()` hook priority** — Pinned to priority 5 (was 10, relying on array order) so it deterministically seeds the live base value before override callbacks like `maybe_pause_cdn_for_inactive_subscription` (priority 10) run.

### Architecture & design rationale

- **Ban gate placement**: Added to `should_reject_rocketcdn_activation()` only, not `should_display_licence_expired_notice()` (would stack the expired + banned banners) or `is_forced_paused()` (already covered by `is_revoked()`; adding a redundant term implies independence when one subsumes the other).
- **Tooltip order**: Mirrors `should_reject_rocketcdn_activation()` so the two cannot drift.
- **Priority hardening**: Makes the seeder's role explicit and removes the hidden dependency on `inc/Plugin.php` subscriber registration order that PR #8751 introduced.
- **Non-destructive force-off**: Confirmed that the license-expiry path never writes `cdn`/`cdn_type`, only filters reads. Restoration on renewal is automatic when `is_forced_paused()` goes false.

### New dependencies

None.

### Risks

None identified. The changes are minimal and localized:
- Tooltip: purely additive UI, no logic.
- Ban gate: reuses existing `is_reseller_license_banned()` predicate.
- Priority hardening: makes order explicit; current behavior is preserved.
- All existing tests remain passing; new tests lock down the scenarios.

### Code validation

### Code style

### Unticked items justification

All Code validation and Code style items are satisfied — see "What was tested" above and the linked QA report for evidence. No screenshots are attached for the tooltip's visual hover state; a real-browser confirmation pass is tracked as a follow-up (see Known pre-existing issues + QA recommendations).

## Additional Checks


hellofromahmed:
> [!NOTE]

**QA: ⚠️ PARTIAL**

| Acceptance Criterion | Method | Result | Notes |
|---|---|---|---|
| AC1 — Expiry forces `cdn_state=nothing`, disabled toggle + tooltip, nothing applied on frontend | Browser (Playwright-equivalent, manual Chrome MCP) + Analysis (`licenceExpiryForcesCdnStateNothing.php`, 3/3) | ✅ | Live-filtered `cdn`/`cdn_state` confirmed `false`/`nothing` via direct filter invocation; **tooltip hover screenshot captured this pass** (previously deferred), text matches `get_rocketcdn_toggle_forced_off_tooltip()` exactly. Frontend curl showed no `rocketcdn.me` URL (inconclusive as a differential check in this sandbox — no CDN rewrite was observed even with a valid licence — but backed by the passing integration test). |
| AC2 — Renewal restores `cdn_state` to prior value (or "creation in progress") | Browser + Analysis | ⚠️ | "Resumes to prior value" confirmed live (browser + filter test: `cdn=true`, `cdn_state=rocketcdn_free` immediately after renewal). The "creation in progress" branch could not be exercised: its implementing method, `Controller::maybe_auto_create_rocketcdn_free_subscription()`, no longer exists in this diff (see blockers). |
| AC3 — Unit/integration coverage for force-off, renewal (all 3 outcomes), and priority ordering | Analysis (test suite) | ⚠️ | `licenceExpiryForcesCdnStateNothing` (3/3), `resolveLiveCdnOrdering` (2/2), `SaveCdnMode` (8/8), and the 3 Unit test files (15/15) all pass. `AddPage` integration failures (10 failures/7 errors) are pre-existing `wp_get_environment_type()==='local'` guard noise, not a regression (confirmed via zero diff on `Rest.php`'s guard and identical failure signature on unmodified code, matching the PR's own documented limitation). However, the dedicated "renewal auto-resume, all 3 outcomes" test (`autoResumeOnLicenceRenewal.php`) was added then deleted within this branch's own history (see blockers) — that specific AC3 claim has no test today. |

**Blockers:**
- AC2/AC3: `Controller::maybe_auto_create_rocketcdn_free_subscription()` — the method the PR body's own "Documentation" section credits with the "creation in progress" renewal outcome — was deleted upstream by #8831 ("Stop Creating the rocketcdn free subscription automatically"). Its dedicated test, `autoResumeOnLicenceRenewal.php`, was subsequently removed as "stale" in this branch (commit `99561a54d`) rather than replaced. As shipped, nothing re-triggers subscription creation on renewal, so AC2's "creation in progress" sub-case and AC3's "all 3 outcomes" claim are not currently implemented/tested — please update the PR body's technical description (it still describes the deleted method as live) and either restore equivalent coverage or narrow AC2/AC3's wording to match current behavior.
- New finding (this pass): `DataManagerSubscriber::disable_rocketcdn_free_with_rocket_license_expired()` (added this PR, hooked to the new `rocket_license_expired_or_revoked` action, registered only in `is_admin()` context) persistently overwrites `wp_rocket_settings['cdn_state']` to `'nothing'` whenever the licence transitions to expired/revoked — confirmed via source (`CDNOptionsManager::set_cdn_state()` calls `update_option`) and via a live admin-context test. No listener exists on the symmetric `rocket_license_active` action anywhere in the codebase, so nothing restores the stored value on renewal. This contradicts the PR's own repeated "non-destructive, never writes to storage" design rationale (stated in `CdnStateBridge.php`'s docblocks and the PR body), has zero test coverage, and will surface as a real bug once the temporary `CdnStateBridge` live-filter shim is removed in Story 10 (the stored `cdn_state` will then be read directly and will incorrectly show `nothing` for any account that ever experienced a licence expiry, even long after renewal). AC1's own "Test 1, step 7" only asserts `wp_rocket_settings['cdn']` stays `1`, not `cdn_state`, so this slipped past the PR's own manual verification plan.

**Note on scope:** commits `ec0074f34` (Other CDN status message / toggle-freeze) and `6d2c35f2d` (RocketCDN Free add-page-controls) on this branch are unrelated follow-up fixes for other work in the same session, not part of issue #8704 / these 3 ACs, and were excluded from this validation.

### Screenshots

Captured via Chrome MCP (manual Playwright-equivalent walkthrough) directly against the local environment — real HTTP calls to the RocketCDN subscription API were intercepted with a QA-only mu-plugin (not part of the plugin diff, removed after the run) so the licence-expiry/renewal scenario could be exercised deterministically instead of depending on the sandbox's real (rejecting) network access.

Publishing to a public gist was not authorized in this run, so these are local file paths (available to attach on request) rather than hosted URLs:
- Baseline — RocketCDN Free active, toggle ON, no tooltip: `/tmp/qa-8804-screenshots/1-baseline-free-active-toggle-on.jpg`
- **Licence expired — toggle OFF, hover tooltip "RocketCDN is currently paused because your WP Rocket licence has expired."** (closes the prior pass's tooltip-capture gap): `/tmp/qa-8804-screenshots/2-expired-toggle-disabled-tooltip-hover.jpg`
- Renewed — toggle auto-resumes to ON, expired banner gone: `/tmp/qa-8804-screenshots/3-renewed-toggle-auto-resumed.jpg`

### Smoke tests

- Dashboard (`/wp-admin/`): loads fine, WP Rocket toolbar item renders.
- Settings page (`/wp-admin/options-general.php?page=wprocket`): loads fine.
- Noted (pre-existing, not caused by this PR): a stale `localStorage['wpr-hash']` value from a prior session can make `PageManager.change()` throw `Cannot read properties of null (reading 'style')` on load, blanking the whole tab content area. Reproduced identically with the base branch's own `wpr-admin.js` (zero diff in that code path), so this is an existing environment/JS quirk, not a regression from this PR.

tests/Integration/inc/Engine/CDN/Render/Controller/autoResumeOnLicenceRenewal.php
```diff
@@ -0,0 +1,368 @@
...
+	 * subscription kept running throughout. On renewal, the pause simply lifts -
+	 * no write, no re-creation.
+	 */
+	public function testShouldResumeWithNoWritesWhenSubscriptionStillRunning(): void {
+		$this->seed_expired_baseline();
+		$this->set_subscription_status( 'running' );
+
+		$this->set_licence_state( false );
+
+		do_action( 'admin_init' );
```
hellofromahmed: [CRITICAL] `do_action( 'admin_init' )` here fires the **entire global** `admin_init` hook, not just `Render\Subscriber`'s callback. This cascades into every other admin_init-hooked subscriber in the plugin, including BerlinDB's `Table::maybe_upgrade()` for `rocket_cache`, `performance_monitoring`, `preconnect_external_domains`, `preload_fonts`, `rucss_used_css`, `above_the_fold`, and `lazy_render_content` tables.

CI logs confirm this is the root cause: `WordPress database error: Table 'wptests_wpr_rocket_cache' already exists for query CREATE TEMPORARY TABLE ...` traces directly back to this `do_action('admin_init')` call. The DDL corruption breaks transactional test isolation for the rest of the suite — later, completely unrelated classes (`Test_CleanMinify`, `Test_RegenerateMinifyCssKey`, `Test_RocketAfterSaveOptions`) then fail with `Table 'wptests_wpr_rocket_cdn' doesn't exist`. Confirmed via the GitHub Checks API that the base branch (`feature/8701-rocketcdn-refactor-clear-cache`) passes this exact CI job — this PR is the regression source. Every CI matrix job (WP 5.9/PHP 7.4, WP latest/PHP 8.0-8.5) is currently failing because of this.

**Fix:** Don't fire the global `admin_init` action. Call the exact method `Render\Subscriber` wires to it directly on the `$render_controller` instance already retrieved in `set_up()`, instead of `do_action( 'admin_init' );`:

    $render_controller->maybe_auto_create_rocketcdn_free_subscription();

The method is `public`, so this still exercises the real production method under test, without cascading into every unrelated subscriber's table-migration logic.

tests/Integration/inc/Engine/CDN/Render/Controller/autoResumeOnLicenceRenewal.php
```diff
@@ -0,0 +1,368 @@
...
+				'last_accessed' => current_time( 'mysql' ),
+			]
+		);
+
+		$this->mock_successful_create_free_api();
+
+		$this->set_licence_state( false );
+		$this->force_create_api_client_free_url( 'https://rocketcdn.me/api/website/create-free/' );
+
+		do_action( 'admin_init' );
```
hellofromahmed: [CRITICAL] `do_action( 'admin_init' )` here fires the **entire global** `admin_init` hook, not just `Render\Subscriber`'s callback. This cascades into every other admin_init-hooked subscriber in the plugin, including BerlinDB's `Table::maybe_upgrade()` for `rocket_cache`, `performance_monitoring`, `preconnect_external_domains`, `preload_fonts`, `rucss_used_css`, `above_the_fold`, and `lazy_render_content` tables.

CI logs confirm this is the root cause: `WordPress database error: Table 'wptests_wpr_rocket_cache' already exists for query CREATE TEMPORARY TABLE ...` traces directly back to this `do_action('admin_init')` call. The DDL corruption breaks transactional test isolation for the rest of the suite — later, completely unrelated classes (`Test_CleanMinify`, `Test_RegenerateMinifyCssKey`, `Test_RocketAfterSaveOptions`) then fail with `Table 'wptests_wpr_rocket_cdn' doesn't exist`. Confirmed via the GitHub Checks API that the base branch (`feature/8701-rocketcdn-refactor-clear-cache`) passes this exact CI job — this PR is the regression source. Every CI matrix job (WP 5.9/PHP 7.4, WP latest/PHP 8.0-8.5) is currently failing because of this.

**Fix:** Don't fire the global `admin_init` action. Call the exact method `Render\Subscriber` wires to it directly on the `$render_controller` instance already retrieved in `set_up()`, instead of `do_action( 'admin_init' );`:

    $render_controller->maybe_auto_create_rocketcdn_free_subscription();

The method is `public`, so this still exercises the real production method under test, without cascading into every unrelated subscriber's table-migration logic.

tests/Integration/inc/Engine/CDN/Render/Controller/autoResumeOnLicenceRenewal.php
```diff
@@ -0,0 +1,368 @@
...
+	 */
+	public function testShouldStayNothingWhenCancelledWithNoPages(): void {
+		$this->seed_expired_baseline();
+		$this->set_subscription_status( 'cancelled' );
+
+		$this->mock_successful_create_free_api();
+
+		$this->set_licence_state( false );
+
+		do_action( 'admin_init' );
```
hellofromahmed: [CRITICAL] `do_action( 'admin_init' )` here fires the **entire global** `admin_init` hook, not just `Render\Subscriber`'s callback. This cascades into every other admin_init-hooked subscriber in the plugin, including BerlinDB's `Table::maybe_upgrade()` for `rocket_cache`, `performance_monitoring`, `preconnect_external_domains`, `preload_fonts`, `rucss_used_css`, `above_the_fold`, and `lazy_render_content` tables.

CI logs confirm this is the root cause: `WordPress database error: Table 'wptests_wpr_rocket_cache' already exists for query CREATE TEMPORARY TABLE ...` traces directly back to this `do_action('admin_init')` call. The DDL corruption breaks transactional test isolation for the rest of the suite — later, completely unrelated classes (`Test_CleanMinify`, `Test_RegenerateMinifyCssKey`, `Test_RocketAfterSaveOptions`) then fail with `Table 'wptests_wpr_rocket_cdn' doesn't exist`. Confirmed via the GitHub Checks API that the base branch (`feature/8701-rocketcdn-refactor-clear-cache`) passes this exact CI job — this PR is the regression source. Every CI matrix job (WP 5.9/PHP 7.4, WP latest/PHP 8.0-8.5) is currently failing because of this.

**Fix:** Don't fire the global `admin_init` action. Call the exact method `Render\Subscriber` wires to it directly on the `$render_controller` instance already retrieved in `set_up()`, instead of `do_action( 'admin_init' );`:

    $render_controller->maybe_auto_create_rocketcdn_free_subscription();

The method is `public`, so this still exercises the real production method under test, without cascading into every unrelated subscriber's table-migration logic.

hellofromahmed:
> [!NOTE]

**Review: ✅ PASS**

CRITICAL blocker resolved in `981503467`: the 3 `do_action( 'admin_init' )` calls in `autoResumeOnLicenceRenewal.php` were replaced with direct calls to `$this->render_controller->maybe_auto_create_rocketcdn_free_subscription()`, where `$this->render_controller` is fetched from the container's `cdn_render_controller` singleton. Verified this is the exact same shared instance production wires to `admin_init` via `Render\Subscriber` (constructed with `addArgument('cdn_render_controller')` in `inc/Engine/CDN/ServiceProvider.php`), and the method exists and is unchanged on `Render\Controller::maybe_auto_create_rocketcdn_free_subscription()`. No global hook cascade remains, no other files were touched by the fix, no new issues introduced.

**Nice-to-haves (unchanged, still non-blocking):**
- `inc/ThirdParty/Hostings/OneCom.php` — `disable_cdn_mode_toggle()` still doesn't set a matching `forced_off_tooltip` for the One.com hosting-compat path.
- `AddPage.php` banned-reseller REST test still can't run locally due to the pre-existing `wp_get_environment_type() === 'local'` guard; worth an explicit "verified in CI/staging" follow-up.

tests/Integration/inc/Engine/CDN/Render/Controller/autoResumeOnLicenceRenewal.php
hellofromahmed: Fixed in 981503467 — replaced the do_action('admin_init') calls with a direct call to the production method on the container's cdn_render_controller singleton, avoiding the global hook cascade.

tests/Integration/inc/Engine/CDN/Render/Controller/autoResumeOnLicenceRenewal.php
hellofromahmed: Fixed in 981503467 — replaced the do_action('admin_init') calls with a direct call to the production method on the container's cdn_render_controller singleton, avoiding the global hook cascade.

tests/Integration/inc/Engine/CDN/Render/Controller/autoResumeOnLicenceRenewal.php
hellofromahmed: Fixed in 981503467 — replaced the do_action('admin_init') calls with a direct call to the production method on the container's cdn_render_controller singleton, avoiding the global hook cascade.

hellofromahmed:
This PR is depending on Story3, so if Story3 is merged into feature branch plz make sure to rebase this one onto the feature branch also.

hellofromahmed:
Heads up: #8830 (enhancement/8829-toggle-reuse-existing-component) forked from the same base commit as this PR (`535a53652`, before this PR's changes existed) and independently added a forced-off tooltip for the RocketCDN toggles - built on top of a componentization it did in that same branch that extracted the once-triplicated `<label class="wpr-cdn-mode-toggle">` markup out of `rocketcdn-free.php`/`rocketcdn-paid.php`/`your-own-cdn.php` into a shared partial (`views/settings/partials/cdn/cdn-mode-toggle.php`), rendered via a single `toggle_tooltip` data key instead of duplicating the tooltip markup per view.

This PR's view changes (`wpr-cdn-mode-toggle-wrap` + inline `wpr-tooltip` markup duplicated across the three templates again, plus a `forced_off_tooltip` key) will directly conflict with that refactor when both land on `enhancement/8693-rocketcdn-free-tiers-refactor`. To avoid that turning into a lossy merge (either undoing #8830's dedup, or shipping two overlapping/incomplete tooltip implementations - this PR's has no hosting-compatibility copy, #8830's had no licence/subscription copy until now), we folded this PR's subscription-loading / expired-licence / banned-reseller reasons directly into #8830's existing tooltip method instead (same `should_reject_rocketcdn_activation()` banned-reseller term, same precedence order, own unit tests) - see #8830 for the diff.

Suggest when merging this PR into the base branch: drop this PR's `Controller.php` `get_forced_off_tooltip()`/`forced_off_tooltip` addition and the three view-template/`_contentDelivery.scss` changes, since #8830 already covers the same reasons through its componentized partial. The `CdnStateBridge` priority pin and `resolve_live_cdn()` hardening in this PR are unrelated and should merge cleanly either way.

inc/Engine/CDN/Render/Controller.php
```diff
@@ -165,19 +168,20 @@ public function add_rocketcdn_paid_section( array $sections ): array {
...
 			],
-			'status_indicator'  => $status_indicator_data,
-			'applied_cdn_state' => $this->context->get_applied_cdn_state(),
-			'rocketcdn_state'   => $rocketcdn_state,
-			'is_forced_off'     => $this->should_reject_rocketcdn_activation(),
-			'is_active'         => Context::ROCKETCDN_PAID_TYPE === $rocketcdn_state,
+			'status_indicator'   => $status_indicator_data,
+			'applied_cdn_state'  => $this->context->get_applied_cdn_state(),
+			'rocketcdn_state'    => $rocketcdn_state,
+			'is_forced_off'      => $this->should_reject_rocketcdn_activation(),
```
jeawhanlee: I'm not sure why we're not using this [method](https://github.com/wp-media/wp-rocket/blob/34ae5057acd6bcf9ae9ced63db9eeb2717cee68b/inc/Engine/CDN/Render/Controller.php#L941) to check if forced off state should be applied since we still maintain [this](https://github.com/wp-media/wp-rocket/blob/34ae5057acd6bcf9ae9ced63db9eeb2717cee68b/inc/Engine/CDN/Render/Subscriber.php#L52) part of the code that forces cdn off which cdn_state mirrors down the line.

inc/Engine/CDN/CdnStateBridge.php
```diff
@@ -13,6 +13,23 @@
...
+	 * {@see \WP_Rocket\Engine\CDN\Render\Controller::maybe_pause_cdn_for_inactive_subscription()}
+	 * and {@see \WP_Rocket\ThirdParty\Hostings\OneCom::maybe_enable_cdn_option()}, both
+	 * registered at the default priority 10. Pinning this to an earlier priority makes
+	 * that ordering explicit instead of relying on `inc/Plugin.php`'s subscriber
+	 * registration array order (`cdn_state_bridge` listed before `cdn_render_subscriber`),
+	 * which today is the only reason the licence-expiry force-off survives.
+	 *
+	 * @var int
+	 */
+	private const CDN_SEEDER_PRIORITY = 5;
```
jeawhanlee, line 31: Is this constant necessary? I only see it being used in just 1 place.

inc/Engine/CDN/Render/Controller.php
hellofromahmed: Good catch, and you're right that just adding is_forced_paused() as a redundant check elsewhere wouldn't be enough - the toggle's is_forced_off needs to actually include it. Fixed in ce68b809d by OR-ing is_forced_paused() into should_reject_rocketcdn_activation() as a fourth term (alongside subscription-loading / expired-licence / banned-reseller), with a matching fourth precedence step added to get_forced_off_tooltip() so the tooltip and the disabled state can't disagree.

Kept should_reject_rocketcdn_activation() rather than switching callers over to is_forced_paused() directly, since the two aren't interchangeable:
- is_forced_paused() has no "subscription creation in progress" concept, which exists specifically to stop a second activation request from racing the first.
- is_forced_paused() only catches a banned reseller incidentally on the free tier (via `is_free() && is_license_invalid()`, since a ban implies `is_revoked()`) - it misses a ban on an active *paid* plan entirely, since a ban doesn't necessarily flip RocketCDN's own billing-subscription status. should_reject_rocketcdn_activation()'s explicit `is_reseller_license_banned()` term is what actually covers that, and it's called out in this PR's own description as a gap `is_forced_paused()` alone doesn't close.

Added a dedicated regression test (`testShouldForceOffWhenPaidSubscriptionCancelled` in addRocketcdnPaidSection.php) for the exact scenario you flagged - a paid subscription cancelled outside its grace period now correctly disables the toggle with the new "subscription is no longer active" tooltip.

inc/Engine/CDN/CdnStateBridge.php
hellofromahmed: Fair point that it's used in only one array entry - kept it as a named constant rather than an inline 5 mainly because of the docblock cross-reference: resolve_live_cdn()'s docblock explains at length why priority 5 specifically matters (it has to run before the default-priority-10 callbacks that depend on it as a seeder), and referencing self::CDN_SEEDER_PRIORITY from both that explanation and the registration array keeps "this number is the seeder priority, not an arbitrary choice" visible at the call site instead of requiring a trip to the docblock to learn what 5 means.

That said, it's a single use site and a reasonable person could go either way - happy to inline it with a short comment instead if you'd prefer. Let me know and I'll make the change either way.

Mai-Saad:
@hellofromahmed Thanks for the PR. Please find exploratory test notes below [WIP]
- Refresh WPR settings after expire license (before delete cdn subscription) will clear cache of pages added to free list (regression on PR, working on feature branch )
Steps:
1- fresh install and activate 
2- enable free CDN 
3- Add home and other page
4- disable preload
5- expire the WPR license at website
6- refresh info , then refresh cdn => UI forced paused, cache cleared
7- cache home, added page and other page
8- refresh cdn tab => cache of added pages will be cleared 

Expected:
1- refresh wpr after expire license before enter grace period shouldn't clear cache :white_check_mark: 
2- refresh WPR while being in grace period of free not clear cache :white_check_mark: 
3- refresh wpr while grace period of paid not clear cache :white_check_mark: 
4- refresh wpr while no subscription , not clear cache :white_check_mark: 

- text of free while in grace period needs update i.e orange circle and pause text :white_check_mark: 
Steps:
1- fresh install and activate
2- add home to free
3- expire wpr
4- refresh info/admin
5- check cdn tab
6- delete cdn free subscription without delete website
7- repeat 4,5
image

- When renew license during grace period while URL was in list, cdn is enabled in UI while is shouldn't
Steps:
1- fresh install and activate
2- add home to free 
3- expire wpr
4- refresh info/admin => clear cache of added pages when refresh info , then clear all cache when refresh admin
5- cache pages, then delete free subscription then refresh info/admin => nothing cleared , UI become greencircle as point above => fixed now on epic branch + #8862, nothing cleared and UI remains orange circle ✅ 
6- cache pages if cleared
7- renew license at website
8- refresh info /admin => all cache cleared (should clear only added page in this case) ❌ => still the same now on epic branch + #8862, all cache is cleared
9- check cdn UI => cdn is on (it should be off as we stopped auto creation) :white_check_mark: 
     but with green circle and active text ❌ (tested on epic branch + #8862), updated screenshot below
Screenshot 2026-09-16 at 11 59 54

- cdn transients when add home while external requests blocked arenot deleted when delete WPR ✅ fixed on #8862
image (47)

Mai-Saad:
to be tested here :
When disable YOCDN , text / icon should be updated (better be fixed here) 
Steps:
1- fresh install/activate WPR
2- enable YOCDN , add cname then save
3- disable YOCDN => check UI has green circle/active text
✅ Confirmed as fixed on #8862

hellofromahmed:
First point is DONE
Second point is not valid:
image

Mai-Saad:
To be tested here https://group-onecom.slack.com/archives/C08F4LZB2QG/p1789131430904399 ✅

hellofromahmed:
## Fixed: CDN stayed enabled (`cdn=1`) after RocketCDN Free is disabled on licence expiry

**Root cause:** `Settings::sanitize_callback()` is wired as the `sanitize_option_wp_rocket_settings` filter via `register_setting()`, so it runs on **every** `wp_rocket_settings` save — including internal writes like `CDNOptionsManager::disable()`. It unconditionally did:

```php
$input['cdn'] = (int) $this->options->get( 'cdn', 0 );
```

`$this->options` is an `Options_Data` snapshot built once, early in the request, so this silently reverted `cdn` back to whatever it was at request start — discarding `disable()`'s `cdn = 0` write. Every other similar guard in the same method (`purge_cron_interval`, `purge_cron_unit`, `automatic_cleanup_frequency`) correctly checks `isset($input[...])` first; only the `cdn` line was missing it. Fixed to match that pattern.

Diagnosed by bisecting via `remove_all_filters()` on `pre_update_option`/`sanitize_option` (both generic and option-specific) to isolate which hook was responsible, since it doesn't show up in a search for `add_filter` calls (registered implicitly by `register_setting()`).

**Also fixed along the way:**
- `CDNOptionsManager::disable()` now resets `cdn_state` alongside `cdn`.
- `disable_rocketcdn_free_with_rocket_license_expired()` now calls `disable()` instead of `set_cdn_state()` only, which never touched `cdn`.
- `rocket_pre_main_option()`'s `WP_ROCKET_SLUG` transient merge now only carries forward the license fields it's meant for (`consumer_key`, `consumer_email`, `secret_key`, `license`), instead of letting any other stale key it happened to be holding overwrite the current save.

**Tests:** added regression coverage in `tests/Unit/inc/Engine/Admin/Settings/Settings/sanitizeCallback.php` (verified it fails against the reverted buggy line with the exact real-world symptom) and `tests/Unit/inc/admin/rocketPreMainOption.php`. Full unit suite: 2860 tests passing.

hellofromahmed:
@Mai-Saad all points are resolved

hanna-meda:
@hellofromahmed, The second point is about the green circle and text from `wpr_cdn_indicator`, that should turn to orange circle and the corresponding text (correct copy text to be approved/provided in #8855) => confirmed fixed on https://github.com/wp-media/wp-rocket/pull/8862 ✅ 

Free while WPR license expired:
Screenshot 2026-09-15 at 09 55 33

Pro while cancelled during the grace period:
Screenshot 2026-09-15 at 10 12 05

hellofromahmed:
Going to merge this one based on this discussion: https://group-onecom.slack.com/archives/C08EFCYRQ2X/p1789458528889279

Issue #8817, opened 2026-09-04
Refactor 3 Followup: Full cache is cleared after creating free subscription


Mai-Saad:
We need to handle the following issue:

- Adding 1st page will clear cache of added page then once creation is done will clear all cache => it should delete cache of added page only when creation is done and no clear all cache here

Steps:
1- fresh install and activate
2- enable free cdn
3- disable preload
4- visit some pages
5- add home page to free list => check cache , home is cleared
6- cache home
7- after creation is done => check cache , all cleared

Expected:
- 1st add while no free subscription shouldn't clearing cache 
- once creation is done, cache of added page only is cleared

DahmaniAdame:
> Adding 1st page will clear cache of added page then once creation is done will clear all cache => it should delete cache of added page only when creation is done and no clear all cache here

Not mandatory for release. 

> No auto create free after Paid is cancelled => now we auto create free after cancel Paid (may be handled in other story, please ensure which story handle that if not here)

@Mai-Saad is it just the creation or creation and activation? 
If on creation = not a blocker. 
If creation and activate = we shouldn't activate it automatically. 

> While cdn free has URLs in list , uncheck cdn then upgrade => pro will be unchecked in UI while it should be checked {happening on followup of s2}

Should be fixed before release. 

To be investigated:

> when add page after fresh install can see many subscription requests

@Mai-Saad how may? Anything breaks? 

@hellofromahmed maybe needs a delay between two consecutive requests? 

> Set/unset cdn using MCP not clearing cache nor reflecting in UI

Not mandatory for release.

hellofromahmed:
> While cdn free has URLs in list , uncheck cdn then upgrade => pro will be unchecked in UI while it should be checked {happening on followup of s2}

Again this is not related to Story 3 at all.

Mai-Saad:
**General Test Notes (other GHs)**
- No auto create free after Paid is cancelled => now we auto create free after cancel Paid (may be handled in other story,  please ensure which story handle that if not here) => will be handled in https://github.com/wp-media/wp-rocket/pull/8831
- While cdn free has URLs in list , uncheck cdn then upgrade => pro will be unchecked in UI while it should be checked {happening on followup of s2} => New GH

To be investigated:
- when add page after fresh install can see many subscription requests => to be tested on https://github.com/wp-media/wp-rocket/pull/8831#issuecomment-5580516012 , if still  there create new GH
[...]

Issue #8812, opened 2026-09-03
RocketCDN subscription creation can send many requests in certain scenario


Mai-Saad:
**Describe the bug**
If we block external request to rocketcdn , then try to create free subscription, subscription request will fail and many requests will be sent till request success
**To Reproduce**
1. Fresh install and activate of WPR (no cdn subscription yet)
2. Block external requests in wp-config (exclude wp-rocket.me)
3. Add home page to free CDN 
4. Check requests sent using API debugger

**Expected behavior**
- When subscription status request fail , we don't send any further requests 
- When cdn_status transient expired, manually try to add page and create subscription should send new request if we arenot restricted by retry timeout

**Screenshots**

<img width="1241" height="861" alt="Image" src="https://github.com/user-attachments/assets/50bbe946-a860-490c-bd0b-1790353f1c4b" />

**Additional context**
- Issue is happening on 3.23.3.3 and 3.22.0.3
- Thanks to @hellofromahmed to support reproduce the issue
- For retest: check both free and paid creation

Mai-Saad:
Enhanced at refactor #8817 , now we don't keep sending express checkout but subscription is sent when refresh admin (still we better not send subscription request as well if failure happened till transient is expired)

<img width="1327" height="418" alt="Image" src="https://github.com/user-attachments/assets/8dd9ff0c-e513-4fcb-a89d-16812aadbf02" />

<img width="1669" height="423" alt="Image" src="https://github.com/user-attachments/assets/fc27e009-28ee-4417-90ed-22c406a05825" />

PR #8831, opened 2026-09-08
Refs #8817: Stop Creating the rocketcdn free subscription automatically

hellofromahmed:
# Description

Refs #8817
*Fixes the "when add page after fresh install can see many subscription requests" item investigated in #8817. Removes a code path that could create a second, duplicate RocketCDN free subscription for the same site, and stops an activation-retry check from running on every wp-admin page load instead of just the WP Rocket settings page. This does not address the other bullets tracked in #8817 (cache-clearing duplication on first page add, pro/free checkbox state, cname/page cache clearing while unchecked), which remain open.*

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Investigated via the site's built-in `api_log` post type (Tools > API Logs), which records every outbound RocketCDN API call with a full debug trace. On a fresh install, adding the first page to RocketCDN Free produced 13 API calls in ~2 minutes, including two full `subscription/create`-style flows for the same site: subscription `72953`/website `75747` (created 19:05:45) and a second, distinct subscription `72954`/website `75748` (created 19:06:58). Tracing each call's debug trace showed three independent, uncoordinated code paths all checking/creating RocketCDN subscription state without sharing a request-scoped cache:

1. `Context::get_driver()` → `rocketcdn_resolver()` → `SubscriptionController::has_active_subscription()` — fires on every DI-container resolution of the CDN driver (registered non-shared).
2. `DataManagerSubscriber::maybe_retry_activation()` — hooked on `admin_init`, so it ran on every wp-admin page load, and explicitly deleted the `rocketcdn_status` transient mid-request before re-fetching, invalidating the cache for the other two paths.
3. `Render\Controller::maybe_auto_create_rocketcdn_free_subscription()` — also hooked on `admin_init` via `Render\Subscriber`, and was the only path besides the explicit "add page" REST endpoint able to call `create_subscription()` again. Its only guard against creating a duplicate was `has_active_subscription()`, which path #2 could invalidate moments earlier — this is the most likely source of the second subscription observed in the logs.

This was static code review plus log analysis, not yet a live re-run against the fixed code — see "How to test" below for the manual QA I'd like done before merge.

### How to test

1. Check out this branch on a fresh WP install with no prior RocketCDN subscription/token.
2. Go to Settings > WP Rocket > CDN tab > RocketCDN Free, and add the homepage (or first page) to trigger subscription creation.
3. While the subscription-creation task is pending (or shortly after), navigate to a few unrelated wp-admin pages (Dashboard, Plugins, Posts) and back.
4. Open Tools > API Logs (`edit.php?post_type=api_log`) and confirm:
   - No `subscription/*/status`, `website/*`, or `activate` calls are logged while browsing pages other than the WP Rocket settings page.
   - Only one subscription ID / website ID pair appears for the site — no second subscription gets created.
5. Confirm RocketCDN Free still activates normally and the page list/status indicator update correctly on the settings page (where `maybe_retry_activation` should still run).
6. Regression-check the forced-pause flow (see Risks below): force a paused state (e.g. temporarily invalid license), let it resolve, force it a second time, and confirm the whole cache still clears on the second pause too via `maybe_pause_cdn_for_inactive_subscription()`.

RocketCDN Free tier subscription creation/activation, the CDN settings page (RocketCDN Free/Paid sections), and the CDN forced-pause tracking behavior in `Render\Controller` (shares the `rocket_rocketcdn_forced_pause_state` option touched by the removed method).

### Documentation

Two changes:

1. Removed `Render\Controller::maybe_auto_create_rocketcdn_free_subscription()` entirely, along with its `admin_init` hook registration and thin wrapper in `Render\Subscriber`. This method auto-called `create_subscription()` whenever `has_active_subscription()` read as false, with no locking against the other two subscription-check paths described above.
2. Added a `get_current_screen()->id !== 'settings_page_wprocket'` guard to `DataManagerSubscriber::maybe_retry_activation()`, matching the same guard pattern already used in `RocketCDN\NoticesSubscriber` and `Render\Controller::maybe_sync_forced_pause_tracking_state()`. This stops the retry/re-activate flow (and its forced transient invalidation) from running on every admin page load, scoping it to the one screen where it's actually relevant.

### New dependencies

None.

### Risks

The removed method was also the only place resetting the `persistent` key of the `rocket_rocketcdn_forced_pause_state` option back to `false` when a forced-pause condition resolved. That flag is read by `maybe_pause_cdn_for_inactive_subscription()` to avoid re-clearing the whole cache on every request while paused. Without the reset, after one forced-pause/resume cycle, `persistent` stays `true`, so a *second* forced-pause event later would skip its `clear_all_cache()` call. This needs a decision: either restore the reset somewhere appropriate (e.g. alongside the existing `tracking` key reset in `maybe_sync_forced_pause_tracking_state()`), or confirm the team is fine leaving it as-is. Flagging explicitly rather than silently deciding, since it's a behavior change outside the scope of the two requested fixes.

# Mandatory Checklist

## Code validation

- [ ] I validated all the Acceptance Criteria. If possible, provide screenshots or videos.

## Code style

- [ ] Output messages (errors, notices, logs) are explicit enough for users to understand the issue and are actionnable.

Code validation items are unticked because this PR has only been validated via static code review, `php -l`, and analysis of API request logs from a prior reproduction — live manual QA against this exact diff (see "How to test") is still pending before merge. No new automated tests were added since this is a targeted removal + one guard clause on existing, already-covered call sites; the "Output messages" item is not applicable as no new errors/notices/logs were introduced.

# Additional Checks


Mai-Saad:
Note for test:
- We need to validate this point isnot happening "when add page after fresh install can see many subscription requests"  ==> improved than before but we will add other enhancement to see if it can be more improved https://github.com/wp-media/wp-rocket/issues/8856
=> current status at https://github.com/wp-media/wp-rocket/pull/8831/commits/6a04235b9b7ba49bf014df3c01e823b1c17bf552
image

- No free auto creation when cancel Paid subscription while pages exist before upgrade and when renew expired license (check testrail to confirm no other cases) {related thread https://group-onecom.slack.com/archives/C08F4LZB2QG/p1788363978039429 } => downgrade isnot creating free while pages exist :white_check_mark:  , renew will be checked at s6 https://github.com/wp-media/wp-rocket/issues/8704


- No cache clear with admin refresh while no subscription or in grace period of 
=> no subscription: no cache clear :white_check_mark: , but after fresh install and activate , can see those requests
image => requests improved , no cache clear while no subscription/grace to be checked on 
- No cache clear while no subscription with admin refresh :white_check_mark: 
- No Clear while no subscription and license is expired with every admin refresh => :white_check_mark: only once with 1st admin refresh after refresh info while license expired 
- No cache clear while grace of free with every admin refresh :white_check_mark: 
- No cache clear while grace of paid with every admin refresh :white_check_mark: 


- retest https://github.com/wp-media/wp-rocket/issues/8812
requests like that in this case , but home page kept in list, shouldn't it be removed? => expected now as 1st request is success {to be confirmed with Adame if ok or not}
image
=> enhanced than before and home not added , still we can enhance requests https://github.com/wp-media/wp-rocket/issues/8812#issuecomment-5652083774

jeawhanlee (approved):

Mai-Saad:
Requests:
After install and activate , before enable cdn (6)
image

Issue #8845, opened 2026-09-10
Accept read and write commands from Fleet


Miraeld:
Part of wp-media/fleet#6. Target repository: **wp-rocket**.

> **What is Fleet?** Fleet is a new WP Media application where an agency manages
> WP Rocket across every site on its licence from one dashboard — today they open
> one wp-admin per site. To read or change a setting it has to reach the site, and
> it deliberately holds **no customer credential**: it signs a short-lived request
> instead. This issue is part of that, in this repository.
>
> Epic: wp-media/fleet#6

The route Fleet calls, and the verification that guards it.

Depends on the `GetOptions::schema()` extraction.

## Why this exists at all

Fleet is a dashboard where an agency manages WP Rocket across every site on its
licence. To read or change a setting it has to reach the site — and it
deliberately **holds no customer credential**: no WordPress login, no
application password, not even the licence's `consumer_key`. An application
password would be the easy route and is the wrong one: it authenticates a
*user*, so it grants full admin, and Fleet would have to store one secret per
site.

Instead Fleet signs an Ed25519 assertion per request, and this route verifies
it.

## Two proofs, from two parties

Every request must carry both:

| Credential | Header | Signed by | Proves |
|---|---|---|---|
| Command | `Authorization: Bearer` | Fleet | who is asking, and covers the exact body bytes |
| Consent grant | `X-WP-Rocket-Fleet-Consent` | wp-rocket.me | the licence owner allows this capability, right now |

Neither substitutes for the other. Fleet is the party being granted the access,
so a consent claim Fleet signed itself proves nothing — and a consent flag this
plugin cached yesterday proves nothing either, which is why consent is not read
from the settings channel.

## Why it calls `execute()` directly

`permission_callback` on both abilities is
`current_user_can( 'rocket_manage_options' )`. That gate is for a human with a
WordPress session, which Fleet does not and will not have.

This route's authorisation is the verified assertion instead: licence scoped,
site scoped, single use, and body bound. That is a stronger claim than a
logged-in administrator makes.

It also means **no WordPress version floor**. `wp_register_ability()` is 6.9
core and only `register()` uses it; the `execute()` methods touch nothing from
the Abilities API, and `AbilitiesServiceProvider` is registered
unconditionally. An agency fleet full of older sites therefore works.

## Acceptance criteria

- [ ] `GET` returns every allowlisted option with its current value, plus the schema, plus which are writable
[...]

PR #8846, opened 2026-09-10
Expose GetOptions' per-option schema

### LINKED ISSUE #8844: Expose GetOptions' per-option schema
Part of wp-media/fleet#6. Target repository: **wp-rocket**.

> **What is Fleet?** Fleet is a new WP Media application where an agency manages
> WP Rocket across every site on its licence from one dashboard — today they open
> one wp-admin per site. To read or change a setting it has to reach the site, and
> it deliberately holds **no customer credential**: it signs a short-lived request
> instead. This issue is part of that, in this repository.
>
> Epic: wp-media/fleet#6

Expose the per-option schema `GetOptions` already builds, so something other
than the ability registration can read it.

## Why

`GetOptions::register()` builds a `$schema_definitions` array describing every
allowlisted option — its type, its enum where it has one, and a human
description — and then uses it once, as the ability's `output_schema`. It is a
local variable, so nothing else can see it.

A remote caller reading options over HTTP has only the *values* to go on, and
values do not carry their type: an unset toggle and an empty text field both
arrive falsy. Without the schema, that caller has to keep its own copy of
"which of these 61 options is a boolean" — a copy that goes stale on our next
release.

## What changes

`$schema_definitions` moves into a public `schema(): array` method that returns
it intersected with the allowlist, exactly as `register()` does today.
`register()` calls it.

**No behaviour change.** Same array, same intersection, same `output_schema`.

## Acceptance criteria

- [ ] `GetOptions::schema()` returns the definitions, intersected with `AllowedOptions`
- [ ] `register()` produces byte-identical `output_schema` to before
- [ ] An option outside the allowlist has no schema entry
- [ ] Unit test covering both of the above

## Note

Deliberately separate from the Fleet route that needs it, because it is a pure
refactor of code this team owns and is worth reviewing on its own terms.

Miraeld:
# Description

Fixes #8844

> ### ✅ QA already performed by the author
>
> **Automated:** full unit suite green — **2808 tests**, 9151 assertions. Full PHPCS green. Codacy green.
> **Functional:** the ability still registers with the same output schema, checked in a WordPress 6.9 install.
>
> This is a pure refactor with no behaviour change, so there is no functional surface to QA beyond that. The Fleet feature it enables is in #8847, which carries its own evidence.
>
> **What still needs a human:** a second pair of eyes on the extraction being genuinely equivalent. Nothing environmental.

Pulls the per-option schema out of `GetOptions::register()` into a `schema()` method, so it has one definition and two readers. No behaviour change: `register()` still builds the same `$properties` from the same array, still behind the same `function_exists( 'wp_register_ability' )` guard.

**Why it is wanted.** A remote caller that is shown a list of options and their values cannot render a control for each without knowing the type. An unset boolean and an unset text field both arrive empty, and only the schema says which is which. #8845 serves this over the Fleet route; the ability keeps advertising it as its output schema exactly as before.

Users see nothing from this PR on its own.

## Type of change

- [x] Sub-task of wp-media/fleet#6
- [ ] Chore
- [ ] Release

### What is Fleet, for context

Fleet is a new WP Media dashboard where an agency manages WP Rocket across every site on its licence from one place, instead of opening one wp-admin per site. It is reached from wp-rocket.me. For it to show or change a setting it has to ask the site, and it deliberately holds **no credential** for any site — it signs a short-lived token per request which the plugin verifies. This PR is the first of two on the plugin side; it is a pure refactor and adds nothing that talks to anything.

### What was tested

Automated. The existing unit and integration suites for the abilities cover both readers of the extracted method:

- `tests/Unit/inc/Engine/Abilities/Options/GetOptions/ExecuteTest.php`
- `tests/Integration/inc/Engine/Abilities/Options/GetOptions/RegisterGetOptionsAbilityTest.php`

Full unit suite green (2808 tests), full PHPCS green.

### How to test

Nothing to set up. The ability is unchanged, so the check is that it still registers with the same output schema:

```bash
composer test-unit -- --filter Abilities
wp rocket abilities list       # get-options still present, same schema
```

Or in a WordPress 6.9 install, confirm `wp-rocket/get-options` still advertises a `properties` entry per allowlisted option.

The MCP / Abilities surface only — `wp-rocket/get-options`. Nothing else reads the schema today. Caching, optimisation and the admin UI are untouched.

### Documentation

`register()` used to declare a ~380 line `$schema_definitions` array inline, intersect it with the allowlist, and pass the result to `wp_register_ability()`. The array and the intersection moved to:

```php
public function schema(): array {
    $schema_definitions = [ /* … unchanged … */ ];

    return array_intersect_key( $schema_definitions, array_flip( $this->allowed_options->get() ) );
}
```

and `register()` now reads `$properties = $this->schema();`.

Two properties preserved deliberately:

- **The `function_exists` guard stays in `register()`.** The array is only built when something asks for it, so this costs no work on a site below WordPress 6.9.
- **The intersection with the allowlist stays inside `schema()`.** An option that may not be read has no schema entry either, so a caller cannot be handed a type for something it will never be shown. Putting the intersection at the call site would have made that a caller's responsibility to remember.

### New dependencies

None.

### Risks

Low. One method extracted, no call site outside this class, no change to what is registered. The risk of getting it wrong would be a schema that no longer matches the allowlist, which the existing integration test for the ability's registration covers.

# Mandatory Checklist

## Code validation

- [x] I implemented tests to cover the new/changed code.

## Code style

- [x] Output messages (errors, notices, logs) are explicit enough for users to understand the issue and are actionable.

Tests: covered by the existing suites for both readers rather than by new ones. The extracted method has no behaviour of its own to test that `register()` did not already have tested — a new test asserting `schema()` returns the array would restate the implementation.

# Additional Checks

- [ ] When possible, I prepared ways to observe the implemented system (logs, metrics, etc.).

Observability: nothing to observe. No I/O, no failure mode.

### DIFF
--- a/inc/Engine/Abilities/Options/GetOptions.php
+++ b/inc/Engine/Abilities/Options/GetOptions.php
@@ -36,15 +36,23 @@ class GetOptions implements AbilitiesInterface {
 	}
 
 	/**
-	 * Registers the ability to get WP Rocket options.
+	 * The type of every option Fleet or an MCP client may be shown.
 	 *
-	 * @return void
+	 * Extracted from register() so it has one definition and two readers.
+	 * The ability advertises it as its output schema; the Fleet route serves
+	 * it so a remote caller can render a real control per option instead of
+	 * guessing a type from whatever value happens to be stored — an unset
+	 * boolean and an unset text field both arrive empty, and only this says
+	 * which is which.
+	 *
+	 * Intersected with the allowlist, so an option that may not be read has
+	 * no schema entry either and cannot be rendered as though it did.
+	 *
+	 * @since 3.23.4
+	 *
+	 * @return array Keyed by option name.
 	 */
-	public function register(): void {
-		if ( ! function_exists( 'wp_register_ability' ) ) {
-			return;
-		}
-
+	public function schema(): array {
 		$schema_definitions = [
 			// Cache settings.
 			'cache_logged_user'                         => [
@@ -413,7 +421,20 @@ class GetOptions implements AbilitiesInterface {
 			],
 		];
 
-		$properties = array_intersect_key( $schema_definitions, array_flip( $this->allowed_options->get() ) );
+		return array_intersect_key( $schema_definitions, array_flip( $this->allowed_options->get() ) );
+	}
+
+	/**
+	 * Registers the ability to get WP Rocket options.
+	 *
+	 * @return void
+	 */
+	public function register(): void {
+		if ( ! function_exists( 'wp_register_ability' ) ) {
+			return;
+		}
+
+		$properties = $this->schema();
 
 		wp_register_ability(
 			'wp-rocket/get-options',


inc/Engine/Abilities/Options/GetOptions.php
```diff
@@ -36,15 +36,23 @@ public function __construct( Options_Data $options, AllowedOptions $allowed_opti
...
+	 * guessing a type from whatever value happens to be stored — an unset
+	 * boolean and an unset text field both arrive empty, and only this says
+	 * which is which.
+	 *
+	 * Intersected with the allowlist, so an option that may not be read has
+	 * no schema entry either and cannot be rendered as though it did.
+	 *
+	 * @since 3.23.4
+	 *
+	 * @return array Keyed by option name.
```
remyperona: This needs to be shortened

remyperona (changes requested):

inc/Engine/Abilities/Options/GetOptions.php
Miraeld: Done

remyperona (approved):

PR #8847, opened 2026-09-10
Accept read and write commands from Fleet, on a verified assertion

### LINKED ISSUE #8845: Accept read and write commands from Fleet

Miraeld:
# Description

Fixes #8845

> ### ✅ QA already performed by the author
>
> This was built and driven as a working prototype before being split into PRs, so the security behaviour is **observed, not reasoned about**. Full detail under *Detailed scenario*; the summary:
>
> **Automated:** 29 new unit tests. Full unit suite green — **2808 tests**, 9151 assertions. PHPCS green, Codacy green. Verification itself has **62 tests** in [wp-media/fleet-bridge](https://github.com/wp-media/fleet-bridge) on PHP 7.4 and 8.3.
>
> **End to end** against a real WordPress 6.9 site, a real Fleet and a real wp-rocket.me — every row below was executed, not predicted:
>
> | Attempt | Result |
> |---|---|
> | Valid command + valid grant | 200, 61 settings + 61 schema entries |
> | Valid write | 200, previous value reported |
> | No consent grant | 401 |
> | Read grant used on a PATCH | 401 |
> | Grant minted for a different site | 401 |
> | Grant replayed | 200, then 401 |
> | Command replayed | 401 |
> | Body altered after signing | 401 |
> | Access switched off mid-session, no cache flush | 401 on the next request |
> | `consumer_key` (not allowlisted) | 422 |
> | Access `none` / `read` / `write` | 403+403 / 200+403 / 200+200 |
> | `vendor/wp-media/fleet-bridge` removed | route 404s, **plugin still boots** |
>
> That last row is a bug this QA caught: it used to take the whole plugin down, and a browser request hid it because the page cache answered 200. It only showed up under wp-cli.
>
> **What still needs a human QA pass:** nothing environmental for the negative paths (they need no Fleet — see *How to test*). The positive path needs a Fleet environment; ping me and I will drive it live for you.

Depends on #8844 — merge that first. This PR targets that branch, so its diff here is only its own work.

Adds one REST route, `wp-rocket/v1/fleet/settings`, that lets Fleet read every allowlisted WP Rocket option on a site and change them, but **only** on a request carrying two valid signatures from two different parties.

**What a user sees.** Nothing, unless their agency uses Fleet *and* they have switched Fleet access on for their licence on wp-rocket.me. With access off, the route answers 401 to everything. On a site whose licence has never polled, likewise. The plugin's own behaviour — caching, optimisation, admin UI — is untouched.

## Type of change

- [x] Sub-task of wp-media/fleet#6
- [ ] Chore
- [ ] Release

### What is Fleet, for context

Fleet is a new WP Media dashboard where an agency manages WP Rocket across every site on its licence from one place, instead of opening one wp-admin per site. It is reached from wp-rocket.me, and it is what this route exists for.

The thing worth knowing before reviewing: **Fleet holds no credential for any customer site.** Not a WordPress login, not an application password, not the licence key. An application password would have been the easy route and the wrong one — it authenticates a *user*, so it grants whatever that user can do rather than "this plugin's allowlisted options", and it would mean one stored secret per site, tens of thousands of them, whose leak compromises every site at once. So nothing is stored and every request is signed instead.

### What was tested

**Automated**, 29 new unit tests over the two classes that are the plugin's own:

- `tests/Unit/inc/Engine/Fleet/Route/IsFleetTest.php` — which checks the route demands, with which scope, over which bytes, and that a refusal says nothing useful.
- `tests/Unit/inc/Engine/Fleet/Route/WriteTest.php` — that a body which is not what it claims is refused before any option is written.
- `tests/Unit/inc/Engine/Fleet/TrustStore/IssuerTest.php` — every way of not knowing who to trust, all answering "trust nobody".

Full unit suite green (2808 tests), full PHPCS green.

**Manual**, end to end against a real WordPress 6.9 site driven by a real Fleet and a real wp-rocket.me. Every one of these was observed, not reasoned about:

| Attempt | Answer |
|---|---|
| No consent grant | 401 |
| A read grant on a PATCH | 401 |
| A grant minted for a different site | 401 |
| The same grant used twice | 200, then 401 |
| The same command used twice | 200, then 401 |
| Body altered after signing | 401 |
| Fleet access switched off mid-session, no cache flush | 401 immediately |
| `consumer_key` (not on the allowlist) | 422 |
| Access level `none` / `read` / `write` | 403+403 / 200+403 / 200+200 |

The verification itself is tested in [wp-media/fleet-bridge](https://github.com/wp-media/fleet-bridge) — 62 tests on PHP 7.4 and 8.3 — rather than here.

### How to test

The route refuses everything without a Fleet, so the honest test is the negative one, which needs nothing:

```bash
curl -i https://your-site.test/wp-json/wp-rocket/v1/fleet/settings
# 401, {"code":"rocket_fleet_unauthorised","message":"This request is not authorised."}

curl -i -X PATCH https://your-site.test/wp-json/wp-rocket/v1/fleet/settings \
  -H 'Authorization: Bearer anything' \
  -H 'X-WP-Rocket-Fleet-Consent: anything' \
  -d '{"option_name":"cache_logged_user","option_value":1}'
# 401. Nothing is written.
```

Then confirm the plugin is otherwise unaffected: clear the cache, save the settings page, check a page is cached.

For the positive path you need a Fleet environment and a licence with access switched on; ping me and I will drive it, or follow `docs/site-control.md` in wp-media/fleet.

- **New:** one REST namespace route, and a daily cron event (`rocket_fleet_purge_nonces`).
- **Touched:** `inc/Plugin.php` — the Fleet service provider is registered alongside the abilities providers. `inc/Engine/WPRocketUninstall.php` — the new event and the new options are cleaned up.
- **Reads, changes nothing:** `RemoteSettingsClient` — the same licence poll the plugin already makes, read for two extra fields.
- **Writes through:** the existing `GetOptions` / `SetOption` abilities, pulled from the container rather than rebuilt, so the allowlist and sanitisation have one home.

Not affected: caching, preload, optimisation, the admin UI, the MCP surface.

### Documentation

**Two proofs, from two parties, on every request.**

| Credential | Header | Signed by | Proves |
|---|---|---|---|
| command | `Authorization: Bearer` | Fleet | who is asking, and covers the exact body bytes |
| consent grant | `X-WP-Rocket-Fleet-Consent` | wp-rocket.me | the licence owner allows this scope on this site, right now |

Neither substitutes for the other, and the two reasons are the whole design:

- **Fleet must not sign the consent.** Fleet is the party being granted the access. Its own signature over "the owner allowed me" proves nothing.
- **The consent must not be cached.** It used to ride on the licence settings poll, which is cached for a day — so a customer revoking access left their sites accepting commands for up to 24 hours. A permission that takes a day to withdraw is not a permission. The grant now lives 60 seconds and arrives *with* each command, so revocation takes effect on the next request.

What the licence poll now carries is only *where the keys are*, which is safe to cache for a day because a key rotation publishes two keys at once for exactly that reason.

```
Fleet ──1─▶ wp-rocket.me     "does this licence still allow settings:write on this host?"
      ◀─2── grant, 60s, signed by wp-rocket.me
      ──3─▶ this site        Authorization: command (Fleet)
                             X-WP-Rocket-Fleet-Consent: grant (wp-rocket.me)
```

**Where the code is.**

| File | What it is |
|---|---|
| `Route.php` | The route, the permission callback, and what a believed request may do |
| `TrustStore.php` | Who this install trusts — the only genuinely WP Rocket-specific part |
| `ServiceProvider.php` | Wiring |
| `Subscriber.php` | Route registration, and the housekeeping cron |

Verification lives in **`wp-media/fleet-bridge`**, not here. Assertion verification written twice is assertion verification wrong once, and WP Rocket is the first plugin to install it rather than the last.

**Decisions a reviewer is likely to question:**

- **The route registers unconditionally and refuses on its own.** Not registering it when access is off would make a site with Fleet switched off answer 404 where a site that has simply never polled answers 401 — and that difference tells an outsider which licences have opted in.
- **The scope is derived from the request method,** never read from the request, so a caller cannot nominate the permission it is checked against. Anything that is not a `GET` is treated as a write, rather than matched against a list a future method could be added outside of.
- **The command is verified over the raw body,** not re-encoded parameters. The digest covers the bytes that travelled.
- **The grant is checked after the command,** so a valid grant is not spent on an unsigned request.
- **Every refusal is the same 401 with the same message.** The route is reachable by anyone, so distinguishing "unknown issuer" from "bad signature" would tell a stranger whether the issuer they guessed is the one this site trusts. The reason goes to the log.
- **`args` is deliberately not declared on the PATCH.** The REST validator would reject a bad option name with its own message before `SetOption` saw it, which would make "not on the allowlist" indistinguishable from "malformed request".
- **The abilities come out of the container.** Building a second `SetOption` would be a second place for the allowlist and the sanitisation to live.
- **`read()` takes no argument.** WordPress calls a route callback with the request and PHP passes extra arguments harmlessly, so declaring one we never read would only be a parameter every static analyser correctly reports as unused.

**`TrustStore` normalises two response shapes.** `get_remote_settings_data()` returns the whole decoded response on a cache miss, with settings under `->data`, and the inner `data` on a cache hit. Reading one shape only works for half the calls, and which half depends on transient expiry — so it looks intermittent rather than wrong. Worth knowing: **`rocket_insights_display_post_column` has this same bug today.** Not fixed here, because it is not this PR's, but it is real and worth its own issue.

### New dependencies

**`wp-media/fleet-bridge ^1.1`** — the verifier. Public on Packagist, MIT, no runtime dependencies beyond PHP 7.4 and libsodium (which has shipped with PHP since 7.2). 62 tests, CI on 7.4 and 8.3.

It resolves from Packagist like every other `wp-media/*` dependency here — verified from a clean container with no credentials, which is the shape the release build runs in.

Not mozart-prefixed, matching `apply-filters-typed`, `mcp-oauth` and `wp-mixpanel`; the mozart list is the legacy set and mozart is explicitly removed from `deploy_plugin.yml`.

### Risks

**Security** is the point of the PR rather than a footnote, so the mitigations are above. In summary: only `EdDSA` accepted and compared by equality; every published key tried, and an empty key set refuses; identifiers spent atomically via `INSERT IGNORE` on `wp_options`' unique index rather than `get_transient`/`set_transient`, which is not atomic and would let two concurrent requests both be honoured; key sets fetched over HTTPS only, excepted by host name for local development rather than by a flag one wrong environment variable away from doing it in production; a `ver` claim refused when unknown, because everything else here freezes on release and old installs cannot be patched.

**Performance.** Nothing runs on a normal page load. The permission callback does one signature verification and, on a cold cache, one HTTPS fetch of a key set (cached 10 minutes, failures 30 seconds). The licence poll is the one the plugin already makes.

**A build without the dependency would have taken the whole plugin down.** `fleet_subscriber` is resolved on every request and building it constructs the verifier, so a missing `vendor/wp-media/fleet-bridge` threw from the service provider on every request. Verified, and worth stating how: a front-end request answered 200 because the page cache served it, which is exactly the sort of thing that makes this look survivable. Under wp-cli:

```
Uncaught Error: Class "WPMedia\FleetBridge\Bridge" not found
thrown in inc/Engine/Fleet/ServiceProvider.php on line 67
```

Registration is now guarded on the class existing. A site without it answers 404 on the route and caches pages exactly as before.

**Table growth.** Two `wp_options` rows per honoured request — the command and the consent grant each carry their own one-time identifier and each is recorded. It grows with real use rather than with traffic: a refused token is never recorded, and nothing is written on a normal page load. The rows are `autoload = 'no'`, so they never reach a front-end request. A daily event prunes anything expired. Nothing about it is load bearing: an expired command is refused on its own expiry claim whether or not its row is still there. Uninstall could not remove them either, since the rows are named for a hash and `WPRocketUninstall` only deleted options by exact name; it now sweeps option prefixes the way it already sweeps transient prefixes.

**Tenancy is enforced in Fleet, and can only be.** This plugin can check that a command is Fleet's, that it names *this* site, and that the owner allows it — never *which* Fleet customer asked, because there is no per-site secret shared with Fleet to check that against. That boundary is deliberate and documented rather than an omission.

# Mandatory Checklist

## Code validation

- [x] I implemented tests to cover the new/changed code.

## Code style

- [x] Output messages (errors, notices, logs) are explicit enough for users to understand the issue and are actionable.

On output messages: the 401 a caller receives is deliberately **not** actionable and deliberately identical for every cause, for the oracle reason above. Everything actionable goes to `error_log` with a `[fleet]` prefix, where an operator can read it and a caller cannot. This is the one place the template's intent and the security requirement pull in opposite directions, and the security requirement wins.

# Additional Checks

- [x] When possible, I prepared ways to observe the implemented system (logs, metrics, etc.).


MathieuLamiot:
@Miraeld This PR targets 8844 branch which itself is targeted to merge to develop.
This PR should not land on develop for now. Please use a feature branch for fleet-related stuff. 8844 branch may be OK for develop, but not fleet related stuff

remyperona (changes requested):
Inline comments and docblocks need to be shortened

Miraeld:
Done

inc/Engine/Fleet/Route.php
```diff
@@ -0,0 +1,440 @@
...
+			if ( array_key_exists( $option, $stored ) ) {
+				$settings[ $option ] = $stored[ $option ];
+
+				continue;
+			}
+
+			// An option nobody has saved is absent from the stored array, so
+			// report it with the value the plugin itself acts on: an unwritten
+			// `minify_css` behaves as off.
+			$settings[ $option ] = get_rocket_option( $option, $this->blank_for( $schema[ $option ] ?? [] ) );
```
remyperona, line 209: Calling get_rocket_option for each allowed option seems inefficient. Getting the full settings option array once and then iterating over it would be better.

inc/Engine/Fleet/Route.php
```diff
@@ -0,0 +1,440 @@
...
+	 *
+	 * @param string $message What happened.
+	 *
+	 * @return void
+	 */
+	public static function log( string $message ): void {
+		if ( ! rocket_get_constant( 'WP_DEBUG' ) ) {
+			return;
+		}
+
```
remyperona: it should also check for WP_DEBUG_LOG

inc/Engine/Fleet/ServiceProvider.php
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Wires the Fleet route.
+ *
+ * Verification comes from `wp-media/fleet-bridge` rather than living here, per
+ * wp-media/fleet#55. What stays here is where trust comes from and what a
+ * command may do once believed.
+ *
+ * The abilities are pulled from the container rather than constructed, so the
+ * route calls the same objects the MCP surface does and the allowlist and
+ * sanitisation keep one home.
```
remyperona: We don't need all that

inc/Engine/Fleet/ServiceProvider.php
```diff
@@ -0,0 +1,88 @@
...
+					$this->getContainer()->get( 'fleet_trust_store' ),
+					new HttpKeySets( [ Route::class, 'log' ] ),
+					new WpdbNonceStore(),
+					// Port free: `home_url()` may carry one, wp-rocket.me stores
+					// the domain without it, and both sides derive the subject
+					// from this string.
+					new Config( (string) wp_parse_url( home_url(), PHP_URL_HOST ) ),
+					new SystemClock()
+				);
+			}
```
remyperona: Why is this built like this?

remyperona (changes requested):

inc/Engine/Fleet/Route.php
Miraeld: `read()` already reads the settings array once. `get_options->execute()` does `$this->options->get_options()` and intersects the allowlist, so `get_rocket_option` only runs for keys absent from that array, meaning the ones never written on this site. It is there for the `pre_get_rocket_option_$name` and `rocket_get_option_$name` filters: without it we report a default the plugin does not actually act on. Happy to drop it and fill from the schema default instead if you prefer, since the `unset` list already tells Fleet which ones were never written.

inc/Engine/Fleet/Route.php
Miraeld: Done in 79cb329, it now checks both.

inc/Engine/Fleet/ServiceProvider.php
Miraeld: Cut to one line in 79cb329.

inc/Engine/Fleet/ServiceProvider.php
Miraeld: No good reason, and it was the only closure in 199 registrations. Rebuilt with `addArguments` like the rest in 79cb329.

PR #8871, opened 2026-09-17
Enhancement/8828 rocketcdn mixpanel tracking

### LINKED ISSUE #8828: RocketCDN Refactor - Tracking
## The two state axes

Two properties, never merged into one state string. Merged, "Pro subscription, toggle off" and "no CDN at all" both report as nothing, and `ongoing_activation_free` is invisible.

- [ ] `cdn_mode`: `none`, `rocketcdn_free`, `rocketcdn_paid`, `byocdn`. Which CDN the site is set up with, regardless of whether it is switched on.
- [ ] `cdn_status`: `active`, `inactive`, `forced_off`, `ongoing_activation`. Whether that mode is applied to the front end.

## To track

- [x] `RocketCDN Mode Changed`, on `rocket_cdn_mode_changed`. Properties: `cdn_mode`, `cdn_status`, and `trigger`, one of `user_toggle`, `auto_activate_first_page`, `fresh_install_resolution`, `migration`, `license_expiry`, `license_renewal`, `pro_purchase`, `pro_cancellation`.
- [x] `RocketCDN Free Activation Failed`, on creation API error. Property: `reason`.
- [x] `RocketCDN Pro Activated`, on `rocket_cdn_subscription_created`. Properties: `source` (`dashboard_upgrade`, `banner_cta`), `preserved_free_pages_count`.
- [x] `RocketCDN Forced Off`, on expiry, ban or cancellation detected. Properties: `reason` (`license_expired`, `license_banned`, `pro_cancelled_outside_grace`), `pre_expiry_mode`.
- [x] `RocketCDN Cache Cleared`, on the CDN cache-clear path. Properties: `scope` (`free_pages`, `site_wide`), `trigger`, `pages_count`.
- [ ] `RocketCDN Pending Cancellation Banner Viewed`. No CTA on this banner, so the view is the only signal. Properties: `stage` (`grace_period`, `pending_deletion`), `days_remaining`. Pair with `RocketCDN Mode Changed` to see whether users switch to Free or Other CDN while the cancellation resolves in the background.
- [x] `Button Clicked` / `button: 'rocket cdn retry pro detection'`. Property: `attempts_used`.
- [x] `RocketCDN Upsell Banner Viewed`. Add `cdn_mode` and `cdn_status`. An upsell shown to an active Free user is a different event from one shown to a user with no CDN or to a BYOCDN user, and nothing else in the payload says which.
- [x] `RocketCDN Upsell Banner Expanded` and `Collapsed`. Keep `trigger` (`manual`, `auto_limit_reached`, `auto_limit_released`), add `cdn_mode` and `cdn_status`.
- [x] `RocketCDN Upsell CTA Clicked`. Keep `destination` (`express-checkout`, `iframe`) and `pages_count`, add `cdn_mode` and `cdn_status`. This is the join to `RocketCDN Pro Activated` with `source: banner_cta`.
- [x] `RocketCDN Activation Failed Banner Viewed` and `RocketCDN Activation Failed CTA Clicked`. Both bypass the `track_event()` helper, so they ship without `interaction_channel` and `path`. Route them through the helper.

## Deprecated

- [x] `Button Clicked` / `button: 'rocket cdn pause'`, triggers `user_paused` and `user_resume`. Replaced by `RocketCDN Mode Changed` with `trigger: user_toggle`.
- [x] `Button Clicked` / `button: 'rocket cdn pause'`, triggers `wpr_forced_pause` and `wpr_forced_resume`. Replaced by `RocketCDN Forced Off`, and by `RocketCDN Mode Changed` with `trigger: license_renewal` for the resume leg.
- [x] `RocketCDN Mode` (`cdn_mode` read from the active tab). Replaced by `RocketCDN Mode Changed`. Retire the old name, do not repoint it. It holds tab labels, so it must be gone before the new `cdn_mode` property ships or the two value sets mix.
- [x] `cdn` in the tracked-settings allowlist. Replaced by `RocketCDN Mode Changed`, since `cdn` becomes a derived boolean.
- [x] Instrumentation on the legacy CDN driver endpoint. Replaced by `rocket_cdn_mode_changed` on the new mode endpoint.
- [x] `cdn_type`. No replacement needed, the field is deleted.
- [x] CNAME reachability and `rocket_cname_valid_*`. Replaced by `RocketCDN No CNAME Warning`.

jeawhanlee:
# Description

Fixes #8828

Replaces RocketCDN's old, merged Mixpanel tracking with the new two-axis model (`cdn_mode` + `cdn_status`), so a Pro subscription with the toggle off, and no CDN at all, are no longer indistinguishable. Deprecates `RocketCDN Mode`, `Button Clicked` (cdn pause), and the old `RocketCDN Activation Failed Banner Viewed` payload shape, and adds `RocketCDN Mode Changed`, `RocketCDN Free Activation Failed`, `RocketCDN Pro Activated`, `RocketCDN Forced Off`, and `RocketCDN Cache Cleared`.

## Type of change

- [ ] Chore
- [ ] Release

### What was tested

Automated: new/updated PHPUnit unit tests covering `Context::get_cdn_status()`/`is_forced_off()`/`get_forced_off_reason()`, `Render\Controller::maybe_sync_forced_off_tracking_state()` and `localize_tracking_data()`, `CDN\Subscriber::remove_cdn_from_tracked_options()`, `SubscriptionController`, `CDNOptionsManager::enable()`, and RocketCDN REST `saveCdnMode()`. Manual: confirmed the four `composer test-integration-adminonly` regressions this branch introduced (an undefined `cdn_state` index surfaced by unrelated admin-screen tests) are fixed.

### How to test

With Mixpanel tracking opted in, exercise each trigger and confirm the event/properties in the Mixpanel debug console (or a network inspector on the tracking request):
- Toggle RocketCDN Free/Pro/BYOCDN/off from the settings page → `RocketCDN Mode Changed` with `trigger: user_toggle`.
- Let the free tier auto-activate on first page cached → `trigger: auto_activate_first_page`.
- Upgrade to Pro via the dashboard CTA and via the upsell banner CTA → `RocketCDN Pro Activated` with `source: dashboard_upgrade` / `banner_cta`, and a `RocketCDN Mode Changed` with `trigger: pro_purchase`.
- Let a licence expire, then renew it → `RocketCDN Forced Off` (`reason: license_expired`) then `RocketCDN Mode Changed` (`trigger: license_renewal`).
- Cancel a Pro subscription outside the grace period → `RocketCDN Mode Changed` with `trigger: pro_cancellation`.
- Fail free-tier activation (bad API response) → `RocketCDN Free Activation Failed`.
- Trigger a CDN cache clear (mode switch, and the free-pages-only path) → `RocketCDN Cache Cleared` with the right `scope`/`pages_count`.

RocketCDN settings page (Free/Pro/BYOCDN toggle, upsell banner, checkout iframe/redirect flow), the `current_screen`-driven forced-off/resume detection, CDN cache-clear tracking, and Mixpanel/tracking subscribers generally (deprecated event removal).

### Documentation

- `Context` is now the single source of truth for both tracking axes: `get_cdn_state()`/`get_applied_cdn_state()` give `cdn_mode`; the new `get_cdn_status()` derives `cdn_status` (`active`/`inactive`/`forced_off`/`ongoing_activation`). `is_forced_off()` and `get_forced_off_reason()` moved here from `Render\Controller` so both the settings-page UI and the tracking code read the same forced-off state.
- `Render\Controller::maybe_sync_forced_off_tracking_state()` (on `current_screen`) is the one place that detects a forced-off/resume *transition* and fires `RocketCDN Forced Off` or `RocketCDN Mode Changed`; it replaces the old generic `rocket_rocketcdn_cdn_state_changed` action and its `Tracking::track_rocket_cdn_pause_status()` handler, both removed.
- `Rest::apply_cdn_mode()` and `DataManagerSubscriber` fire `RocketCDN Mode Changed`/`RocketCDN Pro Activated` directly at each mode-change/activation call site, with a `source` (`dashboard_upgrade` vs `banner_cta`) threaded from the JS click handler through a `rocketcdn_source` URL param into the checkout redirect/iframe flow (`src/js/global/rocketcdn.js`), since which CTA was clicked can't be recovered after redirect.
- `CDN\Subscriber::remove_cdn_from_tracked_options()` strips the deprecated `cdn` key from `rocket_mixpanel_tracked_options` at runtime, rather than editing `dynamic-lists.json` directly — that file is overwritten by `DynamicLists`' remote sync, so a static edit wouldn't survive.
- `rocket_cdn_retry_pro_detection_attempts` is a new standalone option tracking cumulative manual Pro-detection retries; added to `WPRocketUninstall`'s cleanup list.

### New dependencies

None.

### Risks

Low — tracking-only changes, no effect on CDN serving/caching behaviour. Main risk is a missed/duplicated event at one of the several call sites now firing `RocketCDN Mode Changed`; mitigated by per-call-site unit test coverage and the manual trigger-by-trigger test plan above.

# Mandatory Checklist

## Code validation

## Code style

N/A

# Additional Checks


jeawhanlee:
Item 6 [here](https://github.com/wp-media/wp-rocket/issues/8828#issue-5373790944) is not implemented because the banner does not exist yet.

Mai-Saad (commented):
Automated code review (Claude Code). 10 correctness findings, most clustered around the new RocketCDN Mixpanel tracking instrumentation — stale-state reads on toggle, untracked/mistracked failure paths, and a script-localization ordering bug. See inline comments for details.

inc/Engine/CDN/Render/Subscriber.php
```diff
@@ -60,6 +60,7 @@ public static function get_subscribed_events(): array {
 				[ 'get_paid_status_indicator_texts', 10, 4 ],
 			],
 			'admin_init'                              => 'maybe_disable_rocketcdn_paid_after_cancellation',
+			'admin_enqueue_scripts'                   => 'localize_tracking_data',
```
Mai-Saad: **Correctness:** `localize_tracking_data()` is hooked on `admin_enqueue_scripts` at the default priority (10), but the `wpr-admin-common` script it localizes onto is only registered at priority 11 (inc/admin/ui/enqueue.php:51), so `wp_localize_script()` silently no-ops.

On every admin page load, this calls `wp_localize_script('wpr-admin-common', 'rocket_cdn_mixpanel_data', ...)` before that handle is registered; `WP_Scripts::add_data()` returns false for an unregistered handle. `rocket_cdn_mixpanel_data` is never output, so `src/js/global/rocketcdn.js`'s fallback (`typeof rocket_cdn_mixpanel_data !== 'undefined' ? ... : {}`) always kicks in, and every JS-side Mixpanel event ships `cdn_mode`/`cdn_status` as `undefined`. The sibling case (`Tracking\Subscriber::localize_optin_status`, same hook/handle) is deliberately hooked at priority 15 to avoid exactly this.

inc/Engine/CDN/RocketCDN/Rest.php
```diff
@@ -634,18 +634,31 @@ public function save_cdn_mode( WP_REST_Request $request ) {
...
+
+		if ( null === $trigger ) {
+			return;
+		}
+
+		$this->track_event(
+			'RocketCDN Mode Changed',
+			[
+				'cdn_mode'   => $this->context->get_cdn_state( $mode ),
+				'cdn_status' => $this->context->get_cdn_status(),
```
Mai-Saad: **Correctness:** `apply_cdn_mode()` passes the freshly-applied `$mode` into `get_cdn_state()` but calls `get_cdn_status()` with no way to override, so `cdn_status` is computed from the stale, per-request `Options_Data` snapshot taken before this method's own write.

Example: user toggles CDN mode from `nothing` to `rocketcdn_free`. The tracked 'RocketCDN Mode Changed' event reports `cdn_mode: 'rocketcdn_free'` (correct) but `cdn_status: 'inactive'` (wrong, derived pre-write) — an internally contradictory payload on every single toggle.

inc/Engine/CDN/Context.php
```diff
@@ -180,6 +197,81 @@ public function get_rocketcdn_state( ?string $cdn_state = null ): string {
...
+	 *
+	 * @since 3.22
+	 *
+	 * @return bool True if the CDN should be force-paused, false otherwise.
+	 */
+	public function is_forced_off(): bool {
+		$this->forced_off_reason = null;
+
+		// Force paused if paid plan cancelled but in grace period.
+		if ( $this->subscription_controller->is_paid() && $this->subscription_controller->is_in_grace_period() ) {
```
Mai-Saad, line 237: **Correctness:** This `is_forced_off()` branch (paid subscription cancelled but still within grace period) returns `true` without ever setting `$this->forced_off_reason`, leaving it `null`.

A paid subscriber cancels but is still inside the grace period: `is_paid()=true`, `is_in_grace_period()=true`. `is_forced_off()` returns true, but `forced_off_reason` stays null (reset to null at the top of the method, never assigned in this branch). `Controller::maybe_sync_forced_off_tracking_state()` then fires 'RocketCDN Forced Off' with `reason => null` instead of a real value, for what is likely the single most common forced-off trigger.

inc/Engine/CDN/Render/Controller.php
```diff
@@ -942,6 +968,36 @@ public function maybe_disable_rocketcdn_paid_after_cancellation(): void {
...
+	public function localize_tracking_data(): void {
+		wp_localize_script( 'wpr-admin-common', 'rocket_cdn_mixpanel_data', $this->get_tracking_data() );
+	}
+
+	/**
+	 * Gets the `cdn_mode` and `cdn_status` tracking axis values together.
+	 *
+	 * @return array{cdn_mode: string, cdn_status: string}
+	 */
+	private function get_tracking_data(): array {
```
Mai-Saad: **Correctness:** `get_tracking_data()` reads `cdn_mode`/`cdn_status` via `$this->context->get_cdn_state()`/`get_cdn_status()` (backed by the stale, per-request `Options_Data` snapshot) immediately after `maybe_disable_rocketcdn_paid_after_cancellation()` wrote the new `cdn_state` through the separate, always-fresh `$this->options_api` service.

When a Pro subscription is cancelled outside the grace period, this method sets `$settings['cdn_state'] = 'nothing'` and persists it via `options_api->set()`, then immediately fires 'RocketCDN Mode Changed' via `get_tracking_data()` — which reports the stale prior mode (e.g. `rocketcdn_paid`) instead of `nothing`. The same method explicitly works around this exact staleness hazard 30 lines earlier (`get_applied_cdn_state( $settings['cdn_state'] ?? null )`) but not here.

src/js/global/rocketcdn.js
```diff
@@ -429,74 +463,11 @@
...
-			mixpanel.identify(rocket_mixpanel_data.user_id);
-		}
-
-		mixpanel.track('RocketCDN Activation Failed CTA Clicked', {
-			context: rocket_mixpanel_data.context,
-			plugin: rocket_mixpanel_data.plugin,
-			brand: rocket_mixpanel_data.brand,
-			application: rocket_mixpanel_data.app
-		});
+		trackRocketCDNUpsellMixpanelEvent( 'RocketCDN Activation Failed CTA Clicked' );
```
Mai-Saad, line 472: **Correctness:** `trackRocketCDNActivationCTA()` now delegates to the shared `trackRocketCDNUpsellMixpanelEvent()` helper, which hard-returns before calling `mixpanel.track()` when `rocket_mixpanel_data.user_id` is falsy. The inline code this replaced only used `user_id` for a best-effort `identify()` and always called `mixpanel.track()` regardless.

`user_id` is server-populated as an empty string whenever `consumer_email` isn't set (self-hosted/unlicensed installs, or before license data syncs). Any such opted-in user clicking the activation-failed CTA now silently generates zero Mixpanel events, where previously the anonymous `track()` call still fired.

inc/Engine/CDN/RocketCDN/Rest.php
```diff
@@ -634,18 +634,31 @@ public function save_cdn_mode( WP_REST_Request $request ) {
...
 	 */
-	private function apply_cdn_mode( string $mode ): void {
+	private function apply_cdn_mode( string $mode, ?string $trigger = null ): void {
 		$settings              = $this->options_api->get( 'settings', [] );
 		$settings['cdn']       = (int) ( Context::CDN_STATE_NOTHING !== $mode );
 		$settings['cdn_type']  = Context::BYOCDN_TYPE === $mode ? 'byocdn' : 'rocketcdn';
 		$settings['cdn_state'] = $mode;
 		$this->options_api->set( 'settings', $settings );
+
+		if ( null === $trigger ) {
```
Mai-Saad, line 649: **Correctness:** This `null === $trigger` early-return means any caller that invokes `apply_cdn_mode()` without a `$trigger` skips tracking entirely — which is exactly what `rollback_failed_subscription()` (line ~495) does: `$this->apply_cdn_mode( Context::CDN_STATE_NOTHING );` with no second argument.

User toggles to `rocketcdn_free` with no active subscription: `apply_cdn_mode($mode, 'user_toggle')` fires 'RocketCDN Mode Changed' immediately; `create_subscription(true)` then fails (API error); `rollback_failed_subscription()` reverts `cdn_state` to `nothing` via this method with `trigger=null`, hitting this early return before `track_event()` runs. Mixpanel permanently records a successful activation that was actually rolled back, with no event ever showing the revert.

inc/Engine/CDN/RocketCDN/SubscriptionController.php
```diff
@@ -334,10 +349,18 @@ public function check_status( string $task_id ) {
...
+				$reason = 'RocketCDN: Received not known response code when check subscription\'s status.';
+
 				$this->logger::error(
-					'RocketCDN: Received not known response code when check subscription\'s status.',
+					$reason,
 					$status
 				);
+
+				$this->track_event(
+					'RocketCDN Free Activation Failed',
```
Mai-Saad, line 361: **Correctness:** This PR adds a tracking call here to the `default:` (failure) branch of the `switch` in `check_status()`, but the sibling `case 'SUCCESS':` branch a few lines above (line 339-340, `$this->options_manager->enable( false );`) still fires no 'RocketCDN Mode Changed'/activation event.

The free tier auto-activates via this async action-scheduler status-check job — a genuine CDN-mode-changing event — but unlike every other activation path this PR instrumented (ajax `enable()`, checkout redirect, REST toggle), free-tier activations completed through this success path are invisible to the new `cdn_mode`/`cdn_status` analytics model.

inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php
```diff
@@ -192,10 +204,26 @@ public function handle_rocketcdn_checkout_parameter(): void {
 		$this->cdn_options->save_token( $token );
 		$this->cdn_options->enable();
 
-		// Force cdn_type to rocketcdn.
-		$current_options             = $this->options_api->get( 'settings', [] );
-		$current_options['cdn_type'] = Context::ROCKETCDN_TYPE;
-		$this->options_api->set( 'settings', $current_options );
+		$this->track_event(
```
Mai-Saad, line 212: **Correctness:** This 'RocketCDN Mode Changed' tracking call (and the 'RocketCDN Pro Activated' one just below it) fires unconditionally, even though `$activation_result` from `activate_subscription()` (line 201, `$activation_result = $this->api_client->activate_subscription( $token, $website_id );`) is never checked.

If the RocketCDN activation API call fails after local token/website_id validation passes, `$activation_result` reflects the failure but is ignored, so `cdn_options->enable()` still runs and these new tracking calls report a successful Pro activation to Mixpanel that the backend never confirmed. Contrast with `maybe_retry_activation()` elsewhere in the same file (line ~620-623), which correctly checks `is_wp_error( $activation_result )` and bails before enabling.

inc/Engine/CDN/Render/Controller.php
```diff
@@ -617,17 +620,40 @@ public function maybe_sync_forced_off_tracking_state( \WP_Screen $screen ): void
...
+					'cdn_status' => $this->context->get_cdn_status(),
+					'trigger'    => $trigger,
+				]
+			);
+
+			return;
+		}
+
+		$settings        = $this->options_api->get( 'settings', [] );
+		$pre_expiry_mode = $this->context->get_cdn_state( $settings['cdn_state'] );
```
Mai-Saad, line 649: **Correctness (minor):** `$this->context->get_cdn_state( $settings['cdn_state'] )` accesses the array key directly with no null-coalescing, unlike the equivalent access 90 lines later in the same class (line 939: `$settings['cdn_state'] ?? null`).

When the `settings` option array lacks a `cdn_state` key (e.g. a fresh install before `CdnStateBridge`'s upgrade backfill runs), this line emits a PHP "Undefined array key" warning — exactly the class of regression the PR description says was already found and fixed elsewhere in this branch, but this instance was missed.

src/js/global/rocketcdn.js
```diff
@@ -192,13 +189,23 @@
...
 		if ( !iframeVisit ) {
+			window.rocketcdnButtonUrl = setSourceParam( window.rocketcdnButtonUrl, source );
+
 			// Small delay to ensure Mixpanel event is sent before navigation
 			setTimeout( function() {
 				window.location.href = window.rocketcdnButtonUrl;
 			}, 100 );
 		} else {
+			// Remembered so enableCDN() can report which CTA led to this iframe checkout.
+			window.rocketcdnIframeSource = source;
```
Mai-Saad, line 207: **Correctness (minor):** `window.rocketcdnIframeSource` is set on a CTA click but never cleared afterward, contradicting its own doc comment that it should be "absent when...no click to attribute."

After one real click sets `window.rocketcdnIframeSource` (e.g. `banner_cta`), nothing resets it. A later `enableCDN()` call within the same page load that was NOT preceded by a fresh click (e.g. modal re-opened via `maybeOpenModal()`/`maybeOpenModalFromURL()`, or a retried subscribe from inside the iframe) still sends the stale source from the earlier click, misattributing which CTA led to that activation.

inc/Engine/CDN/RocketCDN/Rest.php
jeawhanlee: It was not mentioned to track failure at this point.

inc/Engine/CDN/RocketCDN/DataManagerSubscriber.php
jeawhanlee: Valid & Resolved

inc/Engine/CDN/RocketCDN/SubscriptionController.php
jeawhanlee: It was not mentioned to track at this point.

inc/Engine/CDN/Render/Controller.php
jeawhanlee: cdn_state is already set at this point, if it isn't, that's an actual bug.

src/js/global/rocketcdn.js
jeawhanlee: Valid

hellofromahmed (approved):

PR #8872, opened 2026-09-17
Closes wp-media/wp-rocket.me#6493: Add strict format validation for Lazy Render Content location hashes

hellofromahmed:

Closes wp-media/wp-rocket.me#6493

## Description

This fix adds strict format validation for values submitted through the Lazy Render Content beacon endpoint, so only well-formed location hashes are stored and used internally. It also adds a defensive escape when those stored values are used to build an internal string-matching pattern, and fixes an unrelated edge case in Above-the-Fold/LCP where a failed pattern match could return an unexpected type.

## Type of change

- [ ] Sub-task
- [ ] Chore
- [ ] Release

### What was tested

QA independently verified the validation and escaping logic against a local WP Rocket install:

- Confirmed a malformed `results.lrc[]` value (not matching the expected 32-character hex hash format) submitted via the beacon endpoint is rejected and never persisted.
- Confirmed a well-formed 32-hex-char hash still results in `data-wpr-lazyrender="1"` being correctly applied to the corresponding element (no regression to normal Lazy Render Content behavior).
- Confirmed a stored value that doesn't match the expected format is safely neutralized by the escaping logic on read, with no manual cleanup required for sites upgrading from a previous version.
- Confirmed the Above-the-Fold/LCP code path always returns a string, even when the internal pattern-replace call doesn't find a match — covered by a new unit test (`tests/Unit/inc/Engine/Media/AboveTheFold/Frontend/Controller/optimize.php`).
- Ran repeated DB read/write and cache purge/regenerate cycles with no fatal errors observed.
- Not completed: a manual admin-side (wp-admin dashboard/settings) smoke test — assessed as low risk since this fix touches no admin-rendered code paths, but flagged as a pre-merge nice-to-have for a human reviewer.

Independent code review also confirmed the affected data has exactly one write path and one read path in the codebase, so the fix is reached by every consumer. Full unit suite: 2785 tests / 9061 assertions, 0 failures. PHPStan and PHPCS clean.

### How to test

1. **Format validation**: Submit a `results.lrc[]` value that is not a well-formed 32-character lowercase hex string via the beacon endpoint, for a URL with Lazy Render Content enabled. Confirm it is not persisted.
2. **Valid hashes still work**: Submit a well-formed 32-hex-char lowercase hash (matching an actual `data-rocket-location-hash` attribute on the page) via the beacon flow. Reload and confirm the corresponding element receives the `data-wpr-lazyrender="1"` attribute, with no regression to legitimate behavior.
3. **Defensive escaping**: If the DB contains a row with a pre-existing value that doesn't match the expected format, confirm the page still renders normally with no `TypeError` or fatal errors.
4. **Above-the-Fold/LCP type safety**: On a page using Above-the-Fold/LCP optimization, if the stored `src` produces a pattern that fails to match, confirm no `TypeError` is raised and the page renders without fatal errors.

- **Lazy Render Content** (`inc/Engine/Optimization/LazyRenderContent/`): input validation and consumption logic for beacon submissions.
- **Above-the-Fold/LCP** (`inc/Engine/Media/AboveTheFold/Frontend/Controller.php`): return-type safety for `preg_replace_callback()`.
- **Beacon endpoint** (`rocket_beacon` action): input handling for performance hints.

### Documentation

**Background**: The `rocket_beacon` endpoint accepts submissions of `results.lrc[]` (an array of "location hash" values meant to be 32-character lowercase hex digests). These values were previously persisted with only `sanitize_text_field()` validation, which does not constrain the value to the expected hash format. The stored values are later used, unescaped, to build an internal string-matching pattern used for content rendering — a value that doesn't match the expected format could reach that pattern-building step in an unintended way. Full technical detail is documented on the private tracking issue.

**Fixes implemented**:

1. **Input validation (LazyRenderContent/AJAX/Controller.php)**: Before persisting any element from `results.lrc[]`, validate it against a strict allowlist regex `^[a-f0-9]{32}$` (lowercase hex only, exactly 32 characters). Skip (do not store) any element that doesn't match.

2. **Defensive escaping (LazyRenderContent/Frontend/Controller.php)**: After retrieving stored hashes from the database, pass each through `preg_quote()` before building the internal matching pattern. This neutralizes any stored value — new or pre-existing — into a literal string that can only match itself. For legitimate 32-hex hashes, `preg_quote()` is a no-op (hex characters have no special meaning in this context).

3. **Type contract guard (Media/AboveTheFold/Frontend/Controller.php)**: The `set_fetchpriority()` method's return type is declared `string`, but the `preg_replace_callback()` result was not null-checked before being returned. If pattern matching fails, `preg_replace_callback()` returns `null`, which flows into the method's return and causes a `TypeError`. Now capture the result, check for `null`, and return the original HTML if replacement failed.

**Migration**: No database migration or cleanup script is required. Any pre-existing stored value that doesn't match the expected format is neutralized by the escaping logic on next render.

### New dependencies

None.

### Risks

- **Potential functional regression (resolved)**: The strict `^[a-f0-9]{32}$` regex rejects any hash that is not lowercase 32-char hex. Traced hash generation to `md5()` calls in `LazyRenderContent/Frontend/Processor/Dom.php` and `Processor/Regex.php` (server-side, deterministic, always lowercase 32-char hex) and confirmed the beacon collector only echoes this value back verbatim — the regex can never reject a legitimate hash.
- **Pre-existing stored values**: Sites may have stored values from before this fix that don't match the expected format. These are automatically neutralized on next render via the escaping logic — no manual cleanup or cache purge required upon upgrade.

# Mandatory Checklist

## Code validation

## Code style

None — all items validated by implementation, lead review, and QA (see "What was tested" above).

# Additional Checks

## Follow-up tickets

Non-blocking hardening items surfaced during this pipeline run were consolidated into one follow-up: [#8874 — Hardening follow-ups: Lazy Render Content beacon validation](https://github.com/wp-media/wp-rocket/issues/8874).


hellofromahmed:
> [!NOTE]

**Review: ✅ PASS**

Verified independently:
- `LazyRenderContent/AJAX/Controller.php::add_data()` — `is_string()` guard + `^[a-f0-9]{32}$` allowlist correctly reject/skip malformed `results.lrc[]` elements before persistence; partial acceptance semantics preserved.
- `LazyRenderContent/Frontend/Controller.php::optimize()` — `preg_quote($hash, '/')` applied with the delimiter matching the actual `preg_replace()` pattern delimiter; confirmed a no-op for legitimate hex hashes.
- `Media/AboveTheFold/Frontend/Controller.php::set_fetchpriority()` — `preg_replace_callback()` result is null-guarded before use, matching the existing pattern used elsewhere in the same file/RegexTrait; return type contract (`string`) is now honored.
- Traced hash generation (`md5()` in `Dom.php`/`Regex.php`/`SimpleHtmlDom.php`) to confirm the allowlist regex cannot reject any legitimate hash — resolves the spec's open question, no regression risk.
- Confirmed no other code path writes to or reads `below_the_fold` outside the two fixed controllers (single write path, single read path) — the defensive escaping is reached by every consumer.
- Ran `vendor/bin/phpcs` on all three changed files — clean.
- Ran the full updated unit test suite for LazyRenderContent + AboveTheFold (72 tests) — all pass, including the new validation-scenario tests which assert on the actual malformed-input case, not just the happy path.

**Nice-to-haves:**
- `inc/Engine/Optimization/LazyRenderContent/Frontend/Controller.php` — `(string) $hash` in the `array_map` callback will emit a PHP "Array to string conversion" warning if a stored row ever contains a non-scalar JSON element; not currently reachable given upstream validation, but worth a defensive `is_scalar()` check in the NTH follow-up.
- Consider prioritizing the two hardening items filed in the follow-up ticket (`$max_lrc_hashes_number` cap enforcement, clearing performance-hint rows on cache purge).

hellofromahmed:
> [!NOTE]

**QA: ✅ PASS**

Validated via Strategy A (direct HTTP reproduction against `wp-admin/admin-ajax.php` and the front-end page), plus Strategy C (source diff analysis) for the defensive-escaping logic. Backend-only fix, no UI surface — Strategy B not applicable.

**Verification approach:** submitted a `results.lrc[]` value that does not match the expected 32-character hex format to the beacon endpoint, and confirmed it is rejected before persistence. Verified the same on the pre-fix code path to confirm the check is meaningful (not a no-op), then confirmed on the fixed code that a non-conforming stored value cannot corrupt page rendering, thanks to the defensive escaping applied on read. Full detail is documented on the private tracking issue.

|---|---|---|
| 1. Malformed `lrc[]` values rejected/not persisted (AJAX Controller) | API | ✅ Non-conforming value correctly rejected; DB row confirmed empty, not the submitted value |
| 2. `preg_quote()` defensive escaping on read (Frontend Controller) | Analysis + API | ✅ Diff confirms `array_map(preg_quote, $hashes)` before pattern build; behaviorally confirmed no corruption even when a stored value doesn't match the expected format |
| 3. `set_fetchpriority()` null-safe return (AboveTheFold Controller) | Analysis | ✅ Diff confirms `preg_replace_callback` result captured, null-checked, falls back to original `$html`, preserving `string` return type contract; covered by new unit tests in `tests/Unit/.../AboveTheFold/Frontend/Controller/optimize.php` |
| 4. Primary regression: non-conforming input must not corrupt/blank page | API | ✅ Confirmed reproducible on pre-fix commit, confirmed **not reproducible** on PR branch — front-end page fully intact (only cosmetic hash-attribute stripping via normal `remove_hashes()` path, identical to the no-data case) |
| 5. Legitimate 32-hex hash still works (no regression) | API | ✅ Submitted real hash matching an on-page `data-rocket-location-hash`; confirmed `data-wpr-lazyrender="1"` applied to that exact element after cache purge |

**Smoke tests:** DB writes/reads via WP-CLI behaved as expected across all runs; no fatal errors observed in any request. Admin-side smoke test (settings page / dashboard load) was not completed due to time constraints in this session — recommend a quick manual check before merge, though this fix touches no admin-rendered code paths.

**Note:** Criterion 3's null-safety guard was validated via code/test analysis rather than triggering a live `preg_replace_callback()` null return over HTTP (would require crafting a specific LCP `src`/pattern edge case); the accompanying new unit test exercises this path directly.

PR #8876, opened 2026-09-18
Harden per-user cache key validation

hellofromahmed:

## Description

Adds validation to the per-user cache key derivation to ensure that the username used to select a cache bucket is confirmed by a companion cookie set during login, rather than relying solely on an unauthenticated client-supplied value.

### What was tested

Backend-only change (no UI/frontend files touched). Validated with a mix of live functional testing against a local environment (cookie manipulation via curl, real login/logout flows, on-disk cache inspection) and review of the new unit tests for branch coverage.

|---|---|---|
| Per-user cache bucket cannot be selected using only a self-chosen cookie value | API | ✅ |
| Cookie with a username segment but missing/invalid companion validation is served the anonymous cached page (not an error, not another user's content) | API | ✅ |
| Genuine, valid logged-in session continues to get the correct per-user cached page (no regression) | API | ✅ |
| Existing on-disk cache naming unaffected; `rocket_clean_user()` and purge logic still work | API | ✅ |
| Shared logged-in cache branch (`common_cache_logged_users` filter) is gated by the same validation | Analysis (unit tests + code trace) | ✅ |

- Sent a request with a manually-set `wordpress_logged_in_*` cookie carrying a username segment but no companion validation cookie at all — the response matched the anonymous cached page byte-for-byte, and no new per-user cache directory was created.
- Repeated with a companion cookie set to an arbitrary, non-matching value — same result: anonymous page served, no error, no new directory.
- Performed a real login through `wp-login.php` and confirmed the companion cookie is issued alongside the native logged-in cookie at login, and cleared alongside it at logout.
- Fetched the homepage with the real session cookies and confirmed a new cache directory following the existing naming convention was created and served correctly.
- Ran `rocket_clean_user()` against that user and confirmed the per-user cache directory was correctly located and removed, with no errors.
- Reviewed the new unit test suite — good coverage of valid/missing/wrong-secret/wrong-username/expired/tampered-expiration cases for both the per-user and shared logged-in-cache branches.

Smoke tests: WP Rocket settings page and wp-admin dashboard both loaded normally after the change.

### How to test

1. The per-user cache bucket can no longer be selected using only an unauthenticated, self-chosen cookie value.
2. A request with a cookie whose username segment is set but whose accompanying validation cookie is missing or invalid is served the anonymous (not another user's) cached page.
3. A request with a genuine, valid logged-in session continues to be served that user's correct per-user cached page (no regression).
4. Existing cache directories are not broken; no manual migration required.

- **Cache module:** Per-user cache bucket selection and validation logic.
- **Login/logout:** User cache cookie lifecycle.
- **Shared logged-in cache:** The same validation applies to the common-cache-logged-users branch.

## Type of change

- [ ] Sub-task
- [ ] Chore
- [ ] Release

### Documentation

**Companion Cookie Validation:**

A new, time-bound cookie (`wp_rocket_ucc_<COOKIEHASH>`) is set whenever a user logs in and cleared on logout. Its value contains an expiration timestamp and an HMAC signature over the username and expiration, keyed by the site's `secret_cache_key`.

**Hardened Bucket Selection:**

The `get_request_cache_path()` method now validates the companion cookie before trusting the username segment of the `wordpress_logged_in_*` cookie:
- If the companion cookie is absent, invalid, or expired, the request falls through to the anonymous cache bucket.
- If it is valid and non-expired, the per-user bucket is selected as before (no change to on-disk path format).

This ensures that only requests whose username has been confirmed by a login event (server-side cookie issuance) can access a per-user cache bucket; any other request is served the anonymous cached page instead.

**Backward Compatibility:**

- On-disk bucket naming remains unchanged; no cache-directory migration required.
- Existing `rocket_clean_user()` and purge logic work without modification.
- Users already logged in before the fix deploy will have cache access restored as soon as their auth cookie is naturally reissued (within ~1 day of activity, or at next login/logout).

### New dependencies

None.

### Risks

None identified. This change hardens an existing validation point without introducing new dependencies or external calls on the hot cache-serving path. Two minor, non-blocking follow-up considerations were raised during review — see [this comment](https://github.com/wp-media/wp-rocket/pull/8876#issuecomment-5729332901) for the team to weigh in on.

# Mandatory Checklist

## Code validation

## Code style

N/A — all items validated by the automated pipeline (implementation, lead review, and QA passes).

# Additional Checks


tests/Unit/inc/Engine/Cache/UserCacheKeySubscriber/clearUserCacheCookie.php
```diff
@@ -0,0 +1,76 @@
...
+		$this->options    = Mockery::mock( Options_Data::class );
+		$this->subscriber = new ClearCookieRecordingUserCacheKeySubscriber( $this->options );
+	}
+
+	public function testShouldClearCookieAtBothPathsWithPastExpiry() {
+		$before = time();
+
+		$this->subscriber->clear_user_cache_cookie();
+
+		$this->assertCount( 2, $this->subscriber->calls );
```
hellofromahmed: [MEDIUM] This assertion hardcodes the assumption that `COOKIEPATH !== SITECOOKIEPATH`, unlike the sibling `setUserCacheCookie.php` test which derives the expected call count dynamically from the actual constant values (see its own comment explaining why). Both test files define these constants process-wide with an `if ( ! defined() )` guard, so whichever of the two loads first in the same PHPUnit process wins for both. If `setUserCacheCookie.php` happens to load first (it falls back to `SITECOOKIEPATH = '/'`, same as `COOKIEPATH`), this test's hardcoded `assertCount( 2, ... )` and the two `assertContains()` calls will fail because only one call is recorded. It currently passes only because alphabetical file discovery loads this file first.

**Fix:** Derive the expected call count/paths dynamically the same way `setUserCacheCookie.php` does (e.g. `COOKIEPATH === SITECOOKIEPATH ? 1 : 2`), or force distinct values for the two constants in a way that cannot be short-circuited by load order (e.g. only define them in one shared bootstrap include).

hellofromahmed:
> [!NOTE]

**Review: ✅ PASS**

Previous MEDIUM blocker (hardcoded `assertCount(2, ...)` order-dependency in `clearUserCacheCookie.php`) resolved in `4eff6a7ed`: the expected call count/paths are now derived dynamically from `COOKIEPATH`/`SITECOOKIEPATH`, matching the pattern already used in the sibling `setUserCacheCookie.php`. No other changes introduced by this commit.

hellofromahmed:
> [!NOTE]

**QA: ✅ PASS**

Backend-only change (no UI/frontend files touched). Validated with a mix of live functional testing against the local environment (cookie manipulation via curl, real login/logout flows, on-disk cache inspection) and review of the new unit tests for the branch coverage.

|---|---|---|
| Per-user cache bucket cannot be selected using only a self-chosen cookie value | API | ✅ |
| Cookie with a username segment but missing/invalid companion validation is served the anonymous cached page (not an error, not another user's content) | API | ✅ |
| Genuine, valid logged-in session continues to get the correct per-user cached page (no regression) | API | ✅ |
| Existing on-disk cache naming unaffected; `rocket_clean_user()` and purge logic still work | API | ✅ |
| Shared logged-in cache branch (`common_cache_logged_users` filter) is gated by the same validation | Analysis (unit tests + code trace) | ✅ |

**What was tested:**
- Sent a request with a manually-set `wordpress_logged_in_*` cookie carrying a username segment but no companion validation cookie at all — the response matched the anonymous cached page byte-for-byte (same `Last-Modified`), and no new per-user cache directory was created.
- Repeated with the same cookie plus a companion cookie set to an arbitrary, non-matching value — same result: anonymous page served, no error, no new directory.
- Performed a real login through `wp-login.php` and confirmed the companion cookie is issued alongside the native logged-in cookie at login, and cleared (expired, `Max-Age=0`) alongside it at logout.
- Fetched the homepage with the real session cookies and confirmed a new cache directory following the existing `host-username-secret` naming convention was created and served correctly.
- Ran `rocket_clean_user()` against that user and confirmed the per-user cache directory was correctly located and removed, with no errors.
- Reviewed the new unit test suite (`getRequestCachePath.php`, `setUserCacheCookie.php`, `clearUserCacheCookie.php`) — good coverage of valid/missing/wrong-secret/wrong-username/expired/tampered-expiration cases for both the per-user and shared logged-in-cache branches.

No blocking guards (license/subscription/live-site checks) affect this code path — cache bucket selection runs independently of those gates.

**Smoke tests:** WP Rocket settings page and wp-admin dashboard both loaded normally (HTTP 200) after the change.

tests/Unit/inc/Engine/Cache/UserCacheKeySubscriber/clearUserCacheCookie.php
hellofromahmed: Fixed in 4eff6a7ed.

hellofromahmed:
> [!NOTE]
> Two minor follow-up considerations surfaced during review — flagging here for the team to decide on, not blocking:
>
> 1. **Cookie filter argument:** `UserCacheKeySubscriber::set_user_cache_cookie()` passes `is_ssl()` directly as the third argument to the `secure_logged_in_cookie` filter, whereas WordPress core's `wp_set_auth_cookie()` passes its already-filtered `$secure` value at the equivalent point. Functionally identical in the default case; only relevant if a plugin hooks that specific filter and expects the pre-filtered core value.
> 2. **Key normalization consistency:** the companion cookie is signed using the raw `secret_cache_key` option value, while validation compares against the `sanitize_key()`-normalized copy already baked into the cache config file. These always match today (the value is auto-generated lowercase hex and isn't exposed as an editable field), but the two call sites rely on that being true implicitly rather than by construction — applying the same normalization on both sides would remove the implicit dependency.

Happy to pick either up in a follow-up if the team decides it's worth it.

